PR #5, opened 2018-02-09
only deploy on master

Inveracity:
dont want to deploy on every push, same as the website

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -1,0 +1,21 @@
+import os
+
+import requests
+
+
+branch = os.environ.get("TRAVIS_BRANCH")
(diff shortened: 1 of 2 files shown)


deploy.py
```diff
@@ -0,0 +1,21 @@
+import os
+
+import requests
+
+
+branch = os.environ.get("TRAVIS_BRANCH")
+url = os.environ.get("AUTODEPLOY_WEBHOOK")
+token = os.environ.get("AUTODEPLOY_TOKEN")
+
```
triacle: this blank line is a bit curious to me

AkiraSama:
can't see a problem with that

PR #9, opened 2018-02-16
Prefix specific commands

mental32:
Added a check to allow commands that are advertised to target certain objects to actually work.

Previously any prefix that was advertised for certain objects was unrestricted in accessing commands intended to be accessed with another prefix. This, of course, triggered me so I bugged GDUDE about it and he told me to make a pull request instead.

Also, re worked the logic for the 'with_role' check. This was the only thing not tested so keep it or not I couldn't care less :]

commit desc:
Added the check 'invoke_with_prefix' to allow for prefix specific commands and I spruced up the logic in 'with_role'


bot/decorators.py
```diff
@@ -3,15 +3,21 @@
 from discord.ext.commands import Context
 
 
+def invoke_with_prefix(prefix: str) -> commands.check:
+    '''Allow a command to only be invoked with one prefix
```
shadycake: We use double quotes instead of single quotes, consistency would be nice :P

bot/decorators.py
```diff
@@ -3,15 +3,21 @@
...
 def with_role(*role_ids: int):
     async def predicate(ctx: Context):
         if not ctx.guild:  # Return False in a DM
             return False
 
-        for role in ctx.author.roles:
-            if role.id in role_ids:
-                return True
-        return False
+        return any((role.id in role_ids) for role in ctx.author.roles)
```
shadycake: The logic is the exact same, but the readability is lowered slightly. I'd suggest keeping it as it was before, but that's open to discussion.

bot/decorators.py
mental32: I'm consistent with PEP8 but it's ok, I can change the quotes.

bot/decorators.py
mental32: Sure, I don't mind.

JoeBanks13: quotes

Just going to leave this here...

bot/decorators.py
```diff
@@ -3,15 +3,21 @@
 from discord.ext.commands import Context
 
 
+def invoke_with_prefix(prefix: str) -> commands.check:
```
1mn: this doesnt fit all use cases. how about `ignore_prefix` and `allow_prefix` , and they both take lists.that gives us enough configurability to suit all cases without being terrible for readability.

bot/decorators.py
1mn: please do.

bot/decorators.py
1mn: readability is worse. the original logic is basically an english sentence. for these if this then that. yours reads if any of these are true and by these I mean that the id of the role is in this list and by role I mean the elements that are generated by this expression - it's all backwards. sequence is a critical part of how the brain parses information.

bot/decorators.py
mental32: by default, it'll ignore all other prefixes, I'll change it so it'll take a sequence of arguments considered prefixes and just apply a presence check

1mn:
After some deliberation, this approach is just too simplistic. we're gonna need the commands to display correctly on the help output, we want to return an AttributeError when you have the right command but the wrong namespace, we're gonna need a more comprehensive solution. So, thanks for the PR, we're probably gonna use this idea, but this isn't how.

PR #10, opened 2018-02-17
Clickup commands #z20f

gdude2002:


bot/cogs/clickup.py
```diff
@@ -0,0 +1,152 @@
...
+    @command(name="clickup.tasks()", aliases=["clickup.tasks", "tasks"])
+    @with_role(MODERATOR_ROLE, ADMIN_ROLE, OWNER_ROLE, DEVOPS_ROLE)
+    async def tasks(self, ctx: Context, list: int = None, status: str = None):
+        """
+        Get a list of tasks, optionally on a specific list or with a specific status
+        """
+
+        params = {}
+
+        if list:
```
1mn: naw, man. don't do it. you'll break my heart.

gdude2002: I'm wondering what else to call this, since it really is a clickup list that you're putting in

1mn: `task_list`? even `list_` would be better, but I don't think it needs to come to that.

gdude2002: I was just thinking about it because this will appear as a command argument in the help as well

1mn: `task_list` still seems fairly inoffensive to me. Isn't that what it is?

gdude2002: Hm, yeah, I suppose you're right about that

bot/cogs/clickup.py
```diff
@@ -0,0 +1,162 @@
+# coding=utf-8
+import asyncio
```
1mn: Do you need this import? Doesn't seem like you're using it.

gdude2002:
The following commands have been implemented:

* `tasks` AKA `clickup.tasks()` - Search for a list of tasks, optionally filtering by status and task list
* `task` AKA `clickup.task()` - Get detailed information about a specific task
* `team` AKA `clickup.team()` - Get a list of every member of the team
* `lists` AKA `clickup.lists()` - Get a list of every task list on every project
* `open` AKA `clickup.open()` - Open a new task on specific task list, with a title
* `set_status` AKA `clickup.set_status()` - Set the status on a specific task

bot/utils.py
```diff
@@ -45,3 +57,100 @@ def _convert_keys(self):
...
+    embed.description = paginator.pages[current_page]
+
+    message = await ctx.send(embed=embed)
+
+    if len(paginator.pages) <= 1:
+        return  # There's only one page
+
+    embed.set_footer(text=f"Page {current_page + 1}/{len(paginator.pages)}")
+
+    def event_check(reaction_: Reaction, user_: Member):
```
1mn: I'm not a fan of this function def being in the middle of the other function body. I'd prefer it right at the start..

bot/utils.py
```diff
@@ -45,3 +57,100 @@ def _convert_keys(self):
...
+            )
+        )
+
+    for emoji in PAGINATION_EMOJI:
+        # Add all the applicable emoji to the message
+        await message.add_reaction(emoji)
+
+    while True:
+        try:
+            reaction, user = await ctx.bot.wait_for("reaction_add", timeout=300, check=event_check)
```
1mn: maybe paginate can take the `timeout` as an arg with a default value of `300`? We already have 8000 arguments, what harm could another one do? :stuck_out_tongue:

1mn:
Very nice.

PR #11, opened 2018-02-19
Fixed formatter.py

shadycake:
image
(Command isn't changed in this branch, it was temporarily changed for testing purposes)

- fixed multiline docstrings
- changed full modulename.classname in type hints to only classname
- fixed the format of the arguments to do_<command>


bot/formatter.py
```diff
@@ -61,16 +61,29 @@ class <cog2>:
...
 
+            # indent every line in the help message
+            helptext = [f"    {line}" for line in self.command.help.split("\n")]
+            helptext = "\n".join(helptext)
+
             # prepare the different sections of the help output, and add them to the paginator
             definition = f"async def {stripped_command}{arguments}:"
-            docstring = f"    \"\"\"\n    {self.command.help}\n    \"\"\""
-            invocation = f"    await do_{stripped_command}({args_no_type_hints})"
+            docstring = f"    \"\"\"\n{helptext}\n    \"\"\""
```
1mn: readability for the `docstring` part sucks with all those escapes, and the listcomp for `helptext` is overkill. how about this:
```py
helptext = "\n    ".join(self.command.help.split())
docstring = (f'''
    """
    {helptext}
    """
''')
```

PR #16, opened 2018-02-26
Patched the grievous error we overlooked

AnonGuy:
• Past time must be subtracted from current time for cooldowns


1mn:
100 comments and we didn't catch this. what the hell is wrong with us.

AnonGuy:
Finish building, already. This is torture!

AnonGuy:
We did it boys

PR #18, opened 2018-02-27
Add cog for turning math expressions into LaTeX 

martmists:
Note that this does NOT require eval, and does NOT evaluate the expression, and is therefore 100% safe.


bot/__main__.py
```diff
@@ -34,5 +34,6 @@
 bot.load_extension("bot.cogs.fun")
 bot.load_extension("bot.cogs.eval")
 bot.load_extension("bot.cogs.verification")
+bot.load_extension("bot.cogs.math")
```
AnonGuy: Could this be on the previous line? Not a must-have, but these lines would certainly look nicer with ascending lengths :D

bot/cogs/math.py
```diff
@@ -0,0 +1,39 @@
...
+
+        else:
+            ltx = latex(parsed)
+            urlsafe = urlescape(ltx)
+
+            async with ClientSession() as session:
+                async with session.get(LATEX_URL + urlsafe) as resp:
+                    bytes_img = await resp.read()
+
+            await ctx.send(file=File(fp=BytesIO(bytes_img), filename="latex.png"))
```
AnonGuy: Might we have this `File(fp=BytesIO(bytes_img)` on a previous line. Looks a bit too condensed to me.

AnonGuy:
The current font that the LaTeX is rendered with is Helvetica. Any way to change the request to use the Latin Modern font? Seems to be the one used in most LaTeX rendered expressions

bot/cogs/math.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+
+from io import BytesIO
+from urllib.parse import urlescape
```
gdude2002: This doesn't exist

hargoniX:
There is https://en.wikibooks.org/wiki/LaTeX/Fonts

AnonGuy:
It looks like the font is passed directly in the URL as a query string: `fn_phv`. In my opinion, Latin Modern (`fn_lm`) looks much nicer.

bot/cogs/math.py
```diff
@@ -0,0 +1,39 @@
...
+from aiohttp import ClientSession
+
+from discord import File
+from discord.ext.commands import command
+
+from sympy import latex
+from sympy.parsing.sympy_parser import parse_expr
+
+
+LATEX_URL = "https://latex.codecogs.com/png.download?%5Cdpi%7B150%7D%20%5Cbg_white%20%5Cfn_phv%20"
```
AnonGuy: The return PNG from this base URL is much too small. Change it to something like `https://latex.codecogs.com/gif.latex?%5Cdpi%7B300%7D%20%5Cbg_white%20%5Clarge%20ABC`
e.g add `300` and `large` to the query string to make it readable.

martmists:
- [ ] cog order by lenght
- [ ] larger image and Latin Modern
- [ ] fix urlescape
- [ ] sympy in requirements.txt
- [ ] Shorten file send line

I will fix these issues tomorrow

bot/cogs/math.py
```diff
@@ -0,0 +1,41 @@
...
+from aiohttp import ClientSession
+
+from discord import File
+from discord.ext.commands import command
+
+from sympy import latex
+from sympy.parsing.sympy_parser import parse_expr
+
+
+LATEX_URL = "https://latex.codecogs.com/png.download?%5Cdpi%7B300%7D%20%5Cbg_white%20%5Chuge%20"
```
AnonGuy: Consider removing `bg_white` from the query string completely, to return an expression with a transparent background. You might also want to decrease that text size a little. `150` seems to work well in discord.

martmists: bg_white works better for users on dark or custom themes, I will change the text size though

bot/cogs/math.py
```diff
@@ -0,0 +1,41 @@
...
+from discord.ext.commands import command
+
+from sympy import latex
+from sympy.parsing.sympy_parser import parse_expr
+
+
+LATEX_URL = "https://latex.codecogs.com/png.download?%5Cdpi%7B150%7D%20%5Cbg_white%20%5Chuge%20"
+
+
+class MathCog:
```
gdude2002: Should be named "Math"

Should have an `__init__()` that takes a `bot` object and stores it under `self

bot/cogs/math.py
```diff
@@ -0,0 +1,41 @@
...
+            ltx = latex(parsed)
+            urlsafe = quote(ltx)
+
+            async with ClientSession() as session:
+                async with session.get(LATEX_URL + urlsafe) as resp:
+                    bytes_img = await resp.read()
+
+            file = File(fp=BytesIO(bytes_img), filename="latex.png")
+
+            await ctx.send(file=file)
```
gdude2002: No `setup(bot)` function

martmists:
`__init__` is not actually required, but I was requested to add it for consistency

bot/cogs/math.py
```diff
@@ -0,0 +1,48 @@
...
+
+LATEX_URL = "https://latex.codecogs.com/png.download?%5Cdpi%7B150%7D%20%5Cbg_white%20%5Chuge%20"
+
+
+class Math:
+    def __init__(self, bot):
+        self.bot = bot
+
+    @command()
+    async def latexify(self, ctx, expr: str):
```
gdude2002: Could shove a splat before `expr` so that the expression can contain spaces?

AnonGuy:
Not a must-have, but consider using a POST request to http://latex2png.com/ with the following form data, where `latex` is the parsed LaTeX expression:
image
A HTML colour of `808080` guarantees that the transparent image can be viewed on both themes, and you'll find the image URL here:
image
Again, not exactly necessary, but we'll get a much nicer transparent image:
image

AnonGuy:
This is looking great so far! A couple issues with the discord width limit, but the images look good.

martmists:
Discord width limit is something I cannot change, unless any of you have ideas

bot/cogs/math.py
```diff
@@ -0,0 +1,96 @@
...
+            file = File(fp=BytesIO(bytes_img), filename="latex.png")
+
+            await ctx.send(file=file)
+
+    @command()
+    async def calc(self, ctx, *, expr: str):
+        fixed_expr = expr.replace('^', '**').strip('`')
+
+        if any(x in fixed_expr for x in ("**", "__")):
+            await ctx.send(
```
gdude2002: I would prefer
```py
return await ctx.send(
    ...
)
```

martmists:
There is no need to return though, and either none or all branches should return if we want a consistent style.

gdude2002:
Well, you _have a return there_ on the next line, so..

martmists:
Oh really? My bad then

gdude2002:
Unfortunately, the merge conflicts are too complex for GitHub to handle - I trust I can leave that with you?

martmists:
How do I fix them?

gdude2002:
Huh, you don't have a mergetool?

You'll have to merge master into your branch and deal with the conflicts as usual.

martmists:
I never had any issues merging so I wouldn't really know how ^^"

PR #19, opened 2018-02-28
Cog management #1404e

gdude2002:


bot/cogs/cogs.py
```diff
@@ -0,0 +1,196 @@
...
+        if cog in self.cogs:
+            full_cog = self.cogs[cog]
+        elif "." in cog:
+            full_cog = cog
+        else:
+            full_cog = None
+            embed.description = f"Unknown cog: {cog}"
+
+        if full_cog == "bot.cogs.cogs":
+            embed.description = "You may not reload the cog management cog!"
```
1mn: this will never happen. The cog can't be unloaded, and if it was, you couldn't do this anyway cause it'd be unloaded.

bot/cogs/cogs.py
```diff
@@ -0,0 +1,196 @@
...
+                    cog = filename[:-3]
+
+                    self.cogs[cog] = f"bot.cogs.{cog}"
+
+        # Allow reverse lookups by reversing the pairs
+        self.cogs.update({v: k for k, v in self.cogs.items()})
+
+    @command(name="cogs.load()", aliases=["cogs.load", "load_cog"])
+    @with_role(MODERATOR_ROLE, ADMIN_ROLE, OWNER_ROLE, DEVOPS_ROLE)
+    async def load_command(self, ctx: Context, cog: str):
```
1mn: missing docstrings.

bot/cogs/cogs.py
```diff
@@ -0,0 +1,196 @@
...
+                cog = self.cogs[cog]
+
+            if loaded:
+                chevron = GREEN_CHEVRON
+            else:
+                chevron = RED_CHEVRON
+
+            lines.append(f"{chevron}  {cog}")
+
+        await paginate(lines, ctx, embed, max_size=300, empty=False)
```
1mn: max_size seems a bit large. maybe 200 or even 150 is better. most cogs have very short names, and we don't want a list of 50 cogs with 6 letters average names.

1mn: (to be honest, paginate really should be able to set a max-lines... but that's for a different PR.)

gdude2002: 300 is set because of the emoji - each emoji is always 34 chars on its own.

I was actually kinda thinking 300 wasn't enough, but it's probably fine.

bot/cogs/cogs.py
gdude2002: Got this one already

bot/cogs/cogs.py
gdude2002: Whoops, yep, I'd better get on that

bot/cogs/cogs.py
1mn: ahh, that makes sense.
okay, carry on then!

bot/cogs/cogs.py
```diff
@@ -0,0 +1,194 @@
...
+    async def load_command(self, ctx: Context, cog: str):
+        cog = cog.lower()
+
+        embed = Embed()
+        embed.colour = Colour.red()
+
+        embed.set_author(
+            name="Python Bot (Cogs)",
+            url="https://github.com/discord-python/bot",
+            icon_url="https://raw.githubusercontent.com/discord-python/branding/master/logos/logo_circle.png"
```
AnonGuy: Will these icons work with a shortened url? I was thinking use something like git.io for the particularly long links.
We could maybe add the urls to our constants?

bot/cogs/cogs.py
```diff
@@ -0,0 +1,194 @@
...
+        if full_cog not in self.bot.extensions:
+            try:
+                self.bot.load_extension(full_cog)
+            except ClientException:
+                embed.description = f"Invalid cog: {cog}\n\nCog does not have a `setup()` function"
+            except ImportError:
+                embed.description = f"Invalid cog: {cog}\n\nCould not find cog module {full_cog}"
+            except Exception as e:
+                embed.description = f"Failed to load cog: {cog}\n\n```{e}```"
+            else:
```
AnonGuy: Any reason why the "Cog loaded" message isn't under the `try` clause?

bot/cogs/cogs.py
```diff
@@ -0,0 +1,194 @@
...
+            if key in self.bot.extensions:
+                cogs[key] = True
+            else:
+                cogs[key] = False
+
+        for key in self.bot.extensions.keys():
+            if key in self.cogs:
+                continue
+
+            cogs[key] = True
```
AnonGuy: Could we have this line in the `if` clause, and change it to a `if key not in self.cogs`?

bot/cogs/cogs.py
gdude2002: Because we're not trying it - that should fail hard.

bot/cogs/cogs.py
```diff
@@ -0,0 +1,218 @@
+# coding=utf-8
+import os
+
+from discord import ClientException, Colour, Embed
+from discord.ext.commands import AutoShardedBot, Context, command
+
+from bot.constants import (
+    ADMIN_ROLE, BOT_AVATAR_URL, DEVOPS_ROLE, GITHUB_URL_BOT, GREEN_CHEVRON, MODERATOR_ROLE, OWNER_ROLE, RED_CHEVRON
```
1mn: this line is long and it makes me uncomfortable.

bot/cogs/cogs.py
```diff
@@ -0,0 +1,218 @@
...
+                    cog = filename[:-3]
+
+                    self.cogs[cog] = f"bot.cogs.{cog}"
+
+        # Allow reverse lookups by reversing the pairs
+        self.cogs.update({v: k for k, v in self.cogs.items()})
+
+    @command(name="cogs.load()", aliases=["cogs.load", "load_cog"])
+    @with_role(MODERATOR_ROLE, ADMIN_ROLE, OWNER_ROLE, DEVOPS_ROLE)
+    async def load_command(self, ctx: Context, cog: str):
```
1mn: do the methods really have to be called x_command, can't they just all be called x? it's clear that they are commands from the decorator.

bot/cogs/cogs.py
```diff
@@ -0,0 +1,218 @@
...
+        else:
+            embed.description = f"Cog {cog} is not loaded"
+
+        await ctx.send(embed=embed)
+
+    @command(name="cogs.reload()", aliases=["cogs.reload", "reload_cog"])
+    @with_role(MODERATOR_ROLE, ADMIN_ROLE, OWNER_ROLE, DEVOPS_ROLE)
+    async def reload_command(self, ctx: Context, cog: str):
+        """
+        Reload up an unloaded cog, given the module containing it
```
1mn: this docstring is on acid

bot/cogs/cogs.py
1mn: wait, this one actually could happen. it's the load_cog one that couldn't.  anyway it seems fixed now.

bot/cogs/cogs.py
gdude2002: I started doing that with ClickUp because I was shadowing builtins and reserved words

1mn: okay. then I suppose I should do the same in tags?

gdude2002: I'm not super worried about it, but I guess it's a good habit

PR #21, opened 2018-03-01
Tag feature

1mn:
image


hargoniX:
Soo its seems snekchek will try to run all linters if there is no config file existent, that is what causes CI to fail I guess, let me try to fix (Mart will not publish to pypi today anymore I guess)
Also why do you still have snekchek in CI if you deleted .snekrc?

martmists:
Why did you remove .snekrc if we're using snekchek for flake8 and soon pytest? If you want Travis to pass, simply set .travis.yml to use flake8 instead, but I'm confused as to why.

bot/cogs/clickup.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, bot: AutoShardedBot):
         self.lists = CaseInsensitiveDict()
 
     async def on_ready(self):
-        with ClientSession() as session:
+        async with ClientSession() as session:
```
1mn: These are out of scope. Should I remove them?

1mn: they are probably improvements, but maybe it'd be better if gdude made the changes to this file.

1mn: For the record, I did these so I could get aiohttp3 to work, but then later realized I should be using 2.5 with this version of discord.py rewrite.

1mn: eh, I'll remove them.

bot/cogs/tags.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+class Tags:
+    """
+    Save new tags and fetch existing tags.
+    """
+
+    def __init__(self, bot: AutoShardedBot):
+        self.bot = bot
+        self.api_key = os.environ.get("BOT_API_KEY")
```
gdude2002: `constants.SITE_API_KEY`

bot/cogs/tags.py
```diff
@@ -0,0 +1,266 @@
...
+        self.tag_cooldowns = {}
+
+    async def get_tag_data(self, tag_name: Union[str, None] = None) -> dict:
```
gdude2002: You can replace `Union[anything, None]` with `Optional[anything]`

bot/cogs/tags.py
```diff
@@ -0,0 +1,266 @@
...
+    async def info(self, ctx: Context):
+        """
+        Show available methods for this class.
+
+        :param ctx: Discord message context
+        """
+
+        return await ctx.invoke(self.bot.get_command("help"), "Tags")
+
+    @command(name="tags.get()", aliases=["tags.get", "tags.show()", "tags.show"])
```
gdude2002: So far, I've been adding command aliases that make sense without the namespace, but that's entirely up to you I think. Like, ClickUp has `open_task` for example.

bot/cogs/tags.py
```diff
@@ -0,0 +1,266 @@
...
+    async def get(self, ctx: Context, tag_name: str = None):
+        """
+        Get a list of all tags or a specified tag.
+
+        :param ctx: Discord message context
```
gdude2002: Is this a thing we're doing in command docstrings? It's a good idea, I think

bot/cogs/tags.py
```diff
@@ -0,0 +1,266 @@
...
+        :param tag_name: The name of the tag to create or edit.
+        :param tag_content: The content of the tag.
+        """
+
+        embed = Embed()
+
+        if "\n" in tag_name:
+            embed.colour = Colour.red()
+            embed.title = "Please don't do that"
+            embed.description = "Don't be ridiculous. Newlines are obviously not allowed in the tag name."
```
gdude2002: 👍

bot/cogs/tags.py
```diff
@@ -0,0 +1,266 @@
...
+
+        else:
+            if tag_name and tag_content:
```
gdude2002: This could be reorganised, for example since the `else` here contains a return, you could invert the conditional and drop the extra branch.

```py
if not (tag_name and tag_content):
    # ....
    return

tag_data = ...
```

bot/cogs/tags.py
```diff
@@ -0,0 +1,266 @@
...
+                embed.title = "Missing parameters!",
+                embed.description = "The tag needs both a name and some content"
+                return await ctx.send(embed=embed)
+
+            if tag_data.get("success"):
+                embed.colour = Colour.blurple()
+                embed.title = "Tag successfully added!"
+                embed.description = f"**{tag_name}** added to tag database."
+            else:
+                print(f"Something terrible happened. The API returned {tag_data}")
```
gdude2002: It's bad UX to just return the json - it should be clear what an error state will give us, and we should be able to provide a useful error message

bot/cogs/tags.py
```diff
@@ -0,0 +1,266 @@
...
+        """
+        Remove a tag from the database.
+
+        :param ctx: discord message context
+        :param tag_name: The name of the tag to delete.
+        """
+
+        embed = Embed()
+
+        if tag_name:
```
gdude2002: Again, this conditional could be inverted in order to drop a branch

bot/cogs/tags.py
1mn: I don't see why not, but I think it was mostly just something pycharm did :P

gdude2002: It's basically Sphinx markup - the sphinx document generator uses it when it's generating documentation from your code.

I think it is a good idea to explain the params to the users as well, though.

bot/cogs/tags.py
1mn: yeah this was more for debugging. forgot I left it in.

bot/cogs/tags.py
```diff
@@ -0,0 +1,266 @@
...
+            embed.title = "Please don't do that"
+            embed.description = "Don't be ridiculous. Newlines are obviously not allowed in the tag name."
+
+        else:
+            if tag_name and tag_content:
+                tag_name = tag_name.lower()
+                tag_data = await self.post_tag_data(tag_name, tag_content)
+
+            else:
+                embed.colour = Colour.red(),
```
1mn: lol wtf is this comma

bot/__main__.py
```diff
@@ -34,5 +34,6 @@
 bot.load_extension("bot.cogs.fun")
 bot.load_extension("bot.cogs.eval")
 bot.load_extension("bot.cogs.verification")
+bot.load_extension("bot.cogs.tags")
```
AnonGuy: Can these be in alphabetical size order pls

bot/cogs/tags.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        Show available methods for this class.
+
+        :param ctx: Discord message context
+        """
+
+        return await ctx.invoke(self.bot.get_command("help"), "Tags")
+
+    @command(name="tags.get()", aliases=["tags.get", "tags.show()", "tags.show", "get_tag"])
+    async def get_command(self, ctx: Context, tag_name: str = None):
```
AnonGuy: Why isn't `tag_name` type hinted as an `Optional[str]`?

bot/cogs/tags.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        Get a list of all tags or a specified tag.
+
+        :param ctx: Discord message context
+        :param tag_name:
+        If provided, this function shows data for that specific tag.
+        If not provided, this function shows the caller a list of all tags.
+        """
+
+        def _command_on_cooldown(tag_name: Optional[str]) -> bool:
```
AnonGuy: Why is `tag_name` type hinted as an `Optional`? It doesn't have a default value.

bot/cogs/tags.py
```diff
@@ -0,0 +1,265 @@
...
+            """
+
+            now = time.time()
+
+            cooldown_conditions = (
+                tag_name
+                and tag_name in self.tag_cooldowns
+                and (now - self.tag_cooldowns[tag_name]["time"]) < TAG_COOLDOWN
+                and self.tag_cooldowns[tag_name]["channel"] == ctx.channel.id
+            )
```
AnonGuy: This is :ok_hand:. Will write my conditions like this from now on.

bot/cogs/tags.py
```diff
@@ -0,0 +1,265 @@
...
+class Tags:
+    """
+    Save new tags and fetch existing tags.
+    """
+
+    def __init__(self, bot: AutoShardedBot):
+        self.bot = bot
+        self.tag_cooldowns = {}
+
+    async def get_tag_data(self, tag_name: Optional[str] = None) -> dict:
```
AnonGuy: should there be spaces around the `=` in this keyword assignment?

bot/cogs/tags.py
AnonGuy: Also shouldn't keyword assignments have no spaces around the `=`?

bot/cogs/tags.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        Retrieve the tag_data from our API
+
+        :param tag_name: the tag to retrieve
+        :return:
+        if tag_name was provided, returns a dict with tag data.
+        if not, returns a list of dicts with all tag data.
+
+        """
+        headers = {"X-API-KEY": SITE_API_KEY}
```
AnonGuy: You're assigning this headers dict with the same value in 3 different functions.
Maybe move it outside of the function scope?

bot/__main__.py
```diff
@@ -35,5 +35,6 @@
 bot.load_extension("bot.cogs.eval")
 bot.load_extension("bot.cogs.fun")
 bot.load_extension("bot.cogs.verification")
+bot.load_extension("bot.cogs.tags")
```
AnonGuy: Can these be in size order pl0x

bot/cogs/tags.py
```diff
@@ -0,0 +1,265 @@
...
+
+            tag_name = tag_name.lower()
+            tag_data = await self.post_tag_data(tag_name, tag_content)
+
+            if tag_data.get("success"):
+                embed.colour = Colour.blurple()
+                embed.title = "Tag successfully added!"
+                embed.description = f"**{tag_name}** added to tag database."
+            else:
+                embed.title = "Database error",
```
AnonGuy: Some of the embed titles have punctuation but some don't. It might be more consistent to add periods to these?

bot/cogs/tags.py
```diff
@@ -0,0 +1,265 @@
...
+        embed.colour = Colour.red()
+
+        if "\n" in tag_name:
+            embed.title = "Please don't do that"
+            embed.description = "Don't be ridiculous. Newlines are obviously not allowed in the tag name."
+
+        else:
+            if not (tag_name and tag_content):
+                embed.title = "Missing parameters!"
+                embed.description = "The tag needs both a name and some content"
```
AnonGuy: Can we have a period at the end of this description pl0x

PR #22, opened 2018-03-03
More codeblock hints

hargoniX:
It's me again bringing you more codeblock hints.


mental32:
Just a suggestion but what do you think of the bot deleting a users message and reposting the code in markdown. This way it wouldn't be:
- code-blockless code
- embed with code-block on how to use code-blocks

(assuming the user gets it right the first time) 
- code reposted in a code-block. 

Instead, it would just swap out the users code by deleting it and reposting it in a code block.

I realize this seems counter-intuitive as to teaching users about markdown code blocks my solution to that would be DM'ing the user instructions on markdown or something

hargoniX:
We already discussed that in my first pr but (actually that was  the first implementation from me) Joseph decided that I should change it to markdown hint

1mn:
I'll review this properly later.

As for Mentals suggestion, we're not doing that. there are a number of potential problems with deleting users messages and reposting them, and the bot can't edit the users message. The main problem is that users might not like it. The second is that they lose the ability to edit (which is commonly done), and the third is that if the bot makes a mistake it will be very rude indeed. It's simply too destructive to have the bot automatically detect and delete messages. It's just a matter of time until it starts deleting users, servers, and then before you know it we're all dead.

PR #23, opened 2018-03-03
Improving autocodeblocks

hargoniX:
It's me again bringing more auto codeblocks


martmists:
Why are you using msg.content rather than a separate string, if I may ask?

hargoniX:
Why should i use a seperate string instead of msg.content?

bot/cogs/bot.py
```diff
@@ -78,32 +80,35 @@ def __init__(self, bot: AutoShardedBot):
...
-                        # Some users formatted multi-line code using `'sip("
-                        # This fixes that by treating them as nothing.
-                        if "`" in msg.content:
-                            msg.content = msg.content[:-1][1:]
-                        tree = ast.parse(msg.content)
+                        # Filtering valid Python codeblocks and exiting if a valid Python codeblock is found
+                        if re.match(msg.content, "```(python|py)\n((?:.*\n*)+)```", re.IGNORECASE):
+                            return
+                        else:
+                            # Format the block into python source
```
1mn: this comment is kinda unclear. just say that you're stripping backticks from every line of the message.

bot/cogs/bot.py
```diff
@@ -78,32 +80,35 @@ def __init__(self, bot: AutoShardedBot):
...
+                            content = ""
+                            for line in msg.content.split("\n"):
```
1mn: use `.splitlines()`

bot/cogs/bot.py
```diff
@@ -78,32 +80,35 @@ def __init__(self, bot: AutoShardedBot):
             if time.time()-self.previous_format_times[msg.channel.id] > 300 or msg.channel.id == DEVTEST_CHANNEL:
                 if msg.content.count("\n") >= 3:
                     try:
-                        # Some users formatted multi-line code using `'sip("
-                        # This fixes that by treating them as nothing.
-                        if "`" in msg.content:
-                            msg.content = msg.content[:-1][1:]
-                        tree = ast.parse(msg.content)
+                        # Filtering valid Python codeblocks and exiting if a valid Python codeblock is found
+                        if re.match(msg.content, "```(python|py)\n((?:.*\n*)+)```", re.IGNORECASE):
```
1mn: `re.match` is too strict for this. If there's a valid python codeblock anywhere in the message, we want to ignore it. Use `re.search` instead, since `re.match` only matches from the beginning of the line.

bot/cogs/bot.py
```diff
@@ -78,32 +80,35 @@ def __init__(self, bot: AutoShardedBot):
...
+                            return
+                        else:
+                            # Format the block into python source
+                            content = ""
+                            for line in msg.content.split("\n"):
+                                content += line.strip("`") + "\n"
+                            # Remove "Python" or "Py" from top of the message if exists
```
1mn: blocks like this one should have a blank line above them, or the file starts to feel too dense. it's like a book without paragraphs, you know?

bot/cogs/bot.py
```diff
@@ -78,32 +80,35 @@ def __init__(self, bot: AutoShardedBot):
...
+                            if content.startswith("python") or content.startswith("Python"):
```
1mn: `if content.lower().startswith("python")`

bot/cogs/bot.py
```diff
@@ -78,32 +80,35 @@ def __init__(self, bot: AutoShardedBot):
...
+                                content = content[6:]
+                            elif content.startswith("py") or content.startswith("Py"):
+                                content = content[2:]
+                            # Strip again to remove the whitespace(s) left if the msg looks like
+                            # " Python"
```
1mn: this comment doesn't make sense considering you've just removed "python" from content if it exists. The whitespace you're removing is the one _after_ python but before the code if the user had typed this:

\`\`\`
python if condition:
\`\`\`

bot/cogs/bot.py
```diff
@@ -78,32 +80,35 @@ def __init__(self, bot: AutoShardedBot):
...
-                                     "print(\"Hello world!\")"
-                                     "```"
-                            )  # noqa. E124
-                            information = Embed(title="Code formatting", description=howto)
-                            await msg.channel.send(embed=information)
+                        # Multiple lines of single words could be interpreted as expressions.
+                        # This check is to avoid all nodes being parsed as expressions.
+                        # (e.g. words over multiple lines)
+                        if not all(isinstance(node, ast.Expr) for node in tree.body):
+                            header = (f"Hey {msg.author.mention} I noticed you tried to paste code. ",
```
1mn: this message needs work. my suggestion:

```py
"Hey {msg.author.mention}! I noticed you were trying to paste code into this channel." 
"Discord supports something called Markdown, which allows you to make beautiful "
"code blocks with Python syntax highlighting! Here's how they work:"
get_tag_data("codeblock")
```

bot/cogs/bot.py
```diff
@@ -75,40 +75,54 @@ def __init__(self, bot: AutoShardedBot):
...
+                    content = content[6:]
+                elif content.lower().startswith("py"):
+                    content = content[2:]
+                # Strip again to remove the whitespace(s) left before the code
+                # If the msg looked like "Python <code>" before removing Python
+                content = content.strip()
+                return content
+
     async def on_message(self, msg: Message):
         if msg.channel.id in self.previous_format_times:
```
1mn: Can you change this variable name? `previous_format_times` doesn't seem to make sense anymore. format? what format. what did we format?

bot/cogs/bot.py
```diff
@@ -75,40 +75,54 @@ def __init__(self, bot: AutoShardedBot):
...
             if time.time()-self.previous_format_times[msg.channel.id] > 300 or msg.channel.id == DEVTEST_CHANNEL:
```
1mn: This line is far too dense. Can you do the first expression on the line before and name it something logical? That way this line could just be something like 
```py
if not on_cooldown or msg.channel.id == DEVTEST_CHANNEL
```

bot/cogs/bot.py
```diff
@@ -75,40 +75,54 @@ def __init__(self, bot: AutoShardedBot):
...
                         # Invalid Python code will raise a SyntaxError.
+                        content = self.codeblock_stripping(msg.content)
+                        if not content:
+                            return
+
                         tree = ast.parse(content)
 
                         # Multiple lines of single words could be interpreted as expressions.
                         # This check is to avoid all nodes being parsed as expressions.
                         # (e.g. words over multiple lines)
```
1mn: These block comments are :ok_hand:

bot/cogs/bot.py
```diff
@@ -75,40 +75,54 @@ def __init__(self, bot: AutoShardedBot):
...
                         if not all(isinstance(node, ast.Expr) for node in tree.body):
-                            header = (f"Hey {msg.author.mention} I noticed you tried to paste code. ",
-                                      "Here's how you should have done it:")
-                            information = Embed(title="Codeblocks", description=get_tag_data("codeblock"))
-                            await msg.channel.send(header, embed=information)
+                            howto = (f"Hey {msg.author.mention} I noticed you tried to paste code. ",
```
1mn: You didn't change this message yet, and it's still neither grammatically correct nor expressed in the right tone. The bot should have a whimsical yet polite tone. This is missing a period, as well.  Please see my earlier comment that was dismissed by gdude even though it wasn't addressed :stuck_out_tongue:

bot/cogs/bot.py
```diff
@@ -103,8 +103,9 @@ def codeblock_stripping(self, msg: str):
                 return content
 
     async def on_message(self, msg: Message):
-        if msg.channel.id in self.previous_format_times:
-            if time.time()-self.previous_format_times[msg.channel.id] > 300 or msg.channel.id == DEVTEST_CHANNEL:
+        if msg.channel.id in self.previous_hint_times:
+            on_cooldown = time.time()-self.previous_hint_times[msg.channel.id] > 300
```
1mn: that doesn't make sense, if `time.time()-self.previous_hint_times[msg.channel.id] > 300` is True then this is **not** on cooldown. you should flip the logic in this expression so that the result is False if it's not cooling down, then use a `not` in the next line like in my example. Also, `-` should be surrounded by spaces.

bot/cogs/bot.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, bot: AutoShardedBot):
         self.bot = bot
 
         # Stores allowed channels plus unix timestamp from last call
-        self.previous_format_times = {HELP1_CHANNEL: 0,
+        self.previous_hint_times = {HELP1_CHANNEL: 0,
```
1mn: I still hate this variable name. what hint? we don't really use the word hint in any variable names or comments. how about `self.channel_cooldowns` or something so we're consistent. Also I'm not sure about that, and if you can come up with something better then use that instead.

bot/cogs/bot.py
```diff
@@ -22,7 +24,7 @@ def __init__(self, bot: AutoShardedBot):
         self.bot = bot
 
         # Stores allowed channels plus unix timestamp from last call
-        self.previous_format_times = {HELP1_CHANNEL: 0,
+        self.channel_cooldowns = {HELP1_CHANNEL: 0,
                                       HELP2_CHANNEL: 0,
```
1mn: :scream:

hargoniX: I am as shocked as you are

hargoniX:
Tested it, everything works as it should (at least when not using the get_tag func cause i can not really test it with that atm)

bot/cogs/bot.py
```diff
@@ -73,38 +74,60 @@ def __init__(self, bot: AutoShardedBot):
...
+
+                        # Multiple lines of single words could be interpreted as expressions.
+                        # This check is to avoid all nodes being parsed as expressions.
+                        # (e.g. words over multiple lines)
+                        if not all(isinstance(node, ast.Expr) for node in tree.body):
+                            codeblock_tag = await self.bot.get_cog("Tags").get_tag_data("codeblock")
+                            howto = (f"Hey {msg.author.mention}!\n I noticed you were trying to paste code into this "
+                                     "channel. Discord supports something called Markdown, which allows you to make "
+                                     "beautiful code blocks with Python syntax highlighting!\n"
+                                     f"{codeblock_tag['tag_content']}")
```
1mn: Actually, when you do these implicit joins you don't need f's on more than the very first string. The others will get implicit f's when they join the first one. so this f does nothing.

1mn: nvm I'm stupid.

bot/cogs/bot.py
```diff
@@ -73,38 +74,61 @@ def __init__(self, bot: AutoShardedBot):
...
+    def codeblock_stripping(self, msg: str):
+        """
+        Strip msg in order to find Python code.
+
+        Tries to strip out Python code out of msg and returns the stripped block or
+        None if the block is a valid Python codeblock.
+        """
+        if msg.count("\n") >= 3:
+            # Filtering valid Python codeblocks and exiting if a valid Python codeblock is found
+            if re.search(msg, "```(python|py)\n((?:.*\n*)+)```", re.IGNORECASE):
```
1mn: wrong order of args. it's (pattern, string), not the other way around.

bot/cogs/bot.py
```diff
@@ -73,38 +74,61 @@ def __init__(self, bot: AutoShardedBot):
...
-                            self.previous_format_times[msg.channel.id] = time.time()
+                        tree = ast.parse(content)
+
+                        # Multiple lines of single words could be interpreted as expressions.
+                        # This check is to avoid all nodes being parsed as expressions.
+                        # (e.g. words over multiple lines)
+                        if not all(isinstance(node, ast.Expr) for node in tree.body):
+                            codeblock_tag = await self.bot.get_cog("Tags").get_tag_data("codeblock")
+                            if codeblock_tag == {}:
+                                # todo: add logging
```
1mn: me and inver are gonna go through the bot and fix these very soon, so it's okay to leave them in.

PR #27, opened 2018-03-09
Aiohttp fixes

gdude2002:
Attempt to fix the bot responsiveness problem by moving aiohttp to an asyncio-based DNS resolver.

This also modifies all the cogs to use a single client session applied across all of the bot. And no, not the one d.py uses.


gdude2002:
We'll rewrite how the HTTP stuff is handled later on when we rewrite the cogs to use a subclass-based system

PR #34, opened 2018-03-15
Added logging to formatter.py

shadycake:
I wasn't entirely sure about how verbose I should be, if I need more or less logging, do not be afraid to say so!


shadycake:
I am adding some `log.trace()`s

shadycake:
Whenever I mention `self.command.cog_name` for command that isn't from a cog, it will simply say None, for example: `Help for help from None generated.` I am open to suggestions how I should go about this issue.

PR #40, opened 2018-04-02
Improve tag validation

kingdom5500:
it now (hopefully) replaces the whitespace from evaluating `bot.tags.get("\n\n\n\n\n\n")` and stuff like that.

### DIFF
--- a/bot/cogs/tags.py
+++ b/bot/cogs/tags.py
@@ -168,7 +168,7 @@ class Tags:
             embed.colour = Colour.red()
 
             # Replace all whitespace (Probably not the best place to put this though)
-            tag_data = tag_date.translate(
+            tag_data = tag_data.translate(
                 {
                     0x8: "\\b",
                     0x9: "\\t",
--- a/bot/cogs/tags.py
+++ b/bot/cogs/tags.py
@@ -142,7 +142,7 @@ class Tags:
             embed.description = "Tag names can't be numbers."
 
         else:
-            return
+            return None
 
         return embed
 
--- a/bot/cogs/tags.py
+++ b/bot/cogs/tags.py
@@ -1,5 +1,6 @@
 import logging
 import time
+from typing import Optional
 
 from discord import Colour, Embed, User
 from discord.ext.commands import AutoShardedBot, Context, command
@@ -85,7 +86,7 @@ class Tags:
         return tag_data
 
     @staticmethod
-    async def validate(author: User, tag_name: str, tag_content: str = None):
+    async def validate(author: User, tag_name: str, tag_content: str = None) -> Optional[Embed]:
         """
         Create an embed based on the validity of a tag's name and content
 
--- a/bot/cogs/tags.py
+++ b/bot/cogs/tags.py
@@ -85,15 +85,14 @@ class Tags:
         return tag_data
 
     @staticmethod
-    async def validate(embed: Embed, author: User, tag_name: str, tag_content: str = None) -> bool:
+    async def validate(author: User, tag_name: str, tag_content: str = None):
         """
-        Edit an embed based on the validity of a tag name
+        Create an embed based on the validity of a tag's content
 
-        :param embed: The embed to modify.
         :param author: The user that called the command
         :param tag_name: The name of the tag to validate.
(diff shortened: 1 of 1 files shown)


1mn:
Okay, so this is fine but I'd like a bit more from this PR. See how we're doing some validation for the set function? If we add validation to the get as well, we're doubling up validation and I feel like we should just pull validation out into a helper method for that class and call the validate method from both get and set.

Can you move all validation into a seperate method and implement it into both set and get, please? that means whitespace and newline checks as well as the is_number stuff, everything that's in the set. You can still use translate and your approach with all the control characters though, that's pretty good stuff. Make sure you comment it well so people know what the hell you're actually replacing.

bot/cogs/tags.py
```diff
@@ -84,6 +84,66 @@ def __init__(self, bot: AutoShardedBot):
 
         return tag_data
 
+    @staticmethod
+    async def validate(embed: Embed, author: User, tag_name: str, tag_content: str = None) -> bool:
+        """
+        Edit an embed based on the validity of a tag name
```
1mn: inaccurate. it also validates the `tag_content` a bit.

bot/cogs/tags.py
```diff
@@ -84,6 +84,66 @@ def __init__(self, bot: AutoShardedBot):
...
+
+        :param embed: The embed to modify.
+        :param author: The user that called the command
+        :param tag_name: The name of the tag to validate.
+        :param tag_content: The tag's content, if any.
```
1mn: that's not how you use `'`

bot/cogs/tags.py
```diff
@@ -84,6 +84,66 @@ def __init__(self, bot: AutoShardedBot):
...
+                0x8: "\\b",  # Backspace
+                0x9: "\\t",  # Horizontal tab
+                0xA: "\\n",  # Linefeed
+                0xB: "\\v",  # Vertical tab
+                0xC: "\\f",  # Form feed
+                0xD: "\\r"   # Carriage return
+            }
+        )
+
+        # 'tag_name' has any special characters
```
1mn: `'tag_name' has at least one special character.`

bot/cogs/tags.py
```diff
@@ -84,6 +84,66 @@ def __init__(self, bot: AutoShardedBot):
...
+        if tag_name != raw_name:
+            log.warning(f"{author} tried to put a special character in a tag name. "
+                        "Rejecting the request.")
+            embed.title = "Please don't do that"
+            embed.description = "Don't be ridiculous, special characters obviously aren't allowed in the tag name"
+
+        # 'tag_content' (when not None) or 'tag_name' consists of nothing but whitespace
```
1mn: `'tag_content' or 'tag_name' are either empty, or consist of nothing but whitespace`

bot/cogs/tags.py
```diff
@@ -84,6 +84,66 @@ def __init__(self, bot: AutoShardedBot):
...
+        elif is_number(tag_name):
+            log.error("inside the is_number")
+            log.warning(f"{author} tried to create a tag with a digit as its name. "
+                        "Rejecting the request.")
+            embed.title = "Please don't do that"
+            embed.description = "Tag names can't be numbers."
+
+        # No checks failed, return True (valid)
+        else:
+            return True
```
1mn: I don't like this because passing in stuff "by reference" and modifying it in your helper function is kinda implicit and C-ey. I'd prefer if you simply created and returned an embed instead.

bot/cogs/tags.py
```diff
@@ -84,6 +84,66 @@ def __init__(self, bot: AutoShardedBot):
...
+
+        return False
```
1mn: if `return True` returns an embed, this could simply `return None`, and you could still basically use the same method inside the commands.

bot/cogs/tags.py
```diff
@@ -84,6 +84,67 @@ def __init__(self, bot: AutoShardedBot):
 
         return tag_data
 
+    @staticmethod
+    async def validate(author: User, tag_name: str, tag_content: str = None):
```
1mn: return type hint should be `Optional[Embed]`

PR #47, opened 2018-04-12
Allow using double quotes within triple double quotes.

Volcyy:
`1nbze` on clickup: "modify the python syntax parser so that it will automatically escape any double quotes within a triple double quote block. currently it is necessary to manually escape these."

When using double quotes within a triple quote block, the Python syntax parser would previously error. This fixes that by escaping any double quotes:

img1
img2


1mn:
Great first PR. Even though it's just one line of code, you've solved the problem, provided a reference for the clickup ticket, and even added screenshots of it in action. I wish all PRs were this good.

PR #48, opened 2018-04-12
Allow `__getitem__` / `__setitem__`-like command access

Volcyy:
clickup task: `1aw8t` "Allow bot.tags['zen'] in addition to bot.tags.get('zen') and bot.tags['zen'] = 'superzen' in addition to bot.tags.set('zen', 'superzen')"

This allows users to use cogs containing a `get` and `set` command as they would use a dictionary:

- `bot.tags["ask"]` delegates to `bot.tags.get("ask")`
- `bot.tags["ask"] = "don't"` delegates to `bot.tags.set("ask", "don't")`

examples
more examples

### DIFF
--- a/bot/__init__.py
+++ b/bot/__init__.py
@@ -95,7 +95,7 @@ def _get_word(self) -> str:
     while not self.eof:
         try:
             current = self.buffer[self.index + pos]
-            if current.isspace() or current == "(":
+            if current.isspace() or current == "(" or current == "[":
                 break
             pos += 1
         except IndexError:
@@ -170,6 +170,37 @@ def _get_word(self) -> str:
         pos += 2
         result = self.buffer[self.previous:self.index + (pos+2)]
         self.index += 2
+
+    # Check if a command in the form of `bot.tags['ask'] {= 'whatever'}` was used
+    elif current == "[":
+        def clean_argument(arg: str) -> str:
+            return arg.strip("[]'\" ") \
+                      .replace("\"", "\\\"") \
+                      .replace("'", "\\'")
+
+        # Syntax is `bot.tags['ask']` => mimic `getattr`
+        if self.buffer.endswith("]"):
+            key = clean_argument(self.buffer[self.index:])
+            arg = f"\"{key}\""
+            result = self.buffer[self.previous:self.index] + ".get"
+
+        # Syntax is `bot.tags['ask'] = 'whatever'` => mimic `setattr`
+        elif "=" in self.buffer and not self.buffer.endswith("="):
+            equals_pos = self.buffer.find("=")
+            key = clean_argument(self.buffer[self.index:equals_pos])
+            value = clean_argument(self.buffer.split("=")[1])
+            result = self.buffer[self.previous:self.index] + ".set"
+            arg = f"\"{key}\" \"{value}\""
+
+        # Syntax is god knows what, pass it along
+        else:
+            result = self.buffer
+            arg = ''
+
+        self.buffer = f"{result} {arg}"
+        self.index = len(result)
+        self.end = len(self.buffer)
+
 
     if isinstance(result, str):
         return result.lower()  # Case insensitivity, baby
--- a/bot/__init__.py
+++ b/bot/__init__.py
@@ -177,8 +177,7 @@ def _get_word(self) -> str:
         def clean_argument(arg: str) -> str:
             """Helper function to remove any characters we don't care about."""
 
-            return arg.strip("[]'\" ") \
-                    .replace('"', '\\"')
+            return arg.strip("[]'\" ").replace('"', '\\"')
 
         log.trace(f"Got a command candidate for getitem / setitem mimick: {self.buffer}")
         # Syntax is `bot.tags['ask']` => mimic `getattr`
--- a/bot/__init__.py
+++ b/bot/__init__.py
@@ -181,6 +181,8 @@ def _get_word(self) -> str:
         # Syntax is `bot.tags['ask']` => mimic `getattr`
         if self.buffer.endswith("]"):
             key = clean_argument(self.buffer[self.index:])
+            # note: if not key, this corresponds to an empty argument
+            #       so this should throw / return a SyntaxError ?
             arg = f"\"{key}\""
             result = self.buffer[self.previous:self.index] + ".get"
 
@@ -193,6 +195,7 @@ def _get_word(self) -> str:
             arg = f"\"{key}\" \"{value}\""
 
         # Syntax is god knows what, pass it along
+        # in the future, this should probably return / throw SyntaxError
         else:
             result = self.buffer
             arg = ''
@@ -200,7 +203,6 @@ def _get_word(self) -> str:
         self.buffer = f"{result} {arg}"
         self.index = len(result)
         self.end = len(self.buffer)
-
 
     if isinstance(result, str):
         return result.lower()  # Case insensitivity, baby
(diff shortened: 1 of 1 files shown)


bot/__init__.py
```diff
@@ -168,6 +171,46 @@ def _get_word(self) -> str:
         result = self.buffer[self.previous:self.index + (pos+2)]
         self.index += 2
 
+    # Check if a command in the form of `bot.tags['ask'] {= 'whatever'}` was used
```
AnonGuy: What are these curly brackets for?

Volcyy: supposed to stand for "optional thing", since both `bot.tags['ask']` and `bot.tags['ask'] = 'whatever'` would work. I would've used `[...]`, but since that was already used there, I thought that was a bit unclear

bot/__init__.py
```diff
@@ -168,6 +171,46 @@ def _get_word(self) -> str:
...
+    elif current == "[":
+        # Helper function to remove any characters we don't care about
```
AnonGuy: Maybe this comment could be in a doc string for the function.

bot/__init__.py
```diff
@@ -168,6 +171,46 @@ def _get_word(self) -> str:
...
+        def clean_argument(arg: str) -> str:
+            return arg.strip("[]'\" ")
+
+        # Syntax is `bot.tags['ask']` => mimic `getattr`
+        if self.buffer.endswith("]"):
+            # Key: The first argument, specified `bot.tags[here]`
+            key = clean_argument(self.buffer[self.index:])
+            # note: if not key, this corresponds to an empty argument
+            #       so this should throw / return a SyntaxError ?
+            args = f"\"{key}\""
```
AnonGuy: It looks like `repr(key)` could come in handy instead of using a format string.

Volcyy: That would wrap the argument into single quotes. Doesn't work.

bot/__init__.py
AnonGuy: gotcha

bot/__init__.py
```diff
@@ -168,6 +171,46 @@ def _get_word(self) -> str:
...
+            result = self.buffer[self.previous:self.index] + ".get"
+
+        # Syntax is `bot.tags['ask'] = 'whatever'` => mimic `setattr`
+        elif "=" in self.buffer and not self.buffer.endswith("="):
+            equals_pos = self.buffer.find("=")
+            # Key: The first argument, specified `bot.tags[here]`
+            key = clean_argument(self.buffer[self.index:equals_pos])
+            # Value: The second argument, specified after the `=`
+            value = (clean_argument(self.buffer.split("=")[1])
+                     .replace("\"", "\\\"")  # escape any unescaped quotes
```
AnonGuy: The triple backslashes are a little unclear. Could we use single quotes for this line?
```py
value = (
    clean_argument(
        self.buffer.split("=")[1]
    )
    .replace('"', '\\"')  # escape any unescaped quotes
    .replace("'", "\\'")  # to mimic triple quote behaviour.
)
```

AnonGuy: This also aligns the comments :^)

bot/__init__.py
1mn: I'm not a fan, it's not obvious that they signify optional input. Just use more words to say it in a more obvious way please.

bot/__init__.py
```diff
@@ -168,6 +171,46 @@ def _get_word(self) -> str:
...
+                     .replace("'", "\\'"))   # to mimic triple quote behaviour.
+            # Use the cog's `set` command.
+            result = self.buffer[self.previous:self.index] + ".set"
+            args = f"\"{key}\" \"{value}\""
```
AnonGuy: Can `f"\"{key}\" \"{value}\""` be `f'"{key}" "{value}"'` so you won't have to escape any quotes

bot/__init__.py
```diff
@@ -168,6 +171,46 @@ def _get_word(self) -> str:
...
+        # Syntax is god knows what, pass it along
+        # in the future, this should probably return / throw SyntaxError
+        else:
+            result = self.buffer
+            args = ''
+
+        # Reconstruct valid discord.py syntax
+        self.buffer = f"{result} {args}"
+        self.index = len(result)
+        self.end = len(self.buffer)
```
1mn: this is being repeated in the () and [] handling. Can the code be refactored so this happens outside the if else block?

bot/__init__.py
```diff
@@ -168,6 +171,51 @@ def _get_word(self) -> str:
...
+            value = (
+                clean_argument(
+                    self.buffer.split("=")[1]
+                )
+                .replace('"', '\\"')  # escape any unescaped quotes
+                .replace("'", "\\'")  # to mimic triple quote behaviour.
+            )
+            # Use the cog's `set` command.
+            result = self.buffer[self.previous:self.index] + ".set"
+            args = f"\"{key}\" \"{value}\""
```
AnonGuy: `f"\"{key}\" \"{value}\""` could be `f'"{key}" "{value}"'`, so you won't have to escape any quotes.

bot/__init__.py
```diff
@@ -168,6 +171,46 @@ def _get_word(self) -> str:
...
+            # note: if not key, this corresponds to an empty argument
+            #       so this should throw / return a SyntaxError ?
+            args = f"\"{key}\""
+            # Use the cog's `get` command.
+            result = self.buffer[self.previous:self.index] + ".get"
+
+        # Syntax is `bot.tags['ask'] = 'whatever'` => mimic `setattr`
+        elif "=" in self.buffer and not self.buffer.endswith("="):
+            equals_pos = self.buffer.find("=")
+            # Key: The first argument, specified `bot.tags[here]`
```
1mn: some comments, like this one and the one below, would probably benefit from a blank line above them so the code wouldn't feel so cramped in this section. Have a look around for other comments that could lighten up the code density by having a blank line above them. Generally you can put blank lines above any comment, but doing it too often can hurt readability sometimes. Balancing blank lines is a readability problem without a solution, so just use your best judgement.

This is just a food for thought sort of comment, not a requirement. If you think it looks best like this, keep it like this. But do consider it, at least.

bot/__init__.py
Volcyy: I thought you were part of some cult of only using double quotes, that's why

bot/__init__.py
```diff
@@ -168,6 +171,52 @@ def _get_word(self) -> str:
...
+
+            return arg.strip("[]'\" ")
+
+        # Syntax is `bot.tags['ask']` => mimic `getattr`
+        if self.buffer.endswith("]"):
+            # Key: The first argument, specified `bot.tags[here]`
+            key = clean_argument(self.buffer[self.index:])
+            # note: if not key, this corresponds to an empty argument
+            #       so this should throw / return a SyntaxError ?
+            args = f"\"{key}\""
```
AnonGuy: The escape characters here can be avoided :eyes: 
`f"\"{key}\""` -> `f'"{key}"'`

bot/__init__.py
Volcyy: i actually don't know where this is being repeated, I don't handle a special case for `[]` (it will just get passed as `bot tags get ""` so the "empty thing), and not sure where `()` is handled. Can you elaborate?

1mn: I mean I basically do this in the regular bot.tags.get("handling") as well

bot/__init__.py
Volcyy: Thanks for the comment. It's very useful information. Will keep it in mind in the future

PR #49, opened 2018-04-13
Allow `__delitem__`-like tag deletion.

Volcyy:
clickup `21v8y`: "Allow tags to be removed with bot.tags["item"] = None. Also add keys as an alias to get, so that bot.tags.keys() will return the tag names. Now that it acts like a dict, we might as well support one more dict method."

Allows users to delete tags using the `__setitem__` syntax, when using `None` or empty strings (`''`, `""`) as the value. Additionally, add `tags.keys()` as an alias for `tags.get()` (doesn't take any arguments).

### DIFF
--- a/bot/__init__.py
+++ b/bot/__init__.py
@@ -201,22 +201,18 @@ def _get_word(self) -> str:
             key = clean_argument(self.buffer[self.index:equals_pos])
 
             # Value: The second argument, specified after the `=`
-            rhs = self.buffer.split("=")[1].strip()
+            right_hand = self.buffer.split("=", maxsplit=1)[1].strip()
 
             # If the value is None or '', mimick `bot.tags.delete(key)`
-            if rhs in ("None", "''", '""'):
+            if right_hand in ("None", "''", '""'):
                 log.trace(f"Command mimicks delitem. Key: {key!r}.")
                 result = self.buffer[self.previous:self.index] + ".delete"
                 args = f'"{key}"'
 
             # Otherwise, assume assignment, for example `bot.tags['this'] = 'that'`
             else:
-                value = (
-                    clean_argument(
-                        rhs
-                    )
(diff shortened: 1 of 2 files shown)


bot/__init__.py
```diff
@@ -201,17 +201,27 @@ def clean_argument(arg: str) -> str:
             key = clean_argument(self.buffer[self.index:equals_pos])
 
             # Value: The second argument, specified after the `=`
-            value = (
-                clean_argument(
-                    self.buffer.split("=")[1]
+            rhs = self.buffer.split("=")[1].strip()
```
1mn: not a fan of this variable name. also you might wanna split it at exactly the first = sign so it doesn't fuck up if there are multiple ='s in the buffer. I mean, consider if buffer was `bot.tags['maths'] = "2+3 =  six"`.

bot/__init__.py
```diff
@@ -201,17 +201,27 @@ def clean_argument(arg: str) -> str:
...
+                result = self.buffer[self.previous:self.index] + ".delete"
+                args = f'"{key}"'
+
+            # Otherwise, assume assignment, for example `bot.tags['this'] = 'that'`
+            else:
+                value = (
+                    clean_argument(
+                        rhs
+                    )
+                    .replace("'", "\\'")  # escape any unescaped quotes
```
1mn: this section is a bit crazy. I actually am not entirely sure how to clean this up.

possibly just like this?

```py
# escape any unescaped quotes
value = clean_argument(right_hand).replace("'", "\\'")
```

Volcyy: I do admit I went a little overboard here. Fixed that.

bot/__init__.py
Volcyy: renamed to `right_hand`, and splitting will now take the entire **r**ight **h**and **s**ide 🐤

Issue #13, opened 2018-02-22
Add `pydoc` command


jchristgit:
The standard library has a module `pydoc` which can be used for generating and viewing documentation. I suggest a command such as `[prefix] pydoc <NAME>` which calls `pydoc.doc` with the given name to show instant documentation. I feel like this could be useful when referring to a certain method when helping someone - here's an example with `str.find`: 
```py
>>> import io
>>> import pydoc
>>> out = io.StringIO()
>>> pydoc.doc('str.find', output=out)
>>> print(out.getvalue())
Python Library Documentation: method_descriptor in str

str.find = find(...)
    S.find(sub[, start[, end]]) -> int
    
    Return the lowest index in S where substring sub is found,
    such that sub is contained within S[start:end].  Optional
    arguments start and end are interpreted as in slice notation.
    
    Return -1 on failure.
```

for "bigger" documentation such as `help(dict)`, this generates quite the long output. It might be a good idea to truncate this or put it into a "pageable embed" (with reactions to move around). Also, I've yet to figure out if it's somehow possible to include a link to the official docs..

martmists:
You might be interested in using `sphinx.intersphinx` instead. It allows adding .inv files which are harmless to extract the data, contrary to pydocs which requires the modules to be installed.

gdude2002:
Are you guys planning on a PR for this one or would you like us to handle
it? We're ready to approve this ticket and triage it in our tracker if the
latter.

martmists:
I could implement this once I get my new PC set up.

gdude2002:
Perhaps leave it for one of the others, mart - this would be a great PR for someone less experienced to get their feet wet.

martmists:
I can work on one nonetheless, and add it as forever-pending PR as suggestion/inspiration

gdude2002:
That would look awful and reflect poorly on both us and you.

JoeBanks13:
It's not important that only installed modules work, we can add modules in high demand. The majority of questions asked in our guild can normally be answered with python built-in functions. I'd say pydocs is fine.

gdude2002:
No, I think the versatility would serve us well - the command will be _a lot_ more useful if we can expand it out that way.

And it's a great idea to not pollute your environment with packages.

hargoniX:
[...]

PR #53, opened 2018-04-19
Add documentation lookup support.

jchristgit:
This PR aims to add what was described in #13 - it gives users the ability to look up Python documentation for various libraries. Technically, anything that has a Sphinx inventory file should be supported. The following currently work:
- `aiohttp`
- `discord` (rewrite branch)
- `django`

and the standard library. As for the implementation:
On startup, `intersphinx` fetches the specified inventory files and populates a dictionary with a mapping of `symbol` => `absolute_documentation_url`, for example `aiohttp.ClientSession.get` => `https://aiohttp.readthedocs.io/en/stable/client_reference.html#aiohttp.ClientSession.get`. When a user requests the symbol, the dictionary is checked for the key, and if it exists, the HTML is grabbed and passed to `BeautifulSoup`. BeautifulSoup will then parse the HTML and extract the documentation from the given `absolute_doc_url`, which is passed to `markdownify`, which in turn converts the HTML to Markdown that we can use in codeblocks.

This works, but I feel like it has a few small-ish problems:
- The overall time from command invocation to bot response seems fairly high. I think that while we can't speed up the HTTP request, we could use the `lxml` parser with `BeautifulSoup` instead of the built-in `html.parser`, which should speed up the parsing. I did not add this yet, as it adds an external C dependency and may not be wanted ... would be happy to hear thoughts about this 
- Intersphinx inventories are only loaded on startup, but this takes around half a second per inventory. Right now, this sits at around 2 seconds total, but it may be nice to somehow run the requests concurrently - `intersphinx.fetch_inventory` could possibly be put in another function, which would then be run through `loop.run_in_executor`, for example
```py
await asyncio.gather(await self.fetch_inventory(inv) for inv in INTERSPHINX_INVENTORIES)
```
- Formatting between docs is sometimes inconsistent. An example for this is `WHITESPACE_AFTER_NEWLINES_RE`, which is used to trim spaces that appeared after multiple newlines in some documentation, but some not.
- `urllib3.connectionpool` does `DEBUG` logging inside of the `intersphinx.fetch_inventory` call. Is this okay, should this be suppressed? 
- A new `aiohttp.ClientSession` is made for each request. Should this be an instance attribute instead, or would it be a good idea to re-use the `self.bot`s client session? 
- I'm not sure how to get the Python syntax thing to work properly. When using `bot.doc('str.split')` for example, the command gets passed `"str.split"`, and when using `bot.doc 'str.split'`, the command gets passed `'str.split'`. Do I need to manually `strip` the quote characters? I mostly took `tags.get` as an example and it only appears to remove whitespace characters (default args to `strip`).

Ok, I think that's more than enough to lose your interest. So here's some (hopefully fancy) screenshots:

Screenshots:
has_perms thing
cs get
split


martmists:
- DO NOT use the bot session, as it contains your token which could be leaked. Instead, you could create a session on cog init
- Try to suppress debug output
- `lxml` speeds up this process enough to make it a valid requirement in my opinion
- To speed up the process (though increase memory use) you could cache the webpages/bs4 objects. 

Feature suggestions:
- Allow dynamic adding to a JSON file of sources, as more than those three may be needed and not reset at boot.

bot/cogs/doc.py
```diff
@@ -0,0 +1,207 @@
...
+    'discord': "https://discordpy.readthedocs.io/en/rewrite/",
+    'django': "https://docs.djangoproject.com/en/dev/",
+    'stdlib': "https://docs.python.org/%d.%d/" % sys.version_info[:2]
+}
+
+INTERSPHINX_INVENTORIES = {
+    'aiohttp': "https://aiohttp.readthedocs.io/en/stable/objects.inv",
+    'discord': "https://discordpy.readthedocs.io/en/rewrite/objects.inv",
+    'django': "https://docs.djangoproject.com/en/dev/_objects/",
+    'stdlib': "https://docs.python.org/%d.%d/objects.inv" % sys.version_info[:2]
```
martmists: Keep this static (currently 3.7)

bot/cogs/doc.py
```diff
@@ -0,0 +1,207 @@
...
+
+    :param max_size:
+    Specifies the maximum size the cache should have.
+    Once it exceeds the maximum size, keys are deleted in FIFO order.
+    :param arg_offset:
+    The offset that should be applied to the coroutine's arguments
+    when creating the cache key. Defaults to `0`.
+    """
+
+    cache = OrderedDict()
```
martmists: Dicts are ordered by default. To remove a key (R58) use `del cache[list(cache.keys())[0]]` instead to save memory.

bot/cogs/doc.py
```diff
@@ -0,0 +1,207 @@
...
+        self.bot = bot
+        self.client_session = aiohttp.ClientSession(loop=self.bot.loop)
+        self.inventories = {}
+        self.fetch_initial_inventory_data()
+
+    def __unload(self):
+        if not self.client_session.closed:
+            self.bot.loop.run_until_complete(self.client_session.close())
+
+    def fetch_initial_inventory_data(self):
```
martmists: Add comments to this

bot/cogs/doc.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+
+        return discord.Embed(
+            title=f'`{symbol}`',
+            url=permalink,
+            description=f"```py\n{signature}```{description}"
+        )
+
+    @commands.command(name='doc()', aliases=['doc'])
+    async def doc(self, ctx, *, symbol: commands.clean_content):
```
martmists: If not found, perform a fuzzy search (fuzzywuzzy or faster implementation which I can give you)

bot/cogs/doc.py
```diff
@@ -0,0 +1,207 @@
...
+        element is the description of the given symbol with HTML
+        markup included. If the given symbol could not be found,
+        returns `None`.
+        """
+
+        url = self.inventories.get(symbol)
+        if url is None:
+            return None
+
+        async with aiohttp.ClientSession() as cs:
```
martmists: Use own ClientSession from `__init__`

Volcyy: 🤦‍♂️ I forgot that

bot/cogs/doc.py
Volcyy: > The order-preserving aspect of this new implementation is considered an implementation detail and should not be relied upon (this may change in the future, but it is desired to have this new dict implementation in the language for a few releases before changing the language spec to mandate order-preserving semantics for all current and future Python implementations; this also helps preserve backwards-compatibility with older versions of the language where random iteration order is still in effect, e.g. Python 3.5).
from [What's New in Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-compactdict) - I also thought it would be more intuitive to use this here.

bot/cogs/doc.py
Volcyy: hmm, why though? wouldn't it be preferable to link to the most recent stable version (which I assume the bot to be running on)?

bot/cogs/doc.py
Volcyy: 👍

bot/cogs/doc.py
Volcyy: `fuzzywuzzy` is licensed under GPL2... doesn't that mean we'd have to use GPL2 for this repo, or did I misunderstand how GPL2 works?

martmists: MIT I guess, had this in my gists somewhere

martmists:
Inviting other reviewers to add to my review if needed.

bot/cogs/doc.py
1mn: we'd have to keep changing it for it to be possible to refer to upcoming features. I prefer Volcyys approach to this tbh, because that never needs to be manually maintained.

bot/cogs/doc.py
1mn: `del cache[list(cache.keys())[0]]`  looks horrible. OrderedDicts are fine. dicts are not guaranteed to retain insertion order until 3.7, and even then there are compelling reasons to just use OrderedDict anyway because of readability. The "save memory" argument has no place in a project of this nature. Code readability trumps marginal memory saving every time.

bot/cogs/doc.py
1mn: we can use fuzzywuzzy.

martmists: fuzzywuzzy is slower though, but sure. I'm not sure about licensing, but my code is an option. If fuzzywuzzy's performance turns out to be fast enough, we can definitely use it.

bot/cogs/doc.py
```diff
@@ -0,0 +1,210 @@
...
+            )
+
+        return discord.Embed(
+            title=f'`{symbol}`',
+            url=permalink,
+            description=f"```py\n{signature}```{description}"
+        )
+
+    @commands.command(name='doc()', aliases=['doc'])
+    async def doc(self, ctx, *, symbol: commands.clean_content):
```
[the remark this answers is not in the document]
martmists: I could maybe PR a fix to that, want me to look into it?

bot/cogs/doc.py
martmists: In that case I do recommend using `str.format` or f-strings instead of %-formatting: 

```py
    'stdlib': f"https://docs.python.org/{'.'.join(sys.version_info[:2]}/objects.inv"
```
As %-formatting in later versions of python is heavily frowned upon

bot/cogs/doc.py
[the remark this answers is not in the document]
1mn: tbh I'd prefer to solve this myself. you could try but I'll probably be unreasonably relentless with that PR.

bot/cogs/doc.py
1mn: This I _do_ agree with.

martmists:
Use single and double quotes consistently, don't switch between them. (Though this should be a project-wide change, maybe a future PR?)

1mn:
yeah, we generally prefer that double quotes are used unless the string contains a double quote. if it does, then single quotes are acceptable because it looks tidier than escape characters.

Volcyy:
I really disagree with this. My general idea is using single quotes for short strings that control e.g. parameters - imagine an ORM - ```py
db.insert(values, on_conflict='update')
```
and double quotes for descriptive things, or stuff going towards a user, for example
```py
bot.send_message(user, f"Hello {user.name}, thanks for adding me!")
```
but if you really want to use a single kind of quote, then sure...

1mn:
well, that's an interesting style convention. My chief concern with that is that it is not the prevailing style in this project, and so would be inconsistent with the rest of the project. While PEP8 does not make known any preference of `"` over `'` (other than requiring that one does use `"""` for docstrings in PEP257), it does make it very clear that one should always strive to be consistent with the rest of the project. 

So if we were to adhere to your admittedly useful-sounding convention, PEP8 would require that someone goes through all our existing code and makes the same changes to the rest of it. This, in turn, would screw up a lot of git blame unless we make all the various authors change their own code...

So, for that reason alone, I'm skeptical to allowing it, sensible as it may be.

I'd actually like to see some other core devs comment on this.

gdude2002:
# Pre-review comment response

> * The overall time from command invocation to bot response seems fairly high. I think that while we can't 
>   speed up the HTTP request, we could use the lxml parser with BeautifulSoup instead of the built-in 
>   html.parser, which should speed up the parsing. I did not add this yet, as it adds an external C 
>   dependency and may not be wanted ... would be happy to hear thoughts about this

Go for it, we can do C extensions in prod.

> * A new aiohttp.ClientSession is made for each request. Should this be an instance attribute instead, 
>   or would it be a good idea to re-use the self.bots client session?

There are some real issues with aiohttp not cleaning up properly when you use multiple sessions. Please use the one on `self.bot`. We provided a safe session on `self.bot.http_session`.

> * I'm not sure how to get the Python syntax thing to work properly. When using bot.doc('str.split') for 
>   example, the command gets passed "str.split", and when using bot.doc 'str.split', the command gets 
>   passed 'str.split'. Do I need to manually strip the quote characters? I mostly took tags.get as an 
>   example and it only appears to remove whitespace characters (default args to strip).

The former uses python-based syntax parsing. The latter completely ignores it. I think it should work correctly if you use double quotes, though.

> Style convention conversation...

Yeah, we should stick with one style of quotes throughout the project. I use double quotes, so much of the project is written that way - but we haven't yet decided on a form and made it consistent yet.

Volcyy:
So I've just given BeautifulSoup using the `lxml` parser a try, and it doesn't really seem to cause any considerable speedup. The main time probably still comes from requesting the web resource and downloading it, but that's now cached with a decorator.

As for adding support for adding / removing / modifying this documentation metadata, the PR on the site for the accompanying API is ready (https://github.com/discord-python/site/pull/57), I just need to add commands to interface with it here. 
Since this will involve adding at least the tags URL to the constants file, I initially planned to hold it off until #57 is merged. Seeing as it's still in review, I guess I'll just add it to the constants.

GhostofGoes:
Is there any caching going on, or does it do a full pull and parse every time docs are requested?

Are there any rate restrictions? In other words, do you think it would be possible for someone slow or crash the bot by making a bunch of documentation requests (intentionally or not).


Sorry in advance if these are dumb questions. 😄

Volcyy:
> Is there any caching going on, or does it do a full pull and parse every time docs are requested?

Up to 128 recent symbols are cached, namely, the embed that the bot would respond with. So if someone looks up `str.join` and someone else looks that up later, it's near instant, since the bot just fetches the embed from the cache. Fetching an Embed for each symbol in advance would be *possible*, but I think it would be quite overkill and also do a metric ton of requests to documentation sites, and slow up startup immensely. 

> Are there any rate restrictions? In other words, do you think it would be possible for someone slow or crash the bot by making a bunch of documentation requests (intentionally or not).

There are no rate restrictions, no. This could easily be implemented if wanted, but I don't think its necessary. The majority of time that it takes to fetch an embed for a symbol is used for getting the HTML from the page, which is asynchronous anyways.

bot/cogs/doc.py
```diff
@@ -0,0 +1,546 @@
...
+from bot.decorators import with_role
+from bot.exceptions import CogBadArgument
+
+
+log = logging.getLogger(__name__)
+logging.getLogger('urllib3').setLevel(logging.WARNING)
+
+
+UNWANTED_SIGNATURE_SYMBOLS = ('[source]', '¶')
+WHITESPACE_AFTER_NEWLINES_RE = re.compile(r"(?<=\n\n)( +)")
```
martmists: Use `\s` rather than ` `

bot/cogs/doc.py
```diff
@@ -0,0 +1,546 @@
...
+    :param max_size:
+    Specifies the maximum size the cache should have.
+    Once it exceeds the maximum size, keys are deleted in FIFO order.
+    :param arg_offset:
+    The offset that should be applied to the coroutine's arguments
+    when creating the cache key. Defaults to `0`.
+    """
+
+    # Assign the cache to the function itself so we can clear it from outside.
+    async_cache.cache = OrderedDict()
```
martmists: Don't set it to the function, make it a local variable instead

bot/cogs/doc.py
```diff
@@ -0,0 +1,546 @@
...
+                if len(async_cache.cache) > max_size:
+                    async_cache.cache.popitem(last=False)
+
+                async_cache.cache[key] = await function(*args)
+            return async_cache.cache[key]
+        return wrapper
+    return decorator
+
+
+class DocMarkdownConverter(MarkdownConverter):
```
martmists: These classes and dummy classes would likely be better of in a utility file or something similar.

bot/cogs/doc.py
```diff
@@ -0,0 +1,546 @@
...
+
+class SphinxConfiguration:
+    """Dummy configuration for use with intersphinx."""
+
+    config = DummyObject()
+    config.intersphinx_timeout = 3
+    config.tls_verify = True
+
+
+class ValidPythonIdentifier(commands.Converter):
```
martmists: Same with these converters. Though they're only used here, it would be cleaner to have them in their own file with this file only containing the cog.

bot/cogs/doc.py
```diff
@@ -0,0 +1,546 @@
...
+        self.bot = bot
+        self.inventories = {}
+        self.headers = {"X-API-KEY": SITE_API_KEY}
+        self.url = f"{SITE_API_URL}/bot/doc"
+
+    async def on_ready(self):
+        await self.refresh_inventory()
+
+    async def refresh_inventory(self):
+        async def update_single(
```
martmists: Use a private function rather than a local function.

bot/cogs/doc.py
```diff
@@ -0,0 +1,546 @@
...
+
+        package_json = {
+            'package': name,
+            'base_url': base_url,
+            'inventory_url': inventory_url
+        }
+
+        async with self.bot.http_session.post(
+            self.url, headers=self.headers, json=package_json
+        ) as resp:
```
martmists: These indents may seem unreadable to some users, try changing indentation.

bot/cogs/doc.py
Volcyy: rather than...?

bot/cogs/doc.py
Volcyy: The entire point of setting it to the function is so we can clear the cache from the outside of it when necessary. Did you read the comment above?

bot/cogs/doc.py
Volcyy: Cogs are currently all in a single file. Unless there's something that we can absolutely use elsewhere, I don't see a point in putting it in the utility module.

bot/cogs/doc.py
Volcyy: Seems pretty readable to me. What would you suggest instead?

bot/cogs/doc.py
Volcyy: This isn't really used anywhere else

bot/cogs/doc.py
martmists: whoops, looks like using \` on a space doesn't work. Whitespace is defined as \s, and documentation may contain tabs or zero-width spaces or similar (happens on a couple ones, actually), so instead of `( +)` you should use `(\s+)`

bot/cogs/doc.py
martmists: Still, local functions are considered bad practice. Having it like `async def _update_single(self, ...)` would both make it easier to understand what code is from what function, as well as keeping function locals small, as this function would get created *every time* the function is ran, whereas keeping it on class level would only create it on class creation. This results in a speedup while running, and a bit longer for startup time (both minimal, but small changes make big changes)

bot/cogs/doc.py
martmists:
```python
async with self.bot.http_session.post(self.url, headers=self.headers, json=package_json) as resp:
    ...
# or 

async with self.bot.http_session.post(
        self.url, headers=self.headers, json=package_json) as resp:  # 8 indents to differ from code indent
    ...
# or 

async with self.bot.http_session.post(self.url, 
                                      headers=self.headers, 
                                      json=package_json) as resp:
    ...
```

basically, while your code works, it may show to inexperienced users as if the `)` might close the `async with` as it's on the same indent level as the `async with`.

bot/cogs/doc.py
```diff
@@ -0,0 +1,542 @@
...
+
+class SphinxConfiguration:
+    """Dummy configuration for use with intersphinx."""
+
+    config = DummyObject()
+    config.intersphinx_timeout = 3
+    config.tls_verify = True
+
+
+class ValidPythonIdentifier(commands.Converter):
```
gdude2002: We do have a converters file, converters should probably be in there.

gdude2002:
**In order to expedite our current plans, we're requiring that all PRs be closed or merged.**

You have some outstanding review comments to address!

bot/cogs/doc.py
jchristgit: That would make the converters module a bit messy, IMO. For example, the `DocumentationBaseURL` and `InventoryURL` converters makes use of:
- `aiohttp.ClientConnectorError`
- `ssl.CertificateError`
- `intersphinx`
- `bot.cogs.doc.SphinxConfiguration`

I can move it there if it's really wanted, but it would make more sense to keep it in the cog IMO - at least the `InventoryURL` converter. I can see `DocumentationBaseURL` (renamed to something like `ValidURL`) and `ValidPythonIdentifier` having some general use, but I can't really see `InventoryURL` being used in another cog.

gdude2002: Have you seen the contents of that file? It has a bunch of snake cog stuff in it.

It's really just a maintainability thing. But if it would be messier to move it, then I guess you could just not.

jchristgit: I have, yes. But I could move it. I'm okay with both, so you tell me 🤷‍♂️

gdude2002: Do you feel like it's more effort than it's worth?

jchristgit: No, will do

PR #54, opened 2018-04-19
[#1eeu1] Hiphopify

1mn:
A bot feature for enforcing temporary nick changes.

Takes a user mention and a duration, and forces the user to temporarily have a random hiphop name selected from https://en.wikipedia.org/wiki/List_of_hip_hop_musicians. 
image

A staff-member can call `bot.unhiphopify` to release them from hiphop-prison.
image

If the user tries to change their nickname back while they are in hiphop-prison, the bot will change it back, and then DM them with a message that says they cannot do that.
image

Bot also adds a mod-log item:
image


bot/cogs/hiphopify.py
```diff
@@ -0,0 +1,184 @@
...
+        If not provided, this function shows the caller a list of all tags.
+        """
+
+        log.debug(
+            f"Attempting to hiphopify {member.display_name} for {duration}. "
+            f"forced_nick is set to {forced_nick}."
+        )
+
+        embed = Embed()
+        embed.colour = Colour.blurple()
```
JoeBanks13: Any reason why this isn't in the creation of the embed, `Embed(colour=Colour.blurple())`?

bot/cogs/hiphopify.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+
+        response = await response.json()
+
+        if response:
+
+            if after.display_name == response.get("forced_nick"):
+                return  # Nick change was triggered by this event. Ignore.
+
+            log.warning(
```
JoeBanks13: Under our previous discussion regarding error levels we decided that the definition for the warning level was
> Someone is interacting with the application in an unexpected way or the application is responding in an unexpected way, but without causing an error.

I think that either DEBUG or INFO would be more suitable for this.

bot/cogs/hiphopify.py
```diff
@@ -0,0 +1,184 @@
...
+                f"{after.display_name} is currently in hiphop-prison. "
+                f"Changing the nick back to {before.display_name}."
+            )
+            await after.edit(nick=response.get("forced_nick"))
+            await after.send(
```
JoeBanks13: Could this raise an error if the user has the bot blocked/DMs disabled?

bot/cogs/hiphopify.py
```diff
@@ -0,0 +1,184 @@
...
+        response = await self.bot.http_session.get(
+            SITE_API_HIPHOPIFY_URL,
+            headers=self.headers,
+            params={"user_id": str(before.id)}
+        )
+
+        response = await response.json()
+
+        if response:
+
```
JoeBanks13: Empty line not needed

bot/cogs/hiphopify.py
```diff
@@ -0,0 +1,184 @@
...
+            await mod_log.send(
+                f":middle_finger: {member.name}#{member.discriminator} (`{member.id}`) "
+                f"has been hiphopified by **{ctx.author.name}**. Their new nickname is `{forced_nick}`. "
+                f"They will not be able to change their nickname again until **{end_time}**"
+            )
+
+            # Change the nick and return the embed
+            log.debug("Changing the users nickname and sending the embed.")
+            await member.edit(nick=forced_nick)
+            return await ctx.send(member.mention, embed=embed)
```
JoeBanks13: Unused/unneeded return.

bot/cogs/hiphopify.py
```diff
@@ -0,0 +1,184 @@
...
+            embed.colour = Colour.red()
+            embed.title = random.choice(NEGATIVE_REPLIES)
+            embed.description = response.get("error_message")
+            log.warning(
+                f"Error encountered when trying to unhiphopify {member.display_name}:\n"
+                f"{response}"
+            )
+
+        log.debug(f"{member.display_name} was successfully released from hiphop-prison.")
+        return await ctx.send(embed=embed)
```
JoeBanks13: Unused/unneeded return.

bot/cogs/hiphopify.py
1mn: I suppose it could be. But we've done it this way in more or less all the other cogs, so it's done this way out of convention. We could change them all, but I'm not sure it's better because it seems to violate the Zen of Python somewhat. Flat is, after all, better than nested. Then again, I'm not sure it's that much worse, either. it's quite ¯\_(ツ)_/¯

bot/cogs/hiphopify.py
1mn: agreed.

bot/cogs/hiphopify.py
JoeBanks13: That's fine then

bot/cogs/hiphopify.py
```diff
@@ -0,0 +1,184 @@
...
+        hiphop-prison. If they are not allowed, we will change it back.
+        :return:
+        """
+
+        if before.display_name == after.display_name:
+            return  # User didn't change their nickname. Abort!
+
+        log.debug(
+            f"{before.display_name} is trying to change their nickname to {after.display_name}. "
+            f"Checking if the user is in hiphop-prison..."
```
AnonGuy: Does this line need to be an f-string?

bot/cogs/hiphopify.py
```diff
@@ -0,0 +1,184 @@
...
+            return await ctx.send(embed=embed)
+
+        else:
+            forced_nick = response.get('forced_nick')
+            end_time = response.get("end_timestamp")
+            image_url = response.get("image_url")
+
+            embed.title = "Congratulations!"
+            embed.description = (
+                f"Your previous nickname was so bad that we have decided to change it. "
```
AnonGuy: Does this line need to be an f-string?

bot/cogs/hiphopify.py
1mn: hm, possibly. but the error would bubble up to events and just get logged, so it wouldn't break the bot or anything. might be worth testing.

PR #55, opened 2018-04-19
Use discord.py's case insensitivity instead of patching `bot.cogs` ourselves

Volcyy:
ClickUp `21rc8`:
> Use d.py's built-in case insensitivity instead of our own.

> Discord rewrite now supports case insensitivity by simply passing case_insensitive=True into the bot constructor. This means we should probably replace our current case insensitivity monkey patching with this system instead, and we might not even need our case insensitive dict anymore. 
However, we need to test this properly before we replace. Does it work for prefixes and commands? Are we losing anything by using this method instead of todays method?

Instead of patching `bot.cogs`, this just passes `case_insensitive=True` to discord.py's bot constructor.

`CaseInsensitiveDict` is still used in the ClickUp cog:
https://github.com/discord-python/bot/blob/master/bot/cogs/clickup.py#L32-L39

### DIFF
--- a/bot/__init__.py
+++ b/bot/__init__.py
@@ -232,8 +232,6 @@ def _get_word(self) -> str:
         self.end = len(self.buffer)
         log.trace(f"Mimicked command: {self.buffer}")
 
-    if isinstance(result, str):
-        return result.lower()  # Case insensitivity, baby
     return result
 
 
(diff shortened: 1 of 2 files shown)


Volcyy:
I removed the case insensitivity from the `_get_word`, but the other monkey patch will have to stay in to allow case-insensitive prefixes.

PR #56, opened 2018-04-19
Use converters for tag command argument validation.

Volcyy:
ClickUp `21rbv`: 
> Turn validation for the tags cog into a converter. Thanks Ava!

This PR changes the `Tags` cog to use converters for validating command arguments instead of calling the `validate` coroutine manually.
This has a small side effect: since `BadArgument` exceptions are handled [in `on_command_error`](https://github.com/discord-python/bot/blob/7554e7d25ca67b23590f9206b11ba7437d826265/bot/cogs/events.py#L52), handling the error manually (in order to send a red embed with one of the `FAIL_TITLES`) would result in sending both the embed *and* triggering `on_command_error` afterwards.


1mn:
that side-effect may actually be desirable.

1mn:
functions won't look better than classes anyway, I guess we should just keep the classes then.

PR #57, opened 2018-04-24
Configure constants with YAML file(s)

Volcyy:
This PR removes the hardcoded constants from `bot/constants.py` and instead loads them from a YML file.

The defaults provided in `config-default.yml` should suffice for the default deployment, but custom per-user configuration is now a lot easier by simple overriding the unwanted values from `config-example.yml` in `config.yml`. 

In a nutshell, this
- adds the `pyyaml` dependency, used to parse the YAML files
- adds two tags that can be used in the YAML file `!ENV` and `!REQUIRED_ENV`
- replaces any previously hardcoded constant from `bot/constants.py` and moves it in `config-default.yml` instead
- "dynamically" loads configuration through class attribute access
- completely keeps custom per-user configuration out of VCS

The original idea for this "class-based configuration" and the original implementation of the `YAMLGetter` class was made by @1mn and slightly altered by me.

I've checked through each file to ensure that I replaced any old constant with the new class-based attribute thing, but it might still be useful to re-check again, just to be sure(although flake8 should catch unknown variables?).


GhostofGoes:
What's the advantage of YAML for simple configurations over JSON? PyYAML is a good project but it's another dependency. Also, if will search for and attempt to build bindings for LibYAML C library if it exists on the system, which can sometimes be problematic.

Volcyy:
I think YAML is a *lot* more readable than JSON. 
Compare the current config file to this:
```json
{
    "bot": {
        "help_prefix": "bot.",
        "token": "dipVwoOFIHmtffzuvzszFPbxzkPqUuamcCNLhmYOTAFjhXJmxoBajOSYUYIO",
        "cooldowns": {
            "tags": 1
        },
        "emojis": {
            "green_chevron": "<:greenchevron:418104310329769993>",
            "red_chevron": "<:redchevron:418112778184818698>",
            "white_chevron": "<:whitechevron:418110396973711363>"
        }
    },
    "guild": {
        "id": 434048682569105409,
        "channels": {
            "bot": 434048683135205387,
            "help1": 434048683135205387,
            "help2": 434048683135205387,
            "help3": 434048683135205387,
            "python": 434048683135205387,
            "devlog": 434048683135205387,
            "devtest": 434048683135205387,
            "verification": 434048683135205387,
            "checkpoint_test": 434048683135205387
        },
        "roles": {
            "admin": 434048999071154205,
            "moderator": 434048999071154205,
            "verified": 434048999071154205,
            "owner": 434048999071154205,
            "devops": 434048999071154205,
            "contributor": 434048999071154205
        }
    },
    "keys": {
        "deploy_bot": null,
        "deploy_site": null,
        "site_api": "abcdefghijklmnopqrstuvwxyz"
    },
    "clickup": {
        "space": 757069,
        "team": 754996,
        "key": null
    },
    "papertrail": {
        "address": null,
        "port": null
    },
    "urls": {
        "deploy": null,
        "status": null,
        "site": "pythondiscord.com",
        "site_tags_api": "https://api.pythondiscord.com/tags",
        "site_user_api": "https://api.pythondiscord.com/users",
        "github_bot_repo": "https://github.com/discord-python/bot",
        "bot_avatar": "https://raw.githubusercontent.com/discord-python/branding/master/logos/logo_circle/logo_circle.png"
    }
}
```

Yes, it's somewhat readable, but having doublequotes around keys and `{}` clutters it up quite a bit. 
YAML's indentation use instead of using brackets feels quite similar to Python, and to me, this just feels right. 🦆 

Another thing is that we can use custom tags in YML to completely move configuration out of the `bot/constants.py` file and configure it through the YML file completely. What made this work is adding `!ENV` and `!REQUIRED_ENV`, which makes it clear what it's doing. I'm not sure whether something like this is easily possible in JSON.

Also, keep in mind that you can use comments in the YAML file, should we ever add a setting that requires documenting or should someone suggest to add comments to the settings in this PR.

That being said, if people want to change it to JSON, I can convert it.

GhostofGoes:
YAML is definitely more readable (and I prefer it usually), but for some
constants that don't change often it just seems a bit overkill. If the same
parsing infrastructure will be used elsewhere, however, then I definitely
agree it'd be worth it.

Chris Goes

1mn:
the killer feature for yaml in this is the !ENV. we'd have to write this ourselves to do it in json, and it is something we absolutely need. stuff like BOT_TOKEN depends on being an env var on the prod side and being in the config on the user side.

I also don't think it's unreasonable to think that we may find another use for this parser down the road.

bot/__main__.py
```diff
@@ -48,7 +48,7 @@
 
 # Local setups usually don't have the clickup key set,
 # and loading the cog would simply spam errors in the console.
-if CLICKUP_KEY is not None:
+if ClickUp.key is not None:
     bot.load_extension("bot.cogs.clickup")
 else:
     log.warning("`CLICKUP_KEY` not set in the environment, not loading the ClickUp cog.")
```
JoeBanks13: I wouldn't consider not having ClickUp key a warning, as I don't expect many, if not any, local setups to have a valid ClickUp Key, info should be fine here

bot/cogs/deployment.py
```diff
@@ -3,7 +3,9 @@
 from discord import Colour, Embed
 from discord.ext.commands import AutoShardedBot, Context, command
 
-from bot.constants import ADMIN_ROLE, DEPLOY_BOT_KEY, DEPLOY_SITE_KEY, DEPLOY_URL, DEVOPS_ROLE, OWNER_ROLE, STATUS_URL
+from bot.constants import (
+    Keys, Roles, URLs
```
JoeBanks13: Is there any point multi-line importing or just 3 things, earlier in this file we import `AutoShardedBot`, `Context` and `command` on a single line

bot/cogs/snakes.py
```diff
@@ -18,15 +18,15 @@
 from PIL import Image, ImageDraw, ImageFont
 
 from bot.constants import (
-    ERROR_REPLIES, OMDB_API_KEY, SITE_API_KEY,
-    SITE_API_URL, YOUTUBE_API_KEY
+    ERROR_REPLIES, Keys, URLs
```
JoeBanks13: This can be on a single line

bot/cogs/verification.py
```diff
@@ -3,7 +3,9 @@
 from discord import Message, Object
 from discord.ext.commands import AutoShardedBot, Context, command
 
-from bot.constants import VERIFICATION_CHANNEL, VERIFIED_ROLE
+from bot.constants import (
+    Channels, Roles
```
JoeBanks13: Can be single line

bot/__main__.py
Volcyy: This is from another PR. But sure

bot/constants.py
```diff
@@ -1,69 +1,204 @@
...
+                (cls.section, cls.subsection, name)
+                if cls.subsection is not None else (cls.section, name)
+            )
+            log.critical(f"Tried accessing configuration variable at `{dotted_path}`, but it could not be found.")
+            raise
+
+    def __getitem__(cls, name):
+        return cls.__getattr__(name)
+
+
```
1mn: this line could do with a block comment 

`# Dataclasses` or whatever

bot/constants.py
```diff
@@ -1,69 +1,204 @@
...
+class Papertrail(metaclass=YAMLGetter):
+    section = "papertrail"
+
+
+class URLs(metaclass=YAMLGetter):
+    section = "urls"
+
 
 # Debug mode
 DEBUG_MODE = True if 'local' in os.environ.get("SITE_URL", "local") else False
```
1mn: actually I'm not sure if we're even using this constant anymore. could you check and just remove it if not?

Volcyy: Pretty sure we are:
```sh
$ grep -r 'DEBUG_MODE' .
./bot/constants.py:DEBUG_MODE = True if 'local' in os.environ.get("SITE_URL", "local") else False
./bot/converters.py:from bot.constants import DEBUG_MODE, Keys, URLs
./bot/converters.py:        if DEBUG_MODE:
./bot/cogs/logging.py:from bot.constants import Channels, DEBUG_MODE
./bot/cogs/logging.py:        if not DEBUG_MODE:
./bot/cogs/events.py:    Channels, DEBUG_MODE, Guild,
./bot/cogs/events.py:                if not DEBUG_MODE:
```

1mn: fair enough.

PR #58, opened 2018-04-24
Declaring the encoding to be utf-8 is not necessary in Python3.

1mn:
UTF-8 is the default encoding in Python3 - see [PEP3120](https://www.python.org/dev/peps/pep-3120/). Encoding declarations are only useful in py3 if you want to declare it to be something _other_ than utf-8.

This is basically a Python2 convention that someone has carried over. Since it doesn't actually do anything at all, it should be removed.

### DIFF
--- a/bot/utils.py
+++ b/bot/utils.py
@@ -1,5 +1,3 @@
-
-
 class CaseInsensitiveDict(dict):
     """
     We found this class on StackOverflow. Thanks to m000 for writing it!
--- a/bot/utils.py
+++ b/bot/utils.py
@@ -1,4 +1,3 @@
-# coding=utf-8
 
 
 class CaseInsensitiveDict(dict):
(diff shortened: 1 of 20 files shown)


bot/utils.py
```diff
@@ -1,4 +1,3 @@
-# coding=utf-8
```
JoeBanks13: Is PEP8 happy with these blank lines at the start of the file?

JoeBanks13: Is PEP8 happy with these blank lines?

gdude2002: I think by convention there should be no blank lines at the top of the file.

1mn:
I tend to agree, and the only thing holding me back is that the PEP8 does not really appear to make any exceptions to surrounding classes with blank lines. but this is, perhaps, because it believes there should always be a module docstring? 

anyway I'm fine with us agreeing to not put blank lines at the tippy top.

bot/utils.py
1mn: sorted.

PR #60, opened 2018-04-27
adds dockerfile that uses pipenv

Inveracity:
```
docker build -t pythondiscord/bot .
```


Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+ENV PIPENV_VENV_IN_PROJECT=1
+ENV PIPENV_IGNORE_VIRTUALENVS=1
+ENV PIPENV_NOSPIN=1
+ENV PIPENV_HIDE_EMOJIS=1
+
+RUN pip install pipenv
+
+COPY . /
+
+RUN pipenv sync
```
Volcyy: would it make sense to use `--bare` for minimal output here?

Inveracity: I prefer having a lot of output instead, makes it easier to debug in case of a bad build

PR #62, opened 2018-04-27
Python parser cleanup

1mn:
Abstracted the python parser into its own function. Added error handling for it, which will print a pythonic looking SyntaxError or ValueError to our log instead of the full traceback, which was in these particular cases completely useless.

This will significantly reduce the logspam we saw after the setattr literal_eval bugfix PR.

This PR also adds a few exceptions to the tag cooldown. Sorry, that's a little bit out of scope but it was infuriating to test this without it and since I already solved it it might as well go in our bot, we've been meaning to do it anyway.


1mn:
this is how it looks when the python parser syntaxerrors now.

image

I wanted to put this in the channel it was called from, but this proved to be extremely tricky to do from within the bowels of the discord.py library, inside the parser. There's just no way to get the error to bubble to somewhere I'd have access to context.

bot/__init__.py
```diff
@@ -104,136 +239,33 @@ def _get_word(self) -> str:
...
-        self.buffer = f"{result} {args}"
-        self.index = len(result)
-        self.end = len(self.buffer)
-        log.trace(f"Mimicked command: {self.buffer}")
+    # If the command looks like a python syntax command, try to parse it.
+    if current == "(" or current == "[":
+        try:
+            python_result = parse_python(pos)
+            if python_result:
+                result = python_result
```
Volcyy: hmm. maybe instead of checking for `None` (or is there another case that may occur?) here, wouldn't it be simpler and / or more intuitive to just `raise SyntaxError` within the parser?

bot/cogs/tags.py
```diff
@@ -9,15 +9,22 @@
...
 from bot.pagination import LinePaginator
 
 log = logging.getLogger(__name__)
 
+TEST_CHANNELS = [
+    DEVTEST_CHANNEL,
+    BOT_COMMANDS_CHANNEL,
+    HELPERS_CHANNEL
+]
+
```
Volcyy: i know that people usually use lists for this anyways, but, since we're not modifying it anyways, why not make it a tuple?

bot/cogs/tags.py
```diff
@@ -209,10 +216,13 @@ def _command_on_cooldown(tag_name) -> bool:
             if tag_name:
                 log.debug(f"{ctx.author} requested the tag '{tag_name}'")
                 embed.title = tag_name
-                self.tag_cooldowns[tag_name] = {
-                    "time": time.time(),
-                    "channel": ctx.channel.id
-                }
+
+                # Cooldown
```
Volcyy: imho, I don't think this comment provides a lot of value. I can see that this is *something to do with cooldown*, but I can also see that by reading the if statement body. maybe something like `# Set the cooldown for this tag on this channel` would be more intuitive, although I don't know whether this really needs a comment anyways.

bot/__init__.py
```diff
@@ -101,131 +101,118 @@ def parse_python(view_pos):
...
-                             r'"\1"',
-                             self.buffer)
+        tempbuffer = self.buffer
+        tempbuffer = re.sub(r"(<(?:@|@!|[#&])\d+>)",
+                            r'"\1"',
+                            tempbuffer)
 
         # Let's parse!
-        python_result = None
+        log.debug(f"A python-style command was used. Attempting to parse. "
```
Volcyy: does this need to be a f-string?

bot/__init__.py
```diff
@@ -101,131 +101,118 @@ def parse_python(view_pos):
...
+            # Other types get converted to strings
+            if not isinstance(arg, str):
+                log.trace(f"{arg} is not a str, casting to str.")
+                arg = str(arg)
+
+            # Allow using double quotes within triple double quotes
+            arg = arg.replace('"', '\\"')
+
+            # Adding double quotes to every argument
+            log.trace(f"Wrapping all args in double quotes.")
```
Volcyy: does this need to be a f-string?

bot/__init__.py
```diff
@@ -101,131 +101,118 @@ def parse_python(view_pos):
...
                     log.trace(f"Command mimicks setitem. Key: {key!r}, value: {value!r}.")
-
-                    # Use the cog's `set` command.
-                    python_result = self.buffer[self.previous:self.index] + ".set"
-                    args = f'"{key}" "{value}"'
+                    parsed_result = self.buffer[self.previous:self.index] + ".set"
+                    args = (key, value)
 
             # Syntax is god knows what, pass it along
-            # in the future, this should probably return / throw SyntaxError
```
Volcyy: so we scrapped the `SyntaxError` altogether as its currently not possible to implement, right?

PR #64, opened 2018-04-29
Show alternative syntax in help texts for `bot.tags.keys()` and `bot.tags`

Volcyy:
clickup 2fmet: "The footer for the bot.tags.keys call does not use python syntax in its example. neither does the help text at the bottom of bot.tags, possibly others as well. investigate and rectify."

This PR makes the embed footer of `bot.tags.keys()` show `bot.tags[<tagname>]` instead of `bot.tags.get <tagname>`, and adds a help section for an alias to `bot.tags.get` there:
help text

This is done by checking whether the current command name ends in `.get()`, and if so, adding the relevant line to the paginator.

It would also be possible to add this for `.set()` if desired.


1mn:
Hokay, let's do it!

PR #73, opened 2018-05-08
Make python parser handle single bracket calls like self.accept(. 

1mn:
This solves a problem where these messages would get stuck in #checkpoint and never get removed. Also added #bot-commands to codeblock parsing whitelist so there is no cooldown in that channel anymore.


bot/cogs/bot.py
```diff
@@ -234,7 +234,7 @@ def repl_stripping(self, msg: str):
     async def on_message(self, msg: Message):
         if msg.channel.id in self.channel_cooldowns and not msg.author.bot and len(msg.content.splitlines()) > 3:
             on_cooldown = time.time() - self.channel_cooldowns[msg.channel.id] < 300
-            if not on_cooldown or msg.channel.id == DEVTEST_CHANNEL:
+            if not on_cooldown or msg.channel.id in [DEVTEST_CHANNEL, BOT_COMMANDS_CHANNEL]:
```
Volcyy: maybe use a tuple instead? we're not modifying it anyways

PR #78, opened 2018-05-13
Snake cog

1mn:


bot/__init__.py
```diff
@@ -241,7 +243,7 @@ def parse_python(buffer_pos):
     self.index += pos
 
     # If the command looks like a python syntax command, try to parse it.
-    if current == "(" or current == "[":
+    if current and current == "(" or current == "[":
```
Volcyy: Is this needed? if `current` is a string, `bool(current) = len(current) > 0`. If `current` is `None`, as above, then `current != "("` and `current != "["`, or am I missing something?

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1231 @@
...
+HOLE_EMOJI = "\u2B1C"         # :white_square: - Used in guesses
+EMPTY_UNICODE = "\u200b"      # literally just an empty space
+
+ANTIDOTE_EMOJI = [
+    SYRINGE_EMOJI,
+    PILL_EMOJI,
+    HOURGLASS_EMOJI,
+    CROSSBONES_EMOJI,
+    ALEMBIC_EMOJI,
+]
```
Volcyy: do we need to modify this? if not, what about using a tuple instead?

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1231 @@
...
+
+ANSWERS_EMOJI_REVERSE = {
+    "\U0001F1E6": "A",  # :regional_indicator_a: 🇦
+    "\U0001F1E7": "B",  # :regional_indicator_b: 🇧
+    "\U0001F1E8": "C",  # :regional_indicator_c: 🇨
+    "\U0001F1E9": "D",  # :regional_indicator_d: 🇩
+}
+
+# Zzzen of pythhhon constant
+ZEN = """
```
Volcyy:
```py
import this

ZEN = ''.join(this.d.get(char) or char for char in this.s)
```
?

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1231 @@
...
+    Commands related to snakes. These were created by our
+    community during the first code jam.
+
+    More information can be found in the code-jam-1 repo.
+
+    https://github.com/discord-python/code-jam-1
+    """
+
+    wiki_brief = re.compile(r'(.*?)(=+ (.*?) =+)', flags=re.DOTALL)
+    valid = ('gif', 'png', 'jpeg', 'jpg', 'webp')
```
Volcyy: `valid` is a bit very generic. maybe `valid_img_extensions`?

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1231 @@
...
+        # Paste it onto the final image
+        full_image.paste(rectangle, (0, 0), mask=rectangle)
+
+        # Draw the text onto the final image
+        draw = ImageDraw.Draw(full_image)
+        for line in textwrap.wrap(description, 36):
+            draw.text([margin + 4, offset], line, font=CARD['font'])
+            offset += CARD['font'].getsize(line)[1]
+
+        del draw
```
Volcyy: why the manual `del`s? isn't this garbage collected anywaysssssss?

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1231 @@
...
+        await game.open_game()
+
+    @command(name="snakes.about()", aliases=["snakes.about"])
+    async def snake_about(self, ctx: Context):
+        """
+        A command that shows an embed with information about the event,
+        it's participants, and its winners.
+        """
+
+        contributors = [
```
Volcyy: do we need to modify this? if not, why not make it a tuple?

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1231 @@
...
+                "walked away as grand champions. Make sure you check out `bot.snakes.sal()`, `bot.snakes.draw()` "
+                "and `bot.snakes.hatch()` to see what they came up with."
+            )
+        )
+
+        embed.add_field(
+            name="Contributors",
+            value=(
+                ", ".join(contributors)
+            )
```
Volcyy: this seems a bit overkill. why not
```py
embed.add_field(
    name="Contributors",
    value=", ".join(contributors)
)
```
?

bot/converters.py
```diff
@@ -0,0 +1,89 @@
...
+            )
+            cls.snakes = await response.json()
+            http_session.close()
+
+    @classmethod
+    async def random(cls):
+        """
+        This is stupid. We should find a way to
+        somehow get the global session into a
+        global context, so I can get it from here.
```
Volcyy: I've seen two invocations of this that are both within cogs. Can't we pass `ctx.bot.http_session`?

bot/decorators.py
```diff
@@ -46,3 +49,24 @@ def in_channel(channel_id):
...
+    """
+    def wrap(func):
+        func.__locks = WeakValueDictionary()
+
+        @wraps(func)
+        async def inner(self, ctx, *args, **kwargs):
+            lock = func.__locks.setdefault(ctx.author.id, Lock())
+            if lock.locked():
+                return
+            async with func.__locks.setdefault(ctx.author.id, Lock()):
```
Volcyy: `func.__locks` is a `WeakValueDictionary`, and I'm not aware of `WeakValueDictionary.__aenter__ ` being a thing. Am I missing something?

bot/utils/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # Pagination was not initiated, only one page
+        if result.author == ctx.bot.user:
+            # Continue the wait_for
+            result = await list(pending)[0]
+
+        # Love that duplicate code
+        for coro in pending:
+            coro.cancel()
+    except asyncio.TimeoutError:
```
Volcyy: Isn't the `await asyncio.wait` the only line that can raise an `asyncio.TimeoutError`? Wouldn't it be better if we would put only that in the `try`?

bot/utils/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+):
+    """
+    Has the user choose between multiple entries in case one could not be chosen automatically.
+    :param ctx: Context object from discord.py
+    :param entries: List of items for user to choose from
+    :param timeout: Number of seconds to wait before canceling disambiguation
+    :param per_page: Entries per embed page
+    :param empty: Whether the paginator should have an extra line between items
+    :param embed: The embed that the paginator will use.
+    :return: Users choice for correct entry.
```
Volcyy: Since this appears to be a general utility, might be nice to document the `BadArgument` exception raised.

bot/utils/snakes/perlin.py
```diff
@@ -0,0 +1,158 @@
+"""
+Perlin noise implementation.
+Taken from: https://gist.github.com/eevee/26f547457522755cb1fb8739d0ea89a1
+Licensed under ISC
+"""
+import math
+import random
+# Licensed under ISC
```
Volcyy: why's this here twice?

bot/cogs/snakes.py
Volcyy: actually, i forgot that spams console
```py
import contextlib
import os


with contextlib.redirect_stdout(os.devnull):
    import this
ZEN = ''.join(this.d.get(char) or char for char in this.s)
```

1mn: naw. it's a fun way to do it, but infinitely less readable.

bot/__init__.py
1mn: yep, you're right.

bot/cogs/snakes.py
1mn: changed all of these to tuples.

bot/converters.py
1mn: converters isn't a cog, there's no access to `bot`. but I'm sure there is a solution for this. I'll inquire.

bot/decorators.py
1mn: this question is better directed at Ava, I have no idea what this code is doing and just blindly implemented it because Ava tends to know what she's doing.

bot/utils/__init__.py
1mn: actually there are more lines that can cause it, so this is fine.

bot/cogs/snakes.py
Volcyy: :(

bot/converters.py
Volcyy: 👍

bot/utils/__init__.py
Volcyy: oh, okay

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1208 @@
...
+from io import BytesIO
+from typing import Any, Dict
+
+import aiohttp
+import async_timeout
+from discord import Colour, Embed, File, Member, Message, Reaction
+from discord.ext.commands import AutoShardedBot, BadArgument, Context, bot_has_permissions, command
+from PIL import Image, ImageDraw, ImageFont
+
+from bot.constants import ERROR_REPLIES, OMDB_API_KEY, SITE_API_KEY, SITE_API_URL, YOUTUBE_API_KEY
```
JoeBanks13: Could this be better in the format change you used in `bot/cogs/events.py`

```py
from bot.constants import (
    ERROR_REPLIES, OMDB_API_KEY, SITE_API_KEY,
    SITE_API_URL, YOUTUBE_API_KEY
)
```

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1208 @@
...
+                Circle: This means you have a CORRECT ingredient in the WRONG position
+                Cross: This means you have a WRONG ingredient in the WRONG position
+
+        Info:   The game automatically ends after 5 minutes inactivity.
+                You should only use each ingredient once.
+
+        This game was created by Lord Bisk and Runew0lf.
+        """
+
+        def event_check(reaction_: Reaction, user_: Member):
```
JoeBanks13: Seems like this should probably be just a lambda or moved out of the function into the class.

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1208 @@
...
+    @command(name="snakes.draw()", aliases=["snakes.draw"])
+    async def draw(self, ctx: Context):
+        """
+        Draws a random snek using Perlin noise
+
+        Written by Momo and kel.
+        Modified by juan and lemon.
+        """
+
+        def beautiful_pastel(hue):
```
JoeBanks13: Same thing here as above, this should probably go in the class instead of the function.

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1208 @@
...
+        Modified by lemon.
+        """
+
+        def predicate(message):
+            """
+            Check if the message was sent by the author.
+            """
+            return message.author == ctx.message.author
+
+        def get_random_long_message(messages, retries=10):
```
JoeBanks13: Could be moved to class

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1208 @@
...
+        :param ctx: context
+        :param message: If this is passed, it will snakify the message.
+                        If not, it will snakify a random message from
+                        the users history.
+
+        Written by Momo and kel.
+        Modified by lemon.
+        """
+
+        def predicate(message):
```
JoeBanks13: The simplicity of this leads me to think it could be a lambda, maybe even just a lambda passed into the `.filter()` call

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1208 @@
...
+                async for message in ctx.channel.history(limit=500).filter(predicate):
+                    messages.append(message.content)
+
+                message = get_random_long_message(messages)
+
+            # Set the avatar
+            if user.avatar is not None:
+                avatar = f"https://cdn.discordapp.com/avatars/{user.id}/{user.avatar}"
+            else:
+                avatar = (
```
JoeBanks13: Is there any reason we can't use the users default avatar (colored clyde image).

The way to get the link to these is through the `ctx.author.default_avatar_url` attribute.

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1208 @@
...
+
+        # Send the user a video
+        if len(data) > 0:
+            num = random.randint(0, len(data) - 1)
+            youtube_base_url = 'https://www.youtube.com/watch?v='
+            await ctx.channel.send(
+                content=f"{youtube_base_url}{data[num]['id']['videoId']}"
+            )
+        else:
+            log.warning(f"CRITICAL ERROR: YouTube API returned {response}")
```
JoeBanks13: This should definetely return some sort of error to the user.

bot/cogs/snakes.py
```diff
@@ -0,0 +1,1208 @@
...
+        zen_quote = self._snakify(zen_quote)
+
+        # Embed and send
+        embed.description = zen_quote
+        await ctx.channel.send(
+            embed=embed
+        )
+    # endregion
+
+    @snake_card.error
```
JoeBanks13: This doesn't have a region, less of a requirement but could it go in a region such as "Error Handlers" or something similar.

bot/decorators.py
JoeBanks13: Also it might be nice to return something to the user saying the command is already running

bot/utils/snakes/hatching.py
```diff
@@ -0,0 +1,44 @@
...
+        -----
+       -----  \\
+     /--|  /---\\
+     |--\\  -\\---|
+     |--\\--/--  /
+      \------- /
+        ------```'''
+
+stages = [h1, h2, h3, h4]
+snakes = {
```
JoeBanks13: Compared to the immense amount of snakes we have in other parts of the cogs, could we not find any more for this part?

bot/cogs/snakes.py
1mn: yeah okay, this one is simple enough to turn into a lambda.

bot/cogs/snakes.py
1mn: moving to helper method.

bot/cogs/snakes.py
1mn: a bit weird to have a region for a single method, but I'll do it anyway.

bot/utils/snakes/hatching.py
1mn: This is perhaps something that can be PR'd in later. I do agree that it should have more, but I'm not gonna prioritize it now.

bot/cogs/snakes.py
JoeBanks13: :ok_hand:

bot/utils/snakes/hatching.py
JoeBanks13: Sounds good to me

PR #199, opened 2018-11-18
Superstarify modlog

sco1:
Resolves: #97

This leverages existing functionality built into the modlog cog to generate the embed.

This has been tested locally as best as I can, but needs to be evaluated within PyDis' framework before merging.


bot/cogs/superstarify.py
```diff
@@ -145,11 +145,14 @@ def moderation(self) -> Moderation:
...
-                f":middle_finger: {member.name}#{member.discriminator} (`{member.id}`) "
-                f"has been superstarified by **{ctx.author.name}**. Their new nickname is `{forced_nick}`. "
-                f"They will not be able to change their nickname again until **{end_time}**"
+            mod_log_message = (
+                f"{member.name}#{member.discriminator} (`{member.id}`)\n\n"
+                f"Superstarified by **{ctx.author.name}**\n"
+                f"New nickname:`{forced_nick}`\n"
+                f"Superstardom ends: **{end_time}**"
+            )
+            await ModLog.send_log_message(
```
jos-b: This is erroring locally because ModLog.send_log_message is not a static method. This means you are technically missing the `text` argument. The error I receive locally is:
```py
Nov 18 19:16:33 pd.beardfist.com Bot: |                 bot.decorators |    DEBUG | joseph#1337 has the 'Moderators' role, and passes the check.
Nov 18 19:16:33 pd.beardfist.com Bot: |          bot.cogs.superstarify |    DEBUG | Attempting to superstarify Nix for 10s. forced_nick is set to None.
Nov 18 19:16:33 pd.beardfist.com Bot: |          bot.cogs.superstarify |    TRACE | Logging to the #mod-log channel. This could fail because of channel permissions.
Ignoring exception in command_error
  File "/home/joseph/.local/share/virtualenvs/bot-2DLfRnSI/lib/python3.6/site-packages/discord/client.py", line 221, in _run_event
    await coro(*args, **kwargs)
  File "/home/joseph/Python/bot/bot/cogs/events.py", line 140, in on_command_error
    raise e.original
  File "/home/joseph/.local/share/virtualenvs/bot-2DLfRnSI/lib/python3.6/site-packages/discord/ext/commands/core.py", line 61, in wrapped
  File "/home/joseph/Python/bot/bot/cogs/superstarify.py", line 155, in superstarify
    Icons.user_update, Colour.gold(), "Member Achieved Superstardom", mod_log_message
TypeError: send_log_message() missing 1 required positional argument: 'text'
```

sco1: oops, I must have imagined the decorator.

sco1: I guess instantiating an instance of `modlog.ModLog` in `superstarify`'s `__init__` *could* fix this, but would that introduce undesired side effects?

MarkKoz: Look at how all the other cogs do it. They have a property with only a getter than uses `bot.get_cog()` or whatever it is.

sco1: Alternatively, I'm assuming calling the already loaded cog is an option, which would be a bit more obscure but probably smarter.

jos-b: Instantiating a class would fix it and probably wouldn't cause side effects but getting the instantiated class from bot is the preferred way.

sco1: Mirrored what was done to pull in the Moderation cog, should be ok to test now

bot/cogs/superstarify.py
```diff
@@ -145,11 +149,14 @@ def moderation(self) -> Moderation:
...
-                f":middle_finger: {member.name}#{member.discriminator} (`{member.id}`) "
-                f"has been superstarified by **{ctx.author.name}**. Their new nickname is `{forced_nick}`. "
-                f"They will not be able to change their nickname again until **{end_time}**"
+            mod_log_message = (
+                f"{member.name}#{member.discriminator} (`{member.id}`)\n\n"
+                f"Superstarified by **{ctx.author.name}**\n"
+                f"New nickname:`{forced_nick}`\n"
+                f"Superstardom ends: **{end_time}**"
+            )
+            await self.modlog.send_log_message(
```
jos-b: This now works fine and works locally, though maybe it'd be nice to have the thumbnail be the user avatar (as is done with other logs). `send_log_message` accepts a `thumbnail` kwarg so you can make use of that.

Issue #200, opened 2018-11-19
Short-term Moderation Cog Refactor


sco1:
To preface, I'm using "short-term" to acknowledge what @MarkKoz mentioned that he and @scragly have been discussing for a broader, long-term refactor of the moderation cog & the related API(s). I'm not sure what the timeline on this is.

There's quite a bit of duplicate code in the moderation cog currently, mostly around the API call for infractions, and mostly with only slight differences in the JSON package sent to the API. It would be helpful in the short term for issues like #103 if this API call was wrapped into a helper method so the full command doesn't have to be invoked.

This seems like it would be a low-effort pass at the current cog but wanted to get some feedback before I did it. I can also just copy the same API call into bb and call it a day for now.

PR #201, opened 2018-11-20
Add note support for bb watch, moderation API call refactor

sco1:
Replace log-to channel specification with an optional moderation note.

To support this, the moderation cog infraction API calls are refactored to delegate to an external helper utility.

Resolves: #103, #200


sco1:
I know it's been discussed but I don't think it's been formalized in an issue (yet). Might be deferred until the Django rewrite is done.

PR #204, opened 2018-11-21
Remove duration from JSON payload when not specified

sco1:
It doesn't seem like the infraction API likes it when `duration` is specified as `None` in the JSON payload. This should bring the payloads back in line with what they were before the moderation refactor.


gdude2002:
You guys are slow.

PR #212, opened 2018-12-03
Makes generating `...:` indents faster

Modelmat:
Uses `rjust` instead of an fstring. Five is arbitrary.
```py
>>> from timeit import Timer
>>> t = Timer('f"{\'\':<{5}}...:"')
>>> t2 = Timer('"...:".zfill(5)')
>>> t3 = Timer('"...:".rjust(5)')
>>> print(t.timeit(), t2.timeit(), t3.timeit())
0.2464568559998952 0.09237145299994154 0.08822837999969124
```

### DIFF
--- a/bot/cogs/eval.py
+++ b/bot/cogs/eval.py
@@ -66,9 +66,9 @@ class CodeEval:
                 # far enough to align them.
                 # we first `str()` the line number
                 # then we get the length
-                # and do a simple {:<LENGTH}
+                # and use `str.rjust()`
                 # to indent it.
-                start = f"{'':<{len(str(self.ln))+2}}...: "
+                start = "...:".rjust(len(self.ln) + 2)
 
             if i == len(lines) - 2:
                 if line.startswith("return"):


scragly:
The changes made here are not correctly implemented, unfortunately.

This is original:
```python
f"{'':<{len(str(self.ln))+2}}...: "
```

This gets the string length of the line count for the output, adds 2 in order to buffer up to alignment of the top line, and adds that number of spaces to the start of the string to indent. The `...:` section of the string is not considered, as it's only plainly adding spaces to the beginning. It also has a single space after the colon.

The change within this PR is:
```python
"...:".rjust(len(self.ln) + 2)
```

This gets the `len` of the line count which is an int, so raises the following exception:
```
TypeError: object of type 'int' has no len()
```
The int will first need to be cast to a string to get the string length, as the original does.

```python
"...:".rjust(len(str(self.ln)) + 2)
```

Then it adds 2, seemingly because the original adds 2. But `rjust` is justifying the entire section, including the `...:` part, unlike the original. Due to this, the code should have been tested and adjusted to add the correct buffer amount to re-establish alignment with the top line again. After checking, this should have been `7`.

Finally, the alteration leaves off the space after the colon, which should be added again.

```py
"...: ".rjust(len(str(self.ln)) + 7)
```

To confirm the behaviors mentioned:  
image

image

PR #219, opened 2018-12-18
Make reason a required input to bb watch

### LINKED ISSUE #218: It should be mandatory to add a reason for !bb watch
Leaving it optional means we don't always know why someone was watched. This is important information, so we should make it mandatory instead.

sco1:
Resolves #218


sco1:
Removed.

PR #222, opened 2018-12-24
CommandNotFound default to tags get before silently ignoring

tagptroll1:
Before any CommandNotFound exceptions were completely ignored, now it will first try to find a tag by invoking `tags get` before ignoring. If it finds a tag it posts it.  

This would act as a shortcut for all tags that don't overlap with command names


heavysaturn:
Yeah I like this. Let's do it.

PR #223, opened 2018-12-26
Show when an infraction notified a user successfully via DM.

scragly:
I've added an envelope emoji to the infraction response message that indicates if an infraction successfully notified the user.

If the emoji is present, a DM was sent.
If it failed, or the infraction doesn't attempt to notify, it won't be present.

In the example below, the infractions to myself (Scragly) successfully send a DM notification. This includes the un-mute action also. For Eevee, the notification failed and so the emoji is not present.

dm-notification

### DIFF
--- a/bot/cogs/moderation.py
+++ b/bot/cogs/moderation.py
@@ -82,7 +82,7 @@ class Moderation(Scheduler):
         :param reason: The reason for the warning.
         """
 
-        await self.notify_infraction(
+        notified = await self.notify_infraction(
             user=user,
             infr_type="Warning",
             reason=reason
@@ -92,12 +92,13 @@ class Moderation(Scheduler):
         if response_object is None:
             return
 
-        if reason is None:
-            result_message = f":ok_hand: warned {user.mention}."
-        else:
-            result_message = f":ok_hand: warned {user.mention} ({reason})."
+        dm_result = ":incoming_envelope: " if notified else ""
+        action = f"{dm_result}:ok_hand: warned {user.mention}"
 
-        await ctx.send(result_message)
+        if reason is None:
+            await ctx.send(f"{action}.")
+        else:
+            await ctx.send(f"{action} ({reason}).")
 
     @with_role(*MODERATION_ROLES)
     @command(name="kick")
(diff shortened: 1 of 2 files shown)


SebastiaanZ:
I really like this addition to the mod tools.

One small thing I'd do differently is to always include the feedback explicitly in the message instead of relying on the presence/absence of an icon. So, instead of showing the envelope on success and omitting it on failure, I'd always include the envelope, but with a check mark (e.g., :white_check_mark:) or a cross (e.g., :x:) next to it to indicate the success of the PM action.

The reason is that it makes the outcome apparent instead of having to notice an absence (so, relying on memory) when, say, you're handing out a warning.

However, other than that, I think this is a very welcome addition.

SebastiaanZ:
That's fair, @scragly. I didn't really consider the perspective of having failures show up in public channels when writing my comment. I think your example would indeed be too distracting from the actual infraction.

My line of reasoning was this:

Sometimes infractions are handed out in #staff. In those cases, the only way for the user to actually receive the warning or the kick reason is by DM, so knowing if they did is important. However, moderators may not always remember to check and, when the DM actually failed, there's nothing in the bot response to remind them of it. So, in the situation where you actually want someone to notice that the DM failed, there's no explicit reminder in the bot response. 

However, your argument for not displaying failure to regular users may trump my argument for adding explicit feedback. I wonder if there's some middle ground, a subtle symbol that does remind the moderator that a DM wasn't sent, but that isn't really apparent to the "normal" user. I'm not sure what would be fitting, though.

Another option would be to have the bot mention the failure in mod-alerts with a mention of moderator who handed out the infraction (instead of the regular @everyone mention). That way, the "regular" uses will never know, but the moderator in question does get an explicit reminder of failure.

That's said, I'd support your addition as it is as well.

heavysaturn:
how about implicit in the channel, explicit in the mod-logs?

scragly:
Good idea @SebastiaanZ, I didn't think about sending the notice to modlog.
I'll adjust to add it in. Thanks!

scragly:
Added the modlog message that pings the moderator who performed the command when a DM notification fails to send.

image

PR #233, opened 2018-12-29
Add mod log event for member warn & shadowwarn

sco1:
See #137


jchristgit:
I think the icon name is fine. We can revisit it later if we're unhappy with it.

Issue #245, opened 2019-01-02
Notify user on invite filter


sco1:
Currently, when the bot detects an invite link it's just deleted without notifying the user, often leading to them trying multiple times until they either realize it's being deleted intentionally or a staff member steps in. 

How about adjusting the cog so it notifies the user, either in-channel or via a DM, that the invite is being removed per Rule 10? 

For invites DMed to the bot we could either use this same notification, or just defer it to staff action.

heavysaturn:
love this idea. let's do it!

Issue #243, opened 2019-01-01
Account for DM channel filter alerts


scragly:
Filter alerts sometimes are triggered when the bot receives a DM with something triggering the filters, such as advertising server invites. It would show the channel as `#deleted-channel` rather than explicitly saying it's a DM. It would be worth adding a check for and adjusting to help make things clearer for staff when the alert is posted.

PR #251, opened 2019-01-03
Invite filter DM & DMed invite modlog

sco1:
- Add configurable user notifications for message filter actions
  - Booleans added to server configuration, with the exception of the token filter since notification is handled by its own cog
  - If the filter's `"user_notification"` is `True`, a message is DMed to the user on filter action. This falls back to an in-channel mention if the user has DMs disabled
  - User notification message is configurable in the filter dictionary
- Adjust the mod log message to read `"via DM"` rather than `"in <#deleted-channel>"` if a filter is triggered in a DM

Closes: #243
Closes: #245

### DIFF
--- a/bot/cogs/filtering.py
+++ b/bot/cogs/filtering.py
@@ -1,7 +1,7 @@
 import logging
 import re
 
-from discord import Colour, Member, Message, TextChannel
+from discord import Colour, DMChannel, Member, Message, TextChannel
 import discord.errors
 from discord.ext.commands import Bot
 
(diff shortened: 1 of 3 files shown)


SebastiaanZ:
I'm testing this PR and one thing I've noticed is the following:

When you enable the user notification for a filter that has no message specified, the bot will throw the obvious exception:

```
discord.errors.HTTPException: BAD REQUEST (status code: 400): Cannot send an empty message
```

This isn't wrong, I think, but as it's anticipated, we may want to handle it. Maybe we can have the bot send this to modlog or mod_alert so it's obvious to the one that decided to turn the feature on that they forgot a crucial step: Setting the actual message to be send to the user.

Except for that, everything works well here, so if we decide against the thing above, I'll approve it directly.

sco1:
I guess it does make sense to prepopulate them if they're going to be a config option

PR #253, opened 2019-01-05
Eval returns complete traceback. Also doesn't break on comments.

FrenchMasterSword:
This looks like the end of a normal traceback, and it includes the name of the raised exception.

### DIFF
--- a/bot/cogs/snekbox.py
+++ b/bot/cogs/snekbox.py
@@ -30,7 +30,7 @@ exec(open(venv_file).read(), dict(__file__=venv_file))
 try:
 {CODE}
 except Exception as e:
-    print(e)
+    print("{}: {}".format(e.__class__.__name__, e))
 """
 
 ESCAPE_REGEX = re.compile("[`\u202E\u200B]{3,}")
--- a/bot/cogs/snekbox.py
+++ b/bot/cogs/snekbox.py
@@ -30,7 +30,8 @@ exec(open(venv_file).read(), dict(__file__=venv_file))
 try:
 {CODE}
 except Exception as e:
-    print("{}: {}".format(e.__class__.__name__, e))
+    import traceback
+    print(traceback.format_exc(), end='')
 """
 
 ESCAPE_REGEX = re.compile("[`\u202E\u200B]{3,}")
@@ -85,6 +86,11 @@ class Snekbox:
 
         if not code:  # None or empty string
             return await ctx.invoke(self.bot.get_command("help"), "eval")
+
+        if all([line.startswith('#') for line in code.strip('\n')]):  # Only comments
+            return await ctx.send(
+                f"{ctx.author.mention} Your eval job has completed.\n\n```py\n[No output]\n```"
+            )
 
         log.info(f"Received code from {ctx.author.name}#{ctx.author.discriminator} for evaluation:\n{code}")
         self.jobs[ctx.author.id] = datetime.datetime.now()
--- a/bot/cogs/snekbox.py
+++ b/bot/cogs/snekbox.py
@@ -87,11 +87,6 @@ class Snekbox:
         if not code:  # None or empty string
             return await ctx.invoke(self.bot.get_command("help"), "eval")
 
-        if all([line.startswith('#') for line in code.strip('\n')]):  # Only comments
-            return await ctx.send(
-                f"{ctx.author.mention} Your eval job has completed.\n\n```py\n[No output]\n```"
-            )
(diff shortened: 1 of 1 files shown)


bot/cogs/snekbox.py
```diff
@@ -30,7 +30,7 @@
 try:
 {CODE}
 except Exception as e:
-    print(e)
+    print("{}: {}".format(e.__class__.__name__, e))
```
heavysaturn: Please use [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings).

jos-b:
Just an idea but how about we use [`traceback.format_exc`](https://docs.python.org/3/library/traceback.html#traceback.format_exc) or [`traceback.print_exc`](https://docs.python.org/3/library/traceback.html#traceback.format_exc) to print out the errors to maybe give more context when more advanced snippets of code are executed?

```py
>>> try:
...     int("t")
... except Exception as e:
...     traceback.print_exc()
... 
  File "<stdin>", line 2, in <module>
ValueError: invalid literal for int() with base 10: 't'
```

heavysaturn:
yeah I think that's a far better option.

FrenchMasterSword:
At first I wanted to do the normal traceback as you said, but seeing the current way it was implemented I thought it was intended to not put it

I also wanted to use f-strings, but I didn't know if the bot was running on Python 3.6+(you might have been using PyPy which doesn't suppport it) so I used format. 

I completely agree with the full traceback idea.

jos-b:
Yeah we run with 3.6+.

I don't know if we are trying to avoid traceback here. I personally think this is something we should have, any other opinions?

sco1:
I don't see any reason to avoid it. Even if for whatever reason it ends up being really long the output is already truncated going into the embed.

heavysaturn:
Yep, don't avoid it. I doubt this was intentionally omitted.

bot/cogs/snekbox.py
```diff
@@ -30,7 +30,8 @@
 try:
 {CODE}
 except Exception as e:
-    print("{}: {}".format(e.__class__.__name__, e))
+    import traceback
```
sco1: This should be at the top of the file with the rest of the imports

heavysaturn:
This seems like a useful PR, but we need to test it once all review comments are satisfied before we consider merging, since this user doesn't have a test environment set up yet.

FrenchMasterSword:
@sco1 : as @heavysaturn said, I'm currently not able to test this in my own environment. I'm also unsure where the code is actually executed. However, tests with the actual bot reveal that imported modules such as datetime aren't (by default) imported in the eval. Importing traceback at the top of this file doesn't make it available in the executed code.

bot/cogs/snekbox.py
heavysaturn: that wouldn't work, as discussed in #meta.

bot/cogs/snekbox.py
```diff
@@ -86,6 +87,11 @@ def rmq(self) -> RMQ:
         if not code:  # None or empty string
             return await ctx.invoke(self.bot.get_command("help"), "eval")
 
+        if all([line.startswith('#') for line in code.strip('\n')]):  # Only comments
```
scragly: You probably meant `code.split()` here, not `strip`.

The issue though is that this section is before the code has been stripped of the codeblock text, so will always return `False`.

It will also not take into consideration comments that are indented, or more than one comments that have different indent levels.

I kinda don't like the placement of the inline comment `# Only comments` either, as it unnecessarily extends the line length; could do better maybe above the if statement or within it.

Regarding the placement and adjustments required for the comment check, I'll put it in the main review comment.

FrenchMasterSword: Wait, I just can't read 😅 You are right, I meant to use `split`.

scragly: ~~No, I'm saying you have put `strip` _not_ `split`. You need to use `split`, but your code is using the wrong thing on that line.~~

And you say `I'm aware that content was already stripped` but this part is _not_ where the code is already stripped. That is why I put the recommendations for where things should be placed in the main comment.

scragly: This is not resolved yet. The only change made at this point is the auto-commit for the recommended code change removing the print `end` kwarg. The rest still needs addressing manually.

scragly:
To make things easier, I pushed the needed code changes to the PR fork for quicker merging.

PR #255, opened 2019-01-07
!free help channel command

fiskenslakt:
The long overdue !free command which checks for inactivity in all the help channels, and lists them as a suggestion to users to move to a help channel, or switch to one that isn't currently in use.

## How it works

* If invoked outside of a help channel, it will check the timestamp of the last message in each of the help channels. 
* However if invoked in one of the help channels, it will do the same except for the channel it was invoked in. 
  * If no user is mentioned, it will check the second to last message to avoid considering the invocation itself as activity in the channel
  * If a user *is* mentioned, it will check the third to last message by default. The assumption here is that a user asked if the channel was free, or they asked a question in a potentially active channel, and we don't want to consider that message as activity in the channel.
  * Optionally one can override the default of 3 to a number more appropriate for the situation. (eg. if the user who asked if the channel was free or asked their question took up more than one message)

note: I really don't expect that last situation to ever be used, but I included it just in case (for you power users out there). 

## Cooldown

I've set the cooldown for this to once every minute, on a per channel basis. I felt this was the most appropriate, but I am open to other suggestions.

EDIT - Helpers are now exempt from the cooldown

## Time inactive

~~Currently a channel must be inactive for more than 5 minutes to be considered potentially free. This part I was unsure about, and welcome suggestions if it seems too short, or too long.~~

EDIT - Activity timeout is 10 minutes

## Constraints

`seek` must be less than or equal to 10. I see no point in checking the channel history any older than that.
`seek` can't be less than 1 for obvious reasons.

## Wording

Finally, any suggestions on my wording for the responses are encouraged.

### DIFF
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -4,18 +4,17 @@ from datetime import datetime
 from discord import Colour, Embed, Member, utils
 from discord.ext.commands import BucketType, Context, command, cooldown
 
-from bot.constants import Categories
+from bot.constants import Categories, Free
 
 
 log = logging.getLogger(__name__)
 
+PYTHON_HELP_ID = Categories.python_help
+TIMEOUT = Free.activity_timeout
+
 
 class Free:
     """Tries to figure out which help channels are free."""
-
-    PYTHON_HELP_ID = Categories.python_help
-    TIME_INACTIVE = 300
-
     @command(name="free", aliases=('f',))
     @cooldown(1, 60.0, BucketType.channel)
     async def free(self, ctx: Context, user: Member = None, seek: int = 2):
@@ -34,7 +33,7 @@ class Free:
         in an active channel, and we want the message before that happened.
         """
         free_channels = []
-        python_help = utils.get(ctx.guild.categories, id=self.PYTHON_HELP_ID)
+        python_help = utils.get(ctx.guild.categories, id=PYTHON_HELP_ID)
 
         if user is not None and seek == 2:
             seek = 3
@@ -49,7 +48,7 @@ class Free:
                 msg = await channel.history(limit=1).next()   # noqa (False positive)
 
             inactive = (datetime.utcnow() - msg.created_at).seconds
-            if inactive > self.TIME_INACTIVE:
+            if inactive > TIMEOUT:
                 free_channels.append((inactive, channel))
 
         embed = Embed()
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -69,7 +69,7 @@ class Free:
 
             for i, (inactive, channel) in enumerate(sorted(free_channels, reverse=True), 1):
                 minutes, seconds = divmod(inactive, 60)
-                if minutes > 60:
+                if minutes > 59:
                     hours, minutes = divmod(minutes, 60)
                     embed.description += f"{i}. {channel.mention} inactive for {hours}h{minutes}m{seconds}s\n\n"
                 else:
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -119,7 +119,7 @@ class Free:
                 # return to avoid needlessly logging the error
                 return await ctx.reinvoke()
 
-        log.error(error)  # Don't ignore other errors
+        log.exception(error)  # Don't ignore other errors
 
 
 def setup(bot):
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -2,9 +2,10 @@ import logging
 from datetime import datetime
 
 from discord import Colour, Embed, Member, utils
+from discord.ext import commands
 from discord.ext.commands import BucketType, Context, command, cooldown
 
-from bot.constants import Categories, Free
+from bot.constants import Categories, Free, Roles
 
 
 log = logging.getLogger(__name__)
@@ -82,6 +83,26 @@ class Free:
                                  "as one will likely be available soon.**")
 
         await ctx.send(embed=embed)
+
+    @free.error
+    async def free_error(self, ctx: Context, error):
+        """
+        Runs if any error is raised during invocation
+        of !free command. Any error aside from
+        CommandOnCooldown is ignored.
+
+        If error raised is CommandOnCooldown, and the
+        user who invoked has the helper role, reset
+        the cooldown and reinvoke the command.
+        """
+        helpers = ctx.guild.get_role(Roles.helpers)
+
+        if isinstance(error, commands.CommandOnCooldown):
+            if helpers in ctx.author.roles:
+                # reset cooldown so second invocation
+                # doesn't bring us back here.
+                ctx.command.reset_cooldown(ctx)
+                await ctx.invoke(ctx.command)
 
 
 def setup(bot):
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -79,7 +79,7 @@ class Free:
             embed.description += ("**\nThese channels aren't guaranteed to be free, "
                                   "so use your best judgement and check for yourself.")
         else:
-            embed.description = ("**Doesn't look like any channels are available to me. "
+            embed.description = ("**Doesn't look like any channels are available right now. "
                                  "You're welcome to check for yourself to be sure. "
                                  "If all channels are truly busy, please be patient "
                                  "as one will likely be available soon.**")
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -43,10 +43,15 @@ class Free:
         elif not 0 < seek < 10:
             seek = 3
 
+        # Iterate through all the help channels
+        # to check latest activity
         for channel in python_help.channels:
+            # Seek further back in the help channel
+            # the command was invoked in
             if channel.id == ctx.channel.id:
                 messages = await channel.history(limit=seek).flatten()
                 msg = messages[seek-1]
+            # Otherwise get last message
             else:
                 msg = await channel.history(limit=1).next()   # noqa (False positive)
 
@@ -63,12 +68,17 @@ class Free:
         else:
             embed.description = ""
 
+        # Display all potentially inactive channels
+        # in descending order of inactivity
         if free_channels:
             embed.description += "**The following channel{0} look{1} free:**\n\n**".format(
                 's' if len(free_channels) > 1 else '',
                 '' if len(free_channels) > 1 else 's'
             )
 
+            # Sort channels in descending order by seconds
+            # Get position in list, inactivity, and channel object
+            # For each channel, add to embed.description
             for i, (inactive, channel) in enumerate(sorted(free_channels, reverse=True), 1):
                 minutes, seconds = divmod(inactive, 60)
                 if minutes > 59:
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -115,7 +115,10 @@ class Free:
                 # reset cooldown so second invocation
                 # doesn't bring us back here.
                 ctx.command.reset_cooldown(ctx)
-                await ctx.invoke(ctx.command)
+                # return to avoid needlessly logging the error
+                return await ctx.invoke(ctx.command)
+
+        log.error(error) # Don't ignore other errors
 
 
 def setup(bot):
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -118,7 +118,7 @@ class Free:
                 # return to avoid needlessly logging the error
                 return await ctx.invoke(ctx.command)
 
-        log.error(error) # Don't ignore other errors
+        log.error(error)  # Don't ignore other errors
 
 
 def setup(bot):
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -55,7 +55,7 @@ class Free:
                 free_channels.append((inactive, channel))
 
         embed = Embed()
-        embed.colour = Colour.gold()
+        embed.colour = Colour.blurple()
         embed.title = "**Looking for a free help channel?**"
 
         if user is not None:
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -29,6 +29,7 @@ class Free:
         :param seek: How far back to check the last active message.
 
         seek is used only when this command is invoked in a help channel.
+        You cannot override seek without mentioning a user first.
 
         When seek is 2, we are avoiding considering the last active message
         in a channel to be the one that invoked this command.
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -103,13 +103,11 @@ class Free:
     @free.error
     async def free_error(self, ctx: Context, error):
         """
-        Runs if any error is raised during invocation
-        of !free command. Any error aside from
-        CommandOnCooldown is ignored.
-
         If error raised is CommandOnCooldown, and the
         user who invoked has the helper role, reset
         the cooldown and reinvoke the command.
+
+        Otherwise log the error.
         """
         helpers = ctx.guild.get_role(Roles.helpers)
 
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -46,7 +46,7 @@ class Free:
                 messages = await channel.history(limit=seek).flatten()
                 msg = messages[seek-1]
             else:
-                msg = await channel.history(limit=1).next()
+                msg = await channel.history(limit=1).next()   # noqa (False positive)
 
             inactive = (datetime.utcnow() - msg.created_at).seconds
             if inactive > self.TIME_INACTIVE:
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -10,12 +10,14 @@ from bot.constants import Categories, Free, Roles
 
 log = logging.getLogger(__name__)
 
-PYTHON_HELP_ID = Categories.python_help
 TIMEOUT = Free.activity_timeout
 
 
 class Free:
     """Tries to figure out which help channels are free."""
+
+    PYTHON_HELP_ID = Categories.python_help
+
     @command(name="free", aliases=('f',))
     @cooldown(1, 60.0, BucketType.channel)
     async def free(self, ctx: Context, user: Member = None, seek: int = 2):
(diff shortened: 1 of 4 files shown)


bot/cogs/free.py
```diff
@@ -0,0 +1,90 @@
...
+        elif not 0 < seek < 10:
+            seek = 3
+
+        for channel in python_help.channels:
+            if channel.id == ctx.channel.id:
+                messages = await channel.history(limit=seek).flatten()
+                msg = messages[seek-1]
+            else:
+                messages = await channel.history(limit=1).flatten()
+                msg = messages[0]
```
SebastiaanZ: This line and line 45 will throw this exception:

```py
  File "/home/zappa/.local/share/virtualenvs/bot-05f_83Kn/src/discord-py/discord/client.py", line 225, in _run_event
    await coro(*args, **kwargs)
  File "/home/zappa/pydis/bot/bot/cogs/events.py", line 146, in on_command_error
    raise e.original
  File "/home/zappa/.local/share/virtualenvs/bot-05f_83Kn/src/discord-py/discord/ext/commands/core.py", line 61, in wrapped
  File "/home/zappa/pydis/bot/bot/cogs/free.py", line 48, in free
    msg = messages[0]
IndexError: list index out of range
```
If there are not enough messages in the channel.

This may not be an issue now (there are plenty of messages in them), but cleaning/creating help-channels on the fly has been discussed in the past and implementing that could break this. Don't know if it's relevant, though.

bot/cogs/free.py
```diff
@@ -0,0 +1,90 @@
...
+            inactive = (datetime.utcnow() - msg.created_at).seconds
+            if inactive > self.TIME_INACTIVE:
+                free_channels.append((inactive, channel.id))
+
+        embed = Embed()
+        embed.colour = Colour.gold()
+        embed.title = "**Looking for a free help channel?**"
+
+        if user is not None:
+            embed.description = f"**Hey <@{user.id}>!**\n\n"
```
SebastiaanZ: If we do the following, we can let discord.py handle the mention instead of manually parsing it:
```suggestion
            embed.description = f"**Hey {user.mention}!**\n\n"
```

bot/cogs/free.py
```diff
@@ -0,0 +1,90 @@
...
+from discord.ext.commands import BucketType, Context, command, cooldown
+
+
+log = logging.getLogger(__name__)
+
+
+class Free:
+    """Tries to figure out which help channels are free."""
+
+    PYTHON_HELP_ID = 356013061213126657
```
SebastiaanZ: It may be a good idea to move these settings to config-default.yml where similar settings are stored (like the channel ids and timeout/interval settings). That keeps all the configuration stuff/constants in one place.

SebastiaanZ:
Sorry, for the messy reviews, seems like I inadvertently deleted some of my comments, so I added them back in later.

sco1:
Looks cool!

Yes, definitely want a cooldown exemption for Helpers+

bot/cogs/free.py
fiskenslakt: Yep I knew someone would point this out. I think it's probably a good idea to implement this just for the sake of completeness. I too have doubts there will ever be a help channel that's empty, but doesn't mean it's not worth doing!

bot/cogs/free.py
fiskenslakt: This is awesome, I did not know about this.

bot/cogs/free.py
```diff
@@ -0,0 +1,90 @@
...
+
+        if free_channels:
+            embed.description += "**The following channel{0} look{1} free:**\n\n**".format(
+                's' if len(free_channels) > 1 else '',
+                '' if len(free_channels) > 1 else 's')
+
+            for i, channel in enumerate(sorted(free_channels, reverse=True), 1):
+                inactive, ID = channel
+                minutes, seconds = divmod(inactive, 60)
+                if minutes > 60:
```
fiskenslakt:
```suggestion
                if minutes > 59:
```

overlooked this, woops

bot/cogs/free.py
```diff
@@ -0,0 +1,90 @@
...
+            seek = 3
+        elif not 0 < seek < 10:
+            seek = 3
+
+        for channel in python_help.channels:
+            if channel.id == ctx.channel.id:
+                messages = await channel.history(limit=seek).flatten()
+                msg = messages[seek-1]
+            else:
+                messages = await channel.history(limit=1).flatten()
```
tagptroll1: Isn't it better to just call next on the asynciterator here if the limit is 1 anyways?

fiskenslakt: oh yea, good point

bot/cogs/free.py
heavysaturn: There will never be an empty help channel. If we switch a new system where we create help channels on the fly, this command will become irrelevant, and will be removed. So I'm gonna resolve this. I'm all for solving edge cases, but this isn't on the edge, it's off the surface entirely.

bot/cogs/free.py
fiskenslakt: I've moved the ID to the config, but I'm wondering if moving the timeout constant there is overkill? This command ultimately is temporary, and I can't imagine this timeout will be used for anything else.

sco1: Centralizing constants means we don't need to hunt through cogs to find their configuration options if we want to change them.

fiskenslakt: Fair enough, I'll add the timeout to the config as well.

bot/cogs/free.py
```diff
@@ -0,0 +1,89 @@
...
+
+from discord import Colour, Embed, Member, utils
+from discord.ext.commands import BucketType, Context, command, cooldown
+
+from bot.constants import Categories, Free
+
+
+log = logging.getLogger(__name__)
+
+PYTHON_HELP_ID = Categories.python_help
```
heavysaturn: I don't see a point in having these module-level constants when you're importing a constant already. I'd rather see `Categories.python_help` in the class itself.

bot/cogs/free.py
```diff
@@ -0,0 +1,89 @@
...
+TIMEOUT = Free.activity_timeout
+
+
+class Free:
+    """Tries to figure out which help channels are free."""
+    @command(name="free", aliases=('f',))
```
heavysaturn: I think when there's only a single alias, having it be a tuple is pointless. the readability is worse due to the trailing comma, and as we've explored in previous PRs, there's no real difference in performance anyway.

bot/cogs/free.py
```diff
@@ -0,0 +1,89 @@
...
+                if minutes > 59:
+                    hours, minutes = divmod(minutes, 60)
+                    embed.description += f"{i}. {channel.mention} inactive for {hours}h{minutes}m{seconds}s\n\n"
+                else:
+                    embed.description += f"{i}. {channel.mention} inactive for {minutes}m{seconds}s\n\n"
+
+            embed.description += ("**\nThese channels aren't guaranteed to be free, "
+                                  "so use your best judgement and check for yourself.")
+        else:
+            embed.description = ("**Doesn't look like any channels are available to me. "
```
heavysaturn: how about "Doesn't look like any channels are available right now." -- "to me" looks weird.

bot/cogs/free.py
```diff
@@ -0,0 +1,89 @@
...
+                msg = messages[seek-1]
+            else:
+                msg = await channel.history(limit=1).next()   # noqa (False positive)
+
+            inactive = (datetime.utcnow() - msg.created_at).seconds
+            if inactive > TIMEOUT:
+                free_channels.append((inactive, channel))
+
+        embed = Embed()
+        embed.colour = Colour.gold()
```
heavysaturn: it strikes me as inconsistent that this should be gold. probably best to stick with blurple or something.

bot/cogs/free.py
```diff
@@ -0,0 +1,89 @@
...
+
+        if user is not None:
+            embed.description = f"**Hey {user.mention}!**\n\n"
+        else:
+            embed.description = ""
+
+        if free_channels:
+            embed.description += "**The following channel{0} look{1} free:**\n\n**".format(
+                's' if len(free_channels) > 1 else '',
+                '' if len(free_channels) > 1 else 's')
```
heavysaturn: this trailing parenthesis is inconsistent with our prevailing style. we prefer that it goes on the next line.

bot/cogs/free.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        free_channels = []
+        python_help = utils.get(ctx.guild.categories, id=PYTHON_HELP_ID)
+
+        if user is not None and seek == 2:
+            seek = 3
+        elif not 0 < seek < 10:
+            seek = 3
+
+        for channel in python_help.channels:
```
heavysaturn: this could use a block comment

bot/cogs/free.py
```diff
@@ -0,0 +1,89 @@
...
+        embed.colour = Colour.gold()
+        embed.title = "**Looking for a free help channel?**"
+
+        if user is not None:
+            embed.description = f"**Hey {user.mention}!**\n\n"
+        else:
+            embed.description = ""
+
+        if free_channels:
+            embed.description += "**The following channel{0} look{1} free:**\n\n**".format(
```
heavysaturn: this definitely wouldn't be hurt by a block comment explaining what it's doing, the readability isn't amazing but it's not terrible either.

bot/cogs/free.py
```diff
@@ -0,0 +1,89 @@
...
+                's' if len(free_channels) > 1 else '',
+                '' if len(free_channels) > 1 else 's')
+
+            for i, (inactive, channel) in enumerate(sorted(free_channels, reverse=True), 1):
```
heavysaturn: block comment? block comment.

bot/cogs/free.py
fiskenslakt: I was going for the color of the helper role lol

bot/cogs/free.py
scragly: Must have been bad luck. It's definitely prevailing style to close on next line. It's ok, it's minor and easy to resolve 😄

fiskenslakt: Haha yea, lets just hope I don't have someone else try and get me to change it back and I get stuck in a loop.

bot/cogs/free.py
heavysaturn: okay, I'll buy it.

bot/cogs/free.py
heavysaturn: if they do I will kill them.

jos-b:
I'm somewhat concerned about the 5 minute wait before a channel is considered inactive. I have seen a question left for over 10 minutes before but since the channel was considered active and wasn't talked in someone else came along and assisted with it. I'm not sure of a better duration but 5 feels too short, any opinions anyone?

heavysaturn:
@jos-b I think it's important to note that the feature simply lists the channels in order of inactivity and says "hey, these have been inactive for a while so they're probably your best bet but they might be active". the wording isn't very assertive.

that said, I think we could probably set the duration a tiny bit higher, yeah. 5 minutes does seem short, it might even be less than it takes for someone to answer _on average_. I'd probably go for something like 10. I think we probably need to test this in production and tune it afterwards, though.

jos-b:
That makes more sense. I was more thinking about peoples questions being moved into the scroll back and I think that we need to be clear on the wording and say something like "make sure there are no pending questions in the channel" or something.

bot/cogs/free.py
```diff
@@ -0,0 +1,123 @@
...
+        else:
+            embed.description = ("**Doesn't look like any channels are available right now. "
+                                 "You're welcome to check for yourself to be sure. "
+                                 "If all channels are truly busy, please be patient "
+                                 "as one will likely be available soon.**")
+
+        await ctx.send(embed=embed)
+
+    @free.error
+    async def free_error(self, ctx: Context, error):
```
sco1: Because there's a local error handler, our global handler ignores any errors coming out of this cog:

```
Jan 09 18:53:48 pd.beardfist.com Bot: |                bot.cogs.events |    DEBUG | Command free has a local error handler, ignoring.
```

I'm not sure if reraising uncaught errors will help here, but they at least need to be logged.

fiskenslakt: This was my initial thought as well, however scragly told me to ignore it and not bother logging the error. Perhaps we still need to discuss this further.

bot/cogs/free.py
```diff
@@ -0,0 +1,123 @@
...
+TIMEOUT = Free.activity_timeout
+
+
+class Free:
+    """Tries to figure out which help channels are free."""
+
+    PYTHON_HELP_ID = Categories.python_help
+
+    @command(name="free", aliases=('f',))
+    @cooldown(1, 60.0, BucketType.channel)
```
sco1: Cooldown should be in the server config

bot/cogs/free.py
sco1: It needs to at least be debug logged, otherwise we can't debug the cog without suppressing  the error handler.

bot/cogs/free.py
fiskenslakt: Ah yes, that's probably a good idea.

bot/cogs/free.py
```diff
@@ -0,0 +1,123 @@
...
+    async def free(self, ctx: Context, user: Member = None, seek: int = 2):
```
sco1: Might need a better way to handle these inputs, d.py can [only handle one kwarg](https://discordpy.readthedocs.io/en/rewrite/ext/commands/commands.html#keyword-only-arguments) so if member is missing then this will error.

fiskenslakt: Ah ok, gotcha

bot/cogs/free.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        Runs if any error is raised during invocation
+        of !free command. Any error aside from
+        CommandOnCooldown is ignored.
+
+        If error raised is CommandOnCooldown, and the
+        user who invoked has the helper role, reset
+        the cooldown and reinvoke the command.
+        """
+        helpers = ctx.guild.get_role(Roles.helpers)
```
sco1: Is there a case where someone with roles above Helper won't have the Helper role?

sco1:
Other than the outstanding review comments this seems to work great:

image
image
image

bot/cogs/free.py
sco1: After chatting with the other admins, it seems like this is an edge case likely not worth considering at the moment. 

If it ends up becoming an issue we can just make a fix then.

bot/cogs/free.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if isinstance(error, commands.CommandOnCooldown):
+            if helpers in ctx.author.roles:
+                # reset cooldown so second invocation
+                # doesn't bring us back here.
+                ctx.command.reset_cooldown(ctx)
+                # return to avoid needlessly logging the error
+                return await ctx.invoke(ctx.command)
+
+        log.error(error)  # Don't ignore other errors
```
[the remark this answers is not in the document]
fiskenslakt: Ahhhh, this is good to know. Yea that's much better.

bot/cogs/free.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        helpers = ctx.guild.get_role(Roles.helpers)
+
+        if isinstance(error, commands.CommandOnCooldown):
+            if helpers in ctx.author.roles:
+                # reset cooldown so second invocation
+                # doesn't bring us back here.
+                ctx.command.reset_cooldown(ctx)
+                # return to avoid needlessly logging the error
+                return await ctx.invoke(ctx.command)
```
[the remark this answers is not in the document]
fiskenslakt: haha dammit, I was debating with myself over which one was correct to use. I'm wondering if this sentence in the docs means I should not reset the cooldown first?

>This is similar to invoke() except that it bypasses checks, cooldowns, and error handlers.

My thinking is that it if a helper invokes this command with the cooldown active, they would reset the cooldown for everyone else since when the command is reinvoked, presumably the invocation from the helper wouldn't count towards the cooldown, allowing normal users to then use the command right after a helper since it would have been reset to zero just beforehand. Thus a user could use the command, then a helper directly after, and then a user could use the command right after that, having the command invoked 3 times within a short period of time.

Obviously I don't see this issue cropping up, but if I'm right and the cooldown doesn't increment during a `reinvoke`, then I might as well remove it.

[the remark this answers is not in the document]
SebastiaanZ: That's an interesting point. I did not consider that.

However, I did some testing and found that a regular user (non-helper) can call `!free` after a helper has bypassed the cooldown with the error function, regardless of whether `invoke` or `reinvoke` was used in the error function. It seem like these two functions may differ in their ability to bypass checks, but that they both don't start a new cooldown period.

So, the cooldown doesn't work after a bypass, as a new period doesn't get triggered by the `invoke` and `reinvoke` commands. (See below, this is with your `ctx.invoke` version) That means we either need to code a custom cooldown system that doesn't suffer from that or accept that it may happen.

I lean to the latter at the moment (accept it), as I think a custom cooldown system is an entire project in itself and should be reusable throughout the code. It does mean we can just use `reinvoke` to propagate the arguments.

no_cooldown_for_nonhelper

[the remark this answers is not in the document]
heavysaturn: When we get around to starting our `bot-core` project properly, a better (universal) cooldown system might be a good fit for that. For now, I think I'll resolve this.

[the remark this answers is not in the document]
SebastiaanZ: If you're going to resolve, I'd recommend us using `ctx.reinvoke` to pass the parameters instead of `ctx.invoke` since it doesn't matter for the cooldown, but does for the parameters getting passed.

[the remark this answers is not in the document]
heavysaturn: yeah you're right. let's do that @fiskenslakt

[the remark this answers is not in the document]
SebastiaanZ: They should be able to commit from the suggest code above, I think. That's what the suggestion feature is supposed to do. Saves another push and may keep your review from going stale.

PR #260, opened 2019-01-07
Add mute infraction check to role restoration

sco1:
Adds a check for active mute infractions for the case when a joining user is one who left before their mute role could be removed, leaving them permamuted until an admin can remove the role. For this case, the muted role persistence is skipped if no active mute infraction is present.

Closes: #149

### DIFF
--- a/bot/cogs/events.py
+++ b/bot/cogs/events.py
@@ -25,6 +25,7 @@ class Events:
 
     def __init__(self, bot: Bot):
         self.bot = bot
+        self.headers = {"X-API-KEY": Keys.site_api}
 
     @property
     def mod_log(self) -> ModLog:
@@ -102,6 +103,29 @@ class Events:
 
         resp = await response.json()
         return resp["data"]
+
+    async def has_active_mute(self, user_id: str) -> bool:
+        """
+        Check whether a user has any active mute infractions
+        """
+        response = await self.bot.http_session.get(
+            URLs.site_infractions_user.format(
+                user_id=user_id
+            ),
+            params={"hidden": "True"},
+            headers=self.headers
+        )
+        infraction_list = await response.json()
+
+        # Check for active mute infractions
+        if len(infraction_list) == 0:
+            # Short circuit
+            return False
+
+        muted_check = any(
+            [infraction["active"] for infraction in infraction_list if infraction["type"].lower() == "mute"]
+        )
+        return muted_check
 
     async def on_command_error(self, ctx: Context, e: CommandError):
         command = ctx.command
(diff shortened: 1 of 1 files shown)


bot/cogs/events.py
```diff
@@ -103,6 +104,30 @@ def mod_log(self) -> ModLog:
...
+        )
+        infraction_list = await response.json()
+
+        # Check for active mute infractions
+        if len(infraction_list) == 0:
+            # Short circuit
+            return False
+
+        muted_check = any(
+            [infraction["active"] for infraction in infraction_list if infraction["type"].lower() == "mute"]
```
SebastiaanZ: https://github.com/python-discord/bot/blob/d5cb479be6925a10cf0c46e3088518f56a5a91b3/bot/cogs/events.py#L126-L128

Since any accepts a generator, we don't need to build the list first with a list comp:

```suggestion
            infraction["active"] for infraction in infraction_list if infraction["type"].lower() == "mute"
```

sco1: Oh, duh, that's much smarter :)

bot/cogs/events.py
```diff
@@ -236,6 +261,10 @@ def mod_log(self) -> ModLog:
 
                 for role in RESTORE_ROLES:
                     if role in old_roles:
+                        # Check for mute roles that were not able to be removed and skip if present
+                        if role == str(Roles.muted) and not await self.has_active_mute(str(member.id)):
+                            continue
```
jchristgit: I think some logging here would be nice.

bot/cogs/events.py
```diff
@@ -103,6 +104,30 @@ def mod_log(self) -> ModLog:
...
+            URLs.site_infractions_user.format(
+                user_id=user_id
+            ),
+            params={"hidden": "True"},
+            headers=self.headers
+        )
+        infraction_list = await response.json()
+
+        # Check for active mute infractions
+        if len(infraction_list) == 0:
```
jchristgit: Prefer `if not infraction_list`

bot/cogs/events.py
```diff
@@ -103,6 +104,30 @@ def mod_log(self) -> ModLog:
...
+
+        # Check for active mute infractions
+        if not infraction_list:
+            # Short circuit
+            return False
+
+        muted_check = any(
+            infraction["active"] for infraction in infraction_list if infraction["type"].lower() == "mute"
+        )
+        return muted_check
```
jchristgit: How about renaming this to the something more explanatory like `is_muted` or just returning from `any` directly?

Issue #262, opened 2019-01-09
The base !help-command does not work in staff channels


SebastiaanZ:
At the moment, the base `!help` command does not work in staff channels. A cursory search in staff indicates that it broke somewhere after 2018-12-17, as this is the last time it was successful there:

![2019-01-09_17-52-27](https://user-images.githubusercontent.com/33516116/50914799-ac6b9800-1437-11e9-8b2a-05b5dac45f87.png)

Currently, the bot just doesn't respond when `!help` is called without arguments. Strangely enough, it does work when used with an argument, .e.g, `!help unban`, whether called directly or as fallback when a command is run without the proper arguments, e.g. `!unban` without arguments.

I've not found any obvious reason for it, but I know @scragly also looked, so maybe he can share his findings as well.

SebastiaanZ:
Okay, the base command can only be called from bot-commands at the moment, so the problem is not restricted to the staff channels. Seems like a whitelisting problem somewhere.

SebastiaanZ:
Running the command `!help` on my test/sandbox server show this in the debug logs:

```
Jan 09 18:29:47 pd.beardfist.com Bot: |                 bot.decorators |    DEBUG | Ves Zappa#3787 tried to call the 'charinfo' command. The command was not used in a whitelisted channel, but the author had a role to bypass the in_channel check.
```

That's weird. I'll look into it.

SebastiaanZ:
Okay, I think I know what's going on.

The help command checks which commands a user can run to determine which should be in the help embed. However, the way it does that throws the `bot.decorators.InChannelCheckFailure` exception for some commands when executed outside of bot-commands. Since that exception is not handled, the help command stops executing.

I'll work on a patch.

sco1:
Closed by #263

PR #263, opened 2019-01-09
Patch to make !help work outside of #bot-commands again

SebastiaanZ:
This patches `!help` not working outside of the bot-commands channel. I think a proper rewrite of the procedure may be warranted, but this will do in the mean time. See issue #262 

Basically what happened is that the `help` cog runs all the checks on the commands to see if the user requesting the help-embed is allowed to run it to determine which commands should be included. Apparently, one of those checks raises an exception when the check fails, which stops function execution, preventing the help-embed from showing up.

I'm now catching it with a `try-except` and saying that the user can't run the command if the exception was raised. That solves the bug, but creates another one that should be solved later (and is less severe, in my opinion):

If a user ask for `!help` outside of bot-commands, commands that are only allowed in bot-commands do not show up since they user is not allowed to run that command in the channel where they are requesting the help-embed. I don't think that's right, so we should probably restructure the checks.

However, I'd rather push this out first and rewrite the `can_run` check later.

### DIFF
--- a/bot/cogs/help.py
+++ b/bot/cogs/help.py
@@ -9,10 +9,12 @@ from discord.ext import commands
 from fuzzywuzzy import fuzz, process
 
 from bot import constants
+from bot.decorators import InChannelCheckFailure
 from bot.pagination import (
     DELETE_EMOJI, FIRST_EMOJI, LAST_EMOJI,
     LEFT_EMOJI, LinePaginator, RIGHT_EMOJI,
 )
+
 
 REACTIONS = {
     FIRST_EMOJI: 'first',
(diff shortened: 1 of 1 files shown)


bot/cogs/help.py
```diff
@@ -427,7 +429,15 @@ def _get_command_params(self, cmd):
...
                     # see if the user can run the command
                     strikeout = ''
-                    can_run = await command.can_run(self._ctx)
+
+                    # Patch to make the !help command work outside of #bot-commands again
+                    # This probably needs a proper rewrite, but this will make it work in
+                    # the mean time.
+                    try:
+                        can_run = await command.can_run(self._ctx)
+                    except InChannelCheckFailure:
```
jchristgit: Shouldn't we apply this to all check failures?

SebastiaanZ: I think this is the only one that raises an exception to be able to return a custom message to the user. The other checks return `True` or `False`, which was the expected return to assign to `can_run`.

SebastiaanZ: But, we may want to think of something in case other checks with exceptions get added.

PR #265, opened 2019-01-09
Add infraction hierarchy check

### LINKED ISSUE #134: Moderation commands don't compare role hierachy of caller and "moderated"
**Originally posted by Thomas Petersson:**

Since the bot straight out bans, kicks, etc. after posting an infraction, won't it be possible for moderators to ban admins. Given the bots role is above admins.  This makes it more of a trust based implementation over restricted.  

In my opinion admins should not be able to moderate other admins, and moderators should definitely not be able to moderate admins (owners are above the bot in the hierarchy and can't be effected)

attempted murder

If it was not for the fact that joseph is an owner (and super duper owner) and the bot is under owner role in the hierarchy that may have very well gone through. (We were curious, didn't intend to try to ban Joseph)

sco1:
Per #134, add a hierarchy check to all permutations of the kick & ban infractions to prevent infraction of someone with a higher top role.

This is implemented as a helper method rather than a Check in order to avoid duplication of input conversion and validation, as handled by the commands extension.

Closes #134

### DIFF
--- a/bot/cogs/moderation.py
+++ b/bot/cogs/moderation.py
@@ -176,7 +176,8 @@ class Moderation(Scheduler):
         :param reason: The reason for the ban.
         """
 
-        if not await self.respect_role_hierarchy(ctx, user, 'ban'):
+        member = ctx.guild.get_member(user.id)
+        if not await self.respect_role_hierarchy(ctx, member, 'ban'):
             # Ensure ctx author has a higher top role than the target user
             # Warning is sent to ctx by the helper method
             return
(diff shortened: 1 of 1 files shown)


bot/cogs/moderation.py
```diff
@@ -1207,6 +1241,30 @@ def _infraction_to_string(self, infraction_object):
...
+        If this check fails, a warning is sent to the invoking ctx
+
+        Implement as a method rather than a check in order to avoid having to reimplement parameter
+        checks & conversions in a dedicated check decorater
+        """
+
+        # Build role hierarchy
+        actor = ctx.author
+        role_hierarchy = {role: rank for rank, role in enumerate(reversed(ctx.guild.roles))}
+        hierarchy_check = role_hierarchy[actor.top_role] <= role_hierarchy[target.top_role]
```
jchristgit: can't we just compare the top roles directly? see https://discordpy.readthedocs.io/en/rewrite/api.html#role

sco1: Sorry, forgot about this.

PR #266, opened 2019-01-10
Add watch reason to big brother header embed & fix small bug in watch command

SebastiaanZ:
**Summary:**
This PR enables staff to see the reason a user is being watched while reading big-brother-logs. The code fetches the reasons out of the infraction table at start-up, caches it and serves it from cache when building the embed. This PR also fixes a small bug in the !bigbrother watch command.

**How it looks in action**
_If a reason was specified:_
Hemlock's Cat with watch reason

_If a reason wasn't specified (for backwards compatibility):_
Hemlock's cat without a watch reason

**The reason is retrieved by looking for the prefix**
Since we store the watch reason as a `note` with a specific prefix, I'm using the prefix to search for the reason. The code than selects the latest reason set by comparing the dates.

A side-effect of this is that if we were to change the prefix, the old watch reasons will no longer be found by the code. I've therefore moved the prefix to the \_\_init\_\_ of the class with a comment not to change it in the future. 

**Small bug fix**
In the old code, if the API-call to insert the user into the watch list failed, the `error_message` would overwrite the `reason` parameter before setting the `note` infraction. This caused the big-brother watch reason to be set to the API `error_message`. To mitigate this, I've renamed the variable name for the `error_message` to `error_reason`. I've also moved the infraction code so that an infraction will only be set if they were actually added to big-brother.

**Update to config-default.yml**
While the API route for infraction_user_type was given a key in `constants.py`, it wasn't actually set in `config-default.yml`. I've added it so I can use it to search the infraction table for the bb watch reason.

**update_cache is now async**
Because the `update_cache` method now needs to fetch the watch reasons from the API, I've made it a coroutine. I've updated all the calls to it so it gets properly `await`ed.

### DIFF
--- a/bot/cogs/bigbrother.py
+++ b/bot/cogs/bigbrother.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 import re
 from collections import defaultdict, deque
-from time import strptime
+from time import strptime, struct_time
 from typing import List, Union
 
 from aiohttp import ClientError
@@ -99,7 +99,7 @@ class BigBrother:
         return "(no reason specified)"
 
     @staticmethod
-    def _parse_time(infraction):
+    def _parse_time(infraction: str) -> struct_time:
         """Takes RFC1123 date_time string and returns time object for sorting purposes"""
 
         date_string = infraction["inserted_at"]
--- a/bot/cogs/bigbrother.py
+++ b/bot/cogs/bigbrother.py
@@ -2,8 +2,10 @@ import asyncio
 import logging
 import re
 from collections import defaultdict, deque
+from time import strptime
 from typing import List, Union
 
+from aiohttp import ClientError
 from discord import Color, Embed, Guild, Member, Message, TextChannel, User
 from discord.ext.commands import Bot, Context, group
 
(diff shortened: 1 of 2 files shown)


SebastiaanZ:
I noticed that the alias we had defined for `!bigbrother watch <user> <reason>`, the `!watch`, still had a default argument for `reason` (`None`). I removed it to make sure a reason is really mandatory.

sco1:
For discussion: though assuming the existence of the watch prefix is a great simplification, it may still be still worth considering a fallback since most of the current watches were probably made before the watch reason was made mandatory.

SebastiaanZ:
> For discussion: though assuming the existence of the watch prefix is a great simplification, it may still be still worth considering a fallback since most of the current watches were probably made before the watch reason was made mandatory.

What do you mean by that? 

The watches for which no reason was specified currently show up as `(no reason specified)` as a fallback, so to say. I don't think we can handle it in a different way, since there was no reason given for those users. (Although I could be missing something obvious.)

sco1:
They may have other infractions that give context

SebastiaanZ:
That's true, maybe we could add the last infraction along side the watch reason.

Personally, I do like to keep the two separate, though: The watch reason was really intended as a reason for the thing we're looking at and the last infraction may not even be relevant. Unfortunately, the embed doesn't allow for two lines in the footer, otherwise we could've just listed the last infraction and the reason on separate lines.

SebastiaanZ:
@sco1 This could be an alternative embed lay-out:

Hemlock's Cat alternative lay-out for headed embed

I'm just worried that if the embed gets too large, it'll hurt readability in bb

heavysaturn:
I don't like the fat embed at all, but I do quite like the one in your original PR. I'm not too worried about a transitional period where old watches don't have reasons. That's a problem that will solve itself, given time. If it's particularly frustrating, we can just rewatch them to sort it out.

SebastiaanZ:
Right, something else that came to my mind is that getting the watch reasons for everyone is really not efficient. It's probably better to cache it the first time a user becomes active, so the bot doesn't have to query the API for everyone at start-up. Most of them are no longer active anyway.

SebastiaanZ:
I have kept the original, small embed for now (see the opening comment), since there doesn't seem to an agreement to change it at the moment and I'm leaning towards that option as well.

PR #268, opened 2019-01-11
Revert "Catching the superclass CheckFailure instead of the subclass"

SebastiaanZ:
This reverts commit 8f30b52fd378bb3546c84d0fdb945a4b492236b8.

Accidentally pushed to master when I thought I was working on a branch to change the exception we catch in help.py.


sco1:
Skipping review since it's a reversion

PR #269, opened 2019-01-11
Disable rich embed filter

sco1:
Discord adding the embeds is causing a false positive. This flips the config flag until we can triage the issue.


sco1:
Skipping review due to severity of the issue & very low risk of this change breaking the bot

PR #272, opened 2019-01-12
Add "last edited" delta to modlog for multi-edit messages

sco1:
After seeing some curious edit histories in recent logs I thought it might be nice to see the edit frequency in the `on_message_edit` event modlogs.

This PR makes two major changes:

- For messages that have been previously edited, add a human-readable delta to the `before` message modlog. The delta is calculated as (`after.edited_at` - `before.edited_at`).
- Use the Message timestamp for the modlog embed rather than the time 
  - `after.edited_at` is always used for the `after` log embed
  - For messages that have not been edited, `before.created_at` is used for the `before` log embed
  - For messages that have been edited, `before.edited_at` is used for the `before` log embed

Some minor changes were also made to the `send_log_message` helper:

- Add `timestamp_override` and `footer_override` to enable the above embed changes. This change should be transparent to all other uses of this helper.
- Visually separated the input kwargs to make the function definition possible to read

Log messages for raw message edit events remain unmodified, pending investigation of the payload to see if they can be similarly reworked.

A sample:
image
image
image

### DIFF
--- a/bot/cogs/modlog.py
+++ b/bot/cogs/modlog.py
@@ -104,9 +104,19 @@ class ModLog:
                 self._ignored[event].append(item)
 
     async def send_log_message(
-            self, icon_url: Optional[str], colour: Colour, title: Optional[str], text: str,
-            thumbnail: str = None, channel_id: int = Channels.modlog, ping_everyone: bool = False,
-            files: List[File] = None, content: str = None, additional_embeds: List[Embed] = None,
+            self,
+            icon_url: Optional[str],
+            colour: Colour,
+            title: Optional[str],
+            text: str,
+            thumbnail: str = None,
+            channel_id: int = Channels.modlog,
+            ping_everyone: bool = False,
+            files: List[File] = None,
+            content: str = None,
+            additional_embeds: List[Embed] = None,
+            timestamp_override: datetime.datetime = None,
+            footer_override: str = None,
     ):
         embed = Embed(description=text)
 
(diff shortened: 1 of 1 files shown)


sco1:
I also toyed around with adding millisecond precision to `bot.utils.time.humanize_delta` for extra special sleuthing, but it seems more noisy than helpful:

image

bot/cogs/modlog.py
```diff
@@ -104,17 +104,34 @@ def ignore(self, event: Event, *items: int):
...
     ):
         embed = Embed(description=text)
 
         if title and icon_url:
             embed.set_author(name=title, icon_url=icon_url)
 
         embed.colour = colour
-        embed.timestamp = datetime.datetime.utcnow()
+
+        if timestamp_override:
```
AnonGuy: Could this read better as:
```python
embed.timestamp = timestamp_override or datetime.datetime.utcnow()
```

bot/cogs/modlog.py
```diff
@@ -104,17 +104,34 @@ def ignore(self, event: Event, *items: int):
...
     async def send_log_message(
-            self, icon_url: Optional[str], colour: Colour, title: Optional[str], text: str,
-            thumbnail: str = None, channel_id: int = Channels.modlog, ping_everyone: bool = False,
-            files: List[File] = None, content: str = None, additional_embeds: List[Embed] = None,
+            self,
+            icon_url: Optional[str],
+            colour: Colour,
+            title: Optional[str],
+            text: str,
+            thumbnail: str = None,
```
AnonGuy: Shouldn't the typehint for these (and the other `None` values) be `Optional[str]`?

PR #276, opened 2019-01-14
A talent pool for potential helpers

SebastiaanZ:
**Summary**
This PR adds a talent-pool for potential helpers to our server. A potential helper can be nominated with the `!nominate <user> <reason>` command by users with the role of helpers and up, after which they will be automatically added to the talent-pool channel. The idea behind the talent-pool channel is that it will help the staff cast an informed vote on the nominated helper.

**Header Embed**
While the talent-pool uses the same underlying code as bb, the header embed is slightly modified to include information on which helper+ nominated the potential helper and when. After some experimenting with the lay-out, this seems to be the most readable and compact format:

talent-pool header embed

**The command**
The command to nominate someone is straightforward and gives feedback about the result:

Nominate command in action

**The command in the help embed**
I've chosen to create a separate class/cog in `bigbrother.py` to hold the command. The reason for that is that ensures that the command gets a nice header in the help embed instead of showing up under the rather unfriendly sounding header of BigBrother:

Command under HelperNomination in help embed

The command in HelperNomination is role_restricted to helper+ and will then invoke the `!bigbrother nominate` command with the right parameters. Doing this ensures that the BigBrother section doesn't suddenly show up for helpers.

### DIFF
--- a/bot/cogs/bigbrother.py
+++ b/bot/cogs/bigbrother.py
@@ -113,8 +113,12 @@ class BigBrother:
         return WatchInformation(reason="(no reason specified)", actor_id=None, inserted_at=None)
 
     @staticmethod
-    def _parse_infraction_time(infraction: str) -> struct_time:
-        """Takes RFC1123 date_time string and returns time object for sorting purposes"""
+    def _parse_infraction_time(infraction: dict) -> struct_time:
+        """
+        Helper function that retrieves the insertion time from the infraction dictionary,
+        converts the retrieved RFC1123 date_time string to a time object, and returns it
+        so infractions can be sorted by their insertion time.
+        """
 
         date_string = infraction["inserted_at"]
         return strptime(date_string, "%a, %d %b %Y %H:%M:%S %Z")
--- a/bot/cogs/bigbrother.py
+++ b/bot/cogs/bigbrother.py
@@ -3,21 +3,28 @@ import logging
 import re
 from collections import defaultdict, deque
 from time import strptime, struct_time
-from typing import List, Union
+from typing import List, NamedTuple, Optional, Union
 
 from aiohttp import ClientError
 from discord import Color, Embed, Guild, Member, Message, TextChannel, User
-from discord.ext.commands import Bot, Context, group
+from discord.ext.commands import Bot, Context, command, group
 
 from bot.constants import BigBrother as BigBrotherConfig, Channels, Emojis, Guild as GuildConfig, Keys, Roles, URLs
 from bot.decorators import with_role
 from bot.pagination import LinePaginator
 from bot.utils import messages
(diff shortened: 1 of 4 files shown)


sco1:
What's the point of a separate cog if we're just going to import pretty much all of it from the bb cog?

SebastiaanZ:
The main reason for doing like this instead of an alias is to separate the two in the help menu. Instead of having the nominate command show under `BigBrother`, it now shows under `HelperNomination`. I think that's cleaner for the `helpers` that normally don't see `BigBrother` in their help-embeds and I don't like the negative connotation that goes with the `Monitor users, NSA-style` docstring of the `BigBrother` class. 

So, the separate cog thing is basically just to be able to create an alias that shows up as a root element in the help embed for helpers (and the other roles as well, obviously). 

Since that's the only reason and the functionality is identical, I thought it was wise to keep it all in one file, instead of the one cog/class per file convention of Java. However, I can change that if more people like that part changed. For me, I like the effect of it (separate help category) over the consistency and I don't think a separate file containing basically only an alias makes sense to me. The functional grouping is much more clear this way.

heavysaturn:
that seems like a good reason to me. I fully agree.

bot/cogs/bigbrother.py
```diff
@@ -84,16 +91,26 @@ def update_cache(self, api_response: List[dict]):
...
-            return latest_reason
+            return WatchInformation(reason=latest_reason, actor_id=actor_id, inserted_at=date)
 
-        log.trace(f"No bb watch reason found for {user_id}; returning default string")
-        return "(no reason specified)"
+        log.trace(f"No bb watch reason found for {user_id}; returning defaults")
+        return WatchInformation(reason="(no reason specified)", actor_id=None, inserted_at=None)
 
     @staticmethod
     def _parse_infraction_time(infraction: str) -> struct_time:
```
sco1: I'm not sure why this hasn't errored out before, but it errors for me when testing.

We've got `infraction` typed as `str` and then attempt to index it as a dictionary:

https://github.com/python-discord/bot/blob/0f501deddfc1ec5bf82b5bfdc1dc07dfd63b1a8e/bot/cogs/bigbrother.py#L115-L120

SebastiaanZ: Oh, that's an oversight from me; the type annotation should say dictionary. I'll also update the docstring to reflect that. 

It's basically a helper function to avoid a longer inline lambda to be able to sort the infraction dictionaries based on a converted key.

SebastiaanZ: Done.

sco1:
Turns out I was getting an error because I didn't configure the bot correctly, which just happened to spit out at the time parser. Whoops...

Issue #280, opened 2019-01-17
Truncate large mod alerts


GhostofGoes:
Currently, if there is a mod alert for an excessively large message which violates a filter (e.g. 600 newlines), the entire message is sent. This is problematic, since it fills the channel and makes it impossible for moderators to effectively see their notifications on a mobile device.

One possible solution would be to truncate anything longer than 50 lines or so, and save the full message elsewhere.

PR #285, opened 2019-01-18
Redirect newline antispam log message, add consecutive newline check

sco1:
In order to avoid modlog messages being dominated by messages with excessive newlines, messages caught by the newline AntiSpam rule are now redirected to the logs API rather than the log channel. This was chosen instead of truncating the message in order to retain context on the message for human reviewing.

A consecutive newline check is also added to the rule. This checks consecutive runs of newline characters in a message against the configured threshold (set to `10`), and triggers the rule if this is exceeded.

Closes: #280


sco1:
I'm not 100% sure how to accomplish the second thing without making changes to the log API endpoint. Since we push the "punish" invocation into the background we don't have access to ID returned by the API when we post the logs so we'd end up inserting the messages twice.

That said, shouldn't the message delete events be getting ignored?

https://github.com/python-discord/bot/blob/5f0a7a738de6a1a38600ccac730df0749d4c7dad/bot/cogs/antispam.py#L152

On a more general note, `message_change_logs` is, by nature, pretty spammy so I'm not sure if these events occur often enough to be that distracting.

GhostofGoes:
Message change logs is generally used for searching after the fact. If the changes required are non-trivial, then I don't think it's necessary.

scragly:
I confirmed the message delete ignoring was working in its current form.
This wasn't lemon's fault for thinking that. I believe he was checking in my server, and there's a strange situation where if the bot doesn't have the right permissions, it failed to correctly register the ignore which isn't actually an issue in the main server.

I'm not 100% sure why as to why it happened, as I didn't look too closely into it to check for the issue as it won't occur during production and didn't have the time.

PR #289, opened 2019-01-20
WIP Remove RMQ, Add snekbox API support.

### LINKED ISSUE #281: Change snekbox to not use RMQ.
With the removal of RMQ, snekbox currently won't be able to receive public code evals for running and returning.

An alternate system needs to be put in place beforehand so we can safely and smoothly proceed with the RMQ removal.

scragly:
Closes #281

Related `snekbox` PR:
https://github.com/python-discord/snekbox/pull/7

This fully removes RMQ from the bot code, as well as adds the necessary changes to send a POST request to snekbox instead in the correct format, and to parse the response for sending back to the user.

I also cleaned the code up to conform to the 100 col standard expected with projects.

This is marked as WIP to avoid merging early, as snekbox has not been tested, merged or checked over regarding the production setup. Once that is done, it can be released for final checks and merging.

If you want to test this PR, you must pull the `snekbox/rmq_removal` branch instead of the default, run in snekbox project directory:
```
pipenv run build
docker-compose up -d
```
Which then you can check it's live on `http://localhost:8060`.

After which you can pull the `bot/rmq_removal` branch, and run it as normal alongside `site`. The default bot config assumes the above localhost url is correct for snekbox with the `/input` endpoint. If you change it, just add the new url in `config.yml` under URLs > snekbox.


heavysaturn:
I'm gonna close this PR, and @scragly will do the same thing to the django branch of the bot instead.

Issue #217, opened 2018-12-18
remove with_roles decorator and replace with has_any_role


lemonsaurus:
Initially, when we wrote the `with_roles` decorator, discord.py's own `has_any_role` did not support IDs as input. Around a month ago (in [this commit](https://github.com/Rapptz/discord.py/commit/016963500b6f19b5c8978f135fce9dd264872007)), this was changed. 

That means, there's no longer any point in us having our own decorator for this. Our decorator should be replaced with `has_any_role`. See https://discordpy.readthedocs.io/en/rewrite/ext/commands/api.html#discord.ext.commands.has_any_role for documentation.

heavysaturn:
Okay, so, I no longer think we should do this.

The discord.py framework does not provide a negated version of this. We have both `with_role` and `without_role`, and we use them both a lot. Changing our `with_role` only would mean that we'd have `has_any_role` and `without_role` decorators that no longer matched, and would be imported from different places. This is uglier than what we've currently got.

Closing this ticket, since it hasn't seen any discussion anyway.

GhostofGoes:
Would suggesting the addition of `without_role` functionality to discord.py be worthwhile?

heavysaturn:
I don't see why not.

heavysaturn:
Okay, so after a chat with @scragly, here's an approach that would solve all of the concerns raised above:

We import the new `has_any_role` straight into our `decorators.py`, then we remove our `with_role` decorator and rewrite our `without_role` to just use `has_any_role` but negate the result. That way the user can import both from the same place, they both have logically similar names, and you get all the speed benefits of the new decorator (which is O(1) instead of O(n)).

kosayoda:
**From #236:**
Some cogs use the `cog_check` method to add a check to every command in the cog. The `cog_check` method is the predicate itself, which poses a problem if we were to use `has_any_role`, which is decorator check.

One solution would be to nab [`has_any_role`](https://github.com/Rapptz/discord.py/blob/f513d831d137d29ee4638a5e6a218f70b56c5ac5/discord/ext/commands/core.py#L1364) from discordpy, place it in `decorators.py`, and split the predicate and check.

That way we get the goodness and the usefulness of having both a decorator and a predicate for `cog_check`s. I can work on this if we're moving forward with this idea.

scragly:
@kosayoda I think that's a good solution. If you get a chance to do it, that'll be great. I've assigned you to the ticket.

MarkKoz:
[...]

PR #298, opened 2019-01-31
WIP Has any role from discord to with and without_role

tagptroll1:
As per issue #217 

This only rewrites the existing decorators to use discords implementation of has_any_role.

Still need to review it on a running environment before I can remove the WIP with confidence


heavysaturn:
This isn't the approach I was envisioning. My idea was far simpler - to just import `has_any_role` from discord and literally call it. I'm not sure why we would copypaste their approach.

This only affects the decorators and should not be in conflict with the django rewrite, so there's no reason to close it.. but please justify your approach.

jchristgit:
Hello @tagptroll1, any updates on this?

lemonsaurus:
No follow-up since February, I'm closing this.

PR #301, opened 2019-02-02
Added more functionality to rules command

### LINKED ISSUE #257: Have !rules (and other alternatives of rules) return the specific rule if a number is provided
`!rules 3` returns a shortwritten version of the rule.  
For even more flavor let it handle "slicing" of rules like `3,5,7` to print the rule titles of rule 3, 5 and 7, or a syntax like `3-6` to get rule 3 to and including 6

Refisio:
Resolves #257 by adding the requested functionality.

- The updated `rules` command still provides the same functionality that the original did by calling the command without added parameters.
- The updated `rules` command allows you to request specific rules by calling the rule number, e.g. `!rules 5`. Called-for rules are split using `/`, `,`, and ` ` (single whitespace). This observes that rules can be called using `, ', and will act accordingly.
- The updated `rules` command allows you to request a range of rules by calling them using the `-` and `:` characters. e.g. `!rules 4:7`, `!rules 3-5`. This does *not* allow for the command to be called through the appearance of list slicing. e.g. `!rules [3-5]`. Both of these numbers will be ignored, and thus, the command will fall back to its original output.
- If someone decides to be silly with the command and call such rules as `!rules lemon 3:5`, the bot will act accordingly and ignore lemon.
- The `rules` command can be aliased through the aliases `r` and `rule`, in case some people are lazy, or want to look grammatically correct when calling for one single rule.


bot/cogs/rules.py
```diff
@@ -0,0 +1,139 @@
+import re
+from typing import Optional
+
+from bot.constants import Channels, Roles
+from bot.decorators import redirect_output
+from bot.pagination import LinePaginator
+
+from discord import Colour, Embed
+from discord.ext.commands import Bot, Context, command
```
scragly: The order and structure of imports are meant to be:
```py
import stdlib
from stdlib import lib

import thirdparty
from thirdparty import lib

import local
from local importlib
```

You have the `bot.` local group before the thirdparty group containing `discord.` and will need to be switched around.

Refisio: Let's thank Github Mobile for allowing me to switch to Desktop mode in order to change that now before I forget about it later.

bot/cogs/rules.py
```diff
@@ -0,0 +1,138 @@
...
+                "We do not allow advertisements for communities (including "
+                "other Discord servers) or commercial projects - Contact "
+                "us directly if you want to discuss a partnership!"
+            )
+        ]
+        self.default_desc = "The rules and guidelines that apply to this " \
+                            "community can be found on our [rules page]" \
+                            "(https://pythondiscord.com/about/rules). We " \
+                            "expect all members of the community to have " \
+                            "read and understood these."
```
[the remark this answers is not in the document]
scragly: This still needs adjusting to match the styling of the block above it, in order to bring it in line with the typical project style.
```py
        self.default_desc = (
            "The rules and guidelines that apply to this community can be found on our [rules page]"
            "(https://pythondiscord.com/about/rules). We expect all members of the community to "
            "have read and understood these."
        )
```

bot/cogs/rules.py
```diff
@@ -0,0 +1,105 @@
...
+        rules_to_get = []
+        split_rules = re.split(r'[/, ]', rules)
+        for item in split_rules:
+            if not item.isdigit():
+                if not item:
+                    continue
+                rule_match = re.search(r'\d?\d[:|-]1?\d', item)
+                if rule_match:
+                    a, b = sorted(
+                        [int(x)-1 for x in re.split(r'[:-]', rule_match.group())])
```
scragly: The hanging indented line continuation isn't necessary.
```py
                    a, b = sorted([int(x)-1 for x in re.split(r'[:-]', rule_match.group())])
```

PR #304, opened 2019-02-05
Have invite filter send guild info with mod-alert.

scragly:
When the invite filter is triggered, we get notified in `#mod_alerts`. Almost every time, a staff member copies the invite and pastes it in a normal message in order to show the guild name and member counts to see where the invite leads to. Unfortunately, it's not possible to copy embed content on mobile either, so if a mod on mobile wants to see the info, they'll have to manually type it out which can be annoying.

This PR makes the bot fetch these details and sends an embed with them automatically instead, automating the whole thing and hopefully bringing a little less annoyance to those on mobile.

image


bot/cogs/filtering.py
```diff
@@ -282,7 +301,7 @@ def mod_log(self) -> ModLog:
 
         return bool(re.search(ZALGO_RE, text))
 
-    async def _has_invites(self, text: str) -> bool:
+    async def _has_invites(self, text: str) -> Union[dict, bool]:
```
sco1: This docstring should be updated with the new behavior

PR #307, opened 2019-02-08
Unwatch reason / Nomination improvements

### LINKED ISSUE #290: minor !nominate improvements
* When a user is nominated multiple times, it would be great if this showed in the #talent-pool embed. Something like "Nominated by: Scragly, lemon, fiskenslakt" would make it clear that the user had been nominated multiple times, which should count for something. The reason should not change, but a note should be added.

* It should not be possible to nominate someone who is already a Helper, Mod, Admin or Owner.

* The notes that are added currently are prefixed with `bb watch`. Perhaps for nominations, these could be prefixed with `Helper nomination:` or something.

SebastiaanZ:
This PR is meant to do a couple of things:
- As requested by @heavysaturn, unwatching (both bb and talent-pool) a user now requires a reason; the reason will be added as a note the user's record.
- Previously, a double bb watch would overwrite the old reason; this PR prevents that and notifies the actor.
- Information on how long a user has been in BB is now added to the header
- Previously, a nomination could overwrite a bb-watch; I've added a check for that.
- If a user receives a secondary nomination, then that will no longer overwrite the initial nomination and time. Instead, additional nominations will be added as a note with the prefix "Additional nomination: "
- To ensure the cache is up to date before checking if a user is already watched/nominated, the cache is updated. To do so, I moved the updating functionality out of the `!watched` command to a separate function to keep it DRY.

This should close #290 and the in-channel request by Lemon.

**Note**: Users who are currently nominated will lose their reason, as I've updated the prefix. I'll manually edit their notes later to update them to the current prefix.

### DIFF
--- a/bot/cogs/alias.py
+++ b/bot/cogs/alias.py
@@ -74,7 +74,7 @@ class Alias:
             self, ctx: Context, user: User, *, reason: str
     ):
         """
-        Alias for invoking <prefix>bigbrother watch user [reason].
+        Alias for invoking <prefix>bigbrother watch [user] [reason].
         """
 
         await self.invoke(ctx, "bigbrother watch", user, reason=reason)
@@ -82,7 +82,7 @@ class Alias:
     @command(name="unwatch", hidden=True)
     async def bigbrother_unwatch_alias(self, ctx, user: User, *, reason: str):
         """
-        Alias for invoking <prefix>bigbrother unwatch user [reason].
+        Alias for invoking <prefix>bigbrother unwatch [user] [reason].
         """
 
         await self.invoke(ctx, "bigbrother unwatch", user, reason=reason)
--- a/bot/cogs/alias.py
+++ b/bot/cogs/alias.py
@@ -74,7 +74,7 @@ class Alias:
             self, ctx: Context, user: User, *, reason: str
     ):
         """
-        Alias for invoking <prefix>bigbrother watch user reason.
+        Alias for invoking <prefix>bigbrother watch user [reason].
         """
 
         await self.invoke(ctx, "bigbrother watch", user, reason=reason)
@@ -82,7 +82,7 @@ class Alias:
     @command(name="unwatch", hidden=True)
     async def bigbrother_unwatch_alias(self, ctx, user: User, *, reason: str):
         """
-        Alias for invoking <prefix>bigbrother unwatch user reason.
+        Alias for invoking <prefix>bigbrother unwatch user [reason].
         """
 
         await self.invoke(ctx, "bigbrother unwatch", user, reason=reason)
@@ -106,7 +106,7 @@ class Alias:
     @command(name="reload", hidden=True)
     async def cogs_reload_alias(self, ctx, *, cog_name: str):
         """
-        Alias for invoking <prefix>cogs reload cog_name.
+        Alias for invoking <prefix>cogs reload [cog_name].
 
         cog_name: str - name of the cog to be reloaded.
         """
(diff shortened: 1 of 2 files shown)


SebastiaanZ:
I forgot to include the check for current staff members for the nomination command. I'll add it.

bot/cogs/bigbrother.py
```diff
@@ -391,8 +425,29 @@ def _parse_infraction_time(infraction: dict) -> struct_time:
...
 
+        # Update watch cache to avoid overwriting active nomination reason
+        await self.update_watched_users()
+
+        if user.id in self.watched_users:
+            if self.watched_users[user.id].id == Channels.talent_pool:
+                prefix = "Additional nomination: "
+            else:
+                # If the user is being watched in big-brother, don't add them to talent-pool
+                await ctx.send(f":x: Can't add {user.mention} to the talent-pool at this moment")
```
sco1: This could be more specific

SebastiaanZ: Yes, I was debating with myself what kind of message to put there. Since the command is available from helper and up, this is currently the only command that reveals information about a user's big brother watch status that's available to non-mod+ roles. 

However, since this is the only way this message can appear anyway, it's not hard to deduce the implication of it anyway, so stating it explicitly probably won't hurt.

SebastiaanZ: I've changed the message.

bot/cogs/alias.py
```diff
@@ -74,20 +74,18 @@ def __init__(self, bot):
             self, ctx: Context, user: User, *, reason: str
     ):
         """
-        Alias for invoking <prefix>bigbrother watch user [text_channel].
+        Alias for invoking <prefix>bigbrother watch user reason.
```
heavysaturn: should maybe `reason` here be encased in some sort of brackets to indicate it's not supposed to be written literally? kind of like how text_channel was in square brackets?

bot/cogs/alias.py
```diff
@@ -74,20 +74,18 @@ def __init__(self, bot):
...
         await self.invoke(ctx, "bigbrother watch", user, reason=reason)
 
     @command(name="unwatch", hidden=True)
-    async def bigbrother_unwatch_alias(self, ctx, user: User):
+    async def bigbrother_unwatch_alias(self, ctx, user: User, *, reason: str):
         """
-        Alias for invoking <prefix>bigbrother unwatch user.
-
-        user: discord.User - A user instance to unwatch
+        Alias for invoking <prefix>bigbrother unwatch user reason.
```
heavysaturn: same as previous comment.

bot/cogs/bigbrother.py
```diff
@@ -73,10 +77,20 @@ def update_cache(self, api_response: List[dict]):
             data = await response.json()
             self.update_cache(data)
 
-    async def get_watch_information(self, user_id: int) -> WatchInformation:
+    async def update_watched_users(self):
+        async with self.bot.http_session.get(URLs.site_bigbrother_api, headers=self.HEADERS) as response:
+            if response.status == 200:
+                data = await response.json()
+                self.update_cache(data)
+                log.trace("Updated watch list cache")
```
heavysaturn: this log could be improved. without the context of big brother, it might be a little hard to understand exactly what this means.
```suggestion
                log.trace("Updated Big Brother watchlist cache")
```

bot/cogs/alias.py
SebastiaanZ: Done.

bot/cogs/alias.py
SebastiaanZ: Done.

PR #310, opened 2019-02-13
tox.ini renamed .flake8 and line-length rule from 120 to 100

tagptroll1:
tox.ini -> .flake8  
linelength rule 120 -> 100  
Applied new rule to all files triggering linter.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-max-line-length=120
+max-line-length=100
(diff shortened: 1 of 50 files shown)


heavysaturn:
I'm closing this because it's a huge pull request which introduces several bugs. It's too risky. I can't review this properly, none of us can without risking missing something critical. I also don't think it improves readability that much.

PR #311, opened 2019-02-16
Add team creation command

jos-b:
This PR adds a `!createteam` command to create a team channel for the code jam.

image

It creates the jams category if it does not exist, adds the jammer and team leader roles and then adds overwrites to the channel so that team members can see it and team leaders can pin and add Webhooks in it.

### DIFF
--- a/bot/__main__.py
+++ b/bot/__main__.py
@@ -78,6 +78,7 @@ bot.load_extension("bot.cogs.token_remover")
 bot.load_extension("bot.cogs.utils")
 bot.load_extension("bot.cogs.wolfram")
 bot.load_extension("bot.cogs.free")
+bot.load_extension("bot.cogs.jams")
 
 if has_rmq:
     bot.load_extension("bot.cogs.rmq")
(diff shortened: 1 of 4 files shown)


bot/__main__.py
```diff
@@ -78,6 +78,7 @@
 bot.load_extension("bot.cogs.utils")
 bot.load_extension("bot.cogs.wolfram")
 bot.load_extension("bot.cogs.free")
+bot.load_extension("bot.cogs.jams")
```
jchristgit: `:sort` please

bot/cogs/jams.py
```diff
@@ -0,0 +1,83 @@
...
+
+from bot.constants import Roles
+from bot.decorators import with_role
+
+log = logging.getLogger(__name__)
+
+
+class CodeJams:
+    """
+    A cog for managing the code-jam related parts of our server
```
jchristgit: cog is implied. prefer `Code jam related management tasks`

heavysaturn:
```suggestion
    Manages the code-jam related parts of our server
```

bot/__main__.py
jos-b: `:sorted

PR #312, opened 2019-02-16
Jam creation command improvements

jos-b:
Add voice channel support to the team creation command and fix a bug which meant some users did not have permissions added.

Earlier today during team creation we noticed that some of the channels did not have all members in because the second member or third member of the command was not valid, this meant that the greedy converter stopped on the first and therefore no team permissions were assigned for certain members.

### DIFF
--- a/bot/cogs/jams.py
+++ b/bot/cogs/jams.py
@@ -35,9 +35,9 @@ class CodeJams:
         # two members or at some times even 1 member. This fixes that by checking that there
         # are always 3 members in the members list.
         if len(members) < 3:
-            await ctx.send(f":no_entry_sign: One of your arguments was invalid - there must be a "
+            await ctx.send(":no_entry_sign: One of your arguments was invalid - there must be a "
                            f"minimum of 3 valid members in your team. Found: {len(members)} "
-                           f"members")
+                           "members")
             return
 
         code_jam_category = utils.get(ctx.guild.categories, name="Code Jam")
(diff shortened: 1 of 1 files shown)


bot/cogs/jams.py
```diff
@@ -24,11 +24,22 @@ def __init__(self, bot: commands.Bot):
...
         """
+
+        # We had a little issue during Code Jam 4 here, the greedy converter did it's job
+        # and ignored anything which wasn't a valid argument which left us with teams of
+        # two members or at some times even 1 member. This fixes that by checking that there
+        # are always 3 members in the members list.
+        if len(members) < 3:
+            await ctx.send(f":no_entry_sign: One of your arguments was invalid - there must be a "
+                           f"minimum of 3 valid members in your team. Found: {len(members)} "
+                           f"members")
```
jchristgit: Does this have to be an f-string?

bot/cogs/jams.py
```diff
@@ -24,11 +24,22 @@ def __init__(self, bot: commands.Bot):
...
 
         The first user passed will always be the team leader.
         """
+
+        # We had a little issue during Code Jam 4 here, the greedy converter did it's job
+        # and ignored anything which wasn't a valid argument which left us with teams of
+        # two members or at some times even 1 member. This fixes that by checking that there
+        # are always 3 members in the members list.
+        if len(members) < 3:
+            await ctx.send(f":no_entry_sign: One of your arguments was invalid - there must be a "
```
jchristgit: Does this have to be an f-string?

bot/cogs/jams.py
jos-b: yes. otherwise flake8 will error and I will have to make it look like
```py
blah(f"abc {d}"
        "world")
```

PR #313, opened 2019-02-21
Use of Constants and Adjusting !user permissions

LordHemlock:
Changed !user so that Helpers don't get redirected to the bot_commands channel.

Added the STAFF_ROLES constant to constants.py

Adjusted any cog use of MODERATION_ROLES or STAFF_ROLES that would have been redundant with the addition of the constants.

### DIFF
--- a/bot/cogs/antispam.py
+++ b/bot/cogs/antispam.py
@@ -11,8 +11,8 @@ from bot.cogs.modlog import ModLog
 from bot.constants import (
     AntiSpam as AntiSpamConfig, Channels,
     Colours, DEBUG_MODE, Event,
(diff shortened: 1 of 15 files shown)


GhostofGoes:
Is  WIP the GitHub check? If so, how the heck do we change the status?

MarkKoz:
> 
> 
> Is WIP the GitHub check? If so, how the heck do we change the status?

It's a 3rd-party app for GitHub. It's supposed to detect when the title of the PR changes but it seems to be broken. My guess is this is an issue on their (the devs of the app) end, not ours.

SebastiaanZ:
No, this is the old WIP check we already had and it's failed to run. I'm adding and removing WIP from the title to trigger it.

PR #314, opened 2019-02-22
Hemlock another role tweak

LordHemlock:
Just a couple quick role constant instances that I missed my first time through.


jchristgit:
Thanks!

PR #316, opened 2019-02-22
Removed STAFF constant, added footer to paginated embed

Refisio:
Removed the `STAFF` constant within the cog in favor of the `STAFF_ROLES` constant present in `bot.constants`.
Added a footer to the paginated embed that points to the Rules page of the site, as the default embed does.
Very slight refactoring: moved the `set_footer` line to the default embed instead of having it set before-hand every time.

### DIFF
--- a/bot/cogs/rules.py
+++ b/bot/cogs/rules.py
@@ -65,11 +65,11 @@ class Rules:
         **`rules`:** The rules a user wants to get.
         """
         rules_embed = Embed(title='Rules', color=Colour.blurple())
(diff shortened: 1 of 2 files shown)


GhostofGoes:
The STAFF_ROLE constant update is already in Hemlock's PR: https://github.com/python-discord/bot/pull/314

heavysaturn:
After some discussion, we decided to change this from a url in the footer to a link in the header. Changes were committed straight to the fork.

PR #330, opened 2019-03-09
utils: Added a command so user's are able to ask for the direct link …

samip5:
…to bot's source code.

This is so the source code link can be found easilly.

Signed-off-by: samip5 <samip5@users.noreply.github.com>

### DIFF
--- a/bot/cogs/utils.py
+++ b/bot/cogs/utils.py
@@ -88,6 +88,13 @@ class Utils:
 
         await ctx.message.channel.send(embed=pep_embed)
 
+    @command(name='src', aliases='source')
+    async def return_source_link(self, ctx: Context):
+        """
+        Returns a link to bot's sources on Github.
+        """
+        await ctx.message.channel.send("The bot's source code can be found at: https://github.com/python-discord/bot/")
+
     @command()
     @in_channel(Channels.bot, bypass_roles=STAFF_ROLES)
     async def charinfo(self, ctx, *, characters: str):


sco1:
Thanks for the suggestion! Our bot is currently on a feature freeze, and this would be better as a tag regardless. 

You can find the meta thread here: https://github.com/python-discord/meta/issues/22

samip5:
> Thanks for the suggestion! Our bot is currently on a feature freeze, and this would be better as a tag regardless.
> 
> You can find the meta thread here: [python-discord/meta#22](https://github.com/python-discord/meta/issues/22)

Could you please elaborate on why this would be better as a tag as even [R.Danny](https://github.com/Rapptz/RoboDanny) bot has a source command?

MarkKoz:
Because your proposed command functions exactly like a tag: when the command is entered, the bot responds with the same exact message every time. Because we already have a tag system in place, it doesn't make sense to "hard-code" your proposed command.

heavysaturn:
I wouldn't be interested in this for the same reasons @sco1 outlined, however there has been mention of writing a `!source` command which would be able to find the source for any part of the bot. e.g. `!source kick` would return a URL to the part of the bot where that command was defined.

If I'm not mistaken, @scragly was going to implement this after the feature freeze.

scragly:
I may have forgotten to make an issue ticket on that, @heavysaturn. Will do that now. This would effectively show the bot source in a similar manner to this PR, but with the rest of a featureset on top.

PR #338, opened 2019-03-24
Invite filter: Remove spoiler syntax

gdude2002:
Because I saw dood doing it innocently in ot0 - best to fix it before someone does it maliciously.

This was a quick edit with the GH editor.


sco1:
This would probably be better located earlier in the process so the spoiler tags are removed before the message hits any of the filters since the vector (where relevant) is essentially the same for all of them:

https://github.com/python-discord/bot/blob/4555d90b60eedbb6a3032512277f307d69034f6f/bot/cogs/filtering.py#L151-L152

Thanks for the catch.

gdude2002:
> This would probably be better located earlier in the process so the spoiler tags are removed before the message hits any of the filters since the vector (where relevant) is essentially the same for all of them:
> 
> [bot/bot/cogs/filtering.py](https://github.com/python-discord/bot/blob/4555d90b60eedbb6a3032512277f307d69034f6f/bot/cogs/filtering.py#L151-L152)
> 
> Lines 151 to 152 in [4555d90](/python-discord/bot/commit/4555d90b60eedbb6a3032512277f307d69034f6f)
>  if _filter["content_only"]: 
>      triggered = await _filter["function"](msg.content) 
> 
> Thanks for the catch.

I extracted the sanitizing stuff to a static method (which this class is full of), and it's now used to sanitize the input for all content-only filters.

jchristgit:
@gdude2002 sorry for the late review, can you change this PR to head into django instead?

gdude2002:
I tried, but it had hundreds of commits. Since this is such a small changeset, you guys might as well just yoink it and add it in directly.

PR #344, opened 2019-04-05
Added ping command

ajgrinds:
Added ping command which can be used to ping a role so there is no fear of it being maliciously pinged if forgotten to be made un-ping-able again.

This may or may not have happened today and I wanted to show I was apologetic :joy:


scragly:
The is not the approach we wish to go with.

PR #345, opened 2019-04-05
Made announcement command and edit command

ajgrinds:
I added this to my bot, so I figured why not add it here. It announces a message and pings the role and allows you to edit the message as well.

### DIFF
--- a/bot/cogs/ping.py
+++ b/bot/cogs/ping.py
@@ -1,8 +1,9 @@
 import logging
 from discord.ext.commands import Bot, command, Context, BadArgument, MissingPermissions
-from discord import Role
+from discord import Role, NotFound, Forbidden
 from bot.decorators import with_role
 from bot.constants import MODERATION_ROLES
+from discord.utils import get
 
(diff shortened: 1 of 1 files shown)


scragly:
As mentioned in the last PR, we don't wish to go with this type of implementation. It's not the inability to edit the message that was the issue, but the fact that we don't wish to rely on a bot for every announcement made in the future.

Ensuring a human is doing the announcement allows for a more personalised feel for our community announcements, and gives a point of contact if people wish to ask about any specifics about the announcement made.

Due to the above reasons, this and the previous PR's are not accepted and won't be part of the possible options that are considered in our next staff meeting that will establish what option we wish to go with. We will decide internally from 5 possible choices that we have gotten already, and implement the decision made ourselves. If this involves a code change to the bot, we will personally do so and have strict reviewing among ourselves during that, so this isn't a thing you should worry about in the future.

For now, the owners of the community have already established that anyone who have or will abuse the mentionable roles will end up with a permanent ban. This has already been actioned for any past cases.

PR #352, opened 2019-04-15
Preventing the ot-names cycling bug

### LINKED ISSUE #351: Bot changing off-topic channel names more than once per daily cycle
The bot should change the channel names of the three off-topic channels only once after hitting UTC midnight. However, we've noticed that it may attempt to set new channel names far more often than that. The root of the cause is still unknown, but after looking at the recent audit log screenshotted by Scragly, I've come up with a conjecture to what's happening.

If you take a look at the logs below, then you'll notice that most of the channel name changes actually happen just before the whole hour (9:59 in Scragly's local time). My guess is that the sleep duration, which is set based on "the start of the current day" + "1 day time delta", is off by one second. The odd thing is that it's not obvious to me from the code why this happens.

However, if this is true, then it would explain the current behavior: The background task is triggered at 23:59:59 (UTC), it will cycle the channel names, and, **as we're still on the same day**, calculate a sleep second delta based on the previous midnight (which will be 0, since it's trying to find the seconds to 23:59:59, which it already is), so it will async.sleep for 0 seconds and run itself again.

The fact that it doesn't trigger more often is then caused because it needs to make API calls and has other `await` points in the execution. 

Since this behavior very much looks like what Discord would call "API Abuse", I think it's important to put in a bug fix as quickly as possible. Since the only thing I can think of is the off-by-one error described above, my proposal is aim for 1 minute past midnight, so we never run into "we're still on the same day" issues again:

```py
        today_at_midnight = datetime.utcnow().replace(microsecond=0, second=0, minute=0, hour=0)
        next_midnight = today_at_midnight + timedelta(days=1, minutes=1)
        seconds_to_sleep = (next_midnight - datetime.utcnow()).seconds
        await asyncio.sleep(seconds_to_sleep)
```

If that doesn't fix it, we need to investigate further, but I think this will work.

---

**Edit: some additional confirmation of the off-by-one-second error:**
```
Apr 15 08:08:35 pd.beardfist.com Bot: |       bot.cogs.off_topic_names |    DEBUG | update_names: seconds to sleep 24
```

I'd set the name change for `08:09:00`, but sleeping for 24 seconds at `08:08:35` will cause it to "wake up" at `08:08:59`.

---

Audit Log Screenshot showing channel name changes

SebastiaanZ:
This PR is meant to solve the bug in #351 ~~by sleeping until 1 minute after midnight instead of aiming for midnight itself. I've outlined the rationale for this in the linked issue. I'm not entirely sure if this is actually the problem, but given the evidence/audit log times we have, this may very well be the cause.~~

Since the effect of this bug is something that Discord will probably call "API abuse" (rapid API calls not triggered by user action), we should probably try to merge this soon.

**Edit: I've found the root cause of the issue**

The problem is that in this step:

```py
seconds_to_sleep = (next_midnight - datetime.utcnow()).seconds
```

We truncrate the time delta needed to reach midnight down to the seconds, which means we "wake up" slightly before midnight (`< 1 second`). This is demonstrated here:

code:
```py
from datetime import datetime, timedelta

mn = datetime.utcnow().replace(microsecond=0, second=0, minute=0, hour=0)
mn2 = mn + timedelta(days=1)
delta_seconds = (mn2 - datetime.utcnow()).seconds
delta_full = mn2 - datetime.utcnow()

print(delta_full)
print(delta_seconds)
print(delta_full - timedelta(seconds=delta_seconds))

print(datetime.utcnow() + timedelta(seconds=delta_seconds))
```

output:
```
16:50:59.998919              # Full time needed to reach midnight
60659                        # Truncated down to seconds
0:00:00.998919               # The part we're missing by truncating the delta
2019-04-15 23:59:59.001132   # When we'll wake up
```

To fix this, I'm now simply adding one second to the computed sleep time so we always go over midnight:

```py
seconds_to_sleep = (next_midnight - datetime.utcnow()).seconds + 1
```

This now closes #351, as the bug was found and fixed.

### DIFF
--- a/bot/cogs/off_topic_names.py
+++ b/bot/cogs/off_topic_names.py
@@ -50,11 +50,8 @@ async def update_names(bot: Bot, headers: dict):
     # To ensure we only cycle once per day, we increase the reference midnight point
     # by one day each time the task runs instead of relying on "the most recent"
     # midnight, since that may fail if the task is triggered early.
-    midnight = None
+    midnight = midnight = datetime.utcnow().replace(microsecond=0, second=0, minute=0, hour=0)
     while True:
-        if midnight is None:
-            midnight = datetime.utcnow().replace(microsecond=0, second=0, minute=0, hour=0)
-
         midnight += timedelta(days=1)
         seconds_to_sleep = (midnight - datetime.utcnow()).seconds
         log.debug(f"update_names: seconds to sleep {seconds_to_sleep}")
--- a/bot/cogs/off_topic_names.py
+++ b/bot/cogs/off_topic_names.py
@@ -48,8 +48,11 @@ async def update_names(bot: Bot, headers: dict):
     """
 
     while True:
+        # To prevent the bot from sleeping until one second before midnight, we aim
+        # for one minute past midnight. This should prevent the ot-names cycling bug.
+
         today_at_midnight = datetime.utcnow().replace(microsecond=0, second=0, minute=0, hour=0)
-        next_midnight = today_at_midnight + timedelta(days=1)
+        next_midnight = today_at_midnight + timedelta(days=1, minutes=1)
         seconds_to_sleep = (next_midnight - datetime.utcnow()).seconds
         await asyncio.sleep(seconds_to_sleep)
 
--- a/bot/cogs/off_topic_names.py
+++ b/bot/cogs/off_topic_names.py
@@ -47,13 +47,17 @@ async def update_names(bot: Bot, headers: dict):
             website via the bot's `http_session`.
     """
 
+    # To ensure we only cycle once per day, we increase the reference midnight point
+    # by one day each time the task runs instead of relying on "the most recent"
+    # midnight, since that may fail if the task is triggered early.
+    midnight = None
     while True:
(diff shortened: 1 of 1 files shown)


SebastiaanZ:
> Should this PR just be squashed? I think so.

Oh, right. I tried to revert first, but apparently I did not succeed. My git-fu is not that honed. How do we go about this?

MarkKoz:
To revert I usually do an interactive rebase since that is what I am familiar with, but `git revert` is the command. I did initially think that the initial commit could be reverted to and we'd call it a day, but I see that you added a second instead of adding a minute like you were initially. Therefore, even if you reverted, it'd still involve and additional commit to change from minute to second. 

That being said I still wouldn't have bothered with cleaning up my suggestion since you were just going to scrap the entire thing anyway (and if you had already commited the clean-up but not pushed, you could have dropped the commit through an interactive rebase or other means).

I do think it should be squashed anyway since we went through three iterations of fixing this bug. I don't think it would be beneficial to anyone viewing the history to include those previous attempts.

lemonsaurus:
I personally don't really think micromanaging commit history in this way is something we should be spending any amount of time on. four commits for a bugfix is fine.

PR #356, opened 2019-04-21
Manually pin urllib3 version to mitigate vulnerability

sco1:
Addresses https://nvd.nist.gov/vuln/detail/CVE-2019-11324, an issue with `urllib3 1.24.1` that is fixed in `1.24.2`, released a few days ago.

`urllib3` is a `dulwich` and `safety`/`sphinx` (via `requests`) requirement. `dulwich` pins to `>=1.23` and `requests` pins to `>=1.21.1,<1.25`, so we should be able manually pin `1.24.2` without conflict.

This PR manually pins this version, resyncs dependencies, and relocks.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -18,6 +18,7 @@ aio-pika = "*"
 python-dateutil = "*"
 deepdiff = "*"
 requests = "*"
+urllib3 = ">=1.24.2,<1.25"
 
 [dev-packages]
 "flake8" = ">=3.6"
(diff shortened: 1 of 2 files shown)


GhostofGoes:
It needs to be re-locked I think

sco1:
Forgot to update after Lemon dropped the dulwich dependency. I'll change it when I get home this afternoon.

PR #360, opened 2019-05-06
Preventing the multiple reddit background tasks bug

### LINKED ISSUE #359: Bot creates multiple Reddit background tasks
Juanita noticed that the bot posted and pinned the overview of the weekly top posts of the Python subreddit multiple times:

The reason is that the `on_ready` event may fire multiple times (for instance, when the bot recovers a dropped connection), creating multiple background tasks:

The code causing the bug can be found in this on ready: https://github.com/python-discord/bot/blob/7a61e582ccb866776136d969c97d23907ca160e2/bot/cogs/reddit.py#L279

The fix should be straightforward: Keep a reference to the Background task as a class attribute and check if there's already a Task running like we do at other places as well.

Since this both hits the Discord API and the Reddit API, we should probably just fix this right away and not wait until the feature freeze is over. I'll put in a PR soon.

SebastiaanZ:
This PR closes #359 

It prevents multiple background tasks for the new post poll and weekly top posts poll by saving a reference to task as a class attribute and checking if one already exists when creating one. The reason this needs to be done is because the `on_ready` event fires every time the bot reestablishes the connection, not just on the initial start of the bot.

### DIFF
--- a/bot/cogs/reddit.py
+++ b/bot/cogs/reddit.py
@@ -31,6 +31,9 @@ class Reddit:
         self.prev_lengths = {}
         self.last_ids = {}
 
+        self.new_posts_task = None
+        self.top_weekly_posts_task = None
+
     async def fetch_posts(self, route: str, *, amount: int = 25, params=None):
         """
         A helper method to fetch a certain amount of Reddit posts at a given route.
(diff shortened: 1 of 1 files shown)


SebastiaanZ:
An exception in a task shouldn't kill the loop, although it will complain if you never retrieve the exception from the Task after a while. I don't think that kills the loop, though.

MarkKoz:
If one were to explicitly write  `raise Exception` within the while loop, it would make the task "done" and the exception could be accessed with `task.exception()`. I am just not sure how exceptions from discord.py would be handled cause they have their own error handler system to which exceptions may get redirected.

MarkKoz:
It's just adding an extra `if task.done()` check so it wouldn't hurt to be safe and add it unless we can be sure of the behaviour to expect.

SebastiaanZ:
Ah, yes. Does this have a relevance to the current bug?

The bug is not related to an exception in a Task itself (they run fine), but that multiple tasks get created when the bot reestablishes a connection after it has been dropped. This causes the `on_ready` to fire again and, in the old version of the code, this would create duplicate tasks running concurrently.

MarkKoz:
No, it isn't directly related. However, as a consequence of this patch, if the task does prematurely finish for whatever reason, the task will never get re-created.

MarkKoz:
Actually, even with what I propose it would only cover cases of when the task is prematurely ended as a result of the bot disconnecting. If, for example, there is an error with the reddit API, that would not be covered, so there's some underlying issue with the implementation of the tasks. Perhaps it's beyond the scope of this PR after all.

PR #365, opened 2019-05-20
Master - removing verification system

lemonsaurus:

PR #366, opened 2019-05-20
Django - removing verification system

lemonsaurus:
This is the same as #365 but for the django branch.


jchristgit:
Closing as discussed in the staff meeting.

PR #372, opened 2019-06-14
Replace RabbitMQ with Snekbox API

MarkKoz:
The front-end behaviour isn't exactly the same. Instead of simply stating "the job has completed", a more detailed statement will be given e.g. "the job timed out" along with the return code and a signal name if applicable. This allows for the output to remain in the code block which would have otherwise been occupied by an error message. 

Furthermore, stderr was merged into stdout which ensures all output is always displayed. The only exception would be if the code explicitly used an exit code of 255, which will have the code block output be overridden by "A fatal NsJail error occurred". A fair compromise I think since it's a strange code use explicitly.

Another notable difference is that line numbers now correctly correspond to the input code because the code template has been eliminated.

Other changes were mainly refactoring: splitting things into separate functions and using a maximum line length of 100.


bot/cogs/snekbox.py
```diff
@@ -57,39 +42,38 @@
...
+        try:
+            async with self.bot.http_session.post(url, data=output, raise_for_status=True) as resp:
+                data = await resp.json()
 
-        log.info(f"Received code from {ctx.author.name}#{ctx.author.discriminator} for evaluation:\n{code}")
-        self.jobs[ctx.author.id] = datetime.datetime.now()
+            if "key" in data:
+                return URLs.paste_service.format(key=data["key"])
+        except Exception:
+            # 400 (Bad Request) means the data is too large
```
MarkKoz: I think this actually happens due to too many characters. More specifically, I believe I pinpointed it to 100k characters. Can this be confirmed?

A check can be added for the output length. Either uploading can be skipped or the output can be truncated to 100k characters.

bot/cogs/snekbox.py
```diff
@@ -101,86 +85,140 @@ def rmq(self) -> RMQ:
...
 
-        code = textwrap.indent(code, "    ")
-        code = CODE_TEMPLATE.replace("{CODE}", code)
+        return code
+
+    @staticmethod
+    def get_results_message(results: dict) -> Tuple[str, str]:
+        """Return a user-friendly message and error corresponding to the process's return code."""
+        stdout, returncode = results["stdout"], results["returncode"]
+        msg = f"Your eval job has completed with return code {returncode}"
```
MarkKoz: Would it be clearer if it was called an "exit code" instead?

lemonsaurus: No, return code is fine.

Issue #317, opened 2019-02-23
Proposal to restructure our watch channel cogs (big-brother, talent-pool)


SebastiaanZ:
With the introduction of the helper nomination feature for the bot, our big-brother cog became much more complex and quite difficult to maintain. It implements similar, but slightly different features, which results in a lot of slightly different methods and/or conditional code within methods. That's why I like the idea of the initial implementation of @jchristgit in the Django branch: Have a separate cog for *Nomination* that inherits from *BigBrother*. In my opinion, this makes a lot of sense now the API endpoints have been separated anyway. However, I would like to propose a slightly different tactic for the inheritance, namely a *WatchChannel* base class both inherit from.

**WatchChannel base class**
In the current structure, with Nomination inheriting from BigBrother, changes to the BigBrother cog may be reflected in Nomination cog if it concerns a method that’s not overwritten in the child class. Since I don't agree with this hierarchy, I would like to propose a *WatchChannel* base class in bot/utils with the following inheritance diagram:

![Inheritance schematic](https://user-images.githubusercontent.com/33516116/53285591-905f5400-3762-11e9-84ff-d60eaf2b7797.png)

The base class should define the basic, but complete mechanisms for a watch channel, including the consumption, cache update, watch and unwatch features, in a **granulated** manner. The level of the granulation should depend on the needs to have different implementations for features in the different child classes. 

As an example, the header embeds for BigBrother and Nomination are different, but the rest of the header functionality is the same. By creating a separate method for header creation, we could easily just overwrite that part in our child classes, while keeping the rest of the code (that sends the header) consistent and DRY. Obviously, the granulation level should make sense, but this is an example to show the basic idea.

This is the gist of it and below is a list of the pros and cons I can think of.

**Pros**:

- Independence between BigBrother and Nomination: A change in one cog does not force a change in the inherited cog;
- DRY: we only have to overwrite that granules of code that are actually different;
- Separation of base functionality and implementation: We no longer have a massive, hybrid cog that implements two related, but ultimately different features (although the Django branch already has some separation with Nomination inheriting from BigBrother)

**Cons**:

- The abstraction separates the code over two files. However, this is already kinda the case with Nomination inheriting from BigBrother in the Django branch;
- Slightly more difficult code for beginning contributors to grasp.

Feel free to add your own thoughts to this thread or shoot my idea down completely.

scragly:
As the two are intrinsically related through the base class, these should likely belong in the one extension:
```
watch_tools/
[...]

Issue #315, opened 2019-02-22
Some users seem to be ununwatchable by bigbrother.


tagptroll1:
I attempted to unwatch someone I put up ages ago and got an error by the convert itself saying it couldn't find the User.  This is going to fill up our list of watched users with ununwatchable Users eventually.  

I propose a force unwatch command that just erases it as best as possible from storage.

SebastiaanZ:
So, to me it looks like the UserConverter throws this error when the user cannot be found. Since I think we should keep the converter (it's very handy when users can be converted), I think we should add a `force_unwatch` subcommand that unwatches by ID. For the API, only an ID is needed, not a full user object.

sco1:
I don't see why we need an entirely new command when we could just use a `Union` on the existing command.

SebastiaanZ:
Oh, didn't know about that, sounds like the better option.

sco1:
Docs: https://discordpy.readthedocs.io/en/rewrite/ext/commands/commands.html#typing-union

It goes left to right, so if the `Member` conversion fails it would convert to `str`/`int`

SebastiaanZ:
Thanks for the link, I'll get a fix in.

SebastiaanZ:
I have a fix waiting; I'll put a PR in once #307 is merged, since the code conflicts and that PR seem close to merging anyway (two positive reviews; but @heavysaturn needs to accept the changes they've requested).

MarkKoz:
Fixed by #376

PR #376, opened 2019-06-28
BigBrother/TalentPool/WatchChannel submodule

### LINKED ISSUE #324: Nominate and BB watch allows the bot to be watched, causing loop.
Nominate and BB watch allows Python, the bot, to have it's messages sent to the respective watch channels, causing a loop of all it's messages to be sent continuously to the watch channel over and over again.

Just needs to check it's not self (or maybe any bot) and deny.

SebastiaanZ:
This PR restructures the watch channel moderation tool we use to monitor potentially problematic users and promising candidates for the Helpers role. For more information about the structure, see the issue this should close: #317. This PR also fixes some bugs and closes #324 and #315. 

**This PR depends on the PR that makes changes to the Nomination endpoint in site: https://github.com/python-discord/site/pull/220**

## Main features

This is an overview of the main features, focusing on the differences between the old and the new implementation. 

- A `WatchChannel` baseclass that implements most of the common functionality

- Two watch channel subclasses, `TalentPool` and `BigBrother`, that inherit from `WatchChannel` and extend it with the watch channel specific commands and functionality.

- To organize the files, the watch channels have now been added to a submodule folder with an `__init__.py` file that loads both `Cogs` when the `bot.cogs.watchchannels` extension is loaded.

- Messages are now relayed using webhooks instead of the regular API. This has two main advantages: Messages are now posted with the display name and avatar of the watched member (instead of the bot; see below) and webhooks have separate rate limit buckets. (A side effect of this is that the header embeds are now also smaller and less disruptive.)

Message with avatar/display name of watched user

- Since the user caches are now separated, the embed that lists all the watched users is now watch channel specific as well. In addition, they are sorted based on the watch date (most recently added first) and the duration of the watch is displayed in the embed:

watch channel list embed

- Since nominations are no longer part of the infraction tables or API in the Django back-end, they won't show up in an `!infraction search` result embed. I've therefore added a new `search` subcommand for the `TalentPool` cog that outputs the user's nomination history. 

Nomination History

- The current implementation in `master` suffers from the bug: the destination channel sometimes fails to be loaded from the channel cache during start-up, preventing the watch channels from functioning. This has been fixed by adding some retry logic to the `Guild.get_channel`. After we update to the latest version of d.py, this should be replaced by `fetch_channel`, which has retry logic built-in and doesn't rely on the bot cache, but fetches the channel from the API. Regardless, if initialization of a watch channel fails, the bot will now send an alert to the mod-log channel. 

mod_alert

- Bot users can no longer be watched, closing #324 

## Commands

The commands for `BigBrother` are the same as they are in the current implementation in `master`. The `TalentPool` cog has one additional subcommand: `history` (alias `search` to be consistent with `!inf search <user>`) that lists the user's nomination history, since that's no longer part of the infraction history of a user.

## Not implemented

The current `on_member_ban` implementation in the `master` branch often fails in unwatching a user when the user is banned from our server. It also doesn't differentiate between temporary banishments and those of a more permanent nature, while it would obviously be a good idea to keep someone who we've just banned for a day in big-brother.

I've not yet implemented an alternative, so I've left this out for now and make an issue for it. It's not absolutely critical and one of the potential solutions to the temp/perm ban problem is by hooking the unwatch into the `!tempban`-command instead of trying to figure out what kind of ban it is in the `on_member_ban` Event listener. However, I don't want to touch those files in the middle of Mark's master-django merge project.


bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+        time_delta = self._get_time_delta(inserted_at)
+
+        reason = self.watched_users[user_id]['reason']
+
+        embed = Embed(description=(
+            f"{msg.author.mention} in [#{msg.channel.name}]({msg.jump_url})\n"
+        ))
+        embed.set_footer(text=(
+            f"Added {time_delta} by {actor} | "
+            f"Reason: {reason}"
```
MarkKoz: This could fit on one line and it's a bit misleading seeing it broken up as I thought there was a newline in there at first.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+            lines.append(f"• <@{user_id}> (added {time_delta})")
+
+        await LinePaginator.paginate(
+            lines or ("There's nothing here yet.",),
+            ctx,
+            Embed(
+                title=f"{self.__class__.__name__} watched users ({'updated' if update_cache else 'cached'})",
+                color=Color.blue()
+            ),
+            empty=False
```
MarkKoz: OK, so this follows our style but it's kinda ugly since it's so disproportionate. Do you think this looks better?

```py
        lines = lines or ("There's nothing here yet.",)
        embed = Embed(
            title=f"{self.__class__.__name__} watched users ({'updated' if update_cache else 'cached'})",
            color=Color.blue()
        )
        await LinePaginator.paginate(lines, ctx, embed empty=False)
```

Preferably the embed's title would also be under 100 columns.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+            self.log.trace("Channel queue not empty: Continuing consuming queues")
+            self._consume_task = self.bot.loop.create_task(
+                self.consume_messages(delay_consumption=False)
+            )
+        else:
+            self.log.trace("Done consuming messages.")
+
+    async def webhook_send(
+        self, content: Optional[str] = None, username: Optional[str] = None,
+        avatar_url: Optional[str] = None, embed: Optional[Embed] = None,
```
MarkKoz: I really don't want to hold up the PR with style shit but I think our preferred style is to put each parameter on a separate line when the signature is too long. If no consensus we'll just take care of this later.

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,276 @@
...
+from discord import Color, Embed, Member, User
+from discord.ext.commands import Context, group
+
+from bot.constants import Channels, Guild, Roles, Webhooks
+from bot.decorators import with_role
+from bot.pagination import LinePaginator
+from .watchchannel import WatchChannel, proxy_user
+
+log = logging.getLogger(__name__)
+STAFF_ROLES = Roles.owner, Roles.admin, Roles.moderator, Roles.helpers    # <- In constants after the merge?
```
MarkKoz: Yes, I'll take care of this when merging. It was probably the most frequent conflict I dealt with in the merge.

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,118 @@
...
+            )
+        )
+        if active_watches:
+            [infraction] = active_watches
+            await self.bot.api_client.patch(
+                f"{self.api_endpoint}/{infraction['id']}",
+                json={'active': False}
+            )
+            await post_infraction(
+                ctx, user, type='watch', reason=f"Unwatched: {reason}", hidden=True, active=False
```
MarkKoz: Another style thing where I think the arguments should be on separate lines. There's more instances of this but I'm not going to leave a comment on each one.

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,118 @@
...
+from bot.constants import Channels, Roles, Webhooks
+from bot.decorators import with_role
+from bot.utils.moderation import post_infraction
+from .watchchannel import WatchChannel, proxy_user
+
+log = logging.getLogger(__name__)
+
+
+class BigBrother(WatchChannel):
+    """User monitoring to assist with moderation"""
```
MarkKoz: Could be more descriptive by specifying their messages are what's monitored.

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,276 @@
...
+            lines = textwrap.dedent(
+                f"""
+                ===============
+                Status: **Active**
+                Date: {start_date}
+                Actor: {actor.mention if actor else actor_id}
+                Reason: {nomination_object["reason"]}
+                Nomination ID: `{nomination_object["id"]}`
+                ===============
+                """
```
MarkKoz: Does this cause issues with leading and trailing newlines? Other instances that use triple quotes as parameters have done it a bit differently:

https://github.com/python-discord/bot/blob/django/bot/cogs/moderation.py#L157

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,276 @@
...
+        try:
+            nomination = await self.bot.api_client.get(f"{self.api_endpoint}/{nomination_id}")
+        except ClientResponseError as e:
+            if e.status == 404:
+                self.log.trace(f"Nomination API 404: Can't nomination with id {nomination_id}")
+                e = Embed(
+                    description=f":x: **Can't find a nomination with id `{nomination_id}`**",
+                    color=Color.red()
+                )
+                await ctx.send(embed=e)
```
MarkKoz: At this point I feel like we could benefit from an error embed utility function... Making a consistent error message has been discussed before. I don't think now is a bad time to start. You don't yet have to migrate the entire codebase over to the new function if you choose to add it now.

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,276 @@
...
+        e = Embed(
+            description=f":white_check_mark: **Messages sent by {user} will now be relayed to TalentPool**",
+            color=Color.green()
+        )
+        await ctx.send(embed=e)
+
+    @nomination_group.command(name='history', aliases=('info', 'search'))
+    @with_role(Roles.owner, Roles.admin, Roles.moderator)
+    async def history_command(self, ctx: Context, user: Union[User, proxy_user]) -> None:
+        """Shows the specified user's nomination history"""
```
MarkKoz: I've noticed some docstrings have full stops and some don't. What's it gonna be?

config-default.yml
```diff
@@ -128,6 +129,10 @@ guild:
         helpers:                            267630620367257601
         rockstars:         &ROCKSTARS_ROLE  458226413825294336
 
+    webhooks:
+        talent_pool:                        569145364800602132
+        big_brother:                        569133704568373283
```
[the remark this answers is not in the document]
SebastiaanZ: Yes, these are created on our server. Test server ones are defined in my config.yml

bot/utils/messages.py
```diff
@@ -78,7 +78,7 @@ def check(reaction, user):
         await message.delete()
 
 
-async def send_attachments(message: Message, destination: TextChannel):
+async def send_attachments(message: Message, destination: Union[TextChannel, Webhook]):
     """
     Re-uploads each attachment in a message to the given channel.
```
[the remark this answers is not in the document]
SebastiaanZ: That's true. I'll change that.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+    try:
+        user_id = int(user_id)
+    except ValueError:
+        raise BadArgument
+    user = Object(user_id)
+    user.mention = user.id
+    user.display_name = f"<@{user.id}>"
+    user.avatar_url_as = lambda static_format: None
+    user.bot = False
+    return user
```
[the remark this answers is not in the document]
SebastiaanZ: I'll change this here, but this should indeed be made into a utility function. This was copied from the moderation cog in the `master` branch, but with two attributes added. We could probably come up with a "mock" user that suits all our needs, as I don't think any of the known use cases use incompatible attributes.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+                    break
+            else:
+                self.log.error(f"Cannot get channel with id `{self.destination}`; cannot watch users")
+                return False
+
+        if self.webhook is None:
+            self.webhook = await self.bot.get_webhook_info(self.webhook_id)  # This is `fetch_webhook` in current
+            if self.webhook is None:
+                self.log.error(f"Cannot get webhook with id `{self.webhook_id}`; cannot watch users")
+                return False
```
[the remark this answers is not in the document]
SebastiaanZ: Unlike the `get_channel`, the `get_webhook_info` is already a coroutine that fetches the Webhook from the API; it's now been renamed `fetch_webhook`, but it should not suffer from the same problem as `get_channel` that gets the channel from a cached list of channels in the Bot instance.

bot/cogs/watchchannels/watchchannel.py
SebastiaanZ: That's an interesting point; I'm not sure what the current consensus is. Both are styles, let's call them parameter Tetris and one-per-line style, are prevalent in our bot code base. 

Here are some examples of the Tetris style:

https://github.com/python-discord/bot/blob/django/bot/pagination.py#L96

https://github.com/python-discord/bot/blob/django/bot/pagination.py#L317

Here, the one-per-line style (more prevalent in django-branch):

https://github.com/python-discord/bot/blob/5e16f4a52d59c73a04323e070e7b4a320e8c1e49/bot/utils/messages.py#L15

https://github.com/python-discord/bot/blob/5e16f4a52d59c73a04323e070e7b4a320e8c1e49/bot/rules/attachments.py#L8

I'm leaning towards the one-per-line style, so I'm going to change it here (since both seem acceptable). There was debate surrounding black recently in core-dev, though, that touched on similar issues (tetris vs one-per-line) that had people arguing for both sides.

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: I don't agree with this one, since this has all the arguments on a single line and this is actually the style that `black` would also produce.

bot/cogs/watchchannels/talentpool.py
SebastiaanZ: The difference between the two is not the newlines (there's still a newline after the `"""` in both cases, but the `strip` takes care of it later), but rather that in this case, the string starts the line after the opening `(`, indented like a list of arguments would if the `()` split over multiple lines.

In my eyes, doing this:
```py
textwrap.dedent("""
    hello
    this is my multiline comment
""")
```
looks like:
```py
some_function(arg1,
    arg2,
    arg3,
arg4)
```
Since part of the string definition is indented, part of it follows the opening `(` and part of it is squeezed before the closing `)` unindented (the closing quotes).

bot/cogs/watchchannels/bigbrother.py
jchristgit: why did we change this in the first place? what's wrong with the old one?

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,118 @@
...
+
+        A `reason` for adding the user to BigBrother is required and will displayed
+        in the header when relaying messages of this user to the watchchannel.
+        """
+        if user.bot:
+            e = Embed(
+                description=f":x: **I'm sorry {ctx.author}, I'm afraid I can't do that. I only watch humans.**",
+                color=Color.red()
+            )
+            await ctx.send(embed=e)
```
jchristgit: why is this an embed?

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,118 @@
...
+            return
+
+        if not await self.fetch_user_cache():
+            log.error("Failed to update user cache; can't watch user {user}")
```
jchristgit: this needs to be sent to the author

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,118 @@
...
+            return
+
+        if user.id in self.watched_users:
+            e = Embed(
+                description=":x: **The specified user is already being watched**",
+                color=Color.red()
+            )
+            await ctx.send(embed=e)
+            return
```
jchristgit: why is this an embed and bold

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,118 @@
...
+
+        response = await post_infraction(
+            ctx, user, type='watch', reason=reason, hidden=True
+        )
+        if response is not None:
+            self.watched_users[user.id] = response
+            e = Embed(
+                description=f":white_check_mark: **Messages sent by {user} will now be relayed to BigBrother**",
+                color=Color.green()
+            )
```
jchristgit: why is this an embed and bold

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,118 @@
...
+            await ctx.send(embed=e)
+            return
+
```
jchristgit: do we not need an else case?

bot/cogs/watchchannels/bigbrother.py
jchristgit: this will override the reason of the existing watch, I think we should append to it instead. i also agree on multiple lines, just because black would produce this style doesn't mean its good

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,118 @@
...
+            )
+            await post_infraction(
+                ctx, user, type='watch', reason=f"Unwatched: {reason}", hidden=True, active=False
+            )
+            e = Embed(
+                description=f":white_check_mark: **Messages sent by {user} will no longer be relayed**",
+                color=Color.green()
+            )
+            await ctx.send(embed=e)
+            self._remove_user(user.id)
```
jchristgit: why is this an embed and bold

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: Just wrote a new one while I was writing the new class. I can copy the old one, I think it sounds a bit aggressive, though, so that's probably why I didn't copy it.

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,118 @@
...
+        else:
+            e = Embed(
+                description=":x: **The specified user is currently not being watched**",
+                color=Color.red()
+            )
+            await ctx.send(embed=e)
```
jchristgit: embed and bold question #\200, ditto for the other cog

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: The bold is a bit overkill, but I personally like the embed-style bot responses instead of the plain-text message ones. Makes it stand out more as a bot response.

bot/cogs/watchchannels/bigbrother.py
jchristgit: if "NSA-style" is aggressive, I guess I'm done.

SebastiaanZ: Hmm, maybe aggressive is not the right word, but it does feel a bit over the top to me. Probably because I'm not really into memeing.

bot/cogs/watchchannels/talentpool.py
```diff
@@ -103,12 +104,12 @@ def __init__(self, bot):
                     color=Color.red()
                 )
                 return await ctx.send(embed=e)
-            elif resp.status >= 400:
-                resp.raise_for_status()
+
+            resp.raise_for_status()
```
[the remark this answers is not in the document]
SebastiaanZ: Done.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+    user.mention = user.id
+    user.display_name = f"<@{user.id}>"
+    user.avatar_url_as = lambda static_format: None
+    user.bot = False
+    return user
+
+
+class WatchChannel(ABC):
+    """
+    Base class for WatchChannels
```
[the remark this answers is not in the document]
SebastiaanZ: I've changed it to the text below, but if anyone has a suggestion for a better alternative, it's very welcome:

```py
class WatchChannel(ABC):
    """ABC that implements watch channel functionality to relay all messages of a user to a watch channel."""
```

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+
+    Abstracts the basic functionality for watchchannels in
+    a granular manner to allow for easy overwritting of
+    methods in the child class.
+    """
+
+    @abstractmethod
+    def __init__(self, bot: Bot, destination, webhook_id, api_endpoint, api_default_params, logger) -> None:
+        """
+        abstractmethod for __init__ which should still be called with super().
```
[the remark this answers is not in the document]
SebastiaanZ: I've removed the docstring for now. It add nothing useful now that the parameters need to be provided as arguments.

Going by the seasonalbot code base, which I think does implement the new docstring linting, `__init__` docstrings are not required.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+        # These attributes can be left as they are in the child class
+        self._consume_task = None
+        self.watched_users = defaultdict(dict)
+        self.message_queue = defaultdict(lambda: defaultdict(deque))
+        self.consumption_queue = {}
+        self.retries = 5
+        self.retry_delay = 10
+        self.channel = None
+        self.webhook = None
+        self.message_history = [None, None, 0]
```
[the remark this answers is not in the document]
SebastiaanZ: Yeah, I had this on my "I might change this later" list, but I didn't get around to it. I'll leave this conversation open for now and implement it later. Maybe we can discuss a good approach in the core-dev channel first. A dataclass could work and has nice attribute access, like named tuples. Nothing wrong with adding a dataclass to a module that implements an ABC either, I think.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+    def consuming_messages(self) -> bool:
+        """Checks if a consumption task is currently running."""
+        if self._consume_task is None:
+            return False
+
+        if self._consume_task.done():
+            exc = self._consume_task.exception()
+            if exc:
+                self.log.exception(
+                    f"{self.__class__.__name__} consume task has failed with:",
```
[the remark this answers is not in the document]
SebastiaanZ: That's very true; it's totally redundant to do that there. This is a leftover from before I passed the logger of the concrete implementation.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+        await self.bot.wait_until_ready()
+
+        if await self.initialize_channel() and await self.fetch_user_cache():
+            self.log.trace(f"Started the {self.__class__.__name__} WatchChannel")
+        else:
+            self.log.error(f"Failed to start the {self.__class__.__name__} WatchChannel")
+
+            # Let's try again in a minute.
+            await asyncio.sleep(60)
+            self._start = self.bot.loop.create_task(self.start_watchchannel())
```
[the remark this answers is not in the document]
SebastiaanZ: I've restructured this part completely.

The TextChannel is the only part that actually needs retries, since the `fetch_*` methods of discord.py already have internal retry logic that takes the type of failure into account (retrying if the API is unreachable, not bothering again if there's a permissions problem or similar). (Note: the `get_webhook_info` is such a `fetch_*` method and has been renamed in newer versions to reflect that.)

I've removed the separate `initialize_channel` function, since it's only called when we start the watch channel anyway and moved the code over to `start_watchchannel`. I've only wrapped the `get_channel` in the retry logic, so we only have to replace those lines of code after we update to a newer version d.py.

If either the TextChannel or Webhook instances fail to be fetched/retrieved, the entire cog becomes non-functional, so I've followed your suggestion and included logic to unload the cog. Basically, it will log it, then tries to use the `ModLog` to send an alert to the mod-log channel, and finally unloads itself. (If the `ModLog` alert fails, then something is entirely wrong with the bot and we probably have no way of sending an alert by other means.)

The final step in the start-up is retrieving the watched user list from our API. If that fails, the cogs stays "functional" (without watched users), so it won't be unloaded, but it will send an alert notifying that we failed to retrieve the watched users. If it's just a temporary hiccup in the API, a simple cache refresh command could then fix the issue. (This occasionally happens now in the old implementation.)

bot/cogs/watchchannels/watchchannel.py
SebastiaanZ: Changed and I've changed a couple of similar lines as well.

bot/cogs/watchchannels/watchchannel.py
SebastiaanZ: I've changed to formatting but not the line length. As it stands now, there are a lot of other lines that go over a `100` as well. Since there's been some discussion of using `black` with a line limit of `120`, I've left this as it is for now.

If we are going to use `black`, it's going to produce a lot of lines with near 120 columns.

bot/cogs/watchchannels/watchchannel.py
SebastiaanZ: I've changed this to the a `one-per-line`-style.

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: I'll change the style, although the argument wasn't "it's good because `black` does it". My personal preference is the way it was, that's why I coded it that way, and not the version that smears it out over multiple lines unnecessarily. The `black` part came from the idea of using `black` on the bot code base, which will change this back to the single line version anyway.

(I know the black idea is controversial, so it may never happen.)

SebastiaanZ: > this will override the reason of the existing watch, I think we should append to it instead.

No, it will add a new `Watch` infraction with the unwatch reason. That way, not only the unwatch reason is stored, but also the date and actor.

bot/cogs/watchchannels/talentpool.py
SebastiaanZ: I think Volcyyy prefers to use plain messages instead of embeds for these type of things. I will probably change the code back to that. I like the embeds because it stands out, but it's not necessary to use embeds for bot responses.

bot/cogs/watchchannels/talentpool.py
SebastiaanZ: Full stops. I don't think we enforce the rule under the new linting rules, because we want to allow other punctuation marks as well, but we still concluded that we prefer to have some kind of punctuation mark at the end of the summary line, since it will also be used in the help embeds.

bot/cogs/watchchannels/talentpool.py
gdude2002: I remember that it was either embeds previously or that I brought up the idea of using embeds - I agree that they look nicer, but from what I remember, it was preferred to use simple messages because embeds take up a lot of room

SebastiaanZ: I guess you're right, it does save some space.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,359 @@
...
+    """
+
+    @abstractmethod
+    def __init__(self, bot: Bot, destination, webhook_id, api_endpoint, api_default_params, logger) -> None:
+        """
+        abstractmethod for __init__ which should still be called with super().
+
+        Note: Some of the attributes below need to be overwritten in the
+        __init__ of the child after the super().__init__(*args, **kwargs)
+        call.
```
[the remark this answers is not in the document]
SebastiaanZ: Removed.

bot/cogs/watchchannels/watchchannel.py
MarkKoz: A while ago it was confirmed we wanted to target 100 columns but just haven't done it yet for certain repos. But that was like a month ago and I guess we change our minds. I don't know what's going on with black really. Let's just worry about that later.

sco1: All new code should adhere to a line length of 100.

`black` was a thought experiment

sco1: Disregard, I can't read properly. The current stance is a line length of 100 for new projects.

bot/cogs/watchchannels/talentpool.py
MarkKoz: Sounds good. Not sure if you've made the changes yet so I'll let you resolve this when it's ready.

bot/cogs/watchchannels/talentpool.py
MarkKoz: Missed the strip afterwards. This is fine as-is.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,346 @@
...
+        # After updating d.py, this block can be replaced by `fetch_channel` with a try-except
+        for attempt in range(1, self.retries+1):
+            self.channel = self.bot.get_channel(self.destination)
+            if self.channel is None:
+                if attempt < self.retries:
+                    await asyncio.sleep(self.retry_delay)
+            else:
+                break
+        else:
+            self.log.error(f"Failed to retrieve the text channel with id `{self.destination}")
```
MarkKoz:
```suggestion
            self.log.error(f"Failed to retrieve the text channel with id {self.destination}")
```

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,346 @@
...
+    user.mention = user.id
+    user.display_name = f"<@{user.id}>"
+    user.avatar_url_as = lambda static_format: None
+    user.bot = False
+
+    return user
+
+
+class WatchChannel(ABC):
+    """ABC that implements watch channel functionality to relay all messages of a user to a watch channel."""
```
MarkKoz:
```suggestion
    """"ABC with functionality for relaying users' messages to a certain channel.""""
```

bot/cogs/watchchannels/watchchannel.py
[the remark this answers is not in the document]
MarkKoz: I made a suggestion for the docstring further down. The naming was addressed so I'll resolve this review.

bot/__init__.py
```diff
@@ -55,7 +55,7 @@ def monkeypatch_trace(self, msg, *args, **kwargs):
 
 
 logging.basicConfig(
-    format="%(asctime)s pd.beardfist.com Bot: | %(name)30s | %(levelname)8s | %(message)s",
+    format="%(asctime)s pd.beardfist.com Bot: | %(name)33.33s | %(levelname)8s | %(message)s",
```
MarkKoz: I've not seen this kind of formatting before. What does `33.33` do?

bot/utils/moderation.py
```diff
@@ -14,16 +14,17 @@
 
 
 async def post_infraction(
-    ctx: Context, user: Union[Member, Object, User],
-    type: str, reason: str, expires_at: datetime = None, hidden: bool = False
+    ctx: Context, user: Union[Member, Object, User], type: str, reason: str,
+    expires_at: datetime = None, hidden: bool = False, active: bool = True
```
MarkKoz: I think you agreed to do one-per-line for some other function so this is a candidate for it too.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,346 @@
...
+            self.log.error("Failed to start the watch channel; unloading the cog.")
+
+            message = textwrap.dedent(
+                f"""
+                An error occurred while loading the text channel or webhook.
+
+                TextChannel: {"**Failed to load**" if self.channel is None else "Loaded successfully"}
+                Webhook: {"**Failed to load**" if self.webhook is None else "Loaded successfully"}
+
+                The Cog has been unloaded."""
```
MarkKoz: Should the closing quotes be on a new line? Is this done to prevent having to strip afterwards?

bot/cogs/watchchannels/watchchannel.py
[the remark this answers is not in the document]
MarkKoz: Looks great now.

bot/cogs/watchchannels/talentpool.py
MarkKoz: Looks like you did it actually so I'll resolve.

bot/cogs/watchchannels/watchchannel.py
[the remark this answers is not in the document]
MarkKoz: In that case I can just figure it out when merging with master. You don't have to make it a util in this PR.

bot/__init__.py
SebastiaanZ: It's probably something that I should reverse, now that I think of it. It's width + truncation. However, full information is probably more useful than alignment.

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: I'm turning those bot responses into plain messages.

bot/cogs/watchchannels/watchchannel.py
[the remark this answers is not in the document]
SebastiaanZ: I've created a `dataclass` for this for now. I think it enhances readability, but it may be a bit overkill. I'll leave this convo open as a reminder for now.

bot/cogs/watchchannels/watchchannel.py
SebastiaanZ: They should indeed both be on a separate line or the same line, for consistency. It doesn't matter, as far as the embed is concerned, since the additional newlines seem to be stripped regardless. I think the version with the quotes on separate lines is more readable, as it will show you the text as it will be displayed in the embed as well.

bot/utils/messages.py
```diff
@@ -97,7 +97,14 @@ def check(reaction, user):
             if attachment.size <= MAX_SIZE - 512:
                 with BytesIO() as file:
                     await attachment.save(file)
-                    await destination.send(file=File(file, filename=attachment.filename))
+                    if isinstance(destination, TextChannel):
+                        await destination.send(file=File(file, filename=attachment.filename))
+                    else:
+                        await destination.send(
+                            file=File(file, filename=attachment.filename),
```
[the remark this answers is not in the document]
SebastiaanZ: Changed.

bot/utils/moderation.py
SebastiaanZ: Done.

bot/__init__.py
SebastiaanZ: I've changed it to 33s. That accommodates for the watch channel log names (`bot.cogs.watchchannels.talentpool`), but won't truncate if something somehow generates an even longer logger name.

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: No, the `post_infraction` utility function already catches the Exception, sends an error message to the `Context` and then returns `None`. So, both notifying the user and logging the issue are taken care of there.

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: Changed.

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: Changed.

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: Changed to a plain message, no bold

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: Changed to a plain message, no bold.

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: Changed to a plain message, removed the bold

bot/cogs/watchchannels/bigbrother.py
SebastiaanZ: Changed to a plain message, removed the bold

bot/cogs/watchchannels/talentpool.py
SebastiaanZ: Okay, I'll leave it as it is for now, then.

bot/cogs/watchchannels/talentpool.py
SebastiaanZ: I've changed all the bot responses to plain messages, as @jchristgit requested.

bot/cogs/watchchannels/watchchannel.py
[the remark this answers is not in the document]
SebastiaanZ: Then I'll just leave this for now.

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,237 @@
...
+            },
+            'raise_for_status': False,
+        }
+        async with session.post(url, **kwargs) as resp:
+            response_data = await resp.json()
+
+            if resp.status == 400 and response_data.get('user', False):
+                e = Embed(
+                    description=":x: The specified user can't be found in the database tables",
+                    color=Color.red()
```
MarkKoz: You missed changing this one to a plain message.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,357 @@
...
+        Gives an overview of the watched user list for this channel.
+
+        The optional kwarg `update_cache` specifies whether the cache should
+        be refreshed by polling the API.
+        """
+        if update_cache:
+            if not await self.fetch_user_cache():
+                e = Embed(
+                    description=f":x: **Failed to update {self.__class__.__name__} user cache, serving from cache**",
+                    color=Color.red()
```
MarkKoz: Another missed embed.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,357 @@
...
+                avatar_url=msg.author.avatar_url
+            )
+
+        if msg.attachments:
+            try:
+                await messages.send_attachments(msg, self.webhook)
+            except (errors.Forbidden, errors.NotFound):
+                e = Embed(
+                    description=":x: **This message contained an attachment, but it could not be retrieved**",
+                    color=Color.red()
```
MarkKoz: Missed embed but maybe this one should stay?

SebastiaanZ: I think this one should stay to make it stand out from the regular messages posted to the watch channels themselves.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def _remove_user(self, user_id: int) -> None:
+        """Removes user from the WatchChannel"""
+        self.watched_users.pop(user_id, None)
+        self.message_queue.pop(user_id, None)
+        self.consumption_queue.pop(user_id, None)
+
+    def cog_unload(self) -> None:
+        """Takes care of unloading the cog and canceling the consumption task."""
+        self.log.trace(f"Unloading {self.__class__._name__} cog")
```
MarkKoz: Remove the name from the message. It was missing an underscore by the way.

```suggestion
        self.log.trace(f"Unloading the cog")
```

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,353 @@
...
+        if not self._consume_task.done():
+            self._consume_task.cancel()
+            try:
+                self._consume_task.result()
+            except asyncio.CancelledError as e:
+                self.log.exception(
+                    f"The {self.__class__._name__} consume task was canceled. Messages may be lost.",
```
MarkKoz: Remove the name from the message. It was missing an underscore by the way.

```suggestion
                    f"The consume task was canceled. Messages may be lost.",
```

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,353 @@
...
+            self.log.trace(f"Received message: {msg.content} ({len(msg.attachments)} attachments)")
+            self.message_queue[msg.author.id][msg.channel.id].append(msg)
+
+    async def consume_messages(self, delay_consumption: bool = True) -> None:
+        """Consumes the message queues to log watched users' messages."""
+        if delay_consumption:
+            self.log.trace(f"Sleeping {BigBrotherConfig.log_delay} seconds before consuming message queue")
+            await asyncio.sleep(BigBrotherConfig.log_delay)
+
+        self.log.trace(f"{self.__class__.__name__} started consuming the message queue")
```
MarkKoz: Remove the name from the message.

```suggestion
        self.log.trace(f"Started consuming the message queue")
```

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,353 @@
...
+        username: Optional[str] = None,
+        avatar_url: Optional[str] = None,
+        embed: Optional[Embed] = None,
+    ) -> None:
+        """Sends a message to the webhook with the specified kwargs."""
+        try:
+            await self.webhook.send(content=content, username=username, avatar_url=avatar_url, embed=embed)
+        except discord.HTTPException as exc:
+            self.log.exception(
+                f"Failed to send message to {self.__class__.__name__} webhook",
```
MarkKoz: Remove the name from the message.

```suggestion
                f"Failed to send a message to the webhook",
```

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -0,0 +1,353 @@
...
+                    color=Color.red()
+                )
+                await self.webhook_send(
+                    embed=e,
+                    username=msg.author.display_name,
+                    avatar_url=msg.author.avatar_url
+                )
+            except discord.HTTPException as exc:
+                self.log.exception(
+                    f"Failed to send an attachment to {self.__class__.__name__} webhook",
```
MarkKoz: Remove the name from the message.

```suggestion
                    f"Failed to send an attachment to the webhook",
```

bot/cogs/watchchannels/watchchannel.py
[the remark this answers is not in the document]
MarkKoz: The dataclass could be simplified by removing unpacking support. While mildly convenient, it's really only used in one place. I think it would make the if statement that uses the message history too long, but I wonder what the net line count will be considering the `__iter__` uses up 3 lines. 

That being said I am not opposed to unpacking support remaining.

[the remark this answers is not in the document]
SebastiaanZ: I've removed the `__iter__` and turned this into a multiline `if`-statement of the sort @lemonsaurus hates. If he wants to murder me to death over it, I'll change it to the assigned condition version he prefers.

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,100 @@
...
+            webhook_id=Webhooks.big_brother,
+            api_endpoint='bot/infractions',
+            api_default_params={'active': 'true', 'type': 'watch', 'ordering': '-inserted_at'},
+            logger=log
+        )
+
+    @group(name='bigbrother', aliases=('bb',), invoke_without_command=True)
+    @with_role(Roles.owner, Roles.admin, Roles.moderator)
+    async def bigbrother_group(self, ctx: Context) -> None:
+        """Monitors users by relaying their messages to the BigBrother watch channel."""
```
MarkKoz:
```suggestion
        """Monitors users by relaying their messages to the Big Brother watch channel."""
```

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,100 @@
...
+        await ctx.invoke(self.bot.get_command("help"), "bigbrother")
+
+    @bigbrother_group.command(name='watched', aliases=('all', 'list'))
+    @with_role(Roles.owner, Roles.admin, Roles.moderator)
+    async def watched_command(self, ctx: Context, update_cache: bool = True) -> None:
+        """
+        Shows the users that are currently being monitored in BigBrother.
```
MarkKoz:
```suggestion
        Shows the users that are currently being monitored by Big Brother.
```

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        await self.list_watched_users(ctx, update_cache)
+
+    @bigbrother_group.command(name='watch', aliases=('w',))
+    @with_role(Roles.owner, Roles.admin, Roles.moderator)
+    async def watch_command(self, ctx: Context, user: Union[User, proxy_user], *, reason: str) -> None:
+        """
+        Relay messages sent by the given `user` to the `#big-brother` channel.
+
+        A `reason` for adding the user to BigBrother is required and will displayed
```
MarkKoz:
```suggestion
        A `reason` for adding the user to Big Brother is required and will displayed
```

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -0,0 +1,100 @@
...
+            await ctx.send(":x: The specified user is already being watched.")
+            return
+
+        response = await post_infraction(
+            ctx, user, type='watch', reason=reason, hidden=True
+        )
+
+        if response is not None:
+            self.watched_users[user.id] = response
+            await ctx.send(f":white_check_mark: Messages sent by {user} will now be relayed to BigBrother.")
```
MarkKoz:
```suggestion
            await ctx.send(f":white_check_mark: Messages sent by {user} will now be relayed to Big Brother.")
```

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,233 @@
...
+            webhook_id=Webhooks.talent_pool,
+            api_endpoint='bot/nominations',
+            api_default_params={'active': 'true', 'ordering': '-inserted_at'},
+            logger=log,
+        )
+
+    @group(name='talentpool', aliases=('tp', 'talent', 'nomination', 'n'), invoke_without_command=True)
+    @with_role(Roles.owner, Roles.admin, Roles.moderator)
+    async def nomination_group(self, ctx: Context) -> None:
+        """Highlights the activity of helper nominees by relaying their messages to TalentPool."""
```
MarkKoz:
```suggestion
        """Highlights the activity of helper nominees by relaying their messages to the talent pool channel."""
```

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,233 @@
...
+from bot.decorators import with_role
+from bot.pagination import LinePaginator
+from .watchchannel import WatchChannel, proxy_user
+
+log = logging.getLogger(__name__)
+STAFF_ROLES = Roles.owner, Roles.admin, Roles.moderator, Roles.helpers    # <- In constants after the merge?
+
+
+class TalentPool(WatchChannel):
+    """Relays messages of helper candidates to the talent-pool channel to observe them."""
```
MarkKoz:
```suggestion
    """Relays messages of helper candidates to a channel to observe them."""
```

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,233 @@
...
+    async def nomination_group(self, ctx: Context) -> None:
+        """Highlights the activity of helper nominees by relaying their messages to TalentPool."""
+
+        await ctx.invoke(self.bot.get_command("help"), "talentpool")
+
+    @nomination_group.command(name='watched', aliases=('all', 'list'))
+    @with_role(Roles.owner, Roles.admin, Roles.moderator)
+    async def watched_command(self, ctx: Context, update_cache: bool = True) -> None:
+        """
+        Shows the users that are currently being monitored in TalentPool.
```
MarkKoz:
```suggestion
        Shows the users that are currently being monitored in the talent pool.
```

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,233 @@
...
+        """
+        await self.list_watched_users(ctx, update_cache)
+
+    @nomination_group.command(name='watch', aliases=('w', 'add', 'a'))
+    @with_role(Roles.owner, Roles.admin, Roles.moderator)
+    async def watch_command(self, ctx: Context, user: Union[Member, User, proxy_user], *, reason: str) -> None:
+        """
+        Relay messages sent by the given `user` to the `#talent-pool` channel.
+
+        A `reason` for adding the user to TalentPool is required and will displayed
```
MarkKoz:
```suggestion
        A `reason` for adding the user to talent pool is required and will displayed
```

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,233 @@
...
+        in the header when relaying messages of this user to the watchchannel.
```
MarkKoz:
```suggestion
        in the header when relaying messages of this user to the channel.
```

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,233 @@
...
+        if isinstance(user, Member) and any(role.id in STAFF_ROLES for role in user.roles):
+            await ctx.send(f":x: Nominating staff members, eh? Here's a cookie :cookie:")
+            return
+
+        if not await self.fetch_user_cache():
+            await ctx.send(f":x: Failed to update the user cache; can't add {user}")
+            return
+
+        if user.id in self.watched_users:
+            await ctx.send(":x: The specified user is already being watched in the TalentPool")
```
MarkKoz:
```suggestion
            await ctx.send(":x: The specified user is already being watched in the talent pool")
```

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,233 @@
...
+            response_data = await resp.json()
+
+            if resp.status == 400 and response_data.get('user', False):
+                await ctx.send(":x: The specified user can't be found in the database tables")
+                return
+            else:
+                resp.raise_for_status()
+
+        self.watched_users[user.id] = response_data
+        await ctx.send(f":white_check_mark: Messages sent by {user} will now be relayed to TalentPool")
```
MarkKoz:
```suggestion
        await ctx.send(f":white_check_mark: Messages sent by {user} will now be relayed to the talent pool channel")
```

bot/cogs/watchchannels/talentpool.py
```diff
@@ -0,0 +1,233 @@
...
+            f"{self.api_endpoint}/{nomination['id']}",
+            json={'end_reason': reason, 'active': False}
+        )
+        await ctx.send(f":white_check_mark: Messages sent by {user} will no longer be relayed")
+        self._remove_user(user.id)
+
+    @nomination_group.group(name='edit', aliases=('e',), invoke_without_command=True)
+    @with_role(Roles.owner, Roles.admin, Roles.moderator)
+    async def nomination_edit_group(self, ctx: Context) -> None:
+        """Highlights the activity of helper nominees by relaying their messages to TalentPool."""
```
MarkKoz: Looks like you forgot to write the docstring after copy-pasting it.

bot/cogs/watchchannels/talentpool.py
SebastiaanZ: Isn't the point of a `TalentPool` cog class to relay message to a talent pool channel instead of "a" channel?

bot/cogs/watchchannels/talentpool.py
SebastiaanZ: Is it really wrong to use a closed compound for a compound of two nouns? I mean, we're not talking about any channel, but a watchchannel specifically.

bot/cogs/watchchannels/talentpool.py
MarkKoz: Yeah I was a bit torn. Saying "talent pool channel" feels redundant cause we already know we're reading a docstring for the talent pool. On the other hand "a channel" is too vague. On second though the former is fine.

bot/cogs/watchchannels/talentpool.py
MarkKoz: The summary already mentions the talent pool channel so it should be clear by this point. Was also inconsistent with it being called `talent pool channel` in other instances but `watch channel` here.

PR #377, opened 2019-07-01
Change get_pip command regarding file extensions

larswijn:
The pip command would assume either `.txt` or `.rst`, which was not always correct. It now always tries both, and should still error the same way when the pip was not found.


sco1:
The line ending change needs to be reverted, this is impossible to review.

larswijn:
Closed due to other pull request having better updated write history

See: #378

Issue #367, opened 2019-05-21
PEP command incorrectly assumes file extension type


kosayoda:
The current implementation of the PEP command assumes that PEPs before PEP 542 are written in `.txt`, and PEPs after PEP 542 are written in `.rst`.
https://github.com/python-discord/bot/blob/301e877beff8436de0b0d6f3360e3bcf0d81bf01/bot/cogs/utils.py#L40-L43

[That is not the case.](https://github.com/python/peps)
PEPs before 542 that are written in `.rst`:
* `pep-0012`, `pep-0013`, `pep-0505`

Peps after 542 that are written in `.txt`:
* `pep-0544`, `pep-0545`, `pep-0546`, `pep-0628`. `pep-0666`, `pep-0754`
* `pep-3xxx`

There was a [PR](https://github.com/python/peps/pull/462) in 2017 that renamed every `.txt` PEP file to `.rst` that got merged, but was later reverted because revision history was lost doing so. Therefore it is very likely these edge cases will remain.

eivl:
so instead of assuming one or the other, could we just check for both?

SebastiaanZ:
I think that's a good idea.

We could still implement a look-up order with the current division (or a better one) to minimize the number of API calls we make. I'm not sure how relevant that is and how often this feature is used, though.

GhostofGoes:
Unfortunately, I think the simplest solution would be to write a script to generate a mapping of pep number to file extension, then hardcode that lookup table in the bot.

lemonsaurus:
simplest solution is to hardcode a lookup table? but new peps get added all the time. that sounds like by far the most complicated solution, the simplest would be to just make two requests. our bot could technically even make them in parallel since it's async, thus taking roughly the same time as it does now.

GhostofGoes:
I suppose making extra requests doesn't hurt, it just unnecessary work for every issue of the command. However, the reduction in maintenance overhead is probably worth it.

eivl:
looks like the command have been used less than 100 times by our users. there are mostly staff that uses it, and that is not even remotely high. We are talking among the magnitude of less than once per day. also, 90% is !pep 8, so maybe make a "normal" tag out of it? one that is more descriptive of what it means and is.

SebastiaanZ:
[...]

PR #378, opened 2019-07-01
Change pep command to use `.txt` by default

larswijn:
The pep command would assume either .txt or .rst, which was not always correct. It now always tries both, and should still error the same way when the pip was not found. This is the 2nd time this pull request has been opened because the write history of the previous one had been destroyed.

Closes: #367

### DIFF
--- a/bot/cogs/utils.py
+++ b/bot/cogs/utils.py
@@ -38,7 +38,7 @@ class Utils:
         else:
             return await ctx.invoke(self.bot.get_command("help"), "pep")
 
-        possible_extensions = ['.rst', '.txt']
+        possible_extensions = ['.txt', '.rst']
         found_pep = False
         for extension in possible_extensions:
             # Attempt to fetch the PEP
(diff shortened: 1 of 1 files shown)


jchristgit:
Couldn't we first try to fetch it by the current logic:
```py
        if pep_number > 542:
	            pep_url = f"{self.base_github_pep_url}{pep_number:04}.rst"
	        else:
	            pep_url = f"{self.base_github_pep_url}{pep_number:04}.txt"
```
and fall back to the other filetype if one returns 404?

larswijn:
You could, but... I personally don't see a reason to do so:
Up until pep 542 almost all of the peps are in `.txt` format, and after pep 542, they're equally split between `.rst` and `.txt`. It would be useful to start with the `.txt` extension first, since that one comes up a lot more often, which would be a simple case of switching around the order that's tried which is just a list in the code.

lemonsaurus:
Yeah, it's weird that we're trying it in the wrong order currently. Sounds like the right approach is:

1. try `.txt`
2. If and only if this fails, try `.rst`

this would lead to far fewer requests, but it would take twice as long to resolve the `.rst` ones because of 2 sequential requests. However, I don't think it'll take so long that most people will notice or mind, so I think that's fine.

scragly:
@larswijn Would you like to revisit this if you get some time.

larswijn:
Updated, should be finished now.

PR #379, opened 2019-07-03
Fix !pep command

Akarys42:
PEPs are now in .txt format


kosayoda:
Hello, thank you for showing interest in contributing to our bot.

However, as it stands now, @larswijn already has a PR (See: #378) in place regarding the issue, so we are letting him handle it. By the way, we are currently in a feature freeze, so it would be good if you coordinate with staff before working on another issue. Thanks!

PR #392, opened 2019-08-05
re-add django event handlers, allow for reading out responses in API errors in case something breaks

jchristgit:
I nuked the event handlers for some reason.

Actually, I think the reason was that I wrote the new `syncers` cog which syncs users and roles from the server over to the API (and that made the old handlers doing that irrelevant), and for some reason thought that completely removing the error handler was a good idea. Oh well.


bot/cogs/error_handler.py
```diff
@@ -0,0 +1,89 @@
...
+from discord.ext.commands import Bot, Context
+
+from bot.api import ResponseCodeError
+
+
+log = logging.getLogger(__name__)
+
+
+class ErrorHandler:
+    """Handles errors emttted from commands."""
```
```suggestion
    """Handles errors emitted from commands."""
```

bot/api.py
```diff
@@ -16,33 +21,40 @@ def __init__(self, **kwargs):
         else:
             kwargs['headers'] = auth_headers
```
MarkKoz: Should this be `.update()` instead so that it doesn't override custom headers passed?

SebastiaanZ:
When you enter a command that does not exist, e.g. `!volcyy`, the bot will fall back to polling the API for the existence of a tag. However, if that tag also does not exist, the result `raise_for_status` is not handled (status=404, obviously):

```py
Ignoring exception in command_error
  File "/home/sebastiaan/.local/share/virtualenvs/djangobot-Oms2CWkR/src/discord-py/discord/client.py", line 225, in _run_event
    await coro(*args, **kwargs)
  File "/home/sebastiaan/pydis/djangobot/bot/cogs/error_handler.py", line 49, in on_command_error
    return await ctx.invoke(tags_get_command, tag_name=ctx.invoked_with)
  File "/home/sebastiaan/.local/share/virtualenvs/djangobot-Oms2CWkR/src/discord-py/discord/ext/commands/context.py", line 127, in invoke
  File "/home/sebastiaan/pydis/djangobot/bot/cogs/tags.py", line 79, in get_command
    tag = await self.bot.api_client.get(f'bot/tags/{tag_name}')
  File "/home/sebastiaan/pydis/djangobot/bot/api.py", line 36, in get
    self.maybe_raise_for_status(resp, raise_for_status)
  File "/home/sebastiaan/pydis/djangobot/bot/api.py", line 32, in maybe_raise_for_status
    raise ResponseCodeError(response=response)
bot.api.ResponseCodeError
```

This doesn't happen in the current bot because the `raise_for_status` is a new feature of the API client.

bot/cogs/error_handler.py
```diff
@@ -0,0 +1,89 @@
...
+            tags_get_command = self.bot.get_command("tags get")
+            ctx.invoked_from_error_handler = True
+
+            # Return to not raise the exception
+            return await ctx.invoke(tags_get_command, tag_name=ctx.invoked_with)
+        elif isinstance(e, BadArgument):
+            await ctx.send(f"Bad argument: {e}\n")
+            await ctx.invoke(*help_command)
+        elif isinstance(e, UserInputError):
+            await ctx.invoke(*help_command)
```
MarkKoz: Should a message still be sent here?

jchristgit:
@SebastiaanZ that's a good point, addressed in bea4e8d24122a3ef80ba000d11a6bd68dad267a5.

bot/cogs/error_handler.py
jchristgit: Yes, I think it makes sense - if the user did something wrong, show help for the command. Or does that not make sense?

bot/api.py
jchristgit: Isn't that what it's doing in case the headers are given?

bot/cogs/error_handler.py
MarkKoz: I mean a message in addition to the help command.

bot/api.py
MarkKoz: You're right. I couldn't see the code above.

bot/cogs/error_handler.py
jchristgit: Added in 74f4162d2859333909c74b038ee3a1dcaee65c9c.

SebastiaanZ: In what kind of situations is this exception raised? When an argument can't be parsed, like an invalid value for a duration, BadArgument gets raised and that one does already provide feedback about which argument is problematic.

I would only add a specific message to the user if it actually conveys useful information, but I'm not sure when this exception even happens.

bot/cogs/error_handler.py
```diff
@@ -0,0 +1,91 @@
...
+            await ctx.send(
+                f"Sorry, it looks like I don't have the permissions I need to do that.\n\n"
+                f"Here's what I'm missing: **{e.missing_perms}**"
+            )
+        elif isinstance(e, CommandInvokeError):
+            if isinstance(e.original, ResponseCodeError):
+                if e.original.response.status == 404:
+                    await ctx.send("There does not seem to be anything matching your query.")
+                elif e.original.response.status == 400:
+                    content = await e.original.resopnse.json()
```
SebastiaanZ: typo in `response`

```suggestion
                    content = await e.original.response.json()
```

jchristgit: Mark lost the test.

PR #396, opened 2019-08-14
Write logs into the site

### LINKED ISSUE #540: Don't show infraction total outside staff channels.
Currently, when we deliver an infraction, it will show the infraction total in the bot's response.

image

This is a cool feature, but should not happen in public channels. So let's do something about that.

### Infraction total should be allowed in the following channels:
```
#admins        ID: 365960823622991872
#admin-spam    ID: 563594791770914816
#mod-spam      ID: 620607373828030464
#mods          ID: 305126844661760000
#helpers       ID: 385474242440986624
#organisation  ID: 551789653284356126
#defcon        ID: 464469101889454091
```

If the command is called in any other channel, **do not show the infraction total**. This applies to all moderation commands that currently show the total.

If any of the above channels are not currently registered as constants, please create new constants for them. The above list of channels can be stored as a group constant called `STAFF_CHANNELS`. Make use of [YAML node anchors](https://yaml.org/spec/1.2/spec.html#&%20anchor//) when you do this.

jchristgit:
5 months later

Closes #540

Example: 
image

### DIFF
--- a/bot/api.py
+++ b/bot/api.py
@@ -1,8 +1,12 @@
+import asyncio
+import logging
 from urllib.parse import quote as quote_url
 
 import aiohttp
 
 from .constants import Keys, URLs
+
+log = logging.getLogger(__name__)
 
 
 class ResponseCodeError(ValueError):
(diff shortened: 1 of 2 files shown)


bot/api.py
```diff
@@ -58,3 +62,76 @@ def maybe_raise_for_status(self, response: aiohttp.ClientResponse, should_raise:
...
+    try:
+        asyncio.get_running_loop()
+    except RuntimeError:
+        return False
+    return True
+
+
+class APILoggingHandler(logging.StreamHandler):
+    def __init__(self, client: APIClient):
+        logging.StreamHandler.__init__(self)
```
MarkKoz: Why not `super()`? Don't trust its black magic?

jchristgit: Nope, nope, nope

PR #403, opened 2019-09-05
Update pythonpackage.yml

opabravo:


sco1:
Hello! Not really sure what the intent of this is, as there's no description provided & we're currently using Azure pipelines.

Issue #384, opened 2019-07-10
Update Linting Dependencies


sco1:
The bot repository needs some linting adjustments to bring it in line with the rest of the organization.

- [x] Add flake8-docstrings as a dev dependency
- [x] Add flake8-annotations as a dev dependency
- [x] Relock dependencies
- [x] Update linting rules (See: https://github.com/python-discord/bot/issues/335)
- [x] Relint repository
- [x] Add pre-commit as a dev dependency
- [x] Add pre-commit configuration
- [x] Add pre-commit install to pipenv scripts

Because of the `>1000` lines of changes that will be required required once flake8-docstrings is added, this should be deferred until the django migration is very near completion so we don't unnecessarily contribute to the merging nightmare.

PR #406, opened 2019-09-11
Update linting

sco1:
Per #384, this PR updates the django branch's linting dependencies & relints the codebase.

To potentially help ease review I've split the diffs into discrete chunks, but can squash them into a single commit if preferable.

Closes: #384


bot/decorators.py
```diff
@@ -50,42 +51,34 @@ def predicate(ctx: Context):
...
-    async def predicate(ctx: Context):
+def without_role(*role_ids: int) -> Callable:
+    """Returns True if the user does not have any of the roles in role_ids."""
+    async def predicate(ctx: Context) -> bool:
         return without_role_check(ctx, *role_ids)
     return commands.check(predicate)
 
 
-def locked():
+def locked() -> Union[Callable, None]:
```
SebastiaanZ: Why is this a `Union` of `Callable` and `None`? Doesn't this always return the callable `wrap`?

```suggestion
def locked() -> Callable:
```

bot/decorators.py
```diff
@@ -50,42 +51,34 @@ def predicate(ctx: Context):
...
-    Subsequent calls to the command from the same author are
-    ignored until the command has completed invocation.
+
+    Subsequent calls to the command from the same author are ignored until the command has completed invocation.
 
     This decorator has to go before (below) the `command` decorator.
     """
-
-    def wrap(func):
+    def wrap(func: Callable) -> Union[Callable, None]:
```
SebastiaanZ: The same here, doesn't this always return the callable `inner`?

```suggestion
    def wrap(func: Callable) -> Callable:
```

bot/decorators.py
```diff
@@ -50,42 +51,34 @@ def predicate(ctx: Context):
...
         func.__locks = WeakValueDictionary()
 
         @wraps(func)
-        async def inner(self, ctx, *args, **kwargs):
+        async def inner(self: Callable, ctx: Context, *args, **kwargs) -> Union[Callable, None]:
```
SebastiaanZ: `inner` does not return a callable, but a union of `None` and the return value of the decorated original callable (`func`) in `return await func(...)`.

Since we don't strictly enforce commands to return `None`, by allowing the `return await ctx.send` pattern, I don't think we can make a reasonable assumption of what `func` will return, so I guess we could use `Optional[Any]` or something.

```suggestion
        async def inner(self: Callable, ctx: Context, *args, **kwargs) -> Optional[Any]:
```
(Note: if you like this suggestion, please also include `Any` and `Optional` in the `from typing import`; I've left a suggestion for that as well, to simplify that.)

bot/decorators.py
```diff
@@ -105,15 +98,15 @@ def wrap(func):
...
     """
+    Changes the channel in the context of the command to redirect the output to a certain channel.
 
-    def wrap(func):
+    Redirect is bypassed if the author has a role to bypass redirection.
+    """
+    def wrap(func: Callable) -> Callable:
         @wraps(func)
-        async def inner(self, ctx, *args, **kwargs):
+        async def inner(self: Callable, ctx: Context, *args, **kwargs) -> Callable:
```
SebastiaanZ: This `inner` returns `None`.

Interestingly, I did not think to capture the return value of `func` and return it, which would probably have been a better practice (then, again, we should only apply this decorator to commands, which we do sometimes allow to return something, but I don't think we ever use the return value).

```suggestion
        async def inner(self: Callable, ctx: Context, *args, **kwargs) -> None:
```

bot/decorators.py
```diff
@@ -1,9 +1,9 @@
 import logging
 import random
-import typing
 from asyncio import Lock, sleep
 from contextlib import suppress
 from functools import wraps
+from typing import Callable, Container, Union
```
SebastiaanZ: Suggestion for if you agree with the suggestion for the `inner` of `redirect_output`.

```suggestion
from typing import Any, Callable, Container, Optional, Union
```

bot/pagination.py
```diff
@@ -100,39 +89,24 @@ def add_line(self, line='', *, empty=False):
     async def paginate(cls, lines: Iterable[str], ctx: Context, embed: Embed,
                        prefix: str = "", suffix: str = "", max_lines: Optional[int] = None, max_size: int = 500,
                        empty: bool = True, restrict_to_user: User = None, timeout: int = 300,
-                       footer_text: str = None, url: str = None, exception_on_empty_embed: bool = False):
+                       footer_text: str = None, url: str = None, exception_on_empty_embed: bool = False) -> None:
```
SebastiaanZ: This functions has the `return await ctx.send` pattern in a couple of its paths, so this should be a `Optional[Message]` return annotation. E.g., see [line 165](https://github.com/python-discord/bot/blob/138567f371fa2db8675c2b5661ff5526e774335a/bot/pagination.py#L165).

I can't suggest it properly, though, since it would bring it over the line limit and the imports needs to be adjusted.(`Message` should be added to the `from discord import` line.)

bot/pagination.py
```diff
@@ -327,50 +297,30 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
-        :param image: image url to be appended
-        """
-
+        """Adds an image to a page."""
         self.images.append(image)
 
     @classmethod
     async def paginate(cls, pages: List[Tuple[str, str]], ctx: Context, embed: Embed,
                        prefix: str = "", suffix: str = "", timeout: int = 300,
                        exception_on_empty_embed: bool = False):
```
SebastiaanZ: This `classmethod` does not have a return annotation. https://github.com/python-discord/bot/blob/138567f371fa2db8675c2b5661ff5526e774335a/bot/pagination.py#L303-L306

There's at least one path that has the `return await ctx.send` pattern, so it's an `Optioonal[Message]` (`Message` should still be added to the import list, see previous comment.)
```suggestion
                       exception_on_empty_embed: bool = False) -> Optional[Message]:
```

bot/cogs/wolfram.py
```diff
@@ -65,14 +54,10 @@
 
 def custom_cooldown(*ignore: List[int]) -> check:
```
SebastiaanZ: Not your code, but this return type annotation is wrong. We call `check` which returns a `Callable`

**Note:** This means we should add `Callable` to the `from typing import` list.
```suggestion
def custom_cooldown(*ignore: List[int]) -> Callable:
```

bot/decorators.py
```diff
@@ -105,15 +98,15 @@ def wrap(func):
...
     """
+    Changes the channel in the context of the command to redirect the output to a certain channel.
 
-    def wrap(func):
+    Redirect is bypassed if the author has a role to bypass redirection.
+    """
+    def wrap(func: Callable) -> Callable:
         @wraps(func)
-        async def inner(self, ctx, *args, **kwargs):
+        async def inner(self: Callable, ctx: Context, *args, **kwargs) -> None:
```
SebastiaanZ: I made a mistake during my previous review, this function does have two paths that include `return await func`, so I think we should make this `Optional[Any]` or just `Any`, since `Any` is compatible with `None`. 
```suggestion
        async def inner(self: Callable, ctx: Context, *args, **kwargs) -> Any:
```

bot/utils/__init__.py
```diff
@@ -16,50 +18,59 @@ class CaseInsensitiveDict(dict):
     """
 
     @classmethod
-    def _k(cls, key):
+    def _k(cls, key: Hashable) -> Any:
```
SebastiaanZ: Bit of a nitpick, but since `key` is Hashable, it returns a Hashable as well (str.lower always returns a Hashable str)

```suggestion
    def _k(cls, key: Hashable) -> Hashable:
```

bot/cogs/eval.py
```diff
@@ -31,7 +29,8 @@ def __init__(self, bot: Bot):
 
         self.interpreter = Interpreter(bot)
 
-    def _format(self, inp, out):  # (str, Any) -> (str, discord.Embed)
+    def _format(self, inp: str, out: Any) -> Tuple[str, Optional[discord.embed]]:
```
SebastiaanZ: Just noticed a type when running this file.

```suggestion
    def _format(self, inp: str, out: Any) -> Tuple[str, Optional[discord.Embed]]:
```

bot/cogs/eval.py
```diff
@@ -174,16 +174,15 @@ def _format(self, inp, out):  # (str, Any) -> (str, discord.Embed)
 
     @group(name='internal', aliases=('int',))
     @with_role(Roles.owner, Roles.admin)
-    async def internal_group(self, ctx):
+    async def internal_group(self, ctx: discord.Context) -> None:
```
SebastiaanZ: It's `discord.ext.commands.Context`; I'd recommend importing as `Context`
```suggestion
    async def internal_group(self, ctx: Context) -> None:
```

bot/cogs/eval.py
```diff
@@ -174,16 +174,15 @@ def _format(self, inp, out):  # (str, Any) -> (str, discord.Embed)
...
         """Internal commands. Top secret!"""
-
         if not ctx.invoked_subcommand:
             await ctx.invoke(self.bot.get_command("help"), "internal")
 
     @internal_group.command(name='eval', aliases=('e',))
     @with_role(Roles.admin, Roles.owner)
-    async def eval(self, ctx, *, code: str):
-        """ Run eval in a REPL-like format. """
+    async def eval(self, ctx: discord.Context, *, code: str) -> None:
```
SebastiaanZ: Here as well

```suggestion
    async def eval(self, ctx: Context, *, code: str) -> None:
```

bot/cogs/eval.py
```diff
@@ -6,6 +6,7 @@
 import textwrap
 import traceback
 from io import StringIO
+from typing import Any, Optional, Tuple
 
 import discord
 from discord.ext.commands import Bot, Cog, group
```
SebastiaanZ:
```suggestion
from discord.ext.commands import Bot, Cog, Context, group
```

SebastiaanZ:
Apparently, I messed a few things during my review; apologies.

bot/cogs/eval.py
```diff
@@ -124,7 +123,8 @@ def _format(self, inp, out):  # (str, Any) -> (str, discord.Embed)
 
         return res  # Return (text, embed)
 
-    async def _eval(self, ctx, code):  # (discord.Context, str) -> None
+    async def _eval(self, ctx: discord.Context, code: str) -> Optional[discord.Message]:
```
SebastiaanZ: This seems to be the last one, although this one doesn't prevent the bot from running (I think).

There's one other mention of `discord.Context` in the docstring of the `HelpSession` class, but that's not really within the scope of this PR.
```suggestion
    async def _eval(self, ctx: Context, code: str) -> Optional[discord.Message]:
```

bot/cogs/superstarify/__init__.py
```diff
@@ -224,15 +225,8 @@ def modlog(self) -> ModLog:
...
-        """
-        This command will remove the entry from our database, allowing the user
-        to once again change their nickname.
-
-        :param ctx: Discord message context
-        :param member: The member to unsuperstarify
-        """
-
+    async def unsuperstarify(self, ctx: Context, member: Member) -> Optional[Message]:
+        """This command will the superstarify entry from our database, allowing the user to change their nickname."""
```
lemonsaurus:
```suggestion
        """This command will remove the superstarify entry from our database, allowing the user to change their nickname."""
```

bot/interpreter.py
```diff
@@ -8,13 +11,20 @@
 
 
 class Interpreter(InteractiveInterpreter):
+    """
+    Subclass InteractiveInterpreter to specify custom run functionality.
+
+    Helper class for internal eval
```
lemonsaurus:
```suggestion
    Helper class for internal eval.
```

bot/utils/messages.py
```diff
@@ -17,42 +17,18 @@
...
-
-        client (Optional[discord.Client]):
-            The client instance handling the original command.
-            If not given, will take the client from the guild
-            of the message.
+    attach_emojis: bool = True,
+    client: Optional[Client] = None
+) -> None:
     """
+    Waits for up to `timeout` seconds for a reaction by any of the specified `user_ids` to delete the message.
```
lemonsaurus:
```suggestion
    Wait for up to `timeout` seconds for a reaction by any of the specified `user_ids` to delete the message.
```

bot/cogs/doc.py
```diff
@@ -121,31 +119,33 @@ class InventoryURL(commands.Converter):
...
-        :param config: A `SphinxConfiguration` instance to mock the regular sphinx
-                       project layout. Required for use with intersphinx.
+        Where:
+            * `package_name` is the package name to use, appears in the log
+            * `base_url` is the root documentation URL for the specified package, used to build
+                         absolute paths that link to specific symbols
+            * `inventory_url` is the absolute URL to the intersphinx inventory, fetched by running
+                              `intersphinx.fetch_inventory` in an executor on the bot's event loop
+            * `config` is a `SphinxConfiguration` instance to mock the regular sphinx
+                       project layout, required for use with intersphinx
```
MarkKoz: I'm not a fan of this sort of indentation. I think it'd look better if there was no indentation (the `*` may visually be enough for a reader to recognise a new item in the list) or at least indenting by a consistent amount (e.g. 4 spaces).

bot/cogs/error_handler.py
```diff
@@ -30,7 +30,8 @@ def __init__(self, bot: Bot):
         self.bot = bot
 
     @Cog.listener()
-    async def on_command_error(self, ctx: Context, e: CommandError):
+    async def on_command_error(self, ctx: Context, e: CommandError) -> None:
+        """Provide command error handling."""
```
MarkKoz: I think this could benefit from a more detailed docstring but perhaps that is out of scope of this PR.

bot/cogs/filtering.py
```diff
@@ -94,28 +91,29 @@ def __init__(self, bot: Bot):
...
+        """Invoke message filter for new messages."""
         await self._filter_message(msg)
 
     @Cog.listener()
-    async def on_message_edit(self, before: Message, after: Message):
+    async def on_message_edit(self, before: Message, after: Message) -> None:
+        """
+        Invoke message filter for message edits.
+
+        If there have been multiple edits, calculate the time delta from the previous edit
```
MarkKoz:
```suggestion
        If there have been multiple edits, calculate the time delta from the previous edit.
```

bot/cogs/off_topic_names.py
```diff
@@ -19,7 +19,8 @@ class OffTopicName(Converter):
     """A converter that ensures an added off-topic name is valid."""
 
     @staticmethod
-    async def convert(ctx: Context, argument: str):
+    async def convert(ctx: Context, argument: str) -> str:
+        """Attempt to replace any invalid characters with their approximate unicode equivalent."""
```
MarkKoz:
```suggestion
        """Attempt to replace any invalid characters with their approximate Unicode equivalent."""
```

MarkKoz:
I'm currently reviewing this by the way

bot/cogs/off_topic_names.py
```diff
@@ -47,7 +48,6 @@ class OffTopicName(Converter):
             The running bot instance, used for fetching data from the
             website via the bot's `api_client`.
```
MarkKoz: There doesn't seem to be a consistent way we're listing parameters in docstrings. For example, this is different compared to one of the docstrings in the docs cog. I was under the impression we'd just add a separate paragraph for each parameter akin to Python's own documentation. Such a list seems unnecessary anyway in this case given there's only one argument.

bot/cogs/help.py
```diff
@@ -68,13 +69,16 @@ class HelpSession:
         Where the help message is to be sent to.
     """
 
-    def __init__(self, ctx, *command, cleanup=False, only_can_run=True, show_hidden=False, max_lines=15):
+    def __init__(
+        self, ctx: Context, *command, cleanup: bool = False, only_can_run: bool = True,
+        show_hidden: bool = False, max_lines: int = 15
```
MarkKoz: Isn't the prevailing style to put each parameter on a separate line?

bot/cogs/help.py
```diff
@@ -68,13 +69,16 @@ class HelpSession:
...
+    ):
         """
         Creates an instance of the HelpSession class.
 
         Parameters
         ----------
-        ctx: :class:`discord.Context`
+        ctx: :class:`discord.ext.commands.Context`
```
MarkKoz: I thought we were doing away with this sort of format in the docstrings. It's also present in class docstrings here but I cannot leave separate review comments for them.

bot/cogs/help.py
```diff
@@ -150,12 +141,11 @@ def _get_query(self, query):
...
         self._handle_not_found(query)
 
-    def _handle_not_found(self, query):
+    def _handle_not_found(self, query: str) -> None:
         """
         Handles when a query does not match a valid command or cog.
 
-        Will pass on possible close matches along with the
-        ``HelpQueryNotFound`` exception.
+        Will pass on possible close matches along with the ``HelpQueryNotFound`` exception.
```
MarkKoz:
```suggestion
        Will pass on possible close matches along with the `HelpQueryNotFound` exception.
```

bot/cogs/help.py
```diff
@@ -150,12 +141,11 @@ def _get_query(self, query):
...
 
         Parameters
         ----------
```
MarkKoz: I thought we were doing away with this sort of format in the docstrings.

bot/cogs/help.py
```diff
@@ -183,15 +172,11 @@ def _handle_not_found(self, query):
         seconds: int
             Number of seconds to wait.
```
MarkKoz: I thought we were doing away with this sort of format in the docstrings.

bot/cogs/help.py
```diff
@@ -211,7 +196,6 @@ def reset_timeout(self):
         user: :class:`discord.User`
             The user who added the reaction.
```
MarkKoz: I thought we were doing away with this sort of format in the docstrings.

bot/cogs/help.py
```diff
@@ -281,44 +250,22 @@ def add_reactions(self):
...
-
-        Parameters
-        ----------
-        cmd: :class:`discord.ext.commands.Command`
-            The command object to get the parameters of.
-
-        Returns
-        -------
-        str
+        This is a custom implementation of ``command.signature`` in order to format the command
```
MarkKoz:
```suggestion
        This is a custom implementation of `command.signature` in order to format the command
```

bot/cogs/help.py
```diff
@@ -281,44 +250,22 @@ def add_reactions(self):
...
-        ``sorted`` and ``groupby``.
-
-        A zero width space is used as a prefix for results with no cogs
-        to force them last in ordering.
-
-        Parameters
-        ----------
-        cmd: :class:`discord.ext.commands.Command`
-            The command object being checked.
+        Returns a cog name of a given command for use as a key for ``sorted`` and ``groupby``.
```
MarkKoz:
```suggestion
        Returns a cog name of a given command for use as a key for `sorted` and `groupby`.
```

bot/cogs/help.py
```diff
@@ -703,13 +571,7 @@ def setup(bot):
 
     If an exception is raised during the loading of the cog, ``unload``
```
MarkKoz:
```suggestion
    If an exception is raised during the loading of the cog, `unload`
```

bot/cogs/modlog.py
```diff
@@ -70,7 +66,8 @@ def __init__(self, bot: Bot):
 
         return f"{URLs.site_logs_view}/{response['id']}"
 
-    def ignore(self, event: Event, *items: int):
+    def ignore(self, event: Event, *items: int) -> None:
+        """Add event to ignored events to suppress log emitting."""
```
MarkKoz:
```suggestion
        """Add event to ignored events to suppress log emission."""
```

bot/cogs/wolfram.py
```diff
@@ -109,8 +94,8 @@ def custom_cooldown(*ignore: List[int]) -> check:
     return check(predicate)
 
 
-async def get_pod_pages(ctx, bot, query: str) -> Optional[List[Tuple]]:
-    # Give feedback that the bot is working.
+async def get_pod_pages(ctx: Context, bot: Bot, query: str) -> Optional[List[Tuple]]:
+    """Give feedback that the bot is working."""
```
MarkKoz: This function does more than just show that the bot is typing.

bot/pagination.py
```diff
@@ -18,6 +18,8 @@
 
 
 class EmptyPaginatorEmbed(Exception):
+    """Empty paginator embed exception."""
```
MarkKoz:
```suggestion
    """Raised when attempting to paginate with empty contents."""
```

bot/pagination.py
```diff
@@ -37,13 +39,13 @@ class LinePaginator(Paginator):
         The maximum amount of lines allowed in a page.
     """
 
-    def __init__(self, prefix='```', suffix='```',
-                 max_size=2000, max_lines=None):
+    def __init__(
+        self, prefix: str = '```', suffix: str = '```', max_size: int = 2000, max_lines: int = None
```
MarkKoz: Isn't the prevailing style to put each parameter on a separate line?

bot/pagination.py
```diff
@@ -97,42 +86,29 @@ def add_line(self, line='', *, empty=False):
...
     @classmethod
-    async def paginate(cls, lines: Iterable[str], ctx: Context, embed: Embed,
-                       prefix: str = "", suffix: str = "", max_lines: Optional[int] = None, max_size: int = 500,
-                       empty: bool = True, restrict_to_user: User = None, timeout: int = 300,
-                       footer_text: str = None, url: str = None, exception_on_empty_embed: bool = False):
+    async def paginate(
+        cls, lines: Iterable[str], ctx: Context, embed: Embed,
+        prefix: str = "", suffix: str = "", max_lines: Optional[int] = None, max_size: int = 500,
+        empty: bool = True, restrict_to_user: User = None, timeout: int = 300,
+        footer_text: str = None, url: str = None, exception_on_empty_embed: bool = False
```
MarkKoz: Isn't the prevailing style to put each parameter on a separate line?

bot/pagination.py
```diff
@@ -327,50 +299,32 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
         self.images.append(image)
 
     @classmethod
-    async def paginate(cls, pages: List[Tuple[str, str]], ctx: Context, embed: Embed,
-                       prefix: str = "", suffix: str = "", timeout: int = 300,
-                       exception_on_empty_embed: bool = False):
+    async def paginate(
+        cls, pages: List[Tuple[str, str]], ctx: Context, embed: Embed,
+        prefix: str = "", suffix: str = "", timeout: int = 300,
+        exception_on_empty_embed: bool = False
```
MarkKoz: Isn't the prevailing style to put each parameter on a separate line?

bot/cogs/moderation.py
```diff
@@ -635,19 +600,14 @@ def mod_log(self) -> ModLog:
...
         """
-        Create a temporary mute infraction in the database for a user.
+        Create a temporary mute infraction for a user with the provided reason.
 
-        **`user`:** Accepts user mention, ID, etc.
-        **`duration`:** The duration for the temporary mute infraction
-        **`reason`:** The reason for the temporary mute.
+        Duration strings are parsed per: http://strftime.org/
+
+        This does not send the user a notification.
```
MarkKoz: This should probably be put in the summary so it can be distinguished from the regular counterpart more easily.

bot/cogs/moderation.py
```diff
@@ -693,19 +653,14 @@ def mod_log(self) -> ModLog:
...
         """
-        Create a temporary ban infraction in the database for a user.
+        Create a temporary ban infraction for a user with the provided reason.
 
-        **`user`:** Accepts user mention, ID, etc.
-        **`duration`:** The duration for the temporary ban infraction
-        **`reason`:** The reason for the temporary ban.
+        Duration strings are parsed per: http://strftime.org/
+
+        This does not send the user a notification.
```
MarkKoz: This should probably be put in the summary so it can be distinguished from the regular counterpart more easily.

bot/cogs/snekbox.py
```diff
@@ -169,7 +168,7 @@ def get_results_message(results: dict) -> Tuple[str, str]:
     @command(name="eval", aliases=("e",))
     @guild_only()
     @in_channel(Channels.bot, bypass_roles=STAFF_ROLES)
-    async def eval_command(self, ctx: Context, *, code: str = None):
+    async def eval_command(self, ctx: Context, *, code: str = None) -> Optional[Message]:
```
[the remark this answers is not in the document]
sco1: I don't understand what this means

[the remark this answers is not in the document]
MarkKoz:
```py
# now
return await ctx.send(...)

# instead do this
await ctx.send()
return
```
 
Thus the return type of the command can be `None`.

bot/cogs/moderation.py
sco1: I don't see this as particularly unclear, and the resulting one-liner for many of them don't fit as single line docstrings.

bot/cogs/moderation.py
sco1: I don't see this as particularly unclear, and the resulting one-liner for many of them don't fit as single line docstrings.

PR #417, opened 2019-09-15
Validate bot.cogs.antispam configuration on CI.

jchristgit:


jchristgit:
the continue is since otherwise unknown rules have also set required parameter errors which doesnt make a lot of sense to me

bot/cogs/antispam.py
```diff
@@ -245,16 +246,17 @@ def mod_log(self) -> ModLog:
         await deletion_context.upload_messages(self.bot.user.id, self.mod_log)
 
 
-def validate_config() -> bool:
+def validate_config(rules: Mapping = AntiSpamConfig.rules) -> dict:
```
SebastiaanZ: Since we map strings to strings, this could be:
```suggestion
def validate_config(rules: Mapping = AntiSpamConfig.rules) -> Dict[str, str]:
```

PR #418, opened 2019-09-15
Add tests for `bot.converters`.

jchristgit:
I'm going insane. There is a commented out parameter there where pytest fails and I have no idea why. Specifically:

```py
        # ('x' * 128, "Are you insane? That's way too long!"),  # ???ß
```

Please give that a shot. I have no idea why it's failing, because to the blank eye as well as a Python interpreter (and I directly copied the error string!), the exception raised and the one in the code are the exact same.

### DIFF
--- a/tests/test_converters.py
+++ b/tests/test_converters.py
@@ -1,0 +1,93 @@
+import asyncio
+from datetime import datetime
+from unittest.mock import MagicMock
+
+import pytest
+from discord.ext.commands import BadArgument
+
+from bot.converters import (
+    ExpirationDate,
+    TagContentConverter,
+    TagNameConverter,
+    ValidPythonIdentifier,
+)
+
+
+@pytest.mark.parametrize(
+    ('value', 'expected'),
+    (
+        # sorry aliens
+        ('2199-01-01T00:00:00', datetime(2199, 1, 1)),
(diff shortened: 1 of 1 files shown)


kosayoda:
Looks to be a `pytest` problem, minimal reproducible example is this:

```py
import pytest


def func():
    raise ValueError("?")


@pytest.mark.parametrize(
    ("expected"), (("?"),)
)
def test_func(expected):
    with pytest.raises(ValueError, match=expected):
        func()
```

I have no idea what the reason is, but escaping the `?` character with a backslash in *only the `parametrize` call* works and will pass the test case.

```py
@pytest.mark.parametrize(
    ("expected"), (("\?"),)
)
```

avayert:
in the [documentation for pytest.raises](https://docs.pytest.org/en/latest/reference.html#pytest.raises) it says
 ```
match –

if specified, a string containing a regular expression, or a regular expression object, that is tested against the string representation of the exception using re.search
```

passing the result of `re.escape` as the expected value should work properly

kosayoda:
Ah, that’s definitely it then, I was dumb.

I checked the source and docs for `parametrize` instead.

jchristgit:
Thanks people

PR #421, opened 2019-09-15
Add tests for `bot.utils.time`.

jchristgit:


tests/utils/test_time.py
```diff
@@ -0,0 +1,48 @@
...
+        max_units: int,
+        expected: str
+):
+    assert time.humanize_delta(delta, precision, max_units) == expected
+
+
+@pytest.mark.parametrize(
+    ('stamp', 'expected'),
+    (
+        ('Sun, 15 Sep 2019 12:00:00 GMT', datetime(2019, 9, 15, 12, 0, 0, tzinfo=timezone.utc)),
```
MarkKoz: Parameterisation seems unnecessary when there is only one.

tests/helpers.py
```diff
@@ -0,0 +1,10 @@
+from unittest.mock import MagicMock
+
+
+__all__ = ('AsyncMock',)
+
+
+# TODO: Remove me on 3.8
+class AsyncMock(MagicMock):
```
MarkKoz: Can you add a comment explaining what this is or why it is needed? I have no idea.

tests/utils/test_time.py
jchristgit: Yes, but:

- If we want to add a new test case, we only need to add a new line here
- It's consistent with the other test function

tests/helpers.py
jchristgit: Added

tests/utils/test_time.py
MarkKoz: What other test cases could be added? This seems too simple to warrant that, is all.

jchristgit: That's hard to tell now. But if we ever find something here that requires adding a new test case (e.g. some input causing an error), then this would be simpler to modify

lemonsaurus:
Seems reasonable to me to test erroneous inputs.

MarkKoz:
Are you planning on adding erroneous input tests here @jchristgit ?

jchristgit:
yes, when we find them

MarkKoz:
Here are a few:

* `parse_rfc1123`: a string that is not in the correct format.
    * Maybe this shouldn't be tested since it really just tests if `strptime` raises an exception (which I think we can safely assume will always happen since it's the standard library?)
* `humanize_delta`: an unsupported `precision` 
* `humanize_delta`: `max_units` above what's supported or negative
* `wait_until`: a `start` that's in the future relative to `time` (difference results in negative seconds)

jchristgit:
Agree about parse_rfc1123. I've added test cases for the others. The negative max_units one found an interesting case.
I think we should raise `ValueError` on `max_units <= 0`. What do you think?

jchristgit:
Sleeping for negative seconds will just be gobbled by Python itself, as `asyncio.sleep` will just yield control to the event loop on sleep(n <= 0)

MarkKoz:
>I think we should raise ValueError on max_units <= 0. What do you think?

If it exhibits strange behaviour then yes, just do that. Do you think that when it's too high it should also be raised, for consistency? Or could take it further and also do it for the precision.

jchristgit:
7625d2a

I don't think we should raise for high values as that doesn't cause any odd behaviour.

PR #426, opened 2019-09-15
Add tests for `bot.rules.attachments`.

jchristgit:
This also fixes an issue with the `attachments` rule not respecting the most recent message sent by a user.

### DIFF
--- a/tests/rules/test_attachments.py
+++ b/tests/rules/test_attachments.py
@@ -1,0 +1,52 @@
+import asyncio
+from dataclasses import dataclass
+from typing import Any, List
+
+import pytest
+
+from bot.rules import attachments
+
+
+# Using `MagicMock` sadly doesn't work for this usecase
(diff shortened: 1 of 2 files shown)


tests/rules/test_attachments.py
```diff
@@ -0,0 +1,52 @@
...
+# Using `MagicMock` sadly doesn't work for this usecase
+# since it's __eq__ compares the MagicMock's ID. We just
+# want to compare the actual attributes we set.
+@dataclass
+class FakeMessage:
+    author: str
+    attachments: List[Any]
+
+
+def m(total_attachments: int):
```
scragly: I understand it's probably in order to keep `parametrize` decos shorter, but I'm not a fan of the single-letter function name. Something a bit more self-explanatory would be nicer.

jchristgit: renamed to `msg`

PR #432, opened 2019-09-15
Add coverage reporting & JUnit XML to tests.

jchristgit:


jchristgit:
Do we do that on the site? It doesn't seem like we do, and I have no idea how we would publish the HTML on Azure..

MarkKoz:
I did it on snekbox. It's actually pretty nice. You can reference that repo.

azure-pipelines.yml
```diff
@@ -38,9 +38,21 @@ jobs:
...
+  - task: PublishCodeCoverageResults@1
+    displayName: 'Publish Coverage Results'
+    condition: succeededOrFailed()
+    inputs:
+      codeCoverageTool: Cobertura
+      summaryFileLocation: coverage.xml
+
+  - task: PublishTestResults@2
+    inputs:
+      testResultsFiles: junit.xml
```
MarkKoz: I don't know if pytest returns 1 when it fails. If it does then ensuring the results are published regardless would be a good idea.
```suggestion
      condition: succeededOrFailed()
      testResultsFiles: junit.xml
```

jchristgit: Added

jchristgit:
image

jchristgit:
Does Azure have some place to view a graph of how code coverage evolves over time?

MarkKoz:
> 
> 
> Does Azure have some place to view a graph of how code coverage evolves over time?

I don't see anything. If anywhere, my guess is it would be in the analytics tab. Perhaps some configuration in Azure can be done to add more analytics?

gdude2002:
It [definitely publishes it](https://dev.azure.com/python-discord/Python%20Discord/_build/results?buildId=3801&view=codecoverage-tab), I think maybe it'll add it to the analytics tab once the main branch has some?

gdude2002:
Ah, actually, they're using [this external tool](https://github.com/danielpalme/ReportGenerator) to generate that table. Maybe not, then.

jchristgit:
That tool looks like it can provide a history though and we can install it somehow, can't we? I think being able to view coverage history for the bot on there would be something useful to have.

gdude2002:
image

Taken from here: https://docs.microsoft.com/en-us/azure/devops/pipelines/test/review-code-coverage-results?view=azure-devops

jchristgit:
Doesn't "merging" mean this in that case?

image

as in "take multiple coverages and combine them", not "display a coverage summary"

PR #438, opened 2019-09-18
Infraction Search Dates

LordHemlock:
Since the rework was launched, the current format for the Created and Expires dates on the infraction search embeds use the standard ISO format, which is a bit of a bear to read.

This PR proposes changing that format to `"%Y-%m-%d %H:%M"` which will appear as `2019-09-18 13:59`

### DIFF
--- a/bot/cogs/moderation.py
+++ b/bot/cogs/moderation.py
@@ -1260,11 +1260,11 @@ class Moderation(Scheduler, Cog):
         active = infraction_object["active"]
         user_id = infraction_object["user"]
         hidden = infraction_object["hidden"]
-        created = datetime.fromisoformat(infraction_object["inserted_at"].strftime("%Y-%m-%d %H:%M"))
+        created = datetime.fromisoformat(infraction_object["inserted_at"]).strftime("%Y-%m-%d %H:%M")
         if infraction_object["expires_at"] is None:
             expires = "*Permanent*"
         else:
-            expires = datetime.fromisoformat(infraction_object["expires_at"].strftime("%Y-%m-%d %H:%M"))
+            expires = datetime.fromisoformat(infraction_object["expires_at"]).strftime("%Y-%m-%d %H:%M")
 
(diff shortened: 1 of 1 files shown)


bot/cogs/moderation.py
```diff
@@ -1260,6 +1260,11 @@ def _infraction_to_string(self, infraction_object: Dict[str, Union[str, int, boo
         active = infraction_object["active"]
         user_id = infraction_object["user"]
         hidden = infraction_object["hidden"]
+        created = datetime.fromisoformat(infraction_object["inserted_at"].strftime("%Y-%m-%d %H:%M"))
+        if not infraction_object["expires_at"]:
```
jchristgit: is this none or false? if none, i think `is None` would be nicer

LordHemlock: Fair point, fixing now.

bot/cogs/moderation.py
```diff
@@ -1260,6 +1260,11 @@ def _infraction_to_string(self, infraction_object: Dict[str, Union[str, int, boo
         active = infraction_object["active"]
         user_id = infraction_object["user"]
         hidden = infraction_object["hidden"]
+        created = datetime.fromisoformat(infraction_object["inserted_at"].strftime("%Y-%m-%d %H:%M"))
+        if infraction_object["expires_at"] is None:
+            expires = "*Permanent*"
+        else:
+            expires = datetime.fromisoformat(infraction_object["expires_at"].strftime("%Y-%m-%d %H:%M"))
```
MarkKoz: I'm confused. Why is it doing `datetime -> str -> datetime`? Isn't just a `str` enough i.e. leaving out `datetime.fromisoformat`?

LordHemlock: It's actually having to go `str -> datetime -> str`, as the JSON we get from the request has the date returned as a string and not a datetime object.  Unless I'm missing something.  Which is possible.

PR #443, opened 2019-09-22
Change log.error to log.exception

### LINKED ISSUE #382: Log full traceback with `log.exception` in exception handlers
While trying to reload the `reddit` cog, I noticed that the `cogs` cog doesn't log the full traceback in the `except` blocks, but just the exception message using `log.error`. It would be better to use `log.exception` here to make sure that the full traceback is included instead of just a message like ` 'NoneType' object has no attribute 'startswith'`. (`log.exception` automatically includes the full traceback when used in an exception handler, no additional arguments required.)

I suspect that more cogs do this, so I think it's a good idea to check all the cogs after the Django migration is completed to change the log methods to `log.exception` inside exception handlers where appropriate.

Example from the `cogs` cog (it's both in `master` and `django`):
https://github.com/python-discord/bot/blob/5e16f4a52d59c73a04323e070e7b4a320e8c1e49/bot/cogs/cogs.py#L85
https://github.com/python-discord/bot/blob/25640adec9d042ccf249a91540fb09d354b04dfd/bot/cogs/cogs.py#L85

Akarys42:
`log.exception(msg)` write `msg` in the log, followed by the full traceback. This PR changes all `log.error` to `log.exception` if it's inside a try/except block

This PR closes #382


Akarys42:
Almost all `log.error` are outside of a try/except block, so I can't change them :)

MarkKoz:
Thanks!

PR #444, opened 2019-09-23
Docker Build & CI Refinements

### LINKED ISSUE #407: Update Dependency Pinning
Per https://github.com/python-discord/organisation/issues/98, the star dependencies for the `django` branch should be more conservatively pinned to help mitigate any potential issues.

This is a pretty low priority, and can be handled either before or after django goes live.

### LINKED ISSUE #420: Change Dockerfile base image
We're currently using `alpine` for the bot Dockerfile, and it sucks. It's super slow to install the pipenv dependencies, apparently because alpine has a weird C compiler which makes it compile every pip library from scratch instead of using the wheels.

Let's change it to something like `minideb` or even `ubuntu`. They're still pretty lean images, but they don't have this problem so build times should be reduced _significantly_ just by using a different image.

scragly:
Finally, the last major repository to be looked at. 

### Dockerfile
The base image has been adjusted to use the now-preferred `python:3.7-slim` debian image, in-line with both `site` and `seasonalbot` after their respective refinements.
Closes #420 

The Dockerfile has been brought out of it's `docker` directory as it's the only item within.

Comments have been added to the Dockerfile to encourage stages to be documented if changes occur in future and to aid in ensuring it's easy to understand. Occasionally, I've had to dig through blames to find a reason as to why a certain thing is being installed or why an env var is present.

### Dependencies
Pipfile dependencies have now been pinned to minor versions.
Closes #407 

~~The `regex` package is a sub-dependency for the `dateparser` package we use for parsing our time deltas in commands. The package needs to be built from source every time and is a really slow one to build, slowing both CI and docker builds. I've instead created a pre-built wheel to install from saving the hassle of building every time, and shouldn't be a risk now that deps are pinned in the Pipfile.~~

~~The eventual goal though would be to replace `dateparser` completely as a dependancy with a lib that builds wheels for Debian.~~

Thanks to #449 being merged in earlier than this PR due to being an urgent fix, the dependency is now entirely removed.

### PipEnv Scripts
Script entries for `buildbase`, `pushbase`, `buildci` and `pushci` have all been removed as they're no longer needed.

### Azure Pipelines
Variables were tidied and the script for installing all the build dependencies was removed due to the `regex` wheel being pre-built.

The new docker task format was added to replace the `deploy-azure.sh` which was subsequently deleted.

### Docker Compose
A docker-compose.yml was added that is purely for the purpose of deploying a simple PostgreSQL and Site environment for the bot to hook into. This is useful only for newer or prospective developers who want to test and develop the `bot` repository without going through a huge effort. 

Those who already have set up a proper development environment for `site` don't need to use this.

### Improvements
The uncompressed size is down from 429.7MB to 307.2MB.
The local build speed has dropped from 3:02 to 25.8s

### Reminders Filter Fix
ELA dropped a simple fix for reminders to correctly return reminders based on author.


MarkKoz:
#449 removes `dateparser` so we can get rid of that regex doodoo now.

azure-pipelines.yml
```diff
@@ -6,69 +6,59 @@ variables:
...
+      PIP_CACHE_DIR: ".cache/pip"
+
+    steps:
+      - task: UsePythonVersion@0
+        displayName: 'Set Python version'
+        inputs:
+          versionSpec: '3.7.x'
+          addToPath: true
+
+      - script: sudo pip install pipenv
```
MarkKoz: Using `sudo` makes it use Python 2.7 because I suppose the `PATH` is different for the root user. This can be confirmed by looking at the logs and seeing Python 2.7 in the directories. I suppose it's fine for pipenv to be installed for 2.7 but it just seems unnecessary to even use sudo.

scragly: Agreed. I left it in due to there being a previous commit explicitly adding it, but we don't use it elsewhere. This is an anomaly.

PR #448, opened 2019-09-23
Catch error in case of non successfull API call for new ot-names

### LINKED ISSUE #442: The off-topic channel name updating task fails on non-success API response
The background task we create to update off-topic names at midnight UTC fails if it receives a non-success API response. The reason is that our `bot.api_client` will raise the `bot.api.ResponseCodeError` exception on non-success response status codes. This means that the off-topic channel names won't be updated again until either the bot is restarted or the task is started manually again by an admin.

The relevant lines of code:
https://github.com/python-discord/bot/blob/e70c96248bd7b548412811a4f1ffe88bed41f815/bot/cogs/off_topic_names.py#L59-L61

To handle it, we could simply include a `try-except` block and log the exception in the `except` block. I'm not sure if we want to log the entire exception, since the exception text could be a [massive HTML-response generated by cloudflare](https://paste.pythondiscord.com/ohibicedif). Logging the failure with the response code should generally give us enough to determine the cause of the failure.

Akarys42:
This allows to keep the task running even if the call fail. 

This PR closes #442


Akarys42:
I added a `continue` instead of a `return` to keep the code running

bot/cogs/off_topic_names.py
```diff
@@ -49,9 +49,13 @@ class OffTopicName(Converter):
...
         await asyncio.sleep(seconds_to_sleep)
 
-        channel_0_name, channel_1_name, channel_2_name = await bot.api_client.get(
-            'bot/off-topic-channel-names', params={'random_items': 3}
-        )
+        try:
+            channel_0_name, channel_1_name, channel_2_name = await bot.api_client.get(
+                'bot/off-topic-channel-names', params={'random_items': 3}
+            )
+        except bot.api.ResponseCodeError as e:
```
SebastiaanZ: This won't work, since you're now trying to access the `api` attribute of the `Bot` object with the name `bot`, not the `bot.api` module:

```py
During handling of the above exception, another exception occurred:

  File "<string>", line 9, in func
  File "/home/sebastiaan/pydis/repositories/djangobot/bot/cogs/off_topic_names.py", line 56, in update_names
    except bot.api.ResponseCodeError as e:
AttributeError: 'Bot' object has no attribute 'api'
```

One way to fix this is to import `ResponseCodeError` from `bot.api`:
```py
from bot.api import ResponseCodeError
```
and then using `ResponseCodeError` directly:
```suggestion
        except ResponseCodeError as e:
```

The exception will then be handled correctly:
```
Sep 24 11:27:52 pd.beardfist.com Bot: |          bot.cogs.off_topic_names |    ERROR | Failed to get new off topic channel names: code 404
``` 

### Note: For the suggested change to work, you do need to include that import at the top of the file as well.

Akarys42:
Import error fixed and waiting time reduced to half an hour if an exception occurred.

bot/cogs/off_topic_names.py
Akarys42: Sorry I didn't accepted your changes but the fix has been done because of github mobile stupid UI :)

PR #449, opened 2019-09-23
Fix failing duration conversion

### LINKED ISSUE #446: ExpirationDate Converter Not Correctly Parsing Duration Directives
`dateparser`'s date parsing is not correctly parsing our expiration duration directives per what's expected by http://strftime.org/:

image

From limited testing, `m` and `M` are both parsed as minutes and `y` is not correctly parsed as year. Shorter duration directives (e.g. `h`/`H` and `d`/`D`) are parsed as "expected" (but not case sensitive).

This behavior is consistent across dateparser `0.7.0` (2018-02-08), `0.7.1` (2019-02-12), and `0.7.2` (2019-09-17) so I don't think it's a new issue, just one that we might not have noticed.

SebastiaanZ:
The current ExpirationDate converter does not convert duration strings to `datetime.datetime` objects correctly. To remedy the problem, I've written a new Duration converter that uses regex matching to extract the relevant duration units and `dateutil.relativedelta.relativedelta` to compute a `datetime.datetime` that's the given duration in the future.

I've left the old `ExpirationDate` converter in place for now, since the new Duration converter may not be the most optimal method. However, given the importance of being able to convert durations for moderation tasks, I think it's better to implement the `Duration` converter now and rethink the approach later.

This PR closes #446

### DIFF
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -195,10 +195,16 @@ class Duration(Converter):
         """
         Converts a `duration` string to a datetime object that's `duration` in the future.
 
-        The converter supports years (symbols: `years`, `year, `Y`, `y`), months (`months`, `month`,
-        `m`), weeks (`weeks`, `week`, `W`, `w`), days (`days`, `day`, `D`, `d`), hours (`hours`,
-        `hour`, `H`, `h`), minutes (`minutes`, `minute`, `M`), and seconds (`seconds`, `second`,
-        `S`, `s`), The units must be provided in descending order of magnitude.
+        The converter supports the following symbols for each unit of time:
+        - years: `Y`, `y`, `year`, `years`
+        - months: `m`, `month`, `months`
+        - weeks: `w`, `W`, `week`, `weeks`
+        - days: `d`, `D`, `day`, `days`
+        - hours: `H`, `h`, `hour`, `hours`
+        - minutes: `M`, `minute`, `minutes`
+        - seconds: `S`, `s`, `second`, `seconds`
+
+        The units need to be provided in descending order of magnitude.
         """
         match = self.duration_parser.fullmatch(duration)
         if not match:
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from datetime import datetime
 from ssl import CertificateError
 from typing import Union
@@ -6,6 +7,7 @@ from typing import Union
 import dateparser
 import discord
 from aiohttp import ClientConnectorError
+from dateutil.relativedelta import relativedelta
 from discord.ext.commands import BadArgument, Context, Converter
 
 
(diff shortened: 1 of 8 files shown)


tests/test_converters.py
```diff
@@ -91,3 +92,76 @@ def test_valid_python_identifier_for_valid(value: str):
...
+
+        # Missing amount
+        ('ym'),
+
+        # Garbage
+        ('Guido van Rossum'),
+        ('lemon lemon lemon lemon lemon lemon lemon'),
+    )
+)
+def test_duration_converter_for_invalid(duration: str):
```
MarkKoz: Some tests for duplicate units would be good

tests/test_converters.py
```diff
@@ -91,3 +92,76 @@ def test_valid_python_identifier_for_valid(value: str):
...
+
+
+FIXED_UTC_NOW = datetime.datetime.fromisoformat('2019-01-01T00:00:00')
+
+
+@pytest.mark.parametrize(
+    ('duration', 'expected'),
+    (
+        # Simple duration strings
+        ('1Y', datetime.datetime.fromisoformat('2020-01-01T00:00:00')),
```
MarkKoz: Originally I was going to suggest changing the `datetime` import for the test module because this quite verbose. But, of course, `datetime` is a sin against PEP8 so it would possibly add confusion if it was imported another way. So I propose you use a [parametrised fixture](https://docs.pytest.org/en/latest/fixture.html#fixture-parametrize) for this, and the fixture takes care of calling `fromisoformat`. I _think_ a parameterised fixture would be the right way to do that.

SebastiaanZ: Wouldn't that mean that the parametrized fixture would have to yield both the duration string and the datetime object? When you use parametrized fixtures, your test will be run multiple times with all its parameters, once against each of the values yielded by the parametrized fixture.

Is this only because you don't like the verbosity?

tests/test_converters.py
SebastiaanZ: I'll add that.

bot/converters.py
```diff
@@ -197,3 +199,36 @@ class ExpirationDate(Converter):
...
+
+    async def convert(self, ctx: Context, duration: str) -> datetime:
+        """
+        Converts a `duration` string to a datetime object that's `duration` in the future.
+
+        The converter supports years (symbols: `years`, `year, `Y`, `y`), months (`months`, `month`,
+        `m`), weeks (`weeks`, `week`, `W`, `w`), days (`days`, `day`, `D`, `d`), hours (`hours`,
+        `hour`, `H`, `h`), minutes (`minutes`, `minute`, `M`), and seconds (`seconds`, `second`,
+        `S`, `s`), The units must be provided in descending order of magnitude.
+        """
```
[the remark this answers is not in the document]
SebastiaanZ: I've gone back and forth a couple of times before committing, but i think you're right. It just looks a bit like those parameter lists we don't want to have, but I guess we're dealing with something that's actually better displayed as a pseudo-table.

tests/test_converters.py
MarkKoz: Yeah, it would mean returning both. It's meant to replace the current parameter decorator.

> Is this only because you don't like the verbosity?
Pretty much. Just an idea and I don't feel strongly about it. You don't have to do it if you think it's for the worse.

SebastiaanZ: I've changed it to use parametrized fixtures.

MarkKoz: Looks good. You took it a step further and made it easier to add more tests (cause the expected string doesn't need to be calculated by you anymore).

Issue #401, opened 2019-09-02
!free command breaks if 2 or more channels have same timestamp


fiskenslakt:
When the last messages in 2 or more help channels have the same timestamp as each other, the `sorted` function will try and reconcile this tie by comparing the channel objects themselves. 

![Screenshot of error](https://i.imgur.com/LfjeqKU.png)

Relevant code:

https://github.com/python-discord/bot/blob/40f39877b522d8d19025f25b536cdc1afe1d2596/bot/cogs/free.py#L85

https://github.com/python-discord/bot/blob/40f39877b522d8d19025f25b536cdc1afe1d2596/bot/cogs/free.py#L63

The occurrence of this bug appears to be inconsistent when this particular edge case happens, which has made it difficult to pin down until now.

sco1:
Closed by #455

PR #455, opened 2019-09-25
Free Command Fix

LordHemlock:
- Fixed bug where if two channels had the same last message timestamp the command would error out.

Closes issue #401

### DIFF
--- a/bot/cogs/free.py
+++ b/bot/cogs/free.py
@@ -7,7 +7,6 @@ from discord.ext.commands import Bot, Cog, Context, command
 from bot.constants import Categories, Channels, Free, STAFF_ROLES
 from bot.decorators import redirect_output
 
-
 log = logging.getLogger(__name__)
 
 TIMEOUT = Free.activity_timeout
@@ -51,10 +50,10 @@ class Free(Cog):
             # the command was invoked in
             if channel.id == ctx.channel.id:
                 messages = await channel.history(limit=seek).flatten()
-                msg = messages[seek-1]
+                msg = messages[seek - 1]
             # Otherwise get last message
             else:
-                msg = await channel.history(limit=1).next()   # noqa (False positive)
+                msg = await channel.history(limit=1).next()  # noqa (False positive)
 
             inactive = (datetime.utcnow() - msg.created_at).seconds
             if inactive > TIMEOUT:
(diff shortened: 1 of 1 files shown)


bot/cogs/free.py
```diff
@@ -80,7 +79,8 @@ class Free(Cog):
             # Sort channels in descending order by seconds
             # Get position in list, inactivity, and channel object
             # For each channel, add to embed.description
-            for i, (inactive, channel) in enumerate(sorted(free_channels, reverse=True), 1):
+            for i, (inactive, channel) in enumerate(
+                    sorted(free_channels, key=lambda free_channel: free_channel[0], reverse=True), 1):
```
MarkKoz: This is getting a bit messy now with the key. At this point it'd be cleaner to assign the output of `sorted()` to a variable on a separate line and then use the variable inside `enumerate()`.

bot/cogs/free.py
```diff
@@ -80,7 +79,8 @@ class Free(Cog):
             # Sort channels in descending order by seconds
             # Get position in list, inactivity, and channel object
             # For each channel, add to embed.description
-            for i, (inactive, channel) in enumerate(sorted(free_channels, reverse=True), 1):
+            sorted_channels = sorted(free_channels, key=lambda free_channel: free_channel[0], reverse=True)
```
SebastiaanZ: Any reason for using a `lambda` instead of `operator.itemgetter(0)` here? If you `import itemgetter from operator`, we wouldn't need a lambda function: 

```suggestion
            sorted_channels = sorted(free_channels, key=itemgetter(0), reverse=True)
```

MarkKoz: Well, I don't care either way but one reason to use a lambda is because nothing needs to be imported.

LordHemlock: Is there a specific benefit to using the itemgetter vs not needing the import and using a lambda?  And honestly the reason I did the lambda was because I forgot about that itemgetter

SebastiaanZ: Personally, I think that itemgetter here is more idiomatic and lambdas shouldn't be used in these types of situations. That said, the channel lists are relatively small, so the performance benefit of itemgetter does not make a significant difference for us. I do think it's an improvement for the code style, though.

LordHemlock: That's fair.  Yeah for the sake of consistent style I'll swap it over

Issue #325, opened 2019-03-02
Repeating the reason in the confirmation message for infraction commands is redundant


MarkKoz:
When an infraction command is invoked, the bot replies with a message confirming success of the command (with the 👌). An infraction's reason is repeated in this message, but I find that to be redundant, especially when the confirmation is in the same channel as the one in which the command was sent. 

I bring this up because it stems from reasons being long. For short reasons, I think it's OK. The reason could either get truncated and have an ellipses appended, or it could just be entirely removed from the confirmation.

sco1:
I don't think I'd miss it if it went away.

heavysaturn:
This seems like a no-brainer, we should just implement it. but it'll have to wait until after the feature freeze.

PR #457, opened 2019-09-26
Merge infraction edit commands

### LINKED ISSUE #349: Merge `!infr edit reason` and `!infr edit duration` into `!infr edit`
Quite a few times, moderators go to edit an infraction and forget the `reason` or `duration` subcommands. While this is clearly PEBKAC, it's also an indication that if we can do without it, it's likely going to make the usage more intuitive, which would be nice.

After a thoughty, I've come up with maybe something like the following:
```py
async def infraction_edit(infraction_id, duration: Optional[duration], *, reason=None]):
```

This should allow us to do any of the following command types:

- Both `duration` and `reason`
  - `!infr edit 6b3a977a-c279-44e6-91e0-e53513958364 5d Extended because butt.`
- Just `reason`
  - `!infr edit 6b3a977a-c279-44e6-91e0-e53513958364 Disregard, wrong person.`
- Just `duration`
  - `!infr edit 6b3a977a-c279-44e6-91e0-e53513958364 5d`

MarkKoz:
Closes #349.

Also made the format of infraction timestamps consistent.

Some things to note:

* If the duration converter fails, the argument will instead be considered part of a reason. Therefore a typo like `!infraction edit 1337 50` (missing a unit for the duration) will make the reason `50` and expiration unchanged.
* The new reason is no longer displayed in the confirmation message. This is similar to #325.
* The two old commands are gone. Only `infraction edit` can be used.
* Expiration for all infractions can _still_ be edited, even ones for which it doesn't make sense (i.e. notes and warnings, which are meant to always be permanent).
* Editing the expiration for an expired infraction _still_ wont re-apply the infraction (it will reschedule it but the actual moderation action (mute or ban) will not reoccur).

### DIFF
--- a/bot/cogs/moderation.py
+++ b/bot/cogs/moderation.py
@@ -1282,8 +1282,8 @@ class Moderation(Scheduler, Cog):
 
     # endregion
 
-    @staticmethod
-    async def cog_command_error(ctx: Context, error: Exception) -> None:
+    # This cannot be static (must have a __func__ attribute).
+    async def cog_command_error(self, ctx: Context, error: Exception) -> None:
         """Send a notification to the invoking context on a Union failure."""
         if isinstance(error, BadUnionArgument):
             if User in error.converters:
--- a/bot/cogs/moderation.py
+++ b/bot/cogs/moderation.py
@@ -19,7 +19,7 @@ from bot.decorators import with_role
 from bot.pagination import LinePaginator
 from bot.utils.moderation import already_has_active_infraction, post_infraction
 from bot.utils.scheduling import Scheduler, create_task
-from bot.utils.time import wait_until
+from bot.utils.time import INFRACTION_FORMAT, format_infraction, wait_until
 
 log = logging.getLogger(__name__)
 
@@ -313,11 +313,7 @@ class Moderation(Scheduler, Cog):
             reason=reason
         )
 
-        infraction_expiration = (
-            datetime
-            .fromisoformat(infraction["expires_at"][:-1])
-            .strftime('%c')
-        )
+        infraction_expiration = format_infraction(infraction["expires_at"])
 
         self.schedule_task(ctx.bot.loop, infraction["id"], infraction)
 
@@ -390,11 +386,7 @@ class Moderation(Scheduler, Cog):
         except Forbidden:
             action_result = False
 
-        infraction_expiration = (
-            datetime
-            .fromisoformat(infraction["expires_at"][:-1])
-            .strftime('%c')
-        )
+        infraction_expiration = format_infraction(infraction["expires_at"])
 
         self.schedule_task(ctx.bot.loop, infraction["id"], infraction)
 
@@ -630,11 +622,7 @@ class Moderation(Scheduler, Cog):
         self.mod_log.ignore(Event.member_update, user.id)
         await user.add_roles(self._muted_role, reason=reason)
 
-        infraction_expiration = (
-            datetime
-            .fromisoformat(infraction["expires_at"][:-1])
-            .strftime('%c')
-        )
+        infraction_expiration = format_infraction(infraction["expires_at"])
 
         self.schedule_task(ctx.bot.loop, infraction["id"], infraction)
 
@@ -694,11 +682,7 @@ class Moderation(Scheduler, Cog):
         except Forbidden:
             action_result = False
 
-        infraction_expiration = (
-            datetime
-            .fromisoformat(infraction["expires_at"][:-1])
-            .strftime('%c')
-        )
+        infraction_expiration = format_infraction(infraction["expires_at"])
 
         self.schedule_task(ctx.bot.loop, infraction["id"], infraction)
 
@@ -915,7 +899,7 @@ class Moderation(Scheduler, Cog):
             confirm_messages.append("marked as permanent")
         elif expires_at is not None:
             request_data['expires_at'] = expires_at.isoformat()
-            confirm_messages.append(f"set to expire on {expires_at.strftime('%c')}")
+            confirm_messages.append(f"set to expire on {expires_at.strftime(INFRACTION_FORMAT)}")
         else:
             confirm_messages.append("expiry unchanged")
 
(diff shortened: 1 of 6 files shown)


SebastiaanZ:
> Some things to note:
> 
>     * If the duration converter fails, the argument will instead be considered part of a reason. Therefore a typo like `!infraction edit 1337 50` (missing a unit for the duration) will make the reason `50` and expiration unchanged.
> 
>     * The new reason is no longer displayed in the confirmation message. This is similar to #38 

Your solution looks smart.

I'm debating with myself whether or not I should even add, so take this suggestion with a pinch of salt. Your current code does not give feedback when the duration was not changed. If we look at this from a human factors engineering perspective, this means we also don't get feedback about not having changed the duration when we most need it: When we tried to change duration, but made a mistake in the duration string. 

Obviously, the absence of feedback could be seen as feedback itself, but with the above perspective in mind, we now have to rely on human memory for the actor to spot that something that should have been there is not. Since we don't edit the duration of an infraction often, this increases the risk of the actor missing the fact that the duration feedback is absent.

A solution would be to always indicate if the duration field was touched, change or not. This would introduce slightly more noise in the confirmation message, but since we infrequently alter infractions and it would be only a couple of words, I'd say that's not a big problem.

That said, we're not designing a nuclear power plant or oil refinery, so scrutiny up to that point is probably over the top. Feel free to ignore it.

>     * Expiration for all infractions can _still_ be edited, even ones for which it doesn't make sense (i.e. notes and warnings, which are meant to always be permanent).

We could think of adding more validators to the serializer in the back-end. We already have a couple of them and I already proposed two other ones in python-discord/site#260.

https://github.com/python-discord/site/blob/37d989d0b74a1d0f8b0ee60749c79ea4a55ffba5/pydis_site/apps/api/serializers.py#L109-L121

bot/cogs/moderation.py
```diff
@@ -875,121 +884,68 @@ def mod_log(self) -> ModLog:
...
+        request_data = {}
+        confirm_messages = []
+        log_text = ""
+
+        if expires_at == "permanent":
+            request_data['expires_at'] = None
+            confirm_messages.append("marked as permanent")
+        elif expires_at is not None:
+            request_data['expires_at'] = expires_at.isoformat()
+            confirm_messages.append(f"set to expire on {expires_at.strftime('%c')}")
```
SebastiaanZ: In a slightly, let's call it "interesting", vote and debate with the moderation team, we decided that we want to display infraction datetime strings in an ISO-like string (`"%Y-%m-%d %H:%M"`), like here:

https://github.com/python-discord/bot/blob/1685d676008d3092fa1e2377769613a5063e10f8/bot/cogs/moderation.py#L1132

This has not been consistently changed throughout the cog, yet, and many occurrences of `%c` still exist. Nevertheless, I think we should strive for consistency and use the newly chosen format here as well.

```suggestion
            confirm_messages.append(f"set to expire on {expires_at.strftime('"%Y-%m-%d %H:%M"')}")
```

MarkKoz:
> Your current code does not give feedback when the duration was not changed. 

Not explicitly. But if the "confirmation message" that is sent afterwards does not mention the duration but only the reason, then one can assume the duration did not change. Just want to be clear: are you saying that is not adequate? If that's the case, then sure I can agree to this:

> A solution would be to always indicate if the duration field was touched, change or not.

scragly:
Yeah he's saying that in this section:
> the absence of feedback could be seen as feedback itself, but with the above perspective in mind, we now have to rely on human memory for the actor to spot that something that should have been there is not. 

A little convoluted to read, but the meaning is there haha.

I agree explicit feedback is best for now.

bot/cogs/moderation.py
```diff
@@ -875,121 +884,68 @@ def mod_log(self) -> ModLog:
...
-    @infraction_edit_group.command(name="duration")
-    async def edit_duration(
-            self, ctx: Context,
-            infraction_id: int, expires_at: Union[Duration, str]
+    @infraction_group.command(name='edit')
+    async def infraction_edit(
+        self,
+        ctx: Context,
+        infraction_id: int,
+        expires_at: Union[Duration, permanent_duration, None],
```
SebastiaanZ: Too bad we can't really mix "just" comments and requested changes, so I've added this as a separate comment.

Yesterday (for me) we briefly talked about being able to set exact dates instead of a relative duration for the expiration field. I've checked and I've not seen it used, so I'm not married to it. If we do want it, I may be better to have a separate converter that converts datetime strings of a certain format to a datetime instead of creating a hybrid one that both parses relative durations and absolute datetime strings.

I currently see not no real use for it, though, although it could be useful if we ever want to do something with absolute dates (like adding events to the event calendar we've envisioned.)

scragly: Exact dates could be handy for underage bans.

MarkKoz: > may be better to have a separate converter that converts datetime strings of a certain format to a datetime instead of creating a hybrid one that both parses relative durations and absolute datetime strings.

I agree. I came to that conclusion too as I started working on this. I'm not sure if this would be good to implement in this PR or in another.

bot/cogs/moderation.py
MarkKoz: I could just knock these all out in this PR cause it's only for moderation stuff, right?

MarkKoz: Done, for the most part. All infraction time formats should now be consistent. The only exception is for watch channels, which has a similar format but with seconds included too. Was not sure if that should be touched.

MarkKoz:
I've tested my changes and all is good. 👍

bot/cogs/moderation.py
SebastiaanZ: It can be replaced, as far as I'm concerned. There's a utility function in the `WatchChannel` ABC that's used by the talentpool cog for a display that's very similar to the infraction search output. Originally, I had the idea of promoting that local utility function into a general one, but you've already taken care of that now.

The only thing left is testing, but you've probably thought of that as well. I guess we need to wait for #421 for that.

SebastiaanZ: I've replaced it by the new utility function.

PR #460, opened 2019-09-27
Fix help command invocation

### LINKED ISSUE #456: Help command raises AttributeError when argument is a cog name
For example, `!help Moderation` returns

```
AttributeError: 'Command' object has no attribute 'instance'
```

Have not looked into the cause of this yet.

sco1:
Per #456, the `!help` raises an `AttributeError` when fetching information for Cogs due to referencing of an attribute that is no longer present in d.py. The no longer present `instance` attribute has been replaced with `cog_name` .

This also kaizens in a fix for some incorrect kwargs in one of the watchchannel cog's mod log invocations.

Closes #456

### DIFF
--- a/bot/cogs/help.py
+++ b/bot/cogs/help.py
@@ -110,9 +110,9 @@ class HelpSession:
         cog = self._bot.cogs.get(query)
         if cog:
             return Cog(
-                name=cog.cog_name,
+                name=cog.qualified_name,
                 description=inspect.getdoc(cog),
-                commands=[c for c in self._bot.commands if c.cog_name == cog.cog_name]
+                commands=[c for c in self._bot.commands if c.cog is cog]
             )
 
         self._handle_not_found(query)
--- a/bot/cogs/help.py
+++ b/bot/cogs/help.py
@@ -110,9 +110,9 @@ class HelpSession:
         cog = self._bot.cogs.get(query)
         if cog:
             return Cog(
-                name=cog.__class__.__name__,
+                name=cog.cog_name,
                 description=inspect.getdoc(cog),
-                commands=[c for c in self._bot.commands if c.instance is cog]
+                commands=[c for c in self._bot.commands if c.cog_name == cog.cog_name]
             )
 
         self._handle_not_found(query)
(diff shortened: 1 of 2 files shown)


bot/cogs/help.py
```diff
@@ -110,9 +110,9 @@ def _get_query(self, query: str) -> Union[Command, Cog]:
         cog = self._bot.cogs.get(query)
         if cog:
             return Cog(
-                name=cog.__class__.__name__,
+                name=cog.cog_name,
```
MarkKoz: `cog_name` the wrong attribute for Cogs (but not for Commands) and will raise an error
```suggestion
                name=cog.qualified_name,
```

bot/cogs/help.py
```diff
@@ -110,9 +110,9 @@ def _get_query(self, query: str) -> Union[Command, Cog]:
...
                 description=inspect.getdoc(cog),
-                commands=[c for c in self._bot.commands if c.instance is cog]
+                commands=[c for c in self._bot.commands if c.cog_name == cog.cog_name]
```
MarkKoz:
```suggestion
                commands=[c for c in self._bot.commands if c.cog_name == cog.qualified_name]
```

MarkKoz: More akin to what was there before

```suggestion
                commands=[c for c in self._bot.commands if c.cog is cog]
```

sco1: Which of these two suggestions are intended?

bot/cogs/help.py
sco1: Under what circumstance does this raise an exception?

MarkKoz: It's an AttributeError. So almost always whenever a cog name is given to the help command (in my testing `Moderation` still worked for some reason).

bot/cogs/help.py
MarkKoz: Either one. Sorry, I'll add some more context. The latter one would be nicer except that the `cog` attribute is undocumented. Supposedly it would still be considered a "breaking change" were it to be renamed or removed by discord.py.

bot/cogs/help.py
sco1: Weird, worked fine when I was looking:

image

PR #462, opened 2019-09-27
Moderation cog cleanup

MarkKoz:
**This is not extensively tested**. I am hoping for others (i.e. the reviewers) to help out in this regard considering the importance of these cogs.

This aims to reduce redundancy and code density in the moderation cog.

The moderation cog was split in two: `Infractions` in `infractions.py` and `ModManagement` in `management.py` (contains infraction search and edit commands). They were then moved to a `moderation` sub-package along with `ModLog` and `Superstarify`. The sub-package now acts as the extension for all four of these cogs. `utils/moderation.py` was also moved inside the sub-package as `moderation.utils.py`.

To facilitate this move, custom category support was added to the help command. This allows the `Infractions` and `ModManagement` cogs to appear under the _Moderation_ category in the help command's output.

### Behaviour

Just to define some terms:

* **confirmation message** - A message sent to the invocation context of the command. It shows if the command was successful.
* **appyling infractions** - Doing a ban, a mute, a kick, or a warning and any shadow/temporary variations.
* **pardoning infractions** - Doing an unmute or unban.

#### Error Handling

* When applying infractions, most errors are deferred to the regular error handler
* The `Forbidden` exception is currently the only caught exception when applying infractions
* When pardoning infractions, the errors will be displayed in the confirmation message and the mod log
    * Displaying them in the mod log allows for errors which occur during automatic expiration to also be seen easily in Discord
* If multiple active infractions exist when pardoning and an error occurs when trying to remove them, the error will simply say to check the bot's logs (i.e. stdout in the console)
* If an automatic expiration fails, the moderator role will be mentioned in the mod log. Failures for not finding the user or the ban on Discord are excluded as the intention of the mention is to prevent infractions from being stuck.

#### Confirmation Messages

* A message is sent to the invoking context specifying if the action failed or suceeded
* For applying infractions, the failure message will be generic
* When pardoning, the error messages will be appended to the confirmation message
* Expiration date is shown in the confirmation
    * If the infraction is permanent, "permnanetly" is shown instead
    * No expiration is shown if the infraction fails to apply

#### DM Notifications

* A mail emoji is shown in the confirmation message when DM is sent
* The mod log shows a DM status (sent/failed)
* The actor is mentioned in the mod log when a DM fails to send
* If it's a shadow infraction, no emoji is shown and the DM status is absent from the mod log

#### Mod Log

* A mod log is always sent for applying and pardoning infractions, even when done automatically for expired infractions
* The events are always ignored in the mod log to prevent the default messages from showing
* The title will show the infraction type and whether it failed or succeeded
* Manual pardons and automatic expirations are distinguished in the title
* Automatic expirations will show the bot as the actor

#### Miscellaneous

* Pardoning/expiring infractions will now have reasons in the audit log which mention the infraction ID
* When pardoning/expiring an infraction, if one step fails, the others will still be executed. This means that if, for example, the mute role fails to be removed, the PATCH request will still be sent to the API and the expiration task will still be cancelled.

------------

### Todo

- [x] Find a nice way to generalise pardoning infractions (`unmute` and `unban`)
- [x] Move extensions to a `moderation` subpackage.
    - [x] `Moderation` cog will be in `infractions.py` and `Infractions` cog will be in `management.py`
    - [x ] Create a new `group` attribute for cogs so that the `help` command shows all grouped modules under a single name
        - [x] ~Consider how this affects the `reload` command. Should they even be separate extensions or part of a single `moderation` extension?~ Part of a single moderation extension.
    - [x] Possibly move other extensions here (~`watchchannels`~, `superstarify`, and `modlog`)
    - [x] Move `utils/moderation.py` to `cogs/moderation/utils.py`
        * Should only contain utility functions that are used by several modules or that could be in the future. Module-specific utilities should remain in their corresponding modules.
* ~Write unit tests?~ May be better to save for a future PR.

### Manual tests for `apply_infraction()`

- [x] DM not sending
    - [x] Confirmation message doesn't have mail emoji
    - [x] Actor is mentioned in mod log
    - [x] DM status shows as failed in mod log
- [x] DM sending
    - [x] Confirmation message has mail emoji
    - [x] Actor is not mentioned in mod log
    - [x] DM status shows as sent in mod log
    - [x] User recieves DM
        - [x] Expiry correctly formatted
        - [x] Icon correctly corresponds to the infraction type
- [x] Shadow infractions
    - [x] No DM status in mod log
    - [x] User does not recieve a DM
- [x] Mod log
    - [x] Mod log is sent
    - [x] Title appropriately indicates when the infraction failed to apply
    - [x] Icon correctly corresponds to the infraction type
    - [x] Mentions the actor when applying the infraction fails
- [x] Confirmation message
    - [x] Success
        - [x] Shows OK hand emoji and a message of affirmation
        - [x] Mentions the user
        - [x] Shows "permanently" for expiry if duration unspecified
    - [x] Failure
        - [x] Shows X emoji and an appropriate message
        - [x] Does not show expiry
- [x] Test temporary infractions
    - [x] Expiry is shown in mod log w/ correct format
    - [x] Infraction is scheduled and does indeed expire
    - [x] Expiry is shown in confirmation message w/ correct format


MarkKoz:
The infraction search and edit commands should be separated from the moderation cog and put into a new cog. Should both the moderation and the new cog be moved into a moderation-related subpackage (moglog could go in there too)? If so, what would the package be named and what would the modules (yes, I mean the modules, _not_ the cog classes) be named? Would the contents of `utils/moderation.py` be moved to the subpackages `__init__`?

scragly:
I can understand the logic behind splitting the moderation cog up to make it more manageable in the files, but having to put infr search and edit into an entirely different command category (user-facing) doesn't make much sense to me. Infraction management fits perfectly under the category of "Moderation", so I'm not sure I'm a fan.

MarkKoz:
OK, that's fine. I don't really care for splitting the cogs. Really I just want to split the code - the modules. It could be implemented in such a way that it's one cog but two modules.

scragly:
This was discussed further in-server and it was decided splitting the cogs is fine under the condition that we support setting a category for a cog explicitly for help output, which should be a simple thing to support.

bot/cogs/moderation/utils.py
```diff
@@ -0,0 +1,170 @@
...
+    "mute": (Icons.user_mute, Icons.user_unmute),
+    "kick": (Icons.sign_out, None),
+    "ban": (Icons.user_ban, Icons.user_unban),
+    "warning": (Icons.user_warn, None),
+    "note": (Icons.user_warn, None),
+}
+RULES_URL = "https://pythondiscord.com/pages/rules"
+APPEALABLE_INFRACTIONS = ("ban", "mute")
+
+UserTypes = t.Union[discord.Member, discord.User]
```
scragly:
```suggestion
UserTypes = typing.Union[discord.Member, discord.User]
```

bot/cogs/moderation/utils.py
```diff
@@ -0,0 +1,170 @@
...
+MemberObject = t.Union[UserTypes, discord.Object]
+Infraction = t.Dict[str, t.Union[str, int, bool]]
```
scragly:
```suggestion
Infraction = typing.Dict[str, typing.Union[str, int, bool]]
```

bot/cogs/moderation/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+from discord.ext.commands import Bot
+
+from . import utils
+from .infractions import Infractions
+from .management import ModManagement
+from .modlog import ModLog
+from .superstarify import Superstarify
+
+__all__ = ("utils", "Infractions", "ModManagement", "ModLog", "Superstarify")
```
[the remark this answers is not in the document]
MarkKoz: No, it isn't.

bot/cogs/moderation/infractions.py
```diff
@@ -0,0 +1,491 @@
+import logging
+import textwrap
+import typing as t
```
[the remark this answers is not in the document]
MarkKoz: The idea was that future changes may lead to doing this anyway.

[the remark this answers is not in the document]
scragly: That's fine I guess, it's typically best to hold out until that time though usually. It's not a problem to keep it.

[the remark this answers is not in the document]
MarkKoz: You're right, I was overzealous. But the "work" is already done.

scragly:
Great job, thanks!

bot/cogs/help.py
```diff
@@ -107,12 +106,26 @@ def _get_query(self, query: str) -> Union[Command, Cog]:
...
+        if cog_matches:
+            cog = cog_matches[0]
+            cmds = (cog.get_commands() for cog in cog_matches)  # Commands of all cogs
+
             return Cog(
-                name=cog.qualified_name,
-                description=inspect.getdoc(cog),
-                commands=[c for c in self._bot.commands if c.cog is cog]
+                name=cog.category if hasattr(cog, "category") else cog.qualified_name,
+                description=cog.description,
```
MarkKoz: This uses the docstringoif the first cog in the list of matches which isn't great but it is simple. What would be a good way to define a docstring for a specific category?

scragly: That reminds me actually, I was going to suggest also to move the category into scope of the class instead, as it's not instance specific.

One approach regarding the description is the instead find the first instance of another attribute `category_description` across the cogs, so you could define it once anywhere and the rest won't need to. 

```py
description=cog.category_description if hasattr(cog, "category_description") else cog.description
```

bot/cogs/moderation/infractions.py
```diff
@@ -0,0 +1,519 @@
...
+        for infraction in infractions:
+            if infraction["expires_at"] is not None:
+                self.schedule_task(self.bot.loop, infraction["id"], infraction)
+
+    # region: Permanent infractions
+
+    @command()
+    async def warn(self, ctx: Context, user: MemberConverter, *, reason: str = None) -> None:
+        """Warn a user for the given reason."""
+        infraction = await utils.post_infraction(ctx, user, reason, "warning")
```
SebastiaanZ: The reason and the warning type were switched here
```suggestion
        infraction = await utils.post_infraction(ctx, user,  "warning", reason)
```

`!warn @David Bowie Kitties!` led to:
```py
 Oct 02 21:22:29 Bot: |         bot.cogs.moderation.utils |    ERROR | An unexpected ResponseCodeError occurred while adding an infraction:
  File "/home/sebastiaan/pydis/repositories/djangobot/bot/cogs/moderation/utils.py", line 71, in post_infraction
    response = await ctx.bot.api_client.post('bot/infractions', json=payload)
  File "/home/sebastiaan/pydis/repositories/djangobot/bot/api.py", line 76, in post
    await self.maybe_raise_for_status(resp, raise_for_status)
  File "/home/sebastiaan/pydis/repositories/djangobot/bot/api.py", line 56, in maybe_raise_for_status
    raise ResponseCodeError(response=response, response_json=response_json)
bot.api.ResponseCodeError: Status: 400 Response: {'type': ['"Kitties!" is not a valid choice.']}
```

bot/cogs/moderation/infractions.py
```diff
@@ -0,0 +1,519 @@
...
+        """
+        await self.apply_ban(ctx, user, reason, expires_at=duration)
+
+    # endregion
+    # region: Permanent shadow infractions
+
+    @command(hidden=True)
+    async def note(self, ctx: Context, user: MemberConverter, *, reason: str = None) -> None:
+        """Create a private note for a user with the given reason without notifying the user."""
+        infraction = await utils.post_infraction(ctx, user, reason, "note", hidden=True)
```
SebastiaanZ: Likewise:
```suggestion
        infraction = await utils.post_infraction(ctx, user, "note", reason, hidden=True)
```

bot/cogs/help.py
MarkKoz: Here's a completely different idea:

Keep a dictionary which maps categories to cogs (and also stores the description). This would be an attribute of `bot`, similar to how d.py already has a cogs dictionary. Thus, categories would be defined in extensions (well it could be anywhere outside the cog class really). The current solution just seems too hacky.

MarkKoz:
I'm working on resolving the conflict caused by #486. I want to just replace all the redundant infraction deactivation code it brings with a call to `self.deactivate_infraction()`. However, using this method has two problems, both caused by the role failing to be removed (which could be solved by re-adding the role before calling the method but that seems redundant):

1. The mod log will show as "failed". There is already an parameter to disable sending the mod log. However, some mod log probably should be sent and writing separate code for that is easy but just feels redundant and thus not ideal.
2. Because the DM is sent after the role is removed, the role removal failing means there won't be an attempt to send a DM. Generally, I think this is desirable behaviour. However, this specific case would be an exception.

Does anyone have ideas for clean and non-redundant way to handle deactivating the infraction?

bot/cogs/moderation/infractions.py
```diff
@@ -0,0 +1,525 @@
...
+    @commands.Cog.listener()
+    async def on_ready(self) -> None:
+        """Schedule expiration for previous infractions."""
+        infractions = await self.bot.api_client.get(
+            'bot/infractions',
+            params={'active': 'true'}
+        )
+        for infraction in infractions:
+            if infraction["expires_at"] is not None:
+                self.schedule_task(self.bot.loop, infraction["id"], infraction)
```
SebastiaanZ: @scragly made a good point yesterday in #dev-core and I think it needs to be addressed. Because we're using the `on_ready` event to reschedule the expiration of all active infractions with an expiration date and the `on_ready` event does not fire when you reload the cog, reloading the `moderation` extension will erase all of the scheduled tasks:



This is probably a problem that has always existed and has always gone unnoticed, but now is good time to fix it, I think. The solution to this would be to move this rescheduling logic to a regular coroutine that's added to the event loop as a Task at the end of the Infraction cog's `__init__`. I think we only need to add an `await bot.wait_until_ready()` as the first line, but I don't think the rest of the code needs to be changed.

SebastiaanZ:
> I'm working on resolving the conflict caused by #486. I want to just replace all the redundant infraction deactivation code it brings with a call to `self.deactivate_infraction()`. However, using this method has two problems, both caused by the role failing to be removed (which could be solved by re-adding the role before calling the method but that seems redundant):
> 
>     1. The mod log will show as "failed". There is already an parameter to disable sending the mod log. However, some mod log probably should be sent and writing separate code for that is easy but just feels redundant and thus not ideal.
> 
>     2. Because the DM is sent after the role is removed, the role removal failing means there won't be an attempt to send a DM. Generally, I think this is desirable behaviour. However, this specific case would be an exception.
> 
> 
> Does anyone have ideas for clean and non-redundant way to handle deactivating the infraction?

One option is to split this method that does two things, removing the role from the member and requesting the API to mark it as inactive, into two methods that do one thing, optionally having the `remove_role` method call the `mark_inactive` (just filler names; I'd say you're activity is just fine), if we think that's a combination we always want to have. Another option would be to add a parameter for `remove_role` and an `if` check, but that feels messier to me and achieves basically the same (but with a greater cyclomatic complexity).

We would get the most flexibility if we were to add a third, wrapper, method and don't have the `remove_role` call the `mark_inactive`, but I honestly don't see a case in which we want to attempt to remove a role/ban, but not make sure it's also marked as inactive in the database.

MarkKoz:
@SebastiaanZ 
> One option is to split this method that does two things, removing the role from the member and requesting the API to mark an entry in the database as inactive, into two methods that do one thing, optionally having the `remove_role` method call the `mark_inactive` (just filler names; I'd say you're activity is just fine) if we think that's a combination we always want to have. 

If the mute was split, it'd be something like this:
```py
async def pardon_mute(self, ctx: Context, user: Member):
    self.mod_log.ignore(Event.member_update, user.id)
    await user.remove_roles(self._muted_role, reason=reason)

    # DM the user about the expiration.
    notified = await utils.notify_pardon(
        user=user,
        title="You have been unmuted.",
        content="You may now send messages in the server.",
        icon_url=utils.INFRACTION_ICONS["mute"][1]
    )

    return notified
```

However, splitting the function won't solve the issue of the DM not being sent on failure. As I said, normally we'd want the DM to not send if removing the role failed. Therefore, in this case, a try-except would need to be around the `remove_roles` call. How would it know when to ignore the exception and when to propagate it? Well, it'd need a parameter for that or a separate function definition which always exhibits such behaviour (this is redundant and not ideal). 

> Another option would be to add a parameter for `remove_role` and an `if` check, but that feels messier to me and achieves basically the same (but with a greater cyclomatic complexity).

I would prefer to not have such a parameter but it seems like it may be the better option given how convoluted and inter-connected this function already is.

SebastiaanZ:
I don't think I understand what you mean. From the fix you were referencing, I was under the impression we were talking about calling `deactivate_infraction` in the `on_member_join` event after finding an active mute infraction for the user in the database that has since expired.

If so, that means the user does not have the mute role, since they just joined the server. By splitting up the `deactivate_infraction` method into two methods, `remove_role` (related to the user's state on Discord) and `mark_inactive` (related to the infraction state in the database), we could then call `mark_inactive` and skip the `remove_role` function entirely. That way, we don't have to worry about a role and/or a DM, since they are not relevant for people (re)joining the server without any infractions that should be active.

MarkKoz:
So you don't think in that situation a user should be DM'd about being unmuted?

SebastiaanZ:
I don't really think it's relevant. The mute expired while the user was a member, so they don't get a DM of that.

MarkKoz:
I think you mean _was not a member_, but that make sense so I agree. Now I am wondering if we even want to show the expiration as _failed_ in the mod log when the user already doesn't have the muted role or they don't have an active ban. Thinking about it now, it's actually technically still a success in the sense that they indeed are no longer affected by the infraction, which was the goal of the function.

MarkKoz:
I'm just going to go ahead and do that. It makes sense.

MarkKoz:
Turns out `remove_roles` does not fail if the user already doesn't have the role...

bot/cogs/moderation/infractions.py
```diff
@@ -0,0 +1,563 @@
...
+                        json={"active": False}
+                    )
+                except ResponseCodeError:
+                    log.exception(f"Failed to deactivate infraction #{_id} ({infr_type})")
+                    # This is simpler and cleaner than trying to concatenate all the errors.
+                    log_text["Failure"] = "See bot's logs for details."
+
+                # Cancel pending expiration task.
+                if infraction["expires_at"] is not None:
+                    self.cancel_task(infraction["id"])
```
MarkKoz: This code handles multiple active infractions. Does it still need to be here for the time being or can it already be removed given what you are working on? I'm not caught up on the status of that. @SebastiaanZ

bot/cogs/help.py
MarkKoz: @scragly Thoughts on that? Also, what was the other thing you were talking about on Discord regarding the help command when you first had a look at the diff?

bot/cogs/moderation/infractions.py
MarkKoz: Is efd8a30 what you had in mind? Tried it out and it solves the issue. I looked at watchchannels and that assigns the task to a private attribute which is never referenced anywhere. Is that necessary to do here too?

bot/cogs/moderation/infractions.py
SebastiaanZ: Double active infractions may still occur for now. The code for the PR is done, the migrations file as well, but it will be stalled until python-discord/site#269 is resolved, since that PR also adds a migration file (and is actually far more critical). (We can't merge them independently because migration files have a single, explicit parent file mentioned in their body.)

I was debating with myself not to include the data migration and open a PR later, to prevent the conflict and not stall the validation, but that would still leave users with multiple active infractions in the database (17 users have double bans; no users have double mutes), it only guarantees we won't add more using the API.

MarkKoz: Thanks. Maybe everything will get merged in time but that seems unlikely. I'll leave this unresolved so I don't forget.

bot/cogs/help.py
MarkKoz: Discussion on Discord alleviated my concerns about this being "hacky", especially if the attribute is changed from an instance attribute to a class attribute. Additionally, the support for categories will be documented in the `HelpSession` class.

Regarding the description, I agree making it an attribute of the cog is a good idea.

If categories are going to be used more frequently in the code base, a more "structured" and documented approach could be considered. For now, with only one category, this is adequate.

bot/cogs/moderation/superstarify.py
```diff
@@ -1,34 +1,32 @@
...
 from bot.converters import Duration
-from bot.decorators import with_role
-from bot.utils.moderation import post_infraction
+from bot.utils.checks import with_role_check
 from bot.utils.time import format_infraction
+from . import utils
+from .modlog import ModLog
 
 log = logging.getLogger(__name__)
 NICKNAME_POLICY_URL = "https://pythondiscord.com/pages/rules/#wiki-toc-nickname-policy"
```
SebastiaanZ: This URL has changed in the mean time by the removal of the `wiki-toc-` prefix:

```suggestion
NICKNAME_POLICY_URL = "https://pythondiscord.com/pages/rules/#nickname-policy"
```

PR #466, opened 2019-09-30
New reminder features

### LINKED ISSUE #241: Reminder details in confirmation message.
The reminder command currently only replies with `Your reminder has been created successfully!`. 

This can be improved by providing the actual reminder contents in the confirmation and the datetime / humanized timedelta the reminder will be due to ensure the intended duration was processed. 

For the message sent once the reminder is due, a jump url to the original command message so we can jump back into context of the conversation when the reminder was made would be good to add. (Previously had mixed up wording for this section).

This would do well as a first issue if it's not gotten to in short time.

Akarys42:
Hi, this PR add : 
* reminder name and timedelta in the confirmation message
* jump url to the message where the reminder was set in the deliver message

~~I marked this PR as draft because I don’t have my test environment setup yet so theses changes aren’t tested for now.~~ Now tested 

This PR closes #241.


Akarys42:
Now tested and ready to go ! The only problem is that the `jump_url` is never added to the database, gdude said it is apparently because there is no `jump_url` in the model.

bot/cogs/reminders.py
```diff
@@ -99,10 +99,13 @@ def __init__(self, bot: Bot):
...
         embed.set_author(
             icon_url=Icons.remind_blurple,
-            name="It has arrived!"
-        )
+            name="It has arrived!")
 
-        embed.description = f"Here's your reminder: `{reminder['content']}`"
+        if "jump_url" in reminder:  # keep backward compatibility
+            embed.description = (f"Here's your reminder: `{reminder['content']}`.\n"
+                                 f"Jump back when you created the reminder : {reminder['jump_url']}")
```
SebastiaanZ:
```suggestion
                                 f"Jump back to when you created the reminder : {reminder['jump_url']}")
```

SebastiaanZ: Why don't you turn the text into a clickable URL instead of creating a longer sentence by appending the message URL as text? The embed content allows for the use of `[text](url)` markdown and I think it would look much cleaner that way.

bot/cogs/reminders.py
```diff
@@ -165,14 +168,18 @@ def __init__(self, bot: Bot):
             json={
                 'author': ctx.author.id,
                 'channel_id': ctx.message.channel.id,
+                'jump_url': ctx.message.jump_url,
```
SebastiaanZ: This won't work without making sure the API accepts a `jump_url` field by including it in the model. I've added the `Stalled` label until there's support for this in `site`, since otherwise the implementation of python-discord/site#258 will break our reminders feature.

bot/cogs/reminders.py
```diff
@@ -165,14 +168,18 @@ def __init__(self, bot: Bot):
...
             }
         )
 
+        now = datetime.utcnow()
+
         # Confirm to the user that it worked.
         await self._send_confirmation(
-            ctx, on_success="Your reminder has been created successfully!"
+            ctx,
+            on_success=f"Your reminder `{content}` will arrive in {humanize_delta(relativedelta(expiration, now))}!"
```
SebastiaanZ: I know that adding the contents of the reminder to the confirmation message was mentioned in the issue, but I'm not sure that I like it. We've actively started to remove these type of "confirmation echoes" because the message that sets the reminder will have the exact same contents as what will be echoed here.

I think the only necessary addition is the indication of the date.

bot/cogs/reminders.py
Akarys42: Only admin can add a `jump_url` in the model I suppose?

SebastiaanZ: No, anyone can make a PR to the `site` repository that adds such a field in the [`reminder` model](https://github.com/python-discord/site/blob/master/pydis_site/apps/api/models/bot/reminder.py). The `Meta` class of the reminder serializer probably also needs to be adjusted to include the new field.

Akarys42: PR opened : python-discord/site#277. Wdym by meta class?

Akarys42:
I opened python-discord/site#277 to had the `jump_url` field to the reminder model.

lemonsaurus:
https://github.com/python-discord/site/pull/277 was closed. So this is blocked until that change is made.

@Akarys42 did you at least open an issue on the site repo so that someone else can make this change if you don't intend to make it yourself?

Akarys42:
I didn’t opened a issue because I want to do it myself, I’m with my family this weekend, so I will probably do it next week is it okay for you.

bot/cogs/reminders.py
Akarys42: Here it is: python-discord/site#288

Akarys42:
This PR add the `jump_url` field to the reminder model : python-discord/site#288. __This PR need to be merged first.__

bot/cogs/reminders.py
```diff
@@ -101,10 +101,13 @@ def __init__(self, bot: Bot):
         embed.colour = Colour.blurple()
         embed.set_author(
             icon_url=Icons.remind_blurple,
-            name="It has arrived!"
-        )
+            name="It has arrived!")
```
jchristgit: Why this change?

Akarys42: I reverted a change here and I probably forgot to re-add the line break.

lemonsaurus:
This doesn't pass lint. Please address.

bot/cogs/reminders.py
```diff
@@ -104,7 +104,12 @@ def __init__(self, bot: Bot):
             name="It has arrived!"
         )
 
-        embed.description = f"Here's your reminder: `{reminder['content']}`"
+        if "jump_url" in reminder and reminder["jump_url"] != "invalid":  # keep backward compatibility
+            embed.description = (f"Here's your reminder: `{reminder['content']}`.\n"
+                                 f"[Jump back to when you created the reminder]({reminder['jump_url']})"
+            )
+        else:
+            embed.description = f"Here's your reminder: `{reminder['content']}`"
```
SebastiaanZ: I'd like this condition to be simplified, but it depends on the commends I've left for python-discord/site#288. Since `bot` will need to be merged before `site`¹, the best option for that would be:

```py
if reminder.get("jump_url"):
    # Do something if a jump_url is present
```
Using `.get` means we will get a default value (a falsy `None` in this case) when `"jump_url"` is not yet present in the response we get from `site`, an falsy empty string for historical reminders, and a truthy URL string for reminders set after this new feature.

I also think we should reformat this to a more DRY format, since you're currently having almost duplicate lines in your code that make it harder to maintain and to guarantee consistency in the front-end. (In fact, you've got a period after the reminder content in one path, but not in the other.)

We could tackle that like this:
```py
description = ["f"Here's your reminder: `{reminder['content']}`"]
if reminder.get("jump_url"):
    description.append(f"[Jump back to when you created the reminder]({reminder['jump_url']})")

embed.description = "\n".join(description)
```

Although, since we're only combining two strings here, a simple `+` concatenation would also do.

---
1) `bot` needs to be merged before `site` because `site` will require the `jump_url` field to be present after python-discord/site#288 is merged. Merging `site` before `bot` will therefore break the reminder feature.

Akarys42:
Updated.

PR #467, opened 2019-09-30
Forward arguments to !docs get

Ayplow:
A minor usability that allows `!docs Python` in place of `!docs get Python`.

This seems to have been the intention of the old code, however `invoke` doesn't pass the arguments forward by default, so the UX was confusing.

AFAICT this could be written `ctx.invoke(self.get_command, symbol)`, however this would only serve to complicate and slow things down. Lmk if it provides some utility, and I'll change it

### DIFF
--- a/bot/cogs/doc.py
+++ b/bot/cogs/doc.py
@@ -261,7 +261,7 @@ class Doc(commands.Cog):
     @commands.group(name='docs', aliases=('doc', 'd'), invoke_without_command=True)
     async def docs_group(self, ctx: commands.Context, symbol: commands.clean_content = None) -> None:
         """Lookup documentation for Python symbols."""
-        await ctx.invoke(self.get_command)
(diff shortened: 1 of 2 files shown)


bot/cogs/doc.py
```diff
@@ -261,7 +261,7 @@ def __init__(self, bot: commands.Bot):
     @commands.group(name='docs', aliases=('doc', 'd'), invoke_without_command=True)
     async def docs_group(self, ctx: commands.Context, symbol: commands.clean_content = None) -> None:
         """Lookup documentation for Python symbols."""
-        await ctx.invoke(self.get_command)
+        await self.get_command(ctx, symbol)
```
sco1: Commands can't be invoked directly, this needs to be `ctx.invoke`

```suggestion
        await ctx.invoke(self.get_command, symbol)
```

Issue #458, opened 2019-09-26
Support exact timestamps as arguments for moderation commands


MarkKoz:
> Yesterday (for me) we briefly talked about being able to set exact dates instead of a relative duration for the expiration field. I've checked and I've not seen it used, so I'm not married to it. If we do want it, I may be better to have a separate converter that converts datetime strings of a certain format to a datetime instead of creating a hybrid one that both parses relative durations and absolute datetime strings.
> 
> I currently see not no real use for it, though, although it could be useful if we ever want to do something with absolute dates (like adding events to the event calendar we've envisioned.)

_Originally posted by @SebastiaanZ in https://github.com/python-discord/bot/pull/457_

> Exact dates could be handy for underage bans.

_Originally posted by @scragly in https://github.com/python-discord/bot/pull/457_

MarkKoz:
The converter has been added but it currently isn't being anywhere. I think it should be done in a separate PR, after #457.

PR #473, opened 2019-10-01
Add converter for ISO-formatted datetime strings

SebastiaanZ:
This pull requests adds a converter that automatically parses ISO-formatted datetime strings and returns a `datetime.datetime` object. It uses [`dateutil.parser.isoparse`](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse) to do the heavy lifting and therefore supports the same formats.

In addition, I've added tests that "guarantee" that certain standard formats will be accepted and added these "guaranteed" formats to the docstring of the `ISODate.convert` method.

This pull request should make it easy to implement #458. However, I did not yet want to touch the moderation cog, since it's currently being worked on by @MarkKoz. If we merge this soon, Mark may be able to directly implement it during the work he's doing anyway, though.

### DIFF
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -219,7 +219,7 @@ class Duration(Converter):
 
 
 class ISODateTime(Converter):
-    """"Converts an ISO-8601 datetime string into a datetime.datetime."""
+    """Converts an ISO-8601 datetime string into a datetime.datetime."""
 
     async def convert(self, ctx: Context, datetime_string: str) -> datetime:
         """
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from ssl import CertificateError
 from typing import Union
 
+import dateutil.parser
 import discord
 from aiohttp import ClientConnectorError
 from dateutil.relativedelta import relativedelta
@@ -215,3 +216,35 @@ class Duration(Converter):
         now = datetime.utcnow()
 
         return now + delta
+
+
+class ISODateTime(Converter):
+    """"Converts an ISO-8601 datetime string into a datetime.datetime."""
+
+    async def convert(self, ctx: Context, datetime_string: str) -> datetime:
+        """
+        Converts a ISO-8601 `datetime_string` into a `datetime.datetime` object.
+
+        The converter is flexible in the formats it accepts, as it uses the `isoparse` method of
+        `dateutil.parser`. In general, it accepts datetime strings that start with a date,
+        optionally followed by a time.
+
+        See: <https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse>
+
+        Formats that are guaranteed to be valid by our tests are:
+
+        - `YYYY-mm-ddTHH:MM:SS` | `YYYY-mm-dd HH:MM:SS`
+        - `YYYY-mm-ddTHH:MM` | `YYYY-mm-dd HH:MM`
+        - `YYYY-mm-dd`
+        - `YYYY-mm`
+        - `YYYY`
+
+        Note: ISO-8601 specifies a `T` as the separator between the date and the time part of the
+        datetime string. The converter accepts both a `T` and a single space character.
+        """
+        try:
+            dt = dateutil.parser.isoparse(datetime_string)
+        except ValueError:
+            raise BadArgument(f"`{datetime_string}` is not a valid ISO-8601 datetime string")
+
+        return dt
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -5,6 +5,7 @@ from ssl import CertificateError
 from typing import Union
 
 import dateutil.parser
+import dateutil.tz
 import discord
 from aiohttp import ClientConnectorError
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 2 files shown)


bot/converters.py
```diff
@@ -215,3 +216,35 @@ class Duration(Converter):
         now = datetime.utcnow()
 
         return now + delta
+
+
+class ISODateTime(Converter):
+    """"Converts an ISO-8601 datetime string into a datetime.datetime."""
```
sco1:
```suggestion
    """Converts an ISO-8601 datetime string into a datetime.datetime."""
```

tests/test_converters.py
```diff
@@ -184,3 +185,57 @@ def test_duration_converter_for_invalid(duration: str):
...
+)
+def test_isodatetime_converter_for_valid(datetime_string: str, expected_dt: datetime.datetime):
+    converter = ISODateTime()
+    assert asyncio.run(converter.convert(None, datetime_string)) == expected_dt
+
+
+@pytest.mark.parametrize(
+    ("datetime_string"),
+    (
+        # Make sure it doesn't interfere with the Duration converation
```
MarkKoz:
```suggestion
        # Make sure it doesn't interfere with the Duration conversion
```

bot/converters.py
```diff
@@ -215,3 +216,35 @@ class Duration(Converter):
...
+
+    async def convert(self, ctx: Context, datetime_string: str) -> datetime:
+        """
+        Converts a ISO-8601 `datetime_string` into a `datetime.datetime` object.
+
+        The converter is flexible in the formats it accepts, as it uses the `isoparse` method of
+        `dateutil.parser`. In general, it accepts datetime strings that start with a date,
+        optionally followed by a time.
+
+        See: <https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse>
```
MarkKoz: Why is this between angle brackets?

SebastiaanZ:
> How should time zones be dealt with? There is some other code that expects timezone-naïve datetimes (for example, `utils.time.wait_until`) so we could run into trouble with that. Perhaps the better solution would be to make everything else timezone-aware too.

This converter does not force `tzinfo` at the moment. If you feed it a datetime string without a timezone specifier, it will return a naive `datetime.datetime`, if you feed it something that does specify a timezone, it does return a tz-aware `datetime.datetime`:

```py
>>> dateutil.parser.isoparse("2019-09-01T10:10:10")
datetime.datetime(2019, 9, 1, 10, 10, 10)
>>> dateutil.parser.isoparse("2019-09-01T10:10:10Z")
datetime.datetime(2019, 9, 1, 10, 10, 10, tzinfo=tzutc())
```

Do you think we should restrict this to one or the other at the level of the converter?

bot/converters.py
SebastiaanZ: Annoying habit from typing URLs in Discord. I'll change it.

tests/test_converters.py
SebastiaanZ: Thanks, good eye.

MarkKoz:
> This converter does not force tzinfo at the moment.

Yeah, I was aware it depends on the input. Inputs that result in timezone-aware datetimes could be problematic.

> Do you think we should restrict this to one or the other at the level of the converter?

Well, maybe. But as I said, perhaps a more proper solution would be to make other things work with timezone-aware datetimes. It'd be nice to support timezones in user inputs but right now it isn't an important feature anywhere (but could be in the future). 

I don't know how feasible it is to support timezone-aware datetimes everywhere. Our API already returns a `Z` so it results in timezone-aware datetimes. In other instances we use `utcnow()` which could be switched to `now(datetime.timezone.utc)`. I can't think of where else we work with datetimes at the moment.

SebastiaanZ:
Okay, I've looked into the option of using timezone-aware `datetime` objects everywhere and I like the idea. However, I think it is a much broader and larger issue than this converter and I don't think we should tack such a change on this pull request. That's why my proposal is to make sure that this converter returns the UTC-time stored in timezone-**un**aware `datetime` objects for now. 

The reason is that `discord.py` returns unaware `datetime` objects for datetime related attributes and we wouldn't be able to easily compare "our" tz-aware `datetime` objects with these tz-unaware `datetime` objects returned by discord.py. I think that it's therefore best to return tz-unaware `datetime` objects until we've looked into the broader topic of using tz-aware objects everywhere by default in the bot.

tests/test_converters.py
```diff
@@ -184,3 +185,78 @@ def test_duration_converter_for_invalid(duration: str):
...
+        # `YYYY-mm`
+        ('2019-02-01', datetime.datetime(2019, 2, 1)),
+
+        # `YYYY`
+        ('2025', datetime.datetime(2025, 1, 1)),
+    ),
+)
+def test_isodatetime_converter_for_valid(datetime_string: str, expected_dt: datetime.datetime):
+    converter = ISODateTime()
+    assert asyncio.run(converter.convert(None, datetime_string)) == expected_dt
```
MarkKoz: Does the equality comparison in this assertion already cover checking that the returned `datetime` is timezone-naïve?

SebastiaanZ: Good point, I've added an explicit assertion to check that.

PR #481, opened 2019-10-01
Add checks for valid response and retries to fetch_posts()

### LINKED ISSUE #390: The Reddit cog occassionally fails to relay new /r/python messages
Occasionally, the [Reddit cog](/python-discord/bot/blob/master/bot/cogs/reddit.py) stops relaying new messages to the #reddit channel. From what I've seen, there are two ways in which the cog breaks and both should be easy to fix.

#### 1. The background task fails on unexpected http responses
The background task to fetch new posts, the `self.new_posts_task` created in the `on_ready` event listener, assumes that the response we get from the Reddit API is valid `json` in [this line](/python-discord/bot/blob/master/bot/cogs/reddit.py#L55). If, for some temporary reason, we don't get a `json` response, but a `text/html` response (404, probably), the task fails with the following exception:

```py
In [160]: bot.get_cog("Reddit").new_posts_task
Out[160]: 
  File "/bot/bot/cogs/eval.py", line 167, in _eval
    res = await func()
  File "<string>", line 8, in func
  File "/bot/bot/cogs/eval.py", line 167, in _eval
    res = await func()
  File "<string>", line 8, in func
  File "/bot/bot/cogs/reddit.py", line 142, in poll_new_posts
    posts = await self.fetch_posts(f"{subreddit}/new")
  File "/bot/bot/cogs/reddit.py", line 55, in fetch_posts
    content = await response.json()
  File "/bot/.venv/lib/python3.6/site-packages/aiohttp/client_reqrep.py", line 938, in json
    headers=self.headers)
aiohttp.client_exceptions.ContentTypeError: 0, message='Attempt to decode JSON with unexpected mimetype: text/html; charset=utf-8'
```

This line is in the `fetch_posts` utility function used by both the `new_posts_task` as well as the `top_weekly_posts_task`. This means that can break because of this error. 

The likely solution is to either handle the exception or check the response status code before trying to parse it as json. We probably want to either have some kind of retry logic in the handling of the non-200 response, since otherwise the weekly top posts task skips a week.

#### ~~2. The channel can't be initialized during `on_ready`~~

~~Similar to the issue we had with the watch channels, it can occur that the bot's internal channel cache isn't fully loaded when the `on_ready` event fires. This means that the channel will not be retrieved and the tasks will never be started. This only happens intermittently, but it does occasionally happen.~~

~~In such cases, the `self.bot.get_channel` returns `None`, because the channel has not been loaded into the internal bot cache yet at that point.~~

```py
    async def on_ready(self):
        self.reddit_channel = self.bot.get_channel(Channels.reddit)
```

~~While this could be fixed by adding a slight delay or a bit of retry logic, another option is to wait for the migration to a later version of `discord.py` and use `await self.bot.fetch_channel` here instead. That will fetch the channel from the API, bypassing any issues we may have with the internal bot channel cache.~~

bendiller:
Fixes #390

Added checks for response.status and response.content_type for fetch_posts() to validate the response from self.bot.http_session.get(). Also added retries in case a bad response is received.

It's been tested to verify that the `!reddit top` (and similar) commands still work, and I simulated some retry behavior by specifying intentionally bad routes. Please let me know what changes you'd like to see!

### DIFF
--- a/bot/cogs/reddit.py
+++ b/bot/cogs/reddit.py
@@ -21,6 +21,7 @@ class Reddit(Cog):
 
     HEADERS = {"User-Agent": "Discord Bot: PythonDiscord (https://pythondiscord.com/)"}
     URL = "https://www.reddit.com"
+    MAX_FETCH_RETRIES = 3
 
     def __init__(self, bot: Bot):
         self.bot = bot
@@ -42,16 +43,19 @@ class Reddit(Cog):
         if params is None:
             params = {}
 
-        response = await self.bot.http_session.get(
-            url=f"{self.URL}/{route}.json",
-            headers=self.HEADERS,
-            params=params
-        )
+        for _ in range(self.MAX_FETCH_RETRIES):
+            response = await self.bot.http_session.get(
+                url=f"{self.URL}/{route}.json",
+                headers=self.HEADERS,
+                params=params
+            )
+            if response.status == 200 and response.content_type == 'application/json':
+                # Got appropriate response - process and return.
+                content = await response.json()
+                posts = content["data"]["children"]
+                return posts[:amount]
 
-        content = await response.json()
-        posts = content["data"]["children"]
-
-        return posts[:amount]
+        return list()  # Failed to get appropriate response within allowed number of retries.
 
     async def send_top_posts(
         self, channel: TextChannel, subreddit: Subreddit, content: str = None, time: str = "all"
--- a/bot/cogs/reddit.py
+++ b/bot/cogs/reddit.py
@@ -34,9 +34,7 @@ class Reddit(Cog):
         self.new_posts_task = None
         self.top_weekly_posts_task = None
 
-    async def fetch_posts(
-            self, channel: TextChannel, route: str, *, amount: int = 25, params: dict = None
-    ) -> List[dict]:
+    async def fetch_posts(self, route: str, *, amount: int = 25, params: dict = None) -> List[dict]:
         """A helper method to fetch a certain amount of Reddit posts at a given route."""
         # Reddit's JSON responses only provide 25 posts at most.
         if not 25 >= amount > 0:
@@ -57,8 +55,7 @@ class Reddit(Cog):
                 content = await response.json()
                 posts = content["data"]["children"]
                 return posts[:amount]
-            async with channel.typing():
-                await asyncio.sleep(3)
+            await asyncio.sleep(3)
 
         log.debug(f"Invalid response from: {url} - status code {response.status}, mimetype {response.content_type}")
         return list()  # Failed to get appropriate response within allowed number of retries.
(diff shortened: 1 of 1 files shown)


bot/cogs/reddit.py
```diff
@@ -42,16 +43,19 @@ def __init__(self, bot: Bot):
...
+                params=params
+            )
+            if response.status == 200 and response.content_type == 'application/json':
+                # Got appropriate response - process and return.
+                content = await response.json()
+                posts = content["data"]["children"]
+                return posts[:amount]
 
-        return posts[:amount]
+        return list()  # Failed to get appropriate response within allowed number of retries.
```
MarkKoz: Add some debug logging before it returns that indicates an invalid response was received, including the URL, status code, and mimetype.

bendiller: Okay, understood. Just to be clear, you only want to log the details of the final attempt, correct? Rather than logging each failure in the full retry loop? 

Also, would you want any asyncio.sleep() delay in between retries? I'm not sure if that'd make a valid response significantly more likely, but it seems plausible that it might.

MarkKoz: Yes, just log after the final attempt. A sleep would be good (maybe 3 seconds?) but in that case you should make the bot show as typing when the commands are used so users know they're meant to wait.

bendiller: Okay, I'll get a sleep in there!

bendiller: I've added logging, 3 second sleep, and typing indication during sleep - I did have to modify a few function signatures to achieve the typing indication, just a heads up. I ran the code and tested the commands (including with an intentionally bad route) and it seems to work as intended.

bot/cogs/reddit.py
```diff
@@ -55,6 +57,8 @@ def __init__(self, bot: Bot):
                 content = await response.json()
                 posts = content["data"]["children"]
                 return posts[:amount]
+            async with channel.typing():
```
MarkKoz: If this is done instead in `send_top_posts` and wraps the `fetch_posts` call, the signature of `fetch_posts` wont need to be modified.

bendiller: Alrighty. I'm real new to GitHub and collaboration in general - is it preferable for me to revert the last commit and make the change to `send_top_posts`, or for me to leave git alone and just use a new commit to put the code back how it was and implement the change to `send_top_posts`?

MarkKoz: Do whatever is easier for you. I generally avoid reverts in these situations because they're just an extra step.

lemonsaurus: @bendiller You should probably just make a new commit that addresses the review comments, rather than mucking around with reverts or amends.

bendiller: Thanks for the feedback. I implemented the the `async with channel.typing()...` behavior in `send_top_posts` instead as you suggested. I left it out of the `fetch_posts` call in `poll_new_posts`, because in that case there's no user to inform of the delay. I can change that, if you prefer.

bot/cogs/reddit.py
```diff
@@ -42,16 +43,22 @@ def __init__(self, bot: Bot):
...
+                url=url,
+                headers=self.HEADERS,
+                params=params
+            )
+            if response.status == 200 and response.content_type == 'application/json':
+                # Got appropriate response - process and return.
+                content = await response.json()
+                posts = content["data"]["children"]
+                return posts[:amount]
+            await asyncio.sleep(3)
```
MarkKoz: Putting a space here will make it slightly more readable

```suggestion

            await asyncio.sleep(3)
```

bot/cogs/reddit.py
MarkKoz: Nope, no need to add it to the poll functions. This looks good.

bendiller:
Thanks Mark, that's exciting! Particularly because you've personally assisted me in the Discord more than once, haha. Cheers.

PR #484, opened 2019-10-02
Refactor "Cogs" cog

### LINKED ISSUE #479: The "Cogs" cog does not support loading of extensions which are packages instead of modules
What "packages instead of modules" means is that the extension is an `__init__.py` of a sub-package i.e. the `setup()` function is in  `__init__.py` (this is what `watchchannels` is doing right now).

The culprit is this code https://github.com/python-discord/bot/blob/341fb5200acbf24ab1220339b6ce7cf6816561a5/bot/cogs/cogs.py#L28

It's just an overcomplicated and inaccurate way to accomplish this, especially since the [`Bot.extensions`](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.Bot.extensions) attribute exists (perhaps it didn't when this code was written).

MarkKoz:
Notable user-facing changes:

* Fixes #479
* Embeds have been ditched everywhere except for the list command.
* Renamed to _extensions_ because it really loads extensions, not cogs (cogs just happen to be inside extensions). The old _cogs_ aliases are still available.
* `*` can be used to reload only all loaded extensions and `**` can be used to "reload" all extensions, even unloaded extensions.
* Multiple names of extensions can be given to the reload command. If `*` is also given, a union will be done with all loaded extensions and the extensions named explicitly.
* The fully qualified name of the extension is shown in response messages (e.g. confirmation/failure messages, errors, etc.). The list command is an exception and shows unqualified names.
* When batch reloading, errors have been merged into a single list (no more distinction between load/unload errors).
* [`reload_extension()`](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.Bot.reload_extension) is used for reloading, so if an error occurs while reloading, the extension will be reverted to its previously working state.
* When reloading, if an extension was not initially loaded, it will be implicitly loaded anyway.

---

Plan is to rename it to `extensions` and re-do it based on how I did it in another bot, which has less redundancy in terms of output messages:

https://github.com/MarkKoz/d.py-test-bot/blob/master/testbot/extensions/manager.py
https://github.com/MarkKoz/d.py-test-bot/blob/master/testbot/utils/extensions.py

### DIFF
--- a/bot/cogs/alias.py
+++ b/bot/cogs/alias.py
@@ -5,6 +5,7 @@ from typing import Union
 from discord import Colour, Embed, Member, User
 from discord.ext.commands import Bot, Cog, Command, Context, clean_content, command, group
 
+from bot.cogs.extensions import Extension
 from bot.cogs.watchchannels.watchchannel import proxy_user
 from bot.converters import TagNameConverter
 from bot.pagination import LinePaginator
@@ -84,9 +85,9 @@ class Alias (Cog):
         await self.invoke(ctx, "site rules")
 
     @command(name="reload", hidden=True)
-    async def cogs_reload_alias(self, ctx: Context, *, cog_name: str) -> None:
-        """Alias for invoking <prefix>cogs reload [cog_name]."""
-        await self.invoke(ctx, "cogs reload", cog_name)
+    async def extensions_reload_alias(self, ctx: Context, *extensions: Extension) -> None:
+        """Alias for invoking <prefix>extensions reload [extensions...]."""
+        await self.invoke(ctx, "extensions reload", *extensions)
 
     @command(name="defon", hidden=True)
     async def defcon_enable_alias(self, ctx: Context) -> None:
--- a/bot/cogs/extensions.py
+++ b/bot/cogs/extensions.py
@@ -1,5 +1,5 @@
 import logging
-import os
+import textwrap
 import typing as t
 from enum import Enum
 from pkgutil import iter_modules
@@ -13,7 +13,7 @@ from bot.utils.checks import with_role_check
 
 log = logging.getLogger(__name__)
 
-KEEP_LOADED = ["bot.cogs.extensions", "bot.cogs.modlog"]
+UNLOAD_BLACKLIST = {"bot.cogs.extensions", "bot.cogs.modlog"}
 EXTENSIONS = frozenset(
     ext.name
     for ext in iter_modules(("bot/cogs",), "bot.cogs.")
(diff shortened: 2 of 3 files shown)


bot/cogs/extensions.py
```diff
@@ -0,0 +1,231 @@
...
+
+    @extensions_group.command(name="reload", aliases=("r",))
+    async def reload_command(self, ctx: Context, *extensions: Extension) -> None:
+        """
+        Reload extensions given their fully qualified or unqualified names.
+
+        If an extension fails to be reloaded, it will be rolled-back to the prior working state.
+
+        If `*` is given as the name, all currently loaded extensions will be reloaded.
+        If `**` is given as the name, all extensions, including unloaded ones, will be reloaded.
```
MarkKoz: The asterisks inside the inline code blocks need to be escapes but I don't know how. The help command italicises the docstring, so having these asterisks messes things up.

bot/cogs/extensions.py
```diff
@@ -0,0 +1,231 @@
...
+        """Apply an action to an extension and return the status message and any error message."""
+        verb = action.name.lower()
+        error_msg = None
+
+        try:
+            action.value(self.bot, ext)
+        except (commands.ExtensionAlreadyLoaded, commands.ExtensionNotLoaded):
+            if action is Action.RELOAD:
+                # When reloading, just load the extension if it was not loaded.
+                return self.manage(ext, Action.LOAD)
```
MarkKoz: Is implicitly loading an extension if it isn't loaded undesirable behaviour? If not, another candidate would be to implicitly reload when `!ext load` is done on an already loaded extension.

bot/cogs/alias.py
```diff
@@ -5,6 +5,7 @@
 from discord import Colour, Embed, Member, User
 from discord.ext.commands import Bot, Cog, Command, Context, clean_content, command, group
 
+from bot.cogs.extensions import Extension
```
MarkKoz: The converter uses the `EXTENSIONS` set, which is a module-level constant in `extensions.py`. Importing it here means it iterates the modules again, but the alias must use the converter... Is there a good way to avoid that happening or is it not worth the trouble?

SebastiaanZ: As far as I can tell, it will not iterate twice by importing it here, since it has already been imported when we load the `alias` cog and Python will not import it again. I've added a temporary logging call and that seems to confirm that the code is just executed once. It's only when I reload the `extensions` extension that the logging call is made again, which is to be expected, since `discord.py` explicitly makes sure to re-import the module to refresh the code.

bot/cogs/extensions.py
SebastiaanZ: I've tried a couple of things, but I have no idea how to do this either. We may need to break with our usual style here and go with another solution than an inline code block. I think it may just be impossible for now.

bot/cogs/extensions.py
SebastiaanZ: I think that the intention of loading the cog is clear when someone tries to reload a cog, so I don't mind this behavior. The alternative proposal, reloading when someone tries to load something, sounds more counter-intuitive to me. If I recall correctly, the previous version with the misnomer `cogs` denied both, but I often found it annoying that it wouldn't just load an extension on `reload` when it wasn't loaded.

bot/cogs/extensions.py
```diff
@@ -0,0 +1,216 @@
...
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @group(name="extensions", aliases=("ext", "exts", "c", "cogs"), invoke_without_command=True)
+    async def extensions_group(self, ctx: Context) -> None:
+        """Load, unload, reload, and list loaded extensions."""
+        await ctx.invoke(self.bot.get_command("help"), "extensions")
+
+    @extensions_group.command(name="load", aliases=("l",))
+    async def load_command(self, ctx: Context, extension: Extension) -> None:
```
SebastiaanZ: Is there any particular reason why you allow multiple extensions to be reloaded at once, but don't allow multiple extensions to be loaded or unloaded at once? This feels like an unnecessary break of symmetry between the sub-commands to me. 

Now, in your current version, we could use `reload` to `load` multiple extensions at once if they're unloaded, but that feels like a bit of a hack to me.

bot/cogs/extensions.py
```diff
@@ -0,0 +1,216 @@
...
+        If `**` is given as the name, all extensions, including unloaded ones, will be reloaded.
+        """
+        if not extensions:
+            await ctx.invoke(self.bot.get_command("help"), "extensions reload")
+            return
+
+        if len(extensions) > 1:
+            msg = await self.batch_reload(*extensions)
+        else:
+            msg, _ = self.manage(extensions[0], Action.RELOAD)
```
SebastiaanZ: The docstring leads me to believe that providing a `*` or `**` as the single argument to this subcommand would respectively reload all the loaded cogs or all of the cogs regardless of their status. However, this `if-else` block will call the `self.manage` method directly when a single argument is provided. Since `self.manage` does not have any logic to handle asterisk characters, it will fail with:

```py
Oct 07 09:35:01 Bot: |               bot.cogs.extensions |    ERROR | Extension '*' failed to load.
  File "/home/sebastiaan/pydis/repositories/djangobot/bot/cogs/extensions.py", line 180, in manage
    action.value(self.bot, ext)
  File "/home/sebastiaan/.local/share/virtualenvs/djangobot-Py5o4YeS/lib/python3.7/site-packages/discord/ext/commands/bot.py", line 687, in reload_extension
    raise errors.ExtensionNotLoaded(name)
discord.ext.commands.errors.ExtensionNotLoaded: Extension '*' has not been loaded.

During handling of the above exception, another exception occurred:

  File "/home/sebastiaan/.local/share/virtualenvs/djangobot-Py5o4YeS/lib/python3.7/site-packages/discord/ext/commands/bot.py", line 621, in load_extension
    lib = importlib.import_module(name)
  File "/home/sebastiaan/.local/share/virtualenvs/djangobot-Py5o4YeS/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
ModuleNotFoundError: No module named '*'

The above exception was the direct cause of the following exception:

  File "/home/sebastiaan/pydis/repositories/djangobot/bot/cogs/extensions.py", line 180, in manage
    action.value(self.bot, ext)
  File "/home/sebastiaan/.local/share/virtualenvs/djangobot-Py5o4YeS/lib/python3.7/site-packages/discord/ext/commands/bot.py", line 623, in load_extension
    raise errors.ExtensionNotFound(name, e) from e
discord.ext.commands.errors.ExtensionNotFound: Extension '*' could not be loaded.
```

That means that, effectively, we can only use `*`/`**` in combination with at least one other argument. I don't think that's very intuitive. I think it's better for asterisk arguments to work on their own as well.

MarkKoz: It was my intention for it to work on its own. I broke it when I was attempting to simplify the logic here. I think the fix is

```py
if len(extensions) > 1 or "*" in extensions:
```

on line 99. I'll test this tomorrow.

bot/cogs/extensions.py
MarkKoz: It was like that in the old version so I left it. I could make them all support wildcards but I'd have to rewrite the batch function a bit to work to accept load and unload too. I don't mind looking into that.

bot/cogs/alias.py
MarkKoz: I suspected this may be the cas. I just don't quite understand how imports work I suppose. Thanks for figuring that out though.

bot/cogs/extensions.py
MarkKoz: Yeah, I felt the same way with the second proposal. Leaving this unresolved for now in case anyone else has opinions to share.

bot/cogs/extensions.py
MarkKoz: An alternative style would be to just have it as normal text. Escaping it seems to work then, but I will confirm tomorrow.

bot/cogs/extensions.py
SebastiaanZ: That's on me, then. I never noticed that we had this functionality in the old version, I thought you added this handy feature. I don't think it's necessary, it just struck me as an asymmetry when reading the code. I think it could be a nice refinement to make the sub-commands work in a consistent manner, but I'm not opposed to keeping the current version either. 

I'll leave this conversation open, but if no one else wants this, we can just close it. If we miss it down the line, we can always opt for a refinement later.

bot/cogs/extensions.py
SebastiaanZ: Normal text sounds fine to me. (As far as I can tell, escaping then works as expected.)

mathsman5133: If you add a space between the \` and \*, ie \` *`, it will make the astericks in the codeblock, but the whole description won't be italicised. Other than that I couldn't figure it out either

bot/cogs/extensions.py
```diff
@@ -0,0 +1,216 @@
...
+class Extensions(commands.Cog):
+    """Extension management commands."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @group(name="extensions", aliases=("ext", "exts", "c", "cogs"), invoke_without_command=True)
+    async def extensions_group(self, ctx: Context) -> None:
+        """Load, unload, reload, and list loaded extensions."""
+        await ctx.invoke(self.bot.get_command("help"), "extensions")
```
mathsman5133: Since we upgraded d.py could this be simplified to `await ctx.send_help(ctx.command)`?

MarkKoz: Tried that before. Didn't seem to work with our custom help command so I left it alone and didn't look into it much.

bot/cogs/extensions.py
MarkKoz: I've add support for this now. The only caveat is that for `!ext load`, when multiple extensions are given, if any of those are already loaded, they will still be displayed as loaded (i.e. included in the total load count and no error shown). When a single extension is given, it will show an error if it's already loaded. The same applies to `!ext unload`. I figured this isn't a big deal and makes implementation simpler.

bot/cogs/extensions.py
MarkKoz: Went with normal text.

bot/cogs/extensions.py
SebastiaanZ: That's totally fine with me.

bot/cogs/extensions.py
```diff
@@ -0,0 +1,228 @@
...
+    async def reload_command(self, ctx: Context, *extensions: Extension) -> None:
+        """
+        Reload extensions given their fully qualified or unqualified names.
+
+        If an extension fails to be reloaded, it will be rolled-back to the prior working state.
+
+        If '\*' is given as the name, all currently loaded extensions will be reloaded.
+        If '\*\*' is given as the name, all extensions, including unloaded ones, will be reloaded.
+        """  # noqa: W605
+        if not extensions:
```
mathsman5133: This line will never get reached, as passing in no extensions will raise `commands.MissingRequiredArgument` first (which does a good enough job of telling you arguments are wrong)

mathsman5133: Sorry that's my bad I was using the alias

MarkKoz: This made me realise the other two commands are missing this check.

bot/cogs/extensions.py
lemonsaurus: I think it's fine, too.

PR #485, opened 2019-10-02
Source Command

### LINKED ISSUE #331: Source command.
A source command has been discussed as a good addition in the past for the bot, with the following features:
- `!source` shows the base github repository link
- `!source <command>` will provide a github link to the relevant module:line_number directly

Feel free to recommend any additional suggestions in the comments below.

RohanJnr:
closes #331 

Source command implemented as said in the issue.
The following is a screenshot of it being implemented:

sourcecmd

### DIFF
--- a/bot/cogs/source.py
+++ b/bot/cogs/source.py
@@ -5,7 +5,7 @@ import discord
 from discord.ext import commands
 
 from bot import constants
-
+from bot.decorators import in_channel
 
 logger = logging.getLogger(__name__)
 
@@ -16,15 +16,23 @@ class Source(commands.Cog):
     def __init__(self, bot: commands.Bot) -> None:
         self.bot = bot
 
+    @in_channel(constants.Channels.bot)
     @commands.command(name="source")
-    async def command_source(self, ctx: commands.Context, command_name: str = None) -> None:
+    async def command_source(self, ctx: commands.Context, command_name: str = None, sub_command: str = None) -> None:
         """View the source of a command."""
         if command_name is None:
-            return await ctx.send("> https://github.com/python-discord/bot")
+            await ctx.send("> https://github.com/python-discord/bot")
+            return
 
-        command = self.bot.get_command(command_name)
+        if sub_command:
+            await ctx.send(sub_command)
+            command = self.bot.get_command(f"{command_name} {sub_command}")
+        else:
+            command = self.bot.get_command(command_name)
+
         if command is None:
-            return await ctx.send("No such command found.")
+            await ctx.send("No such command found.")
+            return
 
         url = self.get_command_url(command)
 
@@ -42,19 +50,26 @@ class Source(commands.Cog):
         """Make up the url for the source of the command."""
         # Get the source code
         src_code_object = command.callback.__code__
-
-        # Get module name and replace . with /
-        module_name = command.callback.__module__
-        module_name = module_name.replace(".", "/") + ".py"
+        file_path = src_code_object.co_filename[5:]
 
         # Get line number and set last line number
         lines_list, starting_line_no = inspect.getsourcelines(src_code_object)
         lines = len(lines_list)
+
+        # Get module
+        module_name = command.callback.__module__
+        module_name = module_name.replace(".", "/") + ".py"
+
         last_line_no = starting_line_no + lines - 1
 
         # Make up the url and return
         base_url = "https://github.com/python-discord/bot/tree/master/"
-        final_url = f"<{base_url}{module_name}#L{starting_line_no}-L{last_line_no}>"
+
+        # use the module name if the file path goes into the decorators.
+        if file_path == "bot/decorators.py":
+            final_url = f"{base_url}{module_name}#L{starting_line_no}-L{last_line_no}"
+        else:
+            final_url = f"{base_url}{file_path}#L{starting_line_no}-L{last_line_no}"
 
         return final_url
 
--- a/bot/cogs/source.py
+++ b/bot/cogs/source.py
@@ -1,5 +1,6 @@
 import inspect
 import logging
+import os
 
 import discord
 from discord.ext import commands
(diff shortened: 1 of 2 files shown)


RohanJnr:
> Hey @RohanJnr,
> 
> Like mentioned in the issue, the standard approach does not work consistently for our bot, as it does not always return the right starting line and/or code length. I have not looked into why it's happening, although it may be because of decorators passing the function around.
> 
> As an example, issuing `!source help` sends me to https://github.com/python-discord/bot/blob/master/bot/cogs/help.py#L108-L140, which is obviously not where we want to end up. Help _may_ be the only one for which this happens, but I haven't extensively tested all commands yet.
> 
> One additional problem is way you reconstruct the path to the `.py` file: We've recently started using a subpackage approach (directory with `__init__.py`) to group related functionality together. The command currently does not take that into account and expects all of the extension files to live directly in `bot/cogs/`.
> 
> As an example, issuing `!source superstarify` yields the link https://github.com/python-discord/bot/tree/master/bot/cogs/superstarify.py#L148-L219. This file does not exist and the function is actually located here: https://github.com/python-discord/bot/blob/master/bot/cogs/superstarify/__init__.py#L148-L219. (Notice that the line numbers are correct in this case.)
> 
> I think that if we do add this to the bot, the approach should be robust and not suffer from these edge cases. Additionally, I think that you should consider adding support for subcommands (`!source docs get` now sends you to the source code for `docs`, not the `get` subcommand) and restricting the command to channels like #bot-commands.
> 
> I do think the code you wrote is clean and easy to read and I think the functionality would be cool to have.

Thanks for the suggestion! will work on improving it.

RohanJnr:
btw by subcommands, do u mean commands within a group?

RohanJnr:
the only pending work is the pathfinding for the help command. I am not sure how to go about it. The subcommands URL generation works fine except for the tags group, some weared stuff happening. Maybe you can look through my code and find some fixes?

bot/cogs/source.py
```diff
@@ -16,15 +16,23 @@ class Source(commands.Cog):
     def __init__(self, bot: commands.Bot) -> None:
         self.bot = bot
 
+    @in_channel(constants.Channels.bot)
     @commands.command(name="source")
-    async def command_source(self, ctx: commands.Context, command_name: str = None) -> None:
+    async def command_source(self, ctx: commands.Context, command_name: str = None, sub_command: str = None) -> None:
```
NeKitDS: At the current state, only one command and one subcommand are supported. Why couldn't we do this? : `async def command_source(self, ctx: commands.Context, *, command_name: str = None) -> None:`

bot/cogs/source.py
```diff
@@ -16,15 +16,23 @@ class Source(commands.Cog):
...
         """View the source of a command."""
         if command_name is None:
-            return await ctx.send("> https://github.com/python-discord/bot")
+            await ctx.send("> https://github.com/python-discord/bot")
```
NeKitDS: Not sure why wouldn't we do `return await ctx.send(...)` in one line honestly

bot/cogs/source.py
```diff
@@ -16,15 +16,23 @@ class Source(commands.Cog):
...
+            return
+
+        if sub_command:
+            await ctx.send(sub_command)
+            command = self.bot.get_command(f"{command_name} {sub_command}")
+        else:
+            command = self.bot.get_command(command_name)
 
-        command = self.bot.get_command(command_name)
```
NeKitDS: Instead of that `sub_command` usage, we could do `self.bot.get_command(command_name)`.

bot/cogs/source.py
sco1: If we're annotating that our function returns `None` then it should return `None`, not maybe `None`

bot/cogs/source.py
```diff
@@ -42,19 +50,26 @@ def get_command_url(command: commands.Command) -> str:
         """Make up the url for the source of the command."""
         # Get the source code
         src_code_object = command.callback.__code__
-
-        # Get module name and replace . with /
-        module_name = command.callback.__module__
-        module_name = module_name.replace(".", "/") + ".py"
+        file_path = src_code_object.co_filename[5:]
```
NeKitDS: Rapptz's implementation of getting the location was like this: `file_name = src_code_object.co_filename` and `location = os.path.relpath(filename).replace('\\', '/')`. I think, should work; will do some tests in a bit.

bot/cogs/source.py
NeKitDS: Okay, I have just checked in discord.py. `send` returns a `discord.Message` instance, well then, we should return `None`.

bot/cogs/source.py
```diff
@@ -0,0 +1,80 @@
...
+        embed.description += f"`{prefix}{command.name}`\n\n {url}"
+
+        await ctx.send(embed=embed)
+
+    @staticmethod
+    def get_command_url(command: commands.Command) -> str:
+        """Make up the url for the source of the command."""
+        # Get the source code
+        src_code_object = command.callback.__code__
+        file_path = src_code_object.co_filename[5:]
```
NeKitDS: Rapptz's implementation of getting the location was like this: `file_name = src_code_object.co_filename` and `file_path = os.path.relpath(file_name).replace('\\', '/')`. I think, should work; will do some tests in a bit.

bot/cogs/source.py
```diff
@@ -0,0 +1,80 @@
...
+            return
+
+        url = self.get_command_url(command)
+
+        prefix = constants.Bot.prefix
+
+        embed = discord.Embed(colour=discord.Colour.red())
+        embed.title = "Command Source"
+        embed.description = f"**{command.name.capitalize()}**\n"
+        embed.description += f"`{prefix}{command.name}`\n\n {url}"
```
NeKitDS: Shouldn't we send `command.qualified_name` in the embed?

RohanJnr:
Updated

bot/cogs/source.py
```diff
@@ -18,42 +19,44 @@ def __init__(self, bot: commands.Bot) -> None:
...
-            command = self.bot.get_command(command_name)
-
+        command = self.bot.get_command(command_name)
         if command is None:
             await ctx.send("No such command found.")
             return
 
-        url = self.get_command_url(command)
+        if command.name == "help":
+            url = "https://github.com/python-discord/bot/blob/master/bot/cogs/help.py#L475-L490"
```
NeKitDS: In my opinion, we would better send a link without the lines here.

NeKitDS:
All in all, @SebastiaanZ I think we are ready to merge (though, not sure if we should send the lines in explicit `!source help` implementation.

bot/cogs/source.py
RohanJnr: Which lines

NeKitDS: The `#L475-L490` should not be hard-coded in my opinion. I think we should just return the link to the `help.py` file.

lemonsaurus: I think we should just return the correct lines, but without hardcoding them.

@RohanJnr we can't have any hardcoded edge cases in this solution. Please try to find a solution that solves this without hardcoding.

RohanJnr:
what if I just return the file name for the help command? I will not hard code the file link tho, I can get it by the code.

lemonsaurus:
Returning only the file name for the help command seems like a cheap solution. I think that this is solvable without hardcoding any of the replies. Given how stale this pull request is and how many unaddressed reviews are in it, I'm tempted to just give this task to someone else if you can't come up with an agnostic solution.

PR #488, opened 2019-10-03
Prepare cogs on cog init & wait for bot ready flag

### LINKED ISSUE #101: Review uses of `on_ready` in cogs
**Originally posted by Scragly:**

There are 12 cogs that use the event `on_ready` to perform some form of preparation for that cog.

While that's fine if the extension is loaded before the bot is started, any cases where loading occurs after this stage will result in the preparatory tasks not being done at all, resulting in missing data the cogs rely on to function. This includes cases where a cog may be reloaded during runtime, which often happens during development to refresh to code.

Instead of running when the bot dispatches the ready event, it would be best to instead have it prepare every time a cog is loaded, but have it wait for the bot's ready flag:  
```python
class Cog:
    def __init__(self, bot):
        self.bot = bot
        ...
        bot.loop.create_task(self.prepare_method())

    async def prepare_method(self):
        await self.bot.wait_until_ready()
        # do prep stuff
        ...
```

This is a job for after the Github migration.

K4cePhoenix:
Closes #101. 

Calls preparation methods, when a cog is loaded and waits for the ready flag of the bot as requested in #101.


K4cePhoenix:
> It'd be better to use more specific names for the functions rather than `prepare_cog`. They all have docstrings so they'll give you an idea of what the functions do and what to name them.

Makes sense, I'll look for better names
> Also, in the future, please comment on the issue that you want to work on it before opening a PR.

Sorry for that, will do in the future!

PR #490, opened 2019-10-04
Show previous watch reason and total after invoking watch command

### LINKED ISSUE #404: Using BB watch should show info about any previous unwatch reason.
We provide a reason when unwatching someone, which should be info that is displayed in case someone goes to watch them again. This could be particularly helpful for nominations also.

MarkKoz:
Closes #404 

bild

Here's a much longer reason:
bild
(Note: I changed "watches" to "infractions" in the total after taking this screenshot)

* Are you happy with the format?
* Should the total include the just-added infraction or exclude it? It currently excludes it.

### DIFF
--- a/bot/cogs/watchchannels/bigbrother.py
+++ b/bot/cogs/watchchannels/bigbrother.py
@@ -83,7 +83,7 @@ class BigBrother(WatchChannel, Cog, name="Big Brother"):
             )
 
             if len(history) > 1:
-                total = f"({len(history) // 2} infractions total)"
+                total = f"({len(history) // 2} previous infractions in total)"
                 end_reason = history[0]["reason"]
                 start_reason = f"Watched: {history[1]['reason']}"
                 msg += f"\n\nUser's previous watch reasons {total}:```{start_reason}\n\n{end_reason}```"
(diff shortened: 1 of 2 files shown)


sco1:
Seems like a fine start as-is

> Should the total include the just-added infraction or exclude it? It currently excludes it.

I'd say exclude, with a tweak of the verbiage to something like `n previous watches`

As an aside, these commands are nearly identical in implementation between the two submodules. While perfectly functional, this seems like something that could be refactored into something more generic. From a brief scan, this seems to be the case for much of the watchchannel logic in general, so maybe this is better investigated within the scope of a different PR.

MarkKoz:
> As an aside, these commands are nearly identical in implementation between the two submodules. While perfectly functional, this seems like something that could be refactored into something more generic. From a brief scan, this seems to be the case for much of the watchchannel logic in general, so maybe this is better investigated within the scope of a different PR.

I was fooled into thinking that. While some of the checks at the start certainly could be refactored, the nominations actually use their own endpoint rather than the infractions one. I tried some stuff but realised that and reverted it as I didn't want to deal with moderation cog 2.0 right now.

PR #493, opened 2019-10-05
Create the !mention command.

### LINKED ISSUE #491: !mention command
Currently, all the roles in the community are unmentionable. This is in order to prevent spam. Before we made this change, people would occasionally come into the community and just mentionspam all the role mentions.

The problem with this is that we sometimes need to mention a role. For example, we mention the `@Announcements` role every time we make an announcement. We all desperately wish Discord would allow us to set mention permissions so that only certain roles were allowed to mention other roles, but this is not yet possible.

Currently the manual process looks like this:
1. We set the role to be mentionable by everyone
2. We ping the role.
3. We wait 5-10 seconds
4. We set the role to be unmentionable.

The reason we need to do step 3 is that if you set the role to unmentionable too soon after you do the ping, Discord has a tendency to mess up the mention. Sometimes it will not ping everyone, other times there will be a ghost ping.

We'd like to automate the above process, so we're looking for someone to make a `!mention` command. Here's what we'd like to see:
* Only Moderators, Admins and Owners should be able to use it.
* It takes a role ID or a string as input argument
* The string should not need to be in quotes, even if it has spaces.
* When called, the bot should unlock the role for mention and let the caller know that the role has been unlocked.
* When a role mention for this role is detected in any channel, wait 5 seconds (make this duration configurable), lock the role again and let the caller know that it has been locked in the channel where the unlock was invoked.
* After 5 minutes (this should be configurable), if no role mention has been detected, the unlock should time out and the role should be locked again. Remember here also to let the caller know that it has been locked in the channel where the unlock was invoked.
* If invoked on a role that's already mentionable (like `@Admins`), return an error message
* If invoked on a special role (like `@everyone`), return an error.

mathsman5133:
Create the !mention command.
    - Pass in a role ID or name (no quotes, case sensitive)
    - Bot will set the role to "mentionable"
    - Bot will wait for a default of 300 seconds or until a message mentioning the role is sent
    - If a message is sent, wait an additional default 5 seconds
    - Set the role to be unmentionable and alert the user.

Closes #491


scragly:
Can you please add an informative description to the PR.

Linting has also failed due to a number of issues; please ensure you have the precommit hook installed with `pipenv run precommit` so you will be prevented from committing when linting fails.

To see locally your linting results use `pipenv run lint`.

Your commit history has a PR reference for your own fork at the beginning, can you please rebase to remove that. I'd suggest also fixing your linting issues at the same time to prevent the commit history adding a superfluous `fix linting` commit.

mathsman5133:
Hopefully they're all fixed. Not sure position on using `asyncio.run` to send the message but the alternative is quite messy

mathsman5133:
Done 👍 

I added the ID to the error message if the user was not a staff member - I assumed this might be handy if they ping and run.

bot/cogs/utils.py
```diff
@@ -128,6 +129,59 @@ def get_info(char: str) -> Tuple[str, str]:
...
+    def readable_time(seconds: int) -> str:
+        """Returns a number of seconds into a human-readable minutes/seconds combination."""
+        minutes, seconds = divmod(seconds, 60)
+
+        if minutes:
+            fmt = '{m}min {s}sec'
+        else:
+            fmt = '{s}sec'
+
+        return fmt.format(m=minutes, s=seconds)
```
jchristgit: We have something for this in `bot/utils/time.py` but it only works for deltas. Maybe you can cast to a delta here and use that function instead?

bot/cogs/utils.py
```diff
@@ -128,6 +129,59 @@ def get_info(char: str) -> Tuple[str, str]:
...
+        )
+
+        def check(m: Message) -> bool:
+            """Checks that the message contains the role mention."""
+            return role in m.role_mentions
+
+        try:
+            msg = await self.bot.wait_for("message", check=check, timeout=Mention.message_timeout)
+        except TimeoutError:
+            await role.edit(mentionable=False, reason=f"Automatic role lock - timeout.")
```
jchristgit:
```suggestion
            await role.edit(mentionable=False, reason="Automatic role lock - timeout.")
```

bot/cogs/utils.py
mathsman5133: Done 👍

Issue #320, opened 2019-02-25
Have @Python add @everyone pings to checkpoint on a monthly basis


SebastiaanZ:
We've noticed that adding an @everyone message to the checkpoint channel every once in a while triggers a flood of users accepting our rules to gain the Developers role. Currently, those @everyone messages are added manually every once in a while, but it should be very straightforward to automate the process.

I would leave the current welcome message untouched, since it's a message from a human, but do it like this:

1. Remove previous @everyone message from @Python in that channel
2. Post message that says something like:
    - @everyone Please ping `@Admins` if you encounter any problems during the verification process.

I'm not sure what frequency is best for this. I wouldn't mind it once a month or even once a week. If people get bothered by it, they either accept and join for real or leave the server in some kind of self-purge (they've not accepted our rules anyway).

kraktus:
What do you think is the best approach to set this command? Adding a 

```python
asyncio.sleep(xxx)
``` Does not seems a reliable to do so.

MarkKoz:
Probably the [tasks](https://discordpy.readthedocs.io/en/latest/ext/tasks/index.html) extension.

MarkKoz:
Though that may be problematic if the bot restarts... something could be stored in the database to keep track of the last time it was muted. I don't know how well that'd work with the tasks extension then. But we do have an API endpoint for storing miscellaneous settings in the database as key-value pairs (currently only used for DEFCON I believe).

kraktus:
Yes, basically using the tasks extension raise the same problems than using a simple sleep. 'll look at the API endpoint.

sco1:
I don't know if there's a reason why we can't just have a short-circuit check of the timestamp of the existing message in the channel, in theory there's only ever going to be one message in the channel so iterating over the channel history to get it isn't going to be expensive.

lemonsaurus:
That's alright, I'll assign the issue to you.

Meanwhile, I would like to see the message include something about how they haven't verified yet.

**How about this:**
`@everyone` To verify that you have read our rules, please type `!accept`. Ping `@Admins` if you encounter any problems during the verification process.

For frequency, I think weekly sounds about right.

MarkKoz:
I don't think they have any way of pinging admins, since they only have access to the verification channel and that channel removes all messages IIRC. They'd have to DM, wouldn't they?

sco1:
We still get the notification, which alleviates the burdeon of the user searching for someone to DM, but it can be a pain to find it in the message logs.

[...]

PR #494, opened 2019-10-05
Added to the `.gitignore` a new file to be ignored .DS_Store

kraktus:
Only on Mac OS. `.DS_Store` is a file that stores custom attributes of its containing folder. New contributors on Mac OS won't have to bother anymore about this mysterious file that is created when you fork the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,6 @@ config.yml
 
 # JUnit XML reports from pytest
 junit.xml
+
+# Mac OS .DS_Store, which is a file that stores custom attributes of its containing folder
+.DS_Store


lemonsaurus:
Hi @kraktus, and thanks for the PR!

I'm going to mark this PR as `invalid` and close it. The reason I'm doing this is because, in the spirit of Hacktoberfest, we're not accepting oneliners that don't have an open issue for them at this time.

**However**; that doesn't mean this change isn't welcome. I would suggest you add this change to the PR you open when you solve the issue you're assigned to https://github.com/python-discord/bot/issues/320. A famous programming book called The Pragmatic Programmer refers to this sort of a small improvement added to an otherwise unrelated PR as "kaizen", which is Japanese for improvement. Here at Python Discord, we subscribe to this philosophy of allowing out of scope kaizen improvements, so we're happy to allow this kind of thing in another PR.

kraktus:
Ok, no problem, I just though after reading this: https://github.com/python-discord/bot/blob/master/CONTRIBUTING.md an especially point **4** I should have split my PR in 2 because the changes don't have anything in common.

scragly:
Yep, that's specific to scope of commit messages. You can have multiple commits in a PR as long as they clearly define different purposes.

lemonsaurus:
Yeah exactly. You can make a kaizen commit in an otherwise unrelated PR, just make sure it gets its own commit and that you explain in detail in the commit message what this commit is for.

Here's some good information on how we'd like the commit messages to look:
https://thoughtbot.com/blog/5-useful-tips-for-a-better-commit-message
http://ablogaboutcode.com/2011/03/23/proper-git-commit-messages-and-an-elegant-git-history
https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

PR #495, opened 2019-10-05
Add Periodic Ping to Checkpoint

### LINKED ISSUE #320: Have @Python add @everyone pings to checkpoint on a monthly basis

kraktus:
Hello again,

Per #320, this PR implements a new function `periodic_ping` in `verification.py` to ping weekly users that did not accepted the rules

This PR also adds to `.gitignore` a new file to ignore (created specifically on Mac OS).

Before posting this PR I ran flake8 and a precommit hook.


bot/cogs/verification.py
```diff
@@ -27,12 +29,16 @@
 If you'd like to unsubscribe from the announcement notifications, simply send `!unsubscribe` to <#{Channels.bot}>.
 """
 
+PERIODIC_PING = (f"@everyone To verify that you have read our rules, please type `!accept`."
```
lemonsaurus: this doesn't have to be an f-string.

We also prefer that you use K&R indentation style for these, which would look like this:
```py
PERIODIC_PING = (
    "first string"
    f"other string"
)
```

bot/cogs/verification.py
```diff
@@ -155,6 +161,20 @@ def bot_check(ctx: Context) -> bool:
         else:
             return True
 
+    @tasks.loop(hours=1.0)
```
lemonsaurus: It strikes me as unnecessarily often to check every hour if it's been a week yet. that means we're doing 167 of these checks that return None for every one that does something.

I'm thinking something like 12 or even 24 hours is plenty. Even if the message is skewed by 23 hours every couple of weeks, it's not at all critical that this happens the same time every week or that there's exactly one week between each of these.

bot/cogs/verification.py
```diff
@@ -155,6 +161,20 @@ def bot_check(ctx: Context) -> bool:
...
+    async def periodic_ping(self) -> None:
+        """Post a recap message every week with an @everyone."""
+        message = await self.bot.get_channel(Channels.verification).history(limit=1).flatten()  # check last message
```
lemonsaurus: I'm wondering if it might be better to be more specific about the message to check. There is a small chance that the last message will be the bot saying "`@someuser` Please type `!accept` to verify that you have ...", which the bot will do whenever someone types something other than `!accept`. This message stays for about 10 or 15 seconds before being deleted. in other words, there's a chance this will fail even when it shouldn't.

Of course, the price for such a failure is very low, it just means that we'll wait a little longer before posting this, so I don't know if it's worth fixing, but a slightly more clever method for ensuring that we're checking the correct message would prevent it.

kraktus:
Hopefully it’s all fixed

bot/cogs/verification.py
```diff
@@ -27,12 +29,17 @@
 If you'd like to unsubscribe from the announcement notifications, simply send `!unsubscribe` to <#{Channels.bot}>.
 """
 
+PERIODIC_PING = (
+    "@everyone To verify that you have read our rules, please type `!accept`."
+    f" Ping <@&{Roles.admin}> if you encounter any problems during the verification process.")
```
MarkKoz: The closing parenthesis should be on a new line.

bot/cogs/verification.py
```diff
@@ -155,6 +162,26 @@ def bot_check(ctx: Context) -> bool:
         else:
             return True
 
+    @tasks.loop(hours=12)
+    async def periodic_ping(self) -> None:
+        """Post a recap message every week with an @everyone."""
+        messages = await self.bot.get_channel(Channels.verification).history(limit=5).flatten()  # check lasts messages
+        messages_content = [i.content for i in messages]
+        if PERIODIC_PING not in messages_content:  # if the bot did not posted yet
```
MarkKoz: It strikes me that there is some redundancy in this code in terms of loops. The loop inside the `else` can be used instead of the `messages_content` comprehension e.g.

```py
posted = False

async for message in messages:
    if message.content == PERIODIC_PING:
        posted = True
        ...

if not posted:
    ...send(PERIODIC_PING)
```

**Note that in this case the history iterator would not need to be flattened.**

I think an approach which would be better and still appropriate is to `break` after finding the message (in theory there should only ever be 0 or 1 of the periodic message in the history):

```py
async for message in messages:
    if message.content == PERIODIC_PING:
        ...
        break
else:
    ...send(PERIODIC_PING)
```

kraktus:
All good normally

PR #500, opened 2019-10-07
Token and bad code

### LINKED ISSUE #389: Token remover and auto codeblock response conflict
When a user posts code that's not formatted correctly with a discord.py token the bot deletes their message cause there's a token in it.  However it proceeds to repost their code **with** their token in a "how to format your code" response.

kraktus:
This PR fix #389 

Split the `token_remover` in two functions, one that checks that there is a token in the message (`is_token_in_message`), and the other takes the actions (deleting the message, etc.)
This new function `is_token_in_message` is reused in `cogs/bot.py` to know if the poorly formatted code contains a token. If it true, then the embed message is not displayed, since `token_remover.py` is already in charge.


bot/cogs/token_remover.py
```diff
@@ -86,6 +75,23 @@ def mod_log(self) -> ModLog:
                 channel_id=Channels.mod_alerts,
             )
 
+    def is_token_in_message(self, msg: Message) -> bool:
```
jchristgit: This should probably be a classmethod, so you can call it from the codeblock cog without passing the class as the first argument, whilst still having access to the other methods.

bot/cogs/token_remover.py
```diff
@@ -86,6 +75,24 @@ def mod_log(self) -> ModLog:
                 channel_id=Channels.mod_alerts,
             )
 
+    @classmethod
+    def is_token_in_message(self, msg: Message) -> bool:
```
jchristgit: I was expecting flake8 to nag here, but I guess there's no lint for it.

Via [PEP 8, the first argument to a classmethod should be named `cls`](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments)

kraktus: will be done, thanks

kraktus:
@sco1 Hi, I am sorry but I am not sure about what you mean. Do you say the bot should also checks if a message contains a token every time a message is edited?

sco1:
Yes

lemonsaurus:
@kraktus any ETA on when you can add the requested changes from @sco1? We can't merge until that's done.

kraktus:
Pretty busy right now, will do it as soon as I can.

kraktus:
Not sure about using a `classmethod` or a `static method`.

kraktus:
In case it is not clear, I changed my code to respond to the requested changes.

bot/cogs/bot.py
```diff
@@ -237,7 +238,7 @@ def has_bad_ticks(self, msg: Message) -> bool:
             and len(msg.content.splitlines()) > 3
         )
 
-        if parse_codeblock:
+        if parse_codeblock and not TokenRemover.is_token_in_message(msg):  # no token in the msg
```
tagptroll1: Could this not be in parse_codeblock or a check that returns early before this?
```
if not TokenRemover.is_token_in_message(msg):
    return
...
```

scragly:

PR #501, opened 2019-10-08
Utilise `tasks.loop` and webhooks for reddit postings.

### LINKED ISSUE #391: Using a webhook for the Reddit post relay
We're currently relaying the new Reddit posts (and the weekly top posts overview) as regular messages to a text channel. Since the subreddit sees frequent activity, I propose to change this to a method that uses a webhook to post the messages to the channel. 

This has two advantages:

1. Webhooks use a different rate limit pool than regular messages;
2. We can use a stylized author icon and display_name

The latter is not really important, but will look nice; the former is the biggest advantage of this change.

This should not be difficult to implement, although an admin+ needs to create the proper webhook for the relevant channel.

mathsman5133:
- The auto-posting has been removed as it was spammy and few used it.
- Instead a daily top posting loop has been created. I stuck with top 5 daily for consistency with top 5 weekly and 5 pinned messages, but this can be easily changed
- Both loops will post at midnight
- As before, the weekly top posts will be pinned.
- A webhook id for the Reddit channel has been set in constants and config, but will need to be created by an admin (and real ID set).
- I set the webhook author to be what was previously the content of the message - it looks clean to me, but here is an example:

image

Not sure what to do about the icon, maybe a generic reddit logo would be fine (assuming copyright stuff is fine, but that might be easier set when creating the webhook?)

I've left a few notes about how to simplify the loops (by setting time parameters) when discord.py v1.3 gets released (it's in alpha now - is there any reason not to upgrade?)

Closes #391


bot/cogs/reddit.py
```diff
@@ -113,103 +120,63 @@ def __init__(self, bot: Bot):
...
-                # Only show new posts if we've checked before.
-                if subreddit in self.last_ids:
-                    for post in posts:
-                        data = post["data"]
+        midnight_tomorrow = now.replace(day=now.day + 1, hour=0, minute=0, second=0)
+        seconds_until = (midnight_tomorrow - now).total_seconds()
 
-                        # Convert the ID to an integer for easy comparison.
-                        int_id = int(data["id"], 36)
+        await asyncio.sleep(seconds_until)
```
SebastiaanZ: The `discord.py` Tasks extensions allows you to pass a single (or a list of) `datetime.time` to the `time` parameter of `loop()` to trigger an iteration at the specified time. That way, we don't have to manually calculate `seconds_until` and include a `asyncio.sleep`. 

I think this should do it:

```py
import datetime
@loop(time=datetime.time.min)
```

The use of `datetime.time.min` guarantees we're aiming for the midnight exactly, although I don't think a few seconds really matter.

bot/cogs/reddit.py
```diff
@@ -113,103 +120,63 @@ def __init__(self, bot: Bot):
...
-                    embed.set_footer(text=f"Posted by u/{data['author']} in {subreddit}")
-                    embed.colour = Colour.blurple()
+        # Calculate the amount of seconds until midnight next monday.
+        # when we upgrade d.py this can be simplified to be run every midnight and just check against day
+        monday = now + timedelta(days=7 - now.weekday())
+        monday = monday.replace(hour=0, minute=0, second=0)
+        until_monday = (monday - now).total_seconds()
 
-                    await self.reddit_channel.send(embed=embed)
+        await asyncio.sleep(until_monday)
```
SebastiaanZ: I don't see a way to avoid some logic here by purely using the Task extension, as you only specify a target `time` not a day in the week. One way we could avoid having to calculate the delay and execute the sleep ourselves is by having the `daily` task check if it's Monday and, if so, trigger the weekly top posts function.

This means that the daily posts coroutine wakes up at every midnight and if it's Monday, it will also trigger the weekly top posts coroutine to do its business.

bot/cogs/reddit.py
mathsman5133: I absolutely agree, although the `time` parameter is new in version 1.3 - were in v1.2.3, unless we wish to move to v1.3a (I see no problem, personally).

I've added a note in the comments that this should be fixed once we upgrade

mathsman5133:
I've updated the tasks to use the `time` parameter and combined them into one, however in upgrading discord.py to the latest (v1.3a) version it breaks the linter which seems to want to lint discord.py source, too...

MarkKoz:
This is why we used to have [this](https://github.com/python-discord/bot/commit/5fdacedccb70b2f549439aaa709beb24e84737c1#diff-fec826feae04e51c0d94076385408bdcL18). I guess we deemed it no longer necessary since we did not foresee having to use non-PyPI releases anymore.

And I don't know if it's OK to move to 1.3a. Hopefully someone more familiar with how things are going with that version's development can comment.

mathsman5133:
Sure. FWIW, I've had a (400 guild) bot running on latest versions of v1.3a for months now with absolutely no issues. I don't believe (and just had a quick look at commit history) that any breaking changes have been made so far. Instead, small additions (new methods or not breaking), bugfixes and doc changes have been made.

scragly:
Yeah, please avoid changing dependencies without first discussing it, especially when you're going to entirely change the release source. 

For example, changing things to use v1.3a requires all our docker images and our CI to use git for installing, which we've purposely avoided. It would greatly impact our build times for CI/Docker Images thanks to pipenv version resolution and git installation taking significant extra time.

Slipping this into a PR would also result in our docker images not being ready to handle that and failing when merged to master, as well as the linter not ignoring the git-pulled lib files, because these changes weren't discussed and considered with the devops team, so the image files and linter settings weren't adjusted accordingly.

Please revert these dependency changes.

mathsman5133:
I ran pipfile lock/update and it seemed to update other deps in the `Pipfile.lock`. Do I need to revert these too?

scragly:
If you've not actually changed any deps you can just reset the Pipfile and Pipfile.lock to avoid any edge cases, but for reference, updating within scope of the existing pins are usually less of a concern and are typically done anytime a new dep is added anyhow.

mathsman5133:
Perfect. Thanks for that, and sorry for the issues, taken note for future 👍

mathsman5133:
Should I resolve the conflicts?

scragly:
You can yep, it's up to you how you want to do it though, either by adding a new resolution commit or by doing a rebase to maybe take advantage of cleaning up history on this branch while also applying changes on top of current origin master without conflict.

mathsman5133:
Well that was a trek and a half, but got there in the end. The last thing we need is an actual webhook ID (I've just chucked 123456789 in there for the moment)

lemonsaurus:
Minor note, @mathsman5133 -- force pushing to a branch after reviews have started makes the pull request history difficult to read. I'm missing valuable context here about which changes you made in response to previous reviews, because those changes are all aggregated at the bottom. 

Please try to avoid force pushing after you've posted your pull request, although force pushing before you open it is fine.

bot/cogs/reddit.py
```diff
@@ -63,23 +74,18 @@ def __init__(self, bot: Bot):
         log.debug(f"Invalid response from: {url} - status code {response.status}, mimetype {response.content_type}")
         return list()  # Failed to get appropriate response within allowed number of retries.
 
-    async def send_top_posts(
-        self, channel: TextChannel, subreddit: Subreddit, content: str = None, time: str = "all"
-    ) -> Message:
-        """Create an embed for the top posts, then send it in a given TextChannel."""
+    async def get_top_posts(self, subreddit: Subreddit, time: str = "all") -> Embed:
+        """Returns an embed for the top posts of the given subreddit."""
```
lemonsaurus:
```suggestion
        """Get the top 5 posts for a subreddit within a specified timeframe."""
```
Couple of things here:
* Use imperative mood in docstrings, e.g. `Return an embed`, not `Returns an embed`.
* Don't describe the return type, explain what the method _does_. We have return type annotations to explain what it returns.
* Be specific. Explaining how the `time` parameter works, for example, seems like useful API documentation, and in my opinion you should spend a few more lines doing just that beyond what I suggested here.

bot/cogs/reddit.py
```diff
@@ -63,23 +74,18 @@ def __init__(self, bot: Bot):
...
-                amount=5,
-                params={
-                    "t": time
-                }
-            )
+        embed = Embed(description="")
+
+        posts = await self.fetch_posts(
+            route=f"{subreddit}/top",
+            amount=5,
```
lemonsaurus: Perhaps this method might take the amount as a parameter and default to 5? Would be more future-proof and generic.

bot/cogs/reddit.py
```diff
@@ -63,23 +74,18 @@ def __init__(self, bot: Bot):
...
+            params={
+                "t": time
+            }
```
lemonsaurus: splitting this into three lines seems overkill. a one-liner would probably yield better readability here.

config-default.yml
```diff
@@ -141,6 +141,7 @@ guild:
     webhooks:
         talent_pool:                        569145364800602132
         big_brother:                        569133704568373283
+        reddit:                             123456789
```
lemonsaurus: you can use `635408384794951680`

bot/cogs/reddit.py
```diff
@@ -115,103 +119,51 @@ def __init__(self, bot: Bot):
...
-                if subreddit.lower() == "r/python":
-                    # Remove the oldest pins so that only 5 remain at most.
-                    pins = await self.reddit_channel.pins()
+                # Remove the oldest pins so that only 5 remain at most.
+                pins = await self.channel.pins()
 
-                    while len(pins) >= 5:
-                        await pins[-1].unpin()
-                        del pins[-1]
+                while len(pins) >= 5:
```
lemonsaurus: this isn't really a code review, but I think 5 is unnecessarily low. We can probably double that. Let's set it to, say, 12.

mathsman5133:
Note taken on the force pushes, thanks and sorry.

I have tested the webhooks by running with no sleeping, although have not set them working for a week or so. Am happy to do that if you'd like.

The commands are working fine.

PR #503, opened 2019-10-08
Add raw command

### LINKED ISSUE #334: Message raw info command.
It's been expressed that a command that provides raw info on a set message by it's ID would be beneficial. It would:
- Show raw content
- Show raw embed data
- Show attachment data

avayert:
Example:
Shows a help message embed

Shows output for running the command on the previous embed

Closes #334

### DIFF
--- a/bot/cogs/information.py
+++ b/bot/cogs/information.py
@@ -3,8 +3,7 @@ import logging
 import pprint
 import textwrap
 import typing
-from collections import Mapping
-from typing import Any, Optional
+from typing import Any, Optional, Mapping
 
 import discord
 from discord import CategoryChannel, Colour, Embed, Member, TextChannel, VoiceChannel
--- a/bot/cogs/information.py
+++ b/bot/cogs/information.py
@@ -3,7 +3,7 @@ import logging
 import pprint
 import textwrap
 import typing
-from typing import Any, Optional, Mapping
+from typing import Any, Mapping, Optional
 
 import discord
 from discord import CategoryChannel, Colour, Embed, Member, TextChannel, VoiceChannel
@@ -12,7 +12,7 @@ from discord.ext import commands
 from discord.ext.commands import Bot, Cog, Context, command, group
 
 from bot.constants import Channels, Emojis, MODERATION_ROLES, STAFF_ROLES
-from bot.decorators import InChannelCheckFailure, with_role, in_channel
+from bot.decorators import InChannelCheckFailure, in_channel, with_role
 from bot.utils.checks import with_role_check
 from bot.utils.time import time_since
 
@@ -234,7 +234,8 @@ class Information(Cog):
 
         await ctx.send(embed=embed)
 
-    def format_fields(self, mapping: Mapping[str, Any], field_width: Optional[int] = None):
+    def format_fields(self, mapping: Mapping[str, Any], field_width: Optional[int] = None) -> str:
+        """Format a mapping to be readable to a human."""
         # sorting is technically superfluous but nice if you want to look for a specific field
         fields = sorted(mapping.items(), key=lambda item: item[0])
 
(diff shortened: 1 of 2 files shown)


bot/cogs/information.py
```diff
@@ -1,13 +1,18 @@
...
+import pprint
 import textwrap
 import typing
+from typing import Any, Mapping, Optional
 
-from discord import CategoryChannel, Colour, Embed, Member, Role, TextChannel, VoiceChannel, utils
-from discord.ext.commands import Bot, Cog, Context, command
+import discord
+from discord import CategoryChannel, Colour, Embed, Member, TextChannel, VoiceChannel
+from discord import Role, utils
```
sco1: Any reason why these lines were split?

kraktus: The only reason I can think about is the limitation of 120 characters, but I did not count

avayert: no, I'm guessing PyCharm had a moment

bot/cogs/information.py
```diff
@@ -229,6 +234,81 @@ def __init__(self, bot: Bot):
...
+            # replace backticks as it breaks out of code blocks. Spaces seemed to be the most reasonable solution.
+            # we hope it's not close to 2000
+            paginator.add_line(content.replace('```', '`` `'))
+            paginator.close_page()
+
+        if message.content:
+            add_content('Raw message', message.content)
+
+        transformer = pprint.pformat if json else self.format_fields
+        for field_name in 'embeds attachments'.split():
```
MarkKoz: This makes more sense even though it does the same thing
```suggestion
        for field_name in ('embeds', 'attachments'):
```

bot/cogs/information.py
```diff
@@ -229,6 +234,81 @@ def __init__(self, bot: Bot):
...
+            paginator.add_line(f'== {title} ==\n')
+            # replace backticks as it breaks out of code blocks. Spaces seemed to be the most reasonable solution.
+            # we hope it's not close to 2000
+            paginator.add_line(content.replace('```', '`` `'))
+            paginator.close_page()
+
+        if message.content:
+            add_content('Raw message', message.content)
+
+        transformer = pprint.pformat if json else self.format_fields
```
MarkKoz: Using `json.dumps` with an indent of 2 or 4 will look better and more readable because it puts newlines after brackets.

avayert: I disagree, since it doesn't split up long text into multiple lines. This makes the embed very cumbersome to read.
Compare:
comparison of the two methods

lemonsaurus: I agree with @avayert

bot/cogs/information.py
lemonsaurus: yeah, I prefer this, too.

bot/cogs/information.py
MarkKoz: Is there a way to get the best of both worlds? Or too complicated?

bot/cogs/information.py
```diff
@@ -229,6 +233,81 @@ def __init__(self, bot: Bot):
...
+        # remove trailing whitespace
+        return out.rstrip()
+
+    @group(invoke_without_command=True)
+    @in_channel(Channels.bot, bypass_roles=STAFF_ROLES)
+    async def raw(self, ctx: Context, *, message: discord.Message, json: bool = False) -> None:
+        """Shows information about the raw API response."""
+        # I *guess* it could be deleted right as the command is invoked but I felt like it wasn't worth handling
+        # doing this extra request is also much easier than trying to convert everything back into a dictionary again
+        raw_data = await ctx.bot.http.get_message(message.channel.id, message.id)
```
jchristgit: can a user DOS the bot by spamming this command and making it get stuck in ratelimits?

bot/cogs/information.py
```diff
@@ -229,6 +233,81 @@ def __init__(self, bot: Bot):
...
+
+        paginator = commands.Paginator()
+
+        def add_content(title: str, content: str) -> None:
+            paginator.add_line(f'== {title} ==\n')
+            # replace backticks as it breaks out of code blocks. Spaces seemed to be the most reasonable solution.
+            # we hope it's not close to 2000
+            paginator.add_line(content.replace('```', '`` `'))
```
jchristgit: why not zero-width space to preserve the triple quotes in a readable way?

avayert: Because then if you try to copy paste it's not immediately obvious why the code blocks don't work

bot/cogs/information.py
avayert: Yeah, most likely; this didn't occur to me. It should probably be gated behind a cooldown, and I don't know what the actual ratelimit is but some people said it's `1/1`. How strict should the cooldown be?

jchristgit: very, IMO. I don't see a big need for regular users to run this on a constant basis

jchristgit: but maybe the others have a better idea

lemonsaurus: how about a strict cooldown for regular users and no cooldown for staff?

jchristgit: that sounds like a good idea

bot/cogs/information.py
avayert: I can't think of anything great right now but I'll put it on the backburner in case I have a revelation

bot/cogs/information.py
avayert: I've added a 2 per 3 minutes ratelimit with staff bypasses. I chose 2 because there is a subcommand that they might want to compare to the parent invokation and letting them do that immediately seemed not-too-bad. 3 minutes was arbitrary.

PR #506, opened 2019-10-09
Expand token detection regex character exclusion

sco1:
The existing bot token regex pattern  (`[^\s\.]+\.[^\s\.]+\.[^\s\.]+`) fails to detect tokens in most use-cases beyond simple assignment.

e.g. this would be caught:

```py
token = <token>
```

While these would not:

```py
print(<token>)
bot.run(<token>)
```

To mitigate this, the character exclusion has been expanded to include some of the more commonly used anchors: `[^\s\.()\"']+\.[^\s\.()\"']+\.[^\s\.()\"']+` in order to try and limit what the regex is capturing as a potential token match.

This still seems pretty fragile, but may be the best that can be done (with a regex) unless there's a formal spec for Discord's bot tokens floating around on the internet somewhere.


MarkKoz:
I've regenerated a token a bunch of times and have determined with reasonable certainty that it has a character set of `[\w-]` and has consistent lengths of groups between the dots:

```regex
[\w-]{24}\.[\w-]{6}\.[\w-]{27}
```

sco1:
I don't know if we can assume that tokens are a constant length, per [some discussions](https://github.com/discord-net/Discord.Net/pull/1204) found on the Discord API server. Looking at the test cases they generated in the PR, it looks like only the final field varies, so we should be able to just vary its with and use the simpler regex.

MarkKoz:
That sounds good. Even using the more limited character set should make the regex a lot less fragile.

MarkKoz:
Are the comments about the token format really accurate? If the first part is indeed base64, the character set should be `[a-zA-Z0-9+=\\]`. However, in practice, I have not seen `+=\`. It states the middle is an integer but I have definitely seen both letters and numbers in there as well as a dash.

lemonsaurus:
I feel like without more than anecdotal evidence that `+=/` do not occur, it would be safer to just include the full base64 character set.

I'm also not comfortable assuming the length of the substrings will be constant, neither right now and especially not in the future.

So, for that reason, I think we should just go ahead and merge this as is.

PR #507, opened 2019-10-09
Added a the current nickname of a user in infraction search

### LINKED ISSUE #337: Provide current nickname in embed for infraction search
Will reduce confusion when someone does an infraction search since most users have a nickname completely different to their actual username.

Potential duplicate to [this issue](https://github.com/python-discord/bot/issues/125), not sure if it would be better to just comment on that one.

kraktus:
If the user has not changed its nickname, display *None*, otherwise display its current nickname.

Fix #337


kraktus:
Looks like moderations code changed a lot, I will try to resolve myself the conflits and the repost a PR.

PR #508, opened 2019-10-09
Add the current nickname of a user in infraction search

### LINKED ISSUE #337: Provide current nickname in embed for infraction search

kraktus:
If the user has not changed its nickname, display *None*, otherwise display its current nickname.

closes #337

### DIFF
--- a/bot/cogs/moderation/management.py
+++ b/bot/cogs/moderation/management.py
@@ -228,7 +228,10 @@ class ModManagement(commands.Cog):
         actor = guild.get_member(actor_id)
         active = infraction["active"]
         user_id = infraction["user"]
-        user_nickname = guild.get_member(user_id).nick  # If nickname different from username show nickname
+        if guild.get_member(user_id) is not None:  # If the user is still in the server
+            user_nickname = guild.get_member(user_id).nick  # If nickname different from username show nickname
+        else:
+            user_nickname = None
         hidden = infraction["hidden"]
         created = time.format_infraction(infraction["inserted_at"])
         if infraction["expires_at"] is None:
--- a/bot/cogs/moderation/management.py
+++ b/bot/cogs/moderation/management.py
@@ -228,6 +228,7 @@ class ModManagement(commands.Cog):
         actor = guild.get_member(actor_id)
         active = infraction["active"]
         user_id = infraction["user"]
+        user_nickname = guild.get_member(user_id).nick  # If nickname different from username show nickname
         hidden = infraction["hidden"]
         created = time.format_infraction(infraction["inserted_at"])
         if infraction["expires_at"] is None:
@@ -238,8 +239,7 @@ class ModManagement(commands.Cog):
         lines = textwrap.dedent(f"""
             {"**===============**" if active else "==============="}
             Status: {"__**Active**__" if active else "Inactive"}
-            User: {self.bot.get_user(user_id)} (`{user_id}`)
-            Nickname: {actor.nick or "*None*"}
+            User: {user_nickname or self.bot.get_user(user_id)} (`{user_id}`)
             Type: **{infraction["type"]}**
             Shadow: {hidden}
             Reason: {infraction["reason"] or "*None*"}
--- a/bot/cogs/moderation/management.py
+++ b/bot/cogs/moderation/management.py
@@ -239,6 +239,7 @@ class ModManagement(commands.Cog):
             {"**===============**" if active else "==============="}
             Status: {"__**Active**__" if active else "Inactive"}
             User: {self.bot.get_user(user_id)} (`{user_id}`)
+            Nickname: {actor.nick or "*None*"}
             Type: **{infraction["type"]}**
             Shadow: {hidden}
             Reason: {infraction["reason"] or "*None*"}
--- a/bot/cogs/moderation/management.py
+++ b/bot/cogs/moderation/management.py
@@ -224,10 +224,10 @@ class ModManagement(commands.Cog):
         """Convert the infraction object to a string representation."""
         actor_id = infraction["actor"]
         guild = self.bot.get_guild(constants.Guild.id)
-        actor = guild.get_member(actor_id) or await self.bot.fetch_user(actor_id)
+        actor = guild.get_member(actor_id)
         active = infraction["active"]
         user_id = infraction["user"]
-        user_nickname = self.bot.get_user(user_id) or await self.bot.fetch_user(user_id)
+        user = guild.get_member(user_id)
         hidden = infraction["hidden"]
         created = time.format_infraction(infraction["inserted_at"])
         if infraction["expires_at"] is None:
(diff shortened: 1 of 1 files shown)


kraktus:
> The nickname is supposed to be for the user, not the actor. To be clear, the user is the one which received the infraction. The actor is the one who gave out the infraction. You can see that the actor is already mentioned so there's no need to show their nickname anyway.
> 
> Also, I think it'd be better to replace the user's name with a nickname if it's available rather than adding it on a separate line.

Ok, because I muted myself during tests I didn't see the difference

bot/cogs/moderation/management.py
```diff
@@ -228,6 +228,10 @@ def infraction_to_string(self, infraction: utils.Infraction) -> str:
         actor = guild.get_member(actor_id)
         active = infraction["active"]
         user_id = infraction["user"]
+        if guild.get_member(user_id) is not None:  # If the user is still in the server
+            user_nickname = guild.get_member(user_id).nick  # If nickname different from username show nickname
+        else:
+            user_nickname = None
```
MarkKoz: Can be set to `get_user()` instead of `None` to avoid doing `x or y` later. Keep in mind `get_user()` also returns `None` so it'd be good to set it to an empty string in such case to avoid displaying `"None"` as the username.

MarkKoz: I am not sure but it's quite likely that if a user is not in the guild then they wont be in the cache either and thus `get_user()` won't find them. Could use `fetch_user()` instead, downside being that makes an API call. However, I think that's OK. Anyone else have thoughts on that?

kraktus:
I think using the `<@id>` is the best method because it always works and resolves and the issues!

jchristgit:
Disagree about <@id> being the best method. It doesn't always work, as Discord's clients will often only load online users into the cache, which means that these will display as something like the following:

image

This is especially true on mobile.

kraktus:
> Disagree about <@id> being the best method. It doesn't always work, as Discord's clients will often only load online users into the cache, which means that these will display as something like the following:
> 
> image
> 
> This is especially true on mobile.

Yes I though even if the user was not in the cache clicking on the mention would load it, unfortunately it's not true. I will switch to `fetch_user` instead, even if I would have preferred not to relie on the API.

kraktus:
After some more tests, to use `fetch_user` I have to turn `infraction_to_string` into a `async` function which tons of errors. Do you have any idea to resolve them?

kraktus:
I don't really understand why `actor` is searched in a different way than `user`, so I leaved it the way it was, but instead of doing this:
```
        actor_id = infraction["actor"]
        guild = self.bot.get_guild(constants.Guild.id)
        actor = guild.get_member(actor_id)
```
I think this would be better:
```
       actor_id = infraction["actor"]
       actor = self.bot.get_user(actor_id)
```
I would be also more consistent.

@jchristgit no more use of `<@id>`

jchristgit:
> @jchristgit no more use of <@id>

Using `.mention` is the same as building `<@id>` yourself though: https://github.com/Rapptz/discord.py/blob/c7a1f5e6e9f18fff3128d67b6d098a4ab842fcb7/discord/user.py#L204-L207

The only exception is that you're also fetching the user now which can make this command very slow due to the rate limiting that endpoint has set. `get_user` loads from cache.

I think the most simple approach to this is to:

- Leave the current display of user#discrim as-is
- Move the get_user call before the string
- If the user's nickname != the user's username, display something like "{user} (nickname '{user}')" instead of the current "{user}" setup

kraktus:
> Using `.mention` is the same as building `<@id>` yourself though: https://github.com/Rapptz/discord.py/blob/c7a1f5e6e9f18fff3128d67b6d098a4ab842fcb7/discord/user.py#L204-L207

Damn, how could I forget? I said it myself...
> Also I am pretty sure that `actor.mention` relies on it.

> The only exception is that you're also fetching the user now which can make this command very slow due to the rate limiting that endpoint has set. `get_user` loads from cache.

Fetching only if the user is not in the cache

> If the user's nickname != the user's username, display something like "{user} (nickname '{user}')" instead of the current "{user}" setup

Well that's not what MarkKoz says:
> Also, I think it'd be better to replace the user's name with a nickname if it's available rather than adding it on a separate line.

I agree with him because adding the nickname to the username will longer too much the line 

> Move the get_user call before the string

Already the case I think?

kraktus:
Simplest way to do this I think is not to worry about mentions at all, and display the infos as raw text. I will modify it tonight.

SebastiaanZ:
Hey, @kraktus, I was wondering, are you also a member of our Discord community? What's your handle on there?

kraktus:
@SebastiaanZ I changed it to take the same username as here: kraktus :)

MarkKoz:
> Well that's not what MarkKoz says:
>
>> Also, I think it'd be better to replace the user's name with a nickname if it's available rather than adding it on a separate line.
>
>I agree with him because adding the nickname to the username will longer too much the line

Well, depends on what the staff deems to be useful information. I was thinking both the nickname and discriminator are not needed because the ID will always be shown anyway.

bot/cogs/moderation/management.py
```diff
@@ -221,30 +220,41 @@ def infractions_cog(self) -> Infractions:
...
+        user = guild.get_member(user_id)
         hidden = infraction["hidden"]
         created = time.format_infraction(infraction["inserted_at"])
         if infraction["expires_at"] is None:
             expires = "*Permanent*"
         else:
             expires = time.format_infraction(infraction["expires_at"])
 
+        if user is None:  # if user not in cache
+            user = self.bot.get_user(user_id) or await self.bot.fetch_user(user_id)
```
MarkKoz: I think error handling would be needed for the API calls for two reasons:

1. `NotFound` is probably raised when a user deletes their account
2. It'd be silly to have the entire command fail because of an exception raised for a single non-critical API call. Who knows why `HTTPException` could be raised but it could happen...

MarkKoz: It doesn't look like discord.py will cache the user returned from the API call which isn't ideal. Could we cache it ourselves.

With the caching and error handling, perhaps it's becoming more effort than it's worth. But I think moderators, the ones going to use this command, should be the ones to comment on whether they'd be OK with not displaying names if not in the cache.

kraktus: I will add a exception catch to the `fetch_user` function. I don't change the display output for now, waiting for a mod opinion.

kraktus: @MarkKoz a ```py
try: 

except:``` Would be sufficient or should I use the error_handler?

MarkKoz: Yes, use a try-except to catch `discord.HttpException`.

kraktus:
@MarkKoz I honestly think that's out of reach for me, because I don't understand all the Api-related code of the repo and I've no idea where to begin to implement the requested function in `utils.py`.

MarkKoz:
@kraktus Are you still interested in trying? We can help you through it on Discord as needed. For starters, you could take a look at other code in the bot to see _how_ requests to the API are made.

kraktus:
Unfortunately I won't have time, and after 7 commits didn't progress at all on the objective, I'm closing this PR.

PR #510, opened 2019-10-10
Use OAuth to be Reddit API compliant

### LINKED ISSUE #354: Reddit JSON API OAuth requirement not met.
After a convo in #mods today, we found out that the json feeds on reddit are still required by TOS to have OAUTH. This will need to be added into the bot, as the current method of getting the r/Python feed is without OAUTH, and we're having to tell people to comply in our own server, making this look a bit awkward.

K4cePhoenix:
Closes #354.
- Requests API tokens on extension init
- Client ID and Secrets for Reddit API set up in bot constants and config
- HTTP Requests use the API to receive responses
- Change User-Agent to be complaint with Reddits API rules (https://github.com/reddit-archive/reddit/wiki/API)

Client_ID and Secret have to be created by an admin (https://www.reddit.com/prefs/apps).
The User-Agent string has to be modified with proper values.
Further commits are going to be needed as this PR will conflict with the changes in #501.


bot/cogs/reddit.py
```diff
@@ -19,8 +21,13 @@
 class Reddit(Cog):
     """Track subreddit posts and show detailed statistics about them."""
 
-    HEADERS = {"User-Agent": "Discord Bot: PythonDiscord (https://pythondiscord.com/)"}
+    # Change your client's User-Agent string to something unique and descriptive,
+    # including the target platform, a unique application identifier, a version string,
+    # and your username as contact information, in the following format:
+    # <platform>:<app ID>:<version string> (by /u/<reddit username>)
+    USER_AGENT = "docker:Discord Bot of https://pythondiscord.com/:v?.?.? (by /u/PythonDiscord)"
```
jchristgit: `platform` should probably be something like `python3`, I don't think `docker` makes a lot of sense here.

Also, the versioning doesn't have lots of value. I think we should export an environment variable like `BUILD_COMMIT_SHA` in the build process via `git` and use that here instead.

bot/cogs/reddit.py
```diff
@@ -34,6 +41,66 @@ def __init__(self, bot: Bot):
...
+                "Authorization": "bearer " + self.access_token,
+                "User-Agent": self.USER_AGENT
+            }
+        else:
+            self.client_auth = None
+            self.access_token = None
+            self.refresh_token = None
+            self.headers = None
+
+            log.error("Unable to find client credentials.")
```
jchristgit: What's the actual implication of this? The cog doesn't work?

bot/cogs/reddit.py
K4cePhoenix: Reddits API rules used `android` as an example for platform, so I deemed `docker` to be fitting. We could also be more descriptive and change it to something like `docker-python` or maybe even `docker-python/aiohttp`.

Using `BUILD_COMMIT_SHA` as version string sounds good to me.

bot/cogs/reddit.py
K4cePhoenix: Yes pretty much. Without client id or secret, we can not get an acces token and therefore not use Reddit API. 

Perhaps this should be changed to something more descriptive like `No client credentials were provided. Cannot use Reddit API` and unload the cog?

jchristgit: Yeah, that sounds like a good idea.

MarkKoz: It would be even more robust to check if authentication succeeds rather than just if the credentials are present. In any case, I agree that unloading the extension should be done.

lemonsaurus:
Failing lint. 

`./bot/cogs/reddit.py:120:1: W293 blank line contains whitespace`

bot/cogs/reddit.py
Akarys42: You can also do:
```py
sha = subprocess.run(['git', 'rev-parse', '--short', 'HEAD'], stdout=subprocess.PIPE).stdout.decode('utf-8')
```
to get the current commit sha1.

bot/cogs/reddit.py
```diff
@@ -47,6 +53,61 @@ def channel(self) -> TextChannel:
         """Get the #reddit channel object from the bot's cache."""
         return self.bot.get_channel(Channels.reddit)
 
+    async def get_access_tokens(self) -> None:
```
mathsman5133: Would it be better to utilise the 1-hour expiry of the tokens to not refresh them *every* time a command is run, instead only if a command is run and the access token is older than an hour? For me, if I were to check the top command, I would probably check the weekly and daily at the same time - even if I was just testing/seeing what the command does. This may reduce some unnecessary token regeneration, keeping the command a little faster.

Something like
```py
# setting:
self.access_token = (token, datetime.utcnow() + timedelta(hours=1))

# retrieving
if self.access_token[1] > datetime.utcnow():
    # regenerate
return self.access_token[0]
```
Could use a named tuple if you're feeling fancy.

bot/cogs/reddit.py
K4cePhoenix: I've found, Azure Pipelines has a `BuildNumber` variable, which would be more fitting to use as a version value, imho

However, I haven't used Azure Pipelines and Docker prior to this bot, so implementing it is a bit tricky. ^^

bot/cogs/reddit.py
K4cePhoenix: Agree, I'll change it.

bot/cogs/reddit.py
MarkKoz: I really think it's not worth the hassle. The version number in the user agent is ultimately quite inconsequential. We can just hard-code 1.0.0 or something. Thoughts?

PR #511, opened 2019-10-11
Prevent too similar off-topic channel names

### LINKED ISSUE #465: Add check to prevent too similar off-topic channel names
Right now, there's nothing discouraging us from having these off-topic names added to the list:
```
kosaa
kosaaa
kosaaaa
```
Although that would be nice, we should have the bot by default return that an off-topic name is too similar to an existing one and thus wasn't added.

## Implementation details
- Show the similar name and the similarity
- Don't add the name if it is >X% similar to an existing one
- Add a `!otn forceadd` to bypass the check

kosayoda:
**Closes #465**

Currently, there is nothing stopping very similar names being added to the rotation, eg. `kosa` and `kosaa`.
To solve this, a check is added that rejects names that have a close match using `difflib.get_close_matches` with a cutoff of 0.8.

A command `!otn forceadd` or `!otn fa` was added to override this check and add the name to the rotation anyway.

image


bot/cogs/off_topic_names.py
```diff
@@ -96,14 +96,44 @@ def cog_unload(self) -> None:
...
+    @with_role(*MODERATION_ROLES)
+    async def force_add_command(self, ctx: Context, *names: OffTopicName) -> None:
+        """Forcefully adds a new off-topic name to the rotation."""
         # Chain multiple words to a single one
         name = "-".join(names)
+        await self._add_name(ctx, name)
 
+    async def _add_name(self, ctx: Context, name: str) -> None:
+        """Adds an off-topic channel name to the site storage."""
         await self.bot.api_client.post(f'bot/off-topic-channel-names', params={'name': name})
```
MarkKoz:
```suggestion
         await self.bot.api_client.post('bot/off-topic-channel-names', params={'name': name})
```

bot/cogs/off_topic_names.py
```diff
@@ -96,14 +96,44 @@ def cog_unload(self) -> None:
...
         log.info(
             f"{ctx.author.name}#{ctx.author.discriminator}"
```
MarkKoz: The `__str__` of `discord.Member` will already return the same thing.

bot/cogs/off_topic_names.py
```diff
@@ -96,14 +96,44 @@ def cog_unload(self) -> None:
...
+        # Chain multiple words to a single one
+        name = "-".join(names)
+
+        existing_names = await self.bot.api_client.get('bot/off-topic-channel-names')
+        close_match = difflib.get_close_matches(name, existing_names, n=1, cutoff=0.8)
+
+        if close_match:
+            match = close_match[0]
+            log.info(
+                f"{ctx.author.name}#{ctx.author.discriminator}"
```
MarkKoz: The `__str__` of `discord.Member` will already return the same thing.

jchristgit:
zoop, conflicts

PR #525, opened 2019-10-12
Move the `security` cog tests to `unittest`.

jchristgit:


tests/cogs/test_security.py
```diff
@@ -0,0 +1,59 @@
...
+
+class SecurityCogTests(unittest.TestCase):
+    """Tests the `Security` cog."""
+
+    def setUp(self):
+        """Attach an instance of the cog to the class for tests."""
+        self.bot = MagicMock()
+        self.cog = security.Security(self.bot)
+        self.ctx = MagicMock()
+        self.ctx.author = MagicMock()
```
SebastiaanZ: There are some mock helpers set-up for `Bot` and `Context` in [tests.helpers](/tests/helpers.py) that follow the specifications of the objects they're mocking (using the `spec` parameter of Mock/MagicMock) and have some commonly used attributes already set up (like the `author` attribute for the `Context` mock with a `Member` mock that follows the `discord.Member` spec and `send` for the `Context` mock with an `AsyncMock`).

jchristgit: see 612994a

PR #526, opened 2019-10-12
Move the `token_remover` cog tests to `unittest`.

jchristgit:


tests/cogs/test_token_remover.py
```diff
@@ -0,0 +1,139 @@
...
+from bot.constants import Channels, Colours, Event, Icons
+from tests.helpers import AsyncMock
+
+
+class TokenRemoverTests(unittest.TestCase):
+    """Tests the `TokenRemover` cog."""
+
+    def setUp(self):
+        """Adds the cog, a bot, and a message to the instance for usage in tests."""
+        self.bot = MagicMock()
```
MarkKoz: Should use `MockBot`

tests/cogs/test_token_remover.py
```diff
@@ -0,0 +1,139 @@
...
+        self.bot.get_cog.return_value = MagicMock()
+        self.bot.get_cog.return_value.send_log_message = AsyncMock()
+        self.cog = TokenRemover(bot=self.bot)
+
+        self.msg = MagicMock()
```
MarkKoz: Probably a good idea to create a mock helper for messages and use mocks with specs for some of the attributes here (like the author and channel).

tests/cogs/test_token_remover.py
```diff
@@ -0,0 +1,139 @@
...
+                    channel_id=Channels.mod_alerts
+                )
+
+
+class TokenRemoverSetupTests(unittest.TestCase):
+    """Tests setup of the `TokenRemover` cog."""
+
+    def test_setup(self):
+        """Setup of the cog should log a message at `INFO` level."""
+        bot = MagicMock()
```
MarkKoz: Should use MockBot

tests/cogs/test_token_remover.py
SebastiaanZ: That seems reasonable to me.

I think we should factor out the mocking of those Discord "Model" that we use a lot and typically require the same or a very similar set-up. Since it does mean we'd basically have to maintain and test a small `discord.py` mocking library, I'd say we should restrict it those those base models though. I think that with a set of Member, Guild, Role, Message, TextChannel, Context, Bot, and Role we've covered a lot of things that we typically usr.

SebastiaanZ: @MarkKoz @jchristgit I'm going to add a couple more specific mocks and tests, but I'll resolve this during the merge.

MarkKoz: Do this means this should be left alone for now?

SebastiaanZ:
Uhm, I messed up the git history... But, I've merged this correctly on the command line, so I'm just going to close this. Sorry for that. (Please don't hurt me.)

PR #527, opened 2019-10-12
Reduce `!free` output

### LINKED ISSUE #274: Restricting/redirecting output of commands to bot-commands for regular users
**Edit:** I've decided to make this a general issue, since the `!free` command turns out to be a bit disruptive as well. In python-discussions, there have just been four consecutive calls to `!free` and it looks very disruptive (see below).

My idea would be to create an easy way to redirect the output of certain commands, free and help, to bot commands for regular users, with the bot sending a temporary message to the user in channel redirecting them there. The bot could then delete that message (and, maybe, the invoking message as well) after a certain amount of time has passed.

**Four times !free in python-discussion:**

**Old message for context:**
Now the `!help` command works outside of bot-commands again, we may want to think about restricting the use/output to bot-commands for regular users. The help-embeds are quite sizeable and, therefore, quite disruptive in on-topic channels.

I want to propose to redirect the **output** of help (the help-embed) to bot-commands for non-staff members, prepended by a mention of the user invoking the command. In addition, we could display a small, non-disruptive embed in the original context channel redirecting the user to bot commands. I think this should also be the case for help-calls on specific commands, as the embeds for those can be quite sizeable as well (e.g., `!help site`).

Personally, I'd like this restriction to only apply to regular users, so staff can pull up help on a command quickly in channel and use help in the staff channels.

kraktus:
Closes #274 
current output:

What was proposed in the issue:

My proposal:
Screen Shot 2019-10-12 at 22 31 08

I left the list to make clearer the fact that the output is ordered and stressed the timer.


bot/cogs/free.py
```diff
@@ -72,30 +72,32 @@ class Free(Cog):
...
-
             # Sort channels in descending order by seconds
             # Get position in list, inactivity, and channel object
             # For each channel, add to embed.description
             sorted_channels = sorted(free_channels, key=itemgetter(0), reverse=True)
+            if len(sorted_channels) > 3:  # display 3 channels max
+                sorted_channels = sorted_channels[:3]
+
             for i, (inactive, channel) in enumerate(sorted_channels, 1):
                 minutes, seconds = divmod(inactive, 60)
```
mathsman5133: Just because I got a similar review comment, could this utilise the `utils.time.humanize_time` method, if you cast `inactive` to a timedelta?

bot/cogs/free.py
```diff
@@ -72,30 +72,32 @@ class Free(Cog):
...
-                    embed.description += f"{i}. {channel.mention} inactive for {hours}h{minutes}m{seconds}s\n\n"
+                    embed.description += f"{i}. {channel.mention} **{hours}h {minutes}m {seconds}s** inactive\n"
                 else:
-                    embed.description += f"{i}. {channel.mention} inactive for {minutes}m{seconds}s\n\n"
+                    embed.description += f"{i}. {channel.mention} **{minutes}m {seconds}s** inactive\n"
 
-            embed.description += ("**\nThese channels aren't guaranteed to be free, "
-                                  "so use your best judgement and check for yourself.")
+            embed.description += (
+                "Please confirm these channels "
```
mathsman5133: Could this be shortened to a single line?

bot/cogs/free.py
MarkKoz: Don't think so because it uses full names for units instead of single letters.

However, looking over the time-related code here, it could be improved. The delta object could be appended instead of just the seconds to avoid having to do the redundant math below to get the minutes and hours.

kraktus:
Screen Shot 2019-10-15 at 18 42 27

This is a loser screenshot. We can see that the channels's mention are not aligned, but I don't understand why, according to the code it should be the case.

bot/cogs/free.py
MarkKoz: To be clear, I acknowledge this is out of scope of the issue you are addressing but it would be a welcome addition if you wish to do it @kraktus.

bot/cogs/free.py
MarkKoz: @kraktus Can you please address this?

scragly:
I'm not a fan of the numbering, as it's not really useful (it has the times already present) and only serves to prevent alignment of the mentions, making it slightly uglier.

I think the footer should also still be an actual footer to keep it separated from the main output and to keep the text size slightly smaller.

lemonsaurus:
I agree with Scragly. I don't think your proposal is an improvement on the proposal in the original issue, so I'd like to see it changed to what the original proposal was. No numbers, text in footer.

bot/cogs/free.py
kraktus: I am sorry but I didn't quite understand how to simplify time-related, because time delta object does not have a hour attribute.

kraktus:
Screen Shot 2019-10-21 at 14 39 16

bot/cogs/free.py
MarkKoz: Oh, that's annoying. Well never mind then.

bot/cogs/free.py
```diff
@@ -72,30 +72,29 @@ class Free(Cog):
...
-                's' if len(free_channels) > 1 else '',
-                '' if len(free_channels) > 1 else 's'
-            )
-
             # Sort channels in descending order by seconds
             # Get position in list, inactivity, and channel object
             # For each channel, add to embed.description
             sorted_channels = sorted(free_channels, key=itemgetter(0), reverse=True)
-            for i, (inactive, channel) in enumerate(sorted_channels, 1):
+            if len(sorted_channels) > 3:  # display 3 channels max
```
MarkKoz: This check isn't necessary because when slicing, any end index greater than the end will just return the entire list. So it could be:

```py
for inactive, channel in sorted_channels[:3]:
    ...
```

bot/cogs/free.py
```diff
@@ -72,30 +72,27 @@ class Free(Cog):
...
-            embed.description += ("**\nThese channels aren't guaranteed to be free, "
-                                  "so use your best judgement and check for yourself.")
+            embed.set_footer(text="Please confirm these channels are free before posting")
         else:
-            embed.description = ("**Doesn't look like any channels are available right now. "
-                                 "You're welcome to check for yourself to be sure. "
-                                 "If all channels are truly busy, please be patient "
-                                 "as one will likely be available soon.**")
+            embed.description = (
+                "**Doesn't look like any channels are available right now. "
```
MarkKoz: I think it'd look better if it was not bold:

bild

bild

Issue #471, opened 2019-10-01
Automatic deletion of executable files


GhostofGoes:
Remove executable or potentially executable files that may be malicious (malware, spam, zip bombs, etc.). 

Per Lemon: "Staff meeting consensus is that we can start simple by whitelisting certain extensions and deleting everything else. If people try to get around this with malicious hybrid filetypes or whatever, we will discuss maybe using ClamAV to scan files, but for now, we'll keep it simple. This should just be another filter, like mentionspam."

Org issue: https://github.com/python-discord/organisation/issues/142

Implementation
* Implement rule method (similar one that exists is [`bot/rules/attachments`](https://github.com/python-discord/bot/blob/master/bot/rules/attachments.py)
* Determine an appropriate whitelist (can probably push something reasonable then hammer it out during a meeting or review)

bendiller:
Hi! Is anyone working on this? I'd like to, take care of it if it's open. I already threw together a quick prototype to make sure I understood how to implement this - I should be able to submit a PR pretty quickly.

lemonsaurus:
Nobody is working on this, feel free to have a go at it!

sco1:
Merged by #528

PR #528, opened 2019-10-12
Antimalware cog

bendiller:
Howdy! This PR is intended to address issue #471 (auto-deleting messages with potentially malicious file attachments). After discussing with @lemonsaurus, it became clear that a new Cog would make for a cleaner implementation, instead of a new rule within the `AntiSpam` Cog. I'll be happy to elaborate on those reasons if anyone would like further details.

Here's about the simplest implementation I could conceive, using the suggestions made in the issue's notes. Some points for further discussion:

1. The whitelist (list of allowed extensions) will need vetting - I used a fairly small list of common image and video file formats, and the Cog rejects everything else.
2. I'm fairly new to this code-base, so there may be some preferred implementation details and such that I neglected, either accidentally or otherwise. For example:
        - `AntiSpam.on_message()` begins with a handful of early-out conditions. I don't quite understand the purpose of some of those, so I did not include any, pending further review by staff.
        - `AntiSpam` calls `ModLog.ignore()` before deleting messages - I wasn't clear on why, so I did not do the same.
3. I did not implement punishment for attaching files with prohibited extensions, but I can do so if requested. It seems a little overly punitive perhaps, because I think most people attaching files will be doing so with good intent, but that's not my call.

I'll be happy to make any requested changes. Thanks!

### DIFF
--- a/bot/cogs/antimalware.py
+++ b/bot/cogs/antimalware.py
@@ -25,7 +25,6 @@ class AntiMalware(Cog):
                 break  # Other detections irrelevant because we prioritize the .py message.
             if not attachment.filename.lower().endswith(tuple(AntiMalwareConfig.whitelist)):
                 rejected_attachments = True
-                break
 
         if detected_pyfile or rejected_attachments:
             # Send a message to the user indicating the problem (with special treatment for .py)
--- a/bot/cogs/antimalware.py
+++ b/bot/cogs/antimalware.py
@@ -1,6 +1,6 @@
 import logging
 
-from discord import Message, utils
+from discord import Message, NotFound
 from discord.ext.commands import Bot, Cog
 
 from bot.constants import AntiMalware as AntiMalwareConfig, Channels
@@ -10,6 +10,7 @@ log = logging.getLogger(__name__)
 
 class AntiMalware(Cog):
     """Cog providing anti-malware behavior."""
+
     def __init__(self, bot: Bot):
         self.bot = bot
         self.whitelist = tuple(AntiMalwareConfig.whitelist)
@@ -17,27 +18,31 @@ class AntiMalware(Cog):
     @Cog.listener()
     async def on_message(self, message: Message) -> None:
         """Identify messages with prohibited attachments."""
-        log.trace("Entered AntiMalware.on_message()")
         rejected_attachments = [a for a in message.attachments if
                                 not a.filename.lower().endswith(self.whitelist)]
         detected_pyfile = len([a for a in message.attachments if a.filename.lower().endswith('.py')]) > 0
 
         if len(rejected_attachments) > 0:
-            log.trace("Identified rejected attachment(s)")
-            # Send a message indicating the problem to the user (with special treatment for .py)
+            # Send a message to the user indicating the problem (with special treatment for .py)
             author = message.author
             if detected_pyfile:
                 msg = f"{author.mention}, it looks like you tried to attach a Python file - please " \
                     f"use a code-pasting service such as https://paste.pythondiscord.com/ instead."
             else:
-                meta_channel = utils.get(message.guild.channels, id=Channels.meta)
+                meta_channel = self.bot.get_channel(Channels.meta)
                 msg = f"{author.mention}, it looks like you tried to attach a file type we don't " \
                     f"allow. Feel free to ask in {meta_channel.mention} if you think this is a mistake."
 
             await message.channel.send(msg)
 
+            # Delete the offending message:
+            try:
+                await message.delete()
+            except NotFound:
+                log.info(f"Tried to delete message `{message.id}`, but message could not be found.")
+
 
 def setup(bot: Bot) -> None:
-    """AntiMalware cog load."""
+    """Antimalware cog load."""
     bot.add_cog(AntiMalware(bot))
     log.info("Cog loaded: AntiMalware")
--- a/bot/cogs/antimalware.py
+++ b/bot/cogs/antimalware.py
@@ -20,13 +20,14 @@ class AntiMalware(Cog):
         rejected_attachments = False
         detected_pyfile = False
         for attachment in message.attachments:
-            if not attachment.filename.lower().endswith(tuple(AntiMalwareConfig.whitelist)):
-                rejected_attachments = True
             if attachment.filename.lower().endswith('.py'):
                 detected_pyfile = True
                 break  # Other detections irrelevant because we prioritize the .py message.
+            if not attachment.filename.lower().endswith(tuple(AntiMalwareConfig.whitelist)):
+                rejected_attachments = True
+                break
 
-        if rejected_attachments:
+        if detected_pyfile or rejected_attachments:
             # Send a message to the user indicating the problem (with special treatment for .py)
             author = message.author
             if detected_pyfile:
--- a/bot/cogs/antimalware.py
+++ b/bot/cogs/antimalware.py
@@ -1,0 +1,43 @@
+import logging
+
+from discord import Message, utils
+from discord.ext.commands import Bot, Cog
+
+from bot.constants import AntiMalware as AntiMalwareConfig, Channels
+
+log = logging.getLogger(__name__)
+
+
+class AntiMalware(Cog):
+    """Cog providing anti-malware behavior."""
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.whitelist = tuple(AntiMalwareConfig.whitelist)
+
+    @Cog.listener()
+    async def on_message(self, message: Message) -> None:
+        """Identify messages with prohibited attachments."""
+        log.trace("Entered AntiMalware.on_message()")
+        rejected_attachments = [a for a in message.attachments if
+                                not a.filename.lower().endswith(self.whitelist)]
+        detected_pyfile = len([a for a in message.attachments if a.filename.lower().endswith('.py')]) > 0
+
+        if len(rejected_attachments) > 0:
+            log.trace("Identified rejected attachment(s)")
+            # Send a message indicating the problem to the user (with special treatment for .py)
+            author = message.author
+            if detected_pyfile:
+                msg = f"{author.mention}, it looks like you tried to attach a Python file - please " \
+                    f"use a code-pasting service such as https://paste.pythondiscord.com/ instead."
+            else:
+                meta_channel = utils.get(message.guild.channels, id=Channels.meta)
+                msg = f"{author.mention}, it looks like you tried to attach a file type we don't " \
+                    f"allow. Feel free to ask in {meta_channel.mention} if you think this is a mistake."
+
+            await message.channel.send(msg)
+
+
+def setup(bot: Bot) -> None:
+    """AntiMalware cog load."""
+    bot.add_cog(AntiMalware(bot))
+    log.info("Cog loaded: AntiMalware")
--- a/bot/cogs/antimalware.py
+++ b/bot/cogs/antimalware.py
@@ -17,24 +17,29 @@ class AntiMalware(Cog):
     @Cog.listener()
     async def on_message(self, message: Message) -> None:
         """Identify messages with prohibited attachments."""
-        rejected_attachments = list()
-        detected_pyfile = list()
+        rejected_attachments = False
+        detected_pyfile = False
         for attachment in message.attachments:
             if not attachment.filename.lower().endswith(tuple(AntiMalwareConfig.whitelist)):
-                rejected_attachments.append(attachment)
+                rejected_attachments = True
             if attachment.filename.lower().endswith('.py'):
-                detected_pyfile.append(attachment)
+                detected_pyfile = True
+                break  # Other detections irrelevant because we prioritize the .py message.
 
         if rejected_attachments:
             # Send a message to the user indicating the problem (with special treatment for .py)
             author = message.author
             if detected_pyfile:
-                msg = (f"{author.mention}, it looks like you tried to attach a Python file - please "
-                       f"use a code-pasting service such as https://paste.pythondiscord.com/ instead.")
+                msg = (
+                    f"{author.mention}, it looks like you tried to attach a Python file - please "
+                    f"use a code-pasting service such as https://paste.pythondiscord.com/ instead."
+                )
             else:
                 meta_channel = self.bot.get_channel(Channels.meta)
-                msg = (f"{author.mention}, it looks like you tried to attach a file type we don't "
-                       f"allow. Feel free to ask in {meta_channel.mention} if you think this is a mistake.")
+                msg = (
+                    f"{author.mention}, it looks like you tried to attach a file type we don't "
+                    f"allow. Feel free to ask in {meta_channel.mention} if you think this is a mistake."
+                )
 
             await message.channel.send(msg)
 
--- a/bot/cogs/antimalware.py
+++ b/bot/cogs/antimalware.py
@@ -9,29 +9,32 @@ log = logging.getLogger(__name__)
 
 
 class AntiMalware(Cog):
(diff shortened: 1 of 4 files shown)


bot/cogs/antimalware.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @Cog.listener()
+    async def on_message(self, message: Message) -> None:
+        """Identify messages with prohibited attachments."""
+        log.trace("Entered AntiMalware.on_message()")
+        rejected_attachments = [a for a in message.attachments if
+                                not a.filename.lower().endswith(self.whitelist)]
+        detected_pyfile = len([a for a in message.attachments if a.filename.lower().endswith('.py')]) > 0
+
+        if len(rejected_attachments) > 0:
```
kraktus: You can use 
```
if rejected_attachments:
      ....
```
instead, because an empty list is `False`

bot/cogs/antimalware.py
```diff
@@ -0,0 +1,43 @@
...
+        self.bot = bot
+        self.whitelist = tuple(AntiMalwareConfig.whitelist)
+
+    @Cog.listener()
+    async def on_message(self, message: Message) -> None:
+        """Identify messages with prohibited attachments."""
+        log.trace("Entered AntiMalware.on_message()")
+        rejected_attachments = [a for a in message.attachments if
+                                not a.filename.lower().endswith(self.whitelist)]
+        detected_pyfile = len([a for a in message.attachments if a.filename.lower().endswith('.py')]) > 0
```
kraktus: There is some code redundancy here.  I would rather make only on `for`.
```
for attachement in message.attachments:
      #if python_file add to detected_pyfile

      #if not in whitelist add to rejected_attachments
```

Are you sure one message can have several attachements?

bot/cogs/antimalware.py
```diff
@@ -0,0 +1,43 @@
...
+
+        if len(rejected_attachments) > 0:
+            log.trace("Identified rejected attachment(s)")
+            # Send a message indicating the problem to the user (with special treatment for .py)
+            author = message.author
+            if detected_pyfile:
+                msg = f"{author.mention}, it looks like you tried to attach a Python file - please " \
+                    f"use a code-pasting service such as https://paste.pythondiscord.com/ instead."
+            else:
```
kraktus: if there can be several attachements, why use a `if else` structure?

bot/cogs/antimalware.py
bendiller: Good point.

bot/cogs/antimalware.py
bendiller: Testing seems to suggest that you're correct, they cannot. I misunderstood. Thanks!

bendiller:
Good feedback @kraktus, thank you! As for the message deletion, that begins on line 39, directly above the `def setup()...` function.

bot/cogs/antimalware.py
bendiller: I didn't want to clutter things up with too much response text, and deemed the pastebin advice to be the more important of the two. Just a personal judgment call, I'll be happy to revise if requested by staff.

kraktus:
> Good feedback @kraktus, thank you! As for the message deletion, that begins on line 39, directly above the `def setup()...` function.

Yep I did not see it because I reviewed the first commit :)

bot/cogs/antimalware.py
MarkKoz: Messages can definitely have multiple attachments. Though the suggestion of using a single for loop is still a good suggestion.

bot/cogs/antimalware.py
```diff
@@ -0,0 +1,48 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class AntiMalware(Cog):
+    """Cog providing anti-malware behavior."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.whitelist = tuple(AntiMalwareConfig.whitelist)
```
MarkKoz: Why is this done? As an alias?

config-default.yml
```diff
@@ -322,6 +323,11 @@ anti_spam:
             max: 3
 
 
+anti_malware:
+    whitelist: ['.bmp', '.gif', '.jpg', '.jpeg', '.png', '.tiff', # Images
```
MarkKoz: I think the [more conventional syntax](https://yaml.org/spec/1.2/spec.html#id2759963) for lists should be used as it's more consistent with the rest of the config.

config-default.yml
```diff
@@ -107,6 +107,7 @@ guild:
         help_7:                           587375768556797982
         helpers:                          385474242440986624
         message_log:       &MESSAGE_LOG   467752170159079424
+        meta:                             429409067623251969
```
MarkKoz: This wasn't defined in `constants.py`.

bot/cogs/antimalware.py
```diff
@@ -0,0 +1,48 @@
...
+        rejected_attachments = [a for a in message.attachments if
+                                not a.filename.lower().endswith(self.whitelist)]
+        detected_pyfile = len([a for a in message.attachments if a.filename.lower().endswith('.py')]) > 0
+
+        if len(rejected_attachments) > 0:
+            # Send a message to the user indicating the problem (with special treatment for .py)
+            author = message.author
+            if detected_pyfile:
+                msg = f"{author.mention}, it looks like you tried to attach a Python file - please " \
+                    f"use a code-pasting service such as https://paste.pythondiscord.com/ instead."
```
MarkKoz: It's preferred to encapsulate the string in parenthesis over using `\`.

bot/cogs/antimalware.py
```diff
@@ -0,0 +1,48 @@
...
+            else:
+                meta_channel = self.bot.get_channel(Channels.meta)
+                msg = f"{author.mention}, it looks like you tried to attach a file type we don't " \
+                    f"allow. Feel free to ask in {meta_channel.mention} if you think this is a mistake."
```
MarkKoz: It's preferred to encapsulate the string in parenthesis over using `\`.

bot/cogs/antimalware.py
```diff
@@ -0,0 +1,48 @@
...
+from discord import Message, NotFound
+from discord.ext.commands import Bot, Cog
+
+from bot.constants import AntiMalware as AntiMalwareConfig, Channels
+
+log = logging.getLogger(__name__)
+
+
+class AntiMalware(Cog):
+    """Cog providing anti-malware behavior."""
```
MarkKoz: Original wasn't very helpful. It's pretty obvious it's a cog and we can already see its name is "AntiMalware".

```suggestion
    """Delete messages which contain attachments with non-whitelisted file extensions."""
```

bot/cogs/antimalware.py
bendiller: Yes, to clean up its use later on - `a.filename.lower().endswith(tuple(AntiMalwareConfig.whitelist))` just looked ugly and unclear to me. Also I think I originally intended to use the whitelist more than once for some reason.

What do you prefer?

bot/cogs/antimalware.py
bendiller: Okay. Can you help me understand why? Not trying to be argumentative, just trying to improve. Is the single loop preferred because iterating twice is simply more expensive? I considered that but thought the readability was better in this case, and that the size of the lists being so small would make it a trivial performance hit. 

Is this more like what you'd want to see instead?
```python
rejected_attachments = list()
detected_pyfile = list()
for attachment in message.attachments:
    if not attachment.filename.lower().endswith(self.whitelist):
        rejected_attachments.append(attachment)
    if attachment.filename.lower().endswith('.py'):
        detected_pyfile.append(attachment)
```

lemonsaurus: well yeah, it's twice as expensive if you iterate the same list twice. But I also think it has better readability when you avoid the list comprehensions. Consider for example this line:

```py
len([a for a in message.attachments if a.filename.lower().endswith('.py')]) > 0
```

You'd read this sequentially by saying to yourself "okay, so it's the length of, uh, a list that for every attachment in message.attachment but only if the filename ends with py, a list of items that match this condition, if this list is greater than zero."

Then let's look at the other example (simplified):

```py
for attachment in message.attachments:
    if attachment.filename.lower().endswith('.py'):
        detected_pyfile.append(attachment)
```

For every attachment in message.attachments, if the attachment filename endswith .py, then we append the attachment to this list. It's succinct. It reads almost like an english sentence.

As a rule of thumb, try to keep comprehensions very simple. once you have a comprehension that is complex, just go ahead and refactor it into a loop. And in this case, we have both comprehension complexity, AND a needless second iteration of the same container! it's a no-brainer to refactor.

Hope that's helpful.

bot/cogs/antimalware.py
MarkKoz: Since we're suggesting using a normal for loop, I don't see this alias as useful any more.

bot/cogs/antimalware.py
```diff
@@ -9,29 +9,32 @@
...
         """Identify messages with prohibited attachments."""
-        rejected_attachments = [a for a in message.attachments if
-                                not a.filename.lower().endswith(self.whitelist)]
-        detected_pyfile = len([a for a in message.attachments if a.filename.lower().endswith('.py')]) > 0
-
-        if len(rejected_attachments) > 0:
+        rejected_attachments = list()
+        detected_pyfile = list()
+        for attachment in message.attachments:
+            if not attachment.filename.lower().endswith(tuple(AntiMalwareConfig.whitelist)):
```
MarkKoz: Does it really have to be a tuple or can it just be any sequence? Presumably it is was list before you converted it to a tuple.

bot/cogs/antimalware.py
```diff
@@ -9,29 +9,32 @@
...
     @Cog.listener()
     async def on_message(self, message: Message) -> None:
         """Identify messages with prohibited attachments."""
-        rejected_attachments = [a for a in message.attachments if
-                                not a.filename.lower().endswith(self.whitelist)]
-        detected_pyfile = len([a for a in message.attachments if a.filename.lower().endswith('.py')]) > 0
-
-        if len(rejected_attachments) > 0:
+        rejected_attachments = list()
+        detected_pyfile = list()
```
MarkKoz: It would make more sense for this to be a boolean.

bot/cogs/antimalware.py
```diff
@@ -9,29 +9,32 @@
...
 
     @Cog.listener()
     async def on_message(self, message: Message) -> None:
         """Identify messages with prohibited attachments."""
-        rejected_attachments = [a for a in message.attachments if
-                                not a.filename.lower().endswith(self.whitelist)]
-        detected_pyfile = len([a for a in message.attachments if a.filename.lower().endswith('.py')]) > 0
-
-        if len(rejected_attachments) > 0:
+        rejected_attachments = list()
```
MarkKoz: It would make more sense for this to be a boolean.

bot/cogs/antimalware.py
```diff
@@ -9,29 +9,32 @@
...
+        detected_pyfile = list()
+        for attachment in message.attachments:
```
MarkKoz: If **both** a rejected and a py file are found then you can break out of this loop.

bot/cogs/antimalware.py
```diff
@@ -9,29 +9,32 @@
...
+                detected_pyfile.append(attachment)
+
+        if rejected_attachments:
             # Send a message to the user indicating the problem (with special treatment for .py)
             author = message.author
             if detected_pyfile:
-                msg = f"{author.mention}, it looks like you tried to attach a Python file - please " \
-                    f"use a code-pasting service such as https://paste.pythondiscord.com/ instead."
+                msg = (f"{author.mention}, it looks like you tried to attach a Python file - please "
+                       f"use a code-pasting service such as https://paste.pythondiscord.com/ instead.")
```
MarkKoz: The parenthesis should have newlines like so:

```py
msg = (
    "..."
)
```

bot/cogs/antimalware.py
```diff
@@ -9,29 +9,32 @@
...
             else:
                 meta_channel = self.bot.get_channel(Channels.meta)
-                msg = f"{author.mention}, it looks like you tried to attach a file type we don't " \
-                    f"allow. Feel free to ask in {meta_channel.mention} if you think this is a mistake."
+                msg = (f"{author.mention}, it looks like you tried to attach a file type we don't "
+                       f"allow. Feel free to ask in {meta_channel.mention} if you think this is a mistake.")
```
MarkKoz: The parenthesis should have newlines as shown before.

bot/cogs/antimalware.py
MarkKoz: I agree with Ben.

bot/cogs/antimalware.py
bendiller: Hmm, I think I see what you're saying. If that's desired, couldn't it be slightly simpler - if a .py file is found, it can break, since we are prioritizing the message sent for .py files?

bot/cogs/antimalware.py
bendiller: Oddly enough, it does have to be a tuple or str: https://docs.python.org/3/library/stdtypes.html?highlight=endswith#str.endswith

bot/cogs/antimalware.py
MarkKoz: Yeah, that is a good point. It'd make more sense to have the `.py` check come first. But if another extension is found, it shouldn't break because maybe another attachment later will be a `.py` file.

bot/cogs/antimalware.py
```diff
@@ -17,24 +17,29 @@ def __init__(self, bot: Bot):
...
         """Identify messages with prohibited attachments."""
-        rejected_attachments = list()
-        detected_pyfile = list()
+        rejected_attachments = False
+        detected_pyfile = False
         for attachment in message.attachments:
             if not attachment.filename.lower().endswith(tuple(AntiMalwareConfig.whitelist)):
-                rejected_attachments.append(attachment)
+                rejected_attachments = True
             if attachment.filename.lower().endswith('.py'):
```
MarkKoz: So as I mentioned, this check should come first to make things a bit more efficient. You'd also have to rework the if statements to be like:

`` py
if detected_pyfile:
    ...
elif rejected_attachments:
    ...
```

Comment (bendiller) [reply bot/cogs/antimalware.py]: I'm not sure how to implement it quite the way you've shown without either violating DRY or adding another conditional:
```python
if detected_pyfile or rejected_attachments:
    # Send warning message and delete offending message, etc.
```
but I have implemented the re-ordering as you've requested to skip the one extra conditional evaluation in the cases where a .py file is attached. I hope I understood you correctly.

MarkKoz: That isn't what I meant. Sorry for not being clear. Pseudocode: 

```py
for msg in attachments:
    if py_file:
        detected_pyfile = True
        break
    if not whitelisted:
        rejected_attachments = True
        # Don't break! Since the py message takes priority,
        # we should keep checking attachments to see if maybe
        # a later attachment will be a py file.

if detected_pyfile:
    print("Please use paste service..")
elif rejected_attachments:
    print("File ext not allowed")
else:
    # Do nothing; all attachments sent were whitelisted
    # this else can be omitted but I have shown it here for explanatory purposes
    pass
```

bendiller: Okay, good point, I'll get the second `break` out of there.

MarkKoz: The conditions for sending out the message still needs to be addressed.

MarkKoz: Never mind actually, it's fine too. It's just a different way of doing what I had in mind and I didn't recognise that initially.

PR #529, opened 2019-10-13
Display word triggered by watchlist in #mod-alerts

### LINKED ISSUE #374: Display word triggered by watchlist in #mod-alerts
## Abstract

In `#mod-alerts` when a word is triggered in the `watchlist`, display the word that was triggered in addition to the original message content.

## Rationale

Without knowing the trigger word, reading the original content of the message to figure out what it was is unnecessarily difficult. In order to see if moderation action is required, we have to first find what the word was and where it is in the message, before finally being able to determine what context it was used in.

## Motivation

* Often a trigger will come from a very long message, making it very time consuming to find the instance of the word.
* If the word was triggered in a codeblock, the word is often a substring of some other larger unrelated word, making finding the trigger incredibly difficult

## Specification

Alongside the meta info of who sent the message and in what channel, show what the trigger word was, and if possible, a short snippet of the part of the message that caused the trigger.

**Example:**

```
The watch_tokens watchlist was triggered by lemon#0001 (95872159741644800)
in #admin-spam

Trigger word: "nigg"
Match: "...word niggardly is c..."

Original message:

The word niggardly is controversial due to its phonetic similarity
to the racial slur, but etymologically they're completely unrelated!
```

kosayoda:
**Closes #374**

This pull request adds the word triggered from the watchlist and a short snippet of the message that contains the trigger to the embed sent to `#mod-alerts` whenever the filter catches a naughty message.

Rationale from the issue:
> Without knowing the trigger word, reading the original content of the message to figure out what it was is unnecessarily difficult. In order to see if moderation action is required, we have to first find what the word was and where it is in the message, before finally being able to determine what context it was used in.

Implemented example:
image


bot/cogs/filtering.py
```diff
@@ -237,8 +250,9 @@ def mod_log(self) -> ModLog:
         Only matches words with boundaries before and after the expression.
         """
         for regex_pattern in WORD_WATCHLIST_PATTERNS:
-            if regex_pattern.search(text):
-                return True
+            match = regex_pattern.search(text)
+            if match:
+                return match  # match objects always have a boolean value of True
```
MarkKoz: How about `return match[0]` here? Would have to explicitly check  `triggered is not False` though (perhaps better to return `None` instead of `False` in that case).

bot/cogs/filtering.py
```diff
@@ -184,12 +184,25 @@ def mod_log(self) -> ModLog:
                         else:
                             channel_str = f"in {msg.channel.mention}"
 
+                        # Word and match stats for watch_words and watch_tokens
+                        if filter_name in ("watch_words", "watch_tokens"):
+                            m = triggered
+                            match = m[0]
```
MarkKoz: This is a bit weird cause it could just be `match = triggered[0]`. Maybe you did it that way because the name `triggered` is misleading. In that case, it could just be renamed to something like `match` anyway.

bot/cogs/filtering.py
kosayoda: I return the match because I still need to access the the match start and end positions to get the `Location:` information, here:
`surroundings = m.string[max(m.start() - 10, 0): m.end() + 10]`

bot/cogs/filtering.py
kosayoda: I do suppose I can rename `triggered` to `match`, since they're functionally the same thing and would provide clarity in certain areas. What do you think?

MarkKoz: Yes, that sounds good.

bot/cogs/filtering.py
```diff
@@ -250,11 +262,12 @@ def mod_log(self) -> ModLog:
...
         """
         for regex_pattern in TOKEN_WATCHLIST_PATTERNS:
-            if regex_pattern.search(text):
+            match = regex_pattern.search(text)
+            if match:
 
                 # Make sure it's not a URL
                 if not URL_RE.search(text):
-                    return True
+                    return match  # match objects always have a boolean value of True
```
scragly: It might resolve to bool(match) == True, but it's not returning that so the type annotation is incorrect.

PR #530, opened 2019-10-13
Strip spoiler tags for watchlist triggers

### LINKED ISSUE #299: Strip spoiler tags for watchlist triggers
With the [recent Spoiler Tag addition](https://support.discordapp.com/hc/en-us/articles/360022320632-Spoiler-Tags-) to Discord, it's now possible to take memes to a whole new level. While the message filters are still operable, it's difficult to read the trigger messages in the mod log, particularly on mobile.

Spoilers are wrapped with `||` (e.g. `||text||`). Let's add a helper method to the modlog cog to strip these from the message content before they are sent.

AnonGuy:
Closes #299


bot/cogs/filtering.py
```diff
@@ -26,6 +26,7 @@
     flags=re.IGNORECASE
 )
 
+SPOILER_RE = re.compile(r"(\|\|.+?\|\|)")
```
eivl: .+? matches any character (except for line terminators)

but discord allows you to use new-line in spoiler tags.

AnonGuy: Thanks for the heads up, I've updated the pattern to support multi-line spoilers.

scragly:
I played around and wrote a possible method to split things up into "interpretations" of text messed up with spoilers:

```py
SPOILER_RE = re.compile(r"\|\|", re.DOTALL)

def process_spoilers(text):
    split_text = SPOILER_RE.split(text)
    public_text = "".join(split_text[0:][::2])
    spoiler_text = "".join(split_text[1:][::2])
    cleaned_text =  "".join(split_text)
    return public_text, spoiler_text, cleaned_text
```

AnonGuy:
Thanks @scragly, I'll add it in

lemonsaurus:
@AnonGuy what's the status on this? is it done? it's still labeled WIP.

lemonsaurus:
@AnonGuy Looks like this is a good start, but there are multiple cases that this PR does not resolve, so it feels a bit like a half measure.

This PR does catch the simplest, `||badword||`, but beyond that it does not catch most of the stuff discussed in this PR and in the issue this was based on.

For example, have a look at this embed:

image

Sure, the filter does match it, but the username is spoilered, the location has spoilers, and the original message has spoilers.

It would be better her to display these with the pipes escaped, for example.

It also does not catch the cases that @eivl touched on earlier here, stuff like `n||B||o||A||t||D||h||W||i||O||n||R||g||D||`.

Would you be able to expand this solution to solve all our spoiler-related issues, @AnonGuy?

bot/cogs/filtering.py
```diff
@@ -27,6 +27,7 @@
     flags=re.IGNORECASE
 )
 
+SPOILER_RE = re.compile(r"(\|\|.+?\|\|)", re.DOTALL)
```
lemonsaurus: Okay, I got two issues with this:
1) It's a bit strict. For example, if there are only pipes before but not after the word, this will break. If the words are wrapped in single pipes instead of doubles, this will break. Perhaps it would be better to get rid of _an arbitrary amount of pipes_, or to just get rid of all the pipes in the entire word before filtering.

2) Do we really need regex for just getting rid of pipes? maybe we should solve this with just content.replace or something.

AnonGuy: This pattern is used for expanding a message that may contain a spoilered word, like what gdude mentioned in the issue. For actually getting rid of pipes in watchlist triggers, etc. I'm just using [`discord.utils.escape_markdown`](https://discordpy.readthedocs.io/en/latest/api.html#discord.utils.escape_markdown)

PR #541, opened 2019-10-17
Fix defcon having wrong text when disabling.

### LINKED ISSUE #539: Defcon have wrong text in #mod-log when you disable it.
When you disable defcon the text message in #mod-log is `DEFCON enabled` it should say `DEFCON disabled`

It works like this in the #defcon channel.

Current message in #mod-log
defcon-disabled

but a few months ago you can see this behaviour.
Message from 08.08.2019 in #mod-log
defcon-working

code from [defcon.py](https://github.com/python-discord/bot/blob/master/bot/cogs/defcon.py)

```python
        elif change.lower() == "disabled":
            icon = Icons.defcon_disabled
            color = Colours.soft_red
            status_msg = "DEFCON enabled"
```

Someone looking into this might want to focus on the last line of this snippet.

ikuyarihS:
#### Closes #539

This pull request fixes the issue for wrong status in the embed, as well as unifying certains if else to prevent similar future errors.

Most notable code is here:

```python
log_msg = f"**Staffer:** {actor} (`{actor.id}`)\n"

if change.lower() == "enabled":
    icon = Icons.defcon_enabled
    color = Colours.soft_green
    status_msg = "DEFCON enabled"
    log_msg += f"**Days:** {self.days.days}\n\n"
elif change.lower() == "disabled":
    icon = Icons.defcon_disabled
    color = Colours.soft_red
    status_msg = "DEFCON enabled"
elif change.lower() == "updated":
    icon = Icons.defcon_updated
    color = Colour.blurple()
    status_msg = "DEFCON updated"
    log_msg += f"**Days:** {self.days.days}\n\n"
```

To remedy this issue, the class will now have a ~~private class variable~~ `Enum` ( Thanks to @MarkKoz for the suggestions )

```py
class Action(Enum):
    """Defcon Action."""

    ActionInfo = namedtuple('LogInfoDetails', ['icon', 'color', 'template'])

    ENABLED = ActionInfo(Icons.defcon_enabled, Colours.soft_green, "**Days:** {days}\n\n")
    DISABLED = ActionInfo(Icons.defcon_disabled, Colours.soft_red, "")
    UPDATED = ActionInfo(Icons.defcon_updated, Colour.blurple(), "**Days:** {days}\n\n")
```

Another big change is the introduction of `_defcon_action` to unify the process all defcon actions has to go through:
- Try to do the action.
- Log error if there is any.
- Send log message.


bot/cogs/defcon.py
```diff
@@ -107,6 +115,18 @@ def mod_log(self) -> ModLog:
         """Check the DEFCON status or run a subcommand."""
         await ctx.invoke(self.bot.get_command("help"), "defcon")
 
+    async def _defcon_action(self, ctx: Context, action: Awaitable[None], action_type: str) -> None:
```
MarkKoz: Given that the action is always a PUT request, this may as well just be done here and take parameters for the days and enable/disable. 

Perhaps then you could also get rid of `action_type` and deduce it from the aforementioned arguments. Going with the assumption that the days would never be updated to 0, could "updated" be when enabled = `True` and days = `0`? An alternative is to keep `action_type` and not have an `enabled` parameter since I think it's easy to deduce that value from the value of `action_type`.

bot/cogs/defcon.py
```diff
@@ -30,6 +31,12 @@ class Defcon(Cog):
     days = None  # type: timedelta
     enabled = False  # type: bool
 
+    _defcon_log_info: Dict[str, Tuple] = {
```
MarkKoz: I think an enum would be nicer if this is staying (see other comment).

bot/cogs/defcon.py
ikuyarihS: That's some very good suggestions, now that I inspected the actions closer, it is very possible to move all the puts there, and to accept only days and the bool for `'enabled''. Thank you.

bot/cogs/defcon.py
```diff
@@ -115,11 +123,26 @@ def mod_log(self) -> ModLog:
         """Check the DEFCON status or run a subcommand."""
         await ctx.invoke(self.bot.get_command("help"), "defcon")
 
-    async def _defcon_action(self, ctx: Context, action: Awaitable[None], action_type: str) -> None:
-        """Providing a structured way to do an defcon actions - enabled / disabled / updated."""
+    async def _defcon_action(self, ctx: Context, days: int, enabled: bool) -> None:
```
MarkKoz: Hmm... having second thoughts on the action type inference. I feel like since either way a second parameter (`enabled` boolean at the moment) is needed in addition to `days`, that second parameter may as well just be the enum rather than a boolean to avoid the inference. What do you think?

bot/cogs/defcon.py
```diff
@@ -25,18 +26,25 @@
 BASE_CHANNEL_TOPIC = "Python Discord Defense Mechanism"
 
 
+class DefconLogInfo(Enum):
+    """Defcon Logging Information."""
```
MarkKoz: Since the values are just tuples, I think it is worth mentioning what each element of the tuple is for. An alternative would be to use namedtuples but perhaps that's overkill.

bot/cogs/defcon.py
```diff
@@ -25,18 +26,25 @@
 BASE_CHANNEL_TOPIC = "Python Discord Defense Mechanism"
 
 
+class DefconLogInfo(Enum):
```
MarkKoz: Naming it `Action` would be nicer, or _at the least_ just `LogInfo` (but really the former is a lot better IMO).

bot/cogs/defcon.py
ikuyarihS: Hmm, I think that's a good idea, it is better to be explicit and pass in an `Action.Enabled` / `Action.Disabled` or `Action.Updated` instead of simple boolean, so an Enum will do wonder here.

bot/cogs/defcon.py
ikuyarihS: That's a great idea! After inspecting closely, I've renamed it to `Action` and use it not for `send_log_info()` but for `_defcon_action()` and other actions as well, it should be more consistently now.

bot/cogs/defcon.py
```diff
@@ -139,16 +138,16 @@ def mod_log(self) -> ModLog:
...
                 }
             )
         except Exception as err:
             log.exception("Unable to update DEFCON settings.")
             error = err
         finally:
-            await ctx.send(self.build_defcon_msg(action_type, error))
-            await self.send_defcon_log(action_type, ctx.author, error)
+            await ctx.send(self.build_defcon_msg(action.name.lower(), error))
+            await self.send_defcon_log(action.name.lower(), ctx.author, error)
```
MarkKoz: Why not just pass the enum object instead of using its name? Then you won't even need the `get_info` for anything.

ikuyarihS: That's very true lol, I still am too tunnel vision into what each functions are doing. That's a great call!

bot/cogs/defcon.py
```diff
@@ -226,18 +221,18 @@ def build_defcon_msg(self, change: str, e: Exception = None) -> str:
 
         return msg
 
-    async def send_defcon_log(self, change: str, actor: Member, e: Exception = None) -> None:
+    async def send_defcon_log(self, action: Action, actor: Member, e: Exception = None) -> None:
         """
         Send log message for DEFCON action.
 
         `change` string may be one of the following: ('enabled', 'disabled', 'updated')
```
MarkKoz: This docstring is no longer accurate. It can probably just be removed altogether.

bot/cogs/defcon.py
```diff
@@ -202,17 +197,17 @@ def mod_log(self) -> ModLog:
         self.mod_log.ignore(Event.guild_channel_update, Channels.defcon)
         await self.channel.edit(topic=new_topic)
 
-    def build_defcon_msg(self, change: str, e: Exception = None) -> str:
+    def build_defcon_msg(self, action: Action, e: Exception = None) -> str:
         """
         Build in-channel response string for DEFCON action.
 
         `change` string may be one of the following: ('enabled', 'disabled', 'updated')
```
MarkKoz: This docstring is no longer accurate. It can probably just be removed altogether.

bot/cogs/defcon.py
ikuyarihS: I'll update their docstring accordingly.

MarkKoz: Do you think it's valuable to still keep it? I feel that since it's an enum and the argument is type hinted, one can just look at the definition of the enum to see what the values can be. Before it was just `str` and that was understandably vague so this docstring made sense at the time.

ikuyarihS: At the time, yes, since it can be any string, it is valuable to know what specific string one needs to pass into the functions. Now via type hinting I do not think we'll need to keep this anymore, we would only have one `Action` Enum now, if anything we can just add more to it like an Interface.

bot/cogs/defcon.py
```diff
@@ -223,17 +197,13 @@ def mod_log(self) -> ModLog:
...
+        """Build in-channel response string for DEFCON action."""
+        if action is Action.ENABLED:
             msg = f"{Emojis.defcon_enabled} DEFCON enabled.\n\n"
-        elif change.lower() == "disabled":
+        elif action is Action.DISABLED:
             msg = f"{Emojis.defcon_disabled} DEFCON disabled.\n\n"
-        elif change.lower() == "updated":
+        elif action is Action.UPDATED:
             msg = (
                 f"{Emojis.defcon_updated} DEFCON days updated; accounts must be {self.days} "
```
MarkKoz: Fixes a small issue with the formatting. It was displaying the hours, minutes, and seconds as well.
```suggestion
                 f"{Emojis.defcon_updated} DEFCON days updated; accounts must be {self.days.days} "
```

If you feel like it you could also make the message show singular "day" when it's 1 day.

ikuyarihS: That's a nice spot. I'll also make it a QoL change now and make it `day` when it's 1 day.

Issue #388, opened 2019-07-28
Add Loose Resource Files for Tags


MarkKoz:
Actioning python-discord/organisation#117. 

The files could just be markdown but if we want more customisability of embeds we can use a format like JSON and feed it into [`discord.Embed.from_dict`](https://discordpy.readthedocs.io/en/latest/api.html#discord.Embed.from_dict). 

We'd need to keep in mind that Discord only supports a subset of markdown so a GitHub render of a markdown file will not necessarily be representative of how it'll appear on Discord. [Embed Visualizer](https://leovoel.github.io/embed-visualizer/) can be used for previewing tags.

The existing tags will need to be imported into this new format from our database.

scragly:
Tags are now part of the site API, and are editible by mods+ in the admin panel.

For now, we've added a tag directory with tag contents in individual markdown files within the meta repository:
https://github.com/python-discord/meta/tree/master/tags

For now, it's a manual process of adding or editing tags via bot or site admin whenever we get a PR merged in, however in the future this could possibly be more automated.

Due to the change in situation, I'll close this ticket.

MarkKoz:
It was decided in a meeting that we want to go with this approach.
Therefore, I think that if this is going to be closed, there should be a
consensus to reverse the decision to do this. And, to be clear, when this
was decided, the tags view already existed on the API.

Den fre 15 nov. 2019 06:28scragly <notifications@github.com> skrev:

scragly:
Very well, I'll open an org issue for it to be discussed in the next staff meeting.

MarkKoz:
Closing in favour of lucky number #777

PR #542, opened 2019-10-18
Make it easier for user to search for tags

### LINKED ISSUE #231: Tag fuzzy matching and aliasing
As discussed briefly in #118 and at least a few times on the server in the recent past, I believe tags could stand to gain from fuzzy matching and possibly from aliasing as well.

In short:

- Fuzzy Matching
  - Allows for solving of the plurals problem, like `codeblock` vs `codeblocks` and `f-string` vs `f-strings`
  - Potentially solves things like `fstring` vs `f-string`
  - Embed should indicate to the user that a fuzzy match was used (e.g. `"Did you mean ... ?"`)

- Aliasing (See #187)
  - Requires adding a field to the backend tag model
  - Allows for things that fuzzy matching may not realistically be able to hit, like `ytdl` vs `youtube-dl`

ikuyarihS:
#### Closes #231

Applying the algorithm for `Needles and Haystack` to find and match tag in tags, for example:

Example

This only applies to searching tag_name with more than 3 in length, and at least 80% of its letters are found, from left to right.

There are 3 levels of checking, stop at first found:
- Check if exact name ( case insensitive ) O(1) getting from a dictionary Dict[str, Tag]
- Check for all tags that has 100% matching via algorithm
- Check for all tags that has >= 80% matching

If there are more than one hit, it will be shown as suggestions:

Suggestions

In order to avoid api being called multiple times, I've implemented a cache to only refresh itself when the is a gap of more than 5 minutes from the last api call to get all tags.

Editing / Adding / Deleting tags will also modify the cache directly.

### What about other solution like fuzzywuzzy?

fuzzywuzzy was considered for using, but from testing, it was giving much lower scores than expected:

Code used to test:

```py
from fuzzywuzzy import fuzz

def _fuzzy_search(search: str, target: str) -> bool:
    found = 0
    index = 0
    _search = search.lower().replace(' ', '')
    _target = target.lower().replace(' ', '')
    for letter in _search:
        index = _target.find(letter, index)
        if index == -1:
            break
        found += index > 0
    # return found / len(_search) * 100
    return (
        found / len(_search) * 100,
        fuzz.ratio(search, target),
        fuzz.partial_ratio(search, target)
    )

tests = (
    'this-is-gonna-be-fun',
    'this-too-will-be-fun'
)

for test in tests:
    print(test, '->', _fuzzy_search('this too fun', test))
```

Result from test:
```py
this-is-gonna-be-fun -> (30.0, 50, 50)
this-too-will-be-fun -> (90.0, 62, 58)
```


kosayoda:
Looking at the fuzzy search, have you considered the built-in `difflib` module?

MarkKoz:
We actually discussed if it'd be better to do the fuzzy search server-side on the API. I haven't looked into it deeply but here are some relevant links:

https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/search/
https://github.com/vsemionov/django-rest-fuzzysearch

I'm not sure if it'd be better to do it server or client side. I think that if there is room for fuzzy search to be used in the future with other endpoints (new or existing), then it should be server side. Another factor would be to see how accurate the pg search features are for our needs here.

ikuyarihS:
> Looking at the fuzzy search, have you considered the built-in `difflib` module?

I've taken a look at it, it proves to be quite useful to get the differences in `before` and `after` in the `on_message_edit` event. I've looked at its SequenceMatcher, it provides similar result to fuzzywuzzy

```py
...
s = difflib.SequenceMatcher(lambda x: x in ' -', search, target)
return (
    found / len(_search) * 100,
    ('fuzzy', fuzz.ratio(search, target), fuzz.partial_ratio(search, target)),
    ('difflib', tuple(map(lambda x: x * 100, (s.ratio(), s.real_quick_ratio(), s.quick_ratio()))))
)

# --------------------------------
this too fun & this-is-gonna-be-fun -> (30.0, ('fuzzy', 50, 50), ('difflib', (50.0, 75.0, 50.0)))
this too fun & this-too-will-be-fun -> (90.0, ('fuzzy', 62, 58), ('difflib', (62.5, 75.0, 62.5)))
```

I've thought about either this should be done from the API or from the bot, I think having a cache on the bot will give better performance, specially if we do not modify tags from the site-side and restrict modifying tags to be via bot's commands only, then we can maintain a cache that's perfectly synced with the site.

Postgres search feature looks powerful too, I'll definitely want to see how it performs as well.

SebastiaanZ:
Did we decide on an approach for this? Bot-side or server-side? I kinda like the idea of shipping a query off to the API and having postgres do its thing.

MarkKoz:
Given #388 it's better to keep it client-side as it would eventually have to be client-side anyway. However, that issue is stale so I don't know if we still want to do that. If not, then I agree with doing it on the server-side.

SebastiaanZ:
That's a good point; I'd forgotten about that. Let's ask our tag master, @fiskenslakt, what his current opinion on the matter is to make sure we get this thing moving again.

scragly:
I closed #388 and the meta repo contains markdown files of any of our tags for now  for the public to read through and be able to submit PRs for adding or editing tags. 

At the moment the process of adding or editing the tags is done via bot command in-server or by using the site's admin page (mods+ now have full access to the tag admin page).

There's improvements that can be done to make things easier and to automate/integrate the process, but I'm of the opinion that tags will continue to live on the database, be accessible via API and editable via web admin, and as such we should probably stick to doing fuzzymatching api-side.

bot/cogs/tags.py
```diff
@@ -27,6 +26,44 @@ def __init__(self, bot: Bot):
...
+
+    async def _get_tags(self, is_forced: bool = False) -> None:
+        """Getting all tags."""
+        # Refresh only when there's a more than 5m gap from last call.
+        if is_forced or not self._last_fetch or time.time() - self._last_fetch > 5 * 60:
+            tags = await self.bot.api_client.get('bot/tags')
+            self._cache = {tag['title'].lower(): tag for tag in tags}
+
+    @staticmethod
+    def _fuzzy_search(search: str, target: str) -> bool:
```
MarkKoz:
```suggestion
    def _fuzzy_search(search: str, target: str) -> int:
```

bot/cogs/tags.py
```diff
@@ -27,6 +26,44 @@ def __init__(self, bot: Bot):
...
+                break
+            found += index > 0
+        return found / len(_search) * 100
+
+    def _get_suggestions(self, tag_name: str, score: int) -> list:
+        return sorted(
+            (tag for tag in self._cache.values() if Tags._fuzzy_search(tag_name, tag['title']) >= score),
+            key=lambda tag: Tags._fuzzy_search(tag_name, tag['title']),
+            reverse=True
+        )
```
MarkKoz: Calling `_fuzzy_search` multiple times per tag is pretty inefficient. The scores could be calculated ones and stored in a dictionary e.g. `dict[tag] = score` which could be used in the key function.

bot/cogs/tags.py
```diff
@@ -27,6 +26,44 @@ def __init__(self, bot: Bot):
...
+
+    async def _get_tag(self, tag_name: str) -> list:
+        """Get a specific tag."""
+        await self._get_tags()
+        found = [self._cache.get(tag_name.lower(), None)]
+        if not found[0]:
+            return self._get_suggestions(tag_name, 100) or self._get_suggestions(tag_name, 80)
```
MarkKoz: This could end up doing redundant work because the first suggestions call discards tags which would be in the alternate threshold of 80. What it instead returned something like:

```py
{
    100: ["tag name 1", "another tag"],
    80: ["yet another tag"]
}
```

bot/cogs/tags.py
ikuyarihS: Thank you, I will include this in a single commit with the suggestion below.

MarkKoz: While you're at it, I believe there are a couple of commented out lines that you could probably just delete entirely.

bot/cogs/tags.py
ikuyarihS: That is a great suggestions! I have created a dictionary to store the scores in the form of `tag_title: matching score` ( str, int ) and by doing so, won't be calling `Tags._fuzzy_search()` in other places.

bot/cogs/tags.py
ikuyarihS: Thank you for pointing that out, I've added a list of thresholds as an argument for `self._get_suggestions()` so it can go through this list and return the first one already, avoiding calling this function too many times.

ikuyarihS:
Interesting! I've added a regex to remove all non-alphabet, as well as increasing threshold from `[100, 80]` to `[100, 90, 80, 70, 60]` since it stops as soon as suggestions are found for a threshold, this will give better suggestions that the bot thinks is what the user is searching for. This solved for both the `foo_bar` and `foobars` when searching for `foo-bar`

MarkKoz:
In some cases that still isn't working so well:

* `asks` returns `args-kwargs` instead of `ask`
* `foos` returns `off-topic` and `functions-are-objects` instead of `foo`
* `dict` returns `iterate-dict` without considering `dictcomps` too

Also discovered an unrelated issue in which it can't handle DELETE or GET requests for tags with spaces in them (returns 404). Might be a URL encoding issue since the tag is part of the URL path. It can POST fine because the tag name is instead part of the JSON.

ikuyarihS:
Hmm, I've added another complexity that will force this to search from words to words, here's the snippets I used to test

```py
from typing import Dict, List, Optional

stuff = ['args-kwargs', 'ask', 'class', 'classmethod', 'codeblock', 'decorators', 'dictcomps', 'enumerate', 'except', 'exit()', 'f-strings', 'foo', 'functions-are-objects', 'global', 'if-name-main', 'indent', 'inline', 'iterate-dict', 'listcomps', 'mutable-default-args', 'names', 'no-dm',
         'off-topic', 'open', 'or-gotcha', 'param-arg', 'paste', 'pathlib', 'pep8', 'positional-keyword', 'precedence', 'quotes', 'relative-path', 'repl', 'return', 'round', 'scope', 'seek', 'self', 'star-imports', 'traceback', 'windows-path', 'with', 'xy-problem', 'ytdl', 'zen', 'zip', ]

_cache = dict(zip(stuff, stuff))


def _fuzzy_search(search: str, target: str) -> int:
    """A simple scoring algorithm based on how many letters are found / total, with order in mind."""
    found, index = 0, 0
    _search = search.lower().replace(' ', '')
    _targets = iter(target.lower())
    _target = next(_targets)
    try:
        for letter in _search:
            index = _target.find(letter, index)
            while index == -1:
                _target = next(_targets)
                index = _target.find(letter)
            found += 1
    except StopIteration:
        pass
    return found / len(_search) * 100


def _get_suggestions(tag_name: str, thresholds: Optional[List[int]] = None) -> List[str]:
    """Return a list of suggested tags."""
    scores: Dict[str, int] = {
        tag_title: _fuzzy_search(tag_name, tag)
        for tag_title, tag in _cache.items()
    }

    thresholds = thresholds or [100, 90, 80, 70, 60]

    for threshold in thresholds:
        suggestions = [
            _cache[tag_title]
            for tag_title, matching_score in scores.items()
            if matching_score >= threshold
        ]
        if suggestions:
            return f"{tag_name} - {suggestions}"

    return f"{tag_name} not found"


print(_get_suggestions('fstring'))
print(_get_suggestions('fstrings'))
print(_get_suggestions('fstr'))
print(_get_suggestions('asks'))
print(_get_suggestions('foos'))
print(_get_suggestions('dict'))
print(_get_suggestions('opens'))
print(_get_suggestions('or'))
print(_get_suggestions('or-g'))
print(_get_suggestions('or-'))
print(_get_suggestions('got'))
print(_get_suggestions('path'))
print(_get_suggestions('main'))
print(_get_suggestions('if'))
```

Here are the results:
```py
fstring - ['f-strings']
fstrings - ['f-strings']
fstr - ['f-strings']
asks - ['args-kwargs']
foos - ['functions-are-objects']
dict - ['dictcomps', 'iterate-dict']
opens - ['open']
or - ['decorators', 'functions-are-objects', 'or-gotcha', 'positional-keyword', 'star-imports']
or-g - ['or-gotcha']
or- - ['functions-are-objects', 'or-gotcha']
got - ['or-gotcha']
path - ['pathlib', 'relative-path', 'windows-path']
main - ['if-name-main']
if - ['if-name-main']
```

I've updated the functions accordingly, can you give it a few more tries? If it is still giving inconsistent results I will implement another way to search which will force all letters of `tag_name` to start from the start of each words ( so searching for `or` will only return `or-gotcha` instead of `decorators` and other stuff )

Issue #535, opened 2019-10-15
Restrict ModManagement commands to moderation channels


SebastiaanZ:
The [`ModManagement` cog](https://github.com/python-discord/bot/blob/master/bot/cogs/moderation/management.py) contains commands to search for and edit infractions stored in our database. Since we'd like to keep that data confidential, we want to restrict these commands to the following channels:

- #mods (ID: `305126844661760000`)
- #mod-spam (ID: `620607373828030464`)
- #admins (ID: `365960823622991872`)
- #mod-alerts (ID: `473092532147060736`)

Since this restriction should hold for all the commands in the cog, the implementation should be straightforward by modifying the `cog_check` method:

https://github.com/python-discord/bot/blob/8040724d1b452b79a16c7df073bd2a8f546570a5/bot/cogs/moderation/management.py#L257-L260

Please make sure to add this check in addition to the moderation roles check; it's not a replacement. 

#### Additional info
A constant with the channel id already exists in `constants.py`/`config-default.yml` for the #admins and #mod-alerts channels, but you will need to add a constant for the other two channels. 

It may also be a good idea add a `MODERATION_CHANNELS` constant in a similar way to the `MODERATION_ROLES` constant we have. That makes it easier to modify the specific channels we want to include and as well as it making it easier to use the same set of channels in the future.

atmishra:
I would like to work on this. Thanks!

scragly:
@atmishra I've assigned you to this task, so feel free to start work on it. If you need any assistance with dev environment setup, feel free to ask questions in the `#meta` channel in the discord server.

atmishra:
Thanks! for assigning it to me. I will start working on it.

atmishra:
Hi, I have raised a PR for this ticket https://github.com/python-discord/bot/pull/543 
This is my first PR so I am not sure how to test this. I tried to do the local setup using docker, but its failing.

kosayoda:
Hello and thank you for contributing, are you in the Discord server? We'll be very glad to help you get your local setup working, since manual testing is important for Discord bots.

atmishra:
Thank you! Yes I am on the server, my username is `atmishra#0512`

PR #543, opened 2019-10-19
Restrict ModManagement commands to moderation channels

atmishra:
Closes:  #535
This PR add an extra check `in_channel_check(ctx, *constants.MODERATION_CHANNELS)` to `cog_check`. This conditon will pass only if the channel belongs to the list of moderation channels.


bot/cogs/moderation/management.py
```diff
@@ -257,7 +257,11 @@ def infraction_to_string(self, infraction: utils.Infraction) -> str:
     # This cannot be static (must have a __func__ attribute).
     def cog_check(self, ctx: Context) -> bool:
         """Only allow moderators to invoke the commands in this cog."""
```
lemonsaurus: This docstring needs to be updated to reflect the new check.

atmishra: Thanks! will do the changes.

atmishra:
I wasn't able to test it and need some help with that.

MarkKoz:
@atmishra If you need help don't hesitate to leave your question in the #meta channel on Discord.

bot/cogs/moderation/management.py
atmishra: Updated the doc-string. I am not familiar with terminology yet so, not sure if doc-string makes much sense.

atmishra:
@MarkKoz Thanks! Will post my queries there.

atmishra:
I tested this manually by calling the `!infractions` command. I found an issue, `in_channel_check` just takes a single channel and I am trying to pass a list of channels to it in my PR. Should I modify the `in_channel_check` implementation to support list of channels or just loop over different channels and call this method for each of them?

MarkKoz:
Preferably modify `in_channel_check`. There are also unit tests for this that would need to be adjusted/added. Also there is no need for a loop because the `in` operator can be used.

atmishra:
I have done the changes required. I also did manual testing and this works as expected.

PR #544, opened 2019-10-19
Build GitHub Pages -ready auto-generated documentation

vemel:
### Summary

Since the project is growing, it would be nice to have an index of every object in the repo.It is compatible with [Github Pages](https://pages.github.com/), so they can be enabled in one click, no extra configuration required.

I have just finished working on my docstring documentation generator `handsdown` and think that it might be useful for Discord Bot as well.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,76 @@
+# Python Utility Bot
+
+> Auto-generated documentation index.
+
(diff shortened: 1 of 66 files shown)


SebastiaanZ:
Hi @vemel,

Thank you for your pull request, but we're currently not looking to add this to our repository. 

As a side note, your tool seems to have missed a number of methods in our classes. I haven't checked thoroughly, but it may be related to the use of decorators.

Kind regards,

Ves

vemel:
Thank you for your feedback.

Issue #538, opened 2019-10-15
Docs command refinements


Numerlor:
With how the `symbol`s are currently loaded to the inventories dict, earlier entries get overwritten by newer ones.
This can have negative impact, most notably when loading the 2to3 tool which overrides things like `print`, `zip` or `input` and returns mostly useless docs to them.
It could be solved by never overwriting existing dict entries, or by defining a list of modules that won't override existing dict entries, to keep cases where it's better to not do it (object as an example)

Second issue is how the parsing is handled when a module is fetched from the docs.
The whole page, also containing the module functions, gets parsed into the `signature` which is put into a codeblock without any newlines between paragraphs, making it hard to read.

atmishra:
I would like to work on this.

Numerlor:
I was planning to get to it if it's approved

scragly:
Sounds good to me, feel free to start work on it @Numerlor

Numerlor:
If I were to use the list of symbols/doc module names to specify which can overwrite and which cannot, where should I put it? Skimming through it, it looks like a dozen or so will be needed

Issue #381, opened 2019-07-04
Documentation inventory failed to load


SebastiaanZ:
Some of our users noticed that our `!docs get` command wasn't working. The problem was that the documentation inventory didn't load correctly and was empty. I was unable to trace the problem back to its cause, but I suspect a connection error that wasn't logged (see below).

A manual refresh of the inventory with `!internal eval await bot.get_cog("Doc").refresh_inventory()` fixed the issue for now, but we should probably have some logic in place to catch and handle this problem automatically.

This latter action also made me suspect a connection error, since it gave me the exception below. There's no logic to handle and/or log connection exceptions at the moment, so this exception can't be found in the bot logs.

~~(By the way, someone should probably kaizen the discord.py doc url as well, since the rewrite docs are no longer available, causing the exception below)~~ Edit: I've changed the URLs for the discord.py docs in the database tables; it's working again.

```py
In [193]: await bot.get_cog("Doc").refresh_inventory()
Out[193]: 
  File "/bot/bot/cogs/eval.py", line 167, in _eval
    res = await func()
  File "<string>", line 5, in func
  File "/bot/bot/cogs/doc.py", line 184, in refresh_inventory
    await asyncio.gather(*coros)
  File "/bot/bot/cogs/doc.py", line 153, in update_single
    for _, value in (await self.bot.loop.run_in_executor(None, fetch_func)).items():
  File "/usr/local/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/bot/.venv/lib/python3.6/site-packages/sphinx/ext/intersphinx.py", line 171, in fetch_inventory
    f = _read_from_url(inv, config=app.config)
  File "/bot/.venv/lib/python3.6/site-packages/sphinx/ext/intersphinx.py", line 126, in _read_from_url
    r.raise_for_status()
  File "/bot/.venv/lib/python3.6/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: ('intersphinx inventory %r not fetchable due to %s: %s', 'https://discordpy.readthedocs.io/en/rewrite/objects.inv', <class 'requests.exceptions.HTTPError'>, '404 Client Error: Not Found for url: https://discordpy.readthedocs.io/en/rewrite/objects.inv')
```

Issue #210, opened 2018-11-29
!doc for stdlib items gives 2to3 results


lemonsaurus:
For some reason, calling `!doc get dict` or `!doc get filter` will fetch the docs from `2to3` rather than displaying the docs for the built-ins.

jchristgit:
That is a thing I've spotted for a while but wasn't sure with how to fix it. I believe we should be able to add an exclusion list to the inventory fetching, but that will be very expensive due to the number of symbols involved. Relevant code snippet:
```py
        fetch_func = functools.partial(intersphinx.fetch_inventory, config, '', inventory_url)
        for _, value in (await self.bot.loop.run_in_executor(None, fetch_func)).items():
            # Each value has a bunch of information in the form
            # `(package_name, version, relative_url, ???)`, and we only
            # need the relative documentation URL.
            for symbol, (_, _, relative_doc_url, _) in value.items():
                absolute_doc_url = base_url + relative_doc_url
                self.inventories[symbol] = absolute_doc_url
```

PR #546, opened 2019-10-20
Docs command improvements

Numerlor:
Improves the docs command by:
- [X] Resolving symbol name conflicts
- [X] Auto deleting commands that didn't find anything
- [X] Improving output for docs of modules

closes: #538 
closes: #210 
closes: #381


Numerlor:
any reason `signature_buffer` is not a list comp?
https://github.com/python-discord/bot/blob/f194df838c623d3c1b68809230f396a7241508b1/bot/cogs/doc.py#L210-L219

Numerlor:
Needs some further testing and going to take a look if I can implement #381 here

Numerlor:
@jchristgit any reason 7d2b78483a1df2f7b0fe9b43834410a5df8c5e48 uses` textwrap.shorten`?
Want to get up to 3 signatures for cases where they exist but that collapses all the newlines

Numerlor:
Some issues that I'm aware of but don't know if they should be looked into further:

- Notes and warnings like in [cmath](https://docs.python.org/3/library/cmath.html) don't get put into blocks in embeds
- Some modules like [glob](https://docs.python.org/3/library/glob.html#module-glob) don't get almost any info in their description because they start with a block of text that can't be cut off with the current logic
- Symbols linked to pages ending with `.html` can't be fetched
- A few thousand symbols similar to the format `reference/page/normal_symbol_name` exist as duplicates of normal symbols and can't be fetched through their name

Also not sure if the handling and logging of exceptions when fetching inventories is correct

bot/cogs/doc.py
```diff
@@ -243,25 +305,39 @@ def __init__(self, bot: commands.Bot):
         if len(description) > 1000:
             shortened = description[:1000]
             last_paragraph_end = shortened.rfind('\n\n')
```
scragly:
```suggestion
            last_paragraph_end = shortened.rfind('\n\n', 100)
```

It might be useful to avoid cutting off at a single-sentence paragraph. Doing this resolves cases like glob cutting off at the 'Source code' beginning line if they have too large an introductory paragraph.

bot/cogs/doc.py
```diff
@@ -243,25 +305,39 @@ def __init__(self, bot: commands.Bot):
...
-            description = description[:last_paragraph_end] + f"... [read more]({permalink})"
+            description = description[:last_paragraph_end]
```
scragly: If a paragraph ending is never found, the content is never shortened. This could possibly result in embed content being above 2000 characters.

bot/cogs/doc.py
```diff
@@ -243,25 +305,39 @@ def __init__(self, bot: commands.Bot):
         if len(description) > 1000:
             shortened = description[:1000]
```
scragly: It might be worth increasing this to 1500. Do we have cases where sigs can cover >500 characters? This would also ensure more complete information and doesn't seem too terribly lengthy to read. This is just a general suggestion to look into rather than a suggestion to change though.

bot/cogs/doc.py
Numerlor: resolved in 4795da86d0fef72ac677ae0a8f9e988da1923e17

bot/cogs/doc.py
Numerlor: resolved in 34510f52c6bbe5e2a8bbfc34f8e5d648d0d39a96

bot/cogs/doc.py
```diff
@@ -243,25 +305,39 @@ def __init__(self, bot: commands.Bot):
...
             # https://docs.djangoproject.com/en/dev/ref/views/#module-django.views
-            return discord.Embed(
-                title=f'`{symbol}`',
-                url=permalink,
-                description="This appears to be a generic page not tied to a specific symbol."
-            )
+            embed_description = "This appears to be a generic page not tied to a specific symbol."
+
+        else:
+            embed_description = "".join(f"```py\n{textwrap.shorten(signature, 500)}```" for signature in signatures)
```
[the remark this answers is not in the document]
Numerlor: resolved in 7de5156a7719f0639021e8186f7ea17f5b853af7

bot/cogs/doc.py
```diff
@@ -150,16 +175,37 @@ def __init__(self, bot: commands.Bot):
...
-            # Each value has a bunch of information in the form
-            # `(package_name, version, relative_url, ???)`, and we only
-            # need the relative documentation URL.
-            for symbol, (_, _, relative_doc_url, _) in value.items():
-                absolute_doc_url = base_url + relative_doc_url
-                self.inventories[symbol] = absolute_doc_url
-
-        log.trace(f"Fetched inventory for {package_name}.")
+        package = await self._fetch_inventory(inventory_url, config)
+        if package:
```
scragly: There's no reason to indent all items contained within this if. It's better to invert the if and dedent the contents:
```py
        package = await self._fetch_inventory(inventory_url, config)
        if not package:
            log.trace(f"No inventory for {package_name}.")
            return

        for group, value in package.items():
```

bot/cogs/doc.py
```diff
@@ -150,16 +175,37 @@ def __init__(self, bot: commands.Bot):
...
+            for group, value in package.items():
+                # Each value has a bunch of information in the form
+                # `(package_name, version, relative_url, ???)`, and we only
+                # need the package_name and the relative documentation URL.
+                for symbol, (package_name, _, relative_doc_url, _) in value.items():
```
scragly: The code can already provide the information that your comment gives, so if you want to show that `version` is one of the items ignored, just give it a name with the `_` prefix and remove the unnecessary comments:
```py
        for group, value in package.items():
            for symbol, (package_name, _version, relative_doc_url, _) in value.items():
```

bot/cogs/doc.py
```diff
@@ -150,16 +175,37 @@ def __init__(self, bot: commands.Bot):
...
+                    absolute_doc_url = base_url + relative_doc_url
+
+                    if symbol in self.inventories:
+                        # get `group_name` from _:group_name
+                        group_name = group.split(":")[1]
```
scragly: Another case of code saying the same thing as the comment, making the comment unnecessary.

bot/cogs/doc.py
```diff
@@ -150,16 +175,37 @@ def __init__(self, bot: commands.Bot):
...
+                        if (group_name in NO_OVERRIDE_GROUPS
+                                # check if any package from `NO_OVERRIDE_PACKAGES`
+                                # is in base URL of the symbol that would be overridden
+                                or any(package in self.inventories[symbol].split("/", 3)[2]
+                                       for package in NO_OVERRIDE_PACKAGES)):
+
+                            symbol = f"{group_name}.{symbol}"
+                            # If renamed `symbol` already exists, add library name in front.
+                            if symbol in self.renamed_symbols:
```
scragly: The comment is describing the code again here, but not explaining the why. The comment would be more useful if it describes why it's adding the library name in front such as "for disambiguation" or such.

bot/cogs/doc.py
```diff
@@ -150,16 +175,37 @@ def __init__(self, bot: commands.Bot):
...
+                    absolute_doc_url = base_url + relative_doc_url
+
+                    if symbol in self.inventories:
+                        # get `group_name` from _:group_name
+                        group_name = group.split(":")[1]
+                        if (group_name in NO_OVERRIDE_GROUPS
+                                # check if any package from `NO_OVERRIDE_PACKAGES`
+                                # is in base URL of the symbol that would be overridden
+                                or any(package in self.inventories[symbol].split("/", 3)[2]
+                                       for package in NO_OVERRIDE_PACKAGES)):
```
scragly: This line does not conform to the typical project style. Multilines are dropped after the opening bracket typically and the conditions could be tidied up a bit for readability.

bot/cogs/doc.py
```diff
@@ -366,6 +448,60 @@ def __init__(self, bot: commands.Bot):
...
+
+    async def _fetch_inventory(self, inventory_url: str, config: SphinxConfiguration) -> Optional[dict]:
+        """Get and return inventory from `inventory_url`. If fetching fails, return None."""
+        fetch_func = functools.partial(intersphinx.fetch_inventory, config, '', inventory_url)
+        for retry in range(1, FAILED_REQUEST_RETRY_AMOUNT+1):
+            try:
+                package = await self.bot.loop.run_in_executor(None, fetch_func)
+            except ConnectTimeout:
+                log.error(f"Fetching of inventory {inventory_url} timed out,"
+                          f" trying again. ({retry}/{FAILED_REQUEST_RETRY_AMOUNT})")
```
scragly: Does not match project style, multilines should drop after the opening bracket.
```py
                log.error(
                    f"Fetching of inventory {inventory_url} timed out,"
                    f" trying again. ({retry}/{FAILED_REQUEST_RETRY_AMOUNT})"
                )
```

bot/cogs/doc.py
```diff
@@ -366,6 +448,60 @@ def __init__(self, bot: commands.Bot):
...
+            except ProtocolError:
+                log.error(f"Connection lost while fetching inventory {inventory_url},"
+                          f" trying again. ({retry}/{FAILED_REQUEST_RETRY_AMOUNT})")
```
scragly: Does not match project style, multilines should drop after the opening bracket.
```py
                log.error(
                    f"Connection lost while fetching inventory {inventory_url},"
                    f" trying again. ({retry}/{FAILED_REQUEST_RETRY_AMOUNT})"
                )
```

bot/cogs/doc.py
Numerlor: This comment wasn't made by me, should I remove it anyway?

scragly: If you agree with the comment, yeah it would fit under our idea of kaizen, improving upon aspects we work on that might not be immediately required under the scope of the PR.

bot/cogs/doc.py
```diff
@@ -150,13 +175,32 @@ def __init__(self, bot: commands.Bot):
...
-            # Each value has a bunch of information in the form
-            # `(package_name, version, relative_url, ???)`, and we only
-            # need the relative documentation URL.
-            for symbol, (_, _, relative_doc_url, _) in value.items():
+        package = await self._fetch_inventory(inventory_url, config)
+        if not package:
+            return None
+
+        for group, value in package.items():
+            for symbol, (package_name, _, relative_doc_url, _) in value.items():
```
scragly:
```suggestion
            for symbol, (package_name, _version, relative_doc_url, _) in value.items():
```
Removing the comment was under the idea of not removing any provided info, so the second item in the tuple should still be `_version` to indicate we don't use it (underscore used) and describing the item in case we do need it sometime in future.

bot/cogs/doc.py
```diff
@@ -234,34 +295,51 @@ def __init__(self, bot: commands.Bot):
...
+        embed = discord.Embed(
             title=f'`{symbol}`',
             url=permalink,
-            description=f"```py\n{signature}```{description}"
+            description=embed_description
         )
+        # Show all symbols with the same name that were renamed in the footer.
+        embed.set_footer(
+            text=", ".join(renamed for renamed in self.renamed_symbols - {symbol}
+                           if renamed.endswith(f".{symbol}"))
```
scragly: This does not match the project style, as we drop the line after the bracket. In this case, though the line fits within the allowed 120 line length so can be reduced to a single line.

```py
        embed.set_footer(
            text=", ".join(renamed for renamed in self.renamed_symbols - {symbol} if renamed.endswith(f".{symbol}"))
        )
```

PR #550, opened 2019-10-23
Use standard filter conditions even if DEBUG_MODE is on.

Akarys42:
Hi, the old method filtered only in #dev-test if `DEBUG_MODE` was on but this channel no longer exists.

### DIFF
--- a/bot/cogs/filtering.py
+++ b/bot/cogs/filtering.py
@@ -9,7 +9,7 @@ from discord.ext.commands import Bot, Cog
 
 from bot.cogs.moderation import ModLog
 from bot.constants import (
-    Channels, Colours, DEBUG_MODE,
+    Channels, Colours,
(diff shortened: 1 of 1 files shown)


kraktus:
BTW it would be nice to load:
```
bot.load_extension("bot.cogs.doc")
```
Even in `DEBUG_MODE`

Numerlor:
The docs cog does some larger requests when the database is set up so don't think it'd be that great of an idea to put useless load on the providers

SebastiaanZ:
> The docs cog does some larger requests when the database is set up so don't think it'd be that great of an idea to put useless load on the providers

It's not that much of an issue for a bot used in testing, since it won't have any entries in the database unless you add them during testing. The providers are not hard-coded, so calls won't be made unless they are added to the database.

PR #617, opened 2019-10-25
Autodelete offensive messages after one week.

### LINKED ISSUE #364: Offensive word filtering
Once the bot's word filters are triggered, we will need to call a site endpoint to store the ID of the offending message, along with a "delete time" of one week after the authored time.

The bot will pull a list of offending message IDs daily, and delete the messages that have reached the allotted delete time.

Akarys42:
Hi, this PR add a background deletion task of offensive messages. python-discord/site#300 need to be merged first. 

This PR closes #364. 

### Implementation details 
* If the filter cog filter a message that's considered as offensive, the cog create a new offensive message object in the site db with a delete_date of one week after it was sent.
* A background task run every day, pull up a list of message to delete, find them back, and delete them.


bot/cogs/filtering.py
```diff
@@ -16,13 +19,13 @@
...
+    r"discord(?:[\.,]|dot)com(?:\/|slash)invite|"  # or discord.com/invite/
     r"discordapp(?:[\.,]|dot)com(?:\/|slash)invite|"  # or discordapp.com/invite/
-    r"discord(?:[\.,]|dot)me|"                        # or discord.me
-    r"discord(?:[\.,]|dot)io"                         # or discord.io.
-    r")(?:[\/]|slash)"                                # / or 'slash'
-    r"([a-zA-Z0-9]+)",                                # the invite code itself
+    r"discord(?:[\.,]|dot)me|"  # or discord.me
+    r"discord(?:[\.,]|dot)io"  # or discord.io.
+    r")(?:[\/]|slash)"  # / or 'slash'
+    r"([a-zA-Z0-9]+)",  # the invite code itself
```
sco1: Please revert this whitespace change

bot/cogs/filtering.py
```diff
@@ -36,6 +39,8 @@
     re.compile(fr'{expression}', flags=re.IGNORECASE) for expression in Filter.token_watchlist
 ]
 
+OFFENSIVE_MSG_DELETE_TIME = datetime.timedelta(days=7)  # Time before an offensive msg is deleted.
```
sco1: This would be better in the bot's config

bot/cogs/filtering.py
```diff
@@ -159,6 +185,21 @@ def mod_log(self) -> ModLog:
...
+                            await self.bot.api_client.post(
+                                'bot/offensive-message',
+                                json={
+                                    'id': msg.id,
+                                    'channel_id': msg.channel.id,
+                                    'delete_date': delete_date.isoformat()
+                                }
+                            )
+                            log.trace(f"Offensive message will be deleted on "
+                                      f"{delete_date.isoformat()}")
```
sco1: This should go after the next logic block to remove the possibility that we send a message to the API that the filter has already deleted.

bot/cogs/filtering.py
```diff
@@ -54,7 +59,8 @@ def __init__(self, bot: Bot):
                 "notification_msg": (
                     "Your post has been removed for abusing Unicode character rendering (aka Zalgo text). "
                     f"{_staff_mistake_str}"
-                )
+                ),
+                "offensive_msg": False
```
sco1: Let's use something like `"schedule_deletion"` for these flags instead.

bot/cogs/filtering.py
Akarys42: Oh sorry that's PyCharm autoformating

bot/cogs/filtering.py
```diff
@@ -360,6 +401,34 @@ def mod_log(self) -> ModLog:
...
+            try:
+                channel = bot.get_channel(msg['channel_id'])
+                if channel:
+                    msg_obj = await channel.fetch_message(msg['id'])
+                    await msg_obj.delete()
+            except NotFound:
+                log.info(f"Tried to delete message {msg['id']}, but the message can't be found "
+                         f"(it has been probably already deleted).")
+        log.info(f"Deleted {len(msg_list)} offensive message(s).")
+        await asyncio.sleep(time_until_next.seconds)
```
SebastiaanZ: I know the original issue spelled out `daily`, but we already have a base class (`Scheduler`) that allows us to schedule "something" at a time point in the future. We already use it for infractions and reminders, so I'd rather have this feature use that as well, to avoid having all kinds of different "scheduler"-like logic around the bot. Using a single tried and tested method makes the most sense to me.

Akarys42:
Updated.

bot/cogs/filtering.py
```diff
@@ -175,6 +191,21 @@ def mod_log(self) -> ModLog:
...
+                            delete_date = msg.created_at + OFFENSIVE_MSG_DELETE_TIME
+                            await self.bot.api_client.post(
+                                'bot/offensive-message',
+                                json={
+                                    'id': msg.id,
+                                    'channel_id': msg.channel.id,
+                                    'delete_date': delete_date.isoformat()[:-1]
+                                }
+                            )
+                            log.trace(f"Offensive message will be deleted on "
```
MarkKoz: Should be like
```py
log.trace(
    f"Offensive message..."
    f"..."
)

File: bot/cogs/filtering.py [hunk 7]
```diff
@@ -175,6 +191,21 @@ def mod_log(self) -> ModLog:
...
+                        # If the message is classed as offensive, we store it in the site db and
+                        # it will be deleted it after one week.
+                        if _filter["schedule_deletion"]:
+                            delete_date = msg.created_at + OFFENSIVE_MSG_DELETE_TIME
+                            await self.bot.api_client.post(
+                                'bot/offensive-message',
+                                json={
+                                    'id': msg.id,
+                                    'channel_id': msg.channel.id,
+                                    'delete_date': delete_date.isoformat()[:-1]
```
MarkKoz: Why does the end need to be chopped off when passed to the API? Infractions, for example, don't need to do it.

bot/cogs/filtering.py
```diff
@@ -368,6 +399,46 @@ def mod_log(self) -> ModLog:
         except discord.errors.Forbidden:
             await channel.send(f"{filtered_member.mention} {reason}")
 
+    async def _scheduled_task(self, msg: dict) -> None:
+        """A coroutine which delete the offensive message once the delete date is reached."""
+        delete_at = datetime.datetime.fromisoformat(msg['delete_date'][:-1])
```
MarkKoz: It'd be better to do this with dateutil as done [here](https://github.com/python-discord/bot/blob/master/bot/cogs/moderation/scheduler.py#L410).

bot/cogs/filtering.py
```diff
@@ -368,6 +399,46 @@ def mod_log(self) -> ModLog:
...
+        await self.bot.wait_until_ready()
+        response = await self.bot.api_client.get(
+            'bot/offensive-message',
+        )
+
+        now = datetime.datetime.utcnow()
+        loop = asyncio.get_event_loop()
+
+        for msg in response:
+            delete_at = datetime.datetime.fromisoformat(msg['delete_date'][:-1])
```
MarkKoz: Can be done with dateutil.

config-default.yml
```diff
@@ -174,7 +174,8 @@ filter:
     notify_user_domains:     false
 
     # Filter configuration
-    ping_everyone: true  # Ping @everyone when we send a mod-alert?
+    ping_everyone:             true  # Ping @everyone when we send a mod-alert?
+    offensive_msg_delete_time: 7     # How many days before deleting an offensive message?
```
MarkKoz: This should be called "days" at the end rather than "time" to make it clearer.

bot/cogs/filtering.py
```diff
@@ -75,28 +84,35 @@ def __init__(self, bot: Bot):
...
             "watch_tokens": {
                 "enabled": Filter.watch_tokens,
                 "function": self._has_watchlist_tokens,
                 "type": "watchlist",
                 "content_only": True,
+                "schedule_deletion": True
             },
         }
 
+        self.deletion_task = None
```
MarkKoz: This does not seem to be used for anything.

bot/cogs/filtering.py
```diff
@@ -368,6 +399,46 @@ def mod_log(self) -> ModLog:
...
+
+    async def delete_offensive_msg(self, msg: Mapping[str, str]) -> None:
+        """Delete an offensive message, and then delete it from the db."""
+        try:
+            channel = self.bot.get_channel(msg['channel_id'])
+            if channel:
+                msg_obj = await channel.fetch_message(msg['id'])
+                await msg_obj.delete()
+        except NotFound:
+            log.info(f"Tried to delete message {msg['id']}, but the message can't be found "
```
MarkKoz: Should be like

```py
log.info(
    f"Tried to..."
    f"..."
)
```

bot/cogs/filtering.py
```diff
@@ -368,6 +399,46 @@ def mod_log(self) -> ModLog:
...
+                     f"(it has been probably already deleted).")
+        await self.bot.api_client.delete(f'bot/offensive-message/{msg["id"]}')
```
MarkKoz: A space here makes it easier to read

```suggestion

        await self.bot.api_client.delete(f'bot/offensive-message/{msg["id"]}')
```

bot/cogs/filtering.py
```diff
@@ -368,6 +399,46 @@ def mod_log(self) -> ModLog:
...
+                self.schedule_task(loop, msg['id'], msg)
+
+    async def delete_offensive_msg(self, msg: Mapping[str, str]) -> None:
+        """Delete an offensive message, and then delete it from the db."""
+        try:
+            channel = self.bot.get_channel(msg['channel_id'])
+            if channel:
+                msg_obj = await channel.fetch_message(msg['id'])
+                await msg_obj.delete()
+        except NotFound:
```
MarkKoz: This should also catch all other Discord HTTP errors (perhaps separately so a different error message can be logged) because it's not worth getting spammed tracebacks in the console when a deletion fails.

bot/cogs/filtering.py
```diff
@@ -175,6 +191,21 @@ def mod_log(self) -> ModLog:
                             if _filter["user_notification"]:
                                 await self.notify_member(msg.author, _filter["notification_msg"], msg.channel)
 
+                        # If the message is classed as offensive, we store it in the site db and
+                        # it will be deleted it after one week.
+                        if _filter["schedule_deletion"]:
```
MarkKoz: I don't see where this schedules it for deletion. It only posts it to the API.

bot/cogs/filtering.py
```diff
@@ -368,6 +399,46 @@ def mod_log(self) -> ModLog:
...
+        await self.delete_offensive_msg(msg)
+
+        self.cancel_task(msg['id'])
+
+    async def reschedule_offensive_msg_deletion(self) -> None:
+        """Get all the pending message deletion from the API and reschedule them."""
+        await self.bot.wait_until_ready()
+        response = await self.bot.api_client.get(
+            'bot/offensive-message',
+        )
```
MarkKoz: This could fit on one line; only goes up to column 73 which isn't bad at all.

Akarys42:
Oh sorry, I did some testing before pushing and everything worked fine, I
probably did something wrong, my apologies

>

bot/cogs/filtering.py
```diff
@@ -369,6 +398,48 @@ def mod_log(self) -> ModLog:
         except discord.errors.Forbidden:
             await channel.send(f"{filtered_member.mention} {reason}")
 
+    async def _scheduled_task(self, msg: dict) -> None:
+        """Delete an offensive message once its deletion date is reached."""
+        delete_at = dateutil.parser.isoparse(msg['delete_date'])
```
SebastiaanZ: If `msg['delete_date']` contains a datetime string with a timezone (like the ones returned by the API), then we should make sure to strip the tzinfo here like we do in L418.

Currently, we're getting this each time we try to delete a rescheduled message:

```py
  File "/home/sebastiaan/pydis/repositories/bot/bot/cogs/filtering.py", line 405, in _scheduled_task
    await wait_until(delete_at)
  File "/home/sebastiaan/pydis/repositories/bot/bot/utils/time.py", line 103, in wait_until
    delay = time - (start or datetime.datetime.utcnow())
TypeError: can't subtract offset-naive and offset-aware datetimes
```

This would solve that:

```suggestion
        delete_at = dateutil.parser.isoparse(msg['delete_date']).replace(tzinfo=None)
```

bot/cogs/filtering.py
```diff
@@ -369,6 +398,48 @@ def mod_log(self) -> ModLog:
...
+            if channel:
+                msg_obj = await channel.fetch_message(msg['id'])
+                await msg_obj.delete()
+        except NotFound:
+            log.info(
+                f"Tried to delete message {msg['id']}, but the message can't be found "
+                f"(it has been probably already deleted)."
+            )
+        except HTTPException as e:
+            log.warning(f"Failed to delete message {msg['id']}: status {e.status}")
```
SebastiaanZ: What kind of errors would cause this?

Currently, this error handling means that we would not remove the actual message, but do remove the entry from the database down below. So, the offensive message remains and will never be deleted. If these type of errors are mainly triggered by things like temporary disruptions at Discord's end, then it could be a good idea to keep the rows in the database and attempt removal at a later point in time.

However, that strategy would mean that completely "unremoveable" would linger in the database indefinitely.

I'm not sure. What do you think?

Akarys42: Yeah, I put the except here in order to avoid uncaught exception, I think rescheduling them seems like the way to go. 
I'm not sure about unremoveable messages you're talking about though. Are you talking about errors not caused by a discord downtime, but by an actual error that will not solve with time? If so, yeah, it could be an issue. We could also delete them anyway if the deletion failed more than 3 times for instance.

bot/cogs/filtering.py
```diff
@@ -300,7 +329,7 @@ def mod_log(self) -> ModLog:
 
         Attempts to catch some of common ways to try to cheat the system.
         """
-        # Remove backslashes to prevent escape character aroundfuckery like
+        # Remove backslashes to prevent escape character around fuckery like
```
lemonsaurus: No no. `aroundfuckery` is one word. Well, I mean, technically it isn't a word at all, but it makes zero sense as two words.
```suggestion
        # Remove backslashes to prevent escape character aroundfuckery like
```

bot/cogs/filtering.py
```diff
@@ -369,6 +398,48 @@ def mod_log(self) -> ModLog:
         except discord.errors.Forbidden:
             await channel.send(f"{filtered_member.mention} {reason}")
 
+    async def _scheduled_task(self, msg: dict) -> None:
```
lemonsaurus: This function name does not seem descriptive enough.

bot/cogs/filtering.py
lemonsaurus: What if the message was manually deleted after the filter trigger? This is quite likely to happen a lot, and perhaps warning is not the right log level to use for this if we're going to be seeing it every time somone has deleted the message manually. I mean, as long as the message is gone, that's a good thing and no warning is necessary.

If we rescheduled these so that they would at least attempt to delete the message twice, then we could safely assume that if it fails twice, it's not due to some temporary disruption of the Discord service but because the message no longer exists. In which case, we could probably make this an INFO level log event.

This has the added benefit of not piling up unactionable sentry issues, since a warning would be sent to Sentry for us to deal with.

bot/cogs/filtering.py
SebastiaanZ: It's because it's an override/implementation of the abstract method defined in the Scheduler base class:

https://github.com/python-discord/bot/blob/7dc7ca3427cdd4b2fd1b1083d368e80f55207d9a/bot/utils/scheduling.py#L22-L32

lemonsaurus: Oh, all right. Let's resolve this then.

bot/cogs/filtering.py
SebastiaanZ: > What if the message was manually deleted after the filter trigger? This is quite likely to happen a lot, 

There's a special `except NotFound` clause with an info-level log message. This clause seems to be specifically for messages that the bot either could not fetch for other reasons that `NotFound` (whatever that may be, it could be a temporary API fluke at Discord's end) or where the deletion action fails (for whatever reason).

lemonsaurus: In that case, I suppose this is fine as a warning. I still think rescheduling is something we should probably do. At least once.

Akarys42: @lemonsaurus there is an except block just above for `NotFound` with an appropriate info message

bot/cogs/filtering.py
MarkKoz: Replace "fuckery" with something temporarily to help you understand:

> Around something like [example]

Vs

> Aroundsomething like [example]

bot/cogs/filtering.py
```diff
@@ -369,6 +398,48 @@ def mod_log(self) -> ModLog:
...
+        """Delete an offensive message once its deletion date is reached."""
+        delete_at = dateutil.parser.isoparse(msg['delete_date']).replace(tzinfo=None)
+
+        await wait_until(delete_at)
+        await self.delete_offensive_msg(msg)
+
+        self.cancel_task(msg['id'])
```
MarkKoz: With recent changes to the scheduler, tasks no longer need to cancel themselves. In fact, it's bad for them to do so.

bot/cogs/filtering.py
```diff
@@ -176,6 +191,20 @@ def mod_log(self) -> ModLog:
...
+                        if _filter["schedule_deletion"]:
+                            delete_date = (msg.created_at + OFFENSIVE_MSG_DELETE_TIME).isoformat()
+                            data = {
+                                'id': msg.id,
+                                'channel_id': msg.channel.id,
+                                'delete_date': delete_date
+                            }
+
+                            await self.bot.api_client.post('bot/offensive-messages', json=data)
+                            self.schedule_task(self.bot.loop, msg.id, data)
```
MarkKoz: `schedule_task` no longer takes the loop as an argument. Go through the cog and remove the loop argument from all calls to this function.

bot/cogs/filtering.py
Akarys42: The democracy decided.
image
`around-fuckery` it will be.

lemonsaurus: fuckaroundery will do. even around-fuckery is okay, just please god not `around fuckery`. Anything but that.

bot/cogs/filtering.py
Akarys42: I am facing a little issue. The `OffensiveMessage` model of the site database doesn’t allow PATCH requests, and the PR adding it was merged. I think we have two options:

---

### Option 1
Open a new PR to allow PATCH requests and a new deletion attempts field.
### Option 2
When the deletion needs to be retried, delete the current object from the database and create a new one. To check if we should still retry later, we could check if the message was posted `DELETION_TIME * 2` ago.

---

Which option do you prefer?

scragly:
Accidentally hit a button and destroyed all the things here for a moment, sorry about that!

bot/cogs/filtering.py
MarkKoz: A mix of the two sounds better to me. Make the API support PATCHing the `delete_date` (this is a very simple change to make with DRF). Make the client (i.e. the bot) compare the message's timestamp (keep in mind this can be calculated directly from the snowflake if needed) to determine if a deletion should be re-scheduled. A field for a deletion attempts just doesn't feel right to me.

MarkKoz:
Why is this labelled as stalled? The site PR was merged almost 4 months ago.

Akarys42:
We need a new one to allow PATCH requests, I've been pretty busy lately
because of the covid and stuff, but I'll be back soon(tm)

PR #618, opened 2019-10-26
Schedule superstarify infractions for expiration

### LINKED ISSUE #472: Superstarify prison does not release user when it expires
While the [superstarify command](https://github.com/python-discord/bot/blob/71a3ac9382851845dcb26609d64299bd69b0f0f5/bot/cogs/superstarify/__init__.py#L148-L219) correctly sends the expiration date of the superstarify prison to the API, it does not schedule a task make sure the infraction is actually deactivated when we reach the expiration date. This means that **the user will be kept in superstarify prison until we manually release them** with the `!unsuperstarify` command.

The most obvious solution is to have a superstarify infraction schedule a task like other temporary infractions do. We also need to confirm that this task is recreated if the bot restarts before it was actioned. 

Additionally, it's probably a good idea to confirm that the expiration date is in the future before sending a DM to notify a user of their imprisonment [when they try to change their nickname](https://github.com/python-discord/bot/blob/71a3ac9382851845dcb26609d64299bd69b0f0f5/bot/cogs/superstarify/__init__.py#L63-L83) or [when they rejoin the server](https://github.com/python-discord/bot/blob/71a3ac9382851845dcb26609d64299bd69b0f0f5/bot/cogs/superstarify/__init__.py#L109-L123). It also wouldn't hurt to add that the datetime string is in UTC.

Context: This notified me of an expiration date in the past (08:21 in my local time):

Since @MarkKoz is currently working on the related moderation cog, I've assigned him for now, but I have not yet confirmed that he wants to work on this. Regardless, since we need to schedule this just like regular temporary infractions, it's probably a good idea if he's involved in the discussion.

MarkKoz:
Fixes #472.

A new class `InfractionScheduler` was created and most of the meat of the infractions cog (utility functions for applying/pardoning) was moved there. Superstarify now partially makes use of those functions too, but not fully so that things such as the embed sent when a superstar is applied is retained.

Superstar icons for the mod log were added.

Being consistent with mute infractions, users will not be DM'd about a re-applied superstar infraction when they re-join the server. They will still be DM'd if they try to edit their nickname.

### DIFF
--- a/bot/cogs/moderation/infractions.py
+++ b/bot/cogs/moderation/infractions.py
@@ -209,7 +209,7 @@ class Infractions(InfractionScheduler, commands.Cog):
         await self.pardon_infraction(ctx, "ban", user)
 
     # endregion
-    # region: Base infraction functions
+    # region: Base apply functions
 
     async def apply_mute(self, ctx: Context, user: Member, reason: str, **kwargs) -> None:
         """Apply a mute infraction with kwargs passed to `post_infraction`."""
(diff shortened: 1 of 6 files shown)


MarkKoz:
@SebastiaanZ Sorry if this is a bit confusing, but I've addressed this in #618 because it turned out to be easier for me to do it that way.

PR #619, opened 2019-10-26
Moderation logging

MarkKoz:
**#618 must be merged first.**

The moderation modules were a bit bare in the logging. Most of the new logging is trace logging, but there are some debugs and warnings too.

### DIFF
--- a/bot/cogs/moderation/scheduler.py
+++ b/bot/cogs/moderation/scheduler.py
@@ -39,6 +39,8 @@ class InfractionScheduler(Scheduler):
         """Schedule expiration for previous infractions."""
         await self.bot.wait_until_ready()
 
+        log.trace(f"Rescheduling infractions for {self.__class__.__name__}.")
+
         infractions = await self.bot.api_client.get(
             'bot/infractions',
             params={'active': 'true'}
@@ -59,6 +61,10 @@ class InfractionScheduler(Scheduler):
 
         # Mark as inactive if less than a minute remains.
         if delta < 60:
+            log.info(
+                "Infraction will be deactivated instead of re-applied "
+                "because less than 1 minute remains."
+            )
             await self.deactivate_infraction(infraction)
             return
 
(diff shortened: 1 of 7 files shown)


bot/cogs/moderation/scheduler.py
```diff
@@ -78,6 +84,9 @@ def mod_log(self) -> ModLog:
         icon = utils.INFRACTION_ICONS[infr_type][0]
         reason = infraction["reason"]
         expiry = infraction["expires_at"]
+        id_ = infraction['id']
```
mathsman5133: Should this be a double quote?

bot/cogs/moderation/scheduler.py
```diff
@@ -179,12 +209,16 @@ def mod_log(self) -> ModLog:
         log_text["Member"] = f"{user.mention}(`{user.id}`)"
         log_text["Actor"] = str(ctx.message.author)
         log_content = None
-        footer = f"ID: {response[0]['id']}"
+        id_ = response[0]['id']
```
mathsman5133: Double quote?

bot/cogs/moderation/scheduler.py
```diff
@@ -266,9 +305,9 @@ def mod_log(self) -> ModLog:
         mod_role = guild.get_role(constants.Roles.moderator)
         user_id = infraction["user"]
         _type = infraction["type"]
```
mathsman5133: I know you didn't touch this but while we're fixing the `id_` should this be `type_`?

bot/cogs/moderation/scheduler.py
MarkKoz: There's not an explicit style set though majority prefer double IIRC. There was talks of adding a flake8 plugin for it and if that goes through then this will be taken care of then.

bot/cogs/moderation/scheduler.py
MarkKoz: Yeah I went ahead and fixed some of the other ones in the code base (using the regex `^\s*?_\w+`).

PR #621, opened 2019-10-27
Duck pond!

lemonsaurus:
This cog will listen for duck reactions on any message, and then: 
- If the reaction was added by a staff member 
- and the reaction was a duck 
- and the message has not already been added to the #duck-pond  

It will add the message to the #duck-pond channel and then add a green checkbox to the original message to indicate that the message has ~~been ponded~~ undergone pondification. Once this checkmark has been added, the message will not be processed in the future. If the checkmark is removed and there are more than ducks_required ducks on the message, the bot will automatically add the checkmark back.

Messages are added to the #duck-pond via webhook, so that they can retain the appearance of having their original authors.     

image

If all reactions are removed (with the clear all reactions action), the bot does not have a countermeasure for this. In order to implement a countermeasure, it would be necessary to involve the API and the database. This might be an interesting change to make for a future version of this, because it would also allow us to do stuff like keep track of how many ducks are associated with each post, and create leaderboards -- but it's out of scope for this initial PR.

### DIFF
--- a/bot/cogs/duck_pond.py
+++ b/bot/cogs/duck_pond.py
@@ -62,7 +62,7 @@ class DuckPond(Cog):
                 embed=embed
             )
         except discord.HTTPException:
-            log.exception(f"Failed to send a message to the Duck Pool webhook")
+            log.exception("Failed to send a message to the Duck Pool webhook")
 
     async def count_ducks(self, message: Message) -> int:
         """
@@ -76,8 +76,8 @@ class DuckPond(Cog):
         for reaction in message.reactions:
             async for user in reaction.users():
 
-                # Is the user or member a staff member?
-                if not self.is_staff(user) or not user.id not in duck_reactors:
+                # Is the user a staff member and not already counted as reactor?
+                if not self.is_staff(user) or user.id in duck_reactors:
                     continue
 
                 # Is the emoji a duck?
@@ -90,6 +90,35 @@ class DuckPond(Cog):
                         duck_count += 1
                         duck_reactors.append(user.id)
         return duck_count
+
+    async def relay_message_to_duck_pond(self, message: Message) -> None:
+        """Relays the message's content and attachments to the duck pond channel."""
+        clean_content = message.clean_content
+
+        if clean_content:
+            await self.send_webhook(
+                content=message.clean_content,
+                username=message.author.display_name,
+                avatar_url=message.author.avatar_url
+            )
+
+        if message.attachments:
+            try:
+                await send_attachments(message, self.webhook)
+            except (errors.Forbidden, errors.NotFound):
+                e = Embed(
+                    description=":x: **This message contained an attachment, but it could not be retrieved**",
+                    color=Color.red()
+                )
+                await self.send_webhook(
+                    embed=e,
+                    username=message.author.display_name,
+                    avatar_url=message.author.avatar_url
+                )
+            except discord.HTTPException:
+                log.exception(f"Failed to send an attachment to the webhook")
+
+        await message.add_reaction("✅")
 
     @Cog.listener()
     async def on_raw_reaction_add(self, payload: RawReactionActionEvent) -> None:
@@ -124,32 +153,7 @@ class DuckPond(Cog):
 
         # If we've got more than the required amount of ducks, send the message to the duck_pond.
         if duck_count >= constants.DuckPond.threshold:
-            clean_content = message.clean_content
-
-            if clean_content:
-                await self.send_webhook(
-                    content=message.clean_content,
-                    username=message.author.display_name,
-                    avatar_url=message.author.avatar_url
-                )
-
-            if message.attachments:
-                try:
-                    await send_attachments(message, self.webhook)
-                except (errors.Forbidden, errors.NotFound):
-                    e = Embed(
-                        description=":x: **This message contained an attachment, but it could not be retrieved**",
-                        color=Color.red()
-                    )
-                    await self.send_webhook(
-                        embed=e,
-                        username=message.author.display_name,
-                        avatar_url=message.author.avatar_url
-                    )
-                except discord.HTTPException:
-                    log.exception(f"Failed to send an attachment to the webhook")
-
-            await message.add_reaction("✅")
+            await self.relay_message_to_duck_pond(message)
 
     @Cog.listener()
     async def on_raw_reaction_remove(self, payload: RawReactionActionEvent) -> None:
--- a/bot/cogs/duck_pond.py
+++ b/bot/cogs/duck_pond.py
@@ -91,7 +91,7 @@ class DuckPond(Cog):
                         duck_reactors.append(user.id)
         return duck_count
 
-    async def relay_message_to_duck_pond(self, message: Message) -> None:
+    async def relay_message(self, message: Message) -> None:
         """Relays the message's content and attachments to the duck pond channel."""
         clean_content = message.clean_content
 
@@ -120,6 +120,17 @@ class DuckPond(Cog):
 
         await message.add_reaction("✅")
 
+    @staticmethod
+    def _payload_has_duckpond_emoji(payload: RawReactionActionEvent) -> bool:
+        """Test if the RawReactionActionEvent payload contains a duckpond emoji."""
+        if payload.emoji.is_custom_emoji():
+            if payload.emoji.id in constants.DuckPond.custom_emojis:
+                return True
+        elif payload.emoji.name == "🦆":
+            return True
+
+        return False
+
     @Cog.listener()
     async def on_raw_reaction_add(self, payload: RawReactionActionEvent) -> None:
         """
@@ -130,10 +141,7 @@ class DuckPond(Cog):
         send the message off to the duck pond.
         """
         # Is the emoji in the reaction a duck?
-        if payload.emoji.is_custom_emoji():
-            if payload.emoji.id not in constants.DuckPond.custom_emojis:
-                return
-        elif payload.emoji.name != "🦆":
+        if not self._payload_has_duckpond_emoji(payload):
             return
 
         channel = discord.utils.get(self.bot.get_all_channels(), id=payload.channel_id)
@@ -153,7 +161,7 @@ class DuckPond(Cog):
 
         # If we've got more than the required amount of ducks, send the message to the duck_pond.
         if duck_count >= constants.DuckPond.threshold:
-            await self.relay_message_to_duck_pond(message)
+            await self.relay_message(message)
 
     @Cog.listener()
     async def on_raw_reaction_remove(self, payload: RawReactionActionEvent) -> None:
--- a/bot/cogs/duck_pond.py
+++ b/bot/cogs/duck_pond.py
@@ -129,19 +129,19 @@ class DuckPond(Cog):
         amount of ducks specified in the config under duck_pond/threshold, it will
         send the message off to the duck pond.
         """
+        # Is the emoji in the reaction a duck?
+        if payload.emoji.is_custom_emoji():
+            if payload.emoji.id not in constants.DuckPond.custom_emojis:
+                return
+        elif payload.emoji.name != "🦆":
+            return
+
         channel = discord.utils.get(self.bot.get_all_channels(), id=payload.channel_id)
         message = await channel.fetch_message(payload.message_id)
         member = discord.utils.get(message.guild.members, id=payload.user_id)
 
         # Is the member a human and a staff member?
         if not self.is_staff(member) or member.bot:
-            return
-
-        # Is the emoji in the reaction a duck?
-        if payload.emoji.is_custom_emoji():
-            if payload.emoji.id not in constants.DuckPond.custom_emojis:
-                return
-        elif payload.emoji.name != "🦆":
             return
 
         # Does the message already have a green checkmark?
--- a/bot/cogs/duck_pond.py
+++ b/bot/cogs/duck_pond.py
@@ -1,8 +1,8 @@
 import logging
-from typing import List, Optional, Union
+from typing import Optional, Union
 
 import discord
-from discord import Color, Embed, Member, Message, PartialEmoji, RawReactionActionEvent, Reaction, User, errors
+from discord import Color, Embed, Member, Message, RawReactionActionEvent, User, errors
 from discord.ext.commands import Bot, Cog
 
 from bot import constants
@@ -16,7 +16,6 @@ class DuckPond(Cog):
 
     def __init__(self, bot: Bot):
         self.bot = bot
-        self.log = log
         self.webhook_id = constants.Webhooks.duck_pond
         self.bot.loop.create_task(self.fetch_webhook())
 
@@ -27,7 +26,7 @@ class DuckPond(Cog):
         try:
             self.webhook = await self.bot.fetch_webhook(self.webhook_id)
         except discord.HTTPException:
-            self.log.exception(f"Failed to fetch webhook with id `{self.webhook_id}`")
+            log.exception(f"Failed to fetch webhook with id `{self.webhook_id}`")
 
     @staticmethod
     def is_staff(member: Union[User, Member]) -> bool:
@@ -39,22 +38,11 @@ class DuckPond(Cog):
         return False
 
     @staticmethod
-    def has_green_checkmark(message: Optional[Message] = None, reaction_list: Optional[List[Reaction]] = None) -> bool:
+    def has_green_checkmark(message: Message) -> bool:
         """Check if the message has a green checkmark reaction."""
-        assert message or reaction_list, "You can either pass message or reactions, but not both, or neither."
-
-        if message:
-            reactions = message.reactions
-        else:
-            reactions = reaction_list
-
-        for reaction in reactions:
-            if isinstance(reaction.emoji, str):
-                if reaction.emoji == "✅":
-                    return True
-            elif isinstance(reaction.emoji, PartialEmoji):
-                if reaction.emoji.name == "✅":
-                    return True
+        for reaction in message.reactions:
+            if reaction.emoji == "✅":
+                return True
         return False
 
     async def send_webhook(
@@ -72,52 +60,34 @@ class DuckPond(Cog):
                 avatar_url=avatar_url,
                 embed=embed
             )
-        except discord.HTTPException as exc:
-            self.log.exception(
-                f"Failed to send a message to the Duck Pool webhook",
-                exc_info=exc
-            )
+        except discord.HTTPException:
+            log.exception(f"Failed to send a message to the Duck Pool webhook")
 
-    async def count_ducks(
-        self,
-        message: Optional[Message] = None,
-        reaction_list: Optional[List[Reaction]] = None
-    ) -> int:
+    async def count_ducks(self, message: Message) -> int:
         """
         Count the number of ducks in the reactions of a specific message.
 
         Only counts ducks added by staff members.
         """
-        assert message or reaction_list, "You can either pass message or reactions, but not both, or neither."
-
         duck_count = 0
         duck_reactors = []
 
-        if message:
-            reactions = message.reactions
-        else:
-            reactions = reaction_list
-
-        for reaction in reactions:
+        for reaction in message.reactions:
             async for user in reaction.users():
 
                 # Is the user or member a staff member?
-                if self.is_staff(user) and user.id not in duck_reactors:
+                if not self.is_staff(user) or not user.id not in duck_reactors:
+                    continue
 
-                    # Is the emoji a duck?
-                    if hasattr(reaction.emoji, "id"):
-                        if reaction.emoji.id in constants.DuckPond.duck_custom_emojis:
-                            duck_count += 1
-                            duck_reactors.append(user.id)
-                    else:
-                        if isinstance(reaction.emoji, str):
-                            if reaction.emoji == "🦆":
-                                duck_count += 1
-                                duck_reactors.append(user.id)
-                        elif isinstance(reaction.emoji, PartialEmoji):
-                            if reaction.emoji.name == "🦆":
-                                duck_count += 1
-                                duck_reactors.append(user.id)
+                # Is the emoji a duck?
+                if hasattr(reaction.emoji, "id"):
+                    if reaction.emoji.id in constants.DuckPond.custom_emojis:
+                        duck_count += 1
+                        duck_reactors.append(user.id)
+                elif isinstance(reaction.emoji, str):
+                    if reaction.emoji == "🦆":
+                        duck_count += 1
+                        duck_reactors.append(user.id)
         return duck_count
 
     @Cog.listener()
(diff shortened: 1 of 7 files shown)


bot/cogs/duck_pond.py
```diff
@@ -0,0 +1,213 @@
...
+
+        elif isinstance(payload.emoji, PartialEmoji):
+            if payload.emoji.name == "✅":
+                duck_count = await self.count_ducks(message)
+                if duck_count >= constants.DuckPond.ducks_required:
+                    await message.add_reaction("✅")
+
+
+def setup(bot: Bot) -> None:
+    """Token Remover cog load."""
```
MarkKoz:
```suggestion
    """Load the duck pond cog."""
```

bot/constants.py
```diff
@@ -501,6 +510,30 @@ class RedirectOutput(metaclass=YAMLGetter):
     delete_delay: int
 
 
+class Event(Enum):
```
MarkKoz: Why was this moved?

lemonsaurus: kaizen. It was in a stupid place, and belongs with the other classes.

bot/cogs/duck_pond.py
```diff
@@ -0,0 +1,213 @@
...
+        except discord.HTTPException as exc:
+            self.log.exception(
+                f"Failed to send a message to the Duck Pool webhook",
+                exc_info=exc
+            )
+
+    async def count_ducks(
+        self,
+        message: Optional[Message] = None,
+        reaction_list: Optional[List[Reaction]] = None
```
MarkKoz: Don't see a reason to support this parameter since it's not being used.

bot/cogs/duck_pond.py
```diff
@@ -0,0 +1,213 @@
...
+    def is_staff(member: Union[User, Member]) -> bool:
+        """Check if a specific member or user is staff."""
+        if hasattr(member, "roles"):
+            for role in member.roles:
+                if role.id in constants.STAFF_ROLES:
+                    return True
+        return False
+
+    @staticmethod
+    def has_green_checkmark(message: Optional[Message] = None, reaction_list: Optional[List[Reaction]] = None) -> bool:
```
MarkKoz: Don't see a reason to support the `reaction_list` parameter since it's not being used.

bot/cogs/duck_pond.py
```diff
@@ -0,0 +1,213 @@
...
+        if message:
+            reactions = message.reactions
+        else:
+            reactions = reaction_list
+
+        for reaction in reactions:
+            if isinstance(reaction.emoji, str):
+                if reaction.emoji == "✅":
+                    return True
+            elif isinstance(reaction.emoji, PartialEmoji):
```
MarkKoz: I don't think supporting `PartialEmoji` is necessary. In practice, only message objects retrieved with `fetch_message()` are used:

> This model (PartialEmoji) will be given in two scenarios:
>
>    * “Raw” data events such as on_raw_reaction_add()
>    * Custom emoji that the bot cannot see from e.g. Message.reactions

Neither apply since the message object is neither from a raw event nor is the green checkmark a custom emoji. I feel like we shouldn't bother since this is function is meant for internal use rather than as a general-purpose utility.

bot/constants.py
```diff
@@ -236,6 +236,14 @@ class Colours(metaclass=YAMLGetter):
     soft_orange: int
 
 
+class DuckPond(metaclass=YAMLGetter):
+    section = "duck_pond"
+
+    ducks_required: int
+    duck_custom_emojis: List[int]
+    duck_pond_channel: int
```
MarkKoz: Prefixing with `duck` for some of these is redundant since the whole attribute is already under the `DuckPond` namespace. How about something like this?

```py
    threshold: int
    custom_emojis: List[int]
    channel: int
```

bot/cogs/duck_pond.py
```diff
@@ -0,0 +1,213 @@
...
+                    if hasattr(reaction.emoji, "id"):
+                        if reaction.emoji.id in constants.DuckPond.duck_custom_emojis:
+                            duck_count += 1
+                            duck_reactors.append(user.id)
+                    else:
+                        if isinstance(reaction.emoji, str):
+                            if reaction.emoji == "🦆":
+                                duck_count += 1
+                                duck_reactors.append(user.id)
+                        elif isinstance(reaction.emoji, PartialEmoji):
```
MarkKoz: Same point as line 55.

bot/cogs/duck_pond.py
lemonsaurus: This, as well as the other `reaction_list` options for the two helpers comes from an `on_reaction_clear` handler that I have removed because it wasn't really working, just FYI. They did serve a purpose, but they don't anymore. Good spot.

bot/cogs/duck_pond.py
```diff
@@ -0,0 +1,213 @@
...
+            await self.webhook.send(
+                content=content,
+                username=username,
+                avatar_url=avatar_url,
+                embed=embed
+            )
+        except discord.HTTPException as exc:
+            self.log.exception(
+                f"Failed to send a message to the Duck Pool webhook",
+                exc_info=exc
```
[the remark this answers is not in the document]
lemonsaurus: I'm not exactly sure what you're saying. I lifted this approach from the watch_channels implementation, which we're using for bigbrother and talent pool, so I assumed it had been vetted, but honestly I'm not really sure if this is necessary.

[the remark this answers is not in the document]
SebastiaanZ: It's not necessary, that's my mistake. `log.exception` only needs `exc_info` if you're logging outside of the exception handling context. Other logging methods do not log the traceback by default, but still support his kwarg to enable it. (I think the whole point of `log.exception` is to log an `error`-level message that automatically includes the relevant traceback info.)

bot/cogs/duck_pond.py
```diff
@@ -0,0 +1,213 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class DuckPond(Cog):
+    """Relays messages to #duck-pond whenever a certain number of duck reactions have been achieved."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.log = log
```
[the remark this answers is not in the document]
SebastiaanZ: This probably comes from the watch channels module as well. The reason it was used there is to make sure logging messages emitted from the base class are logged with the `__name__` of the module implementing the ABC, not the module containing the ABC itself.

bot/cogs/duck_pond.py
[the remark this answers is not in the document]
lemonsaurus: Okay, cool. I will change those then. Thanks @SebastiaanZ

lemonsaurus:
All review comments are addressed, but I guess I should probably write tests for this feature, so don't merge it yet.

lemonsaurus:
I started on the tests, will try to finish them later tonight.

SebastiaanZ:
I just realized that I did not yet add tests for the new mock types I've added. I'll do that tomorrow, sleep first.

tests/helpers.py
```diff
@@ -303,6 +303,25 @@ def __init__(self, roles: Optional[Iterable[MockRole]] = None, **kwargs) -> None
...
+
+    Instances of this class will follow the specifications of `discord.User` instances. For more
+    information, see the `MockGuild` docstring.
+    """
+    def __init__(self, **kwargs) -> None:
+        default_kwargs = {'name': 'user', 'id': next(self.discord_id), 'bot': False}
+        super().__init__(spec_set=user_instance, **collections.ChainMap(kwargs, default_kwargs))
+
+        if 'mention' not in kwargs:
+            self.mention = f"@{self.name}"
```
MarkKoz: Shouldn't this be `<@id>`? That's how discord.py does it.

SebastiaanZ: I can change it; I used this because `name` is usually more descriptive in the test output and I did not think having the actual mention format here would aid in testing. If a particular piece of code does need the correct format, an explicit mention can still be passed as a kwarg.

MarkKoz: Nah it's fine.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+        with self.assertLogs(logger=log, level=logging.ERROR) as log_watcher:
+            asyncio.run(self.cog.fetch_webhook())
+
+        self.bot.wait_until_ready.assert_called_once()
+        self.bot.fetch_webhook.assert_called_once_with(1)
+
+        self.assertEqual(len(log_watcher.records), 1)
+
+        [record] = log_watcher.records
+        self.assertEqual(record.message, f"Failed to fetch webhook with id `{self.cog.webhook_id}`")
```
MarkKoz: I think asserting the exact message may be going too far.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+        # Assert that we've made it past `self.is_staff`
+        is_staff.assert_called_once()
+
+    @patch(f"{MODULE_PATH}.DuckPond.relay_message_to_duck_pond", new_callable=helpers.AsyncMock)
+    @patch(f"{MODULE_PATH}.DuckPond.count_ducks", new_callable=helpers.AsyncMock)
+    @helpers.async_test
+    async def test_on_raw_reaction_add_does_not_relay_below_duck_threshold(self, count_ducks, message_relay):
+        """The `on_raw_reaction_add` listener should not relay messages or attachments below the duck threshold."""
+        test_cases = (
+            (constants.DuckPond.threshold-1, False),
```
MarkKoz:
```suggestion
            (constants.DuckPond.threshold - 1, False),
```

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+            (constants.DuckPond.threshold, True),
+            (constants.DuckPond.threshold+1, True),
```
MarkKoz:
```suggestion
            (constants.DuckPond.threshold + 1, True),
```

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+
+        payload.emoji = self.duck_pond_emoji
+
+        for duck_count, should_relay in test_cases:
+            count_ducks.return_value = duck_count
+            with self.subTest(duck_count=duck_count, should_relay=should_relay):
+                await self.cog.on_raw_reaction_add(payload)
+
+                # Confirm that we've made it past counting
+                count_ducks.assert_called_once()
```
MarkKoz: Eh... this doesn't seem necessary to assert at this point. But I won't fight you to death over this point; it can stay if you wish.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+        self.bot.get_all_channels.return_value = (channel, )
+
+        payload = MagicMock(channel_id=channel.id, message_id=message.id, emoji=checkmark)
+
+        test_cases = (
+            (constants.DuckPond.threshold - 1, False),
+            (constants.DuckPond.threshold, True),
+            (constants.DuckPond.threshold + 1, True),
+        )
+        for duck_count, should_readd_checkmark in test_cases:
```
MarkKoz: Does the second `d` in `readd` have a meaning ("d" for "ducks"?) or is it a typo?

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+    """Tests setup of the `DuckPond` cog."""
+
+    def test_setup(self):
+        """Setup of the cog should log a message at `INFO` level."""
+        bot = helpers.MockBot()
+        log = logging.getLogger('bot.cogs.duck_pond')
+
+        with self.assertLogs(logger=log, level=logging.INFO) as log_watcher:
+            duck_pond.setup(bot)
+            line = log_watcher.output[0]
```
MarkKoz: Wouldn't this raise an `IndexError` if it had no log messages? That would seem undesirable compared to an assertion failure.

bot/cogs/duck_pond.py
```diff
@@ -0,0 +1,174 @@
...
+                    if reaction.emoji.id in constants.DuckPond.custom_emojis:
+                        duck_count += 1
+                        duck_reactors.append(user.id)
+                elif isinstance(reaction.emoji, str):
+                    if reaction.emoji == "🦆":
+                        duck_count += 1
+                        duck_reactors.append(user.id)
+        return duck_count
+
+    async def relay_message_to_duck_pond(self, message: Message) -> None:
```
MarkKoz: This name feels too verbose. How about `relay_messages`?

bot/cogs/duck_pond.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # Time to count our ducks!
+        duck_count = await self.count_ducks(message)
+
+        # If we've got more than the required amount of ducks, send the message to the duck_pond.
+        if duck_count >= constants.DuckPond.threshold:
+            await self.relay_message_to_duck_pond(message)
+
+    @Cog.listener()
+    async def on_raw_reaction_remove(self, payload: RawReactionActionEvent) -> None:
```
MarkKoz: Is this really needed? It only cares about the bot's checkmark and ignores those of other users. Only the bot can remove its own reaction (which it would never do since we didn't program it to do that), right?

tests/bot/cogs/test_duck_pond.py
SebastiaanZ: It's to assert it's not another logging message, but the one we actually expect. I can remove it, but it's a pattern that's used more often in our tests (both in `bot`, but also in `site`, if I recall correctly).

tests/bot/cogs/test_duck_pond.py
SebastiaanZ: Hmm, yes, I failed to rewrite this.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+        test_values = (
+            (helpers.MockMessage(clean_content="", attachments=[]), False, False),
+            (helpers.MockMessage(clean_content="message", attachments=[]), True, False),
+            (helpers.MockMessage(clean_content="", attachments=["attachment"]), False, True),
+            (helpers.MockMessage(clean_content="message", attachments=["attachment"]), True, True),
+        )
+
+        for message, expect_webhook_call, expect_attachment_call in test_values:
+            with patch(send_webhook_path, new_callable=helpers.AsyncMock) as send_webhook:
+                with patch(send_attachments_path, new_callable=helpers.AsyncMock) as send_attachments:
```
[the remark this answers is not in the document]
SebastiaanZ: That automatically resets the mocks as they are recreated in each iteration. I should probably do the same for the `send_webhook` patch in the next test to remove the necessity of resetting it manually.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+                    with self.subTest(clean_content=message.clean_content, attachments=message.attachments):
+                        await self.cog.relay_message_to_duck_pond(message)
+
+                        self.assertEqual(expect_webhook_call, send_webhook.called)
+                        self.assertEqual(expect_attachment_call, send_attachments.called)
+
+                        message.add_reaction.assert_called_once_with(self.checkmark_emoji)
+                        message.reset_mock()
```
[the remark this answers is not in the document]
SebastiaanZ: Good point.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+        for side_effect in side_effects:
+            send_attachments.side_effect = side_effect
+            with self.subTest(side_effect=type(side_effect).__name__):
+                with self.assertNotLogs(logger=log, level=logging.ERROR):
+                    await self.cog.relay_message_to_duck_pond(message)
+
+                self.assertEqual(send_webhook.call_count, 2)
+                send_webhook.reset_mock()
+
+        # Subtests for the second `except` block
```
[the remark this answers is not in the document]
SebastiaanZ: I'll move it to a separate test.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+                helpers.MockMessage(reactions=[self._get_reaction(emoji=self.duck_pond_emoji, staff=3, nonstaff=2)]),
+                3
+            ),
+            # A staffer with multiple duck reactions only counts once
+            (
+                "Two different duck reactions from the same staffer",
+                helpers.MockMessage(reactions=[
+                    helpers.MockReaction(emoji=self.duck_pond_emoji, users=[self.staff_member]),
+                    helpers.MockReaction(emoji=self.unicode_duck_emoji, users=[self.staff_member]),
+                ]),
```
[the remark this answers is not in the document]
SebastiaanZ: On second thought, there are just two of these; I'll change it.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+        """The `has_green_checkmark` method should only return `True` if one is present."""
+        test_cases = (
+            (
+                "No reactions", helpers.MockMessage(), False
+            ),
+            (
+                "No green check mark reactions",
+                helpers.MockMessage(reactions=[
+                    helpers.MockReaction(emoji=self.unicode_duck_emoji),
+                    helpers.MockReaction(emoji=self.thumbs_up_emoji)
```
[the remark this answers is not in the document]
MarkKoz: Let's say the function would return `True` if it saw the bot reacted, even if the reaction wasn't a green check mark. This is not desirable of course, so we'd want to test that doesn't happen. If the author of the reactions in this test case is made to be the bot, then this test case can also be used to test for the situation described above.

tests/bot/cogs/test_duck_pond.py
MarkKoz: The log is captured from a specific logger and the level is asserted. You don't think that is adequate? It seems that the contents of the message don't actually matter but checking them is merely the way to be sure the correct message is sent. In practice, it's unlikely to be a nuisance but it does seem silly to have to change a test case when one wants to edit a log message.

If you'd like, we can leave this and save the discussion for another place since it affects all tests.

bot/cogs/duck_pond.py
MarkKoz: You're right. I tried but didn't see how do it at first.

tests/bot/cogs/test_duck_pond.py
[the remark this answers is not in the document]
MarkKoz: What does a reset even do here? Does it make it awaitable again?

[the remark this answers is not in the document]
MarkKoz: Also, I think the context managers are kind of ugly. Do you think it still looks better as decorators even if it means having to call reset?

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+            message.guild.members = (member, )
+            with self.subTest(test_case=description), patch(f"{MODULE_PATH}.DuckPond.has_green_checkmark") as checkmark:
+                checkmark.side_effect = AssertionError(
+                    "Expected method to return before calling `self.has_green_checkmark`."
+                )
+                self.assertIsNone(await self.cog.on_raw_reaction_add(payload))
+
+                # Check that we did make it past the payload checks
+                channel.fetch_message.assert_called_once()
+                channel.fetch_message.reset_mock()
```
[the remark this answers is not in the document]
MarkKoz: > Another option would be to factor out the checks to their own helper methods that can be tested at a unit.

Perhaps, but how would that look? I feel like it'd add bloat to the code considering how simple the checks are. Maybe it's just a necessary thing to deal with if we want nicer tests.

bot/cogs/duck_pond.py
lemonsaurus: might be overkill. I don't mind if we keep it or not.

tests/bot/cogs/test_duck_pond.py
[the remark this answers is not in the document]
lemonsaurus: I prefer it this way, the fully expanded way is less readable.

tests/bot/cogs/test_duck_pond.py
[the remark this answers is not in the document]
SebastiaanZ: Could be something like this:

```py
    @staticmethod
    def _payload_has_duckpond_emoji(payload: RawReactionActionEvent) -> bool:
        """Test if the RawReactionActionEvent payload contains a duckpond emoji."""
        if payload.emoji.is_custom_emoji():
            if payload.emoji.id in constants.DuckPond.custom_emojis:
                return True
        elif payload.emoji.name == "🦆":
            return True

        return False

    @Cog.listener()
    async def on_raw_reaction_add(self, payload: RawReactionActionEvent) -> None:
        """
        Determine if a message should be sent to the duck pond.

        This will count the number of duck reactions on the message, and if this amount meets the
        amount of ducks specified in the config under duck_pond/threshold, it will
        send the message off to the duck pond.
        """
        # Is the emoji in the reaction a duck?
        if not self._payload_has_duckpond_emoji(payload):
            return

        channel = discord.utils.get(self.bot.get_all_channels(), id=payload.channel_id)
        member = discord.utils.get(message.guild.members, id=payload.user_id)

        # Is the member a human and a staff member?
        if not self.is_staff(member) or member.bot:
            return

        # Does the message already have a green checkmark?
        if await self.has_green_checkmark(message):
            return

        # Time to count our ducks!
        duck_count = await self.count_ducks(message)

        # If we've got more than the required amount of ducks, send the message to the duck_pond.
        if duck_count >= constants.DuckPond.threshold:
            await self.relay_message_to_duck_pond(message)
```

bot/cogs/duck_pond.py
MarkKoz: Already done so may as well keep it.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,536 @@
...
+        with self.assertLogs(logger=log, level=logging.ERROR) as log_watcher:
+            asyncio.run(self.cog.send_webhook())
+
+        self.assertEqual(len(log_watcher.records), 1)
+
+        [record] = log_watcher.records
+        self.assertEqual(record.message, "Failed to send a message to the Duck Pool webhook")
+        self.assertEqual(record.levelno, logging.ERROR)
+
+    def _get_reaction(
```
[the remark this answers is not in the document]
SebastiaanZ: That would not be my preference.

It creates more visual distance between the helper method and the place where it's being used (which is the test right below the helper method) and this helper method is specifically written for the test that comes just after it. Moving it up would mean having to scroll up and down to understand the test.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,536 @@
...
+        self.cog.webhook = helpers.MockAsyncWebhook()
+        self.cog.webhook.send.side_effect = discord.HTTPException(response=MagicMock(), message="Something failed.")
+
+        log = logging.getLogger('bot.cogs.duck_pond')
+        with self.assertLogs(logger=log, level=logging.ERROR) as log_watcher:
+            asyncio.run(self.cog.send_webhook())
+
+        self.assertEqual(len(log_watcher.records), 1)
+
+        [record] = log_watcher.records
```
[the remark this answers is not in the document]
SebastiaanZ: I've copied the style from other tests that already exist. I'll change it to a style that does not use iterable unpacking.

bot/cogs/duck_pond.py
SebastiaanZ: That's okay with me. I liked the verbosity when reading the name, as it spells out what's going to happen at that point in the code and `relay_message` feels more generic (which this method is not). It doesn't really matter, though, so I'll change it.

tests/bot/cogs/test_duck_pond.py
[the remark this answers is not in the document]
SebastiaanZ: We're trying to assert calls to the mock objects; a reset makes sure that the call history of previous (sub)tests don't carry over to the tests after them.

tests/bot/cogs/test_duck_pond.py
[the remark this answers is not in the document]
SebastiaanZ: Okay, I'll add the bot user to the other reactions.

tests/bot/cogs/test_duck_pond.py
```diff
@@ -0,0 +1,537 @@
...
+            send_webhook.assert_called_once_with(
+                content=message.clean_content,
+                username=message.author.display_name,
+                avatar_url=message.author.avatar_url
+            )
+
+            self.assertEqual(len(log_watcher.records), 1)
+
+            [record] = log_watcher.records
+            self.assertEqual(record.message, "Failed to send an attachment to the webhook")
```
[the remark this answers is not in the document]
SebastiaanZ: I'll remove it.

tests/bot/cogs/test_duck_pond.py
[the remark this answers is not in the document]
lemonsaurus: I hate it but it's fine.

lemonsaurus:
I'mma just merge this, it's gotten plenty of reviews and it's far too well tested that it could possibly break.

Issue #514, opened 2019-10-11
Custom Trash can reaction to replace red X


eivl:
[org-issue-141](https://github.com/python-discord/organisation/issues/141) talks about how certain members are confused about the red X that is added to embeds. 
Like the one that !eval uses
![eval output](https://i.imgur.com/pdSimjS.png)

The consensus is that we change it for a nice trashcan 🗑

MarkKoz:
Ideally we'd have a custom emote that resembles a trash can, but looks nicer than twemoji's.

Numerlor:
I was thinking something like this with simpler lines would look good
![](https://www.shareicon.net/data/48x48/2016/10/11/842408_trash_512x512.png)

lemonsaurus:
yeah, I was also thinking something like that @Numerlor

kraktus:
Duplicate of #297

scragly:
Closing due to duplicate.

Issue #297, opened 2019-01-30
Making the behavior of the red cross emoji consistent for all embed control interfaces


SebastiaanZ:
(As requested by @tagptroll1)

Some of our embeds use an emoji reactions interface to allow the user to interact with them. However, the behavior of one of the emojis, the red cross (:x:), is inconsistent across different type of embeds. In some cases (e.g., `!help`), the cross will have the bot delete the entire embed; in other cases (e.g., `!infr search`), the bot just deletes the reactions/interface, but leaves the embed visible.

It may be a good idea to synchronize the behavior of the red cross (:x:) so the user knows what to expect when clicking it. 

Potential solutions:
- Choose one for all (deleting the entire embed/just deleting the interface)
- Introduce alternative emoji for deletion (e.g., 🗑), keeping ❌ for removal of emoji reactions.

gdude2002:
Trashcan emoji? **Sign me up!**

MarkKoz:
#625 addressed this. it just didn't close properly due to a colon in the "closes #xxx".

PR #625, opened 2019-10-28
Separate message/embed deletion and reaction deletion emojis

kosayoda:
**Closes:** #297 and #514 

Currently, we use a red X for both message/embed deletion and reaction removal. Certain members are hence confused about the meaning of the reaction.

The PR adds a new emoji, `:trashcan:` for deletion of embed or message.
For regular pagination, the red X clears the reactions added by the bot, while the trashcan deletes the embed. The help cog has similar functionality.
image

Replies that use `wait_for_deletion` like `snekbox` now use the trashcan emoji instead of the red X, indicating that it is for message removal.
image

### DIFF
--- a/bot/cogs/filtering.py
+++ b/bot/cogs/filtering.py
@@ -136,9 +136,9 @@ class Filtering(Cog):
             and not msg.author.bot                          # Author not a bot
         )
 
-        # If we're running the bot locally, ignore role whitelist
+        # If we're running the bot locally, ignore role whitelist and only listen to #dev-test
         if DEBUG_MODE:
-            filter_message = not msg.author.bot
+            filter_message = not msg.author.bot and msg.channel.id == Channels.devtest
 
         # If none of the above, we can start filtering.
         if filter_message:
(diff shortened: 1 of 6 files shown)


MarkKoz:
How useful is the ❌ functionality really? I guess it can stop other people from messing with the pagination? I just don't see that get used much and it may be confusing being seen together with a trash can.

kosayoda:
I'm not too sure about that too, there would be cases where you'd not want the emojis there but still want to read the message.

ikuyarihS:
We do automatic reactions removal after a certain period of time iirc, so won't the ❌ be redundant in that case? I can see the confusion between ❌ and `:trashcan:` since both are used to `remove`.

lemonsaurus:
@kosayoda I think it would be clearer if we just removed the :x: now that we're using the trashcan.

@ikuyarihS makes a good point about automatic removals - the reactions should just time out after a while (and get removed when they do). I think they already do this for the paginator.

This PR looks good to merge once that's been addressed.

MarkKoz:
@kosayoda any updates on this?

scragly:

PR #630, opened 2019-10-28
Relay attachments to log channel before filters delete message

### LINKED ISSUE #549: Message filtering did not account for message with only attachments
There was a `mute` issued to an user after posting 5 messages with 5 attachments. The bot issued a `mute` and treated accordingly against message spamming behaviour, however in this case there are two problems:
- They were messages with empty message contents, only attachments.
- The attachments were completely deleted, and there is no way to know what they were.

The log of the case looks like this
```
Deleted by: Python#4329
Date: Oct. 23, 2019, 6:54 a.m.
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
```

The following codes might be of interest:
```py
    """Detects duplicated messages sent by a single user."""
    relevant_messages = tuple(
        msg
        for msg in recent_messages
        if (
            and msg.content == last_message.content
        )
    )

    total_duplicated = len(relevant_messages)

    if total_duplicated > config['max']:
        ...
```
Which is in [duplicates.py](https://github.com/python-discord/bot/blob/master/bot/rules/duplicates.py)

Akarys42:
When filters remove messages with attachments, we are unable to review the deleted attachments for moderation purposes since we do not save them in any way. In the most severe situation, when a filter triggers on messages only containing attachments and no other content, this leads to empty messages being displayed in the deleted messages front-end and not much to go on for moderators. 

This PR solves that issue by relaying the messages to a log channel before deleting the messages. In addition, it will send the link to the new, relayed attachment to the deleted messages end point so mass-deleted messages are displayed as-is in the deleted messages front-end.

Combined with python-discord/site#303, this PR closes #549


Akarys42:
@ikuyarihS @SebastiaanZ I added the re-post of the attachment. I left the ID of mod-log as a placeholder [here in the config file](https://github.com/python-discord/bot/blob/%23549-show-attachments-staff/config-default.yml#L130) so you can change it when you'll have decided where you want them.

bot/cogs/moderation/modlog.py
```diff
@@ -116,6 +118,21 @@ def ignore(self, event: Event, *items: int) -> None:
...
+            channel_id: int = GuildConstant.attachment_repost
+    ) -> t.List[str]:
+        """Re-upload message's attachments to the the channel_id and return the list of re-posted attachments URLs."""
+        channel = self.bot.get_channel(channel_id)
+        out = []
+        for attachment in message.attachments:
+            buffer = BytesIO()
+            await attachment.save(buffer, use_cached=True)
+            reupload = await channel.send(file=discord.File(buffer, filename=attachment.filename))
+            out.append(reupload.attachments[0].url)
```
ikuyarihS: You can use `ContextManager` here to ensure `buffer` will close properly, like this
```py
            with BytesIO() as buffer:
                await attachment.save(buffer, use_cached=True)
                reupload: discord.Message = await channel.send(file=discord.File(buffer, filename=attachment.filename))
                out.append(reupload.attachments[0].url)
```

bot/cogs/antispam.py
```diff
@@ -243,9 +242,6 @@ def mod_log(self) -> ModLog:
 
     async def _process_deletion_context(self, context_id: int) -> None:
         """Processes the Deletion Context queue."""
-        log.trace("Sleeping before processing message deletion queue.")
-        await asyncio.sleep(10)
-
```
ikuyarihS: I'm not sure this part should be deleted, as discussed in [#dev-contrib](https://discordapp.com/channels/267624335836053506/635950537262759947/644820856958222338) I believe we should wait for response from @SebastiaanZ to confirm what it is doing first.

Akarys42: Comment from @SebastiaanZ in #dev-contrib:
> It should not wait 10 seconds to delete the messages after triggering, but it should take action as soon as possible
> However, since you are going to relay the attachments, that's something that could happen earlier
> The DeletionContext tracks the messages that should be deleted
> The `add` method of that class seems like a good place to relay the attachments and save the links to those
> That happens before the messages are deleted

I’m going to move it ASAP.

bot/cogs/moderation/modlog.py
```diff
@@ -116,6 +118,21 @@ def ignore(self, event: Event, *items: int) -> None:
 
         return await self.bot.get_context(log_message)  # Optionally return for use with antispam
 
+    async def reupload_attachments(
+            self,
+            message: discord.Message,
+            channel_id: int = GuildConstant.attachment_repost
```
MarkKoz: The parameters are indented too far.

bot/cogs/moderation/modlog.py
```diff
@@ -116,6 +118,21 @@ def ignore(self, event: Event, *items: int) -> None:
...
+    ) -> t.List[str]:
+        """Re-upload message's attachments to the the channel_id and return the list of re-posted attachments URLs."""
+        channel = self.bot.get_channel(channel_id)
+        out = []
+        for attachment in message.attachments:
+            with BytesIO() as buffer:
+                await attachment.save(buffer, use_cached=True)
```
MarkKoz: Shouldn't there be some error handling for failed saves? According to the docs, the `proxy_url` may still fail.

Akarys42: Well I'm going to move this function so there shouldn't be a case where the `proxy_url` fail, but if we get a traceback, we don't really care, because we just abort the re-upload.

MarkKoz: I would think it's better to try to upload as many attachments as possible rather than abort when any fail.

Akarys42: Anyway, the re-upload will take place before the message deletion, so there shouldn’t be any reason for the re-upload to fail, and if there is, we should better log it and continue re-uploading isn’t?

MarkKoz: Yeah, just a simple log is fine.

SebastiaanZ:
@Akarys42 We've created a new channel to relay attachments to: `#attachment-log` with the ID `649243850006855680`.

Akarys42:
@ikuyarihS @MarkKoz @SebastiaanZ all requested changes have been adressed

bot/cogs/antispam.py
```diff
@@ -254,6 +263,27 @@ def mod_log(self) -> ModLog:
...
+    channel = bot.get_channel(channel_id)
+    out = []
+    for attachment in message.attachments:
+        try:
+            with BytesIO() as buffer:
+                await attachment.save(buffer, use_cached=True)
+                reupload: Message = await channel.send(file=File(buffer, filename=attachment.filename))
+                out.append(reupload.attachments[0].url)
+        except (HTTPException, NotFound):
+            log.warning(f"Tried to re-upload attchment {attachment.id}, but it has failed.")
```
MarkKoz:
```suggestion
            log.warning(f"Tried to re-upload attachment {attachment.id}, but it has failed.")
```

bot/cogs/moderation/modlog.py
```diff
@@ -53,9 +61,10 @@ def __init__(self, bot: Bot):
                         'author': message.author.id,
                         'channel_id': message.channel.id,
                         'content': message.content,
-                        'embeds': [embed.to_dict() for embed in message.embeds]
+                        'embeds': [embed.to_dict() for embed in message.embeds],
+                        'attachments': attachment,
                     }
-                    for message in messages
+                    for message, attachment in zip(messages, attachments)
```
MarkKoz: This would need to be  `zip_longest` I believe.

bot/cogs/antispam.py
```diff
@@ -254,6 +263,27 @@ def mod_log(self) -> ModLog:
...
+        return []
+    channel = bot.get_channel(channel_id)
+    out = []
+    for attachment in message.attachments:
+        try:
+            with BytesIO() as buffer:
+                await attachment.save(buffer, use_cached=True)
+                reupload: Message = await channel.send(file=File(buffer, filename=attachment.filename))
+                out.append(reupload.attachments[0].url)
+        except (HTTPException, NotFound):
```
MarkKoz: Since `NotFound` subclasses `HTTPException`, just specifying `HTTPException` is adequate.

bot/cogs/antispam.py
MarkKoz: I think an attachment ID is meaningless information, at least on its own. It'd be helpful to include the message's ID too and/or the filename.

bot/cogs/moderation/modlog.py
Akarys42: No, because `reupload_attachments()` return an empty list if there isn't any message attached

MarkKoz: Is that not exactly why it _would_ need to be `zip_longest`? If the attachments are empty then the resulting zip is empty too:

```py
>>> messages = [1, 2, 3]
>>> attachments = []
>>> list(zip(messages, attachments))
[]
```

MarkKoz: I think I misunderstood you. Did you mean if attachments are empty it means the messages would also be empty anyway?

Akarys42: AFAIK `messages` will always be a list of `Message` and `attachments` a list of list (which could be empty) so I don’t think we need `zip_longest` since `messages` and `attachments` will always have the same length

config-default.yml
```diff
@@ -134,6 +134,7 @@ guild:
 
     staff_channels: [*ADMINS, *ADMIN_SPAM, *MOD_SPAM, *MODS, *HELPERS, *ORGANISATION, *DEFCON]
     ignored: [*ADMINS, *MESSAGE_LOG, *MODLOG]
+    attachment_repost: 649243850006855680
```
MarkKoz: This should be under the channels with rest of the channel IDs.

scragly: Agree, and it should be named `attachment_log` to match the in-server channel name.

Akarys42: Yep, it was here because it used an anchor node previously, gonna move it

scragly: Oh, and it's best to add the channel to the `ignored` list just above here

config-default.yml
```diff
@@ -138,7 +139,7 @@ guild:
 
     staff_channels: [*ADMINS, *ADMIN_SPAM, *MOD_SPAM, *MODS, *HELPERS, *ORGANISATION, *DEFCON]
     ignored: [*ADMINS, *MESSAGE_LOG, *MODLOG]
-    attachment_repost: 649243850006855680
+    attachment_repost: *ATTCH_LOG
```
MarkKoz: This is redundant. It should be removed and replaced by `attachment_log` entirely in the code.

config-default.yml
```diff
@@ -138,7 +139,7 @@ guild:
 
     staff_channels: [*ADMINS, *ADMIN_SPAM, *MOD_SPAM, *MODS, *HELPERS, *ORGANISATION, *DEFCON]
     ignored: [*ADMINS, *MESSAGE_LOG, *MODLOG]
```
MarkKoz: As Scragly suggested, ignore the attachments channel here.

bot/cogs/antispam.py
```diff
@@ -255,10 +260,10 @@ def mod_log(self) -> ModLog:
         await deletion_context.upload_messages(self.bot.user.id, self.mod_log)
 
 
-def validate_config(rules: Mapping = AntiSpamConfig.rules) -> Dict[str, str]:
+def validate_config(_rules: Mapping = AntiSpamConfig.rules) -> Dict[str, str]:
```
scragly:
```suggestion
def validate_config(rules_: Mapping = AntiSpamConfig.rules) -> Dict[str, str]:
```

bot/cogs/antispam.py
```diff
@@ -255,10 +260,10 @@ def mod_log(self) -> ModLog:
...
     """Validates the antispam configs."""
     validation_errors = {}
-    for name, config in rules.items():
+    for name, config in _rules.items():
```
scragly:
```suggestion
    for name, config in rules_.items():
```

PR #633, opened 2019-10-29
Add unit tests for bot/rules/newlines.py

alberthdev:
Add some unit tests to validate behavior in bot/rules/newlines.py.
This covers both overall newline exceeding behavior, as well as
messages containing consecutive newlines within each message.

This test suite mimicks existing tests for attachment rules, as their
requirements are very similar.

Addresses: #602

### DIFF
--- a/tests/bot/rules/test_newlines.py
+++ b/tests/bot/rules/test_newlines.py
@@ -1,0 +1,79 @@
+import asyncio
+import unittest
+from dataclasses import dataclass
+from typing import Optional
+
+from bot.rules import newlines
+
+
+# Using `MagicMock` sadly doesn't work for this usecase
+# since it's __eq__ compares the MagicMock's ID. We just
+# want to compare the actual attributes we set.
+@dataclass
+class FakeMessage:
+    author: str
+    content: str
+
+
+def msg(new_lines: int, consecutive_new_lines: int, base_content: Optional[str] = "sample msg") -> FakeMessage:
+    content = base_content
+    content += "a\n" * new_lines
+    content += "\n" * consecutive_new_lines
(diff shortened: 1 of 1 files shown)


alberthdev:
Thanks for looking this over and providing feedback! Agreed that we could clarify the tests, and that we don't really test the case where the authors are different. Will take a look at updating these.

On a side note, I think it's probably worth testing the different user case as well in test_attachments.py - can update that as well if you want, though not sure if that should go into this PR or another one.

MarkKoz:
I think it's OK if you also update the attachment tests in this PR.

kwzrd:
I've some changes to the attachments test awaiting review in #655, so you might want to look at that if you're still planning to update the test cases. Sorry, did not realize you were planning to look at it too.

lemonsaurus:
if @kwzrd is looking at the attachment tests, then there's no sense in having @alberthdev do it as well. Besides, it's been a month since there's been any activity here. Let's proceed with this PR without expanding its scope.

lemonsaurus:
@alberthdev when will you have time to look at addressing the requests from @MarkKoz  ?

alberthdev:
Have been busy with other things recently, but I hope to get back to this sometime soon. Hopefully by the end of the year.

kwzrd:
Hi @alberthdev, could we get a status update on this?

kwzrd:
Hello @alberthdev, since this PR has been stale for months now and we haven't been able to establish communication, I've decided to close it. There is a concurrent PR #733, which implements unit tests for all other antispam rules, and I think #602 can be addressed there as well.

I hope this is ok with you.

PR #641, opened 2019-11-01
Add unit test for links antispam rule

kwzrd:
In response to #600.

Provides 100% coverage. I will welcome any feedback.

### DIFF
--- a/tests/bot/rules/test_links.py
+++ b/tests/bot/rules/test_links.py
@@ -39,6 +39,7 @@ class LinksTests(unittest.TestCase):
             [msg("bob", 0)],
             [msg("bob", 2)],
             [msg("bob", 3)],
+            [msg("bob", 1), msg("bob", 1)],
             [msg("bob", 3), msg("alice", 3)]
         )
 
@@ -61,6 +62,12 @@ class LinksTests(unittest.TestCase):
                 [msg("bob", 1), msg("bob", 2)],
                 (msg("bob", 1), msg("bob", 2)),
                 ("bob",),
+                3
+            ),
+            Case(
+                [msg("alice", 1), msg("alice", 1), msg("alice", 1)],
+                (msg("alice", 1), msg("alice", 1), msg("alice", 1)),
+                ("alice",),
                 3
             ),
             Case(
--- a/tests/bot/rules/test_links.py
+++ b/tests/bot/rules/test_links.py
@@ -1,0 +1,93 @@
+import unittest
+from typing import List, NamedTuple, Tuple
+
+from bot.rules import links
+from tests.helpers import async_test
+
+
+class FakeMessage(NamedTuple):
+    author: str
+    content: str
+
+
+class Case(NamedTuple):
+    recent_messages: List[FakeMessage]
+    relevant_messages: Tuple[FakeMessage]
+    culprit: Tuple[str]
+    total_links: int
+
+
+def msg(author: str, total_links: int) -> FakeMessage:
+    """Makes a message with `total_links` links."""
+    content = " ".join(["https://pydis.com"] * total_links)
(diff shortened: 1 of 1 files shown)


tests/bot/rules/test_links.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @async_test
+    async def test_links_within_limit(self):
+        """Messages with an allowed amount of links."""
+        cases = (
+            [msg("bob", 0)],
+            [msg("bob", 2)],
+            [msg("bob", 3)],
+            [msg("bob", 1), msg("bob", 1)],
+            [msg("bob", 3), msg("alice", 3)]
```
MarkKoz: I don't understand how this case passes the filter if the config has `max` set to 2 rather than 3.

MarkKoz: Never mind, it's because `messages_with_links > 1`

kwzrd: The case can be annotated with an inline comment, do you think that would be helpful?

MarkKoz: Yes that would actually be nice. There's also a similar situation with line 41 so perhaps move them together and put a comment above.

kwzrd: I annotated them separately as I intended the last one to test that the filter will only count links from the latest author. Let me know if you think another way would be better.

MarkKoz: Then I think it's better to make that last case have only 2 links per message so it has more focus and less points of failure.

kwzrd: I agree, this is a good point. Adjusted in latest commit.

tests/bot/rules/test_links.py
```diff
@@ -0,0 +1,101 @@
...
+
+class Case(NamedTuple):
+    recent_messages: List[FakeMessage]
+    relevant_messages: Tuple[FakeMessage]
+    culprit: Tuple[str]
+    total_links: int
+
+
+def msg(author: str, total_links: int) -> FakeMessage:
+    """Makes a message with `total_links` links."""
```
aeros: Minor formatting nit; argument names should be in asterisks.
```suggestion
    """Makes a message with *total_links* links."""
```

kwzrd: This is new to me, thank you! Committed in https://github.com/python-discord/bot/pull/641/commits/760f265339b1462b60ea7d5af1fd8d4476c134f0.

Numerlor: args are in back ticks in all the docstrings I encountered so far so I think it'd be better to just be consistent

lemonsaurus: it is true that backticks is the prevailing style for this in our repo, and that until we decide to change that, it's probably best to keep doing that. I've never really seen the asterisks thing, and afaik there's nothing about this in PEP257. Can I ask where you picked up that convention @aeros?

aeros: @lemonsaurus 
> Can I ask where you picked up that convention @aeros, or what the reasoning behind it is?

We follow that convention in the CPython docs and docstrings. I don't know the specific reason behind it, but we use backticks for functions or methods, and asterisks for arg names. For example:

```
`loop.run_in_executor()`
*executor*
```

Backticks for argument names as well though is probably fine, as long as you're consistent with it.

aeros: I'll do some poking around, to see if I can find an actual reason for it or if it's mentioned in a PEP somewhere.

lemonsaurus: Interesting. I suppose I can see the usefulness in denoting callables one way and arguments another, what symbol is used might just be arbitrary.

Thanks for the info.

PR #643, opened 2019-11-02
implement !zen command.

### LINKED ISSUE #642: create command for getting parts of the zen of python
it would be nice to be able to get specific parts of the zen of python from a command. for example,

`!zen 13` would retrieve the fourteenth line (because we index from zero, obviously).

`!zen namespaces` would search for the string "namespaces", and produce the line which matches best. 

`!zen` without any arguments would still produce the entire thing in the same way that the `zen` tag does.

i think this is reasonably simple, and could come in handy from time to time. :D

edit: also, i'd quite like to implement this. seems simple enough while i get used to all of the new changes since last year :D

kingdom5500:
resolves #642.

image


bot/cogs/utils.py
```diff
@@ -174,6 +198,76 @@ def check(m: Message) -> bool:
...
+        if search_value.lstrip("-").isdigit():
+            index = int(search_value)
+
+            try:
+                line = zen_lines[index]
+            except IndexError:
+                embed = Embed(
+                    colour=Colour.red(),
+                    title=random.choice(NEGATIVE_REPLIES),
+                    description="Please provide a valid index."
```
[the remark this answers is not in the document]
kingdom5500: can do.
image

bot/cogs/utils.py
```diff
@@ -174,6 +198,78 @@ def check(m: Message) -> bool:
...
+
+        # check if it's an integer. could be negative. why not.
+        is_negative_integer = search_value[0] == "-" and search_value[1:].isdigit()
+        if search_value.isdigit() or is_negative_integer:
+            index = int(search_value)
+
+            try:
+                line = zen_lines[index]
+            except IndexError:
+                embed = Embed(
```
MarkKoz: Raise a `BadArgument` with a message instead and let the error handler cog take care of the rest.

bot/cogs/utils.py
```diff
@@ -174,6 +198,78 @@ def check(m: Message) -> bool:
...
+
+            await ctx.send(embed=embed)
+            return
+
+        zen_lines = ZEN_OF_PYTHON.splitlines()
+
+        # check if it's an integer. could be negative. why not.
+        is_negative_integer = search_value[0] == "-" and search_value[1:].isdigit()
+        if search_value.isdigit() or is_negative_integer:
+            index = int(search_value)
```
MarkKoz: Is it possible to make discord.py do the conversion for you? Something like `search_value: typing.Union[int, str, None]` perhaps. Or you could make a custom converter and validate the index range inside it too.

MarkKoz:
@kingdom5500 any updates on this?

scragly:

bot/cogs/utils.py
scragly: It's now been adjusted to use converters as suggested.

bot/cogs/utils.py
scragly: This section has been tidied a little and now raises a BadArgument for the global error handler to deal with.

bot/cogs/utils.py
```diff
@@ -174,6 +198,76 @@ def check(m: Message) -> bool:
...
+                description=ZEN_OF_PYTHON
+            )
+
+            return await ctx.send(embed=embed)
+
+        zen_lines = ZEN_OF_PYTHON.splitlines()
+
+        # check if it's an integer. could be negative. why not.
+        if search_value.lstrip("-").isdigit():
+            index = int(search_value)
```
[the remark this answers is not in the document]
scragly: After the adjustment to use converters, this no longer raises a ValueError, as it just falls back to being taken as a string.

bot/cogs/utils.py
```diff
@@ -173,6 +196,64 @@ def check(m: Message) -> bool:
...
+
+            # the match ratio needs to be adjusted because, naturally,
+            # longer lines will have worse ratios than shorter lines when
+            # fuzzy searching for keywords. this seems to work okay.
+            adjusted_ratio = (len(line) - 5) ** 0.5 * matcher.ratio()
+
+            if adjusted_ratio > best_ratio:
+                best_ratio = adjusted_ratio
+                best_match = line
+                match_index = index
```
SebastiaanZ: Shouldn't we use a threshold below which we say "Sorry, I couldn't find anything matching you query."? The current mechanism *always* returns something, even if the string you specify has nothing to do with any of the lines in the zen of Python. 

It does lead to interesting responses.

scragly: I actually agree with you, and I pondered over it for a while. A particularly nice one is:
image

In the end, it's become obvious that we need to implement a decent difflib based partial matching solution that's used as a utility across the entire project(s), and honestly I don't find it absolutely essential for it to be implemented in this PR to get it merged and dealt with for later enhancement.

SebastiaanZ: That's fair. We're now using fuzzy matching in a lot of different places.

PR #646, opened 2019-11-03
Change `on_message_edit` - only show differences, highlighted in blue with hyperlink

### LINKED ISSUE #632: Showing what changed from `on_message_edit`
Right now, for each message edited, both before & after edited message are posted as embed, separately, in full `.content`

Some suggestions to make this better:

1. Merge both embeds and use a jump link to the after edited message
    - Since the message is not yet deleted, it is still in the server, so getting the full content of the message is just a *click away*.
2. Show what changed
    - It is more valuable to know what changed after the edit rather than the full content before and after, since the message was not deleted *yet*.

### Some examples:

Scragly's suggestion - clean, compact, using hyperlink:
![Scragly's][example 1]

Example only using `difflib`:
![Shirayuki's][example 2]

Example implementing Scragly's suggestion:
![Scragly's idea][example 3]

Example of longer messages:
![long edit][example 4]

[example 1]: https://cdn.discordapp.com/attachments/305126844661760000/638557082874019850/unknown.png
[example 2]: https://cdn.discordapp.com/attachments/305126844661760000/638557938100994069/unknown.png
[example 3]: 
https://user-images.githubusercontent.com/19921765/67757307-2411be80-fa6e-11e9-99a3-73f129122c70.png
[example 4]: 
https://user-images.githubusercontent.com/19921765/67757354-3d1a6f80-fa6e-11e9-8064-e44d91b7c0df.png

ikuyarihS:
Closes #632

#### Before edit:
`There's clearly an easier way to handle this particular data by not using a relational db, but with how much of it does work with one, I'm having trouble judging if that solution is fine, or needs other handling.`
#### After edit:
`There's clearly an easier way to handle this data by not using a relational db, but with how much of it does work extremely well with one, I'm having trouble judging if that solution is acceptable, or needs other handling.`

Embed will then show this:
image

### DIFF
--- a/bot/cogs/moderation/modlog.py
+++ b/bot/cogs/moderation/modlog.py
@@ -674,7 +674,7 @@ class ModLog(Cog, name="ModLog"):
             f"**Before**:\n{' '.join(content_before)}\n"
             f"**After**:\n{' '.join(content_after)}\n"
             "\n"
-            f"[jump to message]({msg_after.jump_url})"
+            f"[Jump to message]({msg_after.jump_url})"
         )
 
         if msg_before.edited_at:
--- a/bot/cogs/moderation/modlog.py
+++ b/bot/cogs/moderation/modlog.py
@@ -653,9 +653,9 @@ class ModLog(Cog, name="ModLog"):
         for index, (diff_type, words) in enumerate(diff_groups):
             sub = ' '.join(words)
             if diff_type == '-':
-                _before = _before.replace(sub, f"[{sub}](http://.z)")
+                _before = _before.replace(sub, f"[{sub}](http://o.hi)")
             elif diff_type == '+':
-                _after = _after.replace(sub, f"[{sub}](http://.z)")
+                _after = _after.replace(sub, f"[{sub}](http://o.hi)")
             else:
                 if len(words) > 2:
                     new = (f"{words[0] if index > 0 else ''}"
--- a/bot/cogs/moderation/modlog.py
+++ b/bot/cogs/moderation/modlog.py
@@ -641,20 +641,26 @@ class ModLog(Cog, name="ModLog"):
 
         _before = before.clean_content
         _after = after.clean_content
-        groups = tuple((g[0], tuple(g[1]))
-                       for g in itertools.groupby(difflib.ndiff(_before.split(), _after.split()), key=lambda s: s[0]))
 
-        for index, (name, values) in enumerate(groups):
-            sub = ' '.join(s[2:].strip() for s in values)
-            if name == '-':
+        # Getting the difference per words and group them by type - add, remove, same
+        # Note that this is intended grouping without sorting
+        diff = difflib.ndiff(_before.split(), _after.split())
+        diff_groups = tuple(
+            (diff_type, tuple(s[2:] for s in diff_words))
+            for diff_type, diff_words in itertools.groupby(diff, key=lambda s: s[0])
+        )
+
+        for index, (diff_type, words) in enumerate(diff_groups):
+            sub = ' '.join(words)
+            if diff_type == '-':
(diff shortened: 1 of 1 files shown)


bot/cogs/moderation/modlog.py
```diff
@@ -635,39 +637,43 @@ def ignore(self, event: Event, *items: int) -> None:
...
-                f"{after.clean_content}"
-            )
+        for index, (diff_type, words) in enumerate(diff_groups):
+            sub = ' '.join(words)
+            if diff_type == '-':
+                _before = _before.replace(sub, f"[{sub}](http://o.hi)")
+            elif diff_type == '+':
+                _after = _after.replace(sub, f"[{sub}](http://o.hi)")
+            else:
+                if len(words) > 2:
```
MarkKoz: Can be merged with previous line to `elif`.

bot/cogs/moderation/modlog.py
```diff
@@ -635,39 +637,43 @@ def ignore(self, event: Event, *items: int) -> None:
...
+                    new = (f"{words[0] if index > 0 else ''}"
+                           " ... "
+                           f"{words[-1] if index < len(diff_groups) - 1 else ''}")
```
MarkKoz: Style should be like:

```py
new = (
    "hello"
    "world"
)
```

ikuyarihS: Thank you, I will definitely stick to this style for more consistency.

bot/cogs/moderation/modlog.py
ikuyarihS: That's a great call, I forgot about it after I removed the `else` of the `if`

MarkKoz:
This doesn't work quite well in this case:

bild

ikuyarihS:
It's more of a logic bug, since I'm using `str.replace()` and the deleted word has a length of 1, chances are it will hyperlink more than just one. I'll take a look, if I remember correctly difflib can return the index as well.

bot/cogs/moderation/modlog.py
```diff
@@ -635,39 +637,45 @@ def ignore(self, event: Event, *items: int) -> None:
...
-
-            after_response = (
-                f"**Author:** {author} (`{author.id}`)\n"
-                f"**Channel:** #{channel.name} (`{channel.id}`)\n"
-                f"**Message ID:** `{before.id}`\n"
-                "\n"
-                f"{after.clean_content}"
-            )
+        _before = []
+        _after = []
```
scragly: I's suggest more explicit naming for these and the message_edit args so it's clear what type of object we're working with.

`_before` could be `content_before`
`before` could be `msg_before`
Same for `_after` and `after`.

ikuyarihS: I like the way you name them, `msg_before` and `content_before` both are good name, and explicit, so I am going to steal them!

lemonsaurus:
I love this idea, very creative.

bot/cogs/moderation/modlog.py
```diff
@@ -618,80 +620,81 @@ def ignore(self, event: Event, *items: int) -> None:
...
+
+        response = (
+            f"**Author:** {author} (`{author.id}`)\n"
+            f"**Channel:** {channel_name} (`{channel.id}`)\n"
+            f"**Message ID:** `{msg_before.id}`\n"
+            "\n"
+            f"**Before**:\n{' '.join(content_before)}\n"
+            f"**After**:\n{' '.join(content_after)}\n"
+            "\n"
+            f"[jump to message]({msg_after.jump_url})"
```
lemonsaurus:
```suggestion
            f"[Jump to message]({msg_after.jump_url})"
```

PR #655, opened 2019-11-09
Antispam rules: add unit test for mentions, improve unit tests for links and attachments, fix bug in attachments rule

### LINKED ISSUE #601: Write unit tests for `bot/rules/mentions.py`
Write unit tests for [`bot/rules/mentions.py`](../blob/master/bot/rules/mentions.py).

## Implementation details

## Additional information

kwzrd:
This PR addresses several problems.

- [x] Implement unit test for `bot.rules.mentions`, using the `MockMessage` helper to correctly represent mocked `discord.Message` instances. Closes #601. Gives 100% coverage. See 
0eade46.
- [x] Adjust unit tests for `bot.rules.links` and `bot.rules.attachments` to use `MockMessage` helper instead of a `namedtuple("FakeMessage", ["author", "content"])` that was a nasty oversimplification of actual Message instances. After #660 the mocking system is now more safe as it prevents setting attributes that do not exist for the mocked objects. See 2a9b1fc and a2617d1.
- [x] Small docstring adjustment - use backticks instead of asterisks when referring to args. See 32caead.

Additionally, we have discovered that there is an inconsistency / bug in the behaviour of `bot.rules.attachments`. Snippets below explain.

All rules have the same function signature, taking the following args:

```python
last_message: Message, recent_messages: List[Message], config: Dict[str, int]
```

The [antispam cog](https://github.com/python-discord/bot/blob/c04df6472b2450d273475df94fa4917116b451e0/bot/cogs/antispam.py#L171) calls them in `on_message` by providing the `message` that triggers the event, and a history of messages up to a certain point in time. This history, however, includes the `message` instance itself. So let's say `last_message == recent_messages[0]`. Within the rule, the first thing we do is build a tuple `relevant_messages` which is made up of all messages _m_ where `m.author == last_message.author`, and so this is the only way the `last_message` arg is used - to see which user we're checking (slight deviation in burst rule which applies to all users):

```python
relevant_messages = tuple(
    msg
    for msg in recent_messages
    if msg.author == last_message.author
)
```

The `attachments` rule however deviates from this pattern by always including the `last_message` in `relevant_messages`, _and then_ looping over `recent_messages`. The result of this is that the `last_message` potentially gets included twice, and its attachments then count twice in the sum.

```python
relevant_messages = [last_message] + [
    msg
    for msg in recent_messages
    if (
        and len(msg.attachments) > 0
    )
]
```

Therefore, a test case that has `last_message == recent_messages[0]` will fail, for example

```python
last_message = msg(4)
recent_messages = [msg(4), msg(0), msg(6)]
expected_total = 10
```

... fails with ...

```python
- ('sent 14 attachments in 5s',
?         ^

+ ('sent 10 attachments in 5s',
?         ^
```

... and also the `relevant_messages` containing a duplicate.

So why wasn't the rule failing on the [unit test written for it](https://github.com/python-discord/bot/blob/c04df6472b2450d273475df94fa4917116b451e0/tests/bot/rules/test_attachments.py)?

The unit test incorrectly makes the assumption [here](https://github.com/python-discord/bot/blob/c04df6472b2450d273475df94fa4917116b451e0/tests/bot/rules/test_attachments.py#L33) and [here](https://github.com/python-discord/bot/blob/c04df6472b2450d273475df94fa4917116b451e0/tests/bot/rules/test_attachments.py#L47) that `last_message` is not present in `recent_messages`, therefore making the same mistake as the rule itself.

This PR therefore also contains the following.

- [x] Fix the `attachments` rule to understand `last_message` to be part of `recent_messages` and build `relevant_messages` accordingly. See 54598dd.
- [x] Adjust the unit test to emulate the behaviour of the `on_message` event calling the rule. See eef447a.
- [x] Add slightly more complex test cases. Test the `attachments` rule against multiple authors, not just a hardcoded `lemon`. See 01731a8.

Always happy to hear any feedback.

### DIFF
--- a/tests/bot/rules/test_attachments.py
+++ b/tests/bot/rules/test_attachments.py
@@ -12,13 +12,16 @@ def msg(total_attachments: int) -> MockMessage:
 class AttachmentRuleTests(unittest.TestCase):
     """Tests applying the `attachments` antispam rule."""
 
+    def setUp(self):
+        self.config = {"max": 5}
+
     @async_test
     async def test_allows_messages_without_too_many_attachments(self):
         """Messages without too many attachments are allowed as-is."""
         cases = (
-            (msg(0), msg(0), msg(0)),
-            (msg(2), msg(2)),
-            (msg(0),),
+            [msg(0), msg(0), msg(0)],
+            [msg(2), msg(2)],
+            [msg(0)],
         )
 
         for recent_messages in cases:
@@ -29,7 +32,7 @@ class AttachmentRuleTests(unittest.TestCase):
                 recent_messages=recent_messages
             ):
                 self.assertIsNone(
-                    await attachments.apply(last_message, recent_messages, {'max': 5})
+                    await attachments.apply(last_message, recent_messages, self.config)
                 )
 
     @async_test
@@ -59,6 +62,6 @@ class AttachmentRuleTests(unittest.TestCase):
                 total=total
             ):
                 self.assertEqual(
-                    await attachments.apply(last_message, recent_messages, {'max': 5}),
+                    await attachments.apply(last_message, recent_messages, self.config),
                     (f"sent {total} attachments in 5s", ('lemon',), relevant_messages)
                 )
--- a/tests/bot/rules/test_attachments.py
+++ b/tests/bot/rules/test_attachments.py
@@ -1,8 +1,7 @@
-import asyncio
 import unittest
 
 from bot.rules import attachments
-from tests.helpers import MockMessage
+from tests.helpers import MockMessage, async_test
 
 
 def msg(total_attachments: int) -> MockMessage:
@@ -13,7 +12,8 @@ def msg(total_attachments: int) -> MockMessage:
 class AttachmentRuleTests(unittest.TestCase):
     """Tests applying the `attachments` antispam rule."""
 
-    def test_allows_messages_without_too_many_attachments(self):
+    @async_test
+    async def test_allows_messages_without_too_many_attachments(self):
         """Messages without too many attachments are allowed as-is."""
         cases = (
             (msg(0), msg(0), msg(0)),
@@ -22,17 +22,23 @@ class AttachmentRuleTests(unittest.TestCase):
         )
 
         for last_message, *recent_messages in cases:
-            with self.subTest(last_message=last_message, recent_messages=recent_messages):
-                coro = attachments.apply(last_message, recent_messages, {'max': 5})
-                self.assertIsNone(asyncio.run(coro))
+            with self.subTest(
+                last_message=last_message,
+                recent_messages=recent_messages
+            ):
+                self.assertIsNone(
+                    await attachments.apply(last_message, recent_messages, {'max': 5})
+                )
 
-    def test_disallows_messages_with_too_many_attachments(self):
+    @async_test
+    async def test_disallows_messages_with_too_many_attachments(self):
         """Messages with too many attachments trigger the rule."""
         cases = (
             ([msg(4), msg(0), msg(6)], 10),
             ([msg(6)], 6),
             ([msg(1)] * 6, 6),
         )
+
         for messages, total in cases:
             last_message, *recent_messages = messages
             relevant_messages = [last_message] + [
@@ -48,8 +54,7 @@ class AttachmentRuleTests(unittest.TestCase):
                 relevant_messages=relevant_messages,
                 total=total
             ):
-                coro = attachments.apply(last_message, recent_messages, {'max': 5})
                 self.assertEqual(
-                    asyncio.run(coro),
+                    await attachments.apply(last_message, recent_messages, {'max': 5}),
                     (f"sent {total} attachments in 5s", ('lemon',), relevant_messages)
                 )
--- a/tests/bot/rules/test_attachments.py
+++ b/tests/bot/rules/test_attachments.py
@@ -21,7 +21,9 @@ class AttachmentRuleTests(unittest.TestCase):
             (msg(0),),
         )
 
-        for last_message, *recent_messages in cases:
+        for recent_messages in cases:
+            last_message = recent_messages[0]
+
             with self.subTest(
                 last_message=last_message,
                 recent_messages=recent_messages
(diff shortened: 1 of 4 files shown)


tests/bot/rules/test_mentions.py
```diff
@@ -0,0 +1,99 @@
...
+from typing import List, NamedTuple, Tuple
+
+from bot.rules import mentions
+from tests.helpers import async_test
+
+
+class FakeMessage(NamedTuple):
+    author: str
+    mentions: List[None]
+
```
SebastiaanZ: Why are you not using the `MockMessage` defined in [`tests.helpers`](../../blob/master/tests/helpers.py#L332)? If it's not important to have a `MockMember` for author attribute, you can just provide a string for the author field like you do now: `MockMessage(author="bob", mentions=list(range(3)))` would work and create a mock that follows the specifications of an actual `discord.Message` object.

kwzrd: I'm not sure if that's desirable though, because in that case the `__eq__` will [look at the Message's id](https://github.com/Rapptz/discord.py/blob/4c3e53edf4ff90e3fcf195c429601a553d661cd8/discord/mixins.py#L31). So we would need to be manually controlling the `id` attribute to correctly compare the output tuples. Using a `FakeMessage`, we simply compare them based on the attribs we need for the test. I believe using `MockMessage` would introduce needless complexity as a result.

This idea is further explained in [the unit test for the attachments rule](https://github.com/python-discord/bot/blob/3953aad9c21e5593a980ed996f54dccdbfac36dd/tests/bot/rules/test_attachments.py#L9) which I draw a lot of inspiration from.

scragly: [`discord.Message`](https://github.com/Rapptz/discord.py/blob/4c3e53edf4ff90e3fcf195c429601a553d661cd8/discord/message.py#L176) does not inherit from [`EqualityComparable`](https://github.com/Rapptz/discord.py/blob/4c3e53edf4ff90e3fcf195c429601a553d661cd8/discord/mixins.py#L27).

SebastiaanZ: I'm a bit torn on the complexity argument here.

One of the reasons we use our custom mock objects is because they follow the specifications of an actual instance of the `discord.py` they are mocking. This means that if a version upgrade in `discord.py` changes the API of the class/object in a way that's relevant to the code you're testing, the tests will now fail because the code it's testing is trying to use attributes that no longer exist on the actual object.

By creating a static object that is a snapshot of the current API, you bypass that spec safeguard of our mock objects: Tests will not show that the code is using no-longer-existing attributes, because your `FakeMessage` happily provides an object with the old attributes for the test.

Another point is that we're now back to creating multiple very similar classes to mock the same object in a lot of different files while we have a single helper mock that could be used for all of them. I'm not sure what I find more complex in the end: Having a lot of slightly different classes in a lot of files or a single mock type that works like the actual object we're trying to mock.

Granted, the current mock we have does not implement an `__eq__` method, which makes it so that our mocks are not compared in a similar way to actual message objects. I think a good way forward is implementing a `__eq__` similar to the `__eq__` the `discord.Message` inherits and modifying the `__init__` to make sure we set an `id` at initialization. We could even an auto-incremental ID to make sure each new mock gets a new message id automatically, which will mean we don't have to worry about controlling the `id` in individual tests, unless we want it to be an exact value.

---
Just as a clarification of how the mocks work:

 > because in that case the __eq__ will look at the Message's id.

This not actually the case. The mocks follow the specifications of the `discord.py`, which means that trying to access attributes that do not exist on the original will result in an `AttributeError`. They do not delegate the methods to the original object or subclass the original object, so using an equality operator will not result in the `__eq__` method of the original `discord.py` class being called. That means we'd have to indeed control the `__eq__` and that's not really necessary here.

kwzrd: @scragly Yes, sorry about that, I should have checked before assuming. It looks like `discord.Message` does not have a custom `__eq__` implementation, but I suppose the point is that the `id` obviously plays a role.

@SebastiaanZ Many thanks for the feedback, I will adjust the test cases as per our discussion on discord.

kwzrd: In https://github.com/python-discord/bot/pull/655/commits/16e6c36c6b370300ef7507d2e01421ad0d83f407, `tests.helpers.MockMessage` is used as suggested. Passing an `id` (message id, that is) manually actually ends up being unnecessary, as the `__eq__` falls back on identity comparison. The test cases just needed to be adjusted so that multiple instances aren't used to represent one message.

Let me know what you think. If this looks good, I will adjust the other rule unit tests to use this methodology.

kwzrd: @SebastiaanZ I'm marking this as resolved since we're now using `MockMessage`, feel free to un-resolve if there's more you'd like me to change.

tests/bot/rules/test_attachments.py
```diff
@@ -1,52 +1,98 @@
...
+        for recent_messages in cases:
+            last_message = recent_messages[0]
+
+            with self.subTest(
+                last_message=last_message,
+                recent_messages=recent_messages,
+                config=self.config
+            ):
+                self.assertIsNone(
+                    await attachments.apply(last_message, recent_messages, self.config)
```
MarkKoz: `self.config` is passed to the `subTest` yet `self.config` is still used here. If this works as-is, then there is no need to pass it to `subTest`.

kwzrd: I'll remove it if you think it's unnecessary.

What I had in mind when adding it was that since the rule's behaviour depends on the config, it should be passed to `subTest` so that it shows up in the error report. Please correct me if I'm misunderstanding something.

kwzrd: > `test_disallows_messages_with_too_many_attachments` should be now acting like a regression test against that duplicate message bug for attachments, right?

Yes, I suppose. The rule now calls the test in the same way as the antispam cog's `on_message` does, which wasn't the case previously.

MarkKoz: It can just stay.

PR #671, opened 2019-11-25
Update antimalware.py to be more consistent with other information messages & improve code a slight bit

Denayder:
Hey! I looked at the `antimalware.py` and felt like I could improve a bit, and I hope I've accomplished that. I look forward to your review of this pull request! :grin:

Here's a list of differences that has been made:
- Messages are now only checked if they contain 1 or more attachments.
- Warning message is now in an embed, but the mention is still text so the user still gets pinged. This is more consistent with other reminders (such as this one: https://prnt.sc/q1t6g7)
- Less if statements and variables
- Removed the try/except block: I don't believe  deleting a message from `on_message` can ever return `NotFound`. If I'm wrong, I'm more than happy to add that back in. 
- If a user sends an illegal file, a list of allowed file types are displayed in the message as well, so they know what they can use.

I think this is an overall improvement to the message and code itself, and I hope you guys agree :grin:

### DIFF
--- a/bot/cogs/antimalware.py
+++ b/bot/cogs/antimalware.py
@@ -1,11 +1,12 @@
 import logging
 
-from discord import Message, NotFound
+from discord import Message, Embed
 from discord.ext.commands import Bot, Cog
 
 from bot.constants import AntiMalware as AntiMalwareConfig, Channels
 
 log = logging.getLogger(__name__)
+PASTE_URL = "https://paste.pythondiscord.com/"
 
 
 class AntiMalware(Cog):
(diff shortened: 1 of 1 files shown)


bot/cogs/antimalware.py
```diff
@@ -1,11 +1,12 @@
 import logging
 
-from discord import Message, NotFound
+from discord import Embed, Message
 from discord.ext.commands import Bot, Cog
 
 from bot.constants import AntiMalware as AntiMalwareConfig, Channels
 
 log = logging.getLogger(__name__)
+PASTE_URL = "https://paste.pythondiscord.com/"
```
Akarys42: https://paste.pydis.com/ is shorter (which is important for a paste service)

bot/cogs/antimalware.py
```diff
@@ -17,37 +18,32 @@ def __init__(self, bot: Bot):
...
-            await message.channel.send(msg)
+        if embed.description:
+            await message.channel.send(f"Hey {message.author.mention}!", embed=embed)
 
             # Delete the offending message:
-            try:
-                await message.delete()
-            except NotFound:
-                log.info(f"Tried to delete message `{message.id}`, but message could not be found.")
+            await message.delete()
```
Akarys42: I notice that you removed the try/except, any special? Logging is useful

Denayder: I removed the try/except block because I don't think it's possible for it to error when the message is from the `on_message` event- Am I wrong in thinking so? I will happily add it back if I am 😁

bot/cogs/antimalware.py
```diff
@@ -17,37 +18,32 @@ def __init__(self, bot: Bot):
...
     async def on_message(self, message: Message) -> None:
         """Identify messages with prohibited attachments."""
-        rejected_attachments = False
-        detected_pyfile = False
+        if len(message.attachments) == 0:
+            return
+
+        embed = Embed()
         for attachment in message.attachments:
             if attachment.filename.lower().endswith('.py'):
```
avayert: since `attachment.filename.lower()` is used twice maybe it can be refactored to the top of the loop here to reduce clutter?
```suggestion
            filename = attachment.filename.lower()
            if filename.endswith('.py'):
```

bot/cogs/antimalware.py
sco1: Yes, it is possible, as this is not the only `on_message` event with the potential to delete the new message.

bot/cogs/antimalware.py
sco1: I'd rather it use the value [defined in the constants](https://github.com/python-discord/bot/blob/master/bot/constants.py#L449) so this doesn't have to be hardcoded.

Denayder:
All requested tweaks have been made :)

Akarys42:
Have you created the default constant in the `config-default.yml` file?

Issue #608, opened 2019-10-25
Write unit tests for `bot/utils/time.py`


SebastiaanZ:
Write unit tests for [`bot/utils/time.py`](../blob/master/bot/utils/time.py). This file already has some unit tests, but they are written for `pytest`. The tests are currently located in [`tests/utils/test_time.py`](../blob/master/tests/utils/test_time.py), but should be moved to the appropriate location in the folder hierarchy, `tests/bot/utils/test_time.py` after they have been migrated to the `unittest` framework.

## Implementation details
Please make sure to read the general information in the [meta issue](553) and the [testing README](../blob/master/tests/README.md). We are aiming for a 100% [branch coverage](https://coverage.readthedocs.io/en/stable/branch.html) for this file, but if you think that is not possible, please discuss that in this issue.

## Additional information
If you want to work on this issue, **please make sure that you get assigned to it** by one of the core devs before starting to work on it. We would like to prevent the situation that multiple people are working on the same issue. To get assigned, leave a comment showing your interesting in tackling this issue.

ikuyarihS:
I would like to work on this please, since I've encountered `bot.utils.time.py` while working on another PR.

PR #672, opened 2019-11-27
Enhance timedelta for infraction expiration

### LINKED ISSUE #668: Display time delta for infraction expiration
# Abstract

In addition to showing the expiration datetime in infraction searches, display the time delta between the expiration and creation of the infraction.

This is optional, but it would also be useful knowing how much time is left until the infraction expires. I may create a separate issue for this, and if so, link it here. EDIT - #669

EDIT2 - I would also like to see this time delta in the bot message confirmation after an infraction has been given out.
Example:
:incoming_envelope: :ok_hand: applied mute to @lemon until 2019-12-01 21:20 (7 days; 42 infractions total).

That last thing though is not mandatory for this issue to be closed. But I think it would be nice, because you could confirm immediately whether your duration was parsed as you expected.

# Specification

Current implementation:
```
Created: 2019-11-23 20:09
Expires: 2019-11-30 20:15
```
Suggested implementation:
```
Created: 2019-11-23 20:09
Expires: 2019-11-30 20:15 (7 days, 6 minutes)
```

The actual granularity of this can vary as durations get longer. For example, if a duration is longer than 30 days, we could put `(1 month)` rather than the exact amount of days. Same goes for years.

If the exact implementation of this doesn't warrant discussion, the one who implements this feature should use their best judgement on when to use the following units of time: (years, months, days, hours, minutes).

For fun, we could put `(∞)` next to permanent infractions.

# Rationale

Confirming the infraction duration should take little effort. If the infraction spills over multiple months, days, or hours for example, it requires more than a second of thinking to determine if the duration is right.
Examples:
```
Created: 2019-11-23 20:58
Expires: 2019-11-23 21:03  <-- 5 minutes, not 1 hour
```
```
Created: 2019-11-23 23:59
Expires: 2019-11-24 00:00 <-- 1 minute, not 1 day or 1 hour
```

# Motivation

I'm lazy.

ikuyarihS:
#### Closes #668

Introducing ~~2~~ 1 helper function that will get the absolute duration between two datetime:
- `bot.utils.time.get_duration_from_expiry()` which accepts an expiry in the form of a string isoformat ( raw data from `infraction['expires_at']` ), which will simply convert the expiry and then call `humanize_delta()`
  - It was created to lessen the burden to convert every time, everywhere.

## Applying to
- When applying `mute`.
- When sending DM after a `mute`.
- When listing infractions.

## Images
image
image
image
image


bot/cogs/moderation/scheduler.py
```diff
@@ -84,12 +84,15 @@ def mod_log(self) -> ModLog:
         icon = utils.INFRACTION_ICONS[infr_type][0]
         reason = infraction["reason"]
         expiry = infraction["expires_at"]
+        expiry_at = expiry
```
Akarys42: I think we can safely override `expiry` and add the duration without altering the rest of the cog.

ikuyarihS: You just read my mind, I'm doing a refactor to get it straight from infraction['expires_at']

ikuyarihS: Aaaaand done, now it's minimum changes to the Cog!

tests/utils/test_time.py
```diff
@@ -60,3 +60,47 @@ def test_wait_until(sleep_patch):
...
+         '2019-11-23 20:09 (7 months, 2 weeks, 1 day, 23 hours, 54 minutes)'),
+        ('2019-11-23T20:58:00Z', datetime(2019, 11, 23, 21, 3), 2, '2019-11-23 20:58 (5 minutes)'),
+        ('2019-11-23T23:59:00Z', datetime(2019, 11, 24, 0, 0), 2, '2019-11-23 23:59 (1 minute)'),
+        ('2019-11-23T23:59:00Z', datetime(2022, 11, 23, 23, 0), 2, '2019-11-23 23:59 (3 years, 3 months)'),
+        ('2019-11-23T23:59:00Z', datetime(2019, 11, 23, 23, 49, 5), 2, '2019-11-23 23:59 (9 minutes, 55 seconds)'),
+        (None, datetime(2019, 11, 23, 23, 49, 5), 2, None),
+    )
+)
+def test_get_duration_from_expiry(expiry: str, date_from: datetime, parts: int, expected: str):
+    assert time.get_duration_from_expiry(expiry, date_from, parts) == expected
```
SebastiaanZ: We don't use `pytest` for our unit tests anymore. This file is an unfortunately leftover, as it was merged during the unittest migration. There's [an issue](https://github.com/python-discord/bot/issues/608) to migrate this file to unittest.

bot/utils/time.py
```diff
@@ -111,3 +120,64 @@ def parse_rfc1123(stamp: str) -> datetime.datetime:
...
+
+    date_from = date_from or datetime.datetime.utcnow()
+    date_to = dateutil.parser.isoparse(expiry).replace(tzinfo=None)
+
+    expiry_formatted = format_infraction(expiry)
+
+    duration = get_duration(date_from, date_to, parts)
+    duration_formatted = f" ({duration})" if duration else ''
+
+    return f"{expiry_formatted}{duration_formatted}"
```
SebastiaanZ: I'm a bit confused by these functions since they seem to share a lot of functionality with the `humanize_delta` utility function that converts a `relativedelta` to a human-readable duration string. Why don't we just parse the datetime string, calculate the `dateutils.relativedelta` and use the existing function to get a human-readable duration?

ikuyarihS: Oh my godness, you are correct, I did not realize we have a `humanize_delta` right there lol, and it works the same, if not more accurate across months and years!

I will probably just write an extra function to handle the expiry but call `humanize_delta` instead!

tests/utils/test_time.py
ikuyarihS: Ah I see, I did not realize that, I would like to work on migrating this to `unitest` as well.

lemonsaurus:
@ikuyarihS taking into account the conversations you had with @SebastiaanZ for his review, what is your plan on this PR to get it finished? Will the test be ported to unittest in this PR or in a different one? If the latter, will the other PR be merged before or after this one? Are you planning to rewrite to make use of `humanize_delta`?

ikuyarihS:
@lemonsaurus Yes, sorry for the delay, I will rewrite to use the `humanize_delta` that we already have for more consistency since we already have it, and it is more than enough for this task.

I've also asked to work on #608 since I will definitely write tests / migrate pytests over to unitests as well.

ikuyarihS:
I've switched to using `humanize_delta` and it's working great!

bot/utils/time.py
```diff
@@ -111,3 +111,33 @@ def parse_rfc1123(stamp: str) -> datetime.datetime:
...
     return dateutil.parser.isoparse(timestamp).strftime(INFRACTION_FORMAT)
+
+
+def get_duration_from_expiry(
+    expiry: str = None,
+    date_from: datetime.datetime = None,
+    max_units: int = 2
+) -> Optional[str]:
+    """
+    Returns a human-readable version of the  the duration between datetime.utcnow() and an expiry.
```
MarkKoz:
```suggestion
    Returns a human-readable version of the duration between datetime.utcnow() and an expiry.
```

bot/utils/time.py
```diff
@@ -111,3 +111,33 @@ def parse_rfc1123(stamp: str) -> datetime.datetime:
...
+
+    Unlike the original function, this function will force the precision to be 'seconds' by not passing it.
+    max_units specifies the maximum number of units of time to include (e.g. 1 may include days but not hours).
+    By default, max_units is 2
+
+    :param expiry: A string. If not passed in, will early return a None ( Permanent infraction ).
+    :param date_from: A datetime.datetime object. If not passed in, will use datetime.utcnow().
+    :param parts: An int, to show how many parts will be returned ( year - month or year - month - week - day ...).
```
MarkKoz: This style for documenting parameters is not used in our code base. Instead, they are either part of the paragraphs or, in cases with many arguments, as a bullet point list. I don't remember exactly but surely you'll find some examples if you look around elsewhere in the code base.

bot/utils/time.py
```diff
@@ -111,3 +111,33 @@ def parse_rfc1123(stamp: str) -> datetime.datetime:
...
+
+    date_from = date_from or datetime.datetime.utcnow()
+    date_to = dateutil.parser.isoparse(expiry).replace(tzinfo=None, microsecond=0)
+
+    expiry_formatted = format_infraction(expiry)
+
+    duration = humanize_delta(relativedelta(date_to, date_from), max_units=max_units)
+    duration_formatted = f" ({duration})" if duration else ''
+
+    return f"{expiry_formatted}{duration_formatted}"
```
MarkKoz: It's neither clear that this function returns a timestamp nor that it uses `format_infraction` specifically to format it. This functions seems to have been designed with only returning the duration in mind and then the timestamp was tacked on towards the end. It could only return the duration, *but* that would mean two function calls in the mod cogs. It may be fine to just rename this function to `format_infraction_with_duration` (preferably something shorter if you can think of it) and, of course, document that it returns a timestamp.

bot/utils/time.py
ikuyarihS: Thank you, I will include this in new commits to avoid having too many small commits.

bot/utils/time.py
ikuyarihS: Hmm, I was actually looking at the function right above, `wait_until` and I saw this style and thought that it will automatically build the doc from this. I will simply remove this part, it should be clear enough from the type hint and the docstring.

bot/utils/time.py
ikuyarihS: You are right, I was just replacing `format_infraction` and avoid calling another command on top of it.
I've renamed the function to `format_infraction_with_duration` ( I can't think of any shorter name, `format_infraction_readable` kind of saying `format_infraction` is not readable, so I'd rather go with lengthier name ).

PR #674, opened 2019-11-30
Refactoring out common pattern for sending errors

avayert:
This is a draft/suggestion for a refactoring of this fairly common pattern in the code. I have not gone through thoroughly so I will have missed some spots. It will have strange formatting changes that I didn't notice because my PyCharm settings are a bit funky ATM. I'm mainly asking for opinions.

### DIFF
--- a/bot/cogs/doc.py
+++ b/bot/cogs/doc.py
@@ -21,7 +21,7 @@ from bot.constants import MODERATION_ROLES, RedirectOutput
 from bot.converters import ValidPythonIdentifier, ValidURL
 from bot.decorators import with_role
 from bot.pagination import LinePaginator
-
+from bot.utils.context import Context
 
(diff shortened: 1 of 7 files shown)


lemonsaurus:
A few observations here:
- I don't think this is necessary today, because we've implemented a design where these features just raise exceptions and then we handle the error embed sending inside the bots error handler.
- There are many unrelated style changes in this PR, which makes the PR a bit hard to read. Did you run it through `black` or something?
- It fails lint.
- It's been stale for like 4 months.
- There's a lot of changes and only 1 commit.

All of this taken into account, I think I'm gonna close this PR. I do think the core idea is interesting though, and subclassing Context to add features to it may be a pattern we should make use of in the future, so thanks for writing it!

PR #678, opened 2019-12-04
Antimalware: fix paste service URL showing replacement field

MarkKoz:
`{key}` was being shown at the end of the URL because the constants are expected to be used with `.format()` to make HTTP requests.


sco1:
Can't we just use `site_paste`?

MarkKoz:
Oh yeah didn't notice.

PR #679, opened 2019-12-04
Display time left until expiration of infraction

### LINKED ISSUE #669: Display time left until expiration of infraction
# Abstract

In addition to the creation and expiration datetimes displayed in an infraction search, display how much time is currently left until the infraction expires at the time of invocation.

# Specification

Proposal:
```
Created: 2019-11-23 20:09
Expires: 2019-11-30 20:15
Remaining: 5 days, 1 hour
```

In regards to granularity, my proposal is to only show the 2 largest units. So if it expires in more than a year, show only years and months. If over a month, show only months and days. If over a day, show days and hours. If over an hour, show hours and minutes. And finally, if over a minute, show only minutes, don't show seconds.

If it expires in less than a minute, either display "less than a minute" or just "1 minute" if that's too verbose.

If the infraction has already expired, show "Expired".
If the infraction never expires, perhaps this field should be omitted entirely.

# Rationale

Similar to #668, confirming how long until an infraction expires should require as little effort as possible.

# Motivation

See my motivation in #668.

ikuyarihS:
#### Closes #669

Will show the remaining time until an infraction is expired.

Introduced 1 helper function in `time` module - `until_expiration` that makes use of the nicely, already made `humanize_delta` in the same module.

### Images:
An active infraction:
image

An infraction that has yet to expire, but is inactive:
image

### DIFF
--- a/bot/cogs/moderation/management.py
+++ b/bot/cogs/moderation/management.py
@@ -232,6 +232,12 @@ class ModManagement(commands.Cog):
         user_id = infraction["user"]
         hidden = infraction["hidden"]
         created = time.format_infraction(infraction["inserted_at"])
+
+        if active:
+            remaining = time.until_expiration(infraction["expires_at"]) or 'Expired.'
+        else:
+            remaining = 'Inactive.'
+
         if infraction["expires_at"] is None:
             expires = "*Permanent*"
         else:
@@ -247,6 +253,7 @@ class ModManagement(commands.Cog):
             Reason: {infraction["reason"] or "*None*"}
             Created: {created}
             Expires: {expires}
+            Remaining: {remaining}
             Actor: {actor.mention if actor else actor_id}
             ID: `{infraction["id"]}`
             {"**===============**" if active else "==============="}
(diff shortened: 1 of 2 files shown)


bot/cogs/moderation/management.py
```diff
@@ -232,6 +232,12 @@ def infraction_to_string(self, infraction: utils.Infraction) -> str:
         user_id = infraction["user"]
         hidden = infraction["hidden"]
         created = time.format_infraction(infraction["inserted_at"])
+
+        if active:
+            remaining = time.until_expiration(infraction["expires_at"]) or 'Expired.'
```
MarkKoz:
```suggestion
            remaining = time.until_expiration(infraction["expires_at"]) or "Expired"
```

bot/cogs/moderation/management.py
```diff
@@ -232,6 +232,12 @@ def infraction_to_string(self, infraction: utils.Infraction) -> str:
...
+        else:
+            remaining = 'Inactive.'
```
MarkKoz:
```suggestion
            remaining = "Inactive"
```

PR #681, opened 2019-12-08
Subclass Bot

### LINKED ISSUE #468: Subclass Bot
This allows us to do some backend QoL stuff:

* Override `add_cog()` so that it logs the loading of cogs
* IDEs will recognise additional attributes, such as the API client
* Override `load_extensions()` so that it automatically prepends `bot.cogs` to the name given

And there's maybe more stuff I'm forgetting.

MarkKoz:
Resolves #468 

- [x] Determine if the custom connector for the session is needed, and if it should be used for the API client and/or discord.py's sessions too.
- [ ] Write tests? Testing the closing/re-opening may prove to be complicated, but not sure.

Probably don't want to prepends `bot.cogs` when adding extensions since that would mess up the `Extensions` extension and require the unload and reload methods to also exhibit this behaviour for consistency while making it impossible to load extensions from anywhere else should we want to do that in the future. Not worth it in retrospect.

### DIFF
--- a/bot/api.py
+++ b/bot/api.py
@@ -32,7 +32,7 @@ class ResponseCodeError(ValueError):
 class APIClient:
     """Django Site API wrapper."""
 
-    def __init__(self, **kwargs):
+    def __init__(self, loop: asyncio.AbstractEventLoop, **kwargs):
         auth_headers = {
             'Authorization': f"Token {Keys.site_api}"
         }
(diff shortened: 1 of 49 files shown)


bot/api.py
```diff
@@ -42,12 +42,39 @@ def __init__(self, **kwargs):
...
+            return
+
+        await self.session.close()
+        self._ready.clear()
+
+    def recreate(self) -> None:
+        """Schedule the aiohttp session to be created if it's been closed."""
+        if self.session is None or self.session.closed:
+            # Don't schedule a task if one is already in progress.
+            if self._creation_task is None or self._creation_task.done():
```
MarkKoz: These checks may be a bit paranoid - `if not self._ready.is_set():` may suffice here. I don't think it's too bad though and doesn't hurt to be a bit safer. 🤷‍♂

scragly: Yeah, this is fine.

PR #684, opened 2019-12-10
Kick users from voice when they get muted

### LINKED ISSUE #644: Have !mute command kick from voice channel
# Abstract

The `!mute` command currently removes voice channel permissions, but does not kick the user from a voice channel they may already be in. Have the `!mute` command check if the user is currently in a voice channel, and if so, kick them from that channel.

# Specification

* The API call for checking whether a user is currently in a voice channel can be found [here](https://discordpy.readthedocs.io/en/latest/api.html#discord.Member.voice). The attribute from the `discord.VoiceState` object that is returned that provides this information is [`channel`](https://discordpy.readthedocs.io/en/latest/api.html#discord.VoiceState.channel)
* The API call for kicking a user can be found [here](https://discordpy.readthedocs.io/en/latest/api.html#discord.Member.move_to). As of d.py version 1.1.0, you can now pass `None` in order to kick them rather than have to move them to a different channel.

# Rationale

The `!mute` command should be expected to remove a users ability to participate in the server in all shapes and forms.

# Motivation

A user can currently misbehave from either being explicitly disruptive in a voice channel, or by repeatedly joining and leaving a voice channel. The only course of action via the bot to stop this behavior is by banning the user.

MarkKoz:
Closes #644

If the kick fails, the bot will send a message saying that applying the infraction failed. I felt that's adequate compared to a specific message for a voice channel kick failure. A specific message can be done easily if it's sent separately from the usual "confirmation message". However, integrating it into the confirmation message would require more thought and effort as it wasn't designed to be extensible to such degree.

### DIFF
--- a/bot/cogs/moderation/scheduler.py
+++ b/bot/cogs/moderation/scheduler.py
@@ -329,7 +329,7 @@ class InfractionScheduler(Scheduler):
             log_content = mod_role.mention
         except discord.HTTPException as e:
             log.exception(f"Failed to deactivate infraction #{id_} ({type_})")
-            log_text["Failure"] = f"HTTPException with code {e.code}."
+            log_text["Failure"] = f"HTTPException with status {e.status} and code {e.code}."
             log_content = mod_role.mention
 
         # Check if the user is currently being watched by Big Brother.
--- a/bot/cogs/moderation/scheduler.py
+++ b/bot/cogs/moderation/scheduler.py
@@ -146,14 +146,18 @@ class InfractionScheduler(Scheduler):
                 if expiry:
                     # Schedule the expiration of the infraction.
                     self.schedule_task(ctx.bot.loop, infraction["id"], infraction)
-            except discord.Forbidden:
+            except discord.HTTPException as e:
                 # Accordingly display that applying the infraction failed.
                 confirm_msg = f":x: failed to apply"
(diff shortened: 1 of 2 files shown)


bot/cogs/moderation/scheduler.py
```diff
@@ -146,14 +146,18 @@ def mod_log(self) -> ModLog:
...
                 expiry_msg = ""
                 log_content = ctx.author.mention
                 log_title = "failed to apply"
 
-                log.warning(f"Failed to apply {infr_type} infraction #{id_} to {user}.")
+                log_msg = f"Failed to apply {infr_type} infraction #{id_} to {user}"
+                if isinstance(e, discord.Forbidden):
+                    log.warning(f"{log_msg}: bot lacks permissions.")
+                else:
+                    log.exception(log_msg)
```
scragly: If it's not a `discord.Forbidden`, there's definitely something unusual going on so it's best to probably log the full traceback also for when we need to investigate.

```suggestion
                    log.exception(log_msg, exc_info=True)
```

MarkKoz: `.exception()` already sets it to `True` by default.

scragly: So it does, my mistake.

PR #691, opened 2019-12-12
Clean command improvements

### LINKED ISSUE #690: Minor improvements on the !clean command
There are a couple of bugs that need to fixed with the !clean command, and a few new features I'd like to see.

## Optional channel parameter
I'd like to be able to clean a channel from the comfort of another channel.

For example:
`!clean all 200 #python-discussion` should run this command inside the #python-discussion channel instead of in the channel it was invoked from.

When this happens, the bot should output its feedback into the channel you invoked the command in. So, if I was in #admins and I used this command, the bot should post its feedback in #admins, not in #python-discussion.

## Don't hide it from showing up in !help
Currently, this command doesn't show up when you do a `!help`, even if you have access to it. There's no reason why this should be true, it should be shown in `!help` output for anyone who has access to the command.

## New alias
Please add `purge` as an alias for `!clean` so that @scragly will be able to remember what this command is called. :grin:

MarkKoz:
Resolves #690 

The channel is always optional and is the last parameter for all the clean commands. This means the amount must be specified if one wishes to specify the channel.

### DIFF
--- a/bot/cogs/clean.py
+++ b/bot/cogs/clean.py
@@ -37,10 +37,13 @@ class Clean(Cog):
         return self.bot.get_cog("ModLog")
 
     async def _clean_messages(
-            self, amount: int, ctx: Context,
-            bots_only: bool = False, user: User = None,
-            regex: Optional[str] = None,
-            channel: Optional[TextChannel] = None
+        self,
+        amount: int,
+        ctx: Context,
+        bots_only: bool = False,
+        user: User = None,
+        regex: Optional[str] = None,
+        channel: Optional[TextChannel] = None
     ) -> None:
         """A helper function that does the actual message cleaning."""
         def predicate_bots_only(message: Message) -> bool:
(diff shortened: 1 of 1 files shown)


bot/cogs/clean.py
```diff
@@ -39,7 +39,8 @@ def mod_log(self) -> ModLog:
     async def _clean_messages(
             self, amount: int, ctx: Context,
             bots_only: bool = False, user: User = None,
-            regex: Optional[str] = None
+            regex: Optional[str] = None,
+            channel: Optional[TextChannel] = None
```
scragly: This should be dedented one level. I understand though this wasn't something you defined in this PR.

bot/cogs/clean.py
```diff
@@ -167,35 +172,35 @@ def predicate_regex(message: Message) -> bool:
...
+    @group(invoke_without_command=True, name="clean", aliases=["purge"])
     @with_role(*MODERATION_ROLES)
     async def clean_group(self, ctx: Context) -> None:
         """Commands for cleaning messages in channels."""
         await ctx.invoke(self.bot.get_command("help"), "clean")
 
     @clean_group.command(name="user", aliases=["users"])
     @with_role(*MODERATION_ROLES)
-    async def clean_user(self, ctx: Context, user: User, amount: int = 10) -> None:
+    async def clean_user(self, ctx: Context, user: User, amount: int = 10, channel: TextChannel = None) -> None:
```
scragly:
```suggestion
    async def clean_user(self, ctx: Context, user: User, amount: Optional[int] = 10, channel: TextChannel = None) -> None:
```
This allows the amount to be skipped when specifying a channel.

bot/cogs/clean.py
```diff
@@ -167,35 +172,35 @@ def predicate_regex(message: Message) -> bool:
...
         """Delete messages posted by the provided user, stop cleaning after traversing `amount` messages."""
-        await self._clean_messages(amount, ctx, user=user)
+        await self._clean_messages(amount, ctx, user=user, channel=channel)
 
     @clean_group.command(name="all", aliases=["everything"])
     @with_role(*MODERATION_ROLES)
-    async def clean_all(self, ctx: Context, amount: int = 10) -> None:
+    async def clean_all(self, ctx: Context, amount: int = 10, channel: TextChannel = None) -> None:
```
scragly:
```suggestion
    async def clean_all(self, ctx: Context, amount: Optional[int] = 10, channel: TextChannel = None) -> None:
```
This allows the amount to be skipped when specifying a channel.

bot/cogs/clean.py
```diff
@@ -167,35 +172,35 @@ def predicate_regex(message: Message) -> bool:
...
         """Delete all messages, regardless of poster, stop cleaning after traversing `amount` messages."""
-        await self._clean_messages(amount, ctx)
+        await self._clean_messages(amount, ctx, channel=channel)
 
     @clean_group.command(name="bots", aliases=["bot"])
     @with_role(*MODERATION_ROLES)
-    async def clean_bots(self, ctx: Context, amount: int = 10) -> None:
+    async def clean_bots(self, ctx: Context, amount: int = 10, channel: TextChannel = None) -> None:
```
scragly:
```suggestion
    async def clean_bots(self, ctx: Context, amount: Optional[int] = 10, channel: TextChannel = None) -> None:
```
This allows the amount to be skipped when specifying a channel.

bot/cogs/clean.py
```diff
@@ -167,35 +172,35 @@ def predicate_regex(message: Message) -> bool:
...
         """Delete all messages posted by a bot, stop cleaning after traversing `amount` messages."""
-        await self._clean_messages(amount, ctx, bots_only=True)
+        await self._clean_messages(amount, ctx, bots_only=True, channel=channel)
 
     @clean_group.command(name="regex", aliases=["word", "expression"])
     @with_role(*MODERATION_ROLES)
-    async def clean_regex(self, ctx: Context, regex: str, amount: int = 10) -> None:
+    async def clean_regex(self, ctx: Context, regex: str, amount: int = 10, channel: TextChannel = None) -> None:
```
scragly:
```suggestion
    async def clean_regex(self, ctx: Context, regex: str, amount: Optional[int] = 10, channel: TextChannel = None) -> None:
```
This allows the amount to be skipped when specifying a channel.

scragly:

bot/cogs/clean.py
```diff
@@ -168,35 +176,57 @@ def predicate_regex(message: Message) -> bool:
             channel_id=Channels.modlog,
         )
 
-    @group(invoke_without_command=True, name="clean", hidden=True)
+    @group(invoke_without_command=True, name="clean", aliases=["purge"])
```
mathsman5133: I know you didn't touch this, but while we're here, is there any particular reason why `invoke_without_command=True` is set? If it is left to the default `False`, the `@with_role` check can be on the command group, but not any of the subcommands - it will invoke the group command (and all checks) when any subcommand is called. The only thing that would need to change is to only send the help if there's no subcommand (and do a manual check) - 

```py
if not ctx.invoked_subcommand:
   await ctx.invoke(self.bot.get_command("help"), "clean")
```

Just a thought...

PR #692, opened 2019-12-12
Allow mods+ to use commands and delete bots' messages in checkpoint channel

### LINKED ISSUE #688: Mods+ should be permitted to use any command in #checkpoint
Currently the bot will ignore all commands inside #checkpoint. However, there are cases where this is not ideal, for example if we want to clean the channel up with the !clean command.

Write a fix that allows moderators, admins and owners to use any command in this channel, but still ignores everything but `!accept` from anyone else.

### LINKED ISSUE #689: Make the bot delete any message it posts in #checkpoint after a while
Currently, messages posted in #checkpoint by the bot will usually be posted with a `delete_after` of 10 seconds or so, so they will automatically be removed. This helps the channel stay clean so that the important messages at the top of it will stay there. The bot also automatically deletes any messages posted in this channel by other users.

However, when the bot posts messages in this channel that are not related to the Verification cog, they will never get deleted. For example, if a user in this channel triggers the spam filter, the bot will post a message about muting the user which will never be deleted.

We want this channel to stay clean, so let's implement some logics that ensure that **any message posted by the bot** in this channel will be deleted after a short time, like 10 seconds.

MarkKoz:
Resolves #688 and resolves #689.

### DIFF
--- a/bot/cogs/verification.py
+++ b/bot/cogs/verification.py
@@ -9,9 +9,10 @@ from bot.cogs.moderation import ModLog
 from bot.constants import (
     Bot as BotConfig,
     Channels, Colours, Event,
-    Filter, Icons, Roles
+    Filter, Icons, MODERATION_ROLES, Roles
 )
 from bot.decorators import InChannelCheckFailure, in_channel, without_role
+from bot.utils.checks import without_role_check
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


bot/cogs/verification.py
```diff
@@ -55,7 +57,11 @@ def mod_log(self) -> ModLog:
     async def on_message(self, message: Message) -> None:
         """Check new message event for messages to the checkpoint channel & process."""
         if message.author.bot:
-            return  # They're a bot, ignore
+            # They're a bot, delete their message after the delay.
+            # But not the periodic ping; we like that one.
+            if message.content != PERIODIC_PING:
```
mathsman5133: Since this is an `on_message` listener, won't it trigger for all messages - and since this is before the check that ensures it's in the correct channel, delete any messages sent by the bot after 10 sec?

MarkKoz: Yes you're right. The channel check on l66 needs to come first.

PR #693, opened 2019-12-12
Allow "recent" as infraction ID for infraction edit command

### LINKED ISSUE #624: Helper for editing most recent infraction.
For the command `!infraction edit`, the argument to specify the infraction requires an ID. It would be useful to have a special argument you could provide instead of the ID that would indicate the infraction you most recently gave out.

MarkKoz:
Resolves #624 

Could support multiple phrases but I left it at "recent" to start with. The confirmation message now displays the infraction ID to reduce uncertainty as to which infraction was edited when "recent" is used.

### DIFF
--- a/bot/cogs/moderation/management.py
+++ b/bot/cogs/moderation/management.py
@@ -9,7 +9,7 @@ from discord.ext import commands
 from discord.ext.commands import Context
 
 from bot import constants
-from bot.converters import InfractionSearchQuery
+from bot.converters import InfractionSearchQuery, string
 from bot.pagination import LinePaginator
 from bot.utils import time
 from bot.utils.checks import in_channel_check, with_role_check
@@ -20,15 +20,6 @@ from .modlog import ModLog
 log = logging.getLogger(__name__)
 
 UserConverter = t.Union[discord.User, utils.proxy_user]
-
-
-def permanent_duration(expires_at: str) -> str:
-    """Only allow an expiration to be 'permanent' if it is a string."""
-    expires_at = expires_at.lower()
-    if expires_at != "permanent":
-        raise commands.BadArgument
-    else:
-        return expires_at
 
 
 class ModManagement(commands.Cog):
@@ -61,7 +52,7 @@ class ModManagement(commands.Cog):
         self,
         ctx: Context,
         infraction_id: int,
-        duration: t.Union[utils.Expiry, permanent_duration, None],
+        duration: t.Union[utils.Expiry, string("permanent"), None],
         *,
         reason: str = None
     ) -> None:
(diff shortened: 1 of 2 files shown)


bot/converters.py
```diff
@@ -15,6 +15,25 @@
 log = logging.getLogger(__name__)
 
 
+def string(*values, preserve_case: bool = False) -> t.Callable[[str], str]:
```
scragly: Not sure on this being lowercase and not very descriptive of it's purpose.

Maybe something like `AllowedStrings` or `AcceptedStrings` instead?

bot/cogs/moderation/management.py
```diff
@@ -60,8 +51,8 @@ def infractions_cog(self) -> Infractions:
     async def infraction_edit(
         self,
         ctx: Context,
-        infraction_id: int,
-        duration: t.Union[utils.Expiry, permanent_duration, None],
+        infraction_id: t.Union[int, string("recent")],
```
scragly: I'd also suggest accepting `last` and `L`

bot/cogs/moderation/management.py
```diff
@@ -60,8 +51,8 @@ def infractions_cog(self) -> Infractions:
...
+        duration: t.Union[utils.Expiry, string("permanent"), None],
```
scragly: This could be a chance to give a short alias of `P` for permanent.

bot/converters.py
MarkKoz: I felt it may be more confusing to name it as a class since passing parameters to converters is uncharacteristic.

Yes, I can make the name more descriptive.

scragly:

bot/cogs/moderation/management.py
```diff
@@ -79,21 +70,40 @@ def infractions_cog(self) -> Infractions:
...
         if duration is None and reason is None:
             # Unlike UserInputError, the error handler will show a specified message for BadArgument
             raise commands.BadArgument("Neither a new expiry nor a new reason was specified.")
 
         # Retrieve the previous infraction for its information.
-        old_infraction = await self.bot.api_client.get(f'bot/infractions/{infraction_id}')
+        if isinstance(infraction_id, str):
+            params = {
+                "actor__id": ctx.author.id,
+                "ordering": "-inserted_at"
```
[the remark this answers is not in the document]
MarkKoz: It isn't actually. Regardless, I would have ordered it explicitly.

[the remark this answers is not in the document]
SebastiaanZ: It isn't? It should be.

```py
    class Meta:
        """Defines the meta options for the infraction model."""

        ordering = ['-inserted_at']
```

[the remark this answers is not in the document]
MarkKoz: Something wrong on my end. I was wrong, it is the default.

PR #696, opened 2019-12-13
Use OAuth to be Reddit API compliant

### LINKED ISSUE #354: Reddit JSON API OAuth requirement not met.

jos-b:
Re-implements #510 which had to be reverted due to a crash on the bot.

Closes #354.
- Requests API tokens on extension init
- Client ID and Secrets for Reddit API set up in bot constants and config
- HTTP Requests use the API to receive responses
- Change User-Agent to be complaint with Reddits API rules (https://github.com/reddit-archive/reddit/wiki/API)

Client_ID and Secret have to be created by an admin (https://www.reddit.com/prefs/apps).
The User-Agent string has to be modified with proper values.
Further commits are going to be needed as this PR will conflict with the changes in #501.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,7 +30,7 @@ jobs:
       - script: python -m flake8
         displayName: 'Run linter'
 
-      - script: BOT_API_KEY=foo BOT_TOKEN=bar WOLFRAM_API_KEY=baz coverage run -m xmlrunner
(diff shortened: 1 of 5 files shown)


jos-b:
This PR must be updated to fix whatever the crash was on Friday 13th December from 12:06AM to 12:46AM UTC. Logs may be able to be gathered in the morning.

scragly:
Was a case of accidentally missing credentials. Retested with the newly generated app creds to make sure things work ok and it seems fine, so will remerge.

PR #651, opened 2019-11-04
[WIP] Show plain nick, or nothing for proxy objects instead of ID in infraction mod log

Numerlor:
closes: #533

PR #701, opened 2019-12-16
Support applying infractions to users not in the DB via Converter `FetchedUser`

### LINKED ISSUE #627: Support applying infractions to users who are not in the database
For example, trying to ban a user (via ID) who has never joined the server before will currently error out because the syncer never got a chance to create the user in the database. Therefore, the user will have to be manually created with `in_guild=False` after their details are fetched with `fetch_user`.

I _think_ this can all be done inside `proxy_user` (if it's turned into a `Converter` subclass so it can be async), but not sure.

manusaurio:
Closes #627 

Trying to ban a user uses 3 sources as converter: `discord.Member`, `discord.User` and the `proxy_user` function, in that order. If a user never joined the server, the first two fails, and `proxy_user` doesn't, but the latter is essentially useless anyway in this case, since the user _needs_ to have its own row in the database for the infraction to take place.

## How this change works
Ban-functions now try to convert the `user` argument with sources in the following order:
1. `discord.Member`
2. `discord.User`
3. `bot.converters.FetchedUser` (returns `discord.User` or `discord.Object`)

**There are several rules for this in the previous flow:**
If the `FetchedUser` converter is executed, it will fetch from the Discord API. If there's no error during the fetch, the `discord.User` object will be used to try to post the infraction. If there's an error because the user doesn't exist, the `discord.User` object attributes will be used to add a new user to the database, then infraction application will be posted again.

If there's any error during the fetch, and said error _isn't_ explicitly that the user doesn't exist, a made-up `discord.Object` containing the user ID will be used, and if the user needs to be added to the DB, default values will be used to fill the remaining columns:

- `avatar_hash`: `0`
- `discriminator`: `0`
- `in_guild`: `False`
- `name`: `Name unknown`
- `roles`: `[]`

## Main changes and additions
1. Added `post_user(...)`: adds a new user to the database, uses default values if the passed object doesn't have all the required attributes.
2. Changed `post_infraction`: tries to `post_user(...)` if the first try posting the infraction fails because the user doesn't exist.
3. Moved `proxy_user` to `bot.converters`
4. Added `FetchedUser(Converter)` class, which fetches from the Discord API and returns a `discord.User` or `discord.Object` if the fetching fails (as long as the error doesn't indicate the user doesn't exist.)

## Observations
The flow after `discord.Member`/`discord.User` could be different (since we seem to aim to minimize API calls), but longer, and (presumably) more convoluted:
1. Try to use a proxy `discord.Object` first.
2. Try to apply the infraction with that.
3. If it doesn't exist, then try fetching the user. If it success, try adding it to the database.
4. If the fetch from before fails, then use the original `discord.Object` anyway and use default values for the rows to add the user to the database.

## Examples
image
image


bot/cogs/moderation/utils.py
```diff
@@ -74,22 +107,21 @@ def proxy_user(user_id: str) -> discord.Object:
...
+                    return
+            else:
+                log.exception("An unexpected ResponseCodeError occurred while adding an infraction:")
+                await ctx.send(":x: There was an error adding the infraction.")
+                return
         else:
-            log.exception("An unexpected ResponseCodeError occurred while adding an infraction:")
-            await ctx.send(":x: There was an error adding the infraction.")
-            return
+            break
```
manusaurio: The loop here makes it quite confusing to read. It's somehow equivalent to
```
  try:
      response = ...post infraction
  except ...:
      if exception == user doesn't exist:
          post_user()
          try:
              response = ...post infraction
          except ...:
              log.trace(...)
          return
      else:
          log.trace(...)
          return
```

bot/cogs/moderation/utils.py
```diff
@@ -74,22 +107,21 @@ def proxy_user(user_id: str) -> discord.Object:
...
+            response = await ctx.bot.api_client.post('bot/infractions', json=payload)
+        except ResponseCodeError as exp:
+            if exp.status == 400 and 'user'in exp.response_json:
+                # Only once attempt to try to add the user to the database, not two:
+                if attempt > 0 or await post_user(ctx, user) is None:
+                    return
+            else:
+                log.exception("An unexpected ResponseCodeError occurred while adding an infraction:")
+                await ctx.send(":x: There was an error adding the infraction.")
+                return
```
manusaurio: This loop makes it a bit difficult to understand what's going on. It should be equivalent to something like this:
```
  try:
      response = ...post infraction
  except ...:
      if exception == user doesn't exist:
          post_user()
          try:
              response = ...post infraction
          except ...:
              log.trace(...)
          return
      else:
          log.trace(...)
          return
```

bot/cogs/moderation/utils.py
```diff
@@ -74,22 +107,21 @@ def proxy_user(user_id: str) -> discord.Object:
...
+    # Try to apply the infraction. If it fails because the user doesn't exist, try to add it.
+    for attempt in range(2):
+        try:
+            response = await ctx.bot.api_client.post('bot/infractions', json=payload)
+        except ResponseCodeError as exp:
+            if exp.status == 400 and 'user'in exp.response_json:
+                # Only once attempt to try to add the user to the database, not two:
+                if attempt > 0 or await post_user(ctx, user) is None:
+                    return
+            else:
```
manusaurio: This loop makes it a bit difficult to understand what's going on. It should be equivalent to something like this:

  try:
      response = ...post infraction
  except ...:
      if exception == user doesn't exist:
          post_user()
          try:
              response = ...post infraction
          except ...:
              log.trace(...)
          return
      else:
          log.trace(...)
          return

manusaurio:
I know the fetching itself could have been done in the converter itself, but I felt it was a bit odd to make a converter have side effects like that. I think it should limit its job to just... convert. Not sure if I'm wrong on this.

bot/cogs/moderation/utils.py
MarkKoz: Do you think readability would be better if you did it like:

```py
for should_post_user in (True, False):
    ...
```
(probably could come up with a better name)

MarkKoz:
Also, there is a `fetch_user` somewhere in the infraction scheduler IIRC. Not sure if it should be removed if `proxy_user` is going to stay. After all, if `fetch_user` failed once (and thus we fell back to `proxy_user`), it's likely to fail again. On the other hand, it failing would be some weird, likely intermittent, API issue because we'd not fall back if the user doesn't exist.

manusaurio:
I'll pick this up in a few hours, to refactor it and finish it.

Question: how should `proxy_user` work? I find strange the way it's done. It says `Used when a Member or User object cannot be resolved.`, but it's still required to have a row for that user in the local DB. So, for this to work, it would be necessary to `POST` that user, basically with a made up ID. Isn't that a bit dangerous?

On the other hand, there's a PR by @Numerlor that removes both `proxy_user`s and makes a class of it, so we should keep that in mind:
https://github.com/python-discord/bot/pull/651/files#diff-9d66c9dd2c18a8ecd9bc3be1100ae94eR73-R98

MarkKoz:
The fetch may fail even if the user is already in the database since our
API is, of course, separate from Discord's. It this situation the
interactions will behave as they currently do - they work as long as the
user is in the database. Better than not having anything at all to fall
back on.

If we want it to really be robust, then fake information could be posted to
the API if it's fallen back to a proxy user. It's not a big deal since the
ID is the only important field and that is being supplied by the person
invoking the command (if it's a username instead then just give up).

Den ons 18 dec. 2019 13:22manusaurio <notifications@github.com> skrev:

manusaurio:
A couple of issues came up, so now it looks _worse_.

As you said, @MarkKoz, it would be a good idea to have that `discord.Object` from `proxy_user(id)` still working, in case fetching fails, but the error doesn't imply the user doesn't exist. The problem is that I can't (I think) import the `proxy_users` from the moderation utils to `converters` (I think there's a circular import issue?) so these commits add a new `proxy_user` as a helper in the `FetchedUser` class itself. That's an abomination that should be addressed.

bot/converters.py
```diff
@@ -281,22 +281,50 @@ class ISODateTime(Converter):
...
+        try:
+            user_id = int(user_id)
+        except ValueError:
+            raise BadArgument
+
+        user = discord.Object(user_id)
+        user.mention = user.id
+        user.avatar_url_as = lambda static_format: None
+
+        return user
```
manusaurio: No much effort here because this won't stay.

In my opinion, maybe we should wait for #651's `ProxyUser` class in `utils` (but I can't tell if there would be another issue importing from there, correct me if I'm wrong - I should have tested it.)

bot/cogs/moderation/utils.py
```diff
@@ -31,38 +31,45 @@
...
+        log.trace("Couldn't take all the attributes for the user payload, taking just its ID.")
+        # XXX: Not sure if these default values are ideal.
+        payload = {
+            'avatar_hash': 0,
+            'discriminator': 0,
+            'id': user.id,
+            'in_guild': False,
+            'name': 'Some name',
+            'roles': []
+        }
```
manusaurio: My lack of experience doesn't tell me if doing `t.Union[discord.User, discord.Object]` and an indistinctive try-except is good practice or not. Need feeback on this.

manusaurio: I just remembered `getattr` exists.

MarkKoz:
> in case fetching fails, but the error doesn't imply the user doesn't exist. 

Why not? If it's a `NotFound` AKA 404 then that's exactly what it implies.

> The problem is that I can't (I think) import the `proxy_users`...

There isn't a need to import it. Make the object inside the `FetchedUser` converter. Here's a rough example:

```py
class FetchedUser:
    async def convert():
        ...

        try:
            # If the fetch succeeds just return the user
            return await fetch_user(user_id)
        except discord.HTTPException as e:
            if e.status != 404:
                # There was some weird API issue.
                # Create a mock User object
                user = discord.Object(user_id)
                user.mention = user.id

                ...

                return user
            else:
                # It's a 404 so the user simply doesn't exist.
                raise BadArgument("User does not exist.")
```

The original `proxy_user`s can likely just be completely removed and replaced by `FetchedUser`, unless you think there is a case for some other commands (now or in the future) for which it'd be not worth the API call of `fetch_user`.

manusaurio:
> Why not? If it's a NotFound AKA 404 then that's exactly what it implies.
My sentence should be read as "if there is an error and it doesn't imply the user doesn't exist, then return a proxy," sorry if I worded it poorly, the comma shouldn't be there (`in case fetching fails but the error doesn't imply the user doesn't exist`)

> The original `proxy_users` can likely just be completely removed and replaced by `FetchedUser`
That makes it much easier then. But `management.py` does use `proxy_use` (to search for infractions.) Should it be removed anyway, and should I make it use `FetchedUser` instead too? I assume there shouldn't be a problem about the API limits unless the moderators really want to test it.

MarkKoz:
> But management.py does use proxy_use (to search for infractions.) 

Oh that is a good case for it. I'd say leave it. You should just move `proxy_user` into the converters module to avoid any imports and allow the `FetchedUser` to return the proxy e.g. `return await ProxyUser.convert(user_id)`.

Regarding #651, it's stale so I wouldn't wait for it. I don't think the resulting conflicts will be too bad. If you really want it merged first, you could try poking the author or even ask to take it over yourself if you fancy that.

bot/cogs/moderation/utils.py
```diff
@@ -25,30 +24,42 @@
...
 
-UserTypes = t.Union[discord.Member, discord.User]
+UserTypes = t.Union[discord.Member, discord.User, FetchedUser]
 MemberObject = t.Union[UserTypes, discord.Object]
 Infraction = t.Dict[str, t.Union[str, int, bool]]
 Expiry = t.Union[Duration, ISODateTime]
 
 
-def proxy_user(user_id: str) -> discord.Object:
+async def post_user(ctx: Context, user: t.Union[discord.User, discord.Object]) -> t.Optional[dict]:
```
MarkKoz: Could technically be a `discord.Member` too. Might be worth re-arranging the aliases as admittedly they're getting a bit out of hand...

bot/cogs/moderation/utils.py
```diff
@@ -25,30 +24,42 @@
...
+    Used when an infraction needs to be applied on a user absent in the guild.
     """
-    log.trace(f"Attempting to create a proxy user for the user id {user_id}.")
+    log.trace("Attempting to add user to the database.")
 
-    try:
-        user_id = int(user_id)
-    except ValueError:
-        raise commands.BadArgument
+    if not isinstance(user, discord.User):
```
MarkKoz: This won't catch `discord.Member`.
```suggestion
    if not isinstance(user, (discord.Member, discord.User)):
```

bot/converters.py
```diff
@@ -278,3 +278,50 @@ class ISODateTime(Converter):
...
+        """Convert `user_id` to a `discord.User` object, after fetching from the Discord API."""
+        try:
+            user_id = int(user_id)
+            user = await ctx.bot.fetch_user(user_id)
+        except ValueError:
+            raise BadArgument(f"The provided argument can't be turned into integer: `{user_id}`")
+        except discord.HTTPException as e:
+            # If the Discord error isn't `Unknown user`, save it in the log and return a proxy instead
+            if e.code != 10013:
+                log.warning("Failed to fetch user, returning a proxy instead.")
```
MarkKoz: Include some info about the error
```suggestion
                log.warning("Failed to fetch user, returning a proxy instead: status {e.status}")
```

bot/converters.py
```diff
@@ -278,3 +278,50 @@ class ISODateTime(Converter):
...
+                return proxy_user(user_id)
+            raise BadArgument
```
MarkKoz: I don't believe this error's message will get seen, but in case it does in the future, _some_ descriptive message should be added
```suggestion

            log.debug(f"Failed to fetch user {user_id}: user does not exist.")
            raise BadArgument(f"User `{user_id}` does not exist")
```

bot/cogs/moderation/utils.py
```diff
@@ -25,30 +24,42 @@
...
+async def post_user(ctx: Context, user: t.Union[discord.User, discord.Object]) -> t.Optional[dict]:
     """
-    Create a proxy user object from the given id.
+    Create a new user in the database.
 
-    Used when a Member or User object cannot be resolved.
+    Used when an infraction needs to be applied on a user absent in the guild.
     """
-    log.trace(f"Attempting to create a proxy user for the user id {user_id}.")
+    log.trace("Attempting to add user to the database.")
```
MarkKoz:
```suggestion
    log.trace("Attempting to add user {user.id} to the database.")
```

bot/cogs/moderation/utils.py
```diff
@@ -74,24 +85,22 @@ def proxy_user(user_id: str) -> discord.Object:
...
-                f":x: Cannot add infraction, the specified user is not known to the database."
-            )
-            return
+    # Try to apply the infraction. If it fails because the user doesn't exist, try to add it.
+    for should_post_user in (True, False):
+        try:
+            response = await ctx.bot.api_client.post('bot/infractions', json=payload)
+            return response
+        except ResponseCodeError as exp:
+            if exp.status == 400 and 'user'in exp.response_json:
```
MarkKoz:
```suggestion
            if exp.status == 400 and 'user' in exp.response_json:
```

bot/converters.py
```diff
@@ -278,3 +278,50 @@ class ISODateTime(Converter):
...
+    If the fetching is successful, a `discord.User` object is returned. If it fails and
+    the error doesn't imply the user doesn't exist, then a `discord.Object` is returned
+    via the `user_proxy` function.
+    """
+
+    @staticmethod
+    async def convert(ctx: Context, user_id: str) -> t.Union[discord.User, discord.Object]:
+        """Convert `user_id` to a `discord.User` object, after fetching from the Discord API."""
+        try:
+            user_id = int(user_id)
```
MarkKoz:
```suggestion
            log.trace(f"Fetching user {user_id}...")
            user_id = int(user_id)
```

bot/converters.py
```diff
@@ -278,3 +278,50 @@ class ISODateTime(Converter):
...
+            user = await ctx.bot.fetch_user(user_id)
```
MarkKoz: It's clearer to return up here than all the way at the bottom.

bot/converters.py
```diff
@@ -278,3 +278,50 @@ class ISODateTime(Converter):
...
+        except ValueError:
+            raise BadArgument(f"The provided argument can't be turned into integer: `{user_id}`")
```
MarkKoz:
```suggestion
            log.debug("Failed to fetch user {user_id}: could not convert to int.")
            raise BadArgument(f"User ID `{user_id}` is invalid - could not convert to an integer.")
```

bot/converters.py
```diff
@@ -278,3 +278,50 @@ class ISODateTime(Converter):
...
+    Create a proxy user object from the given id.
+
+    Used when a Member or User object cannot be resolved.
+    """
+    log.trace(f"Attempting to create a proxy user for the user id {user_id}.")
+
+    try:
+        user_id = int(user_id)
+    except ValueError:
+        raise BadArgument
```
MarkKoz:
```suggestion
        log.debug("Failed to create proxy user {user_id}: could not convert to int.")
        raise BadArgument(f"User ID `{user_id}` is invalid - could not convert to an integer.")
```

bot/cogs/moderation/utils.py
```diff
@@ -25,30 +24,42 @@
...
 
-    return user
+    try:
+        response = await ctx.bot.api_client.post('bot/users', json=payload)
+        log.trace(f"User {user.id} added to the DB.")
+        return response
+    except ResponseCodeError as e:
+        log.warn("Couldn't post user.")
+        await ctx.send(
+            "The attempt to add the user to the DB failed: "
```
MarkKoz:
```suggestion
            ":x: The attempt to add the user to the DB failed: "
```

PR #705, opened 2019-12-19
Moderation: explicitly set active value when posting infractions

MarkKoz:
Required due to changes in the API making the active field required.
See python-discord/site/pull/317

### DIFF
--- a/bot/cogs/moderation/infractions.py
+++ b/bot/cogs/moderation/infractions.py
@@ -203,7 +203,7 @@ class Infractions(InfractionScheduler, commands.Cog):
         if await utils.has_active_infraction(ctx, user, "mute"):
             return
 
-        infraction = await utils.post_infraction(ctx, user, "mute", reason, **kwargs)
+        infraction = await utils.post_infraction(ctx, user, "mute", reason, active=True, **kwargs)
         if infraction is None:
             return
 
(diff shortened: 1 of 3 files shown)


bot/cogs/moderation/infractions.py
```diff
@@ -220,7 +220,7 @@ def __init__(self, bot: Bot):
     @respect_role_hierarchy()
     async def apply_kick(self, ctx: Context, user: Member, reason: str, **kwargs) -> None:
         """Apply a kick infraction with kwargs passed to `post_infraction`."""
-        infraction = await utils.post_infraction(ctx, user, "kick", reason, **kwargs)
+        infraction = await utils.post_infraction(ctx, user, "kick", reason, active=False, **kwargs)
```
SebastiaanZ: Just curious, was  this a previously undetected bug? (We typically don't kick people anymore.)

MarkKoz: Do you mean that `active` would be `True` for a kick? Yes, it was an undetected bug. As I wrote in the other issue, the validator in the serializer could be by bypassed by not including the value at all.

SebastiaanZ: In the command route, the `active=False` is already provided by the command, so it's passed on as a kwarg:
```py
    @command()
    async def kick(self, ctx: Context, user: Member, *, reason: str = None) -> None:
        """Kick a user for the given reason."""
        await self.apply_kick(ctx, user, reason, active=False)
```

MarkKoz: Ah well it may as well stay in `apply_kick` too.

PR #710, opened 2019-12-25
Eval cog improvements 

### LINKED ISSUE #588: Write unit tests for `bot/cogs/snekbox.py`
Write unit tests for [`bot/cogs/snekbox.py`](../blob/master/bot/cogs/snekbox.py).

## Implementation details

## Additional information

### LINKED ISSUE #676: Re-evaluate snippet with emoji reaction
# Abstract
It would be nice to be able to re-evaluate the same snippet if it has been edited, using the same message via an emoji interaction. 

# Specification 
An emoji like this one 🔄 should be added to every evaluated user message, and if the user react with the same emoji on his message (along with the bot), a new eval task should trigger with the updated message content. If the message content wasn’t updated since last eval, the new eval task should silently cancel. Either way, the user reaction should be deleted and be ready to be re-evaluated again, once the last eval have finished. 

# Rationale
For now, when a eval didn’t go as plan, the user need to copy the snippet, paste it, modify it, and then send it. This take some time and some place in the chat, plus copy pasting text with markdown can be quite frustrating. Let’s make it easier!

### LINKED ISSUE #699: Eval shouldn’t strip non-blank lines
As talked in https://discordapp.com/channels/267624335836053506/635950537262759947/639484028012986369

Bot strips out /n lines 
example can be 

```python
def reformat_(a):
    len_a = 8 - len(a)
    a = " " * int(len_a / 2) + a + " " * int(len_a / 2)
    return a
print(reformat_("Hii"))
```

**Why change it?**

Changing it can lead to a better precise output.  also bot already limit the output to 10ish lines . so no harm in adding whitespaces.

### LINKED ISSUE #709: Snekbox !eval output of exactly 11 lines is truncated
If snekbox would return an output of exactly 11 lines, with 10 newline characters, it will truncate the final returning line without sending the output to be uploaded. 

The source is a mismatch in the snekbox cog, within `format_output`--the number of lines to output is set to `output.count("\n")`and checked against `> 10` but later output is sliced with `output.split("\n")[:10]`. 

.split() returns up to 11 items if there are exactly 10 `\n`'s which results in the loss of the final line of output.

Testcase:
!e ```py
print(*range(11), sep="\n")
`\``
The final element (10) is missing.

Akarys42:
Now that snekbox support some external libraries (see python-discord/snekbox#51), I think this is a good time to work on some improvements :

- [x] Feature : Re-evaluate snippet with emoji reaction (closes #676)
   - Commit 6991e5fab13893c05d6f220e71f6ffc71509c1aa
- [X] Bug : Eval is unable to run on Windows because of `Signals.SIGKILL` not being defined
   - Commit ab0d44b
- [X] Bug : Formatting strip spaces on blank lines (closes #699) 
   - Commit ab0d44b8e013694c2e92af51c6fdb8ed9239c331
- [x] Bug : The 11th line is not displayed when there is exactly 11 lines of output (closes #709) 
   - Commit e53492e
- [X] Tests : Write unit tests for `bot/cogs/snekbox.py` (closes #588)
   - Commit 8c8f8dd50f376c3398c04e7bbe76b5028b69ff83


Akarys42:
5 tasks done out of 5, it is now ready for reviewing!

ikuyarihS:
At the moment, after calling `!eval`, get a result from bot, **delete bot's message**, change the code in the original message, click the button to re-eval will raise an error

image

bot/cogs/snekbox.py
```diff
@@ -199,32 +203,64 @@ def get_status_emoji(results: dict) -> str:
...
+                    )
+
+                    log.info(f"{ctx.author}'s job had a return code of {results['returncode']}")
+            finally:
+                del self.jobs[ctx.author.id]
+
+            try:
+                _, new_message = await self.bot.wait_for(
+                    'message_edit',
+                    check=partial(predicate_eval_message_edit, ctx),
```
ikuyarihS: This will make the bot creating a partial everytime, even when the `ctx` will not be changed.

Suggestions:
Create the partial outside of the loop, and then call it inside like this
```py
_predicate_eval_message_edit = partial(predicate_eval_message_edit, ctx)

...

await self.bot.wait_for(
    'message_edit',
    check=predicate_eval_message_edit,
    timeout=10
)
```

This way, we will not create 2 new partials via every loop even when someone is re-evaluating their code a lot of times.

bot/cogs/snekbox.py
```diff
@@ -199,32 +203,64 @@ def get_status_emoji(results: dict) -> str:
...
+                    timeout=10
                 )
+                await ctx.message.add_reaction('🔁')
+                await self.bot.wait_for(
+                    'reaction_add',
+                    check=partial(predicate_eval_emoji_reaction, ctx),
```
ikuyarihS: Same with `predicate_eval_message_edit` you can create one outside of the for loop.

bot/cogs/snekbox.py
```diff
@@ -199,32 +203,64 @@ def get_status_emoji(results: dict) -> str:
...
+                    timeout=10
+                )
+
+                log.info(f"Re-evaluating message {ctx.message.id}")
+                code = new_message.content.split(' ', maxsplit=1)[1]
+                await ctx.message.clear_reactions()
```
ikuyarihS: This is always called, so you can move it outside of the loop.

bot/cogs/snekbox.py
```diff
@@ -152,7 +157,7 @@ def get_status_emoji(results: dict) -> str:
         lines = output.count("\n")
 
         if lines > 0:
-            output = output.split("\n")[:10]  # Only first 10 cause the rest is truncated anyway
+            output = output.split("\n")[:11]  # Only first 11 cause the rest is truncated anyway
```
ikuyarihS: Hardcoding and change this from `10` to `11` will lead to the same issue, just that you will be able to show 1 more line, should we think of a way to have a dynamic limitation ( show all if total characters are low ) for example? Otherwise, I can see a future where we will have another issue open for `bot only show 11 lines and truncate the 12th`.

bot/cogs/snekbox.py
```diff
@@ -152,7 +157,7 @@ def get_status_emoji(results: dict) -> str:
         lines = output.count("\n")
```
ikuyarihS: Because we are now not stripping leading `'\n'` anymore, we will have to find a different way to count lines, otherwise we can have this

image

I believe the leading `'\n'` should still be stripped, the leading `' '` *can* be left alone since it does mess up with output of `dis` and `pandas.DataFrame` if we also strip the `' '`

bot/cogs/snekbox.py
```diff
@@ -152,7 +157,7 @@ def get_status_emoji(results: dict) -> str:
...
 
         if lines > 0:
```
ikuyarihS: This entire if can be replaced with
```py
output = ["{i:03d} | {line}" for i, line in enumerate(output.split('\n'), 1)]
output = output[:11]  # Limiting to only 11 lines
```
```

File: bot/cogs/snekbox.py [hunk 7]
```diff
@@ -169,7 +174,6 @@ def get_status_emoji(results: dict) -> str:
         if truncated:
             paste_link = await self.upload_output(original_output)
 
-        output = output.strip()
         if not output:
             output = "[No output]"
```
ikuyarihS: Since `output` is a `string` you can safely do
```py
output = output or "[No output]"
```

bot/cogs/snekbox.py
Akarys42: The point was to leave all the output lines without doing any stripping logic to avoid any misleading result, since we have the paste service anyway

bot/cogs/snekbox.py
Akarys42: We want to call it after each eval, it needs to stay inside the loop

ikuyarihS: Yes I agree, I meant that you call this inside `try` and also inside `except` so you can call this line outside of the `try except` altogether.

bot/cogs/snekbox.py
ikuyarihS: I see, that is a good point, in this case please disregard this comment.

bot/cogs/snekbox.py
ikuyarihS: After a discussion in #dev-contrib, I agree that having 11 lines is fine since we also have a paste service. If this becomes a problem, we can raise another issue and handle it later.

tests/bot/cogs/test_snekbox.py
```diff
@@ -0,0 +1,363 @@
...
+        """Test the predicate_eval_message_edit function."""
+        msg0 = MockMessage()
+        msg0.id = 1
+        msg0.content = 'abc'
+        msg1 = MockMessage()
+        msg1.id = 2
+        msg1.content = 'abcdef'
+        msg2 = MockMessage()
+        msg2.id = 1
+        msg2.content = 'abcdef'
```
MarkKoz: These can be created with kwargs 

> Mocks can also be called with arbitrary keyword arguments. These will be used to set attributes on the mock after it is created.

For example,

```py
msg0 = MockMessage(id=1, content='abc')
```

tests/bot/cogs/test_snekbox.py
```diff
@@ -0,0 +1,363 @@
...
+        valid_user.id = 2
+
+        invalid_reaction_id = MockReaction()
+        invalid_reaction_id.message.id = 42
+        invalid_reaction_id.__str__.return_value = '🔁'
+        invalid_user_id = MockUser()
+        invalid_user_id.id = 42
+        invalid_reaction_str = MockReaction()
+        invalid_reaction_str.message.id = 1
+        invalid_reaction_str.__str__.return_value = ':longbeard:'
```
MarkKoz: As mentioned before, kwargs can be used to set some of these attributes.

tests/bot/cogs/test_snekbox.py
```diff
@@ -0,0 +1,363 @@
...
+
+        cases = (
+            (invalid_reaction_id, valid_user, False, 'invalid reaction ID'),
+            (valid_reaction, invalid_user_id, False, 'invalid user ID'),
+            (invalid_reaction_str, valid_user, False, 'invalid reaction __str__'),
+            (valid_reaction, valid_user, True, 'matching attributes')
+        )
+        for reaction, user, expected, testname in cases:
+            with self.subTest(msg=f'Test with {testname} and expected return {expected}'):
+                self.assertEqual(snekbox.predicate_eval_emoji_reaction(valid_ctx, reaction, user), expected)
```
MarkKoz: Reads better as separate lines.

```suggestion
                actual = snekbox.predicate_eval_emoji_reaction(valid_ctx, reaction, user)
                self.assertEqual(actual, expected)
```

tests/bot/cogs/test_snekbox.py
```diff
@@ -0,0 +1,363 @@
...
+
+        cases = (
+            (msg0, msg0, False, 'same ID, same content'),
+            (msg0, msg1, False, 'different ID, different content'),
+            (msg0, msg2, True, 'same ID, different content')
+        )
+        for ctx_msg, new_msg, expected, testname in cases:
+            with self.subTest(msg=f'Messages with {testname} return {expected}'):
+                ctx = MockContext()
+                ctx.message = ctx_msg
```
MarkKoz: Can use a kwarg to set the message attribute.

tests/bot/cogs/test_snekbox.py
```diff
@@ -0,0 +1,363 @@
...
+                self.assertEqual(snekbox.predicate_eval_message_edit(ctx, ctx_msg, new_msg), expected)
```
MarkKoz: Reads better as separate lines.
```suggestion
                actual = snekbox.predicate_eval_message_edit(ctx, ctx_msg, new_msg)
                self.assertEqual(actual, expected)
```

bot/cogs/snekbox.py
```diff
@@ -199,32 +203,64 @@ def get_status_emoji(results: dict) -> str:
...
+                    msg = f"{ctx.author.mention} {icon} {msg}.\n\n```py\n{output}\n```"
+                    if paste_link:
+                        msg = f"{msg}\nFull output: {paste_link}"
+
+                    response = await ctx.send(msg)
+                    self.bot.loop.create_task(
+                        wait_for_deletion(response, user_ids=(ctx.author.id,), client=ctx.bot)
+                    )
+
+                    log.info(f"{ctx.author}'s job had a return code of {results['returncode']}")
```
MarkKoz: This function is too large now. It seems best to have this bit be in a new function. It'll separate things into smaller units and thus make testing simpler.

bot/cogs/snekbox.py
```diff
@@ -199,32 +203,64 @@ def get_status_emoji(results: dict) -> str:
...
+                    timeout=10
+                )
+
+                log.info(f"Re-evaluating message {ctx.message.id}")
+                code = new_message.content.split(' ', maxsplit=1)[1]
+                await ctx.message.clear_reactions()
+                await response.delete()
+            except asyncio.TimeoutError:
+                await ctx.message.clear_reactions()
+                return
```
MarkKoz: This function is too large now. It seems best to have this bit also be in a new function (different than the other section I commented on). It'll separate things into smaller units and thus make testing simpler.

bot/cogs/snekbox.py
ikuyarihS: Upon further discussion, there is a `return` in the `except` block, moving `await ctx.message.clear_reactions()` outside will make it not clear any reactions if the exception block is hit, so the code is good as is, and so I'll mark this as resolved.

bot/cogs/snekbox.py
```diff
@@ -199,32 +203,68 @@ def get_status_emoji(results: dict) -> str:
...
+            finally:
+                del self.jobs[ctx.author.id]
+
+            try:
+                _, new_message = await self.bot.wait_for(
+                    'message_edit',
+                    check=_predicate_eval_message_edit,
+                    timeout=10
                 )
+                await ctx.message.add_reaction('🔁')
```
lemonsaurus: Using a unicode literal like this can sometimes be problematic. What you see in other parts of the bot is that these emojis are defined as constants that use unicode code points instead (like `"\U0001F195"`), because this tends to be a safer way of dealing with them. Some users have reported running into weird errors when the actual emoji is left inside of code like this.

bot/cogs/snekbox.py
```diff
@@ -169,12 +177,72 @@ def get_status_emoji(results: dict) -> str:
...
+            )
+            await ctx.message.add_reaction(REEVAL_EMOJI)
+            await self.bot.wait_for(
+                'reaction_add',
+                check=_predicate_emoji_reaction,
+                timeout=10
+            )
+
+            code = new_message.content.split(' ', maxsplit=1)[1]
+            await ctx.message.clear_reactions()
```
MarkKoz: `NotFound` error needs to be suppressed in case user deletes the message themselves.

bot/cogs/snekbox.py
```diff
@@ -169,12 +177,72 @@ def get_status_emoji(results: dict) -> str:
...
+            with contextlib.suppress(HTTPException):
+                await response.delete()
+
+        except asyncio.TimeoutError:
+            await ctx.message.clear_reactions()
```
MarkKoz: `NotFound` error needs to be suppressed in case user deletes the message themselves.

PR #711, opened 2019-12-25
Prevent the role syncer from wiping the database table during API latency

### LINKED ISSUE #563: Write unit tests for `bot/cogs/sync/__init__.py`
Write unit tests for [`bot/cogs/sync/__init__.py`](../blob/master/bot/cogs/sync/__init__.py).

## Implementation details

## Additional information

### LINKED ISSUE #564: Write unit tests for `bot/cogs/sync/cog.py`
Write unit tests for [`bot/cogs/sync/cog.py`](../blob/master/bot/cogs/sync/cog.py). There are already some [existing tests](../blob/master/tests/bot/cogs/sync/) for this file, which constitute a 36% branch coverage. Feel free to review the existing tests to see if they need improvement.

## Implementation details

## Additional information

### LINKED ISSUE #565: Write unit tests for `bot/cogs/sync/syncers.py`
Write unit tests for [`bot/cogs/sync/syncers.py`](../blob/master/bot/cogs/sync/syncers.py). There are already some [existing tests](../blob/master/tests/bot/cogs/sync/) for this file, which constitute a 57% branch coverage. Feel free to review the existing tests to see if they need improvement.

## Implementation details

## Additional information

### LINKED ISSUE #704: Prevent the role syncer from wiping the database table during API latency
## Summary
The role synchronization utility occasionally starts before the bot's role cache has been filled resulting in all the roles being deleted from the database table.

## Background
Since we frequently add and delete roles from our guild, we have an automated synchronization utility that keeps the database up to date with the actual roles on our guild. Currently, the mechanism waits for the `READY` signal to be dispatched by `discord.py` after the bot starts, as that typically signals that the bot's caches have been synchronized with the actual state of the guild.

However, we've observed that in rare cases, the `READY` event may be dispatched before this has happened. The most likely cause is high latency between the bot and the Discord API, resulting in the guild information only being received after the maximum delay period `discord.py` uses to wait for this information. Unfortunately, if `discord.py` does not receive the information in time, it instantiates the cache with an empty list, making it appear as if the guild has no roles.

## Related issues
I think the mechanism behind this issue also caused a number of other bugs we have observed in the past, such as channels not being available after `READY` had been dispatched resulting in `get_channel` returning `None` (e.g., part 2 of #390; which was solved by #380). I can't find individual issues other than the Reddit channel/#390, but I've observed the same bug with the watch channels as well. These instances were solved by replacing the `get_channel` method with the `fetch_channel` method, which fetches the channel directly from the API (see #380).

(*Note: all of these simple channel relays have been replaced by webhooks since.*)

## Proposed solution
While the related issues above were solved at the cost of an additional API call (`fetch_channel`), we may be able to do better. After a lot of digging by @MarkKoz and some experimentation by myself, I would like to propose a two-fold solution to prevent this from happening again:

#### 1. Use `GUILD_AVAILABLE` instead of `READY`
Instead of relying on the potentially misleading `READY` event, we should instead rely on the `GUILD_AVAILABLE` signal using the appropriate event listener. In theory, this event should only fire after the guild's data has become fully available to the bot. I think it's still a good idea to have this event listener validate the relevant caches as well. This event fires separately for every guild the bot is a member of, but since we only care about one guild anyway, a simple guild ID check could serve as the sentinel.

Now, I do think this event may also fire after the guild has become temporarily unavailable due to issues with the Discord API. (Similar to what we've all probably experienced in our clients.) However, I think this would be a useful feature, not a bug: If our bot lost the connection to Discord temporarily, it's probably a good idea to synchronize the database after the connection has been re-established to catch up with changes that have happened in the meantime. 

#### 2. Require human intervention for large diffs
Since I can't guarantee that the first part of this solution will prevent an empty role cache in all circumstances, I'd like to introduce a barrier that prevents the bot from sending a request that deletes all the roles from the database. The idea is simple: We rarely add, modify, or delete a huge number of roles within a short time period, so we could have the bot ask for confirmation before sending a very large diff to the API.

The mechanism could be fairly simple: Have the bot ping the core-dev role with a message/embed detailing the diff and make it wait for confirmation in the form of a reaction by a core-dev. Since a "freak" occurrence may happen during both the automatic synchronization and a manually invoked synchronization, I think we should add this check to both forms of synchronization.

Obviously, we'd have to think about the thresholds for these diffs, balancing the false negatives and the false positives. I think these thresholds are reasonable:

- Always ping when the diff would signal that our guild has no roles at all
- Ping when the synchronization would involve changes to 10 roles simultaneously

For the latter, it's important to realize that adding, altering, or deleting a lot of roles manually one-by-one (e.g., when reordering role positions or adding/removing code jam team roles) will trigger individual synchronization events per role. This means that such a relatively common action should not trigger the human intervention mechanism. That said, if it would, I don't think it's a huge sacrifice to make to prevent more malicious synchronizations.

## Implementation details

The implementation details still have to be discussed. For a start, even if we implement the two solutions mentioned above, I don't it's a bad idea to do an explicit `fetch_roles` to request the latest list of roles from the API before doing a bulk synchronization. It's not an option for now, but that method will be introduced in `discord.py 1.3.0` (although its release may take a while).

In addition, since we've seen related bugs in the past, we could opt for a custom `bot.wait_until_guild_available()` as a sibling to the `bot.wait_until_ready()` to implement the first part of our solution. Now that we've subclassed `discord.commands.ext.Bot`, it's trivial to add such a waiting mechanism that waits until `guild.available` is `True` and the core caches have been validated as non-empty. (I think roles, members, channels would do.) This would give us an easy way of delaying tasks that rely on the guild's information being available in a similar way to what `wait_until_ready` does.

MarkKoz:
Fixes #704 
Closes #563, closes #564, closes #565

### wait_until_guild_available

There is now a `bot.wait_until_guild_available()` coroutine which is similar to `wait_until_ready()`.
https://github.com/python-discord/bot/blob/5c394a13ce20f48b33f500083946d64b927eeadc/bot/bot.py#L85
However, it waits for a specific guild (the Python Discord guild) to be be available. When a guild is available, it means the ready event must have already been dispatched and the cache already available (most importantly, it's also available for that specific guild, which the ready event on its own does not guarantee). Thus, this coroutine is a better substitute for `wait_until_ready()`.

### Syncer confirmation prompt

A confirmation prompt is send when a large difference is detected during a sync, both for automatic and manual syncs. Confirmation is done by reacting with a ✅ or ❌ to the message and times out after 5 minutes. To facilitate sending a confirmation prompt, the syncers were overhauled using an object-oriented approach. Everything related to running a syncer is handled in its own class.

bild

bild

bild

#### Manual sync

For manual syncs, the prompt is sent by editing the initial message i.e. https://github.com/python-discord/bot/blob/5c394a13ce20f48b33f500083946d64b927eeadc/bot/cogs/sync/syncers.py#L133 The invoker of the command is not mentioned because I don't think a message edit which adds a mention will trigger a notification. Only the invoker may react to confirm or abort.

#### Automatic syncs

Automatic syncs are ones which are triggered by the bot on start-up after the guild is available. The prompt is sent by creating a new message in the `#dev-core` channel and mentions `@Core Developers`. If it can't retrieve the channel from the cache or via `fetch_channel`, then the sync is aborted. Any user with the `@Core Developers` role can react to confirm or abort.

## Concerns

1. Should the syncers remain in one module (as they are currently) or be split into separate modules?
2. Is 5 minutes a good time-out time for the confirmation?
https://github.com/python-discord/bot/blob/5c394a13ce20f48b33f500083946d64b927eeadc/bot/cogs/sync/syncers.py#L24
3. Is a maximum difference of 10 adequate? Should there be different maximum values for roles and users?
https://github.com/python-discord/bot/blob/5c394a13ce20f48b33f500083946d64b927eeadc/bot/cogs/sync/syncers.py#L25
4. A confirmation is not sent when no roles are fetched from the cache. However, with a sufficient `MAX_DIFF` set, such oddity will be caught anyway, assuming the database is not empty to begin with.
5. What should be done if the cache is detected as empty after the guild available event was dispatched? Currently it just logs and returns without setting the `asyncio.Event` (so syncers, among other things, will not start). 
https://github.com/python-discord/bot/blob/5c394a13ce20f48b33f500083946d64b927eeadc/bot/bot.py#L70 
Something would have to be pretty wrong with the Discord API for this to occur. I doubt it can recover from such an issue unless somehow the guild goes unavailable and then available again (but this time with a populated cache).

## TODO

- [x] Unit tests for Sync stuff - 100% coverage
     - Tests for `Bot` will not be done in this PR. Admittedly, I am burned out from writing unit tests. Also, #748 will include some changes so that's another excuse to hold off.
- [x] Handle API errors while syncing more gracefully
    * Abort when an error occurs or try to sync as many as possible?
    * Either update the message to state an error occurred or go as far as to show how many out of the total succeeded in syncing.
- [x] Look into updating the confirmation message with the success message and totals for automatic syncs. Would require some restructuring so that the message object can be accessed.
- [x] Prevent the following error when an `APIClient` instance is created to be used as a spec for `MockAPIClient`.
    ```
    RuntimeWarning: coroutine 'APIClient._create_session' was never awaited
    ```


bot/cogs/sync/syncers.py
```diff
@@ -1,235 +1,344 @@
...
-                'roles': list(user.roles)
-            }
+class Syncer(abc.ABC):
+    """Base class for synchronising the database with objects in the Discord cache."""
+
+    _CORE_DEV_MENTION = f"<@&{constants.Roles.core_developer}> "
+    _REACTION_EMOJIS = (constants.Emojis.check_mark, constants.Emojis.cross_mark)
+
+    CONFIRM_TIMEOUT = 60 * 5  # 5 minutes
+    MAX_DIFF = 10
```
SebastiaanZ: Why did you decide to add the constants here instead of in config-default/constants.py?

bot/cogs/sync/syncers.py
```diff
@@ -1,235 +1,344 @@
...
+        If the `author` of the prompt is a bot, then a reaction by any core developer will be
+        considered valid. Otherwise, the author of the reaction (`user`) will have to be the
+        `author` of the prompt.
+        """
+        # For automatic syncs, check for the core dev role instead of an exact author
+        has_role = any(constants.Roles.core_developer == role.id for role in user.roles)
+        return (
+            reaction.message.id == message.id
+            and not user.bot
+            and (has_role if author.bot else user == author)
```
SebastiaanZ: I'm not sure if it's strictly necessary, but this is an elegant solution to limit the check to the author for a manual sync. I guess it prevents someone from accidentally confirming a sync action started by someone else. The point of this comment is purely to say it's an elegant solution.

bot/cogs/sync/syncers.py
MarkKoz: Good idea. I've moved it there.

bot/api.py
```diff
@@ -32,6 +32,9 @@ def __str__(self):
 class APIClient:
     """Django Site API wrapper."""
 
+    session: Optional[aiohttp.ClientSession] = None
```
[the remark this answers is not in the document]
MarkKoz: Fair. I've added a comment as you suggested.

MarkKoz:
@SebastiaanZ Thanks. Any thoughts on the concerns (most importantly, № 5)?

SebastiaanZ:
> @SebastiaanZ Thanks. Any thoughts on the concerns (most importantly, № 5)?

I'm honestly not quite sure. It currently pretty much locks all features that rely on the new signal, but it really should be a freak occurrence that will probably only be solved by restarting the bot (or manually triggering that method with an internal eval to recheck the state). We could add something that tries to send an alert to the guild using a method that not relies on the cache (e.g., fetching a webhook from the API directly and using that), but if we're in that much trouble, the only way to really come back from such an incoherent state is when discord refreshed the entire guild (and I think that should send a new guild available event).

MarkKoz:
> We could add something that tries to send an alert to the guild using a method that not relies on the cache (e.g., fetching a webhook from the API directly and using that),

If you think that's helpful then sure, that could be implemented.

bot/cogs/sync/syncers.py
```diff
@@ -1,235 +1,341 @@
...
+        roles_to_update = guild_roles - db_roles - roles_to_create
+        roles_to_delete = {role for role in db_roles if role.id in deleted_role_ids}
+
+        return _Diff(roles_to_create, roles_to_update, roles_to_delete)
+
+    async def _sync(self, diff: _Diff) -> None:
+        """Synchronise the database with the role cache of `guild`."""
+        log.trace("Syncing created roles...")
+        for role in diff.created:
+            await self.bot.api_client.post('bot/roles', json={**role._asdict()})
```
lemonsaurus: Okay, I hate `json={**role._asdict()})`. Can you confirm that it doesn't work to just do `json=role._asdict()` ?

This pattern is found multiple places.

lemonsaurus: If `json` absolutely will not accept an OrderedDict, just do `json=dict(role._asdict())` instead.

bot/cogs/sync/syncers.py
```diff
@@ -1,235 +1,341 @@
...
+        log.info(f"Starting {self.name} syncer.")
+
+        message = None
+        author = self.bot.user
+        if ctx:
+            message = await ctx.send(f"📊 Synchronising {self.name}s.")
+            author = ctx.author
+
+        diff = await self._get_diff(guild)
+        totals = {k: len(v) for k, v in diff._asdict().items() if v is not None}
```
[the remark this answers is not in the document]
MarkKoz: Forgot to make it key, value. It's a casualty. Oh well.

bot/cogs/sync/syncers.py
MarkKoz: Passing an ordered dict is fine. I don't know why I was unpacking...

PR #722, opened 2020-01-12
Set up a Prometheus endpoint for metrics

jchristgit:
Export metrics about guild members and guild messages by channel into Prometheus. This starts a background aiohttp web server on port 9330 which can then be scraped as needed.

Compared to the initial idea of using StatsD to push metrics elsewhere, this has the following advantages:
- No user setup is required (the port is just opened, and done)
- Automatically exports metrics about the Python interpreter (garbage collections, system version information)
- On Linux, automatically exports process metrics (prod runs on Linux)
- Simpler networking setup (see salt repo)

Currently, this tracks guild members by their online status and guild messages by channel. The guild_id is intentionally tracked as a label such that the bot being on multiple guilds does not break the gauge.


bot/cogs/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+    See https://github.com/prometheus/client_python for metric documentation.
+    """
+
+    PREFIX = 'pydis_bot_'
+
+    def __init__(self, bot: Bot) -> None:
+        self.bot = bot
+
+        self.guild_members = Gauge(
+            name=f'{self.PREFIX}guild_members',
```
lemonsaurus: I think it'd be cleaner if `PREFIX` didn't have a trailing underscore and the underscore was instead part of this fstring.

Pipfile
```diff
@@ -19,6 +19,7 @@ deepdiff = "~=4.0"
 requests = "~=2.22"
 more_itertools = "~=7.2"
 urllib3 = ">=1.24.2,<1.25"
+prometheus-async = {extras = ["aiohttp"],version = "~=19.2"}
```
lemonsaurus: what does this extras thing do here?

bot/cogs/metrics.py
jchristgit: see 4dd4efc

Pipfile
jchristgit: Installs dependencies for running an asyncio web server (via aiohttp) for hosting the metrics. prometheus-async also supports Twisted, so you can choose your event loop integration with the extra

jos-b: I believe it's this https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

PR #726, opened 2020-01-12
Prometheus: track command completions by user.

jchristgit:


bot/cogs/metrics.py
```diff
@@ -31,8 +31,8 @@ def __init__(self, bot: Bot) -> None:
         )
         self.command_completions = Counter(
             name=f'{self.PREFIX}_command_completions',
-            documentation="Completed commands by command by guild.",
-            labelnames=('guild_id', 'command')
+            documentation="Completed commands by command by user by guild.",
```
lemonsaurus: this string is difficult to understand. Can we try to reduce the number of consecutive `by`'s a little? Surely there must be a better way to phrase this.

jchristgit: How about 99088f1?

lemonsaurus: yeah that looks good :+1:

PR #739, opened 2020-02-05
Move tools and questions guide to under resources

scragly:
After the wiki pages adjustment to the resources page, these two urls are needing to be updated to point to the new correct locations for each page.

Tools will be under resources, and Asking Good Questions is a guide, so will be under Guides.

### DIFF
--- a/bot/cogs/site.py
+++ b/bot/cogs/site.py
@@ -59,7 +59,7 @@ class Site(Cog):
     @site_group.command(name="tools")
     async def site_tools(self, ctx: Context) -> None:
         """Info about the site's Tools page."""
-        tools_url = f"{PAGES_URL}/tools"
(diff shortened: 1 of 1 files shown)


bot/cogs/site.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, bot: Bot):
     @site_group.command(name="help")
     async def site_help(self, ctx: Context) -> None:
         """Info about the site's Getting Help page."""
-        url = f"{PAGES_URL}/asking-good-questions"
+        url = f"{PAGES_URL}resources/guides/asking-good-questions"
```
scragly:
```suggestion
        url = f"{PAGES_URL}/resources/guides/asking-good-questions"
```
I'm a dumb, forgot the slash.

PR #741, opened 2020-02-06
Refactor information.py with lots of changes that makes the overall code cleaner and faster

Denayder:
Hello. I created #740 yesterday, but with lack of knowledge of forking and proper commit messages, the commit history became a slight bit messy. I'm here again with pretty much the same PR, with a few changes here and there. Thanks to @sco1 for the helpful advice so far. 

A change log can be found here and in the commit messages:
- No longer check if every role is  ` @everyone` or not. Simply skipping the first element in the `.roles` will skip the ` @everyone` role. 
- Channel and status counting has been completely refactored. It now uses `Counter()`, and uses the `__class__` attribute for channel counting.
  - I have updated the tests to reflect this. 
- No longer send a message every time a role conversion fails when using `!role`. Instead, append a list with all the failed roles, and send 1 message. This should be less spammy.
- Imports were updated a slight bit. It didn't make much sense to me that separate elements from `discord` & `typing` was imported, and then the entire module was imported. 
- Users cannot have 0 roles, so there's no need to check if there is any. 
- Changed unnecessary `if` statement to an `elif` statement.
- Made generally code look cleaner; there seemed to be a lot of unnecessary new lines and from my understanding, the general rule of thumb is that newlines should only be made whenever there's a reason to do so. 

Some concern was brought up regarding the use of the `__class__` attribute- I'm not sure if the use of `Counter()` fixes this concern or not. 

Thank you once again for giving this PR a look. Let me know if there is anything you would like changed. :)


bot/cogs/information.py
```diff
@@ -104,36 +100,18 @@ def __init__(self, bot: Bot):
...
-                online += 1
-            elif str(member.status) == "offline":
-                offline += 1
-            elif str(member.status) == "idle":
-                idle += 1
-            elif str(member.status) == "dnd":
-                dnd += 1
+        statuses = Counter({status.value: 0 for status in Status})
+        for member in ctx.guild.members:
+            statuses[member.status.value] += 1
```
MarkKoz: It's better to use the enum itself rather than its value. Strings are more prone to programmer errors.

Denayder: Changed, thank you for the feedback :)

bot/cogs/information.py
```diff
@@ -104,36 +100,18 @@ def __init__(self, bot: Bot):
...
-            elif type(channel) == CategoryChannel:
-                category_channels += 1
-            elif type(channel) == VoiceChannel:
-                voice_channels += 1
+        member_count = ctx.guild.member_count
+
+        # How many of each type of channel?
+        channels = Counter({TextChannel: 0, VoiceChannel: 0, CategoryChannel: 0})
+        for channel in ctx.guild.channels:
+            channels[channel.__class__] += 1
```
MarkKoz: Use the `discord.ChannelType` enum instead of `__class__`. A channel's type can be retrieved with `channel.type`. `group` and `private` types should be ignored.

bot/cogs/information.py
```diff
@@ -213,7 +191,7 @@ def __init__(self, bot: Bot):
                 {custom_status}
                 **Member Information**
                 Joined: {joined}
-                Roles: {roles or None}
+                Roles: {roles}
```
MarkKoz: Why was `or None` removed? It's better to show `None` than an empty string IMO.

bot/cogs/information.py
MarkKoz: I just realised the `Counter` is not being taken advantage of. The for loop can be done away with:

```py
channels = Counter((c.type for c in guild.channels))
```

The same can be done for the the statuses.

bot/cogs/information.py
```diff
@@ -146,15 +124,15 @@ def __init__(self, bot: Bot):
                 **Counts**
                 Members: {member_count:,}
                 Roles: {roles}
-                Text: {text_channels}
-                Voice: {voice_channels}
-                Channel categories: {category_channels}
+                Text Channels: {channels[TextChannel]}
+                Voice Channels: {channels[VoiceChannel]}
+                Channel categories: {channels[CategoryChannel]}
```
MarkKoz: With the `ChannelType` suggestions, this description would not be accounting for all types (most significant is the `news` type currently) or any future types that Discord may introduce. I could live with that, but if you're up for it, you could make this part of the description be dynamically generated from whatever types are inside the `channels` dictionary.

bot/cogs/information.py
Denayder: Every user has the `Developer` role per default, and users with no roles cannot use this command- making the `or None` check rather redundant. If you want it to be re-added, I can easily do so, I just thought that it wasn't needed :)

bot/cogs/information.py
Denayder: Would this not ignore `discord.Category`? I could add an if statement for that specific type, but wouldn't that be less ideal than using `__class__`?

bot/cogs/information.py
MarkKoz: Users who haven't verified i.e. gotten past the checkpoint will have no roles.

MarkKoz: To be clear, moderators can use this command to get info about other users, including those that haven't passed the checkpoint.

bot/cogs/information.py
Denayder: Call me dumb, but I'm not sure if I exactly understand what you mean here. I assume you basically mean something like this? 
```py
        channel_counts = ""
        for ch in channels:
            channel_counts += f"{str(ch).title()} channels: {channels[ch]} \n"
```
If so, that for-loop can be simplified using this. Would this be preferred instead?
```py
        channel_counts = "".join(f"{str(ch).title()} channels: {channels[ch]} \n" for ch in channels)
```
All of the changes have been made, I just wanna get this with the commit. :)

On my own testing server, this yields:
```
Voice channels: 11
Category channels: 8
Text channels: 41
```
We could sort it alphabetically too, if that is wanted.

MarkKoz: Yes, that is what I meant. Sorting it sounds like a good idea too.

MarkKoz: Also, it's perfectly fine, and perhaps even better, for this change to have been a separate commit. No need to wait.

bot/cogs/information.py
MarkKoz: In retrospect, I don't think any code needs to be written to ignore those types. Since it's only looking at guild channels, it should never encounter those channel types in the first place.

Denayder:
All the requested changes should have been made, although, I can't seem to figure out why the test is failing, even though I did update it to reflect the changes made. Help on this would be appreciated, since I have zero clue 😓

MarkKoz:
> 
> 
> All the requested changes should have been made, although, I can't seem to figure out why the test is failing, even though I did update it to reflect the changes made. Help on this would be appreciated, since I have zero clue 😓

It's due to `dedent()` not removing the leading whitespace. It only removes it if it is common to _all_ lines. The lines added by `channel_counts` lack the leading whitespace. Therefore, the whitespace is not common to all lines and it isn't remove for any of the lines.

sco1:
To expand on the above, this is what comes out of the `dedent` call with the pre-formatted channel list:

```
                **Server information**
                Created: 1 year, 11 months and 20 days ago
                Voice region: us-east
                Features: 

                **Counts**
                Members: 8
                Roles: 17
                Category channels: 12
Text channels: 69
Voice channels: 11

                **Members**
                <:status_online:470326272351010816> 1
                <:status_idle:470326266625785866> 1
                <:status_dnd:470326272082313216> 0
                <:status_offline:470326266537705472> 6
```

Which obviously doesn't match what we have in the unit test, though Discord renders it fine because they (usually) strip the leading whitespace.

One approach that can be used is to insert the channel list after `dedent` is done, using a [Template string](https://docs.python.org/3/library/string.html#template-strings), which should fix the formatting issue:

```py
# Because channel_counts lacks leading whitespace, it breaks the dedent if it's inserted directly by the
# f-string. While this is correctly formated by Discord, it makes unit testing difficult. To keep the formatting
# without joining a tuple of strings we can use a Template string to insert the already-formatted channel_counts
# after the dedent is made.
embed_description = Template(
    textwrap.dedent(f"""
        **Server information**
        Created: {created}
        Voice region: {region}
        Features: {features}

        **Counts**
        Members: {member_count:,}
        Roles: {roles}
        $channel_counts

        **Members**
        {constants.Emojis.status_online} {statuses[Status.online]:,}
        {constants.Emojis.status_idle} {statuses[Status.idle]:,}
        {constants.Emojis.status_dnd} {statuses[Status.dnd]:,}
        {constants.Emojis.status_offline} {statuses[Status.offline]:,}
    """)
).substitute({"channel_counts": channel_counts})

embed = Embed(colour=Colour.blurple(), description=embed_description)
```

Alternatively, you could also use string concatenation, with our without a join:

```py
embed = Embed(
    colour=Colour.blurple(),
    description=(
        "**Server information**\n"
        f"Created: {created}\n"
        f"Voice region: {region}\n"
        f"Features: {features}\n\n"
        "**Counts**\n"
        f"Members: {member_count:,}\n"
        f"Roles: {roles}\n"
        f"{channel_counts}\n\n"
        "**Members**\n"
        f"{constants.Emojis.status_online} {statuses[Status.online]:,}\n"
        f"{constants.Emojis.status_idle} {statuses[Status.idle]:,}\n"
        f"{constants.Emojis.status_dnd} {statuses[Status.dnd]:,}\n"
        f"{constants.Emojis.status_offline} {statuses[Status.offline]:,}\n"
    )
)
```

```py
embed = Embed(
    colour=Colour.blurple(),
    description="\n".join(
        (
            "**Server information**",
            f"Created: {created}",
            f"Voice region: {region}",
            f"Features: {features}",
            "\n**Counts**",
            f"Members: {member_count:,}",
            f"Roles: {roles}",
            f"{channel_counts}",
            "\n**Members**",
            f"{constants.Emojis.status_online} {statuses[Status.online]:,}",
            f"{constants.Emojis.status_idle} {statuses[Status.idle]:,}",
            f"{constants.Emojis.status_dnd} {statuses[Status.dnd]:,}",
            f"{constants.Emojis.status_offline} {statuses[Status.offline]:,}",
        )
    )
)
```

The first approach probably makes the most sense.

Denayder:
Apologies for the delay, and thanks for the assistance @sco1. All changes have been made and checks have been fixed. This should be ready for another review 😁

Issue #734, opened 2020-01-30
Migrate Functionality to New discord.py 1.3.x Features


MarkKoz:
Let's discuss:

* ~~Is now a good time or is there a reason to hold off?~~
    * Has been updated with #743.
* What changes would our code base require to facilitate the update?
* What new features could we make use of to improve our code base?
    * IIRC, `fetch_members` and/or `fetch_roles` could have use somewhere. Perhaps it was in the watch channel cog.
    * We have a [patch](https://github.com/python-discord/bot/blob/master/bot/patches/message_edited_at.py) for some bug. It has already been fixed since 1.2.4. Should the patch be disabled, completely removed, or left untouched (it only applies itself if it detects it needs to).
    * `utils.sleep_until` sounds useful for OT names among other things.

For reference, here is the [changelog](https://discordpy.readthedocs.io/en/latest/whats_new.html#changelog).

sco1:
As a general thought, we should strive to keep a relatively close cadence with the library when feasible. The expansion of our test suite & apparent slowing of major breaking changes in the library should help make the process much more straightforward than it has been in the past.

It's probably beneficial to keep updating separate from refactoring outside of scenarios where the two are linked. While we should absolutely be identifying & planning out refactor targets based on update library functionality, it broadens the scope of what's likely to be an already broad review.

> IIRC, `fetch_members` and/or `fetch_roles` could have use somewhere. Perhaps it was in the watch channel cog.

From the transcripts, it appears to be most recently mentioned in the context of the role syncing (#704). Probably worth a more general scrub of our usage of `get_*` vs `fetch_*` to see if there are any places where it would be more appropriate to use one over the other.

> We have a patch for some bug. Is it fixed in 1.3?

This seems like it's [fixed by `1.2.4`](https://discordpy.readthedocs.io/en/latest/whats_new.html#v1-2-4). We're currently locked to `1.2.5` so this patch is likely already unnecessary. (For reference, this patch was introduced in #433)

> `utils.sleep_until sounds` useful for OT names among other things.

Looks awesome, let's update 😄

ks129:
Can I work on it? Also, discord.py v1.3.2 is released with small bug fixes. We should use this version.
Things I think should be used:
- Commands calling: Now you can call commands like `await self.command_function(ctx, kwargs)` and no need for invoking and command getting. Should be used especially inside same Cog commands.
- `utils.sleep_until` (already mentioned)
[...]

PR #743, opened 2020-02-07
Update discord.py to 1.3.1

MarkKoz:
Related to #734. This PR only updated the dependency.

Note that production was already on 1.3.1 due to using `~=1.2` for the version which is equivalent to `>= 1.2, == 1.*` (`1.3` is permitted by the latter portion of the clause). Should this be stricter going forward e.g. `~= 1.3.1`?

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-discord-py = "~=1.2"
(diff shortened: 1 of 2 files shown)


sco1:
`~=1.3.1` makes sense, didn't notice that it got bumped in the lock file when we relocked for #737. 

I guess now we know that everything is probably maybe working fine 👍

Issue #731, opened 2020-01-23
Superstarify Not Appropriately Capturing Reason String


sco1:
The superstarify command function definition isn't correctly capturing the (optional) reason string, causing the command set the first word as the reason, if specified, or raising an error if the first word contains an apostrophe:

https://github.com/python-discord/bot/blob/33dd712d3fe046c196ac0d31e0bfaf121f469ed5/bot/cogs/moderation/superstarify.py#L106-L113

Changing to:

```py
    @command(name="superstarify", aliases=("force_nick", "star"))
    async def superstarify(
        self,
        ctx: Context,
        member: Member,
        duration: utils.Expiry,
        *,
        reason: str = None
    ) -> None:
```

Should capture the entire reason string & resolve the issue.

Issue #663, opened 2019-11-15
Adding extra functionality and refinements to !remind


scragly:
Possible improvements have been discussed for the reminder command:

- Repeating reminders
  e.g. `!remind "@Admins meeting in 5 minutes" every Sunday at 20:55`
- More natural time parsing
- DM reminders
- Allowing reminders in functional channels such as #dev-contribs which currently is disallowed for non-staff.

sco1:
Adding a couple more enhancements:

- Add reminder ID to the confirmation embed
- Add the target datetime to the confirmation embed (in addition to the currently provided timedelta)

scragly:
I seem to have forgotten to add also:
- Jump link to original invoke msg

sco1:
That was added by #466

scragly:
Oh, very nice.

sco1:
For scoping I'm going to split this into a couple discrete chunks:

Chunk the first (low hanging fruit)
  * Reminder ID in confirmation embed
  * Target datetime in confirmation embed
  * Expand whitelist for non-staff

Chunk the second (requires backend changes)
  * Repeating reminders
  * DM reminders

Chunk the third (converter change has broader implications)
  * More natural time parsing

Shivansh-007:
For the more language time parsing feature, can we make use of [`parsedatetime`](https://github.com/bear/parsedatetime)?

ChrisLovering:
we already have [`arrow`](https://arrow.readthedocs.io/en/latest/) as a dependency, so maybe we can make use of that?

Shivansh-007:
> we already have [`arrow`](https://arrow.readthedocs.io/en/latest/) as a dependency, so maybe we can make use of that?

That would do it too! Yes 👍

wookie184:
The points here other than "Repeating reminders" and "DM reminders" seem to have been completed or have their own tracking issues. I don't think repeating/dm reminders are super important so i'll close this and let anyone create separate issues for them if they have a use case.

PR #749, opened 2020-02-12
Reminder Embed Enhancements & Whitelist Expansion

sco1:
Per #663, this PR imlements the first subset of enhancements to the reminders feature.

* Expand channel whitelist for non-staff to include `#dev-contrib`
* Add reminder ID to confirmation message footer so users don't have to use `!remind list` to figure out what reminder they need to reference for changes.
* Add targeted due date to confirmation message footer for sanity checking. While the initial reminder creation confirmation currently has a humanized delta, the duration edit does not. Rather than change the edit message string, I just made the embed generic for all feedback.

Before:
image

After:
image

Also sneaking in the one-liner fix for #731

Closes: #731


bot/cogs/reminders.py
```diff
@@ -275,9 +286,12 @@ def __init__(self, bot: Bot):
             json={'content': content}
         )
 
+        # Parse the reminder expiration back into a datetime for the confirmation message
+        expiration = datetime.fromisoformat(reminder['expiration'][:-1])
```
MarkKoz: We've been using dateutil for parsing cause I suppose it's less fragile:

https://github.com/python-discord/bot/blob/master/bot/utils/time.py#L133 (ignore the microseconds replacement)

Looks like this cog has several places that could be using this instead.

bot/cogs/reminders.py
```diff
@@ -56,12 +56,21 @@ def __init__(self, bot: Bot):
...
         embed = Embed()
         embed.colour = Colour.green()
         embed.title = random.choice(POSITIVE_REPLIES)
         embed.description = on_success
+
+        if delivery_dt:
+            embed.set_footer(text=f"ID: {reminder_id}, Due: {delivery_dt.strftime('%Y-%m-%dT%H:%M:%S')}")
+        else:
+            # Reminder deletion will have a `None` `delivery_dt`
+            embed.set_footer(text=f"ID: {reminder_id}")
```
MarkKoz: If a string is made with the default value being just the ID, the due date could be conditionally appended to it in order to cut down a bit on redundant code.

bot/cogs/reminders.py
```diff
@@ -56,12 +56,21 @@ def __init__(self, bot: Bot):
                 self.schedule_task(loop, reminder["id"], reminder)
 
     @staticmethod
-    async def _send_confirmation(ctx: Context, on_success: str) -> None:
+    async def _send_confirmation(
+        ctx: Context, on_success: str, reminder_id: str, delivery_dt: Optional[datetime]
```
MarkKoz: Prevailing style is for each parameter should be on a separate line. I notice the style above was already being used before your changes, but it could be a potential improvement here to make things more consistent with the rest of the code base.

bot/cogs/reminders.py
```diff
@@ -20,7 +20,7 @@
 
 log = logging.getLogger(__name__)
 
-WHITELISTED_CHANNELS = (Channels.bot,)
+WHITELISTED_CHANNELS = (Channels.bot, Channels.devcontrib)
```
MarkKoz: Could be a config option. Or do we already have too many of those?

bot/cogs/reminders.py
```diff
@@ -182,7 +191,9 @@ def __init__(self, bot: Bot):
         # Confirm to the user that it worked.
         await self._send_confirmation(
             ctx,
-            on_success=f"Your reminder will arrive in {humanize_delta(relativedelta(expiration, now))}!"
+            on_success=f"Your reminder will arrive in {humanize_delta(relativedelta(expiration, now))}!",
```
MarkKoz: Would be more readable IMO if the humanised delta was created on a separate line and assigned to a variable.

sco1:
I've also snuck in the one-liner fix for #731

PR #750, opened 2020-02-14
Create AsyncResolver inside a coroutine to avoid DeprecationWarning

### LINKED ISSUE #748: aiohttp.AsyncResolver needs to be created inside an async function
Despite the efforts in #681, I failed to realise `aiohttp.AsyncResolver` suffers from the same `DeprecationWarning` as the session. 

Thankfully, it seems that the resolver of a `aiohttp.TCPConnector` instance can be swapped out whenever. It just uses whatever the value of `self._resolver` is every time it needs to resolve a host. It is a "private" attribute but ostensibly there isn't any reason it shouldn't be changed. 

So, the resolver can simply be created in `Bot.start()`. I believe it is safe to also close the resolver and even the connector in `Bot.close()`. I was mainly concerned by how discord.py would handle this. The connector goes to discord.py's `HTTPClient` which uses it to create an internal `aiohttp.ClientSession` - basically the same thing we use the connector for. Whenever `Client.clear()` is called, the session gets recreated using `HTTPClient._connector`. Therefore, the connector can be re-created and assigned to that attribute before `Client.clear()` is called.

Recreating the connector would be a bit clumsy. For `HTTPClient`, the `Bot.http._connector` attribute needs to be assigned the new connector instance. For `APIClient`, the `Bot.APIClient._session_args` dictionary needs to be updated with the new connector instance. I think the latter could benefit from allowing kwargs to be passed to `recreate()`, which will override the "default" `_session_args`.

MarkKoz:
Fixes #748.

aiohttp doesn't like the resolver being created outside an async function. It is currently being done in `Bot.__init__`. This PR moves it to `Bot.login()`, which is a coroutine.

Most of the work here is ensuring the client doesn't end up in a broken state if it's re-used. Admittedly, I am not confident everything was done right with regards to that. However, our use of the client will not be affected by this or the other edge cases accounted for since we simply start the bot once for the lifetime of the program.


bot/bot.py
```diff
@@ -14,18 +15,19 @@ class Bot(commands.Bot):
...
+                "If login() is called (or the bot is started), the connector will be overwritten "
+                "with an internal one"
+            )
 
-        super().__init__(*args, connector=self.connector, **kwargs)
+        super().__init__(*args, **kwargs)
 
         self.http_session: Optional[aiohttp.ClientSession] = None
-        self.api_client = api.APIClient(loop=self.loop, connector=self.connector)
+        self.api_client = api.APIClient(loop=self.loop)
```
SebastiaanZ: We can probably strip the `loop` parameter/argument from the `APIClient` in the future and use the higher-level interface of `asyncio.create_task` within the `APIClient` instead. Not really relevant for this PR, but my memory is not great, so I'm writing it down anyway.

bot/api.py
```diff
@@ -52,33 +52,49 @@ def __init__(self, loop: asyncio.AbstractEventLoop, **kwargs):
...
-        self.session = aiohttp.ClientSession(**self._session_args)
+    async def _create_session(self, **session_kwargs) -> None:
+        """
+        Create the aiohttp session with `session_kwargs` and set the ready event.
+
+        `session_kwargs` is merged with `_default_session_kwargs` and overwrites its values.
+        If an open session already exists, it will first be closed.
+        """
+        await self.close()
+        self.session = aiohttp.ClientSession(**{**self._default_session_kwargs, **session_kwargs})
```
eivl: Do we ever need to save the default values after they have been overwritten? then dict-merging might not be the best option. ChainMap would fix this issue as long as you have an empty dict as the first dict in the chain.

MarkKoz: > Do we ever need to save the default values after they have been overwritten?

No. The merged dictionary being created is one useful once since the `session_kwargs` as passed as an argument and may be different each call.

bot/bot.py
```diff
@@ -17,43 +18,88 @@ class Bot(commands.Bot):
...
         log.info(f"Cog loaded: {cog.qualified_name}")
 
     def clear(self) -> None:
-        """Clears the internal state of the bot and resets the API client."""
+        """
+        Clears the internal state of the bot and recreates the connector and sessions.
+
+        Will cause a DeprecationWarning if called outside a coroutine.
+        """
+        # Because discord.py recreates the HTTPClient session, may as well follow suite and recreate
```
lemonsaurus:
```suggestion
        # Because discord.py recreates the HTTPClient session, may as well follow suit and recreate
```

bot/api.py
eivl: Thanks for the clarification.

PR #753, opened 2020-02-16
Show created date and duration in expired infraction log

### LINKED ISSUE #685: Add created date and duration when outputting expired infraction.
# Abstract
In addition to displaying `Member`, `Actor` and `Reason` of an expired infraction, display created date as well as duration since applied.
Also as a bonus, show the member via <@id>

# Specification
Right now when an infraction expired, the bot will output this:
image

Proposal:
```
Member: <@248100771634085889> `248100771634085889`
Actor: Python#4329
Created: 2019-03-02 14:20 ( 14 months and 15 days )
Reason: Being disruptive and nonconstructive in other's help sessions, crossposting questions across help channels, disregarding TOS after people explicitly point it out, "I dont wanna spent hours reading While u can just help me" in response to being shown relevant resources. It seems you're not wanting to put the effort in when receiving help from others, and fail to learn anything from previous warnings and mutes. Don't bother returning if you can't adjust the total disregard you have.
Watching: Yes
```

# Rationale
I believe it will be helpful to know how long an user has been tempbanned / tempmuted, otherwise mods will have to either search their memory or do another extra step of `infr search`

# Motivation
I don't wanna do an extra `infr search`, I'm lazy!

MarkKoz:
Closes #685

* Creation date is displayed in the log.
* The duration between the expiration and the creation is displayed as a humanised delta. If the infraction had no expiration, then the duration is relative to the current time rather than an expiration.
* The actual actor is displayed rather than always being the bot user.
* The member is displayed as a mention rather than an ID.

Before:
bild

After:
bild

### DIFF
--- a/bot/cogs/moderation/scheduler.py
+++ b/bot/cogs/moderation/scheduler.py
@@ -309,20 +309,21 @@ class InfractionScheduler(Scheduler):
         guild = self.bot.get_guild(constants.Guild.id)
         mod_role = guild.get_role(constants.Roles.moderator)
         user_id = infraction["user"]
+        actor = infraction["actor"]
         type_ = infraction["type"]
         id_ = infraction["id"]
         inserted_at = infraction["inserted_at"]
         expiry = infraction["expires_at"]
 
+        log.info(f"Marking infraction #{id_} as inactive (expired).")
+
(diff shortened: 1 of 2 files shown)


ikuyarihS:
Hmm, in addition, should we also show the avatar of use if `.get_user()` returns a not None? So it will line up with the other embeds

image

PR #755, opened 2020-02-16
Don't suppress all errors in scheduler

### LINKED ISSUE #754: Scheduler suppresses all exceptions of the task
Suppressing all exceptions obviously hides potentially serious errors in production and also makes debugging/development difficult. It seems the original intent was to suppress `asyncio.CancelledError` only.

MarkKoz:
Closes #754 

* Suppress `CancelledError` 
    * Still not sure why this needs to be suppressed
* Raise all other exceptions which come from the task
* Remove loop parameter from `schedule_task()`
* Merge `create_task()` into `schedule_task()`


bot/cogs/moderation/scheduler.py
```diff
@@ -415,4 +415,6 @@ def mod_log(self) -> ModLog:
         expiry = dateutil.parser.isoparse(infraction["expires_at"]).replace(tzinfo=None)
         await time.wait_until(expiry)
 
-        await self.deactivate_infraction(infraction)
+        # Because deactivate_infraction() explicitly cancels this scheduled task, it runs in
+        # a separate task to avoid prematurely cancelling itself.
+        self.bot.loop.create_task(self.deactivate_infraction(infraction))
```
MarkKoz: I think this is decent, but is there a better workaround? My other idea was to take out the `cancel_task()` from `deactivate_infraction()` and have the other callers of `deactivate_infraction()` be responsible for cancelling the task. I think that's a worse solution than what's above.

SebastiaanZ: Maybe I just don't follow the flow of execution, but if you're afraid that a scheduled task is being cancelled because of its parent being cancelled, you could also use [`asyncio.shield`](https://docs.python.org/3/library/asyncio-task.html#asyncio.shield) to prevent it from being cancelled in such a situation.

MarkKoz: The scheduler base class holds onto a `Task` for `_scheduled_task()`. Inside `_scheduled_task()`, the coroutine `deactivate_infraction()` is awaited. Inside this coroutine, the aforementioned task is cancelled. 

https://github.com/python-discord/bot/blob/f09d6b0646edb62806b60b145986b7cc680fd77c/bot/cogs/moderation/scheduler.py#L381

So a child is effectively cancelling its parent before the child itself can finish. Hope that makes sense. Does shield work to prevent that situation? I'm not familiar with it.

MarkKoz: I am realising there is a side issue here when an infraction gets edited and re-scheduled in the middle of that same infraction being deactivated. Still, it's good that a deactivation doesn't get aborted half-way (could leave it in a poor state e.g. expired in the DB but the user still has the mute role).

bot/utils/scheduling.py
```diff
@@ -29,46 +31,58 @@ def __init__(self):
...
+        if task.cancelled():
+            with contextlib.suppress(asyncio.CancelledError):
+                task.exception()
+        else:
+            # Check if it exists to avoid logging a warning.
+            if task_id in self._scheduled_tasks:
+                # Only cancel if the task is not cancelled to avoid a race condition when a new
+                # task is scheduled using the same ID. Reminders do this when re-scheduling after
+                # editing.
+                self.cancel_task(task_id)
```
SebastiaanZ: Is there any chance of a task ending with another type of exception than `CancelledError` and a new task being rescheduled with the same `id` before this callback is executed? (Callbacks are scheduled with "call_soon", so they don't happen directly after the task is done, although they should happen "soon".)

If so, I think this will cancel the new task as we're cancelling it purely using the `task_id` which, if I read this comment right, may or may not assigned to the task this callback was initiated for.

If that's not realistic or this race condition is just a figment of my imagination (which it probably is), then just ignore this. Otherwise, an option like `if self._scheduled_tasks.get(task_id) is task` could work instead. That not only guarantees the `task_id` still exists but also that it's bound to the same task.

(On a side note, you're purely calling `cancel_task` to delete the reference in `scheduled_tasks` with additional logging right? There's no need to truly cancel it since we're inside of a `done` callback. The term "cancel" confused at first, because if the task were truly cancelled, the `task.exception()` that follows would propagate that `CancelledError`. However, after thinking about it, this can't happen since we're never actually cancelling a task here; we're just cleaning up our scheduled task references.)

bot/utils/scheduling.py
```diff
@@ -13,9 +14,10 @@ class Scheduler(metaclass=CogABCMeta):
     """Task scheduler."""
 
     def __init__(self):
+        # Keep track of the child cog's name so the logs are clear.
+        self.cog_name = self.__class__.__name__
 
-        self.cog_name = self.__class__.__name__  # keep track of the child cog's name so the logs are clear.
-        self.scheduled_tasks: Dict[str, asyncio.Task] = {}
+        self._scheduled_tasks: Dict[str, asyncio.Task] = {}
```
SebastiaanZ: I like this change; it signals that we really should be using the Scheduler API instead of manipulating the mapping of scheduled tasks directly (which bypasses the checks, callbacks, and logging steps done below).

bot/utils/scheduling.py
MarkKoz: You've brought up some good points I think. I'm going to try to rework the task cancellation.

bot/cogs/moderation/scheduler.py
MarkKoz: Was just being consistent with the rest of the code but you're right, we should start taking opportunities to phase out `loop.whatever` where applicable.

Anyway, this has been replaced with a shield as you suggest so it's no longer relevant.

bot/utils/scheduling.py
MarkKoz: OK, I came up with a much better and simpler solution. I don't know what the hell I was thinking before but the previously implementation was just wrong in several ways. See commit messages for details.

bot/utils/scheduling.py
```diff
@@ -29,46 +31,70 @@ def __init__(self):
...
 
-def _silent_exception(future: asyncio.Future) -> None:
-    """Suppress future's exception."""
-    with contextlib.suppress(Exception):
-        future.exception()
+        with contextlib.suppress(asyncio.CancelledError):
+            exception = done_task.exception()
+            # Raise the exception if one exists.
+            if exception:
+                raise exception
```
MarkKoz: Does it make sense to raise it or should it be logged instead? Haven't tested this yet. I am guessing that if raised, it will just show up as 

```
asyncio | ERROR | Task exception was never retrieved
```

If logged here, more context could be provided.

MarkKoz: Tested it and here is what the exception looks like when raised:

```
2020-02-27 10:09:00,551 | asyncio | ERROR | Exception in callback Scheduler._task_done_callback(34)(<Task finishe...Error('aaah')>)
handle: <Handle Scheduler._task_done_callback(34)(<Task finishe...Error('aaah')>)>
  File "/opt/pyenv/versions/3.7.5/lib/python3.7/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/home/mark/repos/python/bot-pydis/bot/utils/scheduling.py", line 100, in _task_done_callback
    raise exception
  File "/home/mark/repos/python/bot-pydis/bot/cogs/moderation/scheduler.py", line 434, in _scheduled_task
    await asyncio.shield(self.deactivate_infraction(infraction))
  File "/home/mark/repos/python/bot-pydis/bot/cogs/moderation/scheduler.py", line 310, in deactivate_infraction
    raise ValueError("aaah")
ValueError: aaah
```

If logged:

```
2020-02-27 10:11:42,925 | bot.utils.scheduling | ERROR | Error in #35 140115635934496!
  File "/home/mark/repos/python/bot-pydis/bot/cogs/moderation/scheduler.py", line 434, in _scheduled_task
    await asyncio.shield(self.deactivate_infraction(infraction))
  File "/home/mark/repos/python/bot-pydis/bot/cogs/moderation/scheduler.py", line 310, in deactivate_infraction
    raise ValueError("aaah")
ValueError: aaah
```

MarkKoz: Logging it ourselves looks nicer. However, a user wouldn't know that an error occurred either way. How could they be notified (if applicable i.e. if it was ultimately invoked by a user)?

SebastiaanZ: I think that user-specific feedback would be rather difficult to accomplish, since there could be a long time between the user action and the failing of scheduled task. It also creates the issue of logistics: How are we going to know what to do from within the generic "done callback"?

Another option would be that, for more critical schedulers, we add an additional, scheduler-specific callback that handles it in some way. For the moderation scheduler, this could be a callback that has the Member that created the task added in using partial that, if a task finished with an error, sends an alert to mod-log accompanied with a ping to the original actor.

I'm mainly just thinking aloud here, though.

MarkKoz:
Okay I think I'm going to log it for now because that's that looks better.
This change feels out of the scope for this pull request because it may
require a relatively significant redesign. As we said we can redesign it so
it's not inherited but composition instead and maybe that pull request can
also include your idea.

Den tors 27 feb. 2020 12:25Sebastiaan Zeeff <notifications@github.com>
skrev:

bot/utils/scheduling.py
```diff
@@ -29,46 +31,73 @@ def __init__(self):
...
-    with contextlib.suppress(Exception):
-        future.exception()
+        with contextlib.suppress(asyncio.CancelledError):
+            exception = done_task.exception()
+            # Log the exception if one exists.
+            if exception:
+                log.error(
+                    f"{self.cog_name}: error in task #{task_id} {id(scheduled_task)}!",
+                    exc_info=exception
+                )
```
SebastiaanZ: This is a smart way of doing it and it handles the three cases I can think of nicely. I also like the logging approach we've got here and I think it's best we can do for now. We could look into signalling an exception back to the user directly, but, honestly, the people that matter get to see it anyway as long as we're using Sentry because if exceptions are raised within a task, they should either be handled or it's a bug waiting to be fixed.

I'm sure we as core-devs can contact the user if it was a mission critical task.

PR #760, opened 2020-02-18
Feature: suggest command usage for misspelt commands

### LINKED ISSUE #742: Suggest a list of available commands in case of typos
When we try to invoke a command, we can make little typos, for example
`!muet @shirayuki 72H oh come on, cant you make a proper issue?`

In this case, the bot can then steps in and do a friendly suggestions:
`did you perhaps mean`
`!mute @shirayuki 72H oh come on, cant you make a proper issue?`

The typo in this example was `!muet` when trying to do `!mute`

This will greatly relieve the stress of typing every command rights, and thus enhancing the experience with the bot much further.

F4zii:
Feature: suggest similar command(s) to a misspelt command

Resolves #742


F4zii:
Hey mark, as I've said, the current structure of the error_handler is a bit hard to change for that usage (since it reinvokes every misspelt command as a tag), I can wait or change its implementation if needed
I agree the tags aren't the place for this, I got a suggestion to base it there since its the easiest way, I will now work on a better way and update you :+1: 
I also had an idea to use the returned value from the command invoke into a useful piece of data

F4zii:
Hello, the feature is ready.
A quick question:  should we suggest a few options or just the closest one?

Note: Few suggestions could result in a spam

bot/cogs/tags.py
```diff
@@ -92,9 +92,8 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
         """Show all known tags, a single tag, or run a subcommand."""
         await ctx.invoke(self.get_command, tag_name=tag_name)
 
-    @tags_group.command(name='get', aliases=('show', 'g'))
-    async def get_command(self, ctx: Context, *, tag_name: TagNameConverter = None) -> None:
-        """Get a specified tag, or a list of all tags if no tag is specified."""
+    async def _get_command(self, ctx: Context, tag_name: TagNameConverter = None) -> None:
+
         def _command_on_cooldown(tag_name: str) -> bool:
```
MarkKoz: While you're at it, it'd be nice if this function could get moved out rather than being nested.

bot/cogs/tags.py
```diff
@@ -92,9 +92,8 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
         """Show all known tags, a single tag, or run a subcommand."""
         await ctx.invoke(self.get_command, tag_name=tag_name)
 
-    @tags_group.command(name='get', aliases=('show', 'g'))
-    async def get_command(self, ctx: Context, *, tag_name: TagNameConverter = None) -> None:
-        """Get a specified tag, or a list of all tags if no tag is specified."""
+    async def _get_command(self, ctx: Context, tag_name: TagNameConverter = None) -> None:
```
MarkKoz: Correct the annotations
```suggestion
    async def _get_command(self, ctx: Context, tag_name: str = None) -> bool:
```

MarkKoz: Considering it is accessed elsewhere (i.e. in `error_handler`), it doesn't seem appropriate to make this function "private" by prefixing its name with an underscore. And since it won't be private, it should have a docstring (explain what it returns too).

bot/cogs/tags.py
```diff
@@ -146,6 +147,7 @@ def _command_on_cooldown(tag_name: str) -> bool:
                     description="**There are no tags in the database!**",
                     colour=Colour.red()
                 ))
+                return True
```
MarkKoz: It'd be better to return `False` so that it can try to suggest a command instead. The embed saying "there are no tags" would still be visible, but I think it should just not be shown at all if it's being called from the error handler (can add a boolean parameter to the function to distinguish it).

bot/cogs/error_handler.py
```diff
@@ -87,9 +88,29 @@ def __init__(self, bot: Bot):
                     return
 
                 # Return to not raise the exception
-                with contextlib.suppress(ResponseCodeError):
-                    await ctx.invoke(tags_get_command, tag_name=ctx.invoked_with)
```
MarkKoz: Is suppressing this exception no longer necessary? Maybe not since the tag suggestion feature, but not sure.

bot/cogs/error_handler.py
```diff
@@ -87,9 +88,29 @@ def __init__(self, bot: Bot):
                     return
 
                 # Return to not raise the exception
```
MarkKoz: This comment may be irrelevant now.

bot/cogs/error_handler.py
```diff
@@ -87,9 +88,29 @@ def __init__(self, bot: Bot):
...
-                with contextlib.suppress(ResponseCodeError):
-                    await ctx.invoke(tags_get_command, tag_name=ctx.invoked_with)
+                tags_cog = self.bot.get_cog("Tags")
```
MarkKoz: Should check if it's `None` in case the cog isn't loaded.

bot/cogs/error_handler.py
```diff
@@ -87,9 +88,29 @@ def __init__(self, bot: Bot):
...
+                sent = await tags_cog._get_command(ctx, command_name)
+                # sent = await tags_get_command.callback(tags_get_command.cog, ctx, ctx.invoked_with)
```
MarkKoz: This can be removed.

bot/cogs/error_handler.py
```diff
@@ -87,9 +88,29 @@ def __init__(self, bot: Bot):
...
+                if sent:
                     return
+                # No similar tag found, or tag on cooldown -
+                # searching for a similar command
+                raw_commands = [
+                    (cmd.name, *cmd.aliases)
+                    for cmd in self.bot.walk_commands()
+                    if not cmd.hidden
```
MarkKoz: Can't these two comprehensions be combined into one? What even is the purpose of the second one? Flattening the list?

bot/cogs/error_handler.py
```diff
@@ -87,9 +88,29 @@ def __init__(self, bot: Bot):
...
+                ]
+                raw_commands = [c for data in raw_commands for c in data]
+                similar_command_data = difflib.get_close_matches(command_name, raw_commands, 1)
+                similar_command_name = similar_command_data[0]
+                similar_command = self.bot.get_command(similar_command_name)
+                if similar_command.can_run(ctx):
+                    misspelled_content = ctx.message.content
+                    await ctx.send(
+                        f"Did you mean:\n**{misspelled_content.replace(command_name, similar_command_name)}**",
```
MarkKoz: This could mistakenly replace something other than the command name too. Should use a more precise method of replacing which only affects the command name.

bot/cogs/error_handler.py
```diff
@@ -87,9 +88,29 @@ def __init__(self, bot: Bot):
...
+                if sent:
                     return
+                # No similar tag found, or tag on cooldown -
+                # searching for a similar command
+                raw_commands = [
+                    (cmd.name, *cmd.aliases)
+                    for cmd in self.bot.walk_commands()
+                    if not cmd.hidden
+                ]
+                raw_commands = [c for data in raw_commands for c in data]
```
MarkKoz: Perhaps this could be cached but it seems a bit complicated to keep it up to date considering commands can be added/removed during runtime (e.g. via unloading extensions). Not sure if worth looking into.

bot/cogs/error_handler.py
```diff
@@ -87,9 +88,29 @@ def __init__(self, bot: Bot):
...
+                similar_command_data = difflib.get_close_matches(command_name, raw_commands, 1)
+                similar_command_name = similar_command_data[0]
+                similar_command = self.bot.get_command(similar_command_name)
+                if similar_command.can_run(ctx):
```
MarkKoz: This will raise exceptions for check failures so they should be caught.

bot/cogs/tags.py
F4zii: Alright, right away

bot/cogs/tags.py
F4zii: Are you referring to _command_on_cooldown?

bot/cogs/tags.py
F4zii: That's correct, I think i meant to do this, my bad

bot/cogs/tags.py
F4zii: This is the command itself, are you referring to the explicit one we separated?

F4zii: The outer function (the one that is being called inside the command) is accessible within the error handler, I applied these changes to the outer function

bot/cogs/tags.py
MarkKoz: Yes

bot/cogs/tags.py
MarkKoz: Was referring to `_get_command` to be clear.

bot/cogs/tags.py
```diff
@@ -92,7 +92,18 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
...
+        """
+        Shows a specific tag or a suggestion within a given context.
+
+        params:
+            ctx - a Context object , being used to send and choose the right tag
+            tag_name - a string representing the searched tag name
+
+        Returns False if no tag is not found or the tag is on cooldown
+
+        Return True if found and sent a tag
```
MarkKoz: Keeping in mind this is not strictly enforced, here are my suggestions for writing docstrings:

1. Use the imperative mood.
2. Try to include information about the return value in the summary, if it fits.
3. A parameters/arguments section is generally something we don't use. If they warrant explicit explanation, then that information should be provided as sentences/paragraphs in the body of the docstring.

My attempt, feel free to modify:

```py
        """
        Show contents of the tag `tag_name` in `ctx` and return True if something is shown.

        If a tag is not found, display similar tag names as suggestions. If a tag is not specified,
        display a paginated embed of all tags.

        Tags are on cooldowns on a per-tag, per-channel basis. If a tag is on cooldown, display
        nothing and return False.
        """
```

bot/cogs/tags.py
```diff
@@ -92,7 +92,18 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
         """Show all known tags, a single tag, or run a subcommand."""
         await ctx.invoke(self.get_command, tag_name=tag_name)
 
-    async def _get_command(self, ctx: Context, tag_name: TagNameConverter = None) -> None:
+    async def get_command(self, ctx: Context, tag_name: str = None) -> bool:
```
MarkKoz: This isn't a command so it should not be named as such. `display_tag` is better, for starters.

bot/cogs/tags.py
```diff
@@ -160,12 +171,12 @@ def _command_on_cooldown(tag_name: str) -> bool:
...
                 return True
 
-        return False
+        return True
 
     @tags_group.command(name='get', aliases=('show', 'g'))
-    async def get_command(self, ctx: Context, *, tag_name: TagNameConverter = None) -> None:
-        """Get a specified tag, or a list of all tags if no tag is specified."""
-        await self._get_command(ctx, tag_name)
+    async def _get_command(self, ctx: Context, *, tag_name: TagNameConverter = None) -> None:
```
MarkKoz: This name could have just been left alone since, as I mentioned, `get_commands` is not an appropriate name for the other function anyway.

bot/cogs/error_handler.py
```diff
@@ -86,10 +92,33 @@ def __init__(self, bot: Bot):
...
+                with contextlib.suppress(CommandError):
+                    if similar_command.can_run(ctx):
+                        misspelled_content = ctx.message.content
+                        e = Embed()
+                        e.set_author(name="Did you mean:", icon_url=Icons.questionmark)
+                        e.description = f"{misspelled_content.replace(command_name, similar_command_name)}"
+                        await ctx.send(
+                            embed=e,
+                            delete_after=7.0
+                        )
```
MarkKoz: This doesn't need to be broken up into multiple lines.

bot/cogs/error_handler.py
F4zii: Yeah, from List[Tuple] (command_name, aliases...) into a list of command names and aliases

bot/cogs/error_handler.py
F4zii: I will note it :ok_hand:

bot/cogs/error_handler.py
MarkKoz: Wouldn't this single loop be equivalent to the two list comprehensions?

```py
raw_commands = []
for cmd in self.bot.walk_commands():
    if not cmd.hidden:
        raw_commands += (cmd.name, *cmd.aliases)
```

bot/cogs/error_handler.py
```diff
@@ -108,16 +108,20 @@ def __init__(self, bot: Bot):
...
-                            delete_after=7.0
-                        )
+                log_msg = "Cancelling attempt to suggest a command due to failed checks."
+                try:
+                    if not similar_command.can_run(ctx):
+                        log.debug(log_msg)
+                        return
+                except CommandError as cmd_error:
+                    log.debug(log_msg)
+                    await self.on_command_error(ctx, cmd_error)
```
MarkKoz: Need to return after this otherwise the embed still gets sent.

bot/cogs/error_handler.py
```diff
@@ -86,10 +92,36 @@ def __init__(self, bot: Bot):
...
+                for cmd in self.bot.walk_commands():
+                    if not cmd.hidden:
+                        raw_commands += (cmd.name, *cmd.aliases)
+                similar_command_data = difflib.get_close_matches(command_name, raw_commands, 1)
+                similar_command_name = similar_command_data[0]
+                similar_command = self.bot.get_command(similar_command_name)
+
+                log_msg = "Cancelling attempt to suggest a command due to failed checks."
+                try:
+                    if not similar_command.can_run(ctx):
```
MarkKoz: `can_run` needs to be awaited.

bot/cogs/tags.py
MarkKoz: This still needs to be addressed.

F4zii: I changed the implementation to return false, I will check what is wrong.

MarkKoz: I think you are confused about what this comment refers to. It's specifically regarding an embed being sent with the message `"**There are no tags in the database!**"`, which I'm suggesting should _not_ be sent at all if called from the error handler (i.e. merely returning `False` will not be adequate).

F4zii: Oh ok, but we also need to return false at the end of the function, correct?

MarkKoz: Don't think it matters since that line will never be reached (every branch of execution already has a return statement), but I suppose no harm in keeping it.

F4zii: Yeah, correct

MarkKoz: Like I said, 

> merely returning False will not be adequate

The **embed should not be sent** when this function is called by the error handler. I've proposed the solution of adding a parameter to indicate that.

bot/cogs/tags.py
```diff
@@ -146,7 +146,7 @@ def command_on_cooldown(self, ctx: Context, tag_name: str) -> bool:
                     title='Did you mean ...',
                     description='\n'.join(tag['title'] for tag in founds[:10])
                 ))
-                return True
+                return False
```
MarkKoz: This should still return True otherwise the bot will send a command suggestion after the tag suggestions, which is not desirable.

F4zii: Yeah, I will add a parameter in that case

MarkKoz:
I don't know how you merged but you did something wrong cause all the commits are co-authored by you and that should absolutely not happen. Do you know how to fix this? If not, I can attempt a fix myself.

F4zii:
> I don't know how you merged but you did something wrong cause all the commits are co-authored by you and that should absolutely not happen. Do you know how to fix this? If not, I can attempt a fix myself.

I think it'll be better if you try, I just meged master into my branch, weird

MarkKoz:
Okay. Please don't push anything until I let you know I've fixed it.

bot/cogs/tags.py
F4zii: Both **Current tags** and **There are no tags in the database!** embeds shouldn't be sent if the error handler invoked the function, correct?

MarkKoz:
OK should be good to go. By the way, there was a lint error - `contextlib` was imported but unused. I fixed it. Do you not have the pre-commit hook installed?

F4zii:
> OK should be good to go. By the way, there was a lint error - `contextlib` was imported but unused. I fixed it. Do you not have the pre-commit hook installed?

Yeah I fixed it already, maybe the merge / pull changed it, thanks for fixing it

bot/cogs/tags.py
MarkKoz: Sorry, never mind. I just realised those only get sent if `tag_name` is None. The error handler never sends `None` as the name so nothing needs to be changed.

bot/cogs/error_handler.py
```diff
@@ -74,9 +75,13 @@ def __init__(self, bot: Bot):
         # Try to look for a tag with the command's name if the command isn't found.
         if isinstance(e, CommandNotFound) and not hasattr(ctx, "invoked_from_error_handler"):
             if not ctx.channel.id == Channels.verification:
+                tags_cog = self.bot.get_cog("Tags")
                 tags_get_command = self.bot.get_command("tags get")
-                ctx.invoked_from_error_handler = True
+                if not tags_cog and not tags_get_command:
```
```suggestion
                if not tags_cog or not tags_get_command:
```

F4zii: Good catch, fixing...

bot/cogs/tags.py
F4zii: Yeah, that makes sense, cool, releasing latest commits and i think we are there

ikuyarihS:
Right now there is a rare chance that it can be a tag instead, can we move it up to check for commands, specially if invoker is a mod+?

image

bot/cogs/tags.py
ikuyarihS: I concur that this should return a True, otherwise this is possible

image

ikuyarihS (dismissed):

F4zii:
> Right now there is a rare chance that it can be a tag instead, can we move it up to check for commands, specially if invoker is a mod+?
> 
> image

@ikuyarihS, I think it's enough to check if the Command was invoked with arguments since if that's a single tag, no following arguments will be provided

lemonsaurus:
> @ikuyarihS, I think it's enough to check if the Command was invoked with arguments since if that's a single tag, no following arguments will be provided

I'm not sure I buy this line of reasoning. some moderation commands (like `!silence`) do not take arguments. I think prefering commands over tags is probably the right suggestion.

@F4zii Do you have time to get this wrapped up? It looks like it's fallen victim to Covid19 and we should try to nurse it back to health and get it merged.

Xithrius:
I screwed up, I am sorry.

Xithrius:
Fixed.

Xithrius:
Works with long commands and their arguments.
image

Works with the smallest commands as well.
image

So far I haven't found any bugs with this command and the way that it's implemented.

Still deciding if this should exist within mod channels at all, since it may just be noise at that point to the mods. At the same time, the error stays up for 10 seconds, so instead of having to retype/edit the command, it can be copy/pasted.

Xithrius:
A user with no roles executing a moderator+ only command along with a regular command (typed incorrectly) that anyone can execute.
image

Xithrius:
This is the outcome of a recent poll. I think the conclusion that I can reach here is that only helpers and below should get command suggestions.
image

Xithrius:
Here are examples of a user at different roles executing commands incorrectly.

No-role user:
image


Helper:
image

Moderator:
image

Admin:
image

jb3 (approved):

jerbob (approved):

PR #757, opened 2020-02-17
Handle more types of CommandError

MarkKoz:
Close #131

* Send an error message for all types of `UserInputError`. The `"Something about your input seems off. Check the arguments:"` message was too ambiguous to be of much help to users. The messages mostly use the default error messages defined in the exceptions.
* Send and error message for `BotMissingRole` and `BotMissingAnyRole`. It shares the message with `BotMissingPermissions`. It's a cleaner and less dramatic way of handling these errors even though they will rarely be seen.
* `on_command_error` was split into several functions. Some focus on handling only specific subtypes of errors. This makes the code more readable.

PR #762, opened 2020-02-21
Sentry SDK setup

jos-b:
Implement the Sentry SDK to the bot.

This PR will also setup basic tags to link errors together.


jos-b:
image

jos-b:
This PR has also had to remove the API logging handler (on our plan anyway) since it was blocking sentry from making it's API calls.

bot/__main__.py
```diff
@@ -1,10 +1,16 @@
...
 from discord.ext.commands import when_mentioned_or
+from sentry_sdk.integrations.aiohttp import AioHttpIntegration
 
 from bot import patches
 from bot.bot import Bot
 from bot.constants import Bot as BotConfig, DEBUG_MODE
 
+sentry_sdk.init(
+    dsn=BotConfig.sentry_dsn,
+    integrations=[AioHttpIntegration()]
```
MarkKoz: Is this going to complain about not being created inside an async function?

jos-b: It won't but I have just realised this is intended for aiohttp servers, not clients.

https://docs.sentry.io/platforms/python/aiohttp/

bot/cogs/error_handler.py
```diff
@@ -147,10 +148,18 @@ def __init__(self, bot: Bot):
             f"Sorry, an unexpected error occurred. Please let us know!\n\n"
             f"```{e.__class__.__name__}: {e}```"
         )
-        log.error(
-            f"Error executing command invoked by {ctx.message.author}: {ctx.message.content}"
-        )
-        raise e
```
MarkKoz: Is this still going to be logged to stdout? (by the way, in #757 I changed this to use `exec_info` instead of raising)

jos-b: log.error should go to stdout shouldn't it?

MarkKoz: Well yes but the point is that you've completely removed these lines.

jos-b: Will use exc_info though, I like that

jos-b: Oh wait, I didn't understand this right, I removed the log.error, adding back.

MarkKoz: Since you're still doing `capture_exception`, will Sentry not end up with duplicate entries for this exception?

jos-b: Yeah, I've removed the capture_exception and added log.error:

image

image

MarkKoz: It still gets those extra tags set?

jos-b: Yeah, because I run the log command from inside a scope I've created

MarkKoz:
Is this ready to be merged @jos-b? Is the environment variable set in prod?

jos-b:
@SebastiaanZ It's something which we can adapt as we find a workflow which suits us well. Right now just storing errors somewhere is useful and we'll find out if we need anything else as time goes on.

@MarkKoz the variable is in prod 👍

PR #782, opened 2020-02-23
Rename constants to match their guild counterparts

### LINKED ISSUE #496: updating/upgrading contributing bot tutorial
Not sure if it the right place, but:

This page: https://pythondiscord.com/pages/contributing/bot/ could be better (although it is already pretty good)

Rename `#bot-commands`and `#ot-offtopics` to `bot` and `off-topic-0` to actually meet with their names in `config.yml`. Explicit more that the `bot_token` has to be put in the `.env`file.

Precise somewhere that not all cogs are loaded in `DEBUG` mode, I lost one hour trying to figure why `verification.py` was not loading.

What do you think?

MarkKoz:
Resolves #496 

Contributors will have an easier time configuring the constants if their names line up with the names in the guild. Not all names were changed to be exact; mostly the staff-only channels were left alone since they were already adequately descriptive.

* Use underscores for spaces (mostly changed the log channel names)
* Roles and channels were "split" by category in the config by putting comments above different "categories"
* Some unused constants were removed

### DIFF
--- a/bot/__main__.py
+++ b/bot/__main__.py
@@ -7,7 +7,7 @@ from sentry_sdk.integrations.logging import LoggingIntegration
 
 from bot import patches
 from bot.bot import Bot
-from bot.constants import Bot as BotConfig, DEBUG_MODE
+from bot.constants import Bot as BotConfig
 
 sentry_logging = LoggingIntegration(
     level=logging.TRACE,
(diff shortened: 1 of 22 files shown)


sco1:
I've resolved the merge conflict & added the role & channel groupings to the YAML that were only present in `constants.py`.

I still need to test the changes but I'm not currently at a machine capable of doing so.

PR #783, opened 2020-02-23
Migrate bot to Python 3.8

### LINKED ISSUE #758: Upgrading to Python 3.8
## Summary
We should upgrade the Python version of the bot to Python 3.8

## Rationale
Python 3.8 offers significant improvements in the area of writing unittests for asyncio; in fact, you could say that Python 3.8 finally brings proper asyncio-support to Python's built-in unittest framework. By upgrading to Python 3.8, we'd no longer have to implement custom mock classes and decorators to make our test suite compatible with the plentiful asynchronous features of the bot that we're trying to test.

## Considerations
One reason to potentially hold back an upgrade is the relative young age of Python 3.8. However, with the release of Python 3.8.1, I'm not convinced we have to wait longer with an upgrade; I'd consider Python 3.8 to be mature enough at this point.

Another obvious consideration is whether or not all of our dependencies support Python 3.8. After upgrading my virtual environment to Python 3.8 locally, I can say that I haven't discerned any issues both in creating the dependency graph as well as running the bot. The test suite, while not showing major issues (i.e., some tests fail due to minor issues), does need to be updated to make use of the now built-in asyncio support in everything unittest.

Edit: The minor issue seen in the test suite was caused by an upstream bug in discord.py instead. I've submitted a PR to fix it: Rapptz/discord.py#2570.

SebastiaanZ:
I have upgraded the Python version we use for the bot to Python 3.8. One of the major upgrades this offers for us is that we can now use the utilities for testing asynchronous code instead of having to define our own Async-helper for every asyncio-aspect we'd like to test.

This means that this PR also removes those helpers and introduced the proper equivalents that now ship with Python where needed. Another major change that impacts the run time of the test suite significantly is that the CustomMockMixin now adapts the new "lazy" way of detecting which attributes of objects should be mocked using an AsyncMock.

### Changes to our test helpers

- There's no longer a need for the `async_test` decorator to be able to define coroutine-based test methods. Python 3.8's unittest module now ships with a new test case class, `unittest.IsolatedAsyncioTestCase`, that allows you to use `async def`-based test methods directly. It also makes sure to clean up after each test method so we truly run each test method in isolation instead of with a "polluted" event loop (hence the name, *Isolated*AsyncioTestCase)

- Our custom mock objects no longer strictly need an `__init__` method. Instead, you can define the object we want to use for `spec_set` as a class attribute, `spec_set`. The class attribute `additional_spec_asyncs` can be used to define an iterable of attribute names (strings) that should be mocked as AsyncMock in addition to the ones unittest can detect automatically. (Note: it's able to detect everything that is an actual coroutine, but some regular methods of d.py objects also return coroutine objects and those require such custom specification).

- The mocks included in `unittest.mock` now support the asynchronous iterator and asynchronous context manager protocols, so our custom versions were removed. The specification of return values is slightly different now as it now just works as in Python 3.8 docs.

- Our custom `LoggingTestCase` subclass that adds an additional logging assertion method was using inheritance to add that functionality. Since there are now different test case classes we want to use (e.g., `unittest.TestCase` and `unittest.IsolatedAsyncioTestCase`), that inheritance strategy means that we would have to create a subclass for each one. That's rather redundant and we're not dealing with a true subclass anyway. Instead, I've now created a mixin class that can be used together with a unittest test case class to add the functionality we desire to whatever test case class we want to use.

### Additional changes

- I've changed all our test classes that used `async_test` to use the new `IsolatedAsyncTestSuite`; see above for the rationale.

- I've fixed one hidden bug in `test_time` causing a test method to not actually test something.

- I've removed a pytest file that was left behind during the unittest migration

- I've upped the coverage of the test suite part of the bot to 100% by adding a few tests and `praga: no cover` parts that don't require testing.

This PR closes #758


tests/test_base.py
```diff
@@ -3,7 +3,11 @@
 import unittest.mock
 
 
-from tests.base import LoggingTestCase, _CaptureLogHandler
+from tests.base import LoggingTestsMixin, _CaptureLogHandler
+
+
+class LoggingTestCase(LoggingTestsMixin):
+    pass
```
MarkKoz: Why?

tests/helpers.py
```diff
@@ -337,14 +254,16 @@ class MockBot(CustomMockMixin, unittest.mock.MagicMock):
...
 
     def __init__(self, **kwargs) -> None:
-        super().__init__(spec_set=bot_instance, **kwargs)
+        super().__init__(**kwargs)
 
         # self.wait_for is *not* a coroutine function, but returns a coroutine nonetheless and
         # and should therefore be awaited. (The documentation calls it a coroutine as well, which
         # is technically incorrect, since it's a regular def.)
-        self.wait_for = AsyncMock()
+        # self.wait_for = unittest.mock.AsyncMock()
```
MarkKoz: Why not remove these comments entirely?

SebastiaanZ: Good point.

tests/test_base.py
SebastiaanZ: I've changed this into a more realistic use case with a class both implementing this mixin as well as a proper TestCase class. That's how it's going to be used when it's implemented as well.

tests/bot/cogs/sync/test_cog.py
```diff
@@ -90,7 +91,6 @@ def test_sync_cog_init(self, sync_guild):
         sync_guild.assert_called_once_with()
         self.bot.loop.create_task.assert_called_once_with(mock_sync_guild_coro)
```
MarkKoz: I wonder... is there still no better way to do this despite all the async support that 3.8 had added?

tests/bot/cogs/sync/test_base.py
```diff
@@ -349,7 +339,10 @@ def setUp(self):
                 if ctx is not None:
                     ctx.send.return_value = message
 
-                self.syncer._get_confirmation_result = helpers.AsyncMock(return_value=(False, None))
+                diff = _Diff({1, 2, 3}, {4, 5}, None)
+                self.syncer._get_diff.return_value = diff
```
MarkKoz: Why was this added?

SebastiaanZ: I should probably just set an explicit `MagicMock` as the return value instead of an actual `_Diff` object, but the "issue" here is that the `AsyncMock` you've used to mock `_get_diff` will return another `AsyncMock` as its child when its called and awaited. (The default behavior for mocks is to return a child mock of its own type.)

Because of that "child is like parent" behavior, this will happen in [lines 194-196 of bot.cogs.sync.syncers](https://github.com/python-discord/bot/blob/c81a4d401ea434e98b0a1ece51d3d10f1a3ad226/bot/cogs/sync/syncers.py#L194-L196):

```py
        # `diff` will be a child AsyncMock as "child is like parent"
        diff = await self._get_diff(guild)

        # `diff._asdict` will be an AsyncMock as "child is like parent" and,
        # after being called, it will return an unawaited coroutine object
        # we assign the name `diff_dict`:
        diff_dict = diff._asdict()

        # `diff_dict` is still an unawaited coroutine object meaning that it
        # doesn't have an `items()` method:
        totals = {k: len(v) for k, v in diff_dict.items() if v is not None}
```

This will lead to the following exception when running the tests:

```py
======================================================================
ERROR: test_sync_confirmation_context_redirect (tests.bot.cogs.sync.test_base.SyncerSyncTests) (ctx=None, author=<MockMember name='mock.user' spec_set='Member' id='140583452034864'>, message=None)
If ctx is given, a new message should be sent and author should be ctx's author.
----------------------------------------------------------------------
  File "/home/sebastiaan/pydis/repositories/bot/tests/bot/cogs/sync/test_base.py", line 348, in test_sync_confirmation_context_redirect
    await self.syncer.sync(guild, ctx)
  File "/home/sebastiaan/pydis/repositories/bot/bot/cogs/sync/syncers.py", line 196, in sync
    totals = {k: len(v) for k, v in diff_dict.items() if v is not None}
AttributeError: 'coroutine' object has no attribute 'items'
```

The solution is to assign an explicit return value so the parent mock doesn't "guess" and return an object of its own type. However, instead of assigning a specific `_Diff` object as the return value, I should probably have gone with a `MagicMock` to signify that it's not an important return value; it's just something that needs to support/mimic the API we use on it.

tests/bot/cogs/sync/test_cog.py
SebastiaanZ: I'm not quite sure. The `create_task` method is not a coroutine but a regular method that, if actually used, returns an awaitable `Task` object. The thing I've changed here is that instead of mocking the `sync_guild` with an `AsyncMock`, I've mocked it with a `MagicMock`; the reason is that we're not actually interested in the async interface, but merely want to check if `create_task` was called with the right thing.

The latter is more difficult with an AsyncMock as when we call an AsyncMock, we get coroutine object back (we're not awaiting it anywhere here) and each time we call it, we get a different coroutine object back. That means that if were to go the route of `AsyncMock`, the call within the "actual" method in the cog will return a different coroutine object than the one we can create in the test method and we can't assert that `create_task` is called with the right argument: The coroutine objects are different.

By using a MagicMock instead, we don't need to deal with this more complicated coroutine object behavior as MagicMock will just return the same return value each time it's called, making this an easy assertion.

SebastiaanZ:
> Is it out of scope to change assertions to use the new [await assertions](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock.assert_awaited)?

We could declare that within the scope of the PR, but it will delay it quite some while, I think. I haven't looked into it too deeply yet, but if we go that route, it basically means reevaluating each test method that uses coroutines to see if we can modernize the assert strategy.

Such a modernization would be a good (and probably sets the right examples to people writing new tests), but I could also argue that the upgraded test suite runs and provides the same level of "testing" as it did before the migration meaning that such a modernization would not be a pure migration but rather an enhancement now that new tools have become available.

I think we should carry out such a modernization, but I'm a bit on the fence on if it should be part of this PR. It will delay it by quite a bit (it's probably more work than the migration itself has been as all the test methods have to be actually looked at in detail) and make the review much more involved as well. We could also go with "the current migration provides the same level of testing as we had before so we'll do the modernization in a future PR".

What do you think?

tests/bot/cogs/sync/test_base.py
MarkKoz: Oh yeah I remember you mentioning that weird issue in VC. However, I ran it without that change locally and the test passed.

MarkKoz: Actually, that may be because I changed the syncer to use autospecs (which I believe is smarter about specing attributes than just using the same mock as the parent).

tests/bot/cogs/sync/test_cog.py
MarkKoz: Thanks for your input. I was never too happy with my solution for this but it was far cleaner than my first attempts.

tests/test_base.py
MarkKoz: OK now I realise what was going on here. The mixin relies on methods from `TestCase` but it doesn't actually subclass it. Hence the class above is needed so it can subclass `TestCase`.

tests/bot/cogs/sync/test_base.py
SebastiaanZ: That's for the `test_cog` file, right? The `test_base` file uses this:

```py
class TestSyncer(Syncer):
    """Syncer subclass with mocks for abstract methods for testing purposes."""

    name = "test"
    _get_diff = mock.AsyncMock()
    _sync = mock.AsyncMock()
```

If I comment out line 343 in `test_base`, I do get the mocking behavior above (and the exception/unawaited coroutine warnings).

MarkKoz:
> What do you think?

Let's get this merged and take care of the await assertions in another PR. The 3.8 upgrade hasn't caused a regression regarding assertions - they are at least as effective as they were before. The await assertions are merely an improvement so they don't block anything. I'd rather have this merged first in case anyone else needs to write tests in their PRs.

tests/bot/cogs/sync/test_base.py
MarkKoz: Yeah you're right about the autospecs. That change is actually irrelavant. Anyway, I ran the test again and it's failing now. I don't know why it appeared to pass last night.

MarkKoz:
Stalled by #755

SebastiaanZ:
This should no longer be stalled. The PR mentioned by Mark has been merged, which means we don't suffer from that changed behavior of the CancelledError anymore.

Issue #780, opened 2020-02-23
Validate Channel Constants on Bot Startup


sco1:
Related to https://github.com/python-discord/seasonalbot/issues/335, the bot's channel constants can occasionally get out of sync, resulting in `#deleted-channel` references in some of the error responses.

In order to alert staff that this desync has occurred, a utility function should be added to the bot's startup routine that validates the channel constants & logs a warning if any of the channels do not exist so the constants file can be updated appropriately.

Numerlor:
I'd like to give this a try if you weren't planning for the same assignee as on the seasonalbot issue.
For the logging, should it also send a message somewhere to notify or just log as a warning/error?

sco1:
Go for it. With Sentry now integrated, logging a warning is sufficient.

PR #785, opened 2020-02-24
Verify channels against server and reddit OAuth2 in constants on startup.

Numerlor:
closes: #780 
closes: #794

Adds new ConfigVerifier cog which now verifies if configured channels are present in server.

Checks if Reddit OAuth2 id and secret were set before attempting to load its cog.

1 unrelated commits included, https://github.com/python-discord/bot/pull/785/commits/6174792c01f238e32aca5cc9222caa4feb788281 for removing an unused function.

### DIFF
--- a/bot/cogs/config_verifier.py
+++ b/bot/cogs/config_verifier.py
@@ -1,0 +1,40 @@
+import logging
+
+from discord.ext.commands import Cog
+
+from bot import constants
+from bot.bot import Bot
+
+
+log = logging.getLogger(__name__)
+
+
+class ConfigVerifier(Cog):
+    """Verify config on startup."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.bot.loop.create_task(self.verify_channels())
+
+    async def verify_channels(self) -> None:
+        """
+        Verifies channels in config.
+
+        If any channels in config aren't present in server, log them in a warning.
+        """
+        await self.bot.wait_until_ready()
+        server = self.bot.get_guild(constants.Guild.id)
+
+        server_channel_ids = {channel.id for channel in server.channels}
+        invalid_channels = [
+            channel_name for channel_name, channel_id in constants.Channels
+            if channel_id not in server_channel_ids
+        ]
+
+        if invalid_channels:
+            log.warning(f"Channels do not exist in server: {', '.join(invalid_channels)}.")
+
+
+def setup(bot: Bot) -> None:
+    """Load the ConfigVerifier cog."""
+    bot.add_cog(ConfigVerifier(bot))
--- a/bot/cogs/config_verifier.py
+++ b/bot/cogs/config_verifier.py
@@ -18,7 +18,7 @@ class ConfigVerifier(Cog):
 
     async def verify_channels(self) -> None:
         """
-        Verifies channels in config.
+        Verify channels.
 
         If any channels in config aren't present in server, log them in a warning.
         """
--- a/bot/cogs/config_verifier.py
+++ b/bot/cogs/config_verifier.py
@@ -32,7 +32,7 @@ class ConfigVerifier(Cog):
         ]
 
         if invalid_channels:
-            log.warning(f"Channels do not exist in server: {', '.join(invalid_channels)}.")
+            log.warning(f"Configured channels do not exist in server: {', '.join(invalid_channels)}.")
 
 
 def setup(bot: Bot) -> None:
--- a/bot/cogs/config_verifier.py
+++ b/bot/cogs/config_verifier.py
@@ -22,7 +22,7 @@ class ConfigVerifier(Cog):
 
         If any channels in config aren't present in server, log them in a warning.
         """
-        await self.bot.wait_until_ready()
+        await self.bot.wait_until_guild_available()
         server = self.bot.get_guild(constants.Guild.id)
 
         server_channel_ids = {channel.id for channel in server.channels}
--- a/bot/cogs/config_verifier.py
+++ b/bot/cogs/config_verifier.py
@@ -14,7 +14,7 @@ class ConfigVerifier(Cog):
 
     def __init__(self, bot: Bot):
         self.bot = bot
-        self.bot.loop.create_task(self.verify_channels())
+        self.channel_verify_task = self.bot.loop.create_task(self.verify_channels())
 
     async def verify_channels(self) -> None:
         """
--- a/bot/cogs/reddit.py
+++ b/bot/cogs/reddit.py
@@ -290,8 +290,7 @@ class Reddit(Cog):
 
 def setup(bot: Bot) -> None:
     """Load the Reddit cog."""
-    invalid_values = "", None
-    if any(value in (RedditConfig.secret, RedditConfig.client_id) for value in invalid_values):
+    if not RedditConfig.secret or not RedditConfig.client_id:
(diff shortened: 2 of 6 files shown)


bot/cogs/config_verifier.py
```diff
@@ -0,0 +1,40 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class ConfigVerifier(Cog):
+    """Verify config on startup."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.bot.loop.create_task(self.verify_channels())
```
SebastiaanZ: While we used to do this (and probably had to us this prior to Python 3.7), I think we should switch to using the higher level asyncio API with `asyncio.create_task` instead of using the lower-level loop API of `loop.create_task`. I think this is as good a point as any to introduce a new habit.

I think it's also a good idea to keep a reference to the task around. That will make it easier to debug since we can inspect exceptions and so on if something breaks.

So, would suggest, **with an additional `import asyncio` at the top**, this:

```suggestion
        self.verification_task = asyncio.create_task(self.verify_channels())
```

If we merge this after the Python 3.8 migration, we can also add an explicit `name` kwarg to give the task a beautiful repr.

bot/cogs/config_verifier.py
```diff
@@ -0,0 +1,40 @@
...
+
+    async def verify_channels(self) -> None:
+        """
+        Verify channels.
+
+        If any channels in config aren't present in server, log them in a warning.
+        """
+        await self.bot.wait_until_ready()
```
SebastiaanZ: We've had issues with the guild cache not being available when the READY signal was dispatched (discord.py only waits for 2 seconds which, during Discord API lag, is occasionally too short). To prevent that, we've implemented a custom [`wait_until_guild_available`](https://github.com/python-discord/bot/blob/c81a4d401ea434e98b0a1ece51d3d10f1a3ad226/bot/bot.py#L90) that ensures that the cache is actually loaded before doing things that rely on the cache being available. 

Since this validator relies entirely on the cache, it's a good idea to use that `wait_for` variant instead:

```suggestion
        await self.bot.wait_until_guild_available()
```

bot/cogs/config_verifier.py
Numerlor: This fails as the bot's event loop is not running yet and it can't fetch it if I'm understanding it right. Should it at least be assigned to a var?

SebastiaanZ: Right, that is true, I guess.

bot/cogs/reddit.py
```diff
@@ -290,4 +290,7 @@ def channel(self) -> TextChannel:
 
 def setup(bot: Bot) -> None:
     """Load the Reddit cog."""
-    bot.add_cog(Reddit(bot))
+    if None not in (RedditConfig.client_id, RedditConfig.secret):
+        bot.add_cog(Reddit(bot))
+        return
+    log.error("Credentials not provided, cog not loaded.")
```
kwzrd: Just a matter of preference, but I think these would read nicer as:
```python
if None in (...):
  log.error(...
  return
bot.add_cog(...
```
Feel free to keep as is though.

bot/constants.py
```diff
@@ -186,6 +186,10 @@ def __getattr__(cls, name):
     def __getitem__(cls, name):
         return cls.__getattr__(name)
 
+    def __iter__(cls):
+        """Returns iterator of key: value pairs of current constants class."""
+        return iter(_CONFIG_YAML[cls.section][cls.subsection].items())
```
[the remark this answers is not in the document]
Numerlor: Guess it depends on the behavior we want from it on sections, list only its config values or the subsections too?

bot/cogs/reddit.py
Numerlor: I stylized it with the intended case as the first one (although the eyes fall more onto the log...), maybe a third opinion?

bot/constants.py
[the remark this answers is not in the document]
kwzrd: Subsections have their own config classes. I think this should only feed values annotated in the class that is being iterated.

bot/cogs/reddit.py
sco1: I prefer @kwzrd's

bot/cogs/reddit.py
```diff
@@ -290,4 +290,7 @@ def channel(self) -> TextChannel:
 
 def setup(bot: Bot) -> None:
     """Load the Reddit cog."""
+    if None in (RedditConfig.client_id, RedditConfig.secret):
```
kwzrd: I missed this: if unconfigured, these will be empty strings, not `None`.

Numerlor: It's `None` when excluded from the .env, empty strings if just the values are unset.
Empty string get caught when the cog tries to authenticate and is then unloaded.

kwzrd: Under docker, they default to empty strings when excluded from `.env`. I think it would make sense to check for both in the setup.

kwzrd: Sorry, I realize now that the issue this addresses is that the cog crashes if these two sneak in as `None`s. I assume that this is then a problem that only happens when running without docker.

I still think that it'd make sense to unload & log if they are empty strings, but I'll re-approve now. If you don't want to do that, then you can mark this as resolved.

Numerlor: At a minimum it'll save 3 requests which are done before the cog is unloaded so won't hurt.

Numerlor:
When going over the last commit I noticed this: https://github.com/python-discord/bot/blob/1ccc0f11fd4511d13c7cdf59070dca86465fe59f/bot/cogs/reddit.py#L43-L47
Is the comparison operator supposed to be checking the other way around?
Tokens auto expire and become unusable in an hour, and now the check only executes if the token's already expired (expired time before current) if I'm getting it right

bot/cogs/reddit.py
```diff
@@ -290,4 +290,8 @@ def channel(self) -> TextChannel:
 
 def setup(bot: Bot) -> None:
     """Load the Reddit cog."""
+    invalid_values = "", None
+    if any(value in (RedditConfig.secret, RedditConfig.client_id) for value in invalid_values):
```
ks129:
```suggestion
    if not RedditConfig.secret or not RedditConfig.client_id:
```
This do same and is more readable.

bot/cogs/reddit.py
```diff
@@ -290,4 +290,8 @@ def channel(self) -> TextChannel:
 
 def setup(bot: Bot) -> None:
     """Load the Reddit cog."""
+    invalid_values = "", None
```
ks129:
```suggestion

```
No need for this after change in L294.

MarkKoz:
@Numerlor Yes, I believe the operator in the datetime comparison should be flipped around i.e. revoke the token if it's still valid. Also realised there needs to be a None check for the access token as it's not guaranteed to have been created before the cog unloads.

Numerlor:
> @Numerlor Yes, I believe the operator in the datetime comparison should be flipped around i.e. revoke the token if it's still valid. Also realised there needs to be a None check for the access token as it's not guaranteed to have been created before the cog unloads.

Pushing a fix through, just made sure it's not intended. Already included a bool if check on the `access_token` in https://github.com/python-discord/bot/pull/785/commits/e8e2fa9ee8f607bb6593b7c8325446dc074a972d

PR #786, opened 2020-02-24
Implement a new help channel system

MarkKoz:
Follows the design outlined in https://github.com/python-discord/organisation/issues/200#issuecomment-590000882

# To do

- [ ] Figure out an implementation for a cooldown for claiming a new available channel

# Outline

## Constants

- [x] Chemical element names
- Messages
    - [x] Available
    - [x] Dormant
- `PermissionOverwrite`
    _Can re-use the same object every time a channel is moved_
    - [ ] Available
    - [ ] Dormant

### Config

- Categories
    - [x] Available
    - [x] In Use
    - [x] Dormant
- [x] Role whitelist for `dormant_command`
- [x] Allowed duration of inactivity before making a channel dormant
- [x] Maximum number of channels to put in the available category

## Functions

- [ ] `__init__`
    _Perform initialisation of the help channel system_
    1. Call `create_channel_queue()` to get the queue of available channels
    2. Assign to `channel_queue`
    3. Call `create_name_queue()` to get the queue of element names
    4. Assign to `name_queue`
    5. Call `init_available()` to create missing available channels
    6. Call `move_idle_channels()` to make channels which were inactive for too long dormant
- [ ] `_scheduled_task(TextChannel, int) -> None`
    _Callback used by the scheduler_
    1. Wait given duration or the duration of the configured idle timeout
    2. Call `get_idle_time()` to get time elapsed since the last message
    3. Compare the elapsed time against the configured idle timeout
        * If greater (i.e. idle):
            1. Call `move_to_dormant()` to make the channel dormant
        * If less:
            1. Subtract the elapsed time from the configured idle timout
            2. Reschedule the task with the computed duration
- [ ] `create_channel_queue() -> asyncio.Queue`
    _Initialise and return the queue of dormant channels which will be used for getting the next available channel_
    1. Get all channels in the _Dormant_ category
    2. Sort them randomly
        * This is simpler and faster than going through the history of all of them and sorting by last messages' timestamps
    3. Place them in a new `asyncio.Queue`
    4. Return the queue
- [ ] `create_dormant() -> TextChannel`
    _Create a new dormant channel_
    1. Pop a name from `name_queue`
    2. Return `None` if the queue was empty
    3. Use `guild.create_text_channel()` to create the channel
        * Specify `overwrites` and `category` kwargs
- [ ] `create_name_queue() -> collections.deqeue`
    _Initialise and return the queue of element names which will be used for creating new dormant channels_
    1. Get the channels of all 3 help categories
    2. Remove them from the list read from JSON
    3. Place the remaining names in a new `collections.deqeue`
    4. Return the queue
- [ ] `dormant_command() -> None`
    _Command to manually make the current channel dormant_
    1. Use role whitelist config to permit invocation
    2. Ensure it's called inside an _In Use_ channel
    3. Get the current channel
    4. Call `make_dormant()`
- [ ] `get_available_candidate() -> TextChannel`
    _Return a dormant channel to turn into an available channel_
    1. Pop a channel using `channel_queue.get_nowait()`
    2. If the queue was empty, use `create_dormant()` to create a new channel
    3. If `None` was returned (i.e. no names available):
        1. Send a warning to staff that there's a shortage of channels (so they may manually free some up)
        2. Wait for a channel to become available with `await channel_queue.get()`
    4. Return the channel
- [ ] `get_idle_time(TextChannel) -> int`
    _Return the time elapsed since the last message sent in the channel._
    1. Steal code from the `!free` command to compute this
- [ ] `init_available() -> None`
    _Initialise the Available category with channels_
    1. Determine current amount of available channels
    2. Determine how many channels to add by subtracting current available channels from the configured maximum
    3. Add the above amount of new available channels with `move_to_available()`
- [ ] `move_idle_channels() -> None`
    _Make all idle in-use channels dormant_
    1. Iterate all in-use channels
    2. Call `get_idle_time()` to get time elapsed since the last message
    3. Compare the above time against the configured idle timeout
    4. If greater, make the channel dormant with `move_to_dormant()`
- [ ] `move_to_available() -> None`
    _Make a channel available_
    1. Get a dormant channel with `get_available_candidate()`
    2. Send available message
        * Edit or delete the dormant message
    3. Move channel to _Available_ category
    4. Add send message permissions (channel overrides)
- [ ] `move_to_dormant(TextChannel) -> None`
    _Make a channel dormant_
    1. Remove send message permissions (channel overrides)
    2. Send dormant message
    3. Move channel to _Dormant_ category
    4. Push the channel into `channel_queue`
- [ ] `on_message(Message) -> None`
    _Move an available channel to the In Use category and replace it with a dormant channel_
    1. Ignore messages outside the _Available_ category
    2. Move the channel to the _In Use_ category
    3. Add a new available channel with `move_to_available()`
    4. Schedule the channel to be moved to the _Dormant_ category


bot/cogs/help_channels.py
```diff
@@ -0,0 +1,665 @@
...
+        except KeyError:
+            log.warning(f"Channel #{channel} ({channel.id}) doesn't have a valid name.")
+            position = None
+
+        log.trace(
+            f"Position of #{channel} ({channel.id}) in Dormant will be {position} "
+            f"(was {channel.position})."
+        )
+
+        return position
```
SebastiaanZ: As we've discovered, this approach does not work. After doing a bit of testing, I found out it's because Discord channels the position integers of channels that are already in the category: If you add a channel to a category, the position of all channels with a higher (or equal?) position integer than the one you're inserting will be increased by one:

```
Moving channel h to position 8
Before move: [('b', 2), ('e', 5), ('f', 7),           ('i', 10), ('g', 11), ('j', 13), ('m', 19)]
After move:  [('b', 2), ('e', 5), ('f', 7), ('h', 8), ('i', 11), ('g', 12), ('j', 14), ('m', 20)]
```

As you can see, all the channels with a position `int` higher than 8 had their position integer increased by one. I can't guarantee that this will always happen, but after quite a few experiments, I've never observed an instance that violated this pattern.

My first approach to circumvent this was by assigning each channel a very large starting int (multiples of 200) to give it a buffer against going over the position int of the next channel. However, since Discord minimizes the integers down to the current largest used integer plus one, this doesn't work at all.

Another approach does seem to work: If you insert a channel at the position int of a channel that's already there, it seems to be that it will reliably be inserted before the existing channel. This means that if you have a channel called "c" with position integer 2 and you want to insert "b", you can insert "b" at 2 and the sorting order appears to end up just like we want it. Again, I can only go by experimentation, but it does seem to work.

I used something like this to do the work:
```py
import bisect

def get_position(channel: discord.TextChannel, destination: discord.CategoryChannel) -> int:
    # If the destination category is empty, return the first position
    if  not destination.channels:
        return 1

    # Make list of channel names for bisect
    channel_names = [c.name for c in destination.channels]

    # Get the bisect location of the channel
    rank = bisect.bisect(channel_names, channel.name)

    # If rank is equal to the length, we should insert the channel at the end, otherwise
    # we insert it at the location of the channel it should be inserted before.
    if rank == len(destination.channels):
        return destination.channels[-1].position + 1
    else:
        return destination.channels[rank].position
```

If a channel ever falls out of order, this procedure won't correct that, although it will get back into order once the "out-of-order" channels get cycled through the available/active categories.

Again, it may be that this system is not entirely robust either, I can only go by experimentation, but it seems to work as expected in the tests I've done.

MarkKoz: I really appreciate you looking into this. I'll try to implement your code into the cog.

MarkKoz: It works! But what would happen if the `channel_names` list is unsorted or partially sorted? discord.py sorts the category's channels by position rather than by name. Would it be better to sort them by name ourselves?

MarkKoz:
Having trouble ignoring mod log events since moving a channel and changing its topic seem to dispatch separate channel update events.

bot/cogs/help_channels.py
```diff
@@ -0,0 +1,672 @@
...
+        * Prioritise using the channels which have been dormant for the longest amount of time
+        * If there are no more dormant channels, the bot will automatically create a new one
+        * Configurable with `constants.HelpChannels.max_available`
+        * If there are no dormant channels to move, helpers will be notified (see `notify()`)
+    * When a channel becomes available, the dormant embed will be edited to show `AVAILABLE_MSG`
+
+    In Use Category
+
+    * Contains all channels which are occupied by someone needing help
+    * Channel moves to dormant category after 45 minutes of being idle
```
lemonsaurus: We shouldn't put the number 45 here, since this is a configurable parameter that may change - and in fact, it's not even set to 45 in the settings in this PR.

bot/cogs/help_channels.py
```diff
@@ -0,0 +1,672 @@
...
+        with Path("bot/resources/elements.json").open(encoding="utf-8") as elements_file:
+            all_names = json.load(elements_file)
+
+        if prefix:
+            return [prefix + name for name in all_names[:count]]
+        else:
+            return all_names[:count]
+
+    def get_used_names(self) -> t.Set[str]:
+        """Return channels names which are already being used."""
```
lemonsaurus: Intentional typo?
```suggestion
        """Return channel names which are already being used."""
```

bot/cogs/help_channels.py
```diff
@@ -0,0 +1,672 @@
...
+        self.name_queue = self.create_name_queue()
+
+        log.trace("Moving or rescheduling in-use channels.")
+        for channel in self.get_category_channels(self.in_use_category):
+            await self.move_idle_channel(channel, has_task=False)
+
+        # Prevent the command from being used until ready.
+        # The ready event wasn't used because channels could change categories between the time
+        # the command is invoked and the cog is ready (e.g. if move_idle_channel wasn't called yet).
+        # This may confused users. So would potentially long delays for the cog to become ready.
```
lemonsaurus: unintentional typo?
```suggestion
        # This may confuse users. So would potentially long delays for the cog to become ready.
```

bot/cogs/help_channels.py
```diff
@@ -0,0 +1,672 @@
...
+
+        try:
+            log.trace("Sending notification message.")
+
+            channel = self.bot.get_channel(constants.HelpChannels.notify_channel)
+            mentions = " ".join(f"<@&{role}>" for role in constants.HelpChannels.notify_roles)
+
+            message = await channel.send(
+                f"{mentions} A new available help channel is needed but there "
+                f"are no more dormant ones. Consider freeing up some in-use channels manually by "
```
lemonsaurus: this doesn't really have to be an f-string, but on the other hand, the f makes it line up neatly with the others so I'm not actually sure which I prefer. Maybe we just leave it.

MarkKoz: I prefer to leave it.

bot/cogs/help_channels.py
MarkKoz: The typo I mentioned in the meeting was fixed in 00c9f092a0c056b21e012fdadd807d0410a3ca09 😄

bot/cogs/help_channels.py
MarkKoz: I've removed references to default values and replaced them with references to the config variable names.

bot/cogs/help_channels.py
SebastiaanZ: If the `channel_names` list is partially sorted and/or entirely out of order, we will slowly start building an order and it will get back into order at the latest when every channel has been cycled once. I'm not quite sure what sorting the list manually would do, except that it will guarantee that we then insert a channel before the one it's supposed to come for.

That could be preferable, but it won't be a magic "sorting fix". We could also try to detect if the thing is out of order and, if so, try to fix it with the minimum number of channel edits as is possible, but I'd have to have to sit down and thing of a clever way of doing that. There's probably an algorithm for that.

MarkKoz: I'm fine with it as it is since it will eventually fix itself. It should only get out of order if someone accidentally moves a channel (or if the implementation turns out to not always work). An separate issue can be opened to tackle this in a more clever way.

bot/cogs/help_channels.py
```diff
@@ -0,0 +1,671 @@
...
+question into it. Once claimed, the channel will move into the **Help: In Use** category, and will \
+be yours until it has been inactive for {constants.HelpChannels.idle_minutes} minutes. When that \
+happens, it will be set to **dormant** and moved into the **Help: Dormant** category.
+
+You may claim a new channel once every {constants.HelpChannels.claim_minutes} minutes. If you \
+currently cannot send a message in this channel, it means you are on cooldown and need to wait.
+
+Try to write the best question you can by providing a detailed description and telling us what \
+you've tried already. For more information on asking a good question, \
+[check out our guide on asking good questions]({ASKING_GUIDE_URL}).
```
aeros, line 50: Typically, embedded links limit the text to just noun, and not the entire subject. I.E. "guide on asking good questions" or just "asking good questions" instead of "check out our guide on asking good questions".

bot/utils/scheduling.py
```diff
@@ -51,20 +51,32 @@ def schedule_task(self, task_id: t.Hashable, task_data: t.Any) -> None:
...
         del self._scheduled_tasks[task_id]
+        task.cancel()
 
         log.debug(f"{self.cog_name}: unscheduled task #{task_id} {id(task)}.")
 
+    def cancel_all(self) -> None:
+        """Unschedule all known tasks."""
+        log.debug(f"{self.cog_name}: unscheduling all tasks")
+
+        for task_id in self._scheduled_tasks.copy():
```
aeros, line 77: What is the reason for this iterating through a copy of `_scheduled_tasks`?

bot/utils/scheduling.py
```diff
@@ -51,20 +51,32 @@ def schedule_task(self, task_id: t.Hashable, task_data: t.Any) -> None:
...
+            self.cancel_task(task_id, ignore_missing=True)
+
```
aeros, line 79: After cancelling the tasks, you may also want to consider propagating an exception for any task that couldn't be cancelled; it can help to make subtle errors more apparent. We use this in the implementation for `asyncio.run()`. For an example of how this could be done, see https://github.com/python/cpython/blob/909f4a30093f74d409711e564f93a43167ca0919/Lib/asyncio/runners.py#L62-L73.

bot/cogs/help_channels.py
SebastiaanZ: I don't mind either way, although I think the noun-only version looks slightly better.

SebastiaanZ: As per @aeros 's suggestion, it would look like this:
```suggestion
check out our guide on [asking good questions]({ASKING_GUIDE_URL}).
```

bot/cogs/help_channels.py
```diff
@@ -0,0 +1,671 @@
...
+
+DORMANT_MSG = f"""
+This help channel has been marked as **dormant**, and has been moved into the **Help: Dormant** \
+category at the bottom of the channel list. It is no longer possible to send messages in this \
+channel until it becomes available again.
+
+If your question wasn't answered yet, you can claim a new help channel from the \
+**Help: Available** category by simply asking your question again. Consider rephrasing the \
+question to maximize your chance of getting a good answer. If you're not sure how, have a look \
+through [our guide for asking a good question]({ASKING_GUIDE_URL}).
```
SebastiaanZ, line 61: Suggestion as per @aeros 's comment above:
```suggestion
through our guide for [asking a good question]({ASKING_GUIDE_URL}).
```

bot/utils/scheduling.py
MarkKoz: @aeros All tasks have a [done callback](https://github.com/python-discord/bot/blob/master/bot/utils/scheduling.py#L80) which retrieves exceptions from tasks. When a task is cancelled, it is also considered done. Did I cover all bases here or am I still missing some scenario?

MarkKoz: Oh I think what you're getting at is when `task.cancel()` is called but `task.cancelled()` still returns False, meaning it wasn't actually cancelled.

aeros: @MarkKoz 
> Oh I think what you're getting at is when task.cancel() is called but task.cancelled() still returns False, meaning it wasn't actually cancelled

Yep, that's what I was referring to. It's specifically when task.cancel() is called, `task.cancelled()` returns false, and `task.exception()` contains an exception of some form. Your `_task_done_callback()` does cover this for the most part (I hadn't looked over it), but indefinitely running tasks may still be a concern with that implementation.

> So what if I see it's not cancelled but really all it needed was a little bit more time?

For this situation, you could consider using a timeout at the end with `asyncio.wait_for()`, to give your tasks N seconds at most to cancel. This would prevent any hanging tasks from indefinitely locking up the shutdown process, if that might be an issue. In your situation, you'd likely want to use it where you use `cancel_all()` in `cog_unload()`. For example:

```py
    def cog_unload(self) -> None:
        """Cancel the init task and scheduled tasks when the cog unloads."""
        log.trace("Cog unload: cancelling the init_cog task")
        self.init_task.cancel()

        log.trace("Cog unload: cancelling the channel queue tasks")
        for task in self.queue_tasks:
            task.cancel()
        
        # You'll likely have to determine an optimal timeout, but 300 secs (10 mins)
        # could be a reasonable default.
        try:
            await asyncio.wait_for(self.cancel_all(), timeout=300)
        except asyncio.TimeoutError:
            log.error("could not cancel all tasks: timeout occurred")
```

MarkKoz: `cancel_task()` essentially just calls `task.cancel()` for each task. I thought `task.cancel()` didn't block, so how would a timeout help?

MarkKoz: Oh and I forgot that a task is also considered done when an exception is raised inside it, so presumably that means the done callback would also trigger if an exception was raised when there was an attempt to cancel a task? In that case, yeah, I do believe you're right that the callback is covering it.

aeros: > I thought task.cancel() didn't block, so how would a timeout help?

I was just about to correct the above, sorry: you'd have to modify `cancel_all()` slightly (or add it in `cog_unload`) to wait for `task.done()` to be true for each task that still isn't cancelled after attempting to call `task.cancel()`.

aeros: But the above is only needed if you think indefinitely running tasks might be an issue (which you'll likely have a better idea than I do, based on what the tasks are doing). It might not be a realistic concern in your case.

MarkKoz: No, not indefinite, but there can be long sleeps (30 minutes). In any case, I think it's safe to leave it alone for now but I will keep this in mind if it turns out to be a problem. Thanks.

aeros: > Oh and I forgot that a task is also considered done when an exception is raised inside it, so presumably that means the done callback would also trigger if an exception was raised when there was an attempt to cancel a task?

Yep, it's only an issue when the exception is in the process of propagating/being handled, but hasn't been cancelled yet. If it finished raising, the task is considered done, so it should trigger the done callback when the exception was raised.

As mentioned above, I'm not certain the above will be a realistic concern in your case. As the one who wrote it, you'll likely have a better understanding of whether or not `_task_done_callback()` will be adequate based on what the tasks are actually doing. If indefinitely hanging tasks are an issue later down the road, I'd recommend revisiting it though and implementing some form of timeout.

PR #790, opened 2020-02-25
Added ignoring staff to Antimalware check.

### LINKED ISSUE #767: Filters should not apply to staff members
Currently, we have certain channels whitelisted for certain filters and some filters apply to all members, even Owners cannot post filetypes that aren't on the whitelist!

Please change this so that absolutely all filters will ignore all staff members. It is not necessary to whitelist staff channels or to keep any other kind of exceptions to the filters once this very simple exception has been added, so please clean up any such exceptions while you're at it.

ks129:
As all other checks already have staff ignore, so I added here too. Closes #767 .

### DIFF
--- a/bot/cogs/antimalware.py
+++ b/bot/cogs/antimalware.py
@@ -23,6 +23,7 @@ class AntiMalware(Cog):
             return
 
         # Check if user is staff, if is, return
+        # Since we only care that roles exist to iterate over, check for the attr rather than a User/Member instance
         if hasattr(message.author, "roles") and any(role.id in STAFF_ROLES for role in message.author.roles):
             return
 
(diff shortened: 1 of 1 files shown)


bot/cogs/antimalware.py
```diff
@@ -18,6 +18,10 @@ def __init__(self, bot: Bot):
     @Cog.listener()
     async def on_message(self, message: Message) -> None:
         """Identify messages with prohibited attachments."""
+        # Check if user is staff, if is, return
+        if type(message.author) is Member and any(role.id in STAFF_ROLES for role in message.author.roles):
```
MarkKoz: It's more efficient if the roles are checked after the check for existence of attachments (line 25). Furthermore, at the least, use `isinstance()` rather than `type() is ...`. However, I think checking if the attribute exists is even better:

```suggestion
        if hasattr(message.author, "roles") and any(role.id in STAFF_ROLES for role in message.author.roles):
```

That being said, this should probably ignore DMs (`if not message.guild`). If that happens, then it may be safe to assume the author is always a `Member` but I am not certain of that. It doesn't hurt to be cautious and explicitly check they have roles.

ks129:
@MarkKoz It's now fixed

PR #797, opened 2020-02-26
Fix invalid tag names being passed to the tag cog.

### LINKED ISSUE #787: ZeroDivisionError in tag fuzzy search
Don't know exactly what triggered it cause I noticed it way up in my bot's log history.

```
Ignoring exception in on_command_error
  File "/home/mark/repos/python/bot-pydis/.venv/lib/python3.7/site-packages/discord/client.py", line 312, in _run_event
    await coro(*args, **kwargs)
  File "/home/mark/repos/python/bot-pydis/bot/cogs/error_handler.py", line 92, in on_command_error
    await ctx.invoke(tags_get_command, tag_name=ctx.invoked_with)
  File "/home/mark/repos/python/bot-pydis/.venv/lib/python3.7/site-packages/discord/ext/commands/context.py", line 132, in invoke
  File "/home/mark/repos/python/bot-pydis/bot/cogs/tags.py", line 128, in get_command
    founds = await self._get_tag(tag_name)
  File "/home/mark/repos/python/bot-pydis/bot/cogs/tags.py", line 87, in _get_tag
    return self._get_suggestions(tag_name)
  File "/home/mark/repos/python/bot-pydis/bot/cogs/tags.py", line 66, in _get_suggestions
    for tag_title, tag in self._cache.items()
  File "/home/mark/repos/python/bot-pydis/bot/cogs/tags.py", line 66, in <dictcomp>
    for tag_title, tag in self._cache.items()
  File "/home/mark/repos/python/bot-pydis/bot/cogs/tags.py", line 60, in _fuzzy_search
    return current / len(_search) * 100
ZeroDivisionError: division by zero
```

CC @ikuyarihS

Numerlor:
The error handler fallback allowed any tag names through and tags not containing any letters were not restricted by the `TagNameConverter`, resulting in a `ZeroDivisionError` when fuzzy matched and passed through `REGEX_NON_ALPHABET`

Fixes #787

### DIFF
--- a/bot/cogs/error_handler.py
+++ b/bot/cogs/error_handler.py
@@ -98,10 +98,7 @@ class ErrorHandler(Cog):
                     )
                 else:
                     with contextlib.suppress(ResponseCodeError):
-                        await ctx.invoke(
-                            tags_get_command,
-                            tag_name=tag_name
-                        )
+                        await ctx.invoke(tags_get_command, tag_name=tag_name)
                 return
         elif isinstance(e, BadArgument):
             await ctx.send(f"Bad argument: {e}\n")
--- a/bot/cogs/error_handler.py
+++ b/bot/cogs/error_handler.py
@@ -89,12 +89,20 @@ class ErrorHandler(Cog):
                     return
 
                 # Return to not raise the exception
-                with contextlib.suppress(BadArgument, ResponseCodeError):
-                    await ctx.invoke(
-                        tags_get_command,
-                        tag_name=await TagNameConverter.convert(ctx, ctx.invoked_with)
+                try:
+                    tag_name = await TagNameConverter.convert(ctx, ctx.invoked_with)
+                except BadArgument:
+                    log.debug(
+                        f"{ctx.author} tried to use an invalid command "
+                        f"and the fallback tag failed validation in TagNameConverter."
                     )
-                    return
+                else:
+                    with contextlib.suppress(ResponseCodeError):
+                        await ctx.invoke(
+                            tags_get_command,
+                            tag_name=tag_name
+                        )
+                return
         elif isinstance(e, BadArgument):
             await ctx.send(f"Bad argument: {e}\n")
             await ctx.invoke(*help_command)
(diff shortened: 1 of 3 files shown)


bot/converters.py
```diff
@@ -163,18 +155,18 @@ def is_number(value: str) -> bool:
...
         # The tag name is longer than 127 characters.
         elif len(tag_name) > 127:
             log.warning(f"{ctx.author} tried to request a tag name with over 127 characters. "
                         "Rejecting the request.")
             raise BadArgument("Are you insane? That's way too long!")
 
+        # The tag name is ascii but does not contain any letters.
+        elif not any(character.isalpha() for character in tag_name):
+            log.warning(f"{ctx.author} tried to request a tag name without letters. "
+                        "Rejecting the request.")
```
MarkKoz: Especially now that we use Sentry, these shouldn't warnings. In fact, with #757, there'll be no need to log this as the exception will already get logged in the command error handler. These log messages can be removed from this converter.

bot/cogs/error_handler.py
```diff
@@ -88,8 +89,11 @@ def __init__(self, bot: Bot):
                     return
 
                 # Return to not raise the exception
-                with contextlib.suppress(ResponseCodeError):
-                    await ctx.invoke(tags_get_command, tag_name=ctx.invoked_with)
+                with contextlib.suppress(BadArgument, ResponseCodeError):
+                    await ctx.invoke(
+                        tags_get_command,
+                        tag_name=await TagNameConverter.convert(ctx, ctx.invoked_with)
+                    )
```
MarkKoz: As a follow up to my other comment, the `BadArgument` should be caught and logged since the converter won't log it any more. Forwarding the error to the command handler is not a good idea since it handles the error undesirably (sends messages to the channel).

Numerlor: How should this be logged, through a `logging` warning like in the converter? (Or a lower level since it's just an invalid command at that point)

MarkKoz: Use the debug level.

PR #801, opened 2020-02-28
Bug/info/fuzzy search/wrong annotation

### LINKED ISSUE #789: Wrong annotation for Tags._fuzzy_search
## Description
The current declaration of _fuzzy_search is:
```py
@staticmethod
def _fuzzy_search(search: str, target: str) -> int
```

while the return value is a float.
```py
return current / len(_search) * 100
```

## Available fixes

We can change the return value from **int** to **float** or convert the division mark to floor the result:

 ```py
 def _fuzzy_search(search: str, target: str) -> float
```
 
**or**

```py
return current // len(_search) * 100
```

If you work on this, consider working on #787

F4zii:
## Description
Wrong annotation for Tags._fuzzy_search

The returned value from _fuzzy_search is a float
The current type annotation is an int,
solved this problem by switching it into a float

Resolves #789

If this is merged make sure to pull the changes from master since this branch can be outdated


F4zii:
I will recreate this pr from another branch

PR #807, opened 2020-03-01
Added Pagination to !roles command

### LINKED ISSUE #791: Truncation of !roles Command Embed
Sentry Issue: [BOT-1Q](https://sentry.io/organizations/python-discord/issues/1531925673/?referrer=github_integration)

The `!roles` information command currently sticks role ID's & a corresponding mention into the `description` field of the embed:

https://github.com/python-discord/bot/blob/1ccc0f11fd4511d13c7cdf59070dca86465fe59f/bot/cogs/information.py#L31-L46

However, the `description` field has a character limit of `2048` characters and we now have more roles than can fit into this space so the command will raise an exception. We currently have `49` roles, the command generates around `45` characters per role, so math makes that `2205` characters.

Since this information is probably all useful to have, an alternative display mechanism will need to be utilized. Possible approaches include: pagination, filtering of roles to display, adjusting the information provided, etc.

ks129:
Also made some fixes with variable names, comments and moved `Total roles` from footer to title. Closes #791


bot/cogs/information.py
```diff
@@ -32,20 +33,18 @@ def __init__(self, bot: Bot):
...
         for role in roles:
-            role_string += f"`{role.id}` - {role.mention}\n"
+            role_list.append(f"`{role.id}` - {role.mention}")
 
         # Build an embed
         embed = Embed(
-            title="Role information",
-            colour=Colour.blurple(),
-            description=role_string
+            title=f"Role information (Total {len(roles)} roles)",
```
eivl: Now it will never happen that we have 0 or 1 role. and adding a role/roles pluralization check will just mess up the unit test. but if it is ever needed you could do something like

```python
f"{'s'*0 if True else 's'}"
```
Where True here is the result of a check for len(roles) being 1 or lower.

I think this PR is fine without this addition, this comment is just meant as a reference.

bot/cogs/information.py
```diff
@@ -32,20 +33,18 @@ def __init__(self, bot: Bot):
...
+            colour=Colour.blurple()
         )
-        embed.set_footer(text=f"Total roles: {len(roles)}")
 
-        await ctx.send(embed=embed)
+        await LinePaginator.paginate(role_list, ctx, embed)
```
sco1:
```suggestion
        await LinePaginator.paginate(role_list, ctx, embed, empty=False)
```
Setting the `empty` kwarg to `False` will remove the newline in between every role.

Before:
image

After:
image

bot/cogs/information.py
ikuyarihS: You can further shorten this by doing
```py
's' * (number_of_roles > 1)
```
since `'s' * 0` is `''`

PR #808, opened 2020-03-01
Optimize Dockerfile for faster rebuilds

spyoungtech:
This change optimizes the docker build so that a cached layer for installing dependencies can be used, instead of installing dependencies every build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,12 +9,15 @@ ENV PIP_NO_CACHE_DIR=false \
 # Install pipenv
 RUN pip install -U pipenv
 
-# Copy project files into working directory
+# Create the working directory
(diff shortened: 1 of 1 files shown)


Akarys42:
And what if the dependencies changed? I had to rebuild the image earlier today because new dependencies were added, maybe that's not the intended way, is it?

spyoungtech:
If the dependencies change, the pipfile/pipfile.lock file needs to change. When either of those files change, the cached layer will not be used and the layer will be built the same way it is today.

spyoungtech:
Yeah, I'm pretty sure CI won't leverage caching. GitLab can do that, but not sure about ADO. 

Mostly it's beneficial for local development :)

Issue #661, opened 2019-11-14
!hush command


lemonsaurus:
Occasionally, two users will be too busy fighting each other to notice a moderator trying to step in and clean up the situation. In these cases, we want to be able to silence the channel for a few minutes while we sort out what's going on and what we're going to do about it.

Enter: `!hush`. This command will remove the ability for anyone except staff to post messages in that channel for a short duration, 10 minutes by default.

This will not need any work on the `site`, so it should be quite approachable.

Since this is a moderator command, [you should write unit tests for it.](https://github.com/python-discord/bot/blob/master/tests/README.md)

## Implementation details:

- [ ] `!hush` should remove posting permissions for non-staff in the channel it is called in for 10 minutes.
- [ ] Add the alias `!silence`
- [ ] `!hush <duration>` should set a custom duration. This duration can either be specified as `5M`, `5m` or just `5`. It should not be possible to exceed 15 minutes in duration.
- [ ] A special alias `!shh` should be added. This alias should be handled by the CommandNotFound handler (the way we do tags aliasing today). That's because we want to be able to support a dynamic amount of `h`'s behind the `sh`. For each `h`, increase the duration by 2 minutes. For example, `!shh` would cause a 4 minute hush, `!shhh` would cause a 6 minute hush, `!shhhh` would cause an 8 minute hush, and so on.  The duration should never exceeed 15 minutes, no matter how many h's you add. The minimum number of `h`'s should be 2.
- [ ] A second command, `!unhush`, should undo the hush immediately. This is important in case the bot restarts in the middle of the duration, since we're not persisting the duration anywhere.
- [ ] The unhush should also have an `!unsilence` alias, and an `!unshhh` with any number of h's.
- [ ] Both commands should have unit tests.

fiskenslakt:
While I love the idea of the dynamic amount of `h`'s idea, I really don't see the point in having a duration for this. Are we assuming that the invoker will forget to unmute the channel? This should absolutely just be a toggle in my opinion. Mute and unmute, that is all.

My rationale for this is two-fold:
1. Duration often causes command misfires due to typing it incorrectly. And given this command is often one that would need to be invoked quickly, pausing to think about a duration is just a waste of time, even if the time wasted is fairly small. Having a duration also makes the command more complicated, and will likely cause some staff to forget how it works, and thus waste time checking the command's help page to see the syntax.
[...]

PR #812, opened 2020-03-02
Implement silencer (hush) cog.

Numerlor:
closes: #661 
Adds silencer cog with requested capabilities.

- [x] `!hush` restricts Developers role in invoked or specified channel.
- [x] `silence`/`unsilence` aliases
- [x] custom duration, including `forever`
- [x] ping mods periodically with channels that are hushed forever.
- [x] `!shh`/`!unshh` with arbitrary amount of `h`s after for specifying duration
- [x] tests
- [ ] ~~list hushed channels~~


bot/cogs/error_handler.py
```diff
@@ -31,7 +31,9 @@ def __init__(self, bot: Bot):
         Error handling emits a single error message in the invoking context `ctx` and a log message,
         prioritised as follows:
 
-        1. If the name fails to match a command but matches a tag, the tag is invoked
+        1. If the name fails to match a command:
+            If it matches shh+ or unshh+, the channel is silenced or unsilenced respectively.
+            otherwise if it matches a tag, the tag is invoked
```
```suggestion
            * If it matches shh+ or unshh+, the channel is silenced or unsilenced respectively.
              Otherwise, if it matches a tag, the tag is invoked
```

bot/cogs/error_handler.py
```diff
@@ -89,6 +95,28 @@ def __init__(self, bot: Bot):
         else:
             return self.bot.get_command("help")
 
+    async def try_silence(self, ctx: Context) -> bool:
+        """
+        Attempt to invoke the silence or unsilence command if invoke with matches a pattern.
+
+        Respecting the checks if:
+                                  invoked with `shh+` silence channel for amount of h's*2 with max of 15.
+                                  invoked with `unshh+` unsilence channel
```
MarkKoz: This is indented way too far.

bot/cogs/error_handler.py
```diff
@@ -89,6 +95,28 @@ def __init__(self, bot: Bot):
...
+        Return bool depending on success of command.
+        """
+        command = ctx.invoked_with.lower()
+        silence_command = self.bot.get_command("silence")
+        if not await silence_command.can_run(ctx):
```
MarkKoz: `can_run` may raise `CommandError`, which should be handled here. Refer to how `try_get_tag` did it.

bot/cogs/error_handler.py
```diff
@@ -49,10 +51,14 @@ def __init__(self, bot: Bot):
...
-            if ctx.channel.id != Channels.verification:
+        if isinstance(e, errors.CommandNotFound):
+            if (
+                    not await self.try_silence(ctx)
+                    and not hasattr(ctx, "invoked_from_error_handler")
+                    and ctx.channel.id != Channels.verification
+            ):
                 await self.try_get_tag(ctx)
-                return  # Exit early to avoid logging.
+            return  # Exit early to avoid logging.
```
MarkKoz: This is wrong. Logging should only be avoided if a tag or the hush command succeeds.

bot/cogs/error_handler.py
```diff
@@ -49,10 +51,14 @@ def __init__(self, bot: Bot):
             return
 
         # Try to look for a tag with the command's name if the command isn't found.
-        if isinstance(e, errors.CommandNotFound) and not hasattr(ctx, "invoked_from_error_handler"):
-            if ctx.channel.id != Channels.verification:
+        if isinstance(e, errors.CommandNotFound):
+            if (
+                    not await self.try_silence(ctx)
+                    and not hasattr(ctx, "invoked_from_error_handler")
+                    and ctx.channel.id != Channels.verification
```
MarkKoz: 1. Though it won't make a difference in this case, generally, the attribute check should be first.
2. These were indented too far.

```suggestion
                not hasattr(ctx, "invoked_from_error_handler")
                and not await self.try_silence(ctx)
                and ctx.channel.id != Channels.verification
```

That being said, I think this is more readable and it maintains control flow similar to what it was originally:

```py
        if isinstance(e, errors.CommandNotFound) and not hasattr(ctx, "invoked_from_error_handler"):
            if await self.try_silence(ctx):
                return
            
            if ctx.channel.id != Channels.verification:
                await self.try_get_tag(ctx)
                return  # Exit early to avoid logging.
```

bot/cogs/moderation/__init__.py
```diff
@@ -10,4 +11,5 @@ def setup(bot: Bot) -> None:
     bot.add_cog(Infractions(bot))
     bot.add_cog(ModLog(bot))
     bot.add_cog(ModManagement(bot))
+    bot.add_cog(Silence(bot))
```
MarkKoz: Add it to the docstring too.

bot/converters.py
```diff
@@ -262,6 +262,34 @@ class ISODateTime(Converter):
...
+    async def convert(self, ctx: Context, argument: str) -> t.Optional[int]:
+        """
+        Convert `argument` to a duration that's max 15 minutes or None.
+
+        If `"forever"` is passed, None is returned; otherwise an int of the extracted time.
+        Accepted formats are:
+                              <duration>,
+                              <duration>m,
+                              <duration>M,
+                              forever.
```
MarkKoz: Indented way too far.

bot/converters.py
```diff
@@ -262,6 +262,34 @@ class ISODateTime(Converter):
...
+        """
+        if argument == "forever":
+            return None
+        match = self.MINUTES_RE.match(argument)
+        if not match:
+            raise BadArgument(f"{argument} is not a valid minutes duration.")
+
+        duration = int(match.group(1))
+        if duration > 15:
+            raise BadArgument("Duration must be below 15 minutes.")
```
MarkKoz:
```suggestion
            raise BadArgument("Duration must be at most 15 minutes.")
```

bot/cogs/moderation/silence.py
```diff
@@ -0,0 +1,175 @@
...
+
+class Silence(commands.Cog):
+    """Commands for stopping channel messages for `verified` role in a channel."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.muted_channels = set()
+        self.bot.loop.create_task(self._get_instance_vars())
+
+    async def _get_instance_vars(self) -> None:
```
MarkKoz: If commands are used before this completes, things will likely blow up. Use an `asyncio.Event` to have commands wait for this to complete.

bot/cogs/moderation/silence.py
```diff
@@ -0,0 +1,175 @@
...
+        if await self._unsilence(ctx.channel):
+            await ctx.send(f"{Emojis.check_mark} unsilenced current channel.")
+
+    async def _silence(self, channel: TextChannel, persistent: bool, duration: Optional[int]) -> bool:
+        """
+        Silence `channel` for `self._verified_role`.
+
+        If `persistent` is `True` add `channel` to notifier.
+        `duration` is only used for logging; if None is passed `persistent` should be True to not log None.
+        """
```
MarkKoz: It'd be helpful to document the return value too.

bot/cogs/moderation/silence.py
```diff
@@ -0,0 +1,175 @@
...
+        log.info(f"Silenced #{channel} ({channel.id}) for {duration} minute(s).")
+        return True
+
+    async def _unsilence(self, channel: TextChannel) -> bool:
+        """
+        Unsilence `channel`.
+
+        Check if `channel` is silenced through a `PermissionOverwrite`,
+        if it is unsilence it and remove it from the notifier.
+        """
```
MarkKoz: It'd be helpful to document the return value too.

bot/cogs/moderation/silence.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __new__(cls, *args):
+        """Construct tuple from `args`."""
+        return super().__new__(cls, args)
+
+    def __hash__(self):
+        return hash((self[0],))
+
+    def __eq__(self, other: "FirstHash"):
+        return self[0] == other[0]
```
MarkKoz: I don't understand how this is really better than a dictionary in this case.

bot/cogs/moderation/silence.py
```diff
@@ -0,0 +1,175 @@
...
+        If `persistent` is `True` add `channel` to notifier.
+        `duration` is only used for logging; if None is passed `persistent` should be True to not log None.
+        """
+        current_overwrite = channel.overwrites_for(self._verified_role)
+        if current_overwrite.send_messages is False:
+            log.info(f"Tried to silence channel #{channel} ({channel.id}) but the channel was already silenced.")
+            return False
+        await channel.set_permissions(
+            self._verified_role,
+            overwrite=PermissionOverwrite(**dict(current_overwrite, send_messages=False))
```
MarkKoz: > The overwrite parameter, if given, must either be None or PermissionOverwrite. **For convenience, you can pass in keyword arguments denoting Permissions attributes.** If this is done, then you cannot mix the keyword arguments with the overwrite parameter.

Passing kwargs is a better option when only interested in setting one permission. I'm pretty sure it will leave the rest of the overrides intact, but I have not tested.

bot/cogs/moderation/silence.py
```diff
@@ -0,0 +1,175 @@
...
+        Unsilence `channel`.
+
+        Check if `channel` is silenced through a `PermissionOverwrite`,
+        if it is unsilence it and remove it from the notifier.
+        """
+        current_overwrite = channel.overwrites_for(self._verified_role)
+        if current_overwrite.send_messages is False:
+            await channel.set_permissions(
+                self._verified_role,
+                overwrite=PermissionOverwrite(**dict(current_overwrite, send_messages=True))
```
MarkKoz: As mentioned before, permissions can be passed as kwargs rather than using an entire `PermissionOverwrite` object.

bot/cogs/error_handler.py
Numerlor: How do we want to handle it here?
Invoking the error handler (and returning `False`) will cause it to be triggered twice because of the `try_get_tag` that follows.
The errors could be silenced for the user with just being logged, or None can be returned to also cancel the tag fallback

bot/cogs/moderation/silence.py
Numerlor: How necessary is this? The vars finish loading in milliseconds after the bot connects and the commands become available for invoking.

MarkKoz: `wait_until_guild_available` may take a long time. It'd be a rare occurrence but I think it's a low effort to make it more robust.

bot/cogs/moderation/silence.py
Numerlor: Providing kwargs creates a permission overwrite object from them [internally](https://github.com/Rapptz/discord.py/blob/5e6335750819588062790c6b93d604d3bc7c3862/discord/abc.py#L620-L626), so it doesn't keep them.

But realized that we can skip the PermissionOverwrite itself here and just unpack the dict into kwargs.

bot/cogs/error_handler.py
MarkKoz: The `invoked_from_error_handler` attribute exists to prevent long or infinite recursion in the error handler when it's a `CommandNotFound`.

Numerlor: Copying that behaviour (and not changing anything else), 
the `await self.on_command_error(ctx, error)` from the methods can get executed twice. Once in the silence attempt and then in the tag attempt.
`invoked_from_error_handler` doesn't come into play here because we're skipping the error handler with a `CommandError` with which it gets invoked.

bot/cogs/error_handler.py
Numerlor: Was aiming for a list like look here (under the colon).
Should it just be a 2-4 spaces?

bot/cogs/error_handler.py
MarkKoz: Yes, the command handler can be triggered twice if both commands error out, and that's fine. That's the point: why write code (logging, sending a message, etc.) to handle errors when the error handler already has code which can handle them?

bot/cogs/error_handler.py
MarkKoz: Yes, 4 spaces should be fine. If you think it helps you can prepend an asterisk or number them.

bot/cogs/error_handler.py
Numerlor: Do we want it to be triggered twice? It can result in duplicate messages being sent on an invalid command.

MarkKoz: No, probably not. Let's just go with what you said and log it directly.

bot/cogs/error_handler.py
Numerlor: Went with an asterisk and a space like in other docstrings in the module.

tests/bot/cogs/moderation/test_silence.py
```diff
@@ -0,0 +1,219 @@
...
+        test_cases = (
+            (0.0001, f"{Emojis.check_mark} silenced current channel for 0.0001 minute(s).", True,),
+            (None, f"{Emojis.check_mark} silenced current channel indefinitely.", True,),
+            (5, f"{Emojis.cross_mark} current channel is already silenced.", False,),
+        )
+        for duration, result_message, _silence_patch_return in test_cases:
+            with self.subTest(
+                    silence_duration=duration,
+                    result_message=result_message,
+                    starting_unsilenced_state=_silence_patch_return
```
MarkKoz:
```suggestion
                silence_duration=duration,
                result_message=result_message,
                starting_unsilenced_state=_silence_patch_return
```

tests/bot/cogs/moderation/test_silence.py
```diff
@@ -0,0 +1,219 @@
...
+            ):
+                with mock.patch.object(self.cog, "_silence", return_value=_silence_patch_return):
+                    await self.cog.silence.callback(self.cog, self.ctx, duration)
+                    self.ctx.send.assert_called_once_with(result_message)
+            self.ctx.reset_mock()
```
MarkKoz: Since this is outside the subtest, assertion errors won't prevent this reset from being performed, right?

tests/bot/cogs/moderation/test_silence.py
```diff
@@ -0,0 +1,219 @@
...
+
+    async def test_unsilence_sent_correct_discord_message(self):
+        """Check if proper message was sent to `alert_chanel`."""
```
MarkKoz:
```suggestion
        """Check if proper message was sent to `alert_channel`."""
```

MarkKoz: On second thought, what does this have to do with `alert_channel`? It's sending to `ctx`.

tests/bot/cogs/moderation/test_silence.py
```diff
@@ -0,0 +1,219 @@
...
+
+        self.assertFalse(await self.cog._silence(channel, True, None))
+        channel.set_permissions.assert_not_called()
+
+    async def test_silence_private_silenced_channel(self):
+        """Channel had `send_message` permissions revoked."""
+        channel = MockTextChannel()
+        self.assertTrue(await self.cog._silence(channel, False, None))
+        channel.set_permissions.assert_called_once()
+        self.assertFalse(channel.set_permissions.call_args.kwargs['send_messages'])
```
MarkKoz: For thoroughness, it could also test that existing overwrites on a channel are preserved.

tests/bot/cogs/moderation/test_silence.py
```diff
@@ -0,0 +1,219 @@
...
+                await self.cog._silence(channel, True, None)
+                self.cog.notifier.add_channel.assert_called_once()
+
+        with mock.patch.object(self.cog, "notifier", create=True):
+            with self.subTest(persistent=False):
+                await self.cog._silence(channel, False, None)
+                self.cog.notifier.add_channel.assert_not_called()
+
+    async def test_silence_private_removed_muted_channel(self):
+        channel = MockTextChannel()
```
MarkKoz: Missing a docstring. Not enforced for tests but since you added them everywhere else I figured I'd point it out.

tests/bot/cogs/moderation/test_silence.py
```diff
@@ -0,0 +1,219 @@
...
+        muted_channels.remove.assert_called_once_with(channel)
+
+    @mock.patch("bot.cogs.moderation.silence.asyncio")
+    @mock.patch.object(Silence, "_mod_alerts_channel", create=True)
+    def test_cog_unload(self, alert_channel, asyncio_mock):
+        """Task for sending an alert was created with present `muted_channels`."""
+        with mock.patch.object(self.cog, "muted_channels"):
+            self.cog.cog_unload()
+            asyncio_mock.create_task.assert_called_once_with(alert_channel.send())
+            alert_channel.send.called_once_with(f"<@&{Roles.moderators}> chandnels left silenced on cog unload: ")
```
MarkKoz: How did this even pass tests?
```suggestion
            alert_channel.send.called_once_with(f"<@&{Roles.moderators}> channels left silenced on cog unload: ")
```
Still, this shouldn't pass because it's missing the list of channel mentions at the end.

tests/bot/cogs/moderation/test_silence.py
```diff
@@ -0,0 +1,219 @@
...
+
+    @mock.patch("bot.cogs.moderation.silence.asyncio")
+    def test_cog_unload1(self, asyncio_mock):
```
MarkKoz: This needs a better name.

tests/bot/cogs/moderation/test_silence.py
MarkKoz: Something seems to be wrong because the assertion passes with any value.

tests/bot/cogs/moderation/test_silence.py
Numerlor: Can't believe I missed this one, thanks.

tests/bot/cogs/moderation/test_silence.py
Numerlor: Missed the assert in here.

tests/bot/cogs/moderation/test_silence.py
Numerlor: Yep, called every time; tried with an `assert False` in the test block.

tests/bot/cogs/moderation/test_silence.py
Numerlor: Good idea, implemented 2 new tests for this when using silence and unsilence.

Numerlor:
Managed to lose my changes to the converter tests when I was fixing up my commits and didn't notice, pushed them in now.

bot/cogs/error_handler.py
```diff
@@ -89,6 +93,33 @@ def __init__(self, bot: Bot):
...
+        ctx.invoked_from_error_handler = True
+        try:
+            if not await silence_command.can_run(ctx):
+                log.debug("Cancelling attempt to invoke silence/unsilence due to failed checks.")
+                return False
+        except errors.CommandError:
+            log.debug("Cancelling attempt to invoke silence/unsilence due to failed checks.")
+            return False
+        if command.startswith("shh"):
+            await ctx.invoke(silence_command, duration=min(command.count("h")*2, 15))
```
eivl: Magic numbers aside, I really like this implementation.

PR #813, opened 2020-03-03
Cache all Python dependencies in CI

MarkKoz:
[Caching](https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops) of Python dependencies to reduce frequency of using pipenv to install dependencies in CI.

## What is a cache?

A cache saves a directory for re-use in a later build. In summary, the cache works by using a key to determine if there is a cache hit or miss. The key can comprise of strings and file hashes. When there's a cache hit, the cached directory is restored. When there's a miss, there will be a job that runs at the end to upload the directory to the cache (i.e. creating a new cache). 

There are also partial cache hits which perform both a restore and an upload. Partial restores can be useful if only one dependency changed and the rest can still be re-used. As a result, **a potential issue is that removed packages may linger in the cache.** Support for partial restores may need to be removed. Please comment if you have any suggestions on this.

## How its used

Dependencies are redirected to be installed to the [user base](https://docs.python.org/3/library/site.html?highlight=site#site.USER_BASE) directory, which gets cached. Because nothing else uses the user site, it is useful for caching _only_ our dependencies, which keeps the cache slim and the operations quick.

## Benchmarks

Only the affected build steps were benchmarked.

### Old times

* Installing pipenv: 5-10s
* Installing dependencies: 34s

Total: 39s

### New times

**Cache Miss**:

Example build [here](https://dev.azure.com/python-discord/Python%20Discord/_build/results?buildId=6405&view=logs&j=5264e576-3c6f-51f6-f055-fab409685f20&t=74cc461b-e097-4214-b1b6-8481db81983d) (note that pre-commit also has a cache miss).

* Restoring cache: 3s (realises there's a miss so it does nothing)
* Installing pipenv: 5-10s
* Installing dependencies: 33s
* Saving cache: 13-16s

Total: 55s (16s slower)

**Cache Hit**:

Example build [here](https://dev.azure.com/python-discord/Python%20Discord/_build/results?buildId=6415&view=logs&j=5264e576-3c6f-51f6-f055-fab409685f20&t=ba73fe70-97c4-5e2d-5cc4-47c0f87130c8).

* Restoring from cache: 10s
    * Inconsistent. One time it took 10s, the next time 18s.

Total: 10s (29s faster)

## Cache speed inconsistency

The speed of restoring or creating a cache varies a bit. This just seems to be an issue with Azure Pipelines. For example, [here](https://dev.azure.com/python-discord/Python%20Discord/_build/results?buildId=6396&view=logs&j=5264e576-3c6f-51f6-f055-fab409685f20&t=419a0abe-6ae2-5290-e72a-c67ad02dc29f) it took 24s and [here](https://dev.azure.com/python-discord/Python%20Discord/_build/results?buildId=6397&view=logs&j=5264e576-3c6f-51f6-f055-fab409685f20&t=419a0abe-6ae2-5290-e72a-c67ad02dc29f) 27s to realise there's a cache miss when it normally takes 3-5s. [Here](https://dev.azure.com/python-discord/Python%20Discord/_build/results?buildId=6402&view=logs&j=5264e576-3c6f-51f6-f055-fab409685f20&t=419a0abe-6ae2-5290-e72a-c67ad02dc29f) it took 21s to upload the cache when it normally takes 13-15s. These slowdowns happen more often than I'd like, but I believe the changes are still going to result in a net positive for time.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ jobs:
     steps:
       - task: UsePythonVersion@0
         displayName: 'Set Python version'
-        name: PythonVersion
+        name: python
         inputs:
           versionSpec: '3.8.x'
           addToPath: true
@@ -34,11 +34,11 @@ jobs:
       - task: Cache@2
         displayName: 'Restore Python environment'
         inputs:
-          key: python | $(Agent.OS) | "$(PythonVersion.pythonLocation)" | 0 | ./Pipfile | ./Pipfile.lock
+          key: python | $(Agent.OS) | "$(python.pythonLocation)" | 0 | ./Pipfile | ./Pipfile.lock
           restoreKeys: |
-            python | "$(PythonVersion.pythonLocation)" | 0 | ./Pipfile.lock
-            python | "$(PythonVersion.pythonLocation)" | 0 | ./Pipfile
-            python | "$(PythonVersion.pythonLocation)" | 0
+            python | "$(python.pythonLocation)" | 0 | ./Pipfile.lock
+            python | "$(python.pythonLocation)" | 0 | ./Pipfile
+            python | "$(python.pythonLocation)" | 0
           cacheHitVar: PY_ENV_RESTORED
           path: $(PYTHONUSERBASE)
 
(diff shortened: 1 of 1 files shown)


MarkKoz:
Not sure if saving 24 seconds is worth it since the cache restoration performance seems unstable. Can be 8s, 10s, 16s, 18s, or even [bug out](https://dev.azure.com/python-discord/Python%20Discord/_build/results?buildId=6287&view=logs&j=5264e576-3c6f-51f6-f055-fab409685f20&t=419a0abe-6ae2-5290-e72a-c67ad02dc29f) and be over a minute.

MarkKoz:
If pipenv supported installing to the user site (or any separate site, really) instead of the global site, then it'd probably be feasible to cache the user site.

MarkKoz:
> We'll likely have to evaluate any potential issues with partial cache hits as they happen in order to determine whether or not it represents a recurring issue. Just something to be vigilant for in our diagnosis of any potential failures.

It will either _always_ happen when a package is removed from the lock file, or it will _never_ happen. I think it will be the former but I have not tested. At the least, it means the cache will get bloated with packages which are no longer needed. They will keep accumulating as more packages get removed unless the cache is unused for 7 days and gets automatically deleted by Azure Pipelines. Lingering packages are also a potential cause for conflicts (e.g. unused package `x` depends on package `y==4.0` and useful package `z` depends on `y==5.0`).

Removing support for partial hits means all dependencies will be installed whenever a single dependency changes. Of course that isn't so bad, but being able to leverage an old cache for some dependencies would have been nice.

It's a shame `pipenv clean` does not support `--system`. I've looked into implementing some sort of cleaning ourselves but it's proven to be complicated due to package name differences (dot vs dash, capital vs lowercase, etc.).

MarkKoz:
Decided to just remove support for partial cache hits to be on the safe side.

MarkKoz:
Here goes nothing.

PR #817, opened 2020-03-05
Writing tests for moderation utils.

### LINKED ISSUE #562: Write unit tests for `bot/cogs/moderation/utils.py`
Write unit tests for [`bot/cogs/moderation/utils.py`](../blob/master/bot/cogs/moderation/utils.py).

## Implementation details

## Additional information

ks129:
Closes #562.

Adds tests for the `moderation/utils` module & factors a few strings out into module-level constants to allow easier assertions about their content from the tests.

Additionally, this PR fixes this regression:

image

Due to the use of [`textwrap.shorten`](https://docs.python.org/3/library/textwrap.html#textwrap.shorten) on the final embed description, we were losing newlines. The new approach simply slices the string if it's too long, and ends it with 3 dots.

image


tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,345 @@
...
+            return_value = case["get_return_value"]
+            expected = case["expected_output"]
+
+            with self.subTest(args=args, return_value=return_value, expected=expected):
+                self.bot.api_client.get.return_value = return_value
+
+                result = await has_active_infraction(*args)
+                self.assertEqual(result, expected)
+
+    async def test_notify_infraction(self):
```
SebastiaanZ: This test currently depends on the working of two separate units of code: the coroutines `notify_infraction` and `send_private_embed`. In fact, you currently assert whether or not `notify_infraction` works by looking at something that `send_private_embed` produces. That means that we're not testing those units independently from each other and, if the test fails, we don't know which unit caused it to fail.

While it obviously happens that one unit uses another unit in its execution, I think that in this specific case, it's better to isolate the two by patching the `send_private_embed` coroutine withing this test. That allows us to later assert whether or not `send_private_embed` was awaited with the correct arguments using one of the `unittest.AsyncMock` test method.

```suggestion
    @patch("bot.cogs.moderation.utils.send_private_embed")
    async def test_notify_infraction(self, send_private_embed_mock):
```

Another important characteristic of the coroutine we're testing here is that it returns the return value of the `send_private_embed` coroutine it calls. This is an "essential characteristic" for the features that use this utility, so we need to assert that it indeed does so. That way, if someone happens to accidentally remove the `return` statement from the last line of the coroutine while working on it the tests will reflect that mistake by failing.

Now, we don't actually have to invoke the `send_private_embed` coroutine for that, as that would make it depend on that other unit again; instead, we can simply provide a return value for the mock we created by patching `send_private_embed` and asserting that the unit we're actually testing does indeed pass on that return value by returning it as well.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+        ]
+
+        for case in test_cases:
+            args = case["args"]
+            expected = case["expected_output"]
+            send, send_raise = case["send_result"], case["send_raise"]
+
+            with self.subTest(args=args, expected=expected):
+                if send_raise:
+                    self.ctx.send.side_effect = send_raise
```
MarkKoz: What is the purpose of this? `notify_pardon()` never calls `send()` directly and `send_private_embed()` is mocked so it won't call `send()` either.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+        ]
+
+        for case in test_cases:
+            args = case["args"]
+            expected = case["expected_output"]
+            send, send_raise = case["send_result"], case["send_raise"]
+
+            with self.subTest(args=args, expected=expected, send=send, send_raise=send_raise):
+                if send_raise:
+                    self.ctx.send.side_effect = send_raise
```
MarkKoz: What is the purpose of this? `notify_infraction()` never calls `send()` directly and `send_private_embed()` is mocked so it won't call `send()` either.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+                if not error:
+                    self.bot.api_client.post.assert_awaited_once_with("bot/users", json=payload)
+
+                self.bot.api_client.post.reset_mock(side_effect=True)
+
+    async def test_send_private_embed(self):
+        """Test does `send_private_embed` return correct value."""
+        test_cases = [
+            {
+                "args": (self.user, Embed(title="Test", description="Test val")),
```
MarkKoz: These args the same across all test cases. At the very least, this should be defined once and assigned to a variable that all these cases use. However, they don't really even need to be in the test cases - the aforementioned variable could just be used directly when calling the function.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+                "expected_output": False,
+                "raised_exception": NotFound(AsyncMock(), AsyncMock())
+            }
+        ]
+
+        for case in test_cases:
+            args = case["args"]
+            expected = case["expected_output"]
+            raised: Union[Forbidden, HTTPException, NotFound, None] = case["raised_exception"]
```
MarkKoz: Why give it a type annotation?

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+
+            with self.subTest(args=args, expected=expected, raised=raised):
+                if raised:
```
MarkKoz: There's no need for this conditional. You used a value of `None` in the test case when an exception shouldn't be raised. Well, if `side_effect = None` then that's the same thing as not having a side effect. Therefore, it's perfectly safe to assign `None` to the `side_effect` when an exception shouldn't be raised.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+                send_private_embed_mock.assert_awaited_once_with(args[0], embed)
+
+                self.ctx.send.reset_mock(side_effect=True)
+                send_private_embed_mock.reset_mock()
+
+    async def test_post_user(self):
+        """Test does `post_user` work correctly."""
+        test_cases = [
+            {
+                "args": (self.ctx, self.user),
```
MarkKoz: Same point as other comments I've made: these args are constant so don't repeat them in every test case (though based on another comment I made, you will need to add a second user to test the payload).

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+                        "avatar": "test",
+                        "name": "Test",
+                        "discriminator": 1234,
+                        "roles": [
+                            1234,
+                            5678
+                        ],
+                        "in_guild": False
+                    }
+                ],
```
MarkKoz: First, this is constant across test cases. Second, there's no need for it to be emulating a real-world value. `post_user` just returns whatever `api_client.post` returns. Because you mock the latter, you control its return value. You just need to assert that the return value of `post_user` equals the return value of `api_client.post` e.g.:

```py
expected = "dummy value"

self.ctx.bot.api_client.post.return_value = expected
result = await post_user(*args)
self.assertEqual(result, expected)
```

The above achieves the same effect.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+                "raise_error": False,
+                "payload": {
+                    "avatar_hash": getattr(self.user, "avatar", 0),
+                    "discriminator": int(getattr(self.user, "discriminator", 0)),
+                    "id": self.user.id,
+                    "in_guild": False,
+                    "name": getattr(self.user, "name", "Name unknown"),
+                    "roles": []
+                }
```
MarkKoz: Don't just copy the `getattr` stuff. To properly test this, use different users - one which has the attributes and one which doesn't. Then, here you'd hard code the values you expect the payload to have.

Also want to point out this is *currently* another instance of test data being repeated across test cases.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+                        "inserted_at": "2018-11-22T07:24:06.132307Z",
+                        "expires_at": "5018-11-20T15:52:00Z",
+                        "active": True,
+                        "user": 1234,
+                        "actor": 1234,
+                        "type": "ban",
+                        "reason": "Test Ban",
+                        "hidden": False
+                    }
+                ],
```
MarkKoz: As I pointed out elsewhere, the data doesn't need to be this accurate. Using "dummy value" would achieve the same results since `post_infraction` just returns whatever `api_client.post` returns.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+            args = case["args"]
+            expected = case["expected_output"]
+            raised = case["raised_error"]
+            payload = case["payload"]
+
+            with self.subTest(args=args, expected=expected, raised=raised, payload=payload):
+                self.ctx.bot.api_client.post.reset_mock(side_effect=True)
+                post_user_mock.reset_mock()
+
+                if raised:
```
MarkKoz: As I wrote elsewhere, it's fine to assign `None` to `side_effect` so this check isn't needed.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+                    self.ctx.bot.api_client.post.side_effect = raised
+
+                post_user_mock.return_value = "foo"
+
+                self.ctx.bot.api_client.post.return_value = expected
+
+                result = await post_infraction(*args)
+
+                self.assertEqual(result, expected)
```
MarkKoz: Why so much space between everything? The top two and bottom two could be grouped together.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+
+                if not raised:
+                    self.bot.api_client.post.assert_awaited_once_with("bot/infractions", json=payload)
+
+                if hasattr(raised, "status") and hasattr(raised, "response_json"):
```
MarkKoz: Why use `hasattr`. As far as I can tell, the errors you use are either `None` or `ResponseCodeError`, of which the latter always has those attributes. It seems `if raised:` would suffice.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+                    if raised.status == 400 and "user" in raised.response_json:
+                        post_user_mock.assert_awaited_once_with(args[0], args[1])
```
MarkKoz: This isn't testing the false branch in which an error message should be sent to the context.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+
+                result = await post_user(*args)
+
+                if error:
+                    self.assertIsNone(result)
+                else:
+                    self.assertEqual(result, expected)
+
+                if not error:
+                    self.bot.api_client.post.assert_awaited_once_with("bot/users", json=payload)
```
MarkKoz: This isn't asserting the true branch in which an error should be sent to the context.

tests/bot/cogs/moderation/test_utils.py
MarkKoz: This doesn't test the control flow for the loop it uses with `should_post_user`. Here's an outline of the behaviour you need to test:

- [x] Post the infraction
- [x] If the infraction post fails cause the user isn't in the db, post the user
- [ ] If the user post fails, exit out of the loop right away
- [ ] If the user post succeeds, attempt to post the infraction again
- [ ] If the infraction fails to post _again_, then do nothing and exit the loop
- [ ] If the error is something besides a 400, send an error message and exit the loop

tests/bot/cogs/moderation/test_utils.py
ks129: Is like for one should I use random `avatar_hash`, but for second 0, same for `discriminator` and `name`?

tests/bot/cogs/moderation/test_utils.py
ks129: How I should do all these other checks? It's confusing for me.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -176,20 +176,18 @@ def setUp(self):
...
                 "args": (self.user, "Test title 1", "Example content 1", Icons.user_update),
-                "expected_output": {
-                    "description": "Example content 1",
-                    "title": "Test title 1",
-                    "icon_url": Icons.user_update
-                },
+                "expected_output": Embed(
+                    description="Example content 1",
+                    colour=PARDON_COLOR
+                ).set_author(name="Test title 1", icon_url=Icons.user_update),
```
MarkKoz: These embeds really only differ by the icon. Change the test cases so that only the icon is provided. The test can directly create the rest of the embed within the subtest.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -41,54 +41,35 @@ def setUp(self):
...
-                    "reason": "Test",
-                    "hidden": False
-                }],
+                "get_return_value": [{"id": 1}],
                 "expected_output": True,
-                "send_params": (
-                    f":x: According to my records, this user already has a ban infraction. "
-                    f"See infraction **#1**."
-                )
+                "infraction_nr": "**#1**"
```
MarkKoz: You could use the ID from `get_return_value` to get this value in the test, so including it here feels redundant. If you make the aforementioned ID some uncommon number like 83492, then you could forego having the `**` and `#` too i.e. it'll be safe to just search for the number directly.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -77,54 +77,76 @@ def setUp(self):
...
-                },
+                    colour=INFRACTION_COLOR,
+                    url=utils.RULES_URL
+                ).set_author(
+                    name=INFRACTION_AUTHOR_NAME,
+                    url=utils.RULES_URL,
+                    icon_url=Icons.defcon_denied
+                ).set_footer(
+                    text=INFRACTION_APPEAL_FOOTER
+                ),
```
MarkKoz:
```suggestion
                ).set_footer(text=INFRACTION_APPEAL_FOOTER),
```

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -212,54 +212,31 @@ def setUp(self):
...
-                        "discriminator": 1234,
-                        "roles": [
-                            1234,
-                            5678
-                        ],
-                        "in_guild": False
-                    }
-                ],
+                "args": (self.ctx, user),
+                "post_result": "bar",
```
MarkKoz: You can just "hard-code" these in the test code, no need to define different values in the test cases.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -212,54 +212,31 @@ def setUp(self):
...
-                        "name": "Test",
-                        "discriminator": 1234,
-                        "roles": [
-                            1234,
-                            5678
-                        ],
-                        "in_guild": False
-                    }
-                ],
+                "args": (self.ctx, user),
```
MarkKoz: The context could be omitted since it's constant. You'd probably want to rename the key from `"args"` to `"user"` in that case.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -276,6 +253,8 @@ def setUp(self):
 
                 if error:
                     self.ctx.bot.api_client.post.side_effect = ResponseCodeError(AsyncMock(), expected)
+                    err = self.ctx.bot.api_client.post.side_effect
+                    err.status = 400
```
MarkKoz: At the least, do this.
```suggestion
                    err = ResponseCodeError(Mock(status=400), expected)
                    self.ctx.bot.api_client.post.side_effect = err
```
You could also include this in the test case rather than using a boolean, which would make the code a tiny bit more direct.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -286,6 +265,8 @@ def setUp(self):
 
                 if not error:
                     self.bot.api_client.post.assert_awaited_once_with("bot/users", json=payload)
+                else:
+                    self.assertTrue(str(err.status) in self.ctx.send.call_args[0][0])
```
MarkKoz: It's a good idea to still assert `send()` was awaited before accessing the `call_args`, because if it isn't called `call_args` will be `None`. It's preferred for tests to fail via assertion failures rather than some `TypeError` as the former is easier to understand and fix.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -77,54 +77,76 @@ def setUp(self):
...
-                    "icon_url": Icons.token_removed,
-                    "footer": INFRACTION_APPEAL_FOOTER,
-                },
+                    colour=INFRACTION_COLOR,
+                    url=utils.RULES_URL
+                ).set_author(
+                    name=INFRACTION_AUTHOR_NAME,
+                    url=utils.RULES_URL,
+                    icon_url=Icons.token_removed
+                ).set_footer(text=INFRACTION_APPEAL_FOOTER),
```
MarkKoz: Ultimately I'm not sure if this turned out better than what existed before. I'll let the other reviewer(s) have the final say and I'm just gonna accept this as it is.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+                "expected_output": True,
+                "get_call": {
+                    "active": "true",
+                    "type": "ban",
+                    "user__id": str(self.member.id)
+                },
+                "send_params": (
+                    f":x: According to my records, this user already has a ban infraction. "
+                    f"See infraction **#1**."
+                )
```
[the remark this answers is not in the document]
MarkKoz: The code has been changed but I'll leave this unresolved for other reviewers to perhaps discuss.

tests/bot/cogs/moderation/test_utils.py
MarkKoz: Yes that is what I meant. It seems to have been implemented correctly.

tests/bot/cogs/moderation/test_utils.py
MarkKoz: For the third point, consider that when the loop repeats, `post()` gets called. It is undesirable for that to be repeated, therefore I phrased it as the loop being exited. But really, think about it as "assert that `post()` doesn't happen again". Such an assertion can be made by looking at the `call_count` attribute of the mock.

When a `post()` is successful, the test should ensure that no error message is sent and no user is posted. If the aforementioned assertions are made, then it effectively means the loop exited at that point. This should also be done when the `post()` fails the _second_ time too (i.e. the case in fifth point above). In most of the outlined cases, you'll want to have assertions for `call_count` to make sure it's not called too few or too many times.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -305,114 +305,71 @@ def setUp(self):
...
+        self.assertTrue("500" in self.ctx.send.call_args[0][0])
 
-                if hasattr(raised, "status") and hasattr(raised, "response_json"):
-                    if raised.status == 400 and "user" in raised.response_json:
-                        post_user_mock.assert_awaited_once_with(args[0], args[1])
+    @patch("bot.cogs.moderation.utils.post_user")
+    async def test_user_not_found_none_post_infraction(self, post_user_mock):
+        """Test does `post_infraction` return `None` correctly due can't create new user."""
+        self.bot.api_client.post.side_effect = ResponseCodeError(MagicMock(status=400), {"user": "foo"})
+        post_user_mock.return_value = None
```
MarkKoz: Can be passed as a kwarg to the patch decorator.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -305,114 +305,71 @@ def setUp(self):
...
+            "actor": self.ctx.message.author.id,
+            "hidden": False,
+            "reason": "Test reason",
+            "type": "mute",
+            "user": self.user.id,
+            "active": True
+        }
+
+        self.bot.api_client.post.side_effect = [ResponseCodeError(MagicMock(status=400), {"user": "foo"}), "foo"]
+        post_user_mock.return_value = "bar"
```
MarkKoz: Can be passed as a kwarg to the patch decorator.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -305,114 +305,71 @@ def setUp(self):
...
+
+        actual = await utils.post_infraction(self.ctx, self.user, "mute", "Test reason")
+        self.assertEqual(actual, "foo")
+        self.bot.api_client.post.assert_awaited_once_with("bot/infractions", json=payload)
```
MarkKoz: Should assert it got called twice.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -377,5 +377,5 @@ def setUp(self):
 
         actual = await utils.post_infraction(self.ctx, self.user, "mute", "Test reason")
         self.assertEqual(actual, "foo")
-        self.bot.api_client.post.assert_awaited_with("bot/infractions", json=payload)
+        self.bot.api_client.post.assert_has_awaits([call("bot/infractions", json=payload)] * 2)
```
MarkKoz: I was thinking of just using call count originally but this is an even better way of doing it.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,378 @@
...
+                self.user.send.side_effect = raised
+
+                result = await utils.send_private_embed(self.user, embed)
+
+                self.assertEqual(result, expected)
+                if expected:
+                    self.user.send.assert_awaited_once_with(embed=embed)
+
+
+class TestPostInfraction(unittest.IsolatedAsyncioTestCase):
```
MarkKoz: A separate test case was a good decision. Keeps things more organised.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,378 @@
...
+INFRACTION_AUTHOR_NAME = "Infraction information"
+INFRACTION_COLOR = Colours.soft_red
+
+INFRACTION_DESCRIPTION_TEMPLATE = (
+    "\n**Type:** {type}\n"
+    "**Expires:** {expires}\n"
+    "**Reason:** {reason}\n"
+)
+
+PARDON_COLOR = Colours.soft_green
```
MarkKoz: I don't feel this constant needs an alias.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,378 @@
...
+from bot.cogs.moderation import utils
+from bot.constants import Colours, Icons
+from tests.helpers import MockBot, MockContext, MockMember, MockUser
+
+APPEAL_EMAIL = "appeals@pythondiscord.com"
+
+INFRACTION_TITLE = f"Please review our rules over at {utils.RULES_URL}"
+INFRACTION_APPEAL_FOOTER = f"To appeal this infraction, send an e-mail to {APPEAL_EMAIL}"
+INFRACTION_AUTHOR_NAME = "Infraction information"
+INFRACTION_COLOR = Colours.soft_red
```
MarkKoz: I don't feel this constant needs an alias.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,378 @@
...
+from bot.api import ResponseCodeError
+from bot.cogs.moderation import utils
+from bot.constants import Colours, Icons
+from tests.helpers import MockBot, MockContext, MockMember, MockUser
+
+APPEAL_EMAIL = "appeals@pythondiscord.com"
+
+INFRACTION_TITLE = f"Please review our rules over at {utils.RULES_URL}"
+INFRACTION_APPEAL_FOOTER = f"To appeal this infraction, send an e-mail to {APPEAL_EMAIL}"
+INFRACTION_AUTHOR_NAME = "Infraction information"
```
MarkKoz: Define these in the utils module instead. If the utils module also uses these constants, it will reduce "desync" between it and the tests should any of these strings need to be changed.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,378 @@
...
+INFRACTION_COLOR = Colours.soft_red
+
+INFRACTION_DESCRIPTION_TEMPLATE = (
+    "\n**Type:** {type}\n"
+    "**Expires:** {expires}\n"
+    "**Reason:** {reason}\n"
+)
```
MarkKoz: As with the other constants, this could be defined and used by the utils module too.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,411 @@
...
+                "infraction_nr": "123987",
+                "send_msg": False
+            },
+            {
+                "get_return_value": [{"id": 123987}],
+                "expected_output": {"id": 123987},
+                "infraction_nr": "123987",
+                "send_msg": True
+            }
+        ]
```
kwzrd, line 48: Not asking you to change these, but wouldn't it be better to have a `namedtuple` or a similar construct to define the cases? The repeated subscription with string literals seems fragile and needlessly inconvenient.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,411 @@
...
+                self.bot.api_client.get.return_value = case["get_return_value"]
+
+                result = await utils.get_active_infraction(self.ctx, self.member, "ban", send_msg=case["send_msg"])
+                self.assertEqual(result, case["expected_output"])
+                self.bot.api_client.get.assert_awaited_once_with("bot/infractions", params=params)
+
+                if case["send_msg"] and case["get_return_value"]:
+                    self.ctx.send.assert_awaited_once()
+                    self.assertTrue(case["infraction_nr"] in self.ctx.send.call_args[0][0])
+                    self.assertTrue("ban" in self.ctx.send.call_args[0][0])
```
kwzrd, line 70: Can use `assertIn` to get a much nicer error message if the test fails.

Perhaps assigning `self.ctx.send.call_args[0][0]` to a variable first would greatly improve readability?
```python
self.assertIn("ban", sent_message)
```

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,477 @@
...
+
+                self.bot.api_client.get.reset_mock()
+                self.ctx.send.reset_mock()
+
+    @patch("bot.cogs.moderation.utils.send_private_embed")
+    async def test_notify_infraction(self, send_private_embed_mock):
+        """Test does `notify_infraction` create correct result."""
+        test_cases = [
+            {
+                "args": (self.user, "ban", "2020-02-26 09:20 (23 hours and 59 minutes)"),
```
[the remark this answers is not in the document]
kwzrd: Considering it's a positional arg I don't mind it being here at all.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,411 @@
...
+class ModerationUtilsTests(unittest.IsolatedAsyncioTestCase):
+    """Tests Moderation utils."""
+
+    def setUp(self):
+        self.bot = MockBot()
+        self.member = MockMember(id=1234)
+        self.user = MockUser(id=1234)
+        self.ctx = MockContext(bot=self.bot, author=self.member)
+
+    async def test_user_get_active_infraction(self):
```
kwzrd, line 23: Since the function is called `get_active_infraction`, I'm just wondering why the "user" in the test name?

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,411 @@
...
+            {
+                "args": (self.user, "Test title", "Example content", Icons.user_update),
+                "icon": Icons.user_update,
+                "send_result": False
+            }
+        ]
+
+        for case in test_cases:
+            args = case["args"]
+            send = case["send_result"]
```
kwzrd, line 211: More of an addition to my other comment about the choice to define cases using dicts. I think a simple tuple unpacked in the for loop would be fine:
```python
for args, icon, send_result in test_cases:
```
But it can be seen as a matter of personal preference. You're probably doing `args = case["args"]` for convenience further down the function, so I'm just trying to ask the question whether we can define the test cases such that they already are convenient to work with, without the intermediate step.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,411 @@
...
+            }
+        ]
+
+        for case in test_cases:
+            test_user = case["user"]
+            expected = case["post_result"]
+            error = case["raise_error"]
+            payload = case["payload"]
+
+            with self.subTest(user=test_user, result=expected, error=error, payload=payload):
```
kwzrd, line 284: I'd strongly prefer if the names you're assigning to matched the keys. This creates even more indirection between the definitions above and the code below. You're using 3 different names (key, local variable, `subTest` kwarg) for the same thing.

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,411 @@
...
+                if not error:
+                    self.bot.api_client.post.assert_awaited_once_with("bot/users", json=payload)
+                else:
+                    self.ctx.send.assert_awaited_once()
+                    self.assertTrue(str(error.status) in self.ctx.send.call_args[0][0])
+
+                if isinstance(test_user, MagicMock):
+                    log_mock.debug.assert_called_once()
+                else:
+                    log_mock.debug.assert_not_called()
```
kwzrd, line 307: I don't like the amount of branching here too much.
```python
if error:
  ...
else:
  ...
if not error:
  ...
else:
  ...
```
Can these be joined together?

You marked @MarkKoz's comment above as resolved, but is it? Why are you only checking if `post` was awaited if it won't raise?

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,411 @@
...
+                embed = send_private_embed_mock.call_args[0][1]
+                self.assertEqual(embed.to_dict(), expected.to_dict())
+
+                send_private_embed_mock.assert_awaited_once_with(args[0], embed)
+
+    @patch("bot.cogs.moderation.utils.log")
+    async def test_post_user(self, log_mock):
+        """Should POST a new user and return the response if successful or otherwise send an error message."""
+        user = MockUser(discriminator=5678, id=1234, name="Test user")
+        some_mock = MagicMock(discriminator=3333)
```
kwzrd, line 238: I'm not finding `some_mock` to be a particularly good name, if it's meant to represent something that is *not* a user, that's a useful piece of information that can be included in the variable name.

You're only setting the `discriminator` attr but later on you're accessing `id` and `name` too - is that by design?

tests/bot/cogs/moderation/test_utils.py
```diff
@@ -0,0 +1,411 @@
...
+
+            with self.subTest(expected=expected, raised=raised):
+                self.user.send.reset_mock(side_effect=True)
+                self.user.send.side_effect = raised
+
+                result = await utils.send_private_embed(self.user, embed)
+
+                self.assertEqual(result, expected)
+                if expected:
+                    self.user.send.assert_awaited_once_with(embed=embed)
```
kwzrd, line 344: Why only check `if expected`? Isn't `send` being called unconditionally?

tests/bot/cogs/moderation/test_utils.py
kwzrd: Additionally I'm finding this fairly awkward:
```python
if isinstance(test_user, MagicMock):
```
I don't think it's necessary to test whether a `debug` log was made (if that doesn't indicate that some other action was also performed). But it can stay. I'm just wondering whether we could come up with a nicer way of branching here.

tests/bot/cogs/moderation/test_utils.py
ks129: Mock is awaited only when no errors raised.

tests/bot/cogs/moderation/test_utils.py
ks129: That was just for avoiding error. I added now id and name too.

tests/bot/cogs/moderation/test_utils.py
ks129: Removed `debug` call assertion. Joined these if statements together. About only asserting `post`, what else this does? I'm also asserting returned result, but this calls nothing else what I could assert.

tests/bot/cogs/moderation/test_utils.py
ks129: Synced all names.

tests/bot/cogs/moderation/test_utils.py
kwzrd: Is that so? Try removing the conditional - test passes just fine.

Correct me if I'm wrong, but with `self.user.send` being an `AsyncMock,` the side effect isn't actioned until after the `send` coro is awaited.

tests/bot/cogs/moderation/test_utils.py
ks129: I removed these when moved to `namedtuple`.

tests/bot/cogs/moderation/test_utils.py
ks129: `user` is now removed from the name.

tests/bot/cogs/moderation/test_utils.py
ks129: Used here `assertIn` instead and in one other location.

tests/bot/cogs/moderation/test_utils.py
ks129: I added `namedtuple` for some test cases, but for others, this will be such mess and no one will understand, what thing is what, especially these what have embeds/many keys.

bot/cogs/moderation/utils.py
```diff
@@ -28,6 +28,18 @@
...
+
+INFRACTION_TITLE = f"Please review our rules over at {RULES_URL}"
+INFRACTION_APPEAL_FOOTER = f"To appeal this infraction, send an e-mail to {APPEAL_EMAIL}"
+INFRACTION_AUTHOR_NAME = "Infraction information"
+
+INFRACTION_DESCRIPTION_TEMPLATE = (
+    "\n**Type:** {type}\n"
+    "**Expires:** {expires}\n"
+    "**Reason:** {reason}\n"
+)
```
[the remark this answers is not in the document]
ks129: I added the use case for this in the file.

tests/bot/cogs/moderation/test_utils.py
ks129: For me, the test doesn't pass when I remove conditional: this says that it's awaited 0 times.

bot/cogs/moderation/utils.py
kwzrd, line 41: The newlines will be replaced by a single empty space when passed through `textwrap.shorten`:

image

I think it's confusing to have the newlines here in the template if we know they won't be shown on Discord.

ks129: I think then we should shorten reason only. How much this should display of reason? 1000 characters?

kwzrd: Sounds like a good idea, although this would ideally be answered by someone who knows more about the character limits in embeds, afaik there's a limit on the total and a limit on the description too? Fwiw this isn't a regression, there's no newlines on the master branch either.

j03b: The [Discord.js guide](https://discordjs.guide/popular-topics/embeds.html#embed-limits) has some clear documentation on embed limits

ks129: Is this better now?

kwzrd:
So this PR now makes an attempt at fixing the regression reported in `#meta` recently:

image

The approach taken seems reasonable. The tests look good and we should move forward merging this soon, but since we added this change which affects a lot of code, I will need to test it properly and perhaps discuss with other core devs.

What is the motivation behind choosing `1000` as the limit?

ks129:
@kwzrd Because the limit is 2048, and 1000 is about half, so half is still available for other things.

tests/bot/cogs/moderation/test_utils.py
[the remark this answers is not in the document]
lemonsaurus: I think this is fine like this.

lemonsaurus (approved):
After what looks like a very intense review process, I think this is now looking quite good. Let's try to get it merged already, it's not good that our most important features are completely untested while duckpond has 8000 tests. @kwzrd would you be able to take a second look at this?

bot/cogs/moderation/utils.py
```diff
@@ -142,25 +154,23 @@ async def notify_infraction(
...
-    text = textwrap.dedent(f"""
-        **Type:** {infr_type.capitalize()}
-        **Expires:** {expires_at or "N/A"}
-        **Reason:** {reason or "No reason provided."}
-    """)
+    text = INFRACTION_DESCRIPTION_TEMPLATE.format(
+        type=infr_type.capitalize(),
+        expires=expires_at or "N/A",
+        reason=textwrap.shorten(reason, 1000, placeholder="...") if reason else "No reason provided."
+    )
```
kwzrd, line 161: @lemonsaurus @ks129 My feedback on this PR has been addressed but a somewhat out-of-scope change was made here that I'm not entirely excited about & @sco1 agreed that it's sub-optimal when we spoke about it some time back.

I requested only that the newlines are removed from the string because they are later on stripped by the `shorten`, and so the result string looks different than what the literal implies, which is misleading. @ks129 made a more adventurous attempt at keeping the newlines by only running the `reason` kwarg through the `shorten`, which helps in the sense that the newlines now remain, but it's only because we're no longer shortening the full string, but only one part of it, and taking an educated guess at best when choosing the length (`1000`) with the justification being that the other fields should never be long enough for this to exceed the maximum length. I'm finding this approach somewhat fragile and I think that a module that our entire moderation suite relies on deserves a more sophisticated approach.

ks129: I solved this issue by force-shortening (ignoring word breaks) in case when description will be still too long.

kwzrd: I think this approach is much better, but why are we still running the `reason` through `textwrap.shorten`? That shouldn't be necessary anymore, right?

ks129: I think this should be still there, because this try first soft-shortening (breaking between words) and when after this string is still too long, this force-break, what may break word.

kwzrd: That is true, but it still suffers from the problem of guessing at what the maximum length should be. I think that for the sake of simplicity and transparency, just the final cut-off would be fine (also easier to make assertions about in the tests). I don't personally see much appeal in keeping whole words if a sentence gets cut off anyway.

ks129: This is now done @kwzrd

kwzrd: Thanks, looks excellent. I will test tomorrow before I merge but I think this is now ready.

kwzrd (approved):
This all looks great to me now, and I think we can merge it. Would you be able to resolve the conflicts here, @ks129?

kwzrd:
FYI, I edited the PR description to cover the bugfix that we decided to include.

kwzrd:
Thanks for all the work you put into this!

PR #819, opened 2020-03-05
Add one second padding to format_infraction_with_duration

Akarys42:
It caused the message to be rounded one second down, for example `59 minutes and 59 seconds` instead of `1 hour`, because of the way `humanize_delta` works. This PR simply add a one second padding in order to make the infraction message proper.

### DIFF
--- a/bot/utils/time.py
+++ b/bot/utils/time.py
@@ -144,7 +144,7 @@ def format_infraction_with_duration(
     if absolute:
         delta = abs(delta)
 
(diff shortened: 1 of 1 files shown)


ikuyarihS:
This won't work if it is not just a second, sometimes it can be up to 5 seconds later, and so adding a simple second will not fix the issue.

It is better to check the actual time and try to round it to round it up if the seconds is **close enough** to 60.

Akarys42:
Closing this PR as the fix will take more work than that.

PR #821, opened 2020-03-06
Automatically Remove Users from BigBrother Watch List on Perma Ban

### LINKED ISSUE #321: No embed is send when a user gets "auto-unwatched" after a ban
When we ban a user that's currently being watched, they are automatically unwatched and Python is supposed to send an embed notifying us of that to the relevant watch channel. See [line 141 of the BigBrother cog](https://github.com/python-discord/bot/blob/master/bot/cogs/bigbrother.py#L141).

However, at the moment, no such embed is being send, although I do think the user is getting unwatched. We should probably look into that and fix it. While we're at it, we may want to consider adding an unwatch note to the user as well, just like we do when we unwatch someone manually.

At the moment, I don't know what's causing the bug. A quick review of the code revealed nothing obvious.

LordHemlock:
Previously we had the bot set up so that when a user was permanently banned it would be automatically unwatched by our "Big Brother" system.  This originally worked just fine, but the issue was that there was no embed being sent telling us about this.  However, after the big rewrite, that automated functionality was lost in the process.  This PR is intended to restore it.  

The discussed solution was to move most of the current `unwatch` command's functionality into a helper method and allow for a flag to be set when it's sent from a permanent ban.  The `watch` command will also have its functionality moved to a helper method for consistency and ease of testing.

Closes #321

### DIFF
--- a/bot/cogs/moderation/infractions.py
+++ b/bot/cogs/moderation/infractions.py
@@ -230,7 +230,11 @@ class Infractions(InfractionScheduler, commands.Cog):
 
     @respect_role_hierarchy()
     async def apply_ban(self, ctx: Context, user: UserSnowflake, reason: str, **kwargs) -> None:
-        """Apply a ban infraction with kwargs passed to `post_infraction`."""
+        """
+        Apply a ban infraction with kwargs passed to `post_infraction`.
+
+        Will also remove the banned user from the Big Brother watch list if applicable.
+        """
         if await utils.has_active_infraction(ctx, user, "ban"):
             return
 
@@ -243,7 +247,6 @@ class Infractions(InfractionScheduler, commands.Cog):
         action = ctx.guild.ban(user, reason=reason, delete_message_days=0)
         await self.apply_infraction(ctx, infraction, user, action)
 
-        # Remove perma banned users from the watch list
         if infraction.get('expires_at') is not None:
             log.trace(f"Ban isn't permanent; user {user} won't be unwatched by Big Brother.")
             return
@@ -256,9 +259,7 @@ class Infractions(InfractionScheduler, commands.Cog):
         log.trace(f"Big Brother cog loaded; attempting to unwatch perma-banned user {user}.")
 
         bb_reason = "User has been permanently banned from the server. Automatically removed."
-        await bb_cog.apply_unwatch(ctx, user, bb_reason banned=True)
-
-        log.debug(f"Perma-banned user {user} was unwatched.")
+        await bb_cog.apply_unwatch(ctx, user, bb_reason, send_message=False)
 
     # endregion
     # region: Base pardon functions
--- a/bot/cogs/moderation/infractions.py
+++ b/bot/cogs/moderation/infractions.py
@@ -244,18 +244,21 @@ class Infractions(InfractionScheduler, commands.Cog):
         await self.apply_infraction(ctx, infraction, user, action)
 
         # Remove perma banned users from the watch list
-        if infraction.get('expires_at') is None:
-            log.trace("Ban was a permanent one.  Attempt to remove from watched list.")
-            bb_cog = self.bot.get_cog("Big Brother")
-            if bb_cog:
-                log.trace("Cog loaded.  Attempting to remove from list.")
-                await bb_cog.apply_unwatch(
-                    ctx,
-                    user,
-                    "User has been permanently banned from the server.  Automatically removed.",
-                    banned=True
-                )
-                log.debug("Perma banned user removed from watch list.")
+        if infraction.get('expires_at') is not None:
+            log.trace(f"Ban isn't permanent; user {user} won't be unwatched by Big Brother.")
+            return
+
+        bb_cog = self.bot.get_cog("Big Brother")
+        if not bb_cog:
+            log.trace(f"Big Brother cog not loaded; perma-banned user {user} won't be unwatched.")
+            return
+
+        log.trace(f"Big Brother cog loaded; attempting to unwatch perma-banned user {user}.")
+
+        bb_reason = "User has been permanently banned from the server. Automatically removed."
+        await bb_cog.apply_unwatch(ctx, user, bb_reason banned=True)
+
+        log.debug(f"Perma-banned user {user} was unwatched.")
 
     # endregion
     # region: Base pardon functions
--- a/bot/cogs/moderation/infractions.py
+++ b/bot/cogs/moderation/infractions.py
@@ -67,7 +67,7 @@ class Infractions(InfractionScheduler, commands.Cog):
 
     @command()
     async def ban(self, ctx: Context, user: FetchedMember, *, reason: str = None) -> None:
-        """Permanently ban a user for the given reason.  Also removes them from the BigBrother watch list."""
+        """Permanently ban a user for the given reason and stop watching them with Big Brother."""
         await self.apply_ban(ctx, user, reason)
 
     # endregion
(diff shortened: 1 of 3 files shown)


bot/cogs/watchchannels/bigbrother.py
```diff
@@ -90,10 +100,8 @@ def __init__(self, bot: Bot) -> None:
 
         await ctx.send(msg)
 
-    @bigbrother_group.command(name='unwatch', aliases=('uw',))
-    @with_role(*MODERATION_ROLES)
-    async def unwatch_command(self, ctx: Context, user: FetchedMember, *, reason: str) -> None:
-        """Stop relaying messages by the given `user`."""
+    async def apply_unwatch(self, ctx: Context, user: FetchedMember, reason: str, banned: bool = False) -> None:
```
MarkKoz: I think a more generic name like `send_message` would be better. Regardless, this parameter should be documented.

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -52,6 +52,16 @@ def __init__(self, bot: Bot) -> None:
...
+        await self.apply_watch(ctx, user, reason)
+
+    @bigbrother_group.command(name='unwatch', aliases=('uw',))
+    @with_role(*MODERATION_ROLES)
+    async def unwatch_command(self, ctx: Context, user: FetchedMember, *, reason: str) -> None:
+        """Stop relaying messages by the given `user`."""
+        await self.apply_unwatch(ctx, user, reason)
+
+    async def apply_watch(self, ctx: Context, user: FetchedMember, reason: str) -> None:
+        """Handles adding a user to the watch list."""
```
MarkKoz: This doesn't do the function justice
```suggestion
        """
        Add `user` to watched users and apply a watch infraction with `reason`.

        A message indicating the result of the operation is sent to `ctx`.
        The message will include `user`'s previous watch infraction history, if it exists.
        """
```

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -90,10 +100,8 @@ def __init__(self, bot: Bot) -> None:
 
         await ctx.send(msg)
 
-    @bigbrother_group.command(name='unwatch', aliases=('uw',))
-    @with_role(*MODERATION_ROLES)
-    async def unwatch_command(self, ctx: Context, user: FetchedMember, *, reason: str) -> None:
-        """Stop relaying messages by the given `user`."""
+    async def apply_unwatch(self, ctx: Context, user: FetchedMember, reason: str, banned: bool = False) -> None:
+        """Handles the actual user removal from the watch list."""
```
MarkKoz: This assumes the parameter was renamed.
```suggestion
        """
        Remove `user` from watched users and mark their infraction as inactive with `reason`.

        If `send_message` is True, a message indicating the result of the operation is sent to
        `ctx`.
        """
```

bot/cogs/watchchannels/bigbrother.py
```diff
@@ -111,8 +120,13 @@ def __init__(self, bot: Bot) -> None:
...
+                log.trace("User is not banned.  Sending message to channel")
+                await ctx.send(f":white_check_mark: Messages sent by {user} will no longer be relayed.")
 
             self._remove_user(user.id)
         else:
-            await ctx.send(":x: The specified user is currently not being watched.")
+            log.trace("No active watches found for user.")
+            if not banned:  # Prevents a message being sent to the channel if part of a permanent ban
+                log.trace("User is not perma banned. Send the error message.")
+                await ctx.send(":x: The specified user is currently not being watched.")
```
MarkKoz: I think this will be cleaner if the message is assigned to a parameter and sent at the very end, similar to what the `apply_watch` function does.

bot/cogs/moderation/infractions.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, bot: Bot):
 
     @command()
     async def ban(self, ctx: Context, user: FetchedMember, *, reason: str = None) -> None:
-        """Permanently ban a user for the given reason."""
+        """Permanently ban a user for the given reason.  Also removes them from the BigBrother watch list."""
```
MarkKoz: I think this reads better as a single sentence and keeps the imperative mood.

```suggestion
        """Permanently ban a user for the given reason and stop watching them with Big Brother."""
```

bot/cogs/moderation/infractions.py
```diff
@@ -243,6 +243,20 @@ def __init__(self, bot: Bot):
...
+            bb_cog = self.bot.get_cog("Big Brother")
+            if bb_cog:
+                log.trace("Cog loaded.  Attempting to remove from list.")
+                await bb_cog.apply_unwatch(
+                    ctx,
+                    user,
+                    "User has been permanently banned from the server.  Automatically removed.",
+                    banned=True
+                )
+                log.debug("Perma banned user removed from watch list.")
```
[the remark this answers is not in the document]
MarkKoz: Oops, forgot to include a comma in the args for `apply_unwatch`. Feel free to amend the commit and force push it.

bot/cogs/watchchannels/bigbrother.py
LordHemlock: Addressed in most recent commit

bot/cogs/watchchannels/bigbrother.py
LordHemlock: Addressed in most recent commit

bot/cogs/moderation/infractions.py
```diff
@@ -243,6 +243,20 @@ def __init__(self, bot: Bot):
         action = ctx.guild.ban(user, reason=reason, delete_message_days=0)
         await self.apply_infraction(ctx, infraction, user, action)
 
+        # Remove perma banned users from the watch list
```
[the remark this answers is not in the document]
LordHemlock: Addressed in most recent commit

bot/cogs/moderation/infractions.py
```diff
@@ -243,6 +247,20 @@ def __init__(self, bot: Bot):
         action = ctx.guild.ban(user, reason=reason, delete_message_days=0)
         await self.apply_infraction(ctx, infraction, user, action)
 
+        if infraction.get('expires_at') is not None:
+            log.trace(f"Ban isn't permanent; user {user} won't be unwatched by Big Brother.")
+            return
+
+        bb_cog = self.bot.get_cog("Big Brother")
+        if not bb_cog:
+            log.trace(f"Big Brother cog not loaded; perma-banned user {user} won't be unwatched.")
```
sco1: While I understand that this is part of the trace logging flow, if this situation is encountered then this should be issued at the error level, not a trace.

MarkKoz: Why?

LordHemlock: Corrected in the most recent commit

sco1: Why wouldn't we want to be notified that the big brother cog isn't loaded?

LordHemlock: For the purpose of the trace logic it makes no difference if this is an error or a trace, but for the sake of knowing that something isn't quite right with one of our cogs, it makes total sense for it to be an error.  I just didn't think about it when writing it.

MarkKoz: If that's the case then IMO a warning is more appropriate. If there's ever a legitimate reason to unload it, even temporarily, then it'd be strange to be greeted with an error when nothing is actually wrong with the code. If you don't want it to be unloaded, consider adding it to the blacklist in the extensions cog.

LordHemlock: Fair point, a warning makes sense to me.  @sco1 that work for you?

PR #823, opened 2020-03-08
Subcommand `search` and `search any` to search inside tags' contents and not names.

### LINKED ISSUE #804: tag-search command to search tags via their contents instead of names.
Currently, it can be difficult to find the specific command for a tag even if you know it exists. A command to allow you to search through tag bodies would help with finding the right tags. For example doing `!tag-search backtick` would return `!code-block` (and any other tags that include the word backtick).

ikuyarihS:
#### Closes #804 

This PR introduces two new subcommands to `!tag` that will allow searching for tags via contents instead of names:

- `!tag search` will search for multiple keywords, separated by comma, and return tags that has ALL of these keywords.
- `!tag search any` is the same as `!tag search` but it return tags that has ANY of the keyword instead.

### Examples:
image

### DIFF
--- a/bot/cogs/tags.py
+++ b/bot/cogs/tags.py
@@ -1,7 +1,7 @@
 import logging
 import re
 import time
-from typing import Dict, List, Optional, Tuple
+from typing import Callable, Dict, Iterable, List, Optional, Tuple
 
 from discord import Colour, Embed
 from discord.ext.commands import Cog, Context, group
@@ -86,7 +86,7 @@ class Tags(Cog):
             return self._get_suggestions(tag_name)
         return found
 
-    async def _get_tags_via_content(self, check: callable, keywords: str) -> Optional[Embed]:
+    async def _get_tags_via_content(self, check: Callable[[Iterable], bool], keywords: str) -> Optional[Embed]:
         """
         Search for tags via contents.
 
@@ -124,9 +124,8 @@ class Tags(Cog):
         Only search for tags that has ALL the keywords.
         """
         result = await self._get_tags_via_content(all, keywords)
-        if not result:
-            return
-        await ctx.send(embed=result)
+        if result:
+            await ctx.send(embed=result)
 
     @search_tag_content.command(name='any')
     async def search_tag_content_any_keyword(self, ctx: Context, *, keywords: Optional[str] = None) -> None:
@@ -136,9 +135,8 @@ class Tags(Cog):
         Search for tags that has ANY of the keywords.
         """
         result = await self._get_tags_via_content(any, keywords or 'any')
-        if not result:
-            return
-        await ctx.send(embed=result)
+        if result:
+            await ctx.send(embed=result)
 
     @tags_group.command(name='get', aliases=('show', 'g'))
     async def get_command(self, ctx: Context, *, tag_name: TagNameConverter = None) -> None:
--- a/bot/cogs/tags.py
+++ b/bot/cogs/tags.py
@@ -94,7 +94,8 @@ class Tags(Cog):
         """
         await self._get_tags()
 
-        keywords_processed: Tuple[str] = tuple(query.strip().casefold() for query in keywords.split(','))
+        keywords_processed: Tuple[str] = tuple(query.strip().casefold() for query in keywords.split(',') if query)
+        keywords_processed = keywords_processed or (keywords,)
         founds: list = [
             tag
             for tag in self._cache.values()
@@ -106,10 +107,13 @@ class Tags(Cog):
         elif len(founds) == 1:
             return Embed().from_dict(founds[0]['embed'])
         else:
-            return Embed(
-                title='Did you mean ...',
+            is_plural: bool = len(keywords_processed) > 1 or any(kw.count(' ') for kw in keywords_processed)
+            embed = Embed(
+                title=f"Here are the tags containing the given keyword{'s' * is_plural}:",
                 description='\n'.join(tag['title'] for tag in founds[:10])
             )
+            embed.set_footer(text=f"Keyword{'s' * is_plural} used: {keywords}"[:1024])
+            return embed
 
     @group(name='tags', aliases=('tag', 't'), invoke_without_command=True)
     async def tags_group(self, ctx: Context, *, tag_name: TagNameConverter = None) -> None:
--- a/bot/cogs/tags.py
+++ b/bot/cogs/tags.py
@@ -86,7 +86,7 @@ class Tags(Cog):
             return self._get_suggestions(tag_name)
         return found
 
-    async def _get_tags_via_content(self, check: Callable[[Iterable], bool], keywords: str) -> Optional[Embed]:
+    async def _get_tags_via_content(self, check: Callable[[Iterable], bool], keywords: str) -> list:
         """
         Search for tags via contents.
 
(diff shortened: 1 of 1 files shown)


bot/cogs/tags.py
```diff
@@ -86,11 +86,60 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
...
+    async def search_tag_content_any_keyword(self, ctx: Context, *, keywords: Optional[str] = None) -> None:
+        """
+        Search inside tags' contents for tags. Allow searching for multiple keywords separated by comma.
+
+        Search for tags that has ANY of the keywords.
+        """
+        result = await self._get_tags_via_content(any, keywords or 'any')
+        if not result:
+            return
+        await ctx.send(embed=result)
```
MarkKoz:
```suggestion
        if result:
            await ctx.send(embed=result)
```

bot/cogs/tags.py
```diff
@@ -86,11 +86,60 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
...
+    async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
+        """
+        Search inside tags' contents for tags. Allow searching for multiple keywords separated by comma.
+
+        Only search for tags that has ALL the keywords.
+        """
+        result = await self._get_tags_via_content(all, keywords)
+        if not result:
+            return
+        await ctx.send(embed=result)
```
MarkKoz:
```suggestion
        if not result:
            await ctx.send(embed=result)
```

bot/cogs/tags.py
```diff
@@ -86,11 +86,60 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
             return self._get_suggestions(tag_name)
         return found
 
+    async def _get_tags_via_content(self, check: callable, keywords: str) -> Optional[Embed]:
```
MarkKoz: `Callable[[Iterable], bool]` would be a more accurate annotation.

bot/cogs/tags.py
ikuyarihS: Oh my, thank you!

bot/cogs/tags.py
ikuyarihS: Nice, thank you, I've never used `Callable` from `typing` before!

bot/cogs/tags.py
```diff
@@ -86,11 +86,58 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
...
+            if check(query in tag['embed']['description'] for query in keywords_processed)
+        ]
+
+        if not founds:
+            return None
+        elif len(founds) == 1:
+            return Embed().from_dict(founds[0]['embed'])
+        else:
+            return Embed(
+                title='Did you mean ...',
```
MarkKoz: This title is a bit misleading. It may be interpreted as a list of search _terms_ being suggested, rather than a list of search _results_ (tag names). It should be clear what is being listed are tag names results for the search.

MarkKoz: Or another way to look at it is that it may suggest that the command is meant to search the names of tags rather than their contents.

ikuyarihS: Hmm, that is a good catch, it does sound like the following *terms* are what you are looking for.
How does `Are you perhaps looking for the following tags?` sound?
Or maybe a simple `Are you looking for these tags?`?

ikuyarihS: `The following tags have {keywords} in their description, perhaps you are looking for one of these?` sounds pretty descriptive, but it might be too long, what do you think?

MarkKoz: That sounds OK. What do you think of  `Here are the tags containing the given keyword(s):`?

ikuyarihS: Yep, that definitely sounds better and shorter, thank you!

MarkKoz: `Are you looking for these tags?` was fine too. I don't think the actual keywords should be included as it may make the title too long.

bot/cogs/tags.py
```diff
@@ -86,11 +86,62 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
...
+        """
+        await self._get_tags()
+
+        keywords_processed: Tuple[str] = tuple(query.strip().casefold() for query in keywords.split(',') if query)
+        keywords_processed = keywords_processed or (keywords,)
+        founds: list = [
+            tag
+            for tag in self._cache.values()
+            if check(query in tag['embed']['description'] for query in keywords_processed)
+        ]
```
lemonsaurus: This whole block is incredibly dense. I have a number of objections to it and think you should simplify and refactor the entire block.

- Why are you using type annotations here? It's perfectly obvious that the first line is a tuple from the `tuple()` call, and that the last line is a list from the fact that it's wrapped in square brackets. Please only use inline type annotations in cases where they add clarity, such as surprising types and cases in which the reader would otherwise have a hard time knowing what the type of some variable is.

- Please don't make list comprehensions with this much complexity. If your list comprehension has more than two expressions in it, just refactor it into regular code. See Brett Slatkins excellent book **Effective Python**, Chapter 1 Item 8 - "Avoid More Than Two Expressions in List Comprehensions" for more information. Not only does this have three expressions in it, but you've had to split it up onto multiple lines to even fit it. If you can't fit your list comprehension on a single line, this is usually a sign that you should consider refactoring.

- The `keywords_processed` lines are quite dense. Think about all the stuff you're doing in two lines here. Let's write it down in plain English: 

  Split the `keyword` variable by comma. Create a tuple of the results this produces. Every result should be stripped, and also casefolded. However, if the result is false-y, do not put it in the tuple. Once this process is done, if this process produces no results at all, replace this tuple with a tuple containing only a single item; whatever is in `keywords`.

  Not only that, but you're using `typing.Tuple`, `tuple()` and `(keywords,)` all in two lines. Imagine how confusing this would be to a beginner. Please try to simplify this part into something a beginner wouldn't be so scared by.

- This is the most complex part of this entire pull request, but there are no comments. Use block comments to explain complexity if you can't refactor it into simpler code, and otherwise, use block comments as a sort of bookmark to describe the block below.

bot/cogs/tags.py
```diff
@@ -86,11 +86,62 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
...
+
+        if not founds:
```
lemonsaurus: `if founds` feels semantically weird. Maybe this variable can be renamed so that this line sounds more semantically correct. I would've used something like `matching_tags` or `matches`.

bot/cogs/tags.py
```diff
@@ -86,11 +86,62 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
...
+            return None
+        elif len(founds) == 1:
+            return Embed().from_dict(founds[0]['embed'])
+        else:
+            is_plural: bool = len(keywords_processed) > 1 or any(kw.count(' ') for kw in keywords_processed)
```
lemonsaurus: wait, can you explain the `or any(kw.count(' ') for kw in keywords_processed)` part of this for me? 

Also, this line is dense, too, and while I don't mind the type annotation here as much, I think the excellent variable name makes it intuitively clear that this is a boolean.

bot/cogs/tags.py
ikuyarihS: Thank you for very detailed reviews!

It does hit me now that you say "If you can't fit your list comprehension on a single line, this is usually a sign that you should consider refactoring." - it is a very good point, and I will definitely refactor it as well as the other dense part of the code.

bot/cogs/tags.py
ikuyarihS: Since I am allowing searching with multiple keywords, separated by comma at the same time, simply counting for `' '` will not work, for example if use search for `code,block` - there will be no space at all in the `keywords`

After thinking about this more, it is definitely over kill with my original intention to check if any `separated query` has any space in it, so the line can be reduced to
```
is_plural = len(keywords_processed) > 1 or keywords.strip().count(' ')
```

bot/cogs/tags.py
ikuyarihS: Since it was a simple `search and return` I decided to name it `founds` - which indeed hide away what it is unintentionally, when in reality it is a simple list of `tag`.
I really like your suggestion of `matching_tags`!

bot/cogs/tags.py
```diff
@@ -90,27 +90,37 @@ def _get_suggestions(self, tag_name: str, thresholds: Optional[List[int]] = None
...
+
+        if not matching_tags:
             return None
-        elif len(founds) == 1:
-            return Embed().from_dict(founds[0]['embed'])
+        elif len(matching_tags) == 1:
+            return Embed().from_dict(matching_tags[0]['embed'])
         else:
-            is_plural: bool = len(keywords_processed) > 1 or any(kw.count(' ') for kw in keywords_processed)
+            is_plural = len(keywords_processed) > 1 or keywords.strip().count(' ') > 1
```
MarkKoz: I think this should be 0, not 1. Otherwise, something like "word1 word2" will be singular and only at "word1 word2 word3" will it become plural.

```suggestion
            is_plural = len(keywords_processed) > 1 or keywords.strip().count(' ') > 0
```

ikuyarihS: Oh you are right, originally I wanted it to be just
```py
is_plural = len(keywords_processed) > 1 or keywords.strip().count(' ')
```
Thank you for the catch!

ikuyarihS:
You are right, the truncating was copied over from `!tags` where if multiple results are found, they are already sorted by matching scores, which is not the case for this.

I've added paginator for it just like in `!tags`

PR #828, opened 2020-03-14
Test for command name and alias shadowing

### LINKED ISSUE #765: Unit test for command alias shadowing
Currently, it is possible for someone to write a command that shadows an existing alias. Discord will not crash, it will simply short circuit and make the other alias unavailable.

It ought to be possible to add an automated check for this in the form of a unit test. This test might make use of `ast` (which has a means of checking decorator arguments) and then iterate over all cogs to determine if there are any alias duplicates.

This feature would probably be nice to have in seasonalbot, too, so once it's finished it might be nice to port it over there, too.

MarkKoz:
Resolves #765 

Adds a test to ensure command and alias names are unique. This allows for any names or aliases which shadow other commands to be caught in CI when tests run. The test will report all conflicts. It will _not_ stop at the first conflict encountered. For conflicts, the test will provide the following in the error message:

1. The command's fully qualified name or alias
2. The command's function's name qualified by its module's name
3. The qualified (similar to № 2) command function names of all commands with which the name or alias conflicts

The test only looks for commands which are defined a cog that is within the `bot.cogs` subpackage. Commands added dynamically via `add_command` will likely not be detected. Basically, this is only intended for commands defined with the `@commands.command()` decorator or group equivalent.

### DIFF
--- a/tests/bot/cogs/test_cogs.py
+++ b/tests/bot/cogs/test_cogs.py
@@ -46,7 +46,7 @@ class CommandNameTests(unittest.TestCase):
     @staticmethod
     def get_qualified_names(command: commands.Command) -> t.List[str]:
         """Return a list of all qualified names, including aliases, for the `command`."""
-        names = [f"{command.full_parent_name} {alias}" for alias in command.aliases]
+        names = [f"{command.full_parent_name} {alias}".strip() for alias in command.aliases]
         names.append(command.qualified_name)
 
         return names
--- a/tests/bot/cogs/test_cogs.py
+++ b/tests/bot/cogs/test_cogs.py
@@ -1,9 +1,14 @@
 """Test suite for general tests which apply to all cogs."""
 
+import importlib
+import pkgutil
 import typing as t
 import unittest
+from types import ModuleType
 
 from discord.ext import commands
+
+from bot import cogs
 
 
 class CommandNameTests(unittest.TestCase):
@@ -17,3 +22,9 @@ class CommandNameTests(unittest.TestCase):
                 yield command
                 if isinstance(command, commands.GroupMixin):
                     yield from command.walk_commands()
+
+    @staticmethod
+    def walk_extensions() -> t.Iterator[ModuleType]:
+        """Yield imported extensions (modules) from the bot.cogs subpackage."""
+        for module in pkgutil.iter_modules(cogs.__path__, "bot.cogs."):
+            yield importlib.import_module(module.name)
--- a/tests/bot/cogs/test_cogs.py
+++ b/tests/bot/cogs/test_cogs.py
@@ -6,6 +6,7 @@ import typing as t
 import unittest
 from collections import defaultdict
 from types import ModuleType
+from unittest import mock
 
 from discord.ext import commands
 
@@ -31,9 +32,10 @@ class CommandNameTests(unittest.TestCase):
         def on_error(name: str) -> t.NoReturn:
             raise ImportError(name=name)
 
-        for module in pkgutil.walk_packages(cogs.__path__, "bot.cogs.", onerror=on_error):
-            if not module.ispkg:
-                yield importlib.import_module(module.name)
+        with mock.patch("discord.ext.tasks.loop"):
+            for module in pkgutil.walk_packages(cogs.__path__, "bot.cogs.", onerror=on_error):
+                if not module.ispkg:
+                    yield importlib.import_module(module.name)
 
     @staticmethod
     def walk_cogs(module: ModuleType) -> t.Iterator[commands.Cog]:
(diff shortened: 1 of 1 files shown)


tests/bot/cogs/test_cogs.py
```diff
@@ -0,0 +1,77 @@
+"""Test suite for general tests which apply to all cogs."""
+
+import importlib
+import pkgutil
+import typing as t
```
eivl: this tripped me up a bit. is this something everyone does?

tests/bot/cogs/test_cogs.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @staticmethod
+    def walk_commands(cog: commands.Cog) -> t.Iterator[commands.Command]:
+        """An iterator that recursively walks through `cog`'s commands and subcommands."""
+        for command in cog.__cog_commands__:
+            if command.parent is None:
+                yield command
+                if isinstance(command, commands.GroupMixin):
+                    # Annoyingly it returns duplicates for each alias so use a set to fix that
+                    yield from set(command.walk_commands())
```
eivl: Damn this is great.

tests/bot/cogs/test_cogs.py
```diff
@@ -0,0 +1,77 @@
...
+        with mock.patch("discord.ext.tasks.loop"):
+            for module in pkgutil.walk_packages(cogs.__path__, "bot.cogs.", onerror=on_error):
+                if not module.ispkg:
+                    yield importlib.import_module(module.name)
+
+    @staticmethod
+    def walk_cogs(module: ModuleType) -> t.Iterator[commands.Cog]:
+        """Yield all cogs defined in an extension."""
+        for obj in module.__dict__.values():
+            is_cog = isinstance(obj, type) and issubclass(obj, commands.Cog)
```
eivl: is there any cases where `obj` is only an instance of commands.Cog and not type?

tests/bot/cogs/test_cogs.py
MarkKoz: No, not really. However, I think I stole this idea from some other code bases. Maybe it's a bad habit of mine. The benefit of this is reducing length of annotations and thus increasing readability. If annotations get too long, the function's arguments also have to be broken up, which makes it take up more space. Doing `from typing import x, y, z, a, b, c, d` can work if there are only a couple things but it eventually gets out of control.

tests/bot/cogs/test_cogs.py
```diff
@@ -0,0 +1,77 @@
...
+                    # Annoyingly it returns duplicates for each alias so use a set to fix that
+                    yield from set(command.walk_commands())
+
+    @staticmethod
+    def walk_modules() -> t.Iterator[ModuleType]:
+        """Yield imported modules from the bot.cogs subpackage."""
+        def on_error(name: str) -> t.NoReturn:
+            raise ImportError(name=name)
+
+        with mock.patch("discord.ext.tasks.loop"):
```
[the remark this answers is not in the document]
MarkKoz: Yes, that's exactly right. I believe I wrote about it in the commit message but didn't leave any comment in code.

tests/bot/cogs/test_cogs.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class CommandNameTests(unittest.TestCase):
+    """Tests for shadowing command names and aliases."""
+
+    @staticmethod
+    def walk_commands(cog: commands.Cog) -> t.Iterator[commands.Command]:
+        """An iterator that recursively walks through `cog`'s commands and subcommands."""
+        for command in cog.__cog_commands__:
+            if command.parent is None:
```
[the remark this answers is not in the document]
MarkKoz: Because that's an instance method.

tests/bot/cogs/test_cogs.py
MarkKoz: The former checks that it's a class. Without that check, these errors may occur:

```
TypeError: issubclass() arg 1 must be a class
```

MarkKoz:
> discord.py does raise an exception if there is a name conflict

Yes, but the bot would have to run for that to happen. This test does static checks.

> but the mechanism doesn't check cross-group which is a weakness that only manifests itself due to our bot.cogs.alias module.

Don't understand what you mean by cross-group exactly nor how the alias cog plays a role.

> If this is correct then this should perhaps be documented in the test's docstring

Document what exactly? I could write that the benefits of this test is that it's done statically if you think that will help.

> I'd probably approach this as a test specifically for the alias cog, rather than a test for all cogs.

There's no way to test that statically without doing the same thing being done currently to walk though cogs and get all commands.

kwzrd:
@MarkKoz Ok, I understand now. I was looking for a way in which a name collision could exist such that one command shadows the other, but `discord.py` doesn't raise. So I managed to convince myself that the `alias` cog somehow circumvents the check.

The collision cannot be detected during tests if the command never gets registered.

> Document what exactly?

Nothing, as I was wrong in my assumption.

PR #830, opened 2020-03-16
ModLog Channel Update formatting fix.

### LINKED ISSUE #793: Formatting issue for empty topic in channel update log
The old and new topics should each be in separate in-line code blocks. The arrow should be between and outside the two code blocks.

bild

ks129:
I added `None` string possibility to modlog channel update event to avoid bad MD formatting when one of values is empty. Closes #793

### DIFF
--- a/bot/cogs/moderation/modlog.py
+++ b/bot/cogs/moderation/modlog.py
@@ -215,7 +215,7 @@ class ModLog(Cog, name="ModLog"):
                 new = value["new_value"]
                 old = value["old_value"]
 
-                changes.append(f"**{key.title()}:** `{old}` **→** `{new}`")
+                changes.append(f"**{key.title()}:** `{old or 'None'}` **→** `{new or 'None'}`")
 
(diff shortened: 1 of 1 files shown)


ks129:
@eivl Here is main change:
Before:
image
After:
Screenshot 2020-03-16 at 16 48 29

This channel topic part especially.

bot/cogs/moderation/modlog.py
```diff
@@ -215,7 +215,7 @@ def ignore(self, event: Event, *items: int) -> None:
                 new = value["new_value"]
                 old = value["old_value"]
 
-                changes.append(f"**{key.title()}:** `{old}` **→** `{new}`")
+                changes.append(f"**{key.title()}:** `{old or 'None'}` **→** `{new or 'None'}`")
```
sco1: There should be a comment here explaining why this substitution is necessary to keep the formatting from breaking when an empty value is provided.

PR #832, opened 2020-03-16
Update token filter logging to match expanded detection

sco1:
While the token detection was successfully expanded by #825 to evaluate all potential token matches in the message, the logging method was not similarly updated & still utilized the `re.search` approach to get the matching substring for use in the log message.

Rather than searching for the token again, instead return the potential token substring from the detection method to be used later.


bot/cogs/token_remover.py
```diff
@@ -91,18 +94,21 @@ def mod_log(self) -> ModLog:
         )
 
     @classmethod
-    def is_token_in_message(cls, msg: Message) -> bool:
-        """Check if `msg` contains a seemly valid token."""
+    def find_token_in_message(cls, msg: Message) -> t.Optional[str]:
+        """Check for a seemingly valid token in the provided `Message` instance."""
```
MarkKoz:
```suggestion
        """Return a seemingly valid token found in `msg` or `None` if no token is found."""
```

PR #833, opened 2020-03-17
Fixed !zen command exact word matching.

### LINKED ISSUE #827: !zen command not working as intended
if you do
```
!zen explicit
```
it returns 
```
**The Zen of Python (line 10):**
Unless explicitly silenced.
```

while I do believe that the intent was to have it return line two
```
**The Zen of Python (line 1):**
Explicit is better than implicit.
```

ks129:
Closes #827 
Now this match exact word first, before trying to use `difflib`'s matching, due like in issue is shown, exact word may not best choice for matcher. 
Result for `!zen explicit`
Before (from issue):
```
**The Zen of Python (line 10):**
Unless explicitly silenced.
```
After:
Screenshot 2020-03-17 at 08 57 06


bot/cogs/utils.py
```diff
@@ -230,7 +230,16 @@ def check(m: Message) -> bool:
...
+        # Try to handle first exact word due difflib.SequenceMatched may use some other similar word instead
+        # exact word.
+        for i, line in enumerate(zen_lines):
+            if search_value.lower() in line.lower():
+                embed.title += f" (line {i}):"
+                embed.description = line
+                await ctx.send(embed=embed)
+                return
+
+        # handle if it's a search string and not exact word
```
SebastiaanZ: While this solution works for this particular case, it now picks the first line that contains the word as a substring. For example, if you swap the lines containing explicit and explicitly around, it will match the wrong one again. @sco1 has an alternative suggestion that matches based on words instead of an entire sentence to truly prefer full word matches.

PR #836, opened 2020-03-20
Tags cog unit tests and tags cog fixes

### LINKED ISSUE #589: Write unit tests for `bot/cogs/tags.py`
Write unit tests for [`bot/cogs/tags.py`](../blob/master/bot/cogs/tags.py).

## Implementation details

## Additional information

ks129:
Wrote unit tests for `Tags` cog + made some fixes there. Closes #589


tests/bot/cogs/test_tags.py
```diff
@@ -0,0 +1,305 @@
...
+            {
+                "args": ("my-test-string", None),
+                "expected": []
+            }
+        ]
+        for case in test_cases:
+            with self.subTest(args=case["args"], expected=case["expected"]):
+                actual = self.cog._get_suggestions(*case["args"])
+
+                self.assertEqual(actual, case["expected"])
```
sco1: Is there any reason to compare the full contents of the tag rather than just the tag name (dictionary key) in this test?

ks129:
I updated to match most of the changes in #866, but I need a bit of help to get one more percentage coverage. I don't know how to test these 2 small areas that are 1%.

ks129:
Now conflicts is fixed and 1 of missing coverage part is fixed. Second should be fixed, but I don't understand why this don't count. (Sorry about force-push, but I had to use it to restore some changes)

Senjan21 (approved):
Code and logic wise, all looks good! Thanks for the PR!

bot/exts/info/tags.py
```diff
@@ -145,8 +145,7 @@ async def _send_matching_tags(self, ctx: Context, keywords: str, matching_tags:
         else:
             is_plural = keywords.strip().count(' ') > 0 or keywords.strip().count(',') > 0
             embed = Embed(
-                title=f"Here are the tags containing the given keyword{'s' * is_plural}:",
-                description='\n'.join(tag['title'] for tag in matching_tags[:10])
+                title=f"Here are the tags containing the given keyword{'s' * is_plural}:"
```
kwzrd, line 148: You say that you've 'made some fixes there', could you expand on that? What is the reason for the diff here?

ks129: `description` field is ignored by the paginator.

kwzrd: Okay, thanks, that makes sense. I'll try to do a full review soon.

bot/exts/info/tags.py
```diff
@@ -160,7 +159,8 @@ async def _send_matching_tags(self, ctx: Context, keywords: str, matching_tags:
     @group(name='tags', aliases=('tag', 't'), invoke_without_command=True)
     async def tags_group(self, ctx: Context, *, tag_name: TagNameConverter = None) -> None:
         """Show all known tags, a single tag, or run a subcommand."""
-        await self.get_command(ctx, tag_name=tag_name)
+        # Use invoke here for testability purposes
+        await ctx.invoke(self.get_command, tag_name=tag_name)
```
kwzrd, line 163: I'm not sure what the difference between these two is, and why one is easier to test than the other (based on the commit message that adds this diff). Would you be able to expand on that please?

tests/bot/exts/info/test_tags.py
```diff
@@ -0,0 +1,430 @@
...
+                "expected": Embed.from_dict({
+                    "description": (
+                        "\n**»**   class\n"
+                        "**»**   classmethod\n"
+                        "**»**   ytdl\n"
+                    ),
+                    "footer": {"text": tags.FOOTER_TEXT},
+                    "title": "Here are the tags containing the given keyword:"
+                })
+            }
```
kwzrd, line 251: I'm guessing `from_dict` is used because you cannot pass the footer text to the embed constructor. An alternative approach would be to build the embed before the list, though this is probably ok.

kwzrd (commented):
I think overall the tests look ok. I probably wouldn't use dictionaries to define test cases (there's always the option of just using tuples and unpacking the items in the subtest loop). I think some of the docstrings could be improved both in terms of grammar and content. Keep in mind that the docstring summary is shown to the user if the test fails, so it's fairly important that it manages to explain what exactly is not working as expected. Having the test break and fail to explain why is frustrating and counter-productive, because then another developer has to spend effort trying to figure out what went wrong. If it's difficult to capture the test's purpose in the docstring, it may be an indication that the test should be broken down into multiple smaller test functions.

tests/bot/exts/info/test_tags.py
ToxicKidz: `Embed(...).set_footer(text=tags.FOOTER_TEXT)` would work too since [`Embed.set_footer`](https://discordpy.readthedocs.io/en/latest/api.html#discord.Embed.set_footer) returns the instance.

Xithrius:
@ks129 will you be able to finish up this PR?

ks129:
@kwzrd Issues should be solved now and I also improved docstrings slightly.

Xithrius:
Due to https://github.com/python-discord/bot/pull/1663 becoming implemented along with other PRs expanding the domain of tags, the tests here are unfortunately going to be continuously changed.

We're thinking of focusing on tests which are more isolated such as utilities, so that the testing of said features aren't needed to be changed often.

Thank you to all who worked on this PR.

PR #845, opened 2020-03-26
Logging Level Adjustments

sco1:
With the updated definition on logging levels (see: #837), there are a few events that are potentially issuing logs at too high of a level, alerting the DevOps team for things that aren't actionable, or at too low of a level.

While some instances are obvious adjustments, I've opened this as a draft for feedback on the changes being made & also to ask questions on non-obvious adjustments.

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -77,7 +77,7 @@ class Bot(commands.Bot):
 
         # Its __del__ does send a warning but it doesn't always show up for some reason.
         if self._connector and not self._connector._closed:
-            log.info(
+            log.warning(
                 "The previous connector was not closed; it will remain open and be overwritten"
             )
 
@@ -94,7 +94,7 @@ class Bot(commands.Bot):
 
         # Its __del__ does send a warning but it doesn't always show up for some reason.
         if self.http_session and not self.http_session.closed:
-            log.info(
+            log.warning(
                 "The previous session was not closed; it will remain open and be overwritten"
             )
 
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -77,7 +77,7 @@ class Bot(commands.Bot):
 
         # Its __del__ does send a warning but it doesn't always show up for some reason.
         if self._connector and not self._connector._closed:
-            log.warning(
+            log.info(
                 "The previous connector was not closed; it will remain open and be overwritten"
             )
 
@@ -94,7 +94,7 @@ class Bot(commands.Bot):
 
         # Its __del__ does send a warning but it doesn't always show up for some reason.
         if self.http_session and not self.http_session.closed:
-            log.warning(
+            log.info(
                 "The previous session was not closed; it will remain open and be overwritten"
             )
 
--- a/bot/cogs/moderation/superstarify.py
+++ b/bot/cogs/moderation/superstarify.py
@@ -59,7 +59,7 @@ class Superstarify(InfractionScheduler, Cog):
             return  # Nick change was triggered by this event. Ignore.
 
         log.info(
-            f"{after.display_name} is currently in superstar-prison. "
+            f"{after.display_name} ({after.id}) tried to escape superstar prison. "
             f"Changing the nick back to {before.display_name}."
         )
         await after.edit(
@@ -80,7 +80,7 @@ class Superstarify(InfractionScheduler, Cog):
         )
 
         if not notified:
-            log.warning("Failed to DM user about why they cannot change their nickname.")
+            log.info("Failed to DM user about why they cannot change their nickname.")
 
     @Cog.listener()
     async def on_member_join(self, member: Member) -> None:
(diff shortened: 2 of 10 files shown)


sco1:
Some initial questions on downgrading from warning to info:

https://github.com/python-discord/bot/blob/ee7cfbfca1b23408d7cb3f603498347fcef00c86/bot/cogs/sync/cog.py#L37-L44

I'm assuming this is still something we want to be aware of, in case the race condition becomes a recurring issue?

https://github.com/python-discord/bot/blob/ee7cfbfca1b23408d7cb3f603498347fcef00c86/bot/cogs/sync/syncers.py#L133-L138

Since I believe this is being done as part of an interactive session with a staff member, does it need to separately emit a warning level log?

https://github.com/python-discord/bot/blob/ee7cfbfca1b23408d7cb3f603498347fcef00c86/bot/utils/messages.py#L94-L95

I'm assuming this is fine as an info log. It appears that we'd only get to this condition from the antispam cog, where we're intentionally skipping the upload to prevent duplicating the spam.

bot/bot.py
```diff
@@ -77,7 +77,7 @@ def _recreate(self) -> None:
 
         # Its __del__ does send a warning but it doesn't always show up for some reason.
         if self._connector and not self._connector._closed:
-            log.warning(
+            log.info(
```
MarkKoz: While the consequences of the situation are not that big, I believe it should remain warning because this would happen due to programmer error (i.e. not manually closing before clearing/logging in).

bot/bot.py
```diff
@@ -94,7 +94,7 @@ def _recreate(self) -> None:
 
         # Its __del__ does send a warning but it doesn't always show up for some reason.
         if self.http_session and not self.http_session.closed:
-            log.warning(
+            log.info(
```
MarkKoz: While the consequences of the situation are not that big, I believe it should remain warning because this would happen due to programmer error (i.e. not manually closing before clearing/logging in).

MarkKoz:
> I'm assuming this is still something we want to be aware of, in case the race condition becomes a recurring issue?

I don't quite get this warning. A 404 would mean the user is not found in the database. The only way I can see that happening is if a user is edited the moment they join the server, and the PATCH for the edit goes through before the user is POSTed to the database. If that's the case, then, if left as a warning, it would remind us to come up with a better solution. On the other hand, it may be safe to ignore since a future sync will eventually update the user, though I'm not sure what implications being temporarily out of sync has for other features which rely on user info from the database. my opinion is to leave it a warning.

> Since I believe this is being done as part of an interactive session with a staff member, does it need to separately emit a warning level log?

It could either be from a manual invocation of the sync command or from an automatic sync when the bot starts. Semantically it makes sense to be a warning, but it's not something we'd want to be notified about in Sentry. Of most concern would be when the confirmation of an automatic sync times out; manual invocations could be a lower log level.

> I'm assuming this is fine as an info log. It appears that we'd only get to this condition from the antispam cog, where we're intentionally skipping the upload to prevent duplicating the spam.

This is also used by Big Brother. Semantically, I think a warning makes sense but it's not something we need to be notified of in Sentry since there's simply nothing we can do about a file being too large.

sco1:
> I suppose it all has to be an info level to avoid sentry bugging us?

I'm not particularly concerned about Sentry being verbose as long as it's telling us about things that require action. In the case of the syncer reaction loop, staff has already been pinged in the same context that the warning would be issued, so unless I'm missing something this is essentially pinging twice.

> This is also used by Big Brother.
I meant the `else` clause specifically. From what I can see, other than the antispam call of this method, all other current calls to this method utilize the default `link_large = True`. which is handled by the previous `elif` condition.

MarkKoz:
> so unless I'm missing something this is essentially pinging twice.

Yes, it would. For that reason, it should be moved down to info. If it weren't for Sentry then I'd have kept this as a warning.

> In the case of the antispam call, reaching the else clause is intentional and not something that we need to be reminded of.

OK, makes sense. I'm fine with an info for it then.

bot/cogs/sync/syncers.py
```diff
@@ -131,7 +131,7 @@ def _reaction_check(
             await message.edit(content=f':ok_hand: {mention}{self.name} sync will proceed.')
             return True
         else:
-            log.warning(f"The {self.name} syncer was aborted or timed out!")
+            log.trace(f"The {self.name} syncer was aborted or timed out!")
```
MarkKoz: Well I still think it should be visible in production logs so I disagree with it being anything lower than `info`. Someone looking at logs may not necessarily be looking at Discord to see the edited message.

bot/cogs/moderation/superstarify.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, bot: Bot):
             return  # Nick change was triggered by this event. Ignore.
 
         log.info(
-            f"{after.display_name} is currently in superstar-prison. "
+            f"{after.display_name} ({after.id}) tried to escape superstar prison. "
```
lemonsaurus: hahaha this is great

PR #846, opened 2020-03-27
Filtering: merge the word and token watch filters

### LINKED ISSUE #322: Watchlist Tokens & Watchlist Words merge and bug
Tokens and Words watchlist filters really should be just be combined I think. The name of `watch_tokens` being shown in mod alerts makes one think of bot tokens, not a character/word list. If there's any particular reason why it was done like this in the beginning though, feel free to let us know.

There's also a bug where if the `retarded` expression gets triggered but it matches one of it's exceptions such as "sorry for all the retarded questions", it'll return `False` right there and then, failing to process any other expressions after it. So technically "what's with the retarded question you dumb cunt" may not trigger the filter since `(re+)tar+(d+|t+)(ed)? questions?` is an exception and `cunts*` comes after it in the list of expressions to be processed.

This can be resolved by replacing `return False` with `continue`:
```diff
                    # stuff like "sorry for all the retarded questions"
                    elif re.search(RETARDED_QUESTIONS_RE, text, re.IGNORECASE):
-                       return False
+                       continue
```

As a side note, this is going to make for an interesting github history check if someone looks it over in the future.

MarkKoz:
Closes #322 

The only difference was the automatic addition of word boundaries. Otherwise, they shared a lot of code. The regex lists were kept separate in the config to retain the convenience of word boundaries automatically being added.

* Rename filter to `watch_regex`
* Expand spoilers for both words and tokens
* Ignore URLs for both words and tokens

### DIFF
--- a/bot/cogs/filtering.py
+++ b/bot/cogs/filtering.py
@@ -38,6 +38,7 @@ WORD_WATCHLIST_PATTERNS = [
 TOKEN_WATCHLIST_PATTERNS = [
     re.compile(fr'{expression}', flags=re.IGNORECASE) for expression in Filter.token_watchlist
 ]
+WATCHLIST_PATTERNS = WORD_WATCHLIST_PATTERNS + TOKEN_WATCHLIST_PATTERNS
 
 
 def expand_spoilers(text: str) -> str:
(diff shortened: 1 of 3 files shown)


MarkKoz:
No, the word boundaries are still added automatically.

PR #847, opened 2020-03-28
Added `Command.can_run` check to !aliases command.

### LINKED ISSUE #844: Limit aliases output to only allowed commands.
## Description
We currently have a system for aliases to be able to be output for informative reasons with the `!aliases` command. It'll show a full paginated list of all command aliases, including any commands that aren't able to be used by the invoker, meaning it's just providing unnecessary information for them that they can't really utilise. 

## Proposal
Just the same as when doing help output, check each command that has an alias if the author can run it, allowing only those they can run to be built into the command output.

ks129:
Closes #844

### DIFF
--- a/bot/cogs/alias.py
+++ b/bot/cogs/alias.py
@@ -35,7 +35,12 @@ class Alias (Cog):
         await ctx.invoke(cmd, *args, **kwargs)
 
     async def check_can_run_alias(self, ctx: Context, alias: str) -> bool:
-        """Check can user use this specific alias."""
+        """
+        Check can user use this specific alias.
+
+        Function force allow `get group` alias, due there is no command `get`: it's only to allow
+        use `get tags|docs`, `get` itself is not command and group is basically argument (`tags` or `docs`).
+        """
         # We don't want arguments, only command itself and max command parts in aliases is 2.
         if len(alias.split()) >= 3:
             alias = " ".join(alias.split()[0:2])
--- a/bot/cogs/alias.py
+++ b/bot/cogs/alias.py
@@ -34,6 +34,14 @@ class Alias (Cog):
 
         await ctx.invoke(cmd, *args, **kwargs)
 
+    async def test_alias_can_run(self, ctx: Context, alias: str) -> bool:
+        """Check can user use this specific alias."""
+        # We don't want arguments, only command itself and max command parts in aliases is 2.
+        if len(alias.split()) >= 3:
+            alias = " ".join(alias.split()[0:2])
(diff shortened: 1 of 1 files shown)


bot/cogs/alias.py
```diff
@@ -34,19 +34,32 @@ def __init__(self, bot: Bot):
 
         await ctx.invoke(cmd, *args, **kwargs)
 
+    async def test_alias_can_run(self, ctx: Context, alias: str) -> bool:
```
eivl, line 37: There probably are some `test_` prefixes scattered around in non-test code. But I personally use the `test_` prefix for only test cases. 
My suggestion is that we change the function name to something that does not use the `test_` prefix.

bot/cogs/alias.py
```diff
@@ -34,19 +34,32 @@ def __init__(self, bot: Bot):
...
+        """Check can user use this specific alias."""
+        # We don't want arguments, only command itself and max command parts in aliases is 2.
+        if len(alias.split()) >= 3:
+            alias = " ".join(alias.split()[0:2])
+        # Force allow `get group` due it's not actual command.
+        return alias == "get group" or await self.bot.get_command(alias).can_run(ctx)
```
eivl, line 43: I'm not aware of what `get group` is, this is a fault of my own, but maybe there should be a comment about this in the docstring?
I'm not saying it should be changed, it's more for a discussion than anything else.

eivl: also, returning both bools is pretty smart. I like it.

bot/cogs/alias.py
```diff
@@ -34,19 +34,32 @@ def __init__(self, bot: Bot):
...
         embed = Embed(
             title='Configured aliases',
             colour=Colour.blue()
         )
+        aliases = {
+            name: value for name, value in inspect.getmembers(self)
+            if isinstance(value, Command)
+            and name.endswith("_alias")
+            and await self.test_alias_can_run(ctx, name[:-len('_alias')].replace('_', ' '))
+        }
```
eivl, line 57: This is really clever.

bot/cogs/alias.py
ks129: `get group` is basically `get <group>`, where group can be `tags`, `docs`. Really `tags group` command doesn't exist, just `group` is in the alias function name. I'll add this info to docstring.

ks129:
@eivl Addressed review.

MarkKoz:
Closing because the cog was removed in #1159/#1198

PR #848, opened 2020-03-28
Hard-coded PEP 0 to !pep command

### LINKED ISSUE #805: Pep command cannot fetch PEP 0.
The [PEP 0](https://www.python.org/dev/peps/) is an index of all the other PEPs;
instead of being in the github repo directly, it's generated automatically through a script present there.
~~If implemented, scraping the site for it seems to be the better option as generating it would require the whole up to date repo.~~
The header is a constant in the pep0 package and can be gotten from there in case of future changes 
 or as suggested in the comment below https://github.com/python/peps/blob/master/pep0/constants.py#L10-L20

ks129:
Closes #805

### DIFF
--- a/bot/cogs/utils.py
+++ b/bot/cogs/utils.py
@@ -40,14 +40,6 @@ If the implementation is easy to explain, it may be a good idea.
 Namespaces are one honking great idea -- let's do more of those!
 """
 
-PEP0_TITLE = "Index of Python Enhancement Proposals (PEPs)"
-PEP0_INFO = {
-    "Status": "Active",
-    "Created": "13-Jul-2000",
-    "Type": "Informational"
-}
-PEP0_LINK = "https://www.python.org/dev/peps/"
-
 ICON_URL = "https://www.python.org/static/opengraph-icon-200x200.png"
 
 
@@ -72,16 +64,7 @@ class Utils(Cog):
         # Handle PEP 0 directly due it's static constant in PEPs GitHub repo in Python file, not .rst or .txt so it
         # can't be accessed like other PEPs.
         if pep_number == 0:
-            pep_embed = Embed(
-                title=f"**PEP 0 - {PEP0_TITLE}**",
-                description=f"[Link]({PEP0_LINK})"
-            )
-            pep_embed.set_thumbnail(url=ICON_URL)
-            for field, value in PEP0_INFO.items():
-                pep_embed.add_field(name=field, value=value)
-
-            await ctx.send(embed=pep_embed)
-            return
+            return await self.send_pep_zero(ctx)
 
         possible_extensions = ['.txt', '.rst']
         found_pep = False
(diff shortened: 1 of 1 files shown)


bot/cogs/utils.py
```diff
@@ -59,6 +67,19 @@ def __init__(self, bot: Bot):
             await ctx.invoke(self.bot.get_command("help"), "pep")
             return
 
+        # Handle PEP 0 directly due it's not available like other PEPs (use constants)
```
sco1: This explanation could be a lot better. What is different for PEP 0 vs. the other PEPs that makes the existing logic unusable?

bot/cogs/utils.py
```diff
@@ -40,6 +40,14 @@
 Namespaces are one honking great idea -- let's do more of those!
 """
 
+PEP0_TITLE = "Index of Python Enhancement Proposals (PEPs)"
+PEP0_INFO = {
+    "Status": "Active",
+    "Created": "13-Jul-2000",
+    "Type": "Informational"
+}
+PEP0_LINK = "https://www.python.org/dev/peps/"
```
MarkKoz: Adding constants feels like clutter when these are unlikely to change and aren't used in more than one place. I suggest you just use the values directly in code.

bot/cogs/utils.py
```diff
@@ -59,6 +67,20 @@ def __init__(self, bot: Bot):
...
+            pep_embed = Embed(
+                title=f"**PEP 0 - {PEP0_TITLE}**",
+                description=f"[Link]({PEP0_LINK})"
+            )
+            pep_embed.set_thumbnail(url="https://www.python.org/static/opengraph-icon-200x200.png")
+            for field, value in PEP0_INFO.items():
+                pep_embed.add_field(name=field, value=value)
+
+            await ctx.send(embed=pep_embed)
+            return
```
MarkKoz: Let's make this a separate function so that the command isn't so monolithic.

bot/cogs/utils.py
```diff
@@ -59,6 +67,20 @@ def __init__(self, bot: Bot):
...
             return
 
+        # Handle PEP 0 directly due it's static constant in PEPs GitHub repo in Python file, not .rst or .txt so it
+        # can't be accessed like other PEPs.
+        if pep_number == 0:
+            pep_embed = Embed(
+                title=f"**PEP 0 - {PEP0_TITLE}**",
+                description=f"[Link]({PEP0_LINK})"
+            )
+            pep_embed.set_thumbnail(url="https://www.python.org/static/opengraph-icon-200x200.png")
```
MarkKoz: This URL should be a constant since it's now used in two spots.

bot/cogs/utils.py
```diff
@@ -59,6 +67,20 @@ def __init__(self, bot: Bot):
             await ctx.invoke(self.bot.get_command("help"), "pep")
             return
 
+        # Handle PEP 0 directly due it's static constant in PEPs GitHub repo in Python file, not .rst or .txt so it
+        # can't be accessed like other PEPs.
```
MarkKoz: I find "static constant" to be unclear. What does that mean? I think it'll be less confusing if you just leave it out and say the reason is that the PEP doesn't have an associated rst or txt. Also, "due" was not used correctly here - "because" would have been a better fit.

PR #851, opened 2020-03-29
Added whitelist command

### LINKED ISSUE #639: `!whitelist` command to whitelist stuff from our filters.
The bot should be able to whitelist stuff from our filters with a command.

This should probably be a command group that allows whitelisting of several different things that we're currently configuring in default-config.yml.

- `!whitelist guild 15409531` should whitelist a guild invite for posting on our server. Currently this is managed in the `guild_invite_whitelist` section in default-config.yml
- `!whitelist format .ipybn` should whitelist a format from the anti-malware filter. Currently managed in the `anti_malware.whitelist` section of the `default-config.yml`

This issue will require a corresponding change to the `site`. You can find the corresponding issue for this at https://github.com/python-discord/site/issues/305

atmishra:
Closes #639 

This PR adds whitelist command to manage whitelists in the backend DB. Here is how I have designed the command. This is the first time I am working on creating a new command group, so feel free to add your inputs. 

### General Syntax 
Screenshot from 2020-03-29 14-42-55

### Listing whitelisted items of a particular type
image (2)

### Adding a new item to whitelist
image

### Deleting an item from whitelist
image (3)


bot/cogs/whitelist.py
```diff
@@ -0,0 +1,113 @@
...
+            whitelist_item.strip(),
+        )
+        existing_whitelist = await self._get_whitelisted_items(whitelist_type)
+        if whitelist_item in existing_whitelist:
+            await ctx.send(
+                embed=Embed(
+                    description=f"**{whitelist_item} already exists in the {whitelist_type} whitelist.**",
+                    colour=Colour.red(),
+                )
+            )
```
MarkKoz: IMO it'd be better to rely on the unique constraint failing and detecting that error.

bot/cogs/whitelist.py
```diff
@@ -0,0 +1,113 @@
...
+                f"whitelist_type: {whitelist_type}\n"
+                f"whitelist_item: '{whitelist_item}'\n"
+            )
+
+            await ctx.send(
+                embed=Embed(
+                    title="Item successfully added",
+                    description=f"**{whitelist_item}** added to {whitelist_type} whitelist database.",
+                    colour=Colour.blurple(),
+                )
```
MarkKoz: As I said elsewhere, don't use embeds.

bot/cogs/whitelist.py
```diff
@@ -0,0 +1,113 @@
...
+            )
+        else:
+            body = {"type": whitelist_type, "whitelisted_item": whitelist_item}
+            await self.bot.api_client.post("bot/whitelist", json=body)
+
+            log.debug(
+                f"{ctx.author} successfully added the following items to our database: \n"
+                f"whitelist_type: {whitelist_type}\n"
+                f"whitelist_item: '{whitelist_item}'\n"
+            )
```
MarkKoz: Avoid using newlines in logs because it makes them harder to read. This should also have an info log level.

```suggestion
            log.info(f"{ctx.author} added '{whitelist_item}' to the {whitelist_type} whitelist.")
```

bot/cogs/whitelist.py
```diff
@@ -0,0 +1,113 @@
...
+    async def delete_command(self, ctx: Context, id_: int) -> None:
+        """Remove Item from whitelist."""
+        await self.bot.api_client.delete(f"bot/whitelist/{id_}")
+
+        log.debug(
+            f"{ctx.author} successfully removed the item with id {id_} from whitelist."
+        )
+        await ctx.send(
+            embed=Embed(
+                description=f"Item has been removed successfully!",
```
MarkKoz: The message should be more explicit, like "Removed whitelist item with ID ..."

bot/cogs/whitelist.py
```diff
@@ -0,0 +1,113 @@
...
+    @with_role(Roles.owners, Roles.admins)
+    async def get_command(self, ctx: Context, *, whitelist_type: str) -> None:
+        """Get a list of whitelisted items under whitelist_type."""
+        params = {"type": whitelist_type}
+        res_json = await self.bot.api_client.get(f"bot/whitelist", params=params)
+        whitelisted_items = [
+            f'{item["whitelisted_item"]} (id: {item["id"]})' for item in res_json
+        ]
+        embed: Embed = Embed(
+            title=f"**Current Whitelisted items under {whitelist_type}**"
```
MarkKoz:
```suggestion
            title=f"**Current items in the {whitelist_type} whitelist**"
```

bot/cogs/whitelist.py
```diff
@@ -0,0 +1,113 @@
...
+            await ctx.send(
+                embed=Embed(
+                    title="Item successfully added",
+                    description=f"**{whitelist_item}** added to {whitelist_type} whitelist database.",
+                    colour=Colour.blurple(),
+                )
+            )
+
+    @whitelist_group.command(name="get", aliases=("show", "g"))
+    @with_role(Roles.owners, Roles.admins)
```
MarkKoz: I think it'd make sense to at least let mods (and maybe helpers?) see what the whitelist contains. After all, this is possible with the current system since the config is public in the repo. Will need input from other staff on this.

atmishra:
Hi, @MarkKoz Thanks! for the detailed review. I need clarity on a few things:

1. Even if we split set command to separate commands for each type, how does it improves the invocation? For invocation, we still have to use the same syntax `!whitelist invite ...`
2. How can I infer the type based on the input, as there some type which takes free form text like the extensions type? For supporting the mentions, my plan is to use the regex to take out the ids from the mentions. Is this the right approach?
3. I liked the idea of making the delete command more user friendly. As the delete command will be the same for all the types, how do I distinguish between a mention or a normal text for looking up in the DB? Also, in case of ambiguity I need to take id from the user to delete the item. How do I support both the normal text input and id input in the delete command?

MarkKoz:
> Even if we split set command to separate commands for each type, how does it improves the invocation?

It has the advantage of verifying that types are valid. If the type is an argument, you'd have to write a custom converter to validate. They will also show up in the list of subcommands in the help command. I don't have a strong feeling for it either way but I do think I prefer them being separate for that reason. 

It also allows for each command to use a separate converter for the value. If it was a single command, the type conversion would have to be done inside the function, which I think is a bit messier. On the other hand, it may avoid having boilerplate code for each command, which makes the system more tedious to expand with more types in the future.

> How can I infer the type based on the input, as there some type which takes free form text like the extensions type?

Consider that roles, channels, and invites all can be converted reliably. This leaves file extensions as the only plain text value really. But to avoid accidentally interpreting typos as file extensions, I think we should just force users to invoke the more explicit form of the command i.e. `!whitelist extension ...` rather than supporting type inference for file extensions.

> For supporting the mentions, my plan is to use the regex to take out the ids from the mentions. Is this the right approach?

Yes, I think that is a better approach - to create custom converters for type inference. Discord.py has converters for these but the problem with them is that they support more than just mentions. They can match by name too, which may be conflicting. Supporting only mentions for type inference is the only reliable way to guarantee the user meant to use a certain type.

For the more explicit commands, you will already know the type so you can use the normal discord.py converters to support more inputs.

> how do I distinguish between a mention or a normal text for looking up in the DB?

Yeah, this is an oversight on my part. As I mentioned earlier, the extension type cannot be safely inferred so perhaps the solution is to have subcommands like the command for adding whitelist items. I'm open to other ideas though.

> How do I support both the normal text input and id input in the delete command?

How about a subcommand for specifically deleting by ID?

lemonsaurus:
We need this done, and this PR is extremely stale. I'm closing it.

PR #852, opened 2020-03-29
Don't let change expiration time when infraction already expired.

### LINKED ISSUE #850: Disallow editing duration of expired infractions
Currently the command assumes that an infraction is still active if a duration is being edited. It tries to cancel the previous infraction but it will fail with a warning if the infraction already expired. Relevant code can be found here:
https://github.com/python-discord/bot/blob/582ddbb1ca8bab2cb883781911f5f35962330995/bot/cogs/moderation/management.py#L130-L142

ks129:
Closes #850

### DIFF
--- a/bot/cogs/moderation/management.py
+++ b/bot/cogs/moderation/management.py
@@ -102,9 +102,7 @@ class ModManagement(commands.Cog):
 
         if duration is not None and not old_infraction['active']:
             if reason is None:
-                await ctx.send(
-                    "Expiry can't be changed (infraction already expired) and new reason not specified."
-                )
+                await ctx.send(":x: Cannot edit the expiration of an expired infraction.")
                 return
             confirm_messages.append("expiry unchanged (infraction already expired)")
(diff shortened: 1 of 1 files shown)


ks129:
@MarkKoz I made this early exiting when an infraction expired and no reason provided.

bot/cogs/moderation/management.py
```diff
@@ -100,7 +100,14 @@ def infractions_cog(self) -> Infractions:
         confirm_messages = []
         log_text = ""
 
-        if isinstance(duration, str):
+        if duration is not None and not old_infraction['active']:
+            if reason is None:
+                await ctx.send(
+                    "Expiry can't be changed (infraction already expired) and new reason not specified."
+                )
```
MarkKoz:
```suggestion
                await ctx.send(":x: Cannot edit the expiration of an expired infraction.")
```

ks129:
@MarkKoz I applied change now.

PR #854, opened 2020-03-30
Tags response + command message deletion with trashcan reaction

### LINKED ISSUE #745: Add ability to remove bot tag responses.
With the new fuzzy tags, there's a chance an user may think something which is not a tag is one; resulting in a similar tag being fetched 
(`!t def` as an example resulting in a tag through the whole page)

Adding the ability to remove them (perhaps when it was fuzzy matched) would solve this and perhaps some others when a tag's purpose is misremembered.

ks129:
Closes #745

### DIFF
--- a/bot/cogs/tags.py
+++ b/bot/cogs/tags.py
@@ -1,14 +1,16 @@
 import logging
 import re
 import time
+from asyncio import TimeoutError
 from pathlib import Path
 from typing import Callable, Dict, Iterable, List, Optional
 
-from discord import Colour, Embed
+from discord import Colour, Embed, Message, Reaction, User
 from discord.ext.commands import Cog, Context, group
 
 from bot import constants
 from bot.bot import Bot
+from bot.constants import Emojis
 from bot.converters import TagNameConverter
 from bot.pagination import LinePaginator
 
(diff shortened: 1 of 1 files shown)


bot/cogs/tags.py
```diff
@@ -1,14 +1,16 @@
 import logging
 import re
 import time
+from asyncio import TimeoutError
```
ikuyarihS: I suggest changing this to
```py
import asyncio
```
then you can do
```py
except asyncio.TimeoutError
```
To avoid shadowing the base `TimeoutError`.

bot/cogs/tags.py
```diff
@@ -139,6 +141,24 @@ def _get_tags_via_content(self, check: Callable[[Iterable], bool], keywords: str
...
+        """Add `trashcan` emoji to Tag and handle deletion when user react to it."""
+        await msg.add_reaction(Emojis.trashcan)
+
+        def check_trashcan(reaction: Reaction, user: User) -> bool:
+            return (
+                reaction.emoji == Emojis.trashcan
+                and user == ctx.author
+                and reaction.message.id == msg.id
+            )
+        try:
```
ikuyarihS: There should be a blank line between the function and the `try:`

bot/cogs/tags.py
```diff
@@ -205,12 +225,14 @@ def _command_on_cooldown(tag_name: str) -> bool:
                         "time": time.time(),
                         "channel": ctx.channel.id
                     }
-                await ctx.send(embed=Embed.from_dict(tag['embed']))
+                msg = await ctx.send(embed=Embed.from_dict(tag['embed']))
+                await self.handle_trashcan_react(ctx, msg)
```
ikuyarihS: Since object `msg` is not used anywhere else, you can combine the `send` and the `handle_trashcan_react()` together, and rename it to something like `send_with_thrashcan()` or something, that way we do not have to create a variable outside. Think of a wrapper on top of `ctx.send()` and we simply pass what we want to send to it.

ks129:
@ikuyarihS Made fixes for these things.

bot/cogs/tags.py
```diff
@@ -139,6 +141,26 @@ def _get_tags_via_content(self, check: Callable[[Iterable], bool], keywords: str
                 max_lines=15
             )
 
+    async def send_embed_with_trashcan(self, ctx: Context, embed: Embed) -> None:
```
MarkKoz: A function for this already exists [here](https://github.com/python-discord/bot/blob/master/bot/utils/messages.py#L16).

bot/cogs/tags.py
```diff
@@ -205,12 +207,22 @@ def _command_on_cooldown(tag_name: str) -> bool:
                         "time": time.time(),
                         "channel": ctx.channel.id
                     }
-                await ctx.send(embed=Embed.from_dict(tag['embed']))
+                await wait_for_deletion(
```
sco1: Do we really want to delete successful tag hits?

ks129: I think this may help in a situation when a user accidentally uses a command in wrong channel. And this is just a small reaction under a message, nothing big and breaking so I think this should okay.

PR #857, opened 2020-03-31
Create generators.md tag

decorator-factory:
from https://github.com/python-discord/meta/pull/35

### DIFF
--- a/bot/resources/tags/generators.md
+++ b/bot/resources/tags/generators.md
@@ -1,0 +1,32 @@
+# Python Generators
+
+Often you'll need to return a list of values from a function.
+
+Imagine how you would implement a function that behaves like `map`,
+that is, it applies a function to each element in a list.:
+```python
+def my_map(function, old_list):
+    new_list = []
+    for element in old_list:
+        new_list.append(function(element))
+    return new_list
+
+for i in my_map(abs, [-1, -2, 3, -4]):
+    print(i)
+# 1, 2, 3, 4
+```
+However, there are two main issues with this approach:
+
+1. Initializing a list, appending to it and returning it is noisy boilerplate code.
+
+2. You will have to compute the result (which might even be infinite) first, and this can take a lot of time and memory.
+
+All of this can be solved with a _generator_:
+```python
+def my_map(function, old_list):
+    for element in old_list:
+        yield function(element)
+```
+A generator can produce new elements on demand without storing them in a list.
+
+You can try to run the loop in the example, and it gives the same result.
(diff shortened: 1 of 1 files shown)


bot/resources/tags/generators.md
```diff
@@ -0,0 +1,32 @@
+# Python Generators
```
MarkKoz: Discord doesn't support markdown headers. The best it has is bold.

```suggestion
**Python Generators**
```

bot/resources/tags/generators.md
```diff
@@ -0,0 +1,32 @@
...
+
+Often you'll need to return a list of values from a function.
+
+Imagine how you would implement a function that behaves like `map`,
+that is, it applies a function to each element in a list.:
```
MarkKoz:
```suggestion
that is, it applies a function to each element in a list:
```

bot/resources/tags/generators.md
```diff
@@ -0,0 +1,32 @@
...
+        new_list.append(function(element))
+    return new_list
+
+for i in my_map(abs, [-1, -2, 3, -4]):
+    print(i)
+# 1, 2, 3, 4
+```
+However, there are two main issues with this approach:
+
+1. Initializing a list, appending to it and returning it is noisy boilerplate code.
```
MarkKoz: Serial comma!
```suggestion
1. Initializing a list, appending to it, and returning it is noisy boilerplate code.
```

bot/resources/tags/generators.md
```diff
@@ -0,0 +1,32 @@
...
+
+2. You will have to compute the result (which might even be infinite) first, and this can take a lot of time and memory.
```
MarkKoz:
```suggestion
2. You will have to compute the result (which might even be infinite) first. This can take a lot of time and memory.
```

bot/resources/tags/generators.md
```diff
@@ -0,0 +1,32 @@
...
+
+All of this can be solved with a _generator_:
+```python
+def my_map(function, old_list):
+    for element in old_list:
+        yield function(element)
+```
+A generator can produce new elements on demand without storing them in a list.
```
MarkKoz: I think something along these lines should be mentioned at the beginning of the tag. This means it will get to the point right away and then be able to elaborate on it.

decorator-factory: Added in the latest suggestion.

bot/resources/tags/generators.md
```diff
@@ -1,5 +1,10 @@
 **Python Generators**
 
+A generator is a function that returns a sequence of values,
+producing them on demand, i.e. without storing them in a list.
+
+---
```
MarkKoz, line 6: Discord doesn't render these either but it still kind of works as a separator I guess.

PR #858, opened 2020-03-31
Add mutability.md tag

decorator-factory:


bot/resources/tags/mutability.md
```diff
@@ -0,0 +1,34 @@
...
+around existing strings or create new ones.
+
+```python
+string = "abcd"
+string = string.upper()
+```
+`string.upper()` creates a new string which is like the old one, but with all
+the letters turned to upper case.
+
+`int`, `float`, `complex`, `tuple`, `frozenset`  are other examples of immutable data types in Python.
```
MarkKoz, line 25:
```suggestion
`int`, `float`, `complex`, `tuple`, `frozenset` are other examples of immutable data types in Python.
```

decorator-factory: Fixed hard-wrapping.

bot/resources/tags/mutability.md
```diff
@@ -0,0 +1,32 @@
...
+`string` didn't change. Why is that so?
+
+That's because strings in Python are _immutable_. You can't change them, you can only pass around existing strings or create new oness.
+
+```python
+string = "abcd"
+string = string.upper()
+```
+
+`string.upper()` creates a new string which is like the old one, but with allthe letters turned to upper case.
```
kwzrd, line 21:
```suggestion
`string.upper()` creates and returns a new string which is like the old one, but with all the letters turned to upper case.
```

bot/resources/tags/mutability.md
```diff
@@ -0,0 +1,32 @@
...
+You might think that this would work:
+```python
+string = "abcd"
+string.upper()
+print(string) # abcd
+```
+
+`string` didn't change. Why is that so?
+
+That's because strings in Python are _immutable_. You can't change them, you can only pass around existing strings or create new oness.
```
kwzrd, line 14:
```suggestion
That's because strings in Python are _immutable_. You can't change them, you can only pass around existing strings or create new ones.
```

bot/resources/tags/mutability.md
```diff
@@ -0,0 +1,32 @@
+**Mutable vs immutable objects**
+
+Imagine that you want to make all letters in a string upper case. Conveniently, strings have an `.upper()` method.
+
+You might think that this would work:
+```python
+string = "abcd"
+string.upper()
+print(string) # abcd
```
kwzrd, line 9: Would it perhaps be better to show an interactive session here?
```python
>>> string = "abcd"
>>> string.upper()
'ABCD'
>>> string
'abcd'
```
This demonstrates that the new string is returned, and needs to be assigned to `string`.

bot/resources/tags/mutability.md
```diff
@@ -0,0 +1,32 @@
...
+`int`, `float`, `complex`, `tuple`, `frozenset` are other examples of immutable data types in Python.
+
+Mutable data types like `list`, on the other hand, can be changed in-place:
+```python
+my_list = [1, 2, 3]
+my_list.append(4)
+print(my_list) # [1, 2, 3, 4]
+```
+
+`dict` and `set` are other examples of mutable data types in Python.
```
kwzrd, line 32:
```suggestion
Other examples of mutable data types in Python are `dict` and `set`.
```
Just a minor nit-pick, but I don't like that this sentence is a copy of the one on line 23. Reversing it like this would remedy that, but it's not necessary.

bot/resources/tags/mutability.md
ikuyarihS: Also, should `string` be renamed to `text` instead, to avoid shadowing the `string` module?

kwzrd: I agree, this would be less confusing.

decorator-factory: changed

PR #859, opened 2020-04-01
Add the generators tag

decorator-factory:


MarkKoz:
The mutability tag is also included for some reason. Have you messed up some git operation?

PR #864, opened 2020-04-04
Ban and kick reason length fix

### LINKED ISSUE #861: Ban command failed with reason has more than 512 letters, but still add to DB
## What's wrong?
Right now, when issuing a ban ( `!tempban` for example ) if the reason is greater than 512, it will fail because discord only allow 512 letters.
However, the bot will still add the infraction to our DB and register it as an active ban, and thus subsequent ban will fail, saying that user is already banned.

## Expected behaviour:
When the reason is greater than 512, either send a warning / confirmation that the reason is greater than 512 and so reason will be cut off to be used for `.ban()`.

Or at the very least, `.ban()` should be wrapped inside a `try` `except` and will actually show the error.

## Suggestion
The order of actions should be revised: ban first, then post to DB only if it succeeded.

The code can be found here:
https://github.com/python-discord/bot/blob/44b33d570642905d4eeb64300a3a2d3d0cb80b72/bot/cogs/moderation/infractions.py#L231-L262

ks129:
Closes #861 

### Description
This PR main direction is to shorten reason in places, where is limit of characters. Main places is: Kick and Ban Audit Log entries, user notification and ModLog entries. This also limits ModLog entry description length. Added limits to watching too.

### Still to-do

- [x] Add test for characters limits
- [x] Remove DB entries when applying infraction fail.


bot/cogs/moderation/infractions.py
```diff
@@ -225,7 +225,10 @@ def __init__(self, bot: Bot):
 
         self.mod_log.ignore(Event.member_remove, user.id)
 
-        action = user.kick(reason=reason)
+        if len(reason) > 512:
+            log.info("Kick reason is longer than 512 characters. Reason will be truncated for Audit Log.")
+
+        action = user.kick(reason=f"{reason[:509]}..." if len(reason) > 512 else reason)
```
ikuyarihS, line 231: This can add a `...` to the middle of a word, like "See you ne..." - You will want to truncate at whitespace, `textwrap` will be of use here.

bot/cogs/moderation/infractions.py
```diff
@@ -244,7 +247,14 @@ def __init__(self, bot: Bot):
...
 
-        action = ctx.guild.ban(user, reason=reason, delete_message_days=0)
+        if len(reason) > 512:
+            log.info("Ban reason is longer than 512 characters. Reason will be truncated for Audit Log.")
+
+        action = ctx.guild.ban(
+            user,
+            reason=f"{reason[:509]}..." if len(reason) > 512 else reason,
+            delete_message_days=0
+        )
```
ikuyarihS, line 257: Should we put this in a try except and move it before we call `utils.post_infraction()`? It will be rare, but there can be situation where the actual discord ban / kick failed, but since we've already put the infraction in the DB the bot will think that the ban is active.

ks129: Should this after coroutine making or after `apply_infraction`?

ks129: Also, I can't move to add into DB to end of the function: `infraction` is required for applying infraction, and in coroutine creation, this will not raise an error. An error will be raised on `self.apply_infraction`, but this need DB returned infraction.

bot/cogs/moderation/infractions.py
```diff
@@ -225,7 +226,10 @@ def __init__(self, bot: Bot):
 
         self.mod_log.ignore(Event.member_remove, user.id)
 
-        action = user.kick(reason=reason)
+        if len(reason) > 512:
+            log.info("Kick reason is longer than 512 characters. Reason will be truncated for Audit Log.")
+
+        action = user.kick(reason=textwrap.shorten(reason, width=509, placeholder="..."))
```
ikuyarihS, line 232: The `width` will ensure the new string will never be greater than the limit, **after appending placeholder** so you can change it to this

```suggestion
        action = user.kick(reason=textwrap.shorten(reason, width=512, placeholder=" ..."))
```

I suggest adding a space for `placeholder` as well.

You can do the same for `ban` and it'll look good to go.

bot/cogs/moderation/infractions.py
```diff
@@ -225,7 +226,10 @@ def __init__(self, bot: Bot):
 
         self.mod_log.ignore(Event.member_remove, user.id)
 
-        action = user.kick(reason=reason)
+        if len(reason) > 512:
+            log.info("Kick reason is longer than 512 characters. Reason will be truncated for Audit Log.")
```
MarkKoz, line 230: I don't think these logs are necessary.

bot/cogs/moderation/infractions.py
```diff
@@ -244,7 +248,14 @@ def __init__(self, bot: Bot):
 
         self.mod_log.ignore(Event.member_remove, user.id)
 
-        action = ctx.guild.ban(user, reason=reason, delete_message_days=0)
+        if len(reason) > 512:
+            log.info("Ban reason is longer than 512 characters. Reason will be truncated for Audit Log.")
```
MarkKoz, line 252: I don't think these logs are necessary.

bot/cogs/moderation/infractions.py
```diff
@@ -244,7 +248,14 @@ def __init__(self, bot: Bot):
...
+
+        action = ctx.guild.ban(
+            user,
+            reason=textwrap.shorten(reason, width=512, placeholder=" ..."),
+            delete_message_days=0
+        )
```
MarkKoz, line 258: If you put the reason in a separate variable above you could avoid having to break up the function arguments for `ban()`.

bot/cogs/moderation/scheduler.py
```diff
@@ -84,7 +84,8 @@ def mod_log(self) -> ModLog:
         """Apply an infraction to the user, log the infraction, and optionally notify the user."""
         infr_type = infraction["type"]
         icon = utils.INFRACTION_ICONS[infr_type][0]
-        reason = infraction["reason"]
+        # Truncate reason when it's too long to avoid raising error on sending ModLog entry
+        reason = textwrap.shorten(infraction["reason"], width=1900, placeholder="...")
```
MarkKoz, line 88: Isn't 1900 cutting it a bit too close? That's only a bit over 100 characters for everything else.

sco1: Why is being done here? The truncation should be happening in the methods that are sending the messages, e.g. `utils.notify_infraction` and `mod_log.send_log_message`. This way you can use the actual character limit instead of having to guess at how long it needs to be in order to not raise an error.

ks129: @sco1 I applied this to `utils.notify_infraction`, but I can't do this in `mod_log.send_log_message`, because this take argument `text`, where is the full content of embed description (that include reason).

ks129:
@MarkKoz Now this should be fixed.

bot/cogs/moderation/scheduler.py
sco1: Yes, that’s the point. You’re focusing on truncating only the reason, arbitrarily guessing at the length you need to truncate to, when you can just truncate the entire string that these methods generate before sending.

ks129: @sco1 But like when reason is too long, this will cut `Expiry` (`apply_infraction`) from ModLog entry, because this is after reason. To make this, I have to put everything else first and after this reason.

sco1: Yes.

ks129:
@sco1 I now made changes, but problem is, I can't change `pardon_infraction`. This buildup and log text generating don't allow this way.

sco1:
Why do you need to change `pardon_infraction`?

bot/cogs/moderation/utils.py
```diff
@@ -132,11 +132,11 @@
     log.trace(f"Sending {user} a DM about their {infr_type} infraction.")
 
     embed = discord.Embed(
-        description=textwrap.dedent(f"""
+        description=textwrap.shorten(textwrap.dedent(f"""
             **Type:** {infr_type.capitalize()}
             **Expires:** {expires_at or "N/A"}
             **Reason:** {reason or "No reason provided."}
-            """),
+            """), width=2048, placeholder="..."),
```
MarkKoz, line 139: This is hard to read. Split them up by assigning the dedent to a variable above.

bot/cogs/moderation/scheduler.py
```diff
@@ -127,7 +128,7 @@ def mod_log(self) -> ModLog:
                 f"Infraction #{id_} actor is bot; including the reason in the confirmation message."
             )
 
-            end_msg = f" (reason: {infraction['reason']})"
+            end_msg = f" (reason: {textwrap.shorten(infraction['reason'], width=1500, placeholder='...')})"
```
MarkKoz, line 131:
```suggestion
            end_msg = f" (reason: {textwrap.shorten(reason, width=1500, placeholder='...')})"
```

bot/cogs/moderation/scheduler.py
```diff
@@ -325,7 +326,7 @@ def mod_log(self) -> ModLog:
         log_text = {
             "Member": f"<@{user_id}>",
             "Actor": str(self.bot.get_user(actor) or actor),
-            "Reason": infraction["reason"],
+            "Reason": textwrap.shorten(infraction["reason"], width=1500, placeholder="..."),
```
MarkKoz, line 329: This is not ideal but I recognise this case is more troublesome than others. Do you think there is a simple way to ensure the reason will be at the end given this gets merged with another dict may have additional keys added by `pardon_infraction`? I suspect it will involve having to delete the reason from the dict then add it back in. Maybe just not worth the trouble.

ks129: Maybe should this overwrite `Reason` in `pardon_infraction` to put this as the last key? And after this, `textwrap.shorten` whole ModLog embed description?

ks129:
@MarkKoz Now should OK.

bot/cogs/watchchannels/talentpool.py
```diff
@@ -224,7 +224,7 @@ def _nomination_to_string(self, nomination_object: dict) -> str:
                 Status: **Active**
                 Date: {start_date}
                 Actor: {actor.mention if actor else actor_id}
-                Reason: {nomination_object["reason"]}
+                Reason: {textwrap.shorten(nomination_object["reason"], width=200, placeholder="...")}
```
MarkKoz, line 227: Did you confirm that these need to be shortened and that the paginator doesn't help?

ks129: Yes, because this takes it as one line.

MarkKoz:
All the truncation seems to be fine though.

lemonsaurus:
@ks129 Can you please update this pull request with a proper description? It's not acceptable to just write "Closing #861" as your pull request description - the description is supposed to provide a high level overview over specifically what this pull request does.

I'd also like to suggest that this PR address two more related and fairly minor issues:
- If the ban _does_ fail, for some other reason, it would be nice if we could reset the state properly so we don't get an erroneous state. This includes removing the record from our database and deleting the DM that was sent to the member so they don't get a false infraction notice.
- We should add a test that tests a ban with a ban reason of over 512 characters.

ks129:
@lemonsaurus I fixed PR description. About tests, there are currently no tests for this moderation side.

lemonsaurus:
@ks129 looks like you're right, I thought we had more tests for moderation. We do have the tests in here https://github.com/python-discord/bot/blob/master/tests/bot/cogs/moderation/test_silence.py but not for most of the other moderator commands.

It would still be nice to see a single test, specifically for this problem. Would you be able to write that? No need to write a complete suite of tests for the entire moderation cog, but we got to start somewhere, right?

ks129:
@lemonsaurus So I should write test only for these functions that have shortening?

lemonsaurus:
doesn't this just apply to bans?

ks129:
@lemonsaurus I added truncation to ban, kick and ModLog entries. Should I write to test to every of them?

lemonsaurus:
Yeah, that seems fine to me. They can all be in the same class, though. You're basically just testing truncation.

also, we should probably get tests up and running for moderation features. That's gotta be one of the most important cogs we have. It's funny that we have extensive testing for Duck Pond but we don't have any tests for our most crucial commands.

ks129:
@lemonsaurus I added now removing infraction from DB when applying fail and moved DM sending to later, so this check is this failed before sending it. Also made tests, but ban and kick reason shortening is in one file, ModLog's one is in another file.

tests/bot/cogs/moderation/test_infractions.py
```diff
@@ -0,0 +1,54 @@
+import textwrap
+import unittest
+from unittest.mock import AsyncMock, Mock, patch
+
+from bot.cogs.moderation.infractions import Infractions
+from tests.helpers import MockBot, MockContext, MockGuild, MockMember, MockRole
+
+
+class ShorteningTests(unittest.IsolatedAsyncioTestCase):
```
lemonsaurus, line 9: Let's call it Truncation, not Shortening. Do you mind changing all the occurrances in this file?

tests/bot/cogs/moderation/test_modlog.py
```diff
@@ -0,0 +1,29 @@
...
+
+class ModLogTests(unittest.IsolatedAsyncioTestCase):
+    """Tests for moderation logs."""
+
+    def setUp(self):
+        self.bot = MockBot()
+        self.cog = ModLog(self.bot)
+        self.channel = MockTextChannel()
+
+    async def test_log_entry_description_shortening(self):
```
lemonsaurus, line 17: this one too.

bot/cogs/moderation/modlog.py
```diff
@@ -98,7 +98,10 @@ async def send_log_message(
         footer: t.Optional[str] = None,
     ) -> Context:
         """Generate log embed and send to logging channel."""
-        embed = discord.Embed(description=text)
+        # Truncate string directly here to avoid removing newlines
+        embed = discord.Embed(
+            description=text[:2046] + "..." if len(text) > 2048 else text
```
thomaspet, line 103: This will create a 2049 length string, if it's above 2048

tests/bot/cogs/moderation/test_infractions.py
```diff
@@ -0,0 +1,55 @@
...
+
+        self.cog.apply_infraction = AsyncMock()
+        self.bot.get_cog.return_value = AsyncMock()
+        self.cog.mod_log.ignore = Mock()
+
+        await self.cog.apply_ban(self.ctx, self.target, "foo bar" * 3000)
+        ban = self.cog.apply_infraction.call_args[0][3]
+        self.assertEqual(
+            ban.cr_frame.f_locals["kwargs"]["reason"],
+            textwrap.shorten("foo bar" * 3000, 512, placeholder=" ...")
```
lemonsaurus, line 35: Is there supposed to be a space before the `...` here? Doesn't seem like you're doing that anywhere else. If so, why?

tests/bot/cogs/moderation/test_infractions.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+        # Await ban to avoid warning
+        await ban
```
lemonsaurus, line 38: Can you be more specific? What warning does this raise, and how does this avoid it? This code seems weird to me.

tests/bot/cogs/moderation/test_modlog.py
```diff
@@ -0,0 +1,29 @@
...
+class ModLogTests(unittest.IsolatedAsyncioTestCase):
+    """Tests for moderation logs."""
+
+    def setUp(self):
+        self.bot = MockBot()
+        self.cog = ModLog(self.bot)
+        self.channel = MockTextChannel()
+
+    async def test_log_entry_description_truncation(self):
+        """Should truncate embed description for ModLog entry."""
```
lemonsaurus, line 18: Should be written in imperial mood.

```suggestion
        """Test that embed description for ModLog entry is truncated."""
```

bot/cogs/moderation/modlog.py
lemonsaurus: Yeah, I'm a little suspicious of these hardcoded values too. Can you justify them? Is there a test for this particular part of the code? Have you tested this manually?

bot/cogs/moderation/infractions.py
```diff
@@ -244,7 +245,9 @@ async def apply_ban(self, ctx: Context, user: UserSnowflake, reason: str, **kwar
 
         self.mod_log.ignore(Event.member_remove, user.id)
 
-        action = ctx.guild.ban(user, reason=reason, delete_message_days=0)
+        truncated_reason = textwrap.shorten(reason, width=512, placeholder=" ...")
```
lemonsaurus, line 248: Here's another placeholder with a leading space.

bot/cogs/moderation/modlog.py
ks129: I will fix this length issue, but I can't change this. Otherwise, we are getting [this issue](https://github.com/python-discord/bot/pull/864#pullrequestreview-390849141). This already tests it [here](https://github.com/ks129/bot/blob/ban-kick-reason-length/tests/bot/cogs/moderation/test_modlog.py#L17), but I have same issue there too.

lemonsaurus: That's fine, just make sure the values are correct. It's mostly this length issue I was concerned about.

tests/bot/cogs/moderation/test_infractions.py
```diff
@@ -0,0 +1,55 @@
...
+        self.cog.mod_log.ignore = Mock()
+
+        await self.cog.apply_ban(self.ctx, self.target, "foo bar" * 3000)
+        ban = self.cog.apply_infraction.call_args[0][3]
+        self.assertEqual(
+            ban.cr_frame.f_locals["kwargs"]["reason"],
+            textwrap.shorten("foo bar" * 3000, 512, placeholder="...")
+        )
+        # Await ban to avoid not awaited coroutine warning
+        await ban
```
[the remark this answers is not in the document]
ks129: Addressed in 96b0261. This is a big simplification.

bot/cogs/moderation/scheduler.py
```diff
@@ -164,11 +149,37 @@ async def apply_infraction(
...
+                # Accordingly display whether the user was successfully notified via DM.
+                if await utils.notify_infraction(user, infr_type, expiry, reason, icon):
+                    dm_result = ":incoming_envelope: "
+                    dm_log_text = "\nDM: Sent"
+
+        if failed:
+            dm_log_text = "\nDM: **Canceled**"
+            dm_result = f"{constants.Emojis.failmail} "
+            log.trace(f"Deleted infraction {infraction['id']} from database because applying infraction failed.")
+            await self.bot.api_client.delete(f"bot/infractions/{infraction['id']}")
```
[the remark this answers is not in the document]
ks129: Did that in f42ddf6, but is this enough and good?

bot/cogs/moderation/scheduler.py
```diff
@@ -164,11 +149,37 @@ async def apply_infraction(
...
 
         # Send a confirmation message to the invoking context.
         log.trace(f"Sending infraction #{id_} confirmation message.")
         await ctx.send(
-            f"{dm_result}{confirm_msg} **{infr_type}** to {user.mention}{expiry_msg}{end_msg}."
+            f"{dm_result}{confirm_msg} "
+            f"{f'**{infr_type}** to {user.mention}{expiry_msg}{end_msg}' if not failed else ''}."
```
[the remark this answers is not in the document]
ks129: Addressed in 7f827ab.

bot/cogs/moderation/scheduler.py
```diff
@@ -84,6 +84,7 @@ async def apply_infraction(
         """Apply an infraction to the user, log the infraction, and optionally notify the user."""
         infr_type = infraction["type"]
         icon = utils.INFRACTION_ICONS[infr_type][0]
+        # Truncate reason when it's too long to avoid raising error on sending ModLog entry
```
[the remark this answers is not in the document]
ks129: Removed in 136ef11.

tests/bot/cogs/moderation/test_infractions.py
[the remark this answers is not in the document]
MarkKoz: I think there's also a simple way to assert `apply_infraction` is called with the `ban` or `kick` coroutine. Since `ban` and `kick` are mocked, they'll always return the same exact mock when you call them.

bot/cogs/moderation/scheduler.py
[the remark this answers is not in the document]
MarkKoz: Yeah, it looks fine.

bot/cogs/moderation/scheduler.py
```diff
@@ -174,12 +174,15 @@ async def apply_infraction(
...
+        else:
+            infr_message = f"**{infr_type}** to {user.mention}{expiry_msg}{end_msg}"
 
         # Send a confirmation message to the invoking context.
         log.trace(f"Sending infraction #{id_} confirmation message.")
         await ctx.send(
             f"{dm_result}{confirm_msg} "
-            f"{f'**{infr_type}** to {user.mention}{expiry_msg}{end_msg}' if not failed else ''}."
+            f"{infr_message}."
         )
```
MarkKoz: Because it's now a lot shorter, it should all be on one line instead of being broken up.

tests/bot/cogs/moderation/test_infractions.py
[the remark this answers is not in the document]
ks129: Did that in 3233174.

bot/cogs/moderation/scheduler.py
ks129: Resolved in e236113.

tests/bot/cogs/moderation/test_infractions.py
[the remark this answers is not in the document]
MarkKoz: Very nice.

bot/cogs/moderation/scheduler.py
MarkKoz: The space should also be moved to the `infr_message` rather than here. Otherwise, there will be a space before the full stop when `failed == True`.

ks129: Space moved in f1f2c48.

PR #866, opened 2020-04-05
feature to restrict tags to specific role(s)

### LINKED ISSUE #778: Role restricted tags
## Abstract
We may want to restrict certain tags to certain roles. For example, we may want tags that are only available to moderators and higher, which explain moderation guidelines, bot commands, stuff like that. We may want tags that are only available to Contributors that explain certain details related to our bot and site repos.

## How do we do that?
Well first of all, **we should probably wait until https://github.com/python-discord/bot/issues/777 is solved.**

Once that has been done, well, I'm not entirely sure how we do it. Off the top of my head, my suggestion would be to simply structure it so that tags restricted to a specific role are put into a subfolder inside the `tags` folder with a name that corresponds with the role. For example, tags inside `/bot/bot/resources/tags/helpers` are available to all Helpers.

Because we have cascading staff roles (every Moderator is also a Helper, every Owner is an Admin, a Moderator and a Helper), we do not necessarily need to support tags that are available to a certain subset of roles - but if you can think of an elegant way to do that, then by all means, let's discuss it.

## Just tell me what to do!
**This is an issue that is intentionally left open-ended.** If you want to handle this, you should be comfortable making changes to the above proposal if, while handling this, you find a major problem with this proposal, or if you come across a better way of solving it. Please discuss this issue with us in #dev-contrib as you work on it, and try to arrive at a good solution.

RohanJnr:
Closes #778 

### Implementation

as suggested in the issue.

#### Details

- Created folders for 3 roles: Helpers, Moderators, and Admins.
- When a user uses the `tag` command, a new method called `check_accessibility` checks if the user has the necessary role to access that `tag`.
- If the user does not have the required role, no message is displayed, the user assumes no such tag exists.

#### Additional

- Implemented this to the methods which search tags by content.
- When displaying all the tags, the tags which are not accessible by the user are not shown.


bot/cogs/tags.py
```diff
@@ -36,18 +36,32 @@ def get_tags() -> dict:
...
+                    "embed": {
+                        "description": file.read_text()
+                    },
+                    "restricted_to": "developers"
                 }
-            }
-            cache[tag_title] = tag
+                if len(file_path) == 5:
+                    restricted_to = file_path[3]
+                    tag["restricted_to"] = restricted_to
```
ks129, line 53: Why is this required? Isn't easier to make `tag["restricted_to"] = file_path[3]`?

bot/cogs/tags.py
```diff
@@ -113,8 +127,9 @@ def _get_tags_via_content(self, check: Callable[[Iterable], bool], keywords: str
 
         matching_tags = []
         for tag in self._cache.values():
-            if check(query in tag['embed']['description'].casefold() for query in keywords_processed):
-                matching_tags.append(tag)
+            if self.check_accessibility(user, tag):
+                if check(query in tag['embed']['description'].casefold() for query in keywords_processed):
```
ks129, line 131: I'd suggest joining these if statements to one.

bot/cogs/tags.py
```diff
@@ -113,7 +126,10 @@ def _get_tags_via_content(self, check: Callable[[Iterable], bool], keywords: str
 
         matching_tags = []
         for tag in self._cache.values():
-            if check(query in tag['embed']['description'].casefold() for query in keywords_processed):
+            if (
+                self.check_accessibility(user, tag)
+                and check(query in tag['embed']['description'].casefold() for query in keywords_processed)
+            ):
```
MarkKoz, line 132: Assigning the value to a separate variable avoids breaking the if statement into multiple lines.
```suggestion
            matches = (query in tag['embed']['description'].casefold() for query in keywords_processed)
            if self.check_accessibility(user, tag) and check(matches):
```

ikuyarihS:
There's a minor config, can you resolve it @RohanJnr then it's good to go!

Issue #867, opened 2020-04-05
Make !dormant available to original claimant, and delete invocation.


lemonsaurus:
By far the most requested feature in **#help-channel-feedback** is to allow the original claimant in a help channel to mark the channel as `!dormant`. After considering the drawbacks vs the benefit, we have agreed that we should do this.

We'd also like to make one small related change in the same pull request - Whenever `!dormant` is called, the bot should delete the invocation message. This keeps our help channels from filling up with `!dormant` invocations, which prevents unfamiliar users from constantly trying to use the command just to see what it does.

This issue asks you to make two small changes to the help channel system:

## Original claimant should be able to mark a channel as dormant

- [ ] When a channel is first claimed, store the ID of the user who claimed it and the ID of the channel in a cache. This can just be a simple dictionary.
- [ ] When `!dormant` is invoked, allow it if the original claimant (or a staff member) is the caller.
- [ ] Don't worry about persisting across reboots - we'll deal with that later.

## Remove the invoking message when `!dormant` is called.
- [ ] Whenever `!dormant` is called, the bot should delete the invocation message (the message that says `!dormant`).

## When the `!dormant` command is invoked, remove the cooldown from the channel owner

- [ ] Whenever `!dormant` is called, the person who owns that help channel should have any outstanding cooldown they have pardoned, so that they can claim a new channel. This should happen both if a staff member and the channel owner invokes the command.

Numerlor:
Can jump at this unless you had someone familiar with the system in mind

MarkKoz:
Something to note: if the command is used, their cooldown for claiming a channel should also prematurely end.

lemonsaurus:
@Numerlor You're welcome to it, as long as you're able to address it quite soon. We'd like this merged maybe within a day or two, since there's an ongoing evaluation of this system going on.

If that's a bit sooner than you were thinking, we'll handle it among the core devs.

Numerlor:
Resetting the cooldown completely seems to be a bad idea because it would allow users to repatedly make one channel at a time in use and put it back into the dormant state.

I think cancelling the initial task and creating a new one that will reset the permissions in a (potentially) shorter time would be fitting here, does a minute sound like a good cooldown if this is wanted? And should that be in the config or just a literal/module const?

lemonsaurus:
[...]

PR #868, opened 2020-04-06
Implement help system feedback.

Numerlor:
closes: #867 

Implements the requested feedback from the linked issue for the dormant command, allowing claimants to close their own channel, restore their send permissions to available channels and delete the invokation message.

I'm not entirely satisfied with the check here but couldn't think of a nice way to make it a decorator, so I'm open to other ideas if there are any.

```python
await self.available_category.set_permissions(member, send_messages=None)
```
Was not sure if the `send_messages` should be reset here or the whole overwrite (and if it matters), but left it at `send_messages`

### DIFF
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -210,6 +210,9 @@ class HelpChannels(Scheduler, commands.Cog):
         log.trace("dormant command invoked; checking if the channel is in-use.")
         if ctx.channel.category == self.in_use_category:
             if await self.dormant_check(ctx):
+                with suppress(KeyError):
+                    del self.help_channel_claimants[ctx.channel]
+
                 with suppress(discord.errors.NotFound):
                     log.trace("Deleting dormant invokation message.")
                     await ctx.message.delete()
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -196,6 +196,7 @@ class HelpChannels(Scheduler, commands.Cog):
         if ctx.channel.category == self.in_use_category:
             self.cancel_task(ctx.channel.id)
             await self.move_to_dormant(ctx.channel)
+            await ctx.message.delete()
         else:
             log.debug(f"{ctx.author} invoked command 'dormant' outside an in-use help channel")
 
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -199,7 +199,13 @@ class HelpChannels(Scheduler, commands.Cog):
 
     @commands.command(name="dormant", aliases=["close"], enabled=False)
     async def dormant_command(self, ctx: commands.Context) -> None:
-        """Make the current in-use help channel dormant."""
+        """
+        Make the current in-use help channel dormant.
+
+        Make the channel dormant if the user passes the `dormant_check`,
+        delete the message that invoked this,
+        and reset the send permissions cooldown for the user who started the session.
+        """
         log.trace("dormant command invoked; checking if the channel is in-use.")
         if ctx.channel.category == self.in_use_category:
             if await self.dormant_check(ctx):
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -5,6 +5,7 @@ import logging
 import random
 import typing as t
 from collections import deque
+from contextlib import suppress
 from datetime import datetime
 from pathlib import Path
 
@@ -211,7 +212,9 @@ class HelpChannels(Scheduler, commands.Cog):
             if await self.dormant_check(ctx):
                 self.cancel_task(ctx.channel.id)
                 await self.move_to_dormant(ctx.channel)
-                await ctx.message.delete()
+                with suppress(discord.errors.NotFound):
+                    await ctx.message.delete()
+                    log.trace("Deleting dormant invokation message.")
                 await self.reset_send_permissions_for_help_user(ctx.channel)
         else:
             log.debug(f"{ctx.author} invoked command 'dormant' outside an in-use help channel")
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -190,12 +190,12 @@ class HelpChannels(Scheduler, commands.Cog):
         return deque(available_names)
 
     async def dormant_check(self, ctx: commands.Context) -> bool:
-        """Return True if the user started the help channel session or passes the role check."""
+        """Return True if the user is the help channel claimant or passes the role check."""
         if self.help_channel_claimants.get(ctx.channel) == ctx.author:
-            log.trace(f"{ctx.author} started the help session, passing the check for dormant.")
+            log.trace(f"{ctx.author} is the help channel claimant, passing the check for dormant.")
             return True
 
-        log.trace(f"{ctx.author} did not start the help session, checking roles.")
+        log.trace(f"{ctx.author} is not the help channel claimant, checking roles.")
         return with_role_check(ctx, *constants.HelpChannels.cmd_whitelist)
 
     @commands.command(name="dormant", aliases=["close"], enabled=False)
@@ -621,12 +621,12 @@ class HelpChannels(Scheduler, commands.Cog):
         await self.ensure_permissions_synchronization(self.available_category)
 
     async def reset_claimant_send_permission(self, channel: discord.TextChannel) -> None:
-        """Reset send permissions in the Available category for member that started the help session in `channel`."""
-        log.trace(f"Attempting to find user for help session in #{channel.name} ({channel.id}).")
+        """Reset send permissions in the Available category for the help `channel` claimant."""
+        log.trace(f"Attempting to find claimant for #{channel.name} ({channel.id}).")
         try:
             member = self.help_channel_claimants[channel]
         except KeyError:
-            log.trace(f"Channel #{channel.name} ({channel.id}) not in help session cache, permissions unchanged.")
+            log.trace(f"Channel #{channel.name} ({channel.id}) not in claimant cache, permissions unchanged.")
             return
         log.trace(f"Resetting send permissions for {member} ({member.id}).")
         await self.available_category.set_permissions(member, overwrite=None)
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -109,7 +109,7 @@ class HelpChannels(Scheduler, commands.Cog):
         super().__init__()
 
         self.bot = bot
-        self.help_channel_users: t.Dict[discord.TextChannel, discord.User] = {}
+        self.help_channel_claimants: t.Dict[discord.TextChannel, discord.User] = {}
 
         # Categories
         self.available_category: discord.CategoryChannel = None
@@ -191,7 +191,7 @@ class HelpChannels(Scheduler, commands.Cog):
 
     async def dormant_check(self, ctx: commands.Context) -> bool:
         """Return True if the user started the help channel session or passes the role check."""
-        if self.help_channel_users.get(ctx.channel) == ctx.author:
+        if self.help_channel_claimants.get(ctx.channel) == ctx.author:
             log.trace(f"{ctx.author} started the help session, passing the check for dormant.")
             return True
 
@@ -215,7 +215,7 @@ class HelpChannels(Scheduler, commands.Cog):
                 with suppress(discord.errors.NotFound):
                     await ctx.message.delete()
                     log.trace("Deleting dormant invokation message.")
-                await self.reset_send_permissions_for_help_user(ctx.channel)
+                await self.reset_claimant_send_permission(ctx.channel)
         else:
             log.debug(f"{ctx.author} invoked command 'dormant' outside an in-use help channel")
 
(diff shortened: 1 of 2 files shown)


bot/cogs/help_channels.py
```diff
@@ -183,15 +185,33 @@ def create_name_queue(self) -> deque:
...
         if ctx.channel.category == self.in_use_category:
-            self.cancel_task(ctx.channel.id)
-            await self.move_to_dormant(ctx.channel)
+            if await self.dormant_check(ctx):
+                self.cancel_task(ctx.channel.id)
+                await self.move_to_dormant(ctx.channel)
+                with suppress(discord.errors.NotFound):
+                    await ctx.message.delete()
+                    log.trace("Deleting dormant invokation message.")
+                await self.reset_send_permissions_for_help_user(ctx.channel)
```
MarkKoz, line 214: I think the order should be different:

1. Delete the command so message lingers for as short as possible
2. Reset permissions so it still happens in case moving the channel fails
3. Move the channel

Also, put blank lines between groups to make the code more readable.

MarkKoz, line 214: Do you think it'd make more sense to call this within `move_to_dormant`? I'm not sure. In practice, it doesn't matter I suppose. It'd only make a difference if the cooldown was configured to be longer than the dormant timeout.

bot/cogs/help_channels.py
```diff
@@ -549,6 +571,19 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
...
+        """Reset send permissions in the Available category for member that started the help session in `channel`."""
+        log.trace(f"Attempting to find user for help session in #{channel.name} ({channel.id}).")
+        try:
+            member = self.help_channel_users[channel]
+        except KeyError:
+            log.trace(f"Channel #{channel.name} ({channel.id}) not in help session cache, permissions unchanged.")
+            return
+        log.trace(f"Resetting send permissions for {member} ({member.id}).")
+        await self.available_category.set_permissions(member, send_messages=None)
+        # Cancel task, ignore no task existing when the claim time passed but idle time has not.
```
MarkKoz, line 584: It took me a while to understand the comment. It could be written more clearly:

```suggestion
        # Ignore missing task when claim cooldown has passed but the channel still isn't dormant.
```

bot/cogs/help_channels.py
MarkKoz: There should be error handling to still move the channel if the resetting permissions fails (e.g. user left the guild?)

bot/cogs/help_channels.py
```diff
@@ -183,15 +185,33 @@ def create_name_queue(self) -> deque:
...
+        log.trace("dormant command invoked; checking if the channel is in-use.")
         if ctx.channel.category == self.in_use_category:
-            self.cancel_task(ctx.channel.id)
-            await self.move_to_dormant(ctx.channel)
+            if await self.dormant_check(ctx):
+                self.cancel_task(ctx.channel.id)
+                await self.move_to_dormant(ctx.channel)
+                with suppress(discord.errors.NotFound):
+                    await ctx.message.delete()
+                    log.trace("Deleting dormant invokation message.")
```
MarkKoz, line 213: It makes more sense for the log to go before the delete given the way you wrote the message.

bot/cogs/help_channels.py
```diff
@@ -549,6 +571,19 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
...
+    async def reset_send_permissions_for_help_user(self, channel: discord.TextChannel) -> None:
+        """Reset send permissions in the Available category for member that started the help session in `channel`."""
+        log.trace(f"Attempting to find user for help session in #{channel.name} ({channel.id}).")
+        try:
+            member = self.help_channel_users[channel]
+        except KeyError:
+            log.trace(f"Channel #{channel.name} ({channel.id}) not in help session cache, permissions unchanged.")
+            return
+        log.trace(f"Resetting send permissions for {member} ({member.id}).")
+        await self.available_category.set_permissions(member, send_messages=None)
```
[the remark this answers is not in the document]
Numerlor: on an unrelated note, does that mean we can also throw them away in the Silence cog?

bot/cogs/help_channels.py
```diff
@@ -549,6 +571,19 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
                 log.trace(f"Resetting send permissions for {member} ({member.id}).")
                 await self.available_category.set_permissions(member, send_messages=None)
 
+    async def reset_send_permissions_for_help_user(self, channel: discord.TextChannel) -> None:
```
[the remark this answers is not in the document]
Numerlor: Have to admit that my creativity went out for the day and I spent most of the time on wording, thanks for pointing these out.
Will also change the cache dict to remain similar with this change

bot/cogs/help_channels.py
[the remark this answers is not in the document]
MarkKoz: Doesn't that operate on a role basis rather than on individuals? If yes, then there's only 1 role and it won't cause any harm to leave it.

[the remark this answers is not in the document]
Numerlor: Sorry, I keep mixing up users and the main role here; came to the same realization quite a few times now... deleted the previous message before I got the response.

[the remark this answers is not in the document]
Numerlor: Was the `send_messages` in `reset_send_permissions` intentional? This will cause the same issue after the bot loads up if I'm understanding it correctly
https://github.com/python-discord/bot/blob/7571cabe65e39d231523e713923cd23b927225bc/bot/cogs/help_channels.py#L547-L550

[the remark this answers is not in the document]
MarkKoz: An oversight. Would appreciate if you fixed that.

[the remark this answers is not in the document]
Numerlor: Will do, this lead to my initial confusion on which was to be used

bot/cogs/help_channels.py
Numerlor: I considered moving it somewhere else as the command itself is doing a bit now but thought it'd be better to leave the `move_to_dormant` method to do only the things relevant to it

Numerlor:
@MarkKoz To be clear here, how do we want the error handling to pan out?
Is the issue described in the comment also present now if the task fails?
So we want to remove from the cache as early as possible to make sure it can't be ran more than once, then try to reset the permissions and if that fails move the channel and attempt it again?
Got a bit confused here 
> If the channel remains intact, then at the least the dormant command could be invoked again by staff. I say staff because there should be error handling to ensure the channel is still dormant even if resetting permissions fails

Is `intact` here remain in use?

MarkKoz:
Sorry, meant to write "if the *cache* remains intact". I don't have a clear idea of how to do it. Ideally if something fails, it'd still be possible to repeat the command to try again. If the cache is cleared, then repeating the command wouldn't work for resetting permissions. Furthermore, the operations should be independent of each other. If the permissions can't be reset, still move the channel and vice versa.

Numerlor:
How should it be handled for the instances where the channel can be moved but the permissions aren't reset? Currently the command will only work for channels that are in use, skipping execution outside of them so we can't attempt it again to reset the permissions after it has been moved in case of some kind of an api error.

MarkKoz:
Right. My idea was to have staff use it in dormant channels but that's not really going to work out it seems. Maybe nothing can be done besides waiting it out in that case.

Numerlor:
The least we could do is notify staff but I'd like more opinion on that.

SebastiaanZ:
I'm not sure if I'm really worried if a permissions update doesn't work out; the worst that can happen is that the user has to wait the normal period of 15 minutes, right? Could be annoying, but it's not the end of the world for the people involved.

A far greater issue at the moment is that the overwrites system doesn't work reliably. Despite there being no reason for it, as far as I can see, channels keep getting out of sync with the category permissions **even with the resent mitigation effort** in place.

I think that, at the moment, the best we can do is get this PR out of the way so we can switch the permissions system to a role based one, because the overwrites issue is something that currently happens multiple times per day. I think we can worry about permissions not resetting during the execution of this command later.

(To be sure, I obviously see the value in the discussion; it's just that the other thing is quite critical.)

Numerlor:
Sorry, didn't get the time to look at this yesterday.
Pushed in changes without the handling now.
In here would it make more sense to cancel the task after moving the channel, in case the moving fails?
https://github.com/python-discord/bot/blob/12c80fb664e612a2319dfde7d341737159934e9c/bot/cogs/help_channels.py#L197-L198

MarkKoz:
> In here would it make more sense to cancel the task after moving the channel, in case the moving fails?

Yes.

MarkKoz:
> I'm not sure if I'm really worried if a permissions update doesn't work out; the worst that can happen is that the user has to wait the normal period of 15 minutes, right? Could be annoying, but it's not the end of the world for the people involved.

OK let's just delete the channel from `help_channel_claimants` and leave it at that.

Numerlor:
Pushed in a fix for logging on win with this cog in the latest commit.
Should be ready now without the handling, but not sure how to handle the discarded config change that went through the merge with master.

bot/cogs/help_channels.py
```diff
@@ -210,6 +210,9 @@ def create_name_queue(self) -> deque:
         log.trace("dormant command invoked; checking if the channel is in-use.")
         if ctx.channel.category == self.in_use_category:
             if await self.dormant_check(ctx):
+                with suppress(KeyError):
+                    del self.help_channel_claimants[ctx.channel]
```
MarkKoz, line 214: I'd always delete when a channel is main dormant - not just in the dormant command. But it's not important.

bot/cogs/help_channels.py
```diff
@@ -217,8 +217,8 @@ def create_name_queue(self) -> deque:
                 with suppress(discord.errors.HTTPException, discord.errors.NotFound):
                     await self.reset_claimant_send_permission(ctx.channel)
 
-                self.cancel_task(ctx.channel.id)
                 await self.move_to_dormant(ctx.channel)
+                self.cancel_task(ctx.channel.id)
```
MarkKoz, line 224: On second thought this could be a race condition but I'm not too worried about it.

bot/cogs/help_channels.py
Numerlor: Is there anything that's always ran when moving except the `move_to_dormant` itself that could be used?
Not losing out on much here through the few extra dicts pairs and as I mentioned in a previous conversation I think it makes more sense to just leave the move method to only do stuff directly relevant to it when possible with the command itself grouping the actions now.

MarkKoz: No, I don't think so. But my point wasn't to move the delete just for the sake of moving it. It was so the dictionary would also be cleared when a move happens automatically. Really, the only benefit is that it prevents it from accumulating useless values. However, I imagine the memory footprint will still be insignificant which is why I said it's not important

SebastiaanZ: > However, I imagine the memory footprint will still be insignificant which is why I said it's not important

That's my intuition as well. Since we're keeping a mapping of channels to members, we should have a maximum of "max help channels" mappings in the dictionary under normal circumstances. Even if all channels go dormant with command invocation, the number of additional object references should be manageable.

bot/cogs/help_channels.py
SebastiaanZ: Just to document, could you describe the race condition you were thinking of? Could be relevant if we do notice something going wrong in the future. (Although I'm not worried about that as you're not too worried about that as well.)

`self.cancel_task` should only be called once `move_to_dormant` returns, which I think is after it has done its work (in the current implementation), I think. There are currently no `create_task` or other non-awaited scheduling methods involved.

bot/cogs/help_channels.py
```diff
@@ -598,6 +625,20 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
...
+        """Reset send permissions in the Available category for the help `channel` claimant."""
+        log.trace(f"Attempting to find claimant for #{channel.name} ({channel.id}).")
+        try:
+            member = self.help_channel_claimants[channel]
+        except KeyError:
+            log.trace(f"Channel #{channel.name} ({channel.id}) not in claimant cache, permissions unchanged.")
+            return
+
+        log.trace(f"Resetting send permissions for {member} ({member.id}).")
+        await self.available_category.set_permissions(member, overwrite=None)
```
SebastiaanZ, line 638: Due to #873 being merged, there's now a help method that tries to ensure that channels stay in sync after a permissions overwrite change has occurred on the category. I think we should use that instead here, although we're planning to change to a role-based permissions system soon anyway. Still, for now, we should probably keep it consistent.

bot/cogs/help_channels.py
MarkKoz: Maybe race condition is the wrong term. If the command is invoked when the channel is about to automatically go dormant, then it may attempt to move an already dormant channel. I am not sure what the consequences of that are. This problem would still exist anyway even if the order was flipped back around. A robust solution would probably require some sort of lock, but that would mean one lock per channel which gets relatively complex.

SebastiaanZ: Thanks, I see what you mean. I guess we'll found out, but this is bound to happen occasionally.

MarkKoz:
I just remembered that some docstrings (at least the class's docstring) will need to be updated because the command is no longer solely role restricted as I believe it currently states. The message constants should maybe also be updated to mention the command, but that may require some discussion.

PR #871, opened 2020-04-06
Add status emojis to help channels and improve bottom sorting

SebastiaanZ:
This PR aims to make two changes to our help channel system:

**1. Add status emojis to help channel names**

The new help channel system appears to be intuitive to most of our members. One comment that has been made a few times is that it would be clearer if the channel names would indicate the status of the channels as well. It still happens that people don't notice the existence of two different categories and add their question to an already claimed channel.

By adding emojis to the channel name, I hope that it's more apparent to the end-user that there's a difference between channels in the "In Use" and in the "Available" category.

The following design was chosen after a conversation in the feedback channel:
channel_status_emojis

**2. Change the bottom sorting strategy to large ints.**

We currently attempt to position channels at the bottom of their new category ("In Use" and "Dormant" only) when we move them by getting the maximum position in their new category and adding 1 to it. However, this doesn't always seem to work as intended and channels may end up somewhere in the bottom third of the list.

Another strategy, setting a really large position integer, does seem to give more reliable results. Discord will automatically "truncate" the position integers by resizing them to the maximum current position + 1 and positioning seems to be predictable.

I'm not 100% sure it will actually work as expected in production, but it does work reliably in the testing environment. I am fairly confident that it will be no worse than our current strategy, so I think we should just try it.

### DIFF
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -223,13 +223,14 @@ class HelpChannels(Scheduler, commands.Cog):
     @staticmethod
     def get_clean_channel_name(channel: discord.TextChannel) -> str:
         """Return a clean channel name without status emojis prefix."""
+        prefix = constants.HelpChannels.name_prefix
         try:
-            # Try to remove the status prefix using the index of "help-"
-            name = channel.name[channel.name.index("help-"):]
+            # Try to remove the status prefix using the index of the channel prefix
+            name = channel.name[channel.name.index(prefix):]
             log.trace(f"The clean name for `{channel}` is `{name}`")
         except ValueError:
             # If, for some reason, the channel name does not contain "help-" fall back gracefully
-            log.info(f"Can't get clean name as `{channel}` does not follow the `help-` naming convention.")
+            log.info(f"Can't get clean name as `{channel}` does not follow the `{prefix}` naming convention.")
             name = channel.name
 
         return name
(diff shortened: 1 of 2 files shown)


SebastiaanZ:
> I like the novelty of using large int pushing newly claimed channels to the bottom. This can somewhat help indicating what channels have been in-use longer.

The ints won't be preserved; Discord automatically resizes them to `max + 1` in a category.

bot/cogs/help_channels.py
```diff
@@ -217,23 +221,18 @@ def create_name_queue(self) -> deque:
...
-
-        log.trace(
-            f"Position of #{channel} ({channel.id}) in {destination.name} will be {position} "
-            f"(was {channel.position})."
-        )
+    def get_clean_channel_name(channel: discord.TextChannel) -> str:
+        """Return a clean channel name without status emojis prefix."""
+        try:
+            # Try to remove the status prefix using the index of "help-"
+            name = channel.name[channel.name.index("help-"):]
```
MarkKoz, line 228: The prefix is configurable so "help-" shouldn't be hard coded.

SebastiaanZ: Good catch.

SebastiaanZ:
> I agree with Marks feedback but otherwise it looks good.

It should be addressed in the latest commit

PR #873, opened 2020-04-07
Mitigate available help channels failing to synchronize their permissions

SebastiaanZ:
This PR aims to mitigate the issue that available help channels fail to synchronize their permissions with the `Help: Available` category despite our best efforts. As we have not yet identified the root cause of the issue, I've implemented an `ensure_permissions_synchronization` helper method that is called to check and, where needed, remedy available help channel permissions synchronization issues.

I've also changed the category name `Help: In Use` to `Help: Occupied` to reflect the current situation on server.

### DIFF
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -427,6 +427,12 @@ class HelpChannels(Scheduler, commands.Cog):
             topic=AVAILABLE_TOPIC,
         )
 
+        log.trace(
+            f"Ensuring that all channels in `{self.available_category}` have "
+            f"synchronized permissions after moving `{channel}` into it."
+        )
+        await self.ensure_permissions_synchronization(self.available_category)
+
     async def move_to_dormant(self, channel: discord.TextChannel) -> None:
         """Make the `channel` dormant."""
         log.info(f"Moving #{channel} ({channel.id}) to the Dormant category.")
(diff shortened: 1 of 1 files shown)


bot/cogs/help_channels.py
```diff
@@ -544,14 +550,53 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
...
+        This method mitigates an issue we have yet to find the cause for: Every so often, a channel in the
+        `Help: Available` category gets in a state in which it will no longer synchronizes its permissions
+        with the category. To prevent that, we iterate over the channels in the category and edit the channels
+        that are observed to be in such a state. If no "out of sync" channels are observed, this method will
+        not make API calls and should be fairly inexpensive to run.
+        """
+        for channel in category.channels:
+            if not channel.permissions_synced:
+                log.info(f"The permissions of channel `{channel}` were out of sync with category `{category}`.")
+                await channel.edit(sync_permissions=True)
```
MarkKoz, line 567: Would it be worth doing this in parallel like with a gather?

SebastiaanZ: Good point; we could do that. (Although it's currently max. two channels as we only apply it to the Available category.)

MarkKoz: Oh never mind then. I thought it was doing it for all categories.

bot/cogs/help_channels.py
```diff
@@ -544,14 +550,53 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
...
+
+    async def update_category_permissions(
+        self, category: discord.CategoryChannel, member: discord.Member, **permissions
```
MarkKoz, line 570: Technically it's all kwargs not just permissions but I can live with it.

bot/cogs/help_channels.py
```diff
@@ -544,14 +550,53 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
...
+        """
+        Ensure that all channels in the `category` have their permissions synchronized.
+
+        This method mitigates an issue we have yet to find the cause for: Every so often, a channel in the
+        `Help: Available` category gets in a state in which it will no longer synchronizes its permissions
+        with the category. To prevent that, we iterate over the channels in the category and edit the channels
+        that are observed to be in such a state. If no "out of sync" channels are observed, this method will
+        not make API calls and should be fairly inexpensive to run.
+        """
+        for channel in category.channels:
```
MarkKoz, line 564: I haven't benchmarked but I try to avoid that attribute because it sorts. That's why I made `self.get_category_channels()`.

PR #876, opened 2020-04-11
StatsD integration

jos-b:
Integrates the bot with Graphite (through statsd).

Provides the following stats:
- `mod_alerts.{rule_name}` - Triggers of the mod alerts
- `defcon.leaves` - Users denied access due to defcon
- `defcon.days` - Current number of days defcon is active for
- `errors.{error_name}` - Different types of errors experienced
- `filters.{filter_name}` - Messages violating content filters
- `help.total.in_use` - Total in use help channels
- `help.total.available` - Total available help channels
- `help.total.dormant` - Total dormant help channels
- `help.out_of_channel_alerts` - Alerts to Helpers for when we hit capacity
- `help.claimed` - Total claimed help channels
- `channels.{reformatted_name}` - Total messages in channels
- `messages` - Total messages
- `commands.{command_name}` - Total command executions
- `guild.total_members` - Total guild member count
- `guild.status.online` - Total online members
- `guild.status.idle` - Total idle members
- `guild.status.do_not_disturb` - Total DND members
- `guild.status.offline` - Total offline members
- `tags.usages.{tag_name}` - Total tag usages
- `tokens.removed_tokens` - Total removed tokens
- `tokens.removed_webhooks` - Total removed webhooks
- `help.dormant_calls.{auto/command}` - Was a help channel auto marked as dormant or was !dormant used.
- `help.in_use_time` - A timer for how long a help channel was claimed for.

If anyone has suggestions for further statistics please add a comment and I'll add them onto this PR. Once merged, additional statistics can be added in future PRs.


bot/cogs/stats.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @Cog.listener()
+    async def on_member_update(self, _before: Member, after: Member) -> None:
+        """Update presence estimates on member update."""
+        members = after.guild.members
+
+        online = len([m for m in members if m.status == Status.online])
+        idle = len([m for m in members if m.status == Status.idle])
+        dnd = len([m for m in members if m.status == Status.do_not_disturb])
+        offline = len([m for m in members if m.status == Status.offline])
```
MarkKoz, line 55: This strikes me as terribly inefficient. This could be done with a single loop. Maybe 40k members is still chump change for python.

MarkKoz: Oh it's only iterating changed users not all.

MarkKoz: Wait, yes it is all. Lol

bot/bot.py
```diff
@@ -33,6 +33,16 @@ def __init__(self, *args, **kwargs):
...
 
+        statsd_url = constants.Bot.statsd_host
+
+        if DEBUG_MODE:
+            # Since statsd is UDP, there are no errors for sending to a down port.
+            # For this reason, setting the statsd host to 127.0.0.1 for development
+            # will effectively disable stats.
+            statsd_url = "127.0.0.1"
+
+        self.stats = statsd.StatsClient(statsd_url, 8125, prefix="bot")
```
MarkKoz, line 44: How does this communicate? Is it async? Should we care?

bot/__main__.py
```diff
@@ -57,6 +57,7 @@
 bot.load_extension("bot.cogs.site")
 bot.load_extension("bot.cogs.snekbox")
 bot.load_extension("bot.cogs.sync")
+bot.load_extension("bot.cogs.stats")
```
MarkKoz, line 60: Not sorted alphabetically :(

bot/cogs/stats.py
```diff
@@ -0,0 +1,65 @@
...
+        """Report message events in the server to statsd."""
+        if message.guild is None:
+            return
+
+        reformatted_name = message.channel.name.replace('-', '_')
+
+        if reformatted_name.startswith("ot"):
+            # Off-topic channels change names, we don't want this for stats.
+            # This will change 'ot1-lemon-in-the-dishwasher' to just 'ot1'
+            reformatted_name = reformatted_name[:3]
```
MarkKoz, line 22: Would rather this check the category rather than relying solely on the prefix.

bot/cogs/stats.py
```diff
@@ -0,0 +1,65 @@
...
+    async def on_command_completion(self, ctx: Context) -> None:
+        """Report completed commands to statsd."""
+        command_name = ctx.command.qualified_name.replace(" ", "_")
+
+        self.bot.stats.incr(f"commands.{command_name}")
+
+    @Cog.listener()
+    async def on_member_join(self, member: Member) -> None:
+        """Update member count stat on member join."""
+        self.bot.stats.gauge(f"guild.total_members", len(member.guild.members))
```
MarkKoz, line 40: The bot is in some other guilds, like the emoji server. If not accounted for, the total could be overwritten. I think all events should explicitly check for the main guild.

bot/bot.py
jos-b: That's something I've been looking into. It's not async but the method for sending stats is basically a UDP payload like `statname:1|c` to increment a counter. As far as I can tell there is little way for this to block.

https://github.com/jsocol/pystatsd/blob/1c90b9fdf322680e2625da659abc2aa5d79b5bff/statsd/client/udp.py#L14-L24

bot/cogs/filtering.py
```diff
@@ -207,6 +207,8 @@ def mod_log(self) -> ModLog:
 
                         log.debug(message)
 
+                        self.bot.stats.incr(f"bot.filters.{filter_name}")
```
MarkKoz, line 210: Why is this name prefixed with `bot`?

bot/cogs/error_handler.py
```diff
@@ -236,6 +236,8 @@ def __init__(self, bot: Bot):
             f"```{e.__class__.__name__}: {e}```"
         )
 
+        ctx.bot.stats.incr("command_error_count")
```
MarkKoz, line 239: Let's name this `commands.error_count` to be consistent with `commands.{command_name}`.

bot/cogs/defcon.py
```diff
@@ -104,6 +104,7 @@ def mod_log(self) -> ModLog:
                     log.exception(f"Unable to send rejection message to user: {member}")
 
                 await member.kick(reason="DEFCON active, user is too new")
+                self.bot.stats.incr("defcon_leaves")
```
MarkKoz, line 107: Naming this `defcon.leaves` would be more consistent with the naming scheme used elsewhere.

bot/cogs/error_handler.py
jos-b: I've avoided that intentionally. Since we may want something like "Total executed commands" and for that we can look at the metric `commands.*` (wildcard) and get all command invocations.

bot/cogs/filtering.py
jos-b: Mistake

bot/bot.py
MarkKoz: Did some digging:

https://github.com/jsocol/pystatsd/issues/93
https://pypi.org/project/aiomeasures/ is async but not being updated for 5 years is concerning.

jos-b: I found the same thing and struggled to make it work as well as `statsd`. If this is a huge issue then statsd is such a simple protocol I could probably write a custom implementation.

MarkKoz: OK, I am fine with trying it as is. But how could we measure if being synchronous is an issue?

bot/cogs/error_handler.py
MarkKoz: That's fair. How about `errors.commands`, in case we add more error stats in the future?

jos-b: :okK_hand:

bot/bot.py
jos-b: That's hard to do, I don't foresee any issues because there is no waiting on a response from statsd and I can't think of any slowdowns over the local Docker network.

jos-b:
Fixed, I'm having a bad git day.

jos-b:
Okay, after speaking with aeros I've decided to write a custom asynchronous transport but keep the utility of a statsd library.

bot/bot.py
jos-b: Async stats client has been implemented in https://github.com/python-discord/bot/commit/bca47a1c2e59fb112b947876cea1836879ac7282 so this should not be a concern.

bot/bot.py
```diff
@@ -101,6 +114,10 @@ def _recreate(self) -> None:
         self.http_session = aiohttp.ClientSession(connector=self._connector)
         self.api_client.recreate(force=True, connector=self._connector)
 
+    async def on_ready(self) -> None:
+        """Construct an asynchronous transport for the statsd client."""
+        await self.stats.create_socket()
```
MarkKoz, line 119: Did you intentional put it in an event that may be called on bot reconnect? If not, this should go in `_recreate()`.

jos-b: Right, but recreate is not an async function, would I have to use `create_task` again on the `loop` attribute of the bot class?

jos-b: It isn't critical, I can have the `create_socket` method check if there is an existing transport and close that before creating another one, but I'm not sure how to do it from `_recreate()`.

MarkKoz: Hmm right. It could go in `login`. If not, at the least the previous socket should be closed before you overwrite it.

MarkKoz: Not being able to have it in `_recreate` kind of makes the `clear()` function incomplete, but in practice we don't use `clear()`.

jos-b: Moved to login

bot/cogs/help_channels.py
```diff
@@ -463,14 +476,14 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
             category=self.in_use_category,
             sync_permissions=True,
             topic=IN_USE_TOPIC,
-            position=10000,
```
MarkKoz, line 466: I think you messed up a merge somewhere because this position was never supposed to exist in the first place.

jos-b: Fixed

bot/cogs/stats.py
```diff
@@ -0,0 +1,93 @@
...
+
+        for member in after.guild.members:
+            if member.status == Status.online:
+                online += 1
+            elif member.status == Status.dnd:
+                dnd += 1
+            elif member.status == Status.idle:
+                idle += 1
+            else:
+                offline += 1
```
MarkKoz, line 83: Use `is` rather than `==` for enums. While you're at it, do an explicit comparison for the offline status in the rare case more statuses get added.

bot/cogs/stats.py
```diff
@@ -0,0 +1,93 @@
...
+    @Cog.listener()
+    async def on_message(self, message: Message) -> None:
+        """Report message events in the server to statsd."""
+        if message.guild is None:
+            return
+
+        if message.guild.id != Guild.id:
+            return
+
+        reformatted_name = message.channel.name.replace('-', '_')
```
MarkKoz, line 30: Will Unicode be an issue. In any case, stripping emojis from help channels would be a good idea.

MarkKoz: Regarding help channels, the cog has a function you can borrow to strip emojis.

jos-b: Stripped to alphanumeric

bot/cogs/stats.py
```diff
@@ -32,6 +36,8 @@ def __init__(self, bot: Bot):
         if CHANNEL_NAME_OVERRIDES.get(message.channel.id):
             reformatted_name = CHANNEL_NAME_OVERRIDES.get(message.channel.id)
 
+        reformatted_name = "".join([char for char in reformatted_name if char in ALLOWED_CHARS])
```
MarkKoz, line 39: List comprehension is not needed.
```suggestion
        reformatted_name = "".join(char for char in reformatted_name if char in ALLOWED_CHARS)
```

bot/cogs/defcon.py
```diff
@@ -146,6 +146,8 @@ def mod_log(self) -> ModLog:
             await ctx.send(self.build_defcon_msg(action, error))
             await self.send_defcon_log(action, ctx.author, error)
 
+            self.bot.stats.gauge("defcon.days", days)
```
MarkKoz, line 149: This is a good stat too but I meant to track the duration it is enabled, not the threshold used.

bot/cogs/help_channels.py
```diff
@@ -457,6 +464,13 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
             topic=DORMANT_TOPIC,
         )
 
+        self.bot.stats.incr(f"help.dormant_calls.{caller}")
+
+        if self.claim_times.get(channel.id):
+            claimed = self.claim_times[channel.id]
+            in_use_time = datetime.now() - claimed
+            self.bot.stats.timer("help.in_use_time", in_use_time)
```
MarkKoz, line 472: Need to suppress key errors since on bot restart claim times will not persist.

jos-b: That's what the if will do though right? If there is no entry in `claim_times` just move on.

MarkKoz: Right, sorry. Didn't notice the `get()`.

MarkKoz: Though an `in` would make more sense.

bot/cogs/defcon.py
jos-b: I see, we don't have that information on demand after a restart though so this would involve changes to the defcon storage in the site KV store right?

MarkKoz: Yeah, I think so. It just stores JSON as the value IIRC. If you think it's worth it then let's go for it.

bot/cogs/help_channels.py
```diff
@@ -367,6 +370,18 @@ def get_used_names(self) -> t.Set[str]:
         log.info("Cog is ready!")
         self.ready.set()
 
+        self.report_stats()
+
+    def report_stats(self) -> None:
+        """Report the channel count stats."""
+        total_in_use = len(list(self.get_category_channels(self.in_use_category)))
+        total_available = len(list(self.get_category_channels(self.available_category)))
+        total_dormant = len(list(self.get_category_channels(self.dormant_category)))
```
aeros, line 379: Since you just need the count, you could avoid creating a list each time the stats are reported with: `sum(1 for _ in self.get_category_channels(self.in_use_category))`. It wouldn't make an overly substantial difference though unless this was being called frequently.

bot/async_stats.py
```diff
@@ -0,0 +1,39 @@
...
+            remote_addr=self._addr
+        )
+
+    def _send(self, data: str) -> None:
+        """Start an async task to send data to statsd."""
+        self._loop.create_task(self._async_send(data))
+
+    async def _async_send(self, data: str) -> None:
+        """Send data to the statsd server using the async transport."""
+        self._transport.sendto(data.encode('ascii'), self._addr)
```
aeros, line 39: I mentioned this in `dev-core`, but this particular implementation (bare `asyncio.create_task()`) could result in the tasks not being completed since the task is just scheduled to be ran in the background but never awaited. It's okay though if it's not particularly critical, and some missing and/or delayed data is acceptable.

bot/cogs/help_channels.py
jos-b: Addressed in 4adcb0b0ab2b8768e31043429bdfcbf4dab607e6

bot/async_stats.py
jos-b: For now I think we'll leave it. Stats are not critical and UDP is already a protocol which is less prone to issues.

jos-b: In future though this rudimentary async stats reporter could be rewritten to have a better error handling structure, but right now my knowledge of asyncio, while gradually improving, is not quite there yet.

If anyone does fancy taking it on at a later date then that's fine by me.

bot/cogs/defcon.py
jos-b: Added

PR #877, opened 2020-04-11
HelpChannels: fix AttributeError getting a category for a DMChannel 

### LINKED ISSUE #870: Check channel type before trying to access category
The fix is the same as https://github.com/python-discord/bot/commit/1509d3b1c39fac825d5e5a09fd0caf780db7c91c. I just overlooked applying it to `help_channels.py` too.

```
Ignoring exception in on_message
  File "/usr/local/lib/python3.8/site-packages/discord/client.py", line 312, in _run_event
    await coro(*args, **kwargs)
  File "/bot/bot/cogs/help_channels.py", line 516, in on_message
    if channel.category and channel.category.id != constants.Categories.help_available:
AttributeError: 'DMChannel' object has no attribute 'category'
```

MarkKoz:
fixes #870


bot/cogs/help_channels.py
```diff
@@ -376,6 +376,12 @@ def is_dormant_message(message: t.Optional[discord.Message]) -> bool:
         embed = message.embeds[0]
         return embed.description.strip() == DORMANT_MSG.strip()
 
+    @staticmethod
+    def is_in_category(channel: discord.TextChannel, category_id: int) -> bool:
+        """Return True if `channel` is within a category with `category_id`."""
+        actual_category = getattr(channel, "category", None)
+        return actual_category and actual_category.id == category_id
```
kwzrd, line 383: I feel like this will return `None` if the category isn't found.
```suggestion
        return actual_category is not None and actual_category.id == category_id
```

kwzrd: Alternatively if this is intentional (since `None` will evaluate to `False` anyway in the conditionals) the return type could be annotated as optional. But I think it's better if it always returns a bool.

MarkKoz: Good catch

PR #885, opened 2020-04-17
Use a role overwrite to keep track of help channel cooldowns

### LINKED ISSUE #839: Use role overwrites for new help channel system
Copying from the main PR #786:

Are per-member overwrites scalable? Discord may have a limit. Furthermore, there is supposedly client performance degradation with a high amount of overwrites. If this becomes an issue, an overwrite for a role can be used instead. The role would be added/removed from users as needed.

Not 100% decided if it should be switched to a role. Given a total channel limit of 32, there's going to be at most a bit over 32 simultaneous overwrites. That should be fine performance wise. I prefer user overwrites since they're "stealthier" than a new role, but a role would technically be a more robust solution. For now, user overwrites will stay and I'll open an issue to switch to roles.

MarkKoz:
Resolves #839

Using user overwrites on the category turned out to be flawed due to syncing issues. Channels would get out of sync with the category which would lead to users being able to claim multiple channels. Using a role should hopefully avoid sync issues since the bot will no longer be constantly adding/removing channel overwrites. Instead, it will be adding/removing roles from users.

For now, the roles aren't persisted. Once we have some better infrastructure for that (or we decided to stick to pgsql), persistence can be revisited.

### DIFF
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -230,7 +230,7 @@ class HelpChannels(Scheduler, commands.Cog):
                     del self.help_channel_claimants[ctx.channel]
 
                 with suppress(discord.errors.HTTPException, discord.errors.NotFound):
-                    await self.reset_claimant_send_permission(ctx.channel)
+                    await self.reset_claimant_send_permission(ctx.author)
 
                 await self.move_to_dormant(ctx.channel, "command")
                 self.cancel_task(ctx.channel.id)
@@ -640,18 +640,8 @@ class HelpChannels(Scheduler, commands.Cog):
                 log.trace(f"Resetting send permissions for {member} ({member.id}).")
                 await member.remove_roles(COOLDOWN_ROLE)
 
-    async def reset_claimant_send_permission(self, channel: discord.TextChannel) -> None:
-        """Reset send permissions in the Available category for the help `channel` claimant."""
-        log.trace(f"Attempting to find claimant for #{channel.name} ({channel.id}).")
-        try:
-            member = self.help_channel_claimants[channel]
-        except KeyError:
-            log.trace(
-                f"Channel #{channel.name} ({channel.id}) not in claimant cache, "
-                f"permissions unchanged."
-            )
-            return
-
+    async def reset_claimant_send_permission(self, member: discord.Member) -> None:
+        """Reset send permissions in the Available category for `member`."""
         log.trace(f"Resetting send permissions for {member} ({member.id}).")
         await member.remove_roles(COOLDOWN_ROLE)
 
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -481,7 +481,6 @@ class HelpChannels(Scheduler, commands.Cog):
             f"Ensuring that all channels in `{self.available_category}` have "
             f"synchronized permissions after moving `{channel}` into it."
         )
-        await self.ensure_permissions_synchronization(self.available_category)
         self.report_stats()
 
     async def move_to_dormant(self, channel: discord.TextChannel, caller: str) -> None:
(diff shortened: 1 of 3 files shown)


MarkKoz:
I don't know to test that the sync issues are gone.

bot/cogs/help_channels.py
```diff
@@ -620,67 +630,46 @@ async def on_message(self, message: discord.Message) -> None:
...
     async def reset_send_permissions(self) -> None:
-        """Reset send permissions for members with it set to False in the Available category."""
+        """Reset send permissions in the Available category for claimants."""
         log.trace("Resetting send permissions in the Available category.")
+        guild = self.bot.get_guild(constants.Guild.id)
 
-        for member, overwrite in self.available_category.overwrites.items():
-            if isinstance(member, discord.Member) and overwrite.send_messages is False:
-                log.trace(f"Resetting send permissions for {member} ({member.id}).")
+        # TODO: replace with a persistent cache cause checking every member is quite slow
```
SebastiaanZ, line 638: Seems like we're now hitting multiple points at which keeping a persistent cache of the claimants would be beneficial to the system (the other one being the `channel->claimant` mapping for the `!dormant` command). Sounds like a nice issue for future implementation.

bot/cogs/help_channels.py
```diff
@@ -620,67 +630,46 @@ async def on_message(self, message: discord.Message) -> None:
...
+        for member in guild.members:
+            if self.is_claimant(member):
+                await self.remove_cooldown_role(member)
 
-                # We don't use the permissions helper function here as we may have to reset multiple overwrites
-                # and we don't want to enforce the permissions synchronization in each iteration.
-                await self.available_category.set_permissions(member, overwrite=None)
+    @classmethod
```
SebastiaanZ, line 643: Out of interest, why did you opt to make this (and `remove_cooldown_role`) a classmethod? Is it just preference because you're only calling a staticmethod or is there another reason behind it?

bot/cogs/help_channels.py
```diff
@@ -620,67 +630,46 @@ async def on_message(self, message: discord.Message) -> None:
...
-        log.trace(f"Resetting send permissions for {member} ({member.id}).")
-        await self.update_category_permissions(self.available_category, member, overwrite=None)
-        # Ignore missing task when claim cooldown has passed but the channel still isn't dormant.
-        self.cancel_task(member.id, ignore_missing=True)
+        `coro` is intended to be `discord.Member.add_roles` or `discord.Member.remove_roles`.
+        """
+        try:
+            await coro
+        except discord.NotFound:
+            log.debug(f"Failed to change role for {member} ({member.id}): member not found")
```
SebastiaanZ, line 665: There's a slight drawback here because of the use of the `discord.Object` to represent the role: `discord.NotFound` will both be raised for the member not being found as well as the role not being found. It's (hopefully) not going to be a huge issue for us in production, but if someone forgot to set an ID for the role in a dev setting, the logger will give the misleading message that the member was not found, while it was actually the role:

```
2020-04-20 11:49:13 | bot.cogs.help_channels | DEBUG | Failed to change role for Ves Zappa#3787 (336843820513755157): member not found
```

That may be a bit misleading. Another option would be to verify if the role exists for the guild the help system is being used in during the `init_cog` process and dispatch a warning when it does not (or even disable the cog, although the role is not as essential as the categories being present).

bot/cogs/help_channels.py
```diff
@@ -481,7 +492,6 @@ async def move_to_available(self) -> None:
             f"Ensuring that all channels in `{self.available_category}` have "
```
SebastiaanZ, line 492: This logging statement can be removed as we're no longer doing the ensure perm thing.

bot/cogs/help_channels.py
MarkKoz: Yes, it is because it's only calling a static method. In retrospect, I don't know if it's worth the bother to use classmethod/staticmethod when this is all unlikely to ever be used outside the class.

bot/cogs/help_channels.py
MarkKoz: Removed.

bot/cogs/help_channels.py
MarkKoz: I've addressed this by making the function always get the role.

bot/cogs/help_channels.py
SebastiaanZ: Okay. It's fine by me. I was just curious to know the reason for turning it into a classmethod instead of calling the staticmethod from a "regular" method. Maybe there was something to learn for me.

MarkKoz:
I realised the command doesn't account for a user having multiple channels claimed. Going to look into it, but at a glance there doesn't seem to be an simple way to address this currently.

SebastiaanZ:
> I realised the command doesn't account for a user having multiple channels claimed. Going to look into it, but at a glance there doesn't seem to be an simple way to address this currently.

Right,  if a member claims two channels (respecting the 15-minute cooldown), they role they received after claiming the second channel will also be removed if the first channel was made dormant as it only does an "is_claimaint" check  that has no knowledge of for which channel the role was assigned. I can live with that.

(I guess this is another one of those "if we really want to fix this consistently, we need a better persistent caching system" as we can then more easily just keep track of for which channel someone received a role.)

bot/cogs/help_channels.py
lemonsaurus: yeah, I added persistance to the road map today, so we'll do it soon.

PR #895, opened 2020-04-20
Allowing `!eval` in help channels

### LINKED ISSUE #893: Allowing eval in help channels
With the new help channel system in effect, we want to change the rules for where the `!eval` command is permitted to be used. However, because our help channels are dynamically created on demand and because we also need to allow it for a few non-help channels, and for certain roles, we need a new decorator and some new constant groups to handle this.

## When should the command be allowed?
- In all help channel categories, including **Topical Help**, **Dormant**, **Available**, and **In Use**.
- In **#esoteric-python** and **#bot-commands**
- For **Helpers**, **Moderators**, **Admins**, **Owners**, **Python Community** and **Partners**.

## Where do we put all these IDs?
In the `default-config.yml`, we can create a new group called `eval` where we have `category_whitelist`, `channel_whitelist` and `role_whitelist` which all contain lists of IDs.

## Okay, but how do we check this?
Let's create a new decorator that will check all of the above.

This decorator, maybe `is_whitelisted`, should take a list of roles, channels and category IDs and also a `redirect_channel`, which is where it will tell the user to go if they try to run the command in a channel where it is not permitted.

SebastiaanZ:
This PR implements the policy change we've agreed upon during the staff meeting: Allowing the `!eval` command to be used by regular members in help channels. However, as our help channels are dynamically generated up to a certain number of channels, using a simple whitelist based on channel IDs won't work. That's why this PR generalizes the `in_channel` decorator to be able to whitelist more aspects of the command's context, including the channel category. This means that we can whitelist channels, including help channels, by the categories that they are in.

This PR closes #893 

## Generalizing the `in_channel` decorator

The previous decorator, `in_channel`, allowed us to whitelist commands based on the channel context they were issued in. However, unlike the name suggested, it could also be used to whitelist commands based on another aspect of the command context: The roles the author of the message containing the commands had. In fact, despite the name `in_channel`, the decorator could in principle be used to whitelist commands solely based on the author's roles using the `bypass_roles` kwarg.

This means that instead of having an `in_channel` decorator, we basically had a decorator that whitelisted commands based on two aspects of the command invocation context (i.e. the channel and the author's roles). To highlight this fact, I've generalized the decorator to a general "on_whitelisted_context` decorator that also allows for a third kind of context whitelist: the category containing the channel.

To force readability, I've made it so all the arguments should given as keyword arguments. This means that it's directly clear on which aspects of the context the decorator acts:

```py
@in_whitelisted_context(
    whitelisted_channels=EVAL_CHANNELS,
    whitelisted_categories=EVAL_CATEGORIES,
    whitelisted_roles=EVAL_ROLES,
    redirect_channel=Channels.bot_commands,
)
```

As it's now no longer obvious if and which redirect channel needs to be used to send a member to in case the check fails, the decorator now takes an optional `redirect_channel` kwarg. If such a channel is not given, the error message will not mention a "go here instead" channel; if it is given, we'll kindly suggest to the member that they should use the command there. (My guess is that this channel will usually be `#bot-commands`.)

If we ever want to add more options for context whitelisting, adding another kwarg to the decorator should be trivial.

### DIFF
--- a/tests/bot/test_decorators.py
+++ b/tests/bot/test_decorators.py
@@ -2,15 +2,15 @@ import collections
 import unittest
 import unittest.mock
 
-from bot.decorators import InWhitelistedContextCheckFailure, in_whitelisted_context
+from bot.decorators import InWhitelistCheckFailure, in_whitelist
 from tests import helpers
 
 
 WhitelistedContextTestCase = collections.namedtuple("WhitelistedContextTestCase", ("kwargs", "ctx"))
 
 
-class InWhitelistedContextTests(unittest.TestCase):
-    """Tests for the `in_whitelisted_context` check."""
+class InWhitelistTests(unittest.TestCase):
+    """Tests for the `in_whitelist` check."""
 
     @classmethod
     def setUpClass(cls):
@@ -23,43 +23,43 @@ class InWhitelistedContextTests(unittest.TestCase):
         cls.staff_role = helpers.MockRole(id=121212)
         cls.staff_member = helpers.MockMember(roles=(cls.staff_role,))
 
-        cls.whitelisted_channels = (cls.bot_commands.id,)
-        cls.whitelisted_categories = (cls.help_channel.category_id,)
-        cls.whitelisted_roles = (cls.staff_role.id,)
+        cls.channels = (cls.bot_commands.id,)
+        cls.categories = (cls.help_channel.category_id,)
+        cls.roles = (cls.staff_role.id,)
 
     def test_predicate_returns_true_for_whitelisted_context(self):
         """The predicate should return `True` if a whitelisted context was passed to it."""
         test_cases = (
             # Commands issued in whitelisted channels by members without whitelisted roles
             WhitelistedContextTestCase(
-                kwargs={"whitelisted_channels": self.whitelisted_channels},
+                kwargs={"channels": self.channels},
                 ctx=helpers.MockContext(channel=self.bot_commands, author=self.non_staff_member)
             ),
-            # `redirect_channel` should be added implicitly to the `whitelisted_channels`
+            # `redirect` should be added implicitly to the `channels`
             WhitelistedContextTestCase(
-                kwargs={"redirect_channel": self.bot_commands.id},
+                kwargs={"redirect": self.bot_commands.id},
                 ctx=helpers.MockContext(channel=self.bot_commands, author=self.non_staff_member)
             ),
 
             # Commands issued in a whitelisted category by members without whitelisted roles
             WhitelistedContextTestCase(
-                kwargs={"whitelisted_categories": self.whitelisted_categories},
+                kwargs={"categories": self.categories},
                 ctx=helpers.MockContext(channel=self.help_channel, author=self.non_staff_member)
             ),
 
             # Command issued by a staff member in a non-whitelisted channel/category
             WhitelistedContextTestCase(
-                kwargs={"whitelisted_roles": self.whitelisted_roles},
+                kwargs={"roles": self.roles},
                 ctx=helpers.MockContext(channel=self.non_whitelisted_channel, author=self.staff_member)
             ),
 
             # With all kwargs provided
             WhitelistedContextTestCase(
                 kwargs={
-                    "whitelisted_channels": self.whitelisted_channels,
-                    "whitelisted_categories": self.whitelisted_categories,
-                    "whitelisted_roles": self.whitelisted_roles,
-                    "redirect_channel": self.bot_commands,
+                    "channels": self.channels,
+                    "categories": self.categories,
+                    "roles": self.roles,
+                    "redirect": self.bot_commands,
                 },
                 ctx=helpers.MockContext(channel=self.help_channel, author=self.staff_member)
             ),
@@ -69,31 +69,31 @@ class InWhitelistedContextTests(unittest.TestCase):
             # patch `commands.check` with a no-op lambda that just returns the predicate passed to it
             # so we can test the predicate that was generated from the specified kwargs.
             with unittest.mock.patch("bot.decorators.commands.check", new=lambda predicate: predicate):
-                predicate = in_whitelisted_context(**test_case.kwargs)
+                predicate = in_whitelist(**test_case.kwargs)
 
             with self.subTest(test_case=test_case):
                 self.assertTrue(predicate(test_case.ctx))
 
     def test_predicate_raises_exception_for_non_whitelisted_context(self):
-        """The predicate should raise `InWhitelistedContextCheckFailure` for a non-whitelisted context."""
+        """The predicate should raise `InWhitelistCheckFailure` for a non-whitelisted context."""
         test_cases = (
-            # Failing check with `redirect_channel`
+            # Failing check with `redirect`
             WhitelistedContextTestCase(
                 kwargs={
-                    "whitelisted_categories": self.whitelisted_categories,
-                    "whitelisted_channels": self.whitelisted_channels,
-                    "whitelisted_roles": self.whitelisted_roles,
-                    "redirect_channel": self.bot_commands.id,
+                    "categories": self.categories,
+                    "channels": self.channels,
+                    "roles": self.roles,
+                    "redirect": self.bot_commands.id,
                 },
                 ctx=helpers.MockContext(channel=self.non_whitelisted_channel, author=self.non_staff_member)
             ),
 
-            # Failing check without `redirect_channel`
+            # Failing check without `redirect`
             WhitelistedContextTestCase(
                 kwargs={
-                    "whitelisted_categories": self.whitelisted_categories,
-                    "whitelisted_channels": self.whitelisted_channels,
-                    "whitelisted_roles": self.whitelisted_roles,
+                    "categories": self.categories,
+                    "channels": self.channels,
+                    "roles": self.roles,
                 },
                 ctx=helpers.MockContext(channel=self.non_whitelisted_channel, author=self.non_staff_member)
             ),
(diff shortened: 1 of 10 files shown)


SebastiaanZ:
> I dislike splitting up decorator calls into multiple lines and the fact that it's practically impossible to use it as one line now, even for the simplest of calls, rubs me the wrong way.
>     1. Remove the `whitelisted_` prefix from kwarg names. I think redundant given the decorator name.

I don't mind that a decorator spans over multiple lines, as the kwarg names make explicit what each individual kwarg does and it's highly readable. From that perspective, I don't think they're necessarily redundant as such a decorator could have other kwargs that alter the behavior but do not define a whitelist.

However, can live with changing them and I already anticipated that people would require it. It's just that I like this version better so decided to go with this instead. (I actually had the shorter kwarg names initially, but I like how explicit the current names are.)

> 
>     2. In most cases, there is a redirect channel and it's the same value for all of them (`#bot-commands`) . Furthermore, it's now impossible to have multiple redirect channels. Previously, it used the whitelisted channels as the redirect channels. How about this:

I don't like that and picking a single, dedicated redirect channel is intended. The main reason is that if we want to redirect members to use a command somewhere instead of in the current context, it should be bot-commands. While the command may technically whitelisted in other channels, that doesn't make them the right place to be redirected to just to perform a bot command as those other channels are not bot-command channels and may have conversations of their own going.

In short, sending users to another whitelisted channel just to perform a bot command will typically result in the member using that channel in an off-topic manner. And the bot reply even told them to go there.

As a side note, we were not actually using multiple redirect channels at the moment. Instead, we added a new kwarg, `hidden_channels`, just to prevent non-botcommands channel from showing up in the redirect list to prevent "misredirection" to a channel where a command is allowed but only in the context of conversations actually happening there; not for random bot commands.

The reason I did not make `bot-commands` the defacto default is because a command may only be whitelisted by role (meaning that a member without that role should also be unable to use it in bot-commands) and/or to allow for a different channel that we consider appropriate to divert bot command invocation to. (As this is only `@Python`, I can't think of a reason for it, but we could want to send users to `seasonalbot-commands` instead.)

>        
>        Furthermore, consider that a wrapper for the decorator could be created so that there is one decorator that has redirects disabled by default and the other which uses the channel whitelist.

We could have a "partial" that already has `bot-commands` as the redirect channel to avoid having it as a default (since that may cause surprises if we whitelist a context only by things like roles). I'm not entirely sure if it really adds anything. I like how explicit the current call is and we'd have to basically move that information into the decorator name to keep it explicit (`in_whitelisted_context_with_bot_commands_redirect` doesn't sound sexy, but I'm also unsure how of a shorter name that is as explicit.

MarkKoz:
> I don't mind that a decorator spans over multiple lines,

A difference of opinion, but I think it's quite unsightly.

> the kwarg names make explicit what each individual kwarg does and 

Being explicit is good but it can become overly verbose, like here.

> it's highly readable

In terms of understanding what it does, I agree. However, I think it's length/size due to its verbosity makes it difficult to read as it really catches one's attention, whereas one could be focusing on more important things like the function signature or body.

> such a decorator could have other kwargs that alter the behavior but do not define a whitelist.

I don't think it should compromise (well, it's a compromise in my opinion anyway) to plan for potential future arguments. When those new arguments come, then we can reconsider if we cannot come up with names that can't be confused with the whitelist args; it's not really an expensive change to make down the road.

---

I think this looks much better (not as confident about renaming `redirect_channels` though):

```py
@in_whitelist(
    channels=EVAL_CHANNELS, 
    categories=EVAL_CATEGORIES, 
    roles=EVAL_ROLES, 
    redirect=Channels.bot_commands
)
```

And if we make the first 3 not kwarg-only:


```py
@in_whitelist(EVAL_CHANNELS, EVAL_CATEGORIES, EVAL_ROLES, redirect=Channels.bot_commands)
```

Now even the longest of invocations fits on one line, and it's still readable to me. I don't need kwargs because the constant names already tell me what the types are. Maybe if it was 3 tuple literals instead it'd be harder to look at (i.e. to visually spot the boundaries of the tuples). However, I think it should be up to the programmer to decide whether or not the code is readable without explicit kwarg names.

A different direction to take this in would be separate decorators for roles, channels, and categories. Didn't think about that idea much but just want to throw it out there.

---

> In short, sending users to another whitelisted channel just to perform a bot command will typically result in the member using that channel in an off-topic manner. And the bot reply even told them to go there.

I can see why from a common sense point of view, this is a bad design. I suggested it because, in practice, 4/6 uses of the decorator only have bot commands as the whitelisted channel. Given my distaste with the verbosity of the decorator, I'd want to address that common use case somehow to shorten the invocation. Maybe this is not the right solution after all, but we should at least look into doing _something_.

> As a side note, we were not actually using multiple redirect channels at the moment.

True, but I do recall it being used in the past for some seasonalbot stuff. If you believe there'd never be a case where we need to have multiple redirect channels, then it's fine to restrict it to one.

> The reason I did not make bot-commands the defacto default is because a command may only be whitelisted by role (meaning that a member without that role should also be unable to use it in bot-commands) and/or to allow for a different channel that we consider appropriate to divert bot command invocation to.

Well if it's either nothing or `#bot-commands`, then how about making it a boolean? I can't think of any other channel to redirect to either, but it would limit us down the road should the need arise.

---

> However, I'm not really sure how we would name such a "partial" that is as explicit as having the redirect kwarg there but makes the use more convenient by being shorted.

I agree it is tricky to name. Giving it a long name would defeat the purpose. How about `in_bot_commands`? The intent would be to use it for commands that only need `#bot-commands` whitelisted, though technically other channels could be specified since the redirect is always implicitly whitelisted. However, I concede that if someone doesn't know `bot_commands` is a channel, then that name would be unclear. Furthermore, it'd be unclear it still allows role and category whitelists. I think the boolean idea for the redirect is more appealing.

SebastiaanZ:
> And if we make the first 3 not kwarg-only:
> 
> ```python
> @in_whitelist(EVAL_CHANNELS, EVAL_CATEGORIES, EVAL_ROLES, redirect=Channels.bot_commands)
> ```
> 
> Now even the longest of invocations fits on one line, and it's still readable to me. I don't need kwargs because the constant names already tell me what the types are. Maybe if it was 3 tuple literals instead it'd be harder to look at (i.e. to visually spot the boundaries of the tuples). However, I think it should be up to the programmer to decide whether or not the code is readable without explicit kwarg names.

I'm not a fan of the the proposed API allowing positional arguments, as it's very easy to get it wrong and, if you do, it will fail silently. I don't think allowing optional arguments with an identical type pattern as positional arguments is a good idea.

Take this alternative call, for example:

 ```py
@in_whitelist(EVAL_CHANNELS, EVAL_ROLES, EVAL_CATEGORIES, redirect=Channels.bot_commands)
```

This looks entirely reasonable when you're reading it and the names of the constants give the impression that it's easy to understand which argument does what. In reality, the `EVAL_ROLES` and `EVAL_CATEGORIES` argument are accidentally swapped making it so that the container of role IDs will be used for the category check and the container with category IDs will be used for the role check.

This will fail silently, as they are both just containers of integers.

An other easy mistake is when we want to remove one of the three whitelists from the decorator; it's very easy to go something like this

```py
@in_whitelist(EVAL_CHANNELS, EVAL_CATEGORIES, EVAL_ROLES, redirect=Channels.bot_commands)

```

to something like this:

```py
@in_whitelist(EVAL_CHANNELS, EVAL_ROLES, redirect=Channels.bot_commands)
```

when you want to remove the category whitelist. It looks like it should work but in reality you've just shifted the roles whitelist to the category position and it will break silently again.

Obviously, we could shift the burden of detecting this to the code review process, but just as it's easy to get these wrong for the original contributors, it's also easy to read over it for a reviewer. As it's a "fail silent", the only other way to catch it is by thoroughly testing it and noticing the difference in behavior. However, while I would very much like for us to test reviews thoroughly, I don't think it's unlikely for someone to miss this in any review larger than just specifically adding the decorator.

All in all, I'm not a fan of the proposed API and I don't think the argument of saving a few lines is strong enough to dismiss these API considerations. I know that made multi-line decorator invocations are a common pet peeve, but I think the API is just a way more important factor than avoiding a decorator that may take up multiple lines if you use all the arguments at all costs.

(I'm obviously shortening the kwarg names by dropping the `whitelisted` prefix and I'll adapt the shorted `in_whitelist` name for the decorator as well. That will probably bring "line break relief" in a lot of the locations where it's used, as they don't pass all three whitelists.)

lemonsaurus:
Okay, just to chime in here with some quick input:
- Shortening the kwarg names to drop whitelisted prefix is nice.
- `in_whitelist` is a good decorator name.
- I agree with @SebastiaanZ about positional arguments being too error prone to be justified.

> The reason I did not make bot-commands the defacto default is because a command may only be whitelisted by role (meaning that a member without that role should also be unable to use it in bot-commands) and/or to allow for a different channel that we consider appropriate to divert bot command invocation to. (As this is only @Python, I can't think of a reason for it, but we could want to send users to seasonalbot-commands instead.)

I'm not sure I buy this. Why would we use this decorator to check only roles? We already have `with_role` for that, do we not? It might even be worth considering whether to assert that at least one of the channel lists has been passed, to prevent people from using it as a role filter.

I think I agree with @MarkKoz here that if we're gonna be redirecting to #bot-commands nine out of ten times anyway, then it would be better to explicitely set redirect to None or something in the few cases where we don't want that behavior, and of course allow a different redirect channel for the cases where we'd want to redirect somewhere else, which _might_ happen in regards to events.

SebastiaanZ:
> We already have `with_role` for that, do we not?

Maybe we should not have.

> So my suggestion is to default to #bot-commands, but to allow the user to pass either None for no redirect message (instead, some generic "you can't use this command" message would be shown), or some other channel ID to redirect to that instead.

Sounds good, that was the direction I was heading as well, so I agree.

lemonsaurus:
> Maybe we should not have. An in_whitelist(roles=ROLES) would do as well.

maybe. `with_role` arguably has better readability for this case, and it sounds like handling this well introduces a little more complexity to the decorator.. but I'm not violently opposed to generalizing around this decorator, I guess.

SebastiaanZ:
> maybe. `with_role` arguably has better readability for this case, and it sounds like handling this well introduces a little more complexity to the decorator.. but I'm not violently opposed to generalizing around this decorator, I guess.

Why would it add complexity?

The only reason I currently see that would warrant additional complexity is if we'd want to raise a specific exception so that our error handler could handle the check failure in a different. However, in that case, we could define a specific decorator for that behavior. We're not currently doing that.

tests/bot/test_decorators.py
```diff
@@ -0,0 +1,115 @@
...
+        cls.staff_member = helpers.MockMember(roles=(cls.staff_role,))
+
+        cls.channels = (cls.bot_commands.id,)
+        cls.categories = (cls.help_channel.category_id,)
+        cls.roles = (cls.staff_role.id,)
+
+    def test_predicate_returns_true_for_whitelisted_context(self):
+        """The predicate should return `True` if a whitelisted context was passed to it."""
+        test_cases = (
+            # Commands issued in whitelisted channels by members without whitelisted roles
```
MarkKoz, line 33: You should add these comments to the test cases so they can be shown as the subtest message.

SebastiaanZ: I will. I'll push them in half an hour from now.

PR #896, opened 2020-04-20
Answered help session statistics

jos-b:
Create a dictionary to gather whether help sessions were closed with no input from anyone but the original claimant.

This should help build an understanding of how many channels are unanswered.

Also introduces two emoji for the in use help channels signifying channels with input from users other than the claimant.

### DIFF
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -62,7 +62,8 @@ through our guide for [asking a good question]({ASKING_GUIDE_URL}).
 """
 
 AVAILABLE_EMOJI = "✅"
-IN_USE_EMOJI = "⌛"
+IN_USE_ANSWERED_EMOJI = "⌛"
+IN_USE_UNANSWERED_EMOJI = "⏳"
 NAME_SEPARATOR = "｜"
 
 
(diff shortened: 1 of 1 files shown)


bot/cogs/help_channels.py
```diff
@@ -133,6 +134,7 @@ def __init__(self, bot: Bot):
 
         # Stats
         self.claim_times = {}
+        self.unanswered = {}
```
MarkKoz, line 137: Can you leave a short comment explaining what's in this dict?

bot/cogs/help_channels.py
```diff
@@ -587,6 +595,17 @@ async def on_message(self, message: discord.Message) -> None:
...
+        if self.is_in_category(channel, constants.Categories.help_in_use):
+            if channel.id in self.unanswered:
+                claimant_id = self.help_channel_claimants[channel].id
+
+                if claimant_id != message.author.id:
+                    self.unanswered[channel.id] = False
+
+                    await channel.edit(
+                        name=f"{IN_USE_ANSWERED_EMOJI}{NAME_SEPARATOR}{self.get_clean_channel_name(channel)}"
+                    )
```
MarkKoz, line 607: This should be moved to a separate function. It will keep the function sizes more manageable and make it easier to write tests down the road.

bot/cogs/help_channels.py
jos-b: Typehinted + commented both this dictionary and the dictionary above.

PR #901, opened 2020-04-21
PEP command improvements

### LINKED ISSUE #891: PEP Command fixes
The PEP command uses the `isdigit` check for making sure the passed number is an int, this however is not always true; `isdecimal` or a `try except` should be used on the int conversion itself
https://github.com/python-discord/bot/blob/b7b48d0239afedbec31e8c05c90430a5b635cec0/bot/cogs/utils.py#L58-L60
PEP digit error

Furthermore the pep command gives no indication whether it was invoked correctly when making the time consuming web request to fetch data, this could be improved by triggering the typing event until the message is sent.

The command could also use caching like the `doc` command has, using the `async_cache` it implements to reduce the time it takes to return PEPs and prevent unnecessary requests. But this would probably need the PEP command to be a separate cog for more atomic code

ks129:
Resolve #891 
- Made `pep_number` type hint to `int` to avoid unnecessary manual converting.
- Added `ctx.trigger_typing` calling to show a user that bot is responding.

### Notes
- I made it automatically convert PEP number to `int` instead of manual (less secure) conversion. This save some lines of code.


Numerlor:
How do we feel about caching the web responses to lower the load and response time with the `async_cache` decorator as talked about in the issue?

While not in the scope of what was talked about in the issue the code of the function can also be improved in some parts; the logging could use a bit of info on the PEPs it's working with, the `found_pep` variable seems to be redundant and probably could find a bit more with a more thorough read if you feel like doing that.

ks129:
This might be good idea, but Core Devs have to decide it.

ks129:
@MarkKoz @Numerlor I implemented caching now and refactored PEP command.

bot/cogs/utils.py
```diff
@@ -305,7 +273,46 @@ async def send_pep_zero(self, ctx: Context) -> None:
         pep_embed.add_field(name="Created", value="13-Jul-2000")
         pep_embed.add_field(name="Type", value="Informational")
 
-        await ctx.send(embed=pep_embed)
+        return pep_embed
```
Numerlor, line 276: This can be safely made into a normal (static) method now

bot/cogs/utils.py
```diff
@@ -51,71 +53,37 @@ def __init__(self, bot: Bot):
...
-                pep_content = await response.text()
+        async with self.bot.http_session.get(self.peps_listing_api_url) as resp:
+            listing = await resp.json()
 
-                # Taken from https://github.com/python/peps/blob/master/pep0/pep.py#L179
-                pep_header = HeaderParser().parse(StringIO(pep_content))
+        for file in listing:
+            name = file["name"]
+            if name.startswith("pep-") and (name.endswith(".txt") or name.endswith(".rst")):
+                self.peps[int(name.split(".")[0].split("-")[1])] = file["download_url"]
```
Numerlor, line 73: Could use a bit more logging in this method if it stays.

bot/cogs/utils.py
```diff
@@ -51,71 +53,37 @@ def __init__(self, bot: Bot):
...
+        self.refresh_peps_urls.start()
 
-            if response.status == 200:
-                log.trace("PEP found")
-                found_pep = True
+    @loop(hours=3)
+    async def refresh_peps_urls(self) -> None:
+        """Refresh PEP URLs listing in every 3 hours."""
+        # Wait until HTTP client is available
+        await self.bot.wait_until_guild_available()
```
Numerlor, line 65: The guild is not needed in this method and the client is set on login, so `wait_until_ready` can be used instead of waiting for the guild

bot/cogs/utils.py
```diff
@@ -51,71 +53,37 @@ def __init__(self, bot: Bot):
...
 
-                pep_content = await response.text()
+        async with self.bot.http_session.get(self.peps_listing_api_url) as resp:
+            listing = await resp.json()
 
-                # Taken from https://github.com/python/peps/blob/master/pep0/pep.py#L179
-                pep_header = HeaderParser().parse(StringIO(pep_content))
+        for file in listing:
+            name = file["name"]
+            if name.startswith("pep-") and (name.endswith(".txt") or name.endswith(".rst")):
```
Numerlor, line 72: endswith can accept a tuple of strings to check so the or can be removed and a single endswith used

bot/cogs/utils.py
Numerlor, line 73: This line feels a bit busy, the splitting could be a bit split up on lines with names; or something like strip used for the surrounding chars

ks129:
@Numerlor Is this better now?

bot/cogs/utils.py
```diff
@@ -40,6 +42,20 @@
...
+    pep_embed = Embed(
+        title=f"**PEP 0 - Index of Python Enhancement Proposals (PEPs)**",
+        description=f"[Link](https://www.python.org/dev/peps/)"
+    )
+    pep_embed.set_thumbnail(url=ICON_URL)
+    pep_embed.add_field(name="Status", value="Active")
+    pep_embed.add_field(name="Created", value="13-Jul-2000")
+    pep_embed.add_field(name="Type", value="Informational")
+
+    return pep_embed
```
Numerlor, line 56: This should still stay as a static method inside of the class instead of being defined outside as it's only useful inside of that class.

bot/cogs/utils.py
```diff
@@ -40,6 +42,20 @@
 ICON_URL = "https://www.python.org/static/opengraph-icon-200x200.png"
 
 
+def get_pep_zero_embed() -> Embed:
+    """Send information about PEP 0."""
```
Numerlor, line 46: The docstring no longer describes the current behavior of the function.

bot/cogs/utils.py
```diff
@@ -250,19 +205,84 @@ async def vote(self, ctx: Context, title: str, *options: str) -> None:
...
+            pep_embed, success = await self.get_pep_embed(pep_number)
         await ctx.send(embed=pep_embed)
 
+        if success:
+            log.trace(f"PEP {pep_number} getting and sending finished successfully. Increasing stat.")
+            self.bot.stats.incr(f"pep_fetches.{pep_number}")
+
+    @async_cache(arg_offset=1)
+    async def get_pep_embed(self, pep_nr: int) -> Tuple[Embed, bool]:
+        """Fetch, generate and return PEP embed. Implement `async_cache`."""
```
Numerlor, line 249: I don't feel like the async cache is worth mentioning in the docstring

bot/cogs/utils.py
```diff
@@ -250,19 +205,84 @@ async def vote(self, ctx: Context, title: str, *options: str) -> None:
...
+        await self.bot.wait_until_ready()
+        log.trace("Started refreshing PEP URLs.")
+
+        async with self.bot.http_session.get(self.peps_listing_api_url) as resp:
+            listing = await resp.json()
+        log.trace("Got PEP URLs listing from GitHub API")
+
+        for file in listing:
+            name = file["name"]
+            name: str
```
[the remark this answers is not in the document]
ks129: Fixed in 36dac33 . I used this for PyCharm auto-suggesting and bug finding, but forget to remove it.

bot/cogs/utils.py
```diff
@@ -250,19 +205,84 @@ async def vote(self, ctx: Context, title: str, *options: str) -> None:
...
+            fields_to_check = ("Status", "Python-Version", "Created", "Type")
+            for field in fields_to_check:
+                # Check for a PEP metadata field that is present but has an empty value
+                # embed field values can't contain an empty string
+                if pep_header.get(field, ""):
+                    pep_embed.add_field(name=field, value=pep_header[field])
+            return pep_embed, True
+        else:
+            log.trace(f"The user requested PEP {pep_nr}, but the response had an unexpected status code: "
+                      f"{response.status}.\n{response.text}")
```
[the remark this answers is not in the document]
ks129: Fixed in 248ce24

ks129:
@Numerlor Is the current way is better?

bot/utils/cache.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def decorator(function: Callable) -> Callable:
+        """Define the async_cache decorator."""
+        @functools.wraps(function)
+        async def wrapper(*args) -> Any:
+            """Decorator wrapper for the caching logic."""
+            key = ':'.join(str(args[arg_offset:]))
+
+            value = async_cache.cache.get(key)
+            if value is None:
```
MarkKoz, line 25: If you're not actually using the value, use `in` instead to check if it's in the dict.
```suggestion
            if key in async_cache.cache:
```

bot/cogs/utils.py
```diff
@@ -261,7 +240,54 @@ async def send_pep_zero(self, ctx: Context) -> None:
...
+            for field in fields_to_check:
+                # Check for a PEP metadata field that is present but has an empty value
+                # embed field values can't contain an empty string
+                if pep_header.get(field, ""):
+                    pep_embed.add_field(name=field, value=pep_header[field])
+            return pep_embed
+        else:
+            log.trace(
+                f"The user requested PEP {pep_nr}, but the response had an unexpected status code: "
+                f"{response.status}.\n{response.text}"
```
MarkKoz, line 283: Please avoid putting newlines in logs; it makes them harder to read.

bot/cogs/utils.py
```diff
@@ -261,7 +240,54 @@ async def send_pep_zero(self, ctx: Context) -> None:
...
+            )
+
+            error_message = "Unexpected HTTP error during PEP search. Please let us know."
+            await ctx.send(
+                embed=Embed(title="Unexpected error", description=error_message, colour=Colour.red())
+            )
```
MarkKoz, line 289: Bit nicer to avoid putting args on a new line when possible IMO.
```suggestion
            embed = Embed(title="Unexpected error", description=error_message, colour=Colour.red())
            await ctx.send(embed=embed)
```

bot/cogs/utils.py
```diff
@@ -261,7 +240,54 @@ async def send_pep_zero(self, ctx: Context) -> None:
...
+
+    @async_cache(arg_offset=1)
+    async def get_pep_embed(self, pep_nr: int, ctx: Context) -> Optional[Embed]:
+        """Fetch, generate and return PEP embed."""
+        if pep_nr not in self.peps:
+            log.trace(f"PEP {pep_nr} was not found")
+            not_found = f"PEP {pep_nr} does not exist."
+            await ctx.send(
+                embed=Embed(title="PEP not found", description=not_found, colour=Colour.red())
+            )
```
MarkKoz, line 253:
```suggestion
            embed=Embed(title="PEP not found", description=not_found, colour=Colour.red())
            await ctx.send(embed=embed)
```

bot/cogs/utils.py
```diff
@@ -250,8 +191,46 @@ async def vote(self, ctx: Context, title: str, *options: str) -> None:
...
+        async with self.bot.http_session.get(self.peps_listing_api_url) as resp:
+            listing = await resp.json()
+        log.trace("Got PEP URLs listing from GitHub API")
+
+        for file in listing:
+            name = file["name"]
+            if name.startswith("pep-") and name.endswith((".rst", ".txt")):
+                pep_number = name.replace("pep-", "").split(".")[0]
+                self.peps[int(pep_number)] = file["download_url"]
+        log.info("Successfully refreshed PEP URLs listing.")
```
MarkKoz, line 212: I find it harder to read a line that's outside and right under a block

```suggestion
                self.peps[int(pep_number)] = file["download_url"]

        log.info("Successfully refreshed PEP URLs listing.")
```

bot/cogs/utils.py
```diff
@@ -250,8 +191,46 @@ async def vote(self, ctx: Context, title: str, *options: str) -> None:
...
+    @loop(hours=3)
+    async def refresh_peps_urls(self) -> None:
+        """Refresh PEP URLs listing in every 3 hours."""
+        # Wait until HTTP client is available
+        await self.bot.wait_until_ready()
+        log.trace("Started refreshing PEP URLs.")
+
+        async with self.bot.http_session.get(self.peps_listing_api_url) as resp:
+            listing = await resp.json()
+        log.trace("Got PEP URLs listing from GitHub API")
```
MarkKoz, line 205:
```suggestion
            listing = await resp.json()

        log.trace("Got PEP URLs listing from GitHub API")
```

bot/cogs/utils.py
```diff
@@ -250,8 +191,46 @@ async def vote(self, ctx: Context, title: str, *options: str) -> None:
         for reaction in options:
             await message.add_reaction(reaction)
 
-    async def send_pep_zero(self, ctx: Context) -> None:
-        """Send information about PEP 0."""
+    # PEPs area
+
+    @loop(hours=3)
+    async def refresh_peps_urls(self) -> None:
```
MarkKoz, line 197: Instead of polling, how about trying to get from the existing cache first and then sending a request if a PEP doesn't exist? A more complex variation of this idea is to store PEPs that couldn't be found in a separate cache which gets cleared once in a while. This would avoid spamming HTTP requests if the same bad PEP is requested within a short time frame, but I don't really think such complexity is necessary.

ks129:
@MarkKoz Addressed changes.

bot/cogs/utils.py
```diff
@@ -244,15 +245,20 @@ def get_pep_zero_embed() -> Embed:
...
+            if pep_nr not in self.peps and (self.last_refreshed_peps + timedelta(minutes=30)) > datetime.now():
+                log.trace(f"PEP {pep_nr} was not found")
+                not_found = f"PEP {pep_nr} does not exist."
+                embed = Embed(title="PEP not found", description=not_found, colour=Colour.red())
+                await ctx.send(embed=embed)
+                return
+            elif pep_nr not in self.peps:
+                await self.refresh_peps_urls()
+            else:
+                break
```
MarkKoz, line 261: It's confusing to follow the control flow of this loop. This could be simplified without a loop: 

```py
if pep_nr not in self.peps and (self.last_refreshed_peps + timedelta(minutes=30)) <= datetime.now():
    await self.refresh_peps_urls()

if pep_nr not in self.peps:
    ...
    await ctx.send(embed=embed)
    ...
```

bot/cogs/utils.py
```diff
@@ -250,8 +192,48 @@ async def vote(self, ctx: Context, title: str, *options: str) -> None:
         for reaction in options:
             await message.add_reaction(reaction)
 
-    async def send_pep_zero(self, ctx: Context) -> None:
-        """Send information about PEP 0."""
+    # PEPs area
```
Numerlor, line 195: this can use `# region: PEP` and later `# endregion` to be collapsible in some editors

ks129:
@Numerlor I don't want to split `get_pep_embed` much. I made small changes there, is this better now?

ks129:
@Numerlor I made class for cache and now get_pep_embed isn's so big function, but I think this is not enough big feature for it's own cog.

bot/exts/utils/utils.py
```diff
@@ -50,71 +55,11 @@ def __init__(self, bot: Bot):
 
         self.base_pep_url = "http://www.python.org/dev/peps/pep-"
         self.base_github_pep_url = "https://raw.githubusercontent.com/python/peps/master/pep-"
+        self.peps_listing_api_url = "https://api.github.com/repos/python/peps/contents?ref=master"
```
Numerlor, line 58: Making these class constants would help in making them easier to track of in case anything changes

Numerlor (changes requested):
The suggestion wasn't necessarily a new cog, it could be a class in the utils module that'd provide one public method for the cog to use (send number, get embed/text)

bot/exts/utils/utils.py
```diff
@@ -246,8 +191,51 @@ async def vote(self, ctx: Context, title: clean_content(fix_channel_mentions=Tru
...
+        # Handle PEP 0 directly because it's not in .rst or .txt so it can't be accessed like other PEPs.
+        if pep_number == 0:
+            pep_embed = self.get_pep_zero_embed()
+        else:
+            if not await self.validate_pep_number(ctx, pep_number):
+                return
+
+            pep_embed = await self.get_pep_embed(ctx, pep_number)
+
+        if pep_embed:
```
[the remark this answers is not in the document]
ks129: I didn't split that to different functions without reason: In some reason, when putting stats increasing to any other function, then this don't actually increase stat sometimes, and I need to only increase stat when this success.

bot/exts/utils/utils.py
ks129: Addressed in 86f2024

bot/utils/cache.py
```diff
@@ -0,0 +1,40 @@
...
+    def __call__(self, max_size: int = 128, arg_offset: int = 0) -> Callable:
+        """Decorator for async cache."""
+
+        def decorator(function: Callable) -> Callable:
+            """Define the async cache decorator."""
+
+            @functools.wraps(function)
+            async def wrapper(*args) -> Any:
+                """Decorator wrapper for the caching logic."""
+                key = ':'.join(str(args[arg_offset:]))
```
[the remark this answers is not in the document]
ks129: Addressed in c58d68e

bot/utils/cache.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Once the cache exceeds the maximum size, keys are deleted in FIFO order.
+
+    An offset may be optionally provided to be applied to the coroutine's arguments when creating the cache key.
+    """
+
+    def __init__(self):
+        self._cache = OrderedDict()
+
+    def __call__(self, max_size: int = 128, arg_offset: int = 0) -> Callable:
```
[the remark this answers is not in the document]
ks129: Max size should be set only once, so I moved this to `__init__` (90103c5).

bot/exts/utils/utils.py
```diff
@@ -41,6 +43,9 @@
 
 ICON_URL = "https://www.python.org/static/opengraph-icon-200x200.png"
 
+# Async cache instance for PEPs
+async_cache = AsyncCache()
```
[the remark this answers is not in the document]
ks129: Addressed in 03560d8

bot/exts/info/doc.py
```diff
@@ -65,34 +65,8 @@
...
-            if value is None:
-                if len(async_cache.cache) > max_size:
-                    async_cache.cache.popitem(last=False)
-
-                async_cache.cache[key] = await function(*args)
-            return async_cache.cache[key]
-        return wrapper
-    return decorator
+# Async cache instance for docs cog
+async_cache = AsyncCache()
```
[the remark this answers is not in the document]
ks129: Addressed in 03560d8

ks129:
@Numerlor changes is made

bot/exts/utils/utils.py
```diff
@@ -257,7 +243,70 @@ async def send_pep_zero(self, ctx: Context) -> None:
...
+            # embed field values can't contain an empty string
+            if pep_header.get(field, ""):
+                pep_embed.add_field(name=field, value=pep_header[field])
+
+        return pep_embed
+
+    @pep_cache(arg_offset=2)
+    async def get_pep_embed(self, ctx: Context, pep_nr: int) -> Optional[Embed]:
+        """Fetch, generate and return PEP embed. When any error occur, use `self.send_pep_error_embed`."""
+        response = await self.bot.http_session.get(self.peps[pep_nr])
```
Numerlor, line 288: We shouldn't need the error handling code below if we raise for status here and let it propagate to the user

bot/exts/utils/utils.py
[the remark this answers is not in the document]
Numerlor: The stat code should be left untouched, but currently message sending is handled over 3 methods: in this one, in `get_pep_embed` and in `validate_pep_number`. From the two checking coros it's not immediately apparent that they will send message from their names and the sending doesn't need to be handled inside of them, as all their paths where the messages are sent are already being checked in here.

bot/exts/utils/utils.py
ks129: You mean setting raise_for_status to `True` and when this send any error?

Numerlor: Ah nevermind that, this shouldn't go directly to the user since we don't have the proper handling for it in the error handler.
Though raising for status and logging the exception (as a warning or an exception) would cover all the status errors along with things like connection errors which are currently unhandled

ks129: I think it's not necessary, as this already logs response status code and I think connection error will raise error anyway?

bot/exts/utils/utils.py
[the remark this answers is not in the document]
ks129: Resolved in c9ffb11.

ks129:
@Numerlor Changes is now made.

Senjan21 (approved):
Looks good, lets merge it.

PR #910, opened 2020-04-25
Converters: handle ValueError when year for duration is out of range

### LINKED ISSUE #906: Handle out of range year in Duration converter
Sentry Issue: [BOT-39](https://sentry.io/organizations/python-discord/issues/1611103057/?referrer=github_integration)

As documented [here](https://docs.python.org/3/library/datetime.html#datetime.date), it has to be `MINYEAR <= year <= MAXYEAR` which is effectively `1 <= year <= 9999`

```
ValueError: year 104039 is out of range
  File "discord/ext/commands/core.py", line 425, in _actual_conversion
    ret = await instance.convert(ctx, argument)
  File "bot/converters.py", line 220, in convert
    return now + delta
  File "dateutil/relativedelta.py", line 405, in __radd__
    return self.__add__(other)
  File "dateutil/relativedelta.py", line 387, in __add__
    ret = (other.replace(**repl)
```

MarkKoz:
Fixes #906

`datetime` objects only support a year up to 9999. To prevent the exception from propagating, it is now caught and the converter raises a different error which will be handled elsewhere to ultimately display an message to the user that their argument has some stupidly high duration.


bot/converters.py
```diff
@@ -217,7 +217,10 @@ async def convert(self, ctx: Context, duration: str) -> datetime:
         delta = relativedelta(**duration_dict)
         now = datetime.utcnow()
 
-        return now + delta
+        try:
+            return now + delta
+        except ValueError:
+            raise BadArgument(f"`{duration}` results in a datetime outside the supported range.")
```
aeros, line 223: Since there seems to be two ways for the custom `BadArgument` exception to be raised from this method, I would recommend explicitly testing for each of them in `test_converters.py` with `assertRaisesRegex()` to ensure the correct one occurs. There seems to be an existing test for the former in: 

https://github.com/python-discord/bot/blob/6e97a6af9a72286b9ec8102e1abd64f634fe8d92/tests/bot/test_converters.py#L193-L199.

(Note that the above *msg* kwarg in `assertRaises()` does not test the message itself; instead, it customizes the message displayed when the assertion fails. That can be quite useful for diagnosing test failures, but doesn't help much with ensuring the correct exception message occurred.)

MarkKoz: Good catch, thanks. Didn't realises there were even any tests for converters.

tests/bot/test_converters.py
```diff
@@ -16,7 +16,7 @@
 )
 
 
-class ConverterTests(unittest.TestCase):
+class ConverterTests(unittest.IsolatedAsyncioTestCase):
```
aeros, line 19: Good idea w/ using `IsolatedAsyncioTestCase`, I was actually just revisiting the PR and about to suggest using it instead of `asyncio.run()`. `asyncio.run()` has a fair bit of additional overhead that isn't needed for testing purposes, especially when you call it multiple times (it creates and closes a new event loop each time it's used, in addition to some other cleanup calls related to tasks, async generators, and in 3.9 the default executor).

PR #923, opened 2020-05-01
Remove the mention command

MarkKoz:
It was made obsolete by a new Discord feature. Users can be granted a permission to mention a role despite the role being set as non-mentionable.

### DIFF
--- a/bot/cogs/utils.py
+++ b/bot/cogs/utils.py
@@ -2,19 +2,16 @@ import difflib
 import logging
 import re
 import unicodedata
-from asyncio import TimeoutError, sleep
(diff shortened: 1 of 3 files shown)


kwzrd:
Should the config for this also be removed? Doesn't look like it's being used anywhere.

https://github.com/python-discord/bot/blob/8cb6fe392a17d7f682af60b30a23cb143a27e87c/bot/constants.py#L553-L557

https://github.com/python-discord/bot/blob/8cb6fe392a17d7f682af60b30a23cb143a27e87c/config-default.yml#L510-L512

PR #927, opened 2020-05-06
Truncate internal eval.

### LINKED ISSUE #907: Truncate internal eval responses over 2000 characters
Sentry Issue: [BOT-1Z](https://sentry.io/organizations/python-discord/issues/1541819058/?referrer=github_integration)

This could be handled by create a paste link like snekbox does or by paginating the response.

```
In content: Must be 2000 or fewer in length.
  File "bot/cogs/eval.py", line 197, in eval
    await self._eval(ctx, code)
  File "bot/cogs/eval.py", line 174, in _eval
    await ctx.send(f"```py\n{out}```", embed=embed)
  File "discord/http.py", line 225, in request
    raise HTTPException(r, data)
```

Numerlor:
Fixes #907

Implements a new util function for uploading strings to paste service.
Truncates internal eval responses in a similar fashion to snekbox's eval and updates snekbox to use the new function.


bot/utils/__init__.py
```diff
@@ -1,9 +1,62 @@
...
+                f"trying again ({attempt}/{FAILED_REQUEST_ATTEMPTS})."
+            )
+            continue
+        elif "key" in response_json:
+            log.info(f"Successfully uploaded contents to paste service behind key {response_json['key']}.")
+            return URLs.paste_service.format(key=response_json['key']) + extension
+        log.warning(
+            f"Got unexpected JSON response from paste service: {response_json}\n"
+            f"trying again ({attempt}/{FAILED_REQUEST_ATTEMPTS})."
+        )
```
MarkKoz, line 62: I think this should be outside the loop.

MarkKoz: Never mind, misread.

MarkKoz:
That is an issue on my end but I cannot resolve it. I won't be able to test this PR myself.

bot/cogs/eval.py
```diff
@@ -171,6 +172,20 @@ async def func():  # (None,) -> Any
...
         out, embed = self._format(code, res)
+        if len(out) > 1500 or out.count("\n") > 15:
+            paste_link = await send_to_paste_service(self.bot.http_session, out, extension="py")
+            if paste_link is not None:
+                paste_text = f"full contents at {paste_link}"
+            else:
+                paste_text = "failed to upload contents to paste service."
+
+            await ctx.send(
+                f"```py\n{out[:1500]}\n```"
```
MarkKoz, line 183: This should also truncate it to 15 lines.

bot/cogs/eval.py
```diff
@@ -197,6 +222,16 @@ async def eval(self, ctx: Context, *, code: str) -> None:
...
 
 
+def find_nth_occurrence(string: str, substring: str, n: int) -> Optional[int]:
+    """Return index of `n`th occurrence of `substring` in `string`, or None if not found."""
+    index = 0
+    for _ in range(n):
+        index = string.find(substring, index+1)
+        if index == -1:
+            return None
+    return index
```
lemonsaurus, line 232: This is generic enough that I think it might be better to keep it in utils. Might be useful somewhere else.

bot/utils/__init__.py
```diff
@@ -13,6 +22,50 @@ class CogABCMeta(CogMeta, ABCMeta):
...
 
+async def send_to_paste_service(http_session: ClientSession, contents: str, *, extension: str = "") -> Optional[str]:
+    """
+    Upload `contents` to the paste service.
+
+    `http_session` should be the current running ClientSession from aiohttp
+    `extension` is added to the output URL
+
+    When an error occurs, `None` is returned, otherwise the generated URL with the suffix.
+    """
```
lemonsaurus, line 33: I don't understand why this code should be in `__init__.py`. Let's move this to an appropriate file instead.

Numerlor: I didn't feel a need to create a new file for a simple function like this. Are you suggesting for example a path like `bot/utils/paste/send_to_paste_service` ?

lemonsaurus: We shouldn't be defining functions in the `__init__.py` file at all, it's an antipattern. I'm not suggesting that it needs to have a whole file just for itself, it'd be fine for it to share a file with other similar functions. `paste.py` would be okay, but even something more generic like `services.py` would be fine by me. Then we can say oh okay, here are helpers for interacting with external services.

But my primary gripe is just that I don't think anything like this should ever live inside an init file.

bot/cogs/eval.py
Numerlor: Where would you suggest putting this, next to `pad_base64` in the utils' init or an another file?

lemonsaurus: you really gonna ask me if I think it should be in the init after I just went on a whole rant about how nothing should be in the init? :grin: 

I don't even think `pad_base64` or `CogABCMeta` should be there. I never would've put it there myself. I wouldn't mind if you took this opportunity to clean that up as a bit of kaizen for this PR.

Numerlor: These are really general helper functions, I'm not a big fan of say, `utils/utils` or `utils/helpers` as utils is fairly indescriptive by itself, but can't think of something better where to group them.

lemonsaurus: `utils.py`, `helpers.py`, or even `general.py` are all fine by me. And of course, then you'd import them into the `__init__.py` (that's one thing you absolutely _should_ do in the init) so when someone wants to import them, they'd still just do `from bot.utils import whatever`.

kosayoda (approved):
Stone solid.

PR #930, opened 2020-05-07
Add tests for the antimalware cog

### LINKED ISSUE #555: Write unit tests for `bot/cogs/antimalware.py`
Write unit tests for [`bot/cogs/antimalware.py`](../blob/master/bot/cogs/antimalware.py).

## Implementation details

## Additional information

MrGrote:
closes #555


tests/bot/cogs/test_antimalware.py
```diff
@@ -0,0 +1,146 @@
...
+
+        with self.assertLogs(logger=logger, level="INFO") as logs:
+            await self.cog.on_message(self.message)
+        self.assertEqual(
+            [
+                f"INFO:{MODULE}:"
+                f"User '{self.message.author}' ({self.message.author.id}) "
+                f"uploaded blacklisted file(s): {blocked_extensions_str}"
+            ],
+            logs.output)
```
MarkKoz, line 136: My comments on logging also apply here.

tests/bot/cogs/test_antimalware.py
```diff
@@ -0,0 +1,146 @@
...
+
+    async def test_message_with_illegal_attachment_logs(self):
+        """Deleting a message with an illegal attachment should result in a log."""
+        attachment = MockAttachment(filename="python.asdfsff")
+        self.message.attachments = [attachment]
+
+        file_extensions = {splitext(attachment.filename.lower())[1] for attachment in self.message.attachments}
+        extensions_blocked = file_extensions - set(AntiMalwareConfig.whitelist)
+        blocked_extensions_str = ', '.join(extensions_blocked)
```
MarkKoz, line 125: Though I don't think the log's contents should be assert at all, I'll mention this anyway:

You shouldn't copy the same code into tests. If the cog had this code in a function, you'd be able to mock the function's return value and then assert your given return value was within the log string.

tests/bot/cogs/test_antimalware.py
```diff
@@ -0,0 +1,146 @@
...
+
+    async def test_python_file_redirect_embed(self):
+        """A message containing a .python file should result in an embed redirecting the user to our paste site"""
+        attachment = MockAttachment(filename="python.py")
+        self.message.attachments = [attachment]
+        self.message.channel.send = AsyncMock()
+
+        await self.cog.on_message(self.message)
+        args, kwargs = self.message.channel.send.call_args
+        embed = kwargs.pop("embed")
```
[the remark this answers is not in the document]
MrGrote: `pop` was used in `test_information.py` in a similar situation.

tests/bot/cogs/test_antimalware.py
```diff
@@ -0,0 +1,146 @@
...
+
+        self.assertEqual(args[0], f"Hey {self.message.author.mention}!")
+        self.assertEqual(embed.description, (
+            "**Uh-oh!** It looks like your message got zapped by our spam filter. "
+            "We currently don't allow `.txt` attachments, so here are some tips to help you travel safely: \n\n"
+            "• If you attempted to send a message longer than 2000 characters, try shortening your message "
+            "to fit within the character limit or use a pasting service (see below) \n\n"
+            "• If you tried to show someone your code, you can use codeblocks \n(run `!code-blocks` in "
+            f"{cmd_channel.mention} for more information) or use a pasting service like: "
+            f"\n\n{URLs.site_schema}{URLs.site_paste}"
```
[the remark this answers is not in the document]
MrGrote: I've been looking into this but since two of the strings use variables that aren't available at that point I was thinking of moving the embed creating to their own functions and mocking the return value of those.

tests/bot/cogs/test_antimalware.py
```diff
@@ -0,0 +1,159 @@
...
+
+class AntiMalwareCogTests(unittest.IsolatedAsyncioTestCase):
+    """Test the AntiMalware cog."""
+
+    def setUp(self):
+        """Sets up fresh objects for each test."""
+        self.bot = MockBot()
+        self.cog = antimalware.AntiMalware(self.bot)
+        self.message = MockMessage()
+        AntiMalwareConfig.whitelist = [".first", ".second", ".third"]
```
SebastiaanZ, line 21: Why are you monkeypatching this directly instead of using `unittest.mock.patch`? The downside of monkeypatching directly is that the monkeypatch will stay behind, as the object in memory is mutated. While `unittest.mock.patch` does the same, in principle, it restores the original after the fact. By providing an argument for `new`, we can determine what it should be patched with and it makes sure that the mock doesn't get passed as an attribute to test methods.

You could use `unittest.mock.patch` as a class decorator to patch this for the entire test case class and it will be cleaned up after the class ends. In this case, you're mutating that object in memory, which may result in unexpected values if other files/tests access it after this line of code has run. Another option would be to just patch it for the methods that rely on it. If there's more than one and the patch value is stable (which its locating in the `setUp` method implies), I'd just go for a class decorator.

```py
from unittest.mock import patch

MODULE = "bot.cogs.antimalware"

@patch(f"{MODULE}.AntiMalwareConfig.whitelist", new=[".first", ".second", ".third"])
class AntiMalwareCogTests(unittest.IsolatedAsyncioTestCase):
    """Test the AntiMalware cog."""
```

MrGrote: I had not considered this, pushed the changes.

PR #937, opened 2020-05-10
Make token detection more robust and completely rewrite its tests

### LINKED ISSUE #928: Token Check Raising for non-ASCII Characters
Due to the structure of the regex & validation for detecting Discord tokens, there is a chance that the filter can be triggered for values that cause the validation step to raise.

`λx.λy.x`, for example, matches the regex pattern but causes the [`is_valid_user_id`](https://github.com/python-discord/bot/blob/007f771b9cdf9de608920feea7f0e4a98b61fcab/bot/cogs/token_remover.py#L127) method's `base64.b64decode` call to raise a `ValueError` when attempting to decode the lambda character.

This can be fixed by doing one or both of:
   * Expanding the [`except` clause](https://github.com/python-discord/bot/blob/007f771b9cdf9de608920feea7f0e4a98b61fcab/bot/cogs/token_remover.py#L138) to also catch `ValueError`
   * Add minimum field widths to the [regex token](https://github.com/python-discord/bot/blob/007f771b9cdf9de608920feea7f0e4a98b61fcab/bot/cogs/token_remover.py#L30)

MarkKoz:
Major changes:

* Only match URL-safe Base64 characters in the regex
* Fix parsing of the timestamp; it was completely wrong since it was wrongly interpreting it as a snowflake
* User IDs must have ASCII digits only
* Split some functions to make testing easier

I rewrote practically all of the tests cause they were not implemented correctly. A particularly strange thing the tests did often was assert the return value of `on_message`. I added a new decorator to help with making autospecced patches.

Fixes #928
Fixes BOT-3X


MarkKoz:
Upon further inspection, the extant tests are not written correctly so I will have to fix them.

tests/bot/cogs/test_token_remover.py
```diff
@@ -1,131 +1,279 @@
...
+        )
 
-    def test_is_valid_user_id_is_false_for_alphabetic_content(self):
-        """A string decoding to alphabetic characters is not a valid user ID."""
-        # YWJj = base64(abc)
-        self.assertFalse(TokenRemover.is_valid_user_id('YWJj'))
+        for user_id, is_valid in subtests:
+            with self.subTest(user_id=user_id, is_valid=is_valid):
+                result = TokenRemover.is_valid_user_id(user_id)
+                self.assertIs(result, is_valid)
```
aeros, line 38: I'm not sure that I agree with the usage of `assertIs()` in this case rather than `assertEqual()`. Although in this case the result is the same, you typically want to use `assertIs()` to specifically convey that you're comparing object identity, which isn't relevant in this case.

tests/bot/cogs/test_token_remover.py
```diff
@@ -1,131 +1,279 @@
...
-        """A string decoding to numeric characters is a valid user ID."""
-        # MTIz = base64(123)
-        self.assertTrue(TokenRemover.is_valid_user_id('MTIz'))
+    def test_is_valid_user_id(self):
+        """Should correctly discern valid user IDs and ignore non-numeric and non-ASCII IDs."""
+        subtests = (
+            ("MTIz", True),  # base64(123)
+            ("YWJj", False),  # base64(abc)
+            ("λδµ", False),
+        )
```
aeros, line 33: I think we should expand upon the subtests to include user IDs that follow a more realistic format, perhaps using the real user IDs of a few staff members (since it's fully public information). Also, having something around 5 valid user IDs and 5 invalid ones would help to improve coverage (not literal code coverage %; I'm referring to coverage of possible inputs).

IMO, the ones might be a bit too unrealistic for testing purposes.

bot/cogs/token_remover.py
```diff
@@ -123,6 +138,8 @@ def is_maybe_token(cls, test_str: str) -> bool:
         if cls.is_valid_user_id(user_id) and cls.is_valid_timestamp(creation_timestamp):
             return True
 
+        return False
+
```
aeros, line 142: With there now being two separate ways to get a `False` return from `is_maybe_token`, we may want to consider adding some logging to clearly separate each case for debugging purposes. I'd also change it to an explicit `else` clause instead of the bare `return False`. For example:

```py
    @classmethod
    def is_maybe_token(cls, test_str: str) -> bool:
        """Check the provided string to see if it is a seemingly valid token."""
        try:
            user_id, creation_timestamp, hmac = test_str.split('.')
        except ValueError:
            log.debug(f"Invalid format in: {test_str}. Could not distinguish user ID, timestamp, and hmac.")
            return False

        if cls.is_valid_user_id(user_id) and cls.is_valid_timestamp(creation_timestamp):
            return True
        else:
            log.debug(f"Invalid user ID or timestamp in: {test_str}.")
            return False 
```

tests/bot/cogs/test_token_remover.py
aeros: Also, this may be more of a personal preference, but I've found that it's significantly easier to read and debug when the valid and invalid cases are separated into 2+ tests, and then using something like `assertTrue(result)` and `assertFalse(result)` for each of the subtests respectively. At a glance, it can sometimes be needlessly difficult to discern whether it should be true or false for that specific input, especially if the list of test inputs grow.

tests/bot/cogs/test_token_remover.py
```diff
@@ -1,131 +1,279 @@
...
+
+        self.assertEqual(return_value, log_message.format.return_value)
+        log_message.format.assert_called_once_with(
+            author=self.msg.author,
+            author_id=self.msg.author.id,
+            channel=self.msg.channel.mention,
+            user_id="MTIz",
+            timestamp="DN9R_A",
+            hmac="xxx",
+        )
```
aeros, line 235: What exactly is the purpose of this specific test being a coroutine method? It's not immediately clear to me.

(FYI: you can have non coroutine test methods defined in an `IsolatedAsyncioTestCase` class)

tests/bot/cogs/test_token_remover.py
```diff
@@ -1,131 +1,279 @@
...
+
+    @mock.patch.object(TokenRemover, "mod_log", new_callable=mock.PropertyMock)
+    @autospec("bot.cogs.token_remover", "log")
+    @autospec(TokenRemover, "delete_message", "format_log_message")
+    async def test_take_action(self, delete_message, format_log_message, logger, mod_log_property):
+        """Should delete the message and send a mod log."""
+        cog = TokenRemover(self.bot)
+        mod_log = mock.create_autospec(ModLog, spec_set=True, instance=True)
```
aeros, line 243: Could `mod_log` not just use the `@autospec` decorator since it allows any valid `**kwargs` for `patch.object()`?

E.g. `@autospec(TokenRemover, "mod_log", new_callable=mock.PropertyMock)` instead of having `@mock.patch.object(TokenRemover, "mod_log", new_callable=mock.PropertyMock)` and `mod_log = mock.create_autospec(ModLog, spec_set=True, instance=True)`? It would behave the same AFAICT, but it seems worthwhile to remove a line of boilerplate.

Note that I don't personally have much experience w/ making use of mocks to this degree in my own unit tests, so the above is more of a question as it is a suggestion.

MarkKoz: `new_callable` cannot be used if autospeccing. I could set `autospec=False` but at that point I felt like I should use the normal unittest decorators.

tests/bot/cogs/test_token_remover.py
MarkKoz: No reason any more. Forgot to remove it.

tests/bot/cogs/test_token_remover.py
MarkKoz: Well you wouldn't consider `if x == True:` correct, right? So why would `assertEqual` be correct when comparing booleans? Sure it works but I've learned it's not good practice to compare booleans and `None` with `==`.

tests/bot/cogs/test_token_remover.py
MarkKoz: I can get 5 valid ones but I don't know would be considered a "realistic" invalid value.

tests/bot/cogs/test_token_remover.py
MarkKoz: I agree on the readability point but I didn't feel it was worth splitting up given the low amount of test inputs and the simplicity of the test. If more inputs are added like you suggested, then I'll probably split it up.

bot/cogs/token_remover.py
```diff
@@ -150,7 +167,7 @@ def is_valid_timestamp(b64_content: str) -> bool:
         try:
             content = base64.urlsafe_b64decode(b64_content)
             snowflake = struct.unpack('i', content)[0]
-        except (binascii.Error, struct.error):
+        except (binascii.Error, struct.error, ValueError):
```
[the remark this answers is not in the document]
MarkKoz: Well, I suppose. This was intended to be a quick fix so I went with catching the exception.

bot/cogs/token_remover.py
```diff
@@ -135,7 +152,7 @@ def is_valid_user_id(b64_content: str) -> bool:
         try:
             content: bytes = base64.b64decode(b64_content)
             return content.decode('utf-8').isnumeric()
```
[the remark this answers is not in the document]
MarkKoz: Good catch. I will change this.

bot/cogs/token_remover.py
MarkKoz: I agree.

bot/cogs/token_remover.py
```diff
@@ -16,6 +16,10 @@
 
 log = logging.getLogger(__name__)
 
+LOG_MESSAGE = (
+    "Censored a seemingly valid token sent by {author} (`{author_id}`) in {channel},"
```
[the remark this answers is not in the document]
MarkKoz: Good catch. Also, the channel name is already being formatted (it uses `channel.mention`).

bot/cogs/token_remover.py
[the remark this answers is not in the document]
MarkKoz: Do you think regex would be clearer, `isdecimal() and isascii()` (not sure if that's right even, the character classes are confusing), or something comparing ordinal range for ASCII digits?

MarkKoz:
> 
> 
> I'm not sure how the `is_valid_timestamp` works as I couldn't find anything about how the token structure works but it doesn't seem to be behaving as expected.
> 
> With the `int` struct unpacking, the resulting number can be negative which just seems invalid if it's a timestamp.
> Then the "snowflake" + `TOKEN_EPOCH` gets passed to `snowflake_time` where the value gets right shifted into some small number and a datetime is returned from that, as that's supposed to work with the id that's in the first part of the token

You're right. I never looked into it and assumed it's correct. Two mistakes with it:

1. The int needs to be unpacked as big endian rather than little endian (this is why it was negative).
2. After the token epoch is added, we already have our Unix timestamp; it's wrong to interpret it as a snowflake

bot/cogs/token_remover.py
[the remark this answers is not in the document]
Numerlor: The bot logs will only include `<#ID>` which is less useful than a name with the id but I guess that's worth the tradeoff because then it'd result in the name being logged twice in the channel

bot/cogs/token_remover.py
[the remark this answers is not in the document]
Numerlor: Checking isdecimal and isascii should be clear while remaining faster than regex or validating it manually.
With the number string methods and isascii, only 0-9 will match

Numerlor:
> 2\. After the token epoch is added, we already have our Unix timestamp; it's wrong to interpret it as a snowflake

From my tests it looks like the decoded int is directly a valid timestamp, without adding the token epoch as shown above. I think `int.from_bytes` can also be used instead of the struct unpacking to be a bit clearer

tests/bot/cogs/test_token_remover.py
aeros: > Well you wouldn't consider if x == True: correct, right? So why would assertEqual be correct when comparing booleans? Sure it works but I've learned it's not good practice to compare booleans and None with ==.

Yeah, disregard that comment; you're correct. It's actually better to use `assertIs()` over `assertEqual()` when directly testing a boolean return, since the truthiness of the object (from `bool(object)`) could give rather misleading results if the function returns anything other than `True`, `False`, or `None` when it shouldn't. My preference is generally towards `assertTrue()` or `assertFalse()`, but in this situation `assertIs()` is fine.

Sorry about that, I'm honestly not quite sure what I was thinking there. I've been just getting over an ear infection and may been a bit too eager to get back into PR review. I'll take this as a sign to take a few more days off before doing anything that I don't need to address immediately. ;-)

tests/bot/cogs/test_token_remover.py
aeros: That's fine. When I mentioned realistic inputs, I was mainly referring to the valid ones. For the invalid inputs, they should just mainly have some variety.

bot/cogs/token_remover.py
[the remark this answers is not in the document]
MarkKoz: Sorry, you're right. I think I'll leave it alone though. I don't see a clean way to have separate messages and it doesn't feel to be worth the trouble.

tests/bot/cogs/test_token_remover.py
MarkKoz: No worries. I hope you don't get discouraged; your reviews would be valuable here.

bot/cogs/token_remover.py
```diff
@@ -65,94 +74,105 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
...
             channel_id=Channels.mod_alerts,
         )
 
         self.bot.stats.incr("tokens.removed_tokens")
 
+    @staticmethod
+    async def delete_message(msg: Message) -> None:
+        """Remove a `msg` containing a token and send an explanatory message in the same channel."""
+        await msg.delete()
+        await msg.channel.send(DELETION_MESSAGE_TEMPLATE.format(mention=msg.author.mention))
```
Numerlor, line 103: I think this could just be used directly where its single use is with the two straight forward expressions; in my opinion it abstracts a bit too much unnecessarily.
As for the issue with this if it stays like it is, `take_action`'s docstring nor the name of this function suggest that a message is sent so it may be confusing if the docstring or the code here is not read.

bot/cogs/token_remover.py
```diff
@@ -65,94 +74,105 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
...
-            snowflake = struct.unpack('i', content)[0]
-        except (binascii.Error, struct.error):
+            decoded_bytes = base64.urlsafe_b64decode(b64_content)
+            timestamp = int.from_bytes(decoded_bytes, byteorder="big")
+        except (binascii.Error, ValueError) as e:
+            log.debug(f"Failed to decode token timestamp '{b64_content}': {e}")
+            return False
+
+        # Seems like newer tokens don't need the epoch added, but add anyway since an upper bound
+        # is not checked.
```
Numerlor, line 170: Do you have an example of a token like that?
Unsurprisingly I had no success finding any tokens online and all tokens I have available are direct timestamps but also quite new. Still, while not a big issue, this does potentially bring a few false positives like the `DISCORD_EPOCH - TOKEN_EPOCH` edge case which is invalid in the "new" format

MarkKoz: The image explaining the token format has an old token. I couldn't find any other old tokens. I think it's better to have false positives than false negatives, and this really is an edge case. On the other hand, old format tokens are going to be infrequent so we could just stop checking for them.

bot/cogs/token_remover.py
MarkKoz: In hindsight, I agree.

MarkKoz:
> I feel like we can also add a timestamp check to is_valid_user_id to be consistent with the is_valid_user_id method. 

This is impossible/redundant, depending on how you look at it. The snowflake format treats a timestamp of 0 as the discord epoch (i.e. timestamps are relative to the epoch). Since it's being parsed as an unsigned integer, the timestamp will always be at least 0.

Numerlor:
> The snowflake format treats a timestamp of 0 as the discord epoch (i.e. timestamps are relative to the epoch).

Ah right didn't consider that we do not have anything to check against (except limiting high values)

PR #947, opened 2020-05-17
RedisCache - Data Persistence 

### LINKED ISSUE #884: Update pyflakes to version 2.2.0
The new version introduces support for assignment expressions ([changelog](https://github.com/PyCQA/pyflakes/blob/master/NEWS.rst)).
Currently when lint is ran with the expression in the code, pyflakes will crash flake8. 
Updating the version fixes this but a `noqa` ignore is still necessary because pycodestyle did not yet receive an update and does not recognize the new syntax with a colon.

The update will introduce these lint errors: https://paste.fuelrats.com/rotukasohe.txt
where `F999`s are f-strings with no expressions inside of them and the `F821`s referring to this method definition's literals in the calls inside the typehints
https://github.com/python-discord/bot/blob/b44caed7c5b26f13b4b31d237ffe07f3157aadfb/bot/cogs/moderation/management.py#L49-L56

lemonsaurus:
This pull requests adds a new feature that the bot can use to add persistent storage.

This works by utilizing a Redis server as a cache, while providing a user-friendly dict-like interface.

Let's have a look at how it works and exactly what this pull request contains.

## Creating a new RedisCache
The star of the show is the `RedisCache`. This must be created as a class attribute.
```py
class SomeClass:
    my_cache = RedisDict()
```
* When we create it as a class attribute, it will automatically create a new hashmap in Redis that uses the class name and the attribute name as its components - for example, `SomeClass.my_cache`, in this case.

## Using a RedisCache
Once you have created this object, it's a lot like an async-friendly dictionary interface. Redis only supports strings and numbers, so all keys and values must be ints, floats, or strings. If you need to store something more complex inside of it, consider writing a `to_json` and `from_json` method for your complex object, so you can easily jsonify the object going in and out of the cache.

The following features are available, and mostly behaves like their dictionary counterparts:
* `.get(key, default=None)` - Get a value. If the `key` doesn't exist, returns `default` instead.
* `.set(key, value)` - Set a value in the cache.
* `.delete(key)` - Delete a specified key value pair.
* `.clear()` - Clear the entire cache.
* `.contains(key)` - Return True if key exists in cache, otherwise False.
* `.pop(key, default=None)` - Get a value and remove it from the cache. If it doesn't exist, return `default` instead.
* `.items()` - This returns an AsyncIterator, so you need to `async for` to iterate it.
* `.update(dictionary)` - Update the cache with values in the provided dict. This will overwrite the value for any key that's already in the cache.
* `.to_dict()` - Returns a dictionary with all key value pairs in the cache.
* `.increment(key, amount)` - increments the value at `key` by `amount`
* `.decrement(key, amount)` - decrements the value at `key` by `amount`

Valid keys are integers and strings.
Valid values are integers, strings, and floats.

## What else is in this PR?
Well, there's a full suite of tests, and we've added Redis to the docker-compose file so it will be possible to use this when running the bot locally.

We also add a few new dependencies, all Redis related.

## How do I test this?
Well, you'll need to add a few new variables to your `config.yml`. Unless you plan to set up Redis (there's a docker-compose service if you want to), you can use this config:
```yml
bot: 
    redis:
        host:  "redis"
        port:  6379
        password: ""
        use_fakeredis: true
```

This will allow the bot to run with fakeredis instead of trying to connect to a Redis server, so you will not need to set up any additional dependencies.

Closes #884


bot/utils/redis_dict.py
```diff
@@ -0,0 +1,136 @@
...
+    def __iter__(self):
+        """Iterate all the items in the Redis cache."""
+        keys = self._redis.hkeys(self._namespace)
+        return iter([key.decode('utf-8') for key in keys])
+
+    def __len__(self):
+        """Return the number of items in the Redis cache."""
+        return self._redis.hlen(self._namespace)
+
+    def copy(self) -> Dict:
```
kwzrd, line 104: I probably would have named this `to_dict`. It feels like the name, `copy`, and the dosctring, `Convert to dict and return`, tell a different story.

bot/utils/redis_dict.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, namespace: Optional[str] = None) -> None:
+        """Initialize the RedisDict with the right namespace."""
+        super().__init__()
+        self._has_custom_namespace = namespace is not None
+
+        if self._has_custom_namespace:
+            self._set_namespace(namespace)
+        else:
+            self.namespace = "global"
```
kwzrd, line 42: Is this supposed to set `self._namespace`?

bot/utils/redis_dict.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def _set_namespace(self, namespace: str) -> None:
+        """Try to set the namespace, but do not permit collisions."""
+        while namespace in self._namespaces:
+            namespace = namespace + "_"
+
+        self._namespaces.append(namespace)
+        self._namespace = namespace
```
kwzrd, line 50: I think I'd prefer if there was a `clean_namespace` function or similar, that would return the namespace, and have the attr always set in the `__init__`. This way, the attr is being set in two methods (actually isn't but I think that's a bug, i.e. `namespace` vs `_namespace`), and isn't annotated anywhere.

In case of collision, the determined namespace depends on order, right? The order in which the `RedisDict` are created. Could this lead to potential problems, if a cog is renamed or replaced? I suppose the worst that can happen is the persisted data is lost. I find it a little difficult to reason about and probably just need to spend more time thinking about what we're doing here.

bot/utils/redis_dict.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def pop(self, key: ValidRedisKey, default: Optional[JSONSerializableType] = None) -> JSONSerializableType:
+        """Get the item, remove it from the cache, and provide a default if not found."""
+        value = self.get(key, default)
+        del self[key]
+        return value
+
+    def popitem(self) -> JSONSerializableType:
+        """Get the last item added to the cache."""
+        key = list(self.keys())[-1]
```
kwzrd, line 127: Does `self.keys()` delegate to `__iter__`? We do not define that method - I'm not sure where it's coming from.

SebastiaanZ: Yes, it's a mixin method provided by the base class.

bot/utils/redis_dict.py
```diff
@@ -0,0 +1,136 @@
...
+    Objects created by this class should mostly behave like a normal dictionary,
+    but will store all the data in our Redis database for persistence between restarts.
+
+    Redis is limited to simple types, so to allow you to store collections like lists
+    and dictionaries, we JSON deserialize every value. That means that it will not be possible
+    to store complex objects, only stuff like strings, numbers, and collections of strings and numbers.
+    """
+
+    _namespaces = []
+    _redis = redis_py.Redis(
```
jos-b, line 29: This won't work in production without a redis password. I have put the actual variable name in the org issue iirc but from memory it is REDIS_PASSWORD.

bot/utils/redis_dict.py
```diff
@@ -0,0 +1,136 @@
...
+        """Return a beautiful representation of this object instance."""
+        return f"RedisDict(namespace={self._namespace!r})"
+
+    def __eq__(self, other: RedisDict) -> bool:
+        """Check equality between two RedisDicts."""
+        return self.items() == other.items() and self._namespace == other._namespace
+
+    def __ne__(self, other: RedisDict) -> bool:
+        """Check inequality between two RedisDicts."""
+        return self.items() != other.items() or self._namespace != other._namespace
```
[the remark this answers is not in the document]
lemonsaurus: Yes. A normal `dict` has both a `__ne__` and an `__eq__`, and so does the `MutableMapping`. Sadly, these don't seem to work with our object, so we must overload them or they'll cause trouble for us.

bot/utils/redis_dict.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __setitem__(self, key: ValidRedisKey, value: JSONSerializableType):
```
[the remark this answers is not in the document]
lemonsaurus: Good catch. I don't see any particular reason why we shouldn't. I think initially I didn't do it because I hadn't come up with the idea of making a JSONSerializableType annotation yet, but now that we've got that, there are obvious candidates for all of these to have return types. I will fix it.

bot/utils/redis_dict.py
lemonsaurus: right, maybe, but this is just an overloaded standard dict method. we're replicating the functionality from `dict.copy()`, which would return a shallow copy of the dict object. We cannot really return a "copy" of the `RedisDict`, so we do the next best thing and just return a `dict`.

`to_dict` is probably a better name, but there's no `dict.to_dict()` - I mean, that would be silly.

bot/utils/redis_dict.py
lemonsaurus: oh. yeah it is. good eye! will fix.

bot/utils/redis_dict.py
lemonsaurus: I think this is a fair point, I don't mind making this change.

> Could this lead to potential problems, if a cog is renamed or moved?

It's very unlikely, but this is why _we should never use Redis to store anything important_, only for caching data we can afford to lose. With that principle firmly in place, I don't really see the need to dive too deep into those kinds of edge cases.

bot/utils/redis_dict.py
lemonsaurus: Fixed in ecf7f24

bot/utils/redis_dict.py
kwzrd: > this is just an overloaded standard dict method

Right, I didn't realize that. That does justify it a decent amount.

bot/utils/redis_dict.py
```diff
@@ -0,0 +1,137 @@
...
+    def popitem(self) -> JSONSerializableType:
+        """Get the last item added to the cache."""
+        key = list(self.keys())[-1]
+        return self.pop(key)
+
+    def setdefault(self, key: ValidRedisKey, default: Optional[JSONSerializableType] = None) -> JSONSerializableType:
+        """Try to get the item. If the item does not exist, set it to `default` and return that."""
+        value = self.get(key)
+
+        if value is None:
```
ks129, line 135: Why `value` variable is required? Isn't better
```suggestion
        if self.get(key) is None:
```

SebastiaanZ:
Hello, this PR is stalled until further notice. It's going to get a proper rewrite.

bot/utils/redis_dict.py
lemonsaurus: With the rewrite, this feels less relevant. `_namespace` is now really only ever set in this method, and is only initialized as None in `__init__`. This is done because `__init__` happens too early in the lifecycle for us to be able to set the attribute yet. We need to wait until `__set_name__`.

bot/utils/redis_dict.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if self._has_custom_namespace:
+            self._set_namespace(namespace)
+        else:
+            self.namespace = "global"
+
+    def _set_namespace(self, namespace: str) -> None:
+        """Try to set the namespace, but do not permit collisions."""
+        while namespace in self._namespaces:
+            namespace = namespace + "_"
```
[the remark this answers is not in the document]
lemonsaurus: Done.

bot/utils/redis_dict.py
```diff
@@ -0,0 +1,137 @@
...
+        else:
+            return default
+
+    def pop(self, key: ValidRedisKey, default: Optional[JSONSerializableType] = None) -> JSONSerializableType:
+        """Get the item, remove it from the cache, and provide a default if not found."""
+        value = self.get(key, default)
+        del self[key]
+        return value
+
+    def popitem(self) -> JSONSerializableType:
```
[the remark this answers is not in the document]
lemonsaurus: This method has been removed.

bot/utils/redis_dict.py
lemonsaurus: No longer relevant.

lemonsaurus:
This PR is now once again ready for review and for testing.

I've written a bunch of tests for it which all pass, but I haven't given it a proper test in a test bot yet, and this should absolutely be done before we merge it. It'll be a nice learning case for any reviewer to get to implement this into a testcog anyway, so I think I'll let you guys handle the bot testing.

bot/bot.py
```diff
@@ -45,8 +45,7 @@ def __init__(self, *args, **kwargs):
             # will effectively disable stats.
             statsd_url = "127.0.0.1"
 
-        asyncio.create_task(self._create_redis_session())
-
+        self.loop.create_task(self._create_redis_session())
```
SebastiaanZ, line 48: Other "lazy" session creators hook into the `_recreate` method below to do their work. What's the reason for doing it here instead of there? I don't think I mind either way, this will ensure that it will be scheduled as soon as the loop starts.

The only thing I'm slightly worried about is that I don't know when Discord calls `Bot.close()`: If that's done before it calls `Bot.clear()`, your redis session will be closed but not recreated.

Edit: A quick glance at `discord.py` makes me think it won't do that latter thing. I don't think we need to recreate our Redis session often as well. Hopefully never.

tests/bot/cogs/test_snekbox.py
```diff
@@ -21,7 +21,10 @@ async def test_post_eval(self):
         """Post the eval code to the URLs.snekbox_eval_api endpoint."""
         resp = MagicMock()
         resp.json = AsyncMock(return_value="return")
-        self.bot.http_session.post().__aenter__.return_value = resp
```
SebastiaanZ, line 24: The reason that this worked before is that when you "call" a Mock, it returns the same child mock each time. That means the lack of an argument to `post()` does not matter here: All you care about is getting the child mock that you want to manipulate the `__aenter__` return value of.

`self.bot.http_session.post()` will return the mock that will be returned later when the actual code we want to test is run and the return value of that mocks `__aenter__` mock is configured.

tests/helpers.py
```diff
@@ -264,10 +265,16 @@ class MockAPIClient(CustomMockMixin, unittest.mock.MagicMock):
     spec_set = APIClient
 
 
-# Create a Bot instance to get a realistic MagicMock of `discord.ext.commands.Bot`
-bot_instance = Bot(command_prefix=unittest.mock.MagicMock())
-bot_instance.http_session = None
-bot_instance.api_client = None
+def _get_mock_loop() -> unittest.mock.Mock:
```
SebastiaanZ, line 282: Nice method to solve this issue.

bot/utils/redis_cache.py
```diff
@@ -41,7 +39,39 @@ def _set_namespace(self, namespace: str) -> None:
...
+        elif isinstance(value, int):
+            return f"i|{value}"
+        elif isinstance(value, str):
+            return f"s|{value}"
+
+    @staticmethod
+    def _from_typestring(value: str) -> ValidRedisType:
+        """Turn a valid Redis type into a typestring."""
+        if value.startswith("f|"):
+            return float(value[2:])
```
SebastiaanZ, line 88: I'm assuming that empty type strings of float/int type are impossible here? Otherwise this may raise a `ValueError` for trying to `float("")` or `int("")`. I don't know enough about redis to be sure of that.

bot/utils/redis_cache.py
```diff
@@ -54,8 +84,11 @@ def __get__(self, instance: RedisCache, owner: Any) -> RedisCache:
         if self.bot:
             return self
 
+        if self._namespace is None:
+            raise RuntimeError("RedisCache must be a class attribute.")
+
         if instance is None:
-            raise NotImplementedError("You must create an instance of RedisCache to use it.")
+            raise RuntimeError("You must create an instance of RedisCache to use it.")
```
SebastiaanZ, line 131: This exception message is incorrect, as `instance` does not mean the instance of `RedisCache` (that would be `self`).

What happens is that we need the cache to be accessed at least once through an instance of the **cog** because we grab the `Bot` instance from the cog. So, say that we have this cog:

```py
class MyCog:
    cache = RedisCache()
```

We need the `RedisCache` to be accessed as an attribute of a `MyCog` instance, like `self.cache` within one of the `MyCog` methods. That's the step that ensures that we can fetch the `bot` attribute (or whatever it is called) from the `MyCog` instance.

This `RuntimeError` only happens when something that will be incredibly rare happens: Someone tries to access the `cache` attribute on the *class* object directly (either `MyCog.cache` or the even more convoluted `self.__class_.cache` from within a cog method) **before** the first access through `self.cache` directly.

Long story, short: It's not about creating a `RedisCache` instance, but rather that the `RedisCache` instance **has to be** accessed through an attribute of the *cog instance* (`self.cache`) the *very first time* it's accessed so we can grab the `Bot` reference.

```suggestion
            raise RuntimeError(
                "You must access the RedisCache instance through the cog instance "
                "before accessing it using the cog's class object."
            )
```

bot/utils/redis_cache.py
```diff
@@ -120,41 +130,49 @@ async def get(self, key: ValidRedisType, default: Optional[ValidRedisType] = Non
...
+        await self._validate_cache()
+        await self._redis.delete(self._namespace)
 
     async def pop(self, key: ValidRedisType, default: Optional[ValidRedisType] = None) -> ValidRedisType:
         """Get the item, remove it from the cache, and provide a default if not found."""
-        # value = await self.get(key, default)
-        # await self.delete(key)
-        # return value
+        value = await self.get(key, default)
+        await self.delete(key)
```
SebastiaanZ, line 206: Is deleting a key that does not exist "safe"? The `default` implies that trying to pop an non-existing key is valid.

Seems like it's supported, from the [HDEL docs](https://redis.io/commands/hdel):

> Removes the specified fields from the hash stored at key. Specified fields that do not exist within this hash are ignored. If key does not exist, it is treated as an empty hash and this command returns 0.

bot/utils/redis_cache.py
```diff
@@ -120,41 +130,49 @@ async def get(self, key: ValidRedisType, default: Optional[ValidRedisType] = Non
...
 
     async def items(self) -> AsyncIterator:
         """Iterate all the items in the Redis cache."""
-        # data = await redis.hgetall(self.get_with_namespace(key))
-        # for item in data:
-        #     yield item
+        await self._validate_cache()
+        data = await self._redis.hgetall(self._namespace)  # Get all the keys
+        for key, value in self._dict_from_typestring(data).items():
+            yield key, value
```
SebastiaanZ, line 187: Neat. Looks beautiful.

tests/bot/utils/test_redis_cache.py
```diff
@@ -42,95 +50,105 @@ async def test_set_get_item(self):
...
+        await self.redis.set('one', 1)
+        await self.redis.set('two', 2)
+        await self.redis.set('three', 3)
+        self.assertEqual(await self.redis.length(), 3)
+
+        await self.redis.set('four', 4)
+        self.assertEqual(await self.redis.length(), 4)
+
+    async def test_to_dict(self):
+        """Test that the .copy method returns a workable dictionary copy."""
```
SebastiaanZ, line 114: This docstring mentions the `.copy` method, but it's testing the `.to_dict` method.

tests/bot/utils/test_redis_cache.py
```diff
@@ -42,95 +50,105 @@ async def test_set_get_item(self):
...
+        await self.redis.clear()
+        self.assertEqual(await self.redis.length(), 0)
+
+    async def test_pop(self):
+        """Test that we can .pop an item from the RedisDict."""
+        await self.redis.clear()
+        await self.redis.set('john', 'was afraid')
+
+        self.assertEqual(await self.redis.pop('john'), 'was afraid')
+        self.assertEqual(await self.redis.pop('pete', 'breakneck'), 'breakneck')
```
SebastiaanZ, line 136: Important test; it indeed makes sure we don't fail on a non-existing key (see the comment I made on the actual method)

tests/bot/utils/test_redis_cache.py
```diff
@@ -42,95 +50,105 @@ async def test_set_get_item(self):
...
+
+    async def test_delete_item(self):
+        """Test that .delete allows us to delete stuff from the RedisCache."""
+        # Add an item and verify that it gets added
+        await self.redis.set("internet", "firetruck")
+        self.assertEqual(await self.redis.get("internet"), "firetruck")
+
+        # Delete that item and verify that it gets deleted
+        await self.redis.delete("internet")
+        self.assertIs(await self.redis.get("internet"), None)
```
SebastiaanZ, line 68: We rely on `.delete` ignoring/working with non-existing keys elsewhere (in `.pop`). I think it's good to test here that deleting a non-existing key is indeed fine. What do you think?

bot/utils/redis_cache.py
```diff
@@ -21,7 +21,37 @@ class communicates with the Redis server.
...
+    class SomeCog(Cog):
+        # To initialize a valid RedisCache, just add it as a class attribute here.
+        # Do not add it to the __init__ method or anywhere else, it MUST be a class
+        # attribute. Do not pass any parameters.
+        cache = RedisCache()
+
+        async def my_method(self):
+            # Now we can store some stuff in the cache just by doing this.
+            # This data will persist through restarts!
+            await self.cache.set("key", "value")
```
SebastiaanZ, line 40: Maybe we could add that it's important to access it like this, on the instance of the cog, with `self.cache`. That's vital the very first time, as we need to grab the `Bot` reference of `self` (the cog instance).

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,212 @@
...
+            return default
+        else:
+            value = self._from_typestring(value)
+            return value
+
+    async def delete(self, key: ValidRedisType) -> None:
+        """Delete an item from the Redis cache."""
+        await self._validate_cache()
+        key = self._to_typestring(key)
+        return await self._redis.hdel(self._namespace, key)
```
SebastiaanZ, line 170: We rely on the "ignore missing keys"-behavior HDEL has elsewhere (in `.pop`). I think we should document the behavior of this method when confronted with missing keys here.

bot/bot.py
lemonsaurus: I've moved it over to `_recreate` anyway. I didn't do it originally because I just wasn't aware of it. Seems like a better fit for that method, as long as it works.

tests/bot/cogs/test_snekbox.py
lemonsaurus: I think this was written by @MarkKoz, I don't recognize it.

tests/helpers.py
lemonsaurus: This should also be @MarkKoz's work. Nice work, Mark.

bot/utils/redis_cache.py
lemonsaurus: hmm.. I can't imagine what an empty float or int would look like. I don't think that's possible to create, no.

bot/utils/redis_cache.py
lemonsaurus: Yeah, I saw the same docs, so I figure it's fine.

bot/utils/redis_cache.py
lemonsaurus: Yeah I really love this syntax for asynciterators that @jos-b found, it's very elegant.

tests/bot/utils/test_redis_cache.py
lemonsaurus: Oops. I'll fix that. :eye:

tests/bot/utils/test_redis_cache.py
lemonsaurus: Yep, I guess we're testing that already.

bot/utils/redis_cache.py
SebastiaanZ: I don't know either and I can't think of a situation where we would set such an empty `int`/`float` string with our serialization method. I came across this before I read the `_to_typestring`.

bot/utils/redis_cache.py
lemonsaurus: I've added some documentation for this, but I also decided to not rely on it for `.pop`, since it's just an additional superfluous API call for no real benefit.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,215 @@
...
+    def _from_typestring(value: Union[bytes, str]) -> ValidRedisType:
+        """Turn a typestring into a valid Redis type."""
+        if isinstance(value, bytes):
+            value = value.decode('utf-8')
+        if value.startswith("f|"):
+            return float(value[2:])
+        if value.startswith("i|"):
+            return int(value[2:])
+        if value.startswith("s|"):
+            return value[2:]
```
kwzrd, line 92: I wonder if these could be defined in one place?

A very simple example:
```python
TEMPLATES = (
    (float, "f|{value}"),
    (int,   "i|{value}"),
    (str,   "s|{value}"),
)
```
Both functions then use this definition, e.g. by lopping through the pairs until it finds a match. I don't love that this is essentially a bi-directional map with each direction defined elsewhere, seems fragile. Both functions also return `None` should no match be found, which is not annotated nor documented. I looked through the entire cog, so maybe that cannot happen - these are just thoughts that I'm having now.

bot/utils/redis_cache.py
lemonsaurus: Done!

bot/utils/redis_cache.py
lemonsaurus: An interesting thought. I'll make an attempt at this.

kwzrd: Perhaps better to define the prefix, not the template, e.g.:
```python
(float, "f|")
```
Makes it easier for the `startswith` check.

lemonsaurus: Yep, that's what I went for. It's addressed in 5120717

bot/utils/redis_cache.py
lemonsaurus: I'll resolve this then.

bot/utils/redis_cache.py
kwzrd: Solid work - thanks for accepting the suggestion. I think this ends up being a lot safer and convenient, as the `TYPESTRING_PREFIXES` definition now also serves as documentation.

tests/bot/cogs/test_snekbox.py
MarkKoz: Thank you for explaining. I forgot about that as it's a bit counter-intuitive to return the same thing (it is convenient for testing of course). I suppose `post.return_value.__aenter__.return_value = ...` would have worked too then, right?

tests/helpers.py
MarkKoz: Thanks 😊

tests/bot/cogs/test_snekbox.py
SebastiaanZ: No, because the `.post` attribute is another mock than the mock it returns when you call it.

MarkKoz: So `post.return_value != post()`?

SebastiaanZ: I misread, you're right.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,258 @@
...
+    """
+    A simplified interface for a Redis connection.
+
+    This class must be created as a class attribute in a class. This is because it
+    uses __set_name__ to create a namespace like MyCog.my_class_attribute which is
+    used as a hash name when we store stuff in Redis, to prevent collisions.
+
+    The class this object is instantiated in must also contains an attribute with an
+    instance of Bot. This is because Bot contains our redis_pool, which is how this
+    class communicates with the Redis server.
```
MarkKoz, line 25: On the fence here, but these paragraphs should be less specific. I don't think they need to mention implementation details like the fact that it uses `__set_name__` or that it relies on the Bot specifically for the redis pool. If you still want this information somewhere, then it's better to put it in some internal docs rather than the class's docstring.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,258 @@
...
+        return {self._from_typestring(key): self._from_typestring(value) for key, value in dictionary.items()}
+
+    def _dict_to_typestring(self, dictionary: Dict) -> Dict:
+        """Turns all contents of a dict into typestrings."""
+        return {self._to_typestring(key): self._to_typestring(value) for key, value in dictionary.items()}
+
+    async def _validate_cache(self) -> None:
+        """Validate that the RedisCache is ready to be used."""
+        if self.bot is None:
+            raise RuntimeError("Critical error: RedisCache has no `Bot` instance.")
```
MarkKoz, line 137: I liked what you did for the namespace error message. Please make this error more specific too. Unless someone reads the class docstring (which we hope they do, but no guarantees), they won't know what this means/how to fix it.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,258 @@
...
+                "before accessing it using the cog's class object."
+            )
+
+        for attribute in vars(instance).values():
+            if isinstance(attribute, Bot):
+                self.bot = attribute
+                self._redis = self.bot.redis_session
+                return self
+        else:
+            raise RuntimeError("Cannot initialize a RedisCache without a `Bot` instance.")
```
MarkKoz, line 174: This message could also be clarified.

bot/utils/redis_cache.py
MarkKoz: The only reason I see why an async iterator is used is so we can do `async for x in y()` rather than `for x in await y()`, right? It feels misleading because, to me, an async iterator implies it's lazy, which of course `hgetall` is not.

tests/bot/utils/test_redis_cache.py
```diff
@@ -0,0 +1,175 @@
...
+        self.assertEqual(await self.redis.pop('john'), 'was afraid')
+        self.assertEqual(await self.redis.pop('pete', 'breakneck'), 'breakneck')
+        self.assertEqual(await self.redis.length(), 0)
+
+    async def test_update(self):
+        """Test that we can .update the RedisDict with multiple items."""
+        await self.redis.clear()
+        await self.redis.set("reckfried", "lona")
+        await self.redis.set("bel air", "prince")
+        await self.redis.update({
```
MarkKoz, line 144: I don't know if this is a "wrong" way to write unit tests, but it's different from my approach. My problem with doing something like this is that your test not only relies on `update` working, but also `clear` and `set`. If `set` breaks, this test will fail even though it's only supposed to fail if `update` is broken (because it's meant to test `update`). It would be misleading to see this test fail for such reason because my first thought would be something is wrong with `update`, but upon closer inspection, I'd discover it's fine and something else is wrong. I don't know if that's such a bad thing but it's just my take on it.

bot/bot.py
```diff
@@ -96,6 +112,14 @@ def _recreate(self) -> None:
                 "The previous connector was not closed; it will remain open and be overwritten"
             )
 
+        if self.redis_session and not self.redis_session.closed:
+            log.warning(
+                "The previous redis pool was not closed; it will remain open and be overwritten"
+            )
+
+        # Create the redis session
+        self.loop.create_task(self._create_redis_session())
```
aeros, line 121: I'm not 100% certain about the usage of a bare `create_task()` here for creating the redis session. To me, it seems like you would want to guarantee that the session is created before proceeding rather than scheduling the initialization to be ran in the background whenever it has the chance to (which is usually while other coroutines are being awaited, but there's no guarantee that it will have time to complete).

In order to guarantee its completion, you could do one of the following, depending on whether or not blocking the event loop is a concern:

If blocking the event loop isn't a concern here, you could use `self.loop.run_until_complete(self._create_redis_session())`. If blocking the event loop is a concern, this function could be converted into a coroutine and then you could use `await self._create_redis_session()`. At a glance, it seems like blocking the event loop wouldn't be a significant issue here since it presumably would only be ran once on restart/setup.

MarkKoz: We've been discussing on Discord about reworking the creation of all the objects here. Creating events for each thing is just too cumbersome. Your proposal to block the loop is quite a novel approach to me. Somehow, it never crossed my mind to do it like this.

I'm inclined to leave this as is and rework  it along with creation of the other objects in a separate PR. Thoughts on this course of action?

aeros: > Your proposal to block the loop is quite a novel approach to me. Somehow, it never crossed my mind to do it like this.

Yeah, I personally think that for initialization purposes when startup/restart is highly infrequent and doesn't take especially long, it's perfectly okay for some calls to block the event loop. This can help simplify the startup process significantly, and the performance impact is usually negligible at best.

> I'm inclined to leave this as is and rework it along with creation of the other objects in a separate PR. Thoughts on this course of action?

I think that would be reasonably okay. I would keep an eye on this part specifically to see if the creation of the redis session is ever not completed when expecting it to, although it looks like this is tracked by `self._redis_ready`. 

Also, a point in favor of converting to a coroutine that I forgot to mention is that it's technically "unsafe" to create async objects outside of a coroutine, especially without passing the event loop explicitly. In practice, this is only of particular significance when using an event loop that's not in the main thread or in general when using multiple event loops, but there's a decent chance that aiohttp will deprecate creating async objects outside of a coroutine and without passing the event loop, we've been doing this recently in the latest development versions of asyncio (such as for `asyncio.Lock()`).

It's caused enough interoperability issues between async libraries that it's being addressed now, although I suspect that a full removal would be decently far out due to the volume of possible breakage.

bot/utils/redis_cache.py
lemonsaurus: I don't think I agree with that. Docstrings _are_ internal docs. I think I have the opposite opinion here: I don't particularly see the need for some guide on our website to contain these sorts of implementation details, a guide should just explain how to use it. Docstrings, however, are for people who are digging into the source, and should provide a fuller idea of how the object works and behaves.

I should fix the wrong variable name, though.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,258 @@
...
+                print(value)
+
+            # We can even iterate in a comprehension!
+            consumed = [value async for key, value in self.cache.items()]
+    """
+
+    _namespaces = []
+
+    def __init__(self) -> None:
+        """Raise a NotImplementedError if `__set_name__` hasn't been run."""
```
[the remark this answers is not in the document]
lemonsaurus: Right, this was true once upon a time but definitely is not now.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,258 @@
...
+        self._namespace = None
+        self.bot = None
+
+    def _set_namespace(self, namespace: str) -> None:
+        """Try to set the namespace, but do not permit collisions."""
+        while namespace in self._namespaces:
+            namespace += "_"
```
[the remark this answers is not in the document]
lemonsaurus: Good suggestion.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,258 @@
...
+        self._namespace = namespace
+
+    @staticmethod
+    def _valid_typestring_types() -> str:
+        """
+        Creates a nice, readable list of valid types for typestrings, useful for error messages.
+
+        This will be dynamically updated if we change the TYPESTRING_PREFIXES constant up top.
+        """
+        valid_types = ", ".join([str(_type).split("'")[1] for _, _type in TYPESTRING_PREFIXES])
```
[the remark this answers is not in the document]
lemonsaurus: using `__name__` is a big improvement, yes. thanks!

tests/bot/utils/test_redis_cache.py
lemonsaurus: I see your point, but I'm not entirely sure how to test this properly without relying on `set` and `clear` to manage the state of the cache. Do you have a suggestion?

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,258 @@
...
+from typing import Any, AsyncIterator, Dict, Optional, Union
+
+from bot.bot import Bot
+
+TYPESTRING_PREFIXES = (
+    ("f|", float),
+    ("i|", int),
+    ("s|", str),
+)
+ValidRedisType = Union[str, int, float]
```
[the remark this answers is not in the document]
lemonsaurus: I thiiink I agree with this. Seems a bit hungarian. I'll remove it.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,258 @@
...
+    def _set_namespace(self, namespace: str) -> None:
+        """Try to set the namespace, but do not permit collisions."""
+        while namespace in self._namespaces:
+            namespace += "_"
+
+        self._namespaces.append(namespace)
+        self._namespace = namespace
+
+    @staticmethod
+    def _valid_typestring_types() -> str:
```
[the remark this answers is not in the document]
lemonsaurus: I did consider that, but yeah, they're staticmethods out of concern for readability mostly. 

Still, I'd be interested to hear more opinions on this. Anyone else feel these should be module constants?

bot/bot.py
lemonsaurus: Okay. This is an interesting bit of feedback, and thanks for that. I think I agree with what Mark is suggesting - that this should be refactored along with the other similar objects in a separate PR - so I'm going to resolve this conversation for now.

bot/utils/redis_cache.py
lemonsaurus: Okay, as a compromise I've moved some of the specifics into the `__set_name__` and `__get__` docstrings, and simplified the stuff in the class docstring to this:

```
Please note that this class MUST be created as a class attribute, and that that class
must also contain an attribute with an instance of our Bot. See `__get__` and `__set_name__`
for more information about how this works.
```

ks129:
Maybe should this have 2 more functions: `incr` and `decr`. These should useful for counters, because then you don't have to get value, increase it and then set. Also, maybe add some trace/info/debug logging?

bot/utils/redis_cache.py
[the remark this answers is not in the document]
lemonsaurus: After testing it out, I think it actually looks fine as a constant. 
image

Let's go with that.

bot/utils/redis_cache.py
lemonsaurus: Yep, this has now been refactored into returning an ItemsView, @MarkKoz

lemonsaurus:
> Maybe should this have 2 more functions: `incr` and `decr`. These should useful for counters, because then you don't have to get value, increase it and then set. Also, maybe add some trace/info/debug logging?
@ks129 I think both of these suggestions are great! I'll work on that now.

lemonsaurus:
All review comments should now have been addressed.

ks129:
Doesn't Redis have in-built commands for increasing and decreasing?

SebastiaanZ:
> Doesn't Redis have in-built commands for increasing and decreasing?

It does, but it does not work well with our current type string serialization system.

lemonsaurus:
> Doesn't Redis have in-built commands for increasing and decreasing?

Yes - but we can't use these, because of how we're storing the data. Since we're converting all data into "typestrings" before it gets stored, Redis will not be able to increment numbers since they are stored as strings with a prefix.

lemonsaurus:
I've added an option to `config.yml` called `use_fakeredis` in the latest commit, see 66d273f.

This will allow you guys to do a feature test of this without having to set up Redis. Just set this option to True, and the bot will use fakeredis instead of trying to connect to a Redis server.

This will make it a lot more contributor friendly, particularly for contributors on Windows. Redis is not even officially supported in Windows at all.

bot/utils/redis_cache.py
MarkKoz: If I want to learn how to use this class, I will first read the class docstring to get an overview. My perspective is that knowing the implementation details is just noise at this point. I don't need to know that it specifically uses `__set_name__` in order to learn to effectively use this class; I just need to know that it has to be instantiated as a class attribute.

I don't know what you mean by website guide. By internal docs, I meant moving this information to one of the "private" methods in the class, like to `__set_name__` itself or `_set_namespace`.

lemonsaurus: Yeah - that's what I ended up doing, so I guess this is good then?

tests/bot/utils/test_redis_cache.py
MarkKoz: One way to look at it is that we assume our redis library is already tested and working. So we can assume that if we use `hmset_dict`, it will work properly. The library already has tests written to ensure this. Thus, all our tests need to do is assert `hmset_dict` was called with the right arguments.

If not that, another way to do it is to use the redis command directly rather than via your class since, again, we assume these to be tested by the library.

On the other hand, these tests could be seen more like integration tests and so they can be left alone? Though the way I've understood integration testing is it tests two separate modules, not just multiple functions in the same module.

I think these tests are still effective and I'm not going to ask you to rewrite them.

bot/bot.py
MarkKoz: @aeros the fact that it's not a coroutine is cause it's following discord.py's design. If a non-coroutine is called by a coroutine, then I think it's fine (which is the case when `login()` calls this). But there's also `clear()`, which isn't a coroutine. And you can see I noted in a comment that calling it outside a coroutine will raise a warning. Discord.py creates its session outside a coroutine and supports re-creating it by calling `clear()` and I'm just going along with this design though we were discussing letting it go.

bot/utils/redis_cache.py
```diff
@@ -287,3 +289,36 @@ async def update(self, items: Dict) -> None:
...
+        This also supports negative amounts, although it would provide better
+        readability to use .decrement() for that.
+        """
+        # Since this has several API calls, we need a lock to prevent race conditions
+        async with self.increment_lock:
+            value = await self.get(key)
+
+            # Can't increment a non-existing value
+            if value is None:
+                raise RuntimeError("The provided key does not exist!")
```
MarkKoz, line 344: Isn't `KeyError` more appropriate?

tests/bot/utils/test_redis_cache.py
lemonsaurus: I appreciate and understand your point. I think I will keep this in mind for future tests, but perhaps leave these to be for the time being, since it would require me to rewrite almost all the tests to achieve this sort of atomic testing at this point.

I think for our purposes, it's probably good enough for now that _some of these tests_ start failing if there's a problem with the RedisCache. But I will try to write more atomic tests in the future.

bot/utils/redis_cache.py
lemonsaurus: It absolutely is. Let's change this.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,360 @@
...
+        return {self._from_typestring(key): self._from_typestring(value) for key, value in dictionary.items()}
+
+    def _dict_to_typestring(self, dictionary: Dict) -> Dict:
+        """Turns all contents of a dict into typestrings."""
+        return {self._to_typestring(key): self._to_typestring(value) for key, value in dictionary.items()}
+
+    async def _validate_cache(self) -> None:
+        """Validate that the RedisCache is ready to be used."""
+        if self.bot is None:
+            log.exception("Attempt to use RedisCache with no `Bot` instance.")
```
kwzrd, line 133: I think the `exception` method is misused here (and throughout the cog). As per the [logging docs](https://docs.python.org/3/library/logging.html#logging.Logger.exception), it's a convenience method for automatically appending the traceback when logging exceptions. It shouldn't be used outside of an `except` block. We should log using `error` directly, as there is no traceback here.

kwzrd: What is the motivation for the log? Wouldn't it be better to have the entire message in the exception that is raised on the line below?

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,360 @@
...
+        log.trace(f"Attempting to retrieve {key}.")
+        value = await self._redis.hget(self._namespace, key)
+
+        if value is None:
+            log.trace(f"Value not found, returning default value {default}")
+            return default
+        else:
+            value = self._from_typestring(value)
+            log.trace(f"Value found, returning value {value}")
+            return value
```
kwzrd, line 236: This can give `None` so return type should be annotated as `Optional`.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,360 @@
...
+    async def pop(self, key: RedisType, default: Optional[RedisType] = None) -> RedisType:
+        """Get the item, remove it from the cache, and provide a default if not found."""
+        log.trace(f"Attempting to pop {key}.")
+        value = await self.get(key, default)
+
+        # No need to try to delete something that doesn't exist,
+        # that's just a superfluous API call.
+        if value != default:
+            log.trace(f"Key {key} exists, deleting it from the cache.")
+            await self.delete(key)
```
kwzrd, line 315: Not sure about this - what if I happen to provide the same `default` as the fetched `value`? Can't that happen?

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,360 @@
...
+
+        return value
+
+    async def update(self, items: Dict) -> None:
+        """Update the Redis cache with multiple values."""
+        await self._validate_cache()
+        log.trace(f"Updating the cache with the following items:\n{items}")
+        await self._redis.hmset_dict(self._namespace, self._dict_to_typestring(items))
```
kwzrd, line 323: If I understand this correctly, the `items` param could be annotated as `Dict[RedisType, ReditType]`.

kwzrd: I'm aware that this mimics a dict update, but maybe the docstring could be a little bit more detailed? If you think it's fine as it is, then it can stay.

I didn't instantly understand what this will do from the function signature & docstring, but I also didn't connect it with a dict at first.

kwzrd:
I played with this a little bit today and it seems to work really well - I think this PR is very close to being ready. Redis worked for me right out of the box with no extra set-up needed (using Docker).

I tested by setting up my own cog that uses two `RedisCache` instances. The namespace resolution is very cool - I was worried a little bit earlier but that was mainly due to the "3 allowed ways to instantiate" part. I don't mind awaiting the getters and setters at all. Trace-level logs help make the review a smooth process - thanks a lot for those! I've only tested with ints for now but I did try the entire provided API and everything seems to work fine. With a little bit more testing I will be happy to approve this.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,360 @@
...
+            consumed = [value async for key, value in self.cache.items()]
+    """
+
+    _namespaces = []
+
+    def __init__(self) -> None:
+        """Initialize the RedisCache."""
+        self._namespace = None
+        self.bot = None
+        self.increment_lock = asyncio.Lock()
```
MarkKoz, line 84: This should be private.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,360 @@
...
+                "the RedisCache inside a class that has a Bot instance "
+                "class attribute."
+            )
+
+        if self._namespace is None:
+            log.exception("Attempt to use RedisCache with no namespace.")
+            raise RuntimeError(
+                "Critical error: RedisCache has no namespace. "
+                "Did you initialize this object as a class attribute?"
+            )
```
MarkKoz, line 147: The order of these two checks should be swapped. If I assign the cache to an instance attribute and then try to use it, it will still complain there's no bot even if my class has one. This makes it confusing cause the real issue is that my cache isn't a class attribute.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,360 @@
...
+    async def _validate_cache(self) -> None:
+        """Validate that the RedisCache is ready to be used."""
+        if self.bot is None:
+            log.exception("Attempt to use RedisCache with no `Bot` instance.")
+            raise RuntimeError(
+                "Critical error: RedisCache has no `Bot` instance. "
+                "This happens when the class RedisCache was created in doesn't "
+                "have a Bot instance. Please make sure that you're instantiating "
+                "the RedisCache inside a class that has a Bot instance "
+                "class attribute."
```
MarkKoz, line 139: Let's drop "class" cause it may imply to some that instance attributes aren't acceptable.

```suggestion
                "attribute."
```

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,360 @@
...
+                self._redis = self.bot.redis_session
+                return self
+        else:
+            log.exception("Attempt to use RedisCache with no `Bot` instance.")
+            raise RuntimeError(
+                "Critical error: RedisCache has no `Bot` instance. "
+                "This happens when the class RedisCache was created in doesn't "
+                "have a Bot instance. Please make sure that you're instantiating "
+                "the RedisCache inside a class that has a Bot instance "
+                "class attribute."
```
MarkKoz, line 204: Let's drop "class" cause it may imply to some that instance attributes aren't acceptable.
```suggestion
                "attribute."
```

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,360 @@
...
+            )
+
+        if self._namespace is None:
+            log.exception("Attempt to use RedisCache with no namespace.")
+            raise RuntimeError(
+                "Critical error: RedisCache has no namespace. "
+                "Did you initialize this object as a class attribute?"
```
MarkKoz, line 146: Being phrased as a question, it isn't clear if a class attribute is the cause of the error or the solution i.e. "Did you do this? Well, you shouldn't have!" vs. "Did you do this? You should have.". I think a statement would be clearer e.g. " You must do x..."

bot/utils/redis_cache.py
lemonsaurus: RuntimeErrors are supposed to help you figure out what you're doing wrong when you're trying to implement the RedisCache.

These logging events are supposed to help us find and diagnose bugs via Sentry.

I would be okay with including the full message, but I guess the motivation was that RuntimeErrors are contributor-facing while the exceptions are mostly just for core devs, who wouldn't mind the brevity.

Anyway, I'll change exceptions into errors and use the same error message.

bot/utils/redis_cache.py
lemonsaurus: damn, that's true. I hadn't considered that. Nice catch!

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,373 @@
...
+
+        # Since this has several API calls, we need a lock to prevent race conditions
+        async with self._increment_lock:
+            value = await self.get(key)
+
+            # Can't increment a non-existing value
+            if value is None:
+                error_message = "The provided key does not exist!"
+                log.error(error_message)
+                raise KeyError(error_message)
```
aeros, line 356: Without knowing how exactly the Redis Cache will be used, this may be a bit hard to predict, but do you have a general estimate of how frequently the caller might attempt to increment a key that doesn't exist? If it's expected to be reasonably often, it may be worth releasing the lock immediately after the `if value is None:` check.

However, if this is more for API safety and it's expected that the caller will be using a key that definitely exists 99% of the time, it's probably fine as it is.

tests/bot/utils/test_redis_cache.py
```diff
@@ -0,0 +1,239 @@
...
+        for _input, expected in conversion_tests:
+            self.assertEqual(self.redis._from_typestring(expected), _input)
+
+        # Test that exceptions are raised on invalid input
+        with self.assertRaises(TypeError):
+            self.redis._to_typestring(["internet"])
+            self.redis._from_typestring("o|firedog")
+
+    async def test_increment_decrement(self):
+        """Test .increment and .decrement methods."""
```
aeros, line 178: I would suggest also having a `test_increment_concurrent()` to ensure it's actually safe from race conditions. It *should be* okay with the current lock implementation, but to actually test for it I'd advise something like the following: 
```py
await self.redis.set("test_key", 0)
tasks = []
# Feel free to adjust the count if needed
for _ in range(100):
    task = asyncio.create_task(
        self.redis.increment("test_key", 1))
    tasks.append(task)

await asyncio.gather(*tasks)
val = await self.redis.get("test_key")
self.assertEqual(val, 100)
```

bot/utils/redis_cache.py
lemonsaurus: > do you have a general estimate of how frequently the caller might attempt to increment a key that doesn't exist?

There is no reason this should ever happen outside of the local environment. If someone wrote a feature that did this, we would probably catch it during review and prevent it from ever being merged. This feature won't be directly exposed to users (at least, I really don't think it should be), so in my opinion this should practically _never_ happen.

tests/bot/utils/test_redis_cache.py
lemonsaurus: oh yeah, I really like this. I'll set that up now. We can probably go even higher than 100, I doubt it'll add much test time if we go for 1000, and I'd feel a bit safer with that.

kwzrd:
During my testing yesterday I also found out that the test cases aren't entirely isolated in the sense that values in the cache persist between them. My approach would have been to make sure the cache is cleared before each test (probably in the `setUp`), but I don't require that this is done. I just thought I'd mention this in case it's not intentional.

tests/bot/utils/test_redis_cache.py
aeros: Awesome! Yeah, 1000 or so should more fully cover your bases; I just had no idea what the actual runtime was like so I used 100 as a more conservative estimate. I'm glad the example was helpful though. :-)

aeros:
@kwzrd 
> My approach would have been to make sure the cache is cleared before each test (probably in the setUp), but I don't require that this is done.

I agree that it would make sense to clear the cache in between tests, otherwise it ends up being somewhat subject to side effects from previous tests and can lead to inconsistent results depending on the order the tests are executed in (which is typically something to avoid).

lemonsaurus:
> My approach would have been to make sure the cache is cleared before each test (probably in the setUp)

Addressed in 35a1de3

tests/bot/utils/test_redis_cache.py
```diff
@@ -0,0 +1,272 @@
...
+        with self.assertRaises(RuntimeError):
+            await cog.cache.get("john")
+
+        # Raises "RedisCache has no namespace"
+        with self.assertRaises(RuntimeError):
+            await cog.other_cache.get("was")
+
+        # Raises "You must access the RedisCache instance through the cog instance"
+        with self.assertRaises(RuntimeError):
+            await MyCog.cache.get("afraid")
```
aeros, line 272: In the case where there are multiple of the same exception class which can be raised from the same method (such as w/ `cache.get()`), I would recommend using [`assertRaisesRegex`](https://docs.python.org/3/library/unittest.html?highlight=assertraisesregex#unittest.TestCase.assertRaisesRegex) to ensure it causes the correct exception message.

lemonsaurus: I don't know if I agree. This kind of testing is very brittle. I don't really like the idea that changing a string should fail a test, especially for a project like a community bot.

If it was absolutely crucial that we arrive at the right RuntimeError, I would, but I really don't think it is. As long as these raise errors at all, it's really all fine.

lemonsaurus: Adressed in a slightly different way in f66a635

aeros: > I don't know if I agree. This kind of testing is very brittle. I don't really like the idea that changing a string should fail a test, especially for a project like a community bot.

Fair enough. I mostly come from experience w/ the exception messages remaining static for long periods of time (if not forever), so the string for the exception message changing isn't an issue.

The custom exception classes are another good approach, and I'd say it probably makes more sense in this case since room in the namespace isn't at a high cost. So the recent commit definitely addresses this concern.

bot/utils/redis_cache.py
```diff
@@ -0,0 +1,408 @@
...
+            error_message = (
+                "Critical error: RedisCache has no `Bot` instance. "
+                "This happens when the class RedisCache was created in doesn't "
+                "have a Bot instance. Please make sure that you're instantiating "
+                "the RedisCache inside a class that has a Bot instance attribute."
+            )
+            log.error(error_message)
+            raise NoBotInstanceError(error_message)
+
+        await self.bot.redis_ready.wait()
```
aeros, line 176: I think there's a bit of an issue with relying on `await self.bot.redis_ready.wait()` after the redis session is closed in `Bot.close()`. Specifically, when the bot calls `self.redis_ready.clear()`, it sets the condition to False, and with how `Condition.wait()` works, it could end up indefinitely suspending the coroutine while waiting for the condition to be set to True. This would be an issue if `_validate_cache()` was called during or after `Bot.close()`.

I'm not overly concerned about something intentionally interacting with the redis session after it's been closed, but it is certainly possible that a redis session method that uses `_validate_cache()` is called while `Bot.close()` is in progress or just finished (specifically after `self.redis_ready.clear()`), such as when restarting the bot.

To address this, I would recommend having a flag in the bot's attributes such as `_redis_closed` that starts as False, setting it to True just before the bot calls `await self.redis_session.wait_closed()` in `Bot.close()`, and adding a check at the start of `_validate_cache()` that raises an exception when `_redis_closed` is True.

(To make it additionally safe, you could check `_redis_closed` with a lock, but I'm not certain that would be necessary.)

lemonsaurus: addressed in b7c30d4

bot/bot.py
```diff
@@ -135,6 +137,7 @@ def _recreate(self) -> None:
 
         # Create the redis session
         self.loop.create_task(self._create_redis_session())
+        self.redis_closed = False
```
SebastiaanZ, line 140: Small comment, this runs before `_create_redis_session` starts. The previous line just schedules that for execution later in the event loop.

lemonsaurus: hm, good point. could move it into the `_create_redis_session` itself

PR #949, opened 2020-05-17
Use `send_help` to ensure that our help command is correctly invoked

SebastiaanZ:
Currently, we're invoking our help command manually like this:

```py
await ctx.invoke(self.bot.get_command("help"), "bot")
```

However, unfortunately, this fails when the command callback function is decorated, leading to exceptions like:

```py
AttributeError: 'Help' object has no attribute 'channel'
  File "bot/cogs/snekbox.py", line 292, in eval_command
    await ctx.invoke(self.bot.get_command("help"), "eval")
  File "discord/ext/commands/context.py", line 132, in invoke
  File "bot/decorators.py", line 149, in inner
    if ctx.channel.id == destination_channel:

Error executing command invoked by <>: !e
```
See Sentry issue: [BOT-4T](https://sentry.io/organizations/python-discord/issues/1670955136/?referrer=github_integration)

The solution is to use the way of invoking the help command built built into `discord.py`: `ctx.send_help`, as the recent refactoring of the help command made our custom help command compatible with that `Context` method.

### DIFF
--- a/bot/cogs/bot.py
+++ b/bot/cogs/bot.py
@@ -41,7 +41,7 @@ class BotCog(Cog, name="Bot"):
     @with_role(Roles.verified)
     async def botinfo_group(self, ctx: Context) -> None:
         """Bot informational commands."""
-        await ctx.invoke(self.bot.get_command("help"), "bot")
+        await ctx.send_help("bot")
 
     @botinfo_group.command(name='about', aliases=('info',), hidden=True)
     @with_role(Roles.verified)
--- a/bot/cogs/bot.py
+++ b/bot/cogs/bot.py
@@ -41,7 +41,7 @@ class BotCog(Cog, name="Bot"):
     @with_role(Roles.verified)
     async def botinfo_group(self, ctx: Context) -> None:
         """Bot informational commands."""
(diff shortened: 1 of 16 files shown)


mathsman5133:
A Quick comment, would it be more efficient to pass in `ctx.command` rather than the name of the command, where applicable? [Passing in a string means an extra lookup](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/context.py#L286) and if for some reason the command name were to change this way means an additional change isn't required

SebastiaanZ:
> A Quick comment, would it be more efficient to pass in `ctx.command` rather than the name of the command, where applicable? [Passing in a string means an extra lookup](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/context.py#L286) and if for some reason the command name were to change this way means an additional change isn't required

The docs don't mention support for passing in a `Context` object. It can either be a `Cog`, a `Command`-object, or a string.

mathsman5133:
Right, but can't you get a `Command` object with `ctx.command`?

SebastiaanZ:
I accidentally hit `edit` instead of `reply`.

PR #950, opened 2020-05-17
Created tests for startup logging

### LINKED ISSUE #583: Write unit tests for `bot/cogs/logging.py`
Write unit tests for [`bot/cogs/logging.py`](../blob/master/bot/cogs/logging.py).

## Implementation details

## Additional information

ks129:
The title says all. Closes #583


tests/bot/cogs/test_logging.py
```diff
@@ -0,0 +1,42 @@
...
+    async def test_debug_mode_false(self):
+        """Should send connected message to dev-log."""
+        self.bot.get_channel.return_value = self.dev_log
+
+        await self.cog.startup_greeting()
+        self.bot.wait_until_guild_available.assert_awaited_once_with()
+        self.bot.get_channel.assert_called_once_with(constants.Channels.dev_log)
+
+        embed = self.dev_log.send.call_args[1]['embed']
+        self.dev_log.send.assert_awaited_once_with(embed=embed)
```
MarkKoz, line 27: If you retrieved the embed from the call args, then you already know it's been called with that value. It doesn't make sense to use a value from its call args in such an assertion. In such case, I would advice you just use `assert_awaited_once()`.

tests/bot/cogs/test_logging.py
```diff
@@ -0,0 +1,42 @@
...
+
+        self.assertEqual(embed.description, "Connected!")
+        self.assertEqual(embed.author.name, "Python Bot")
+        self.assertEqual(embed.author.url, "https://github.com/python-discord/bot")
+        self.assertEqual(
+            embed.author.icon_url,
+            "https://raw.githubusercontent.com/python-discord/branding/master/logos/logo_circle/logo_circle_large.png"
+        )
```
MarkKoz, line 35: This is something I've brought up a few times in various PRs. Basically, I don't think we should assert exact values of strings. In most cases it doesn't really matter what the exact value is. What the test should really care about is if it uses the value we wanted to. For example, rather than testing that it shows "x" every time, we should test that if we give it "x", it will show "x", but if we gave it "y", it'd show "y". This can be achieved by defining and using constants rather than using string literals. For embeds, you also have the option of using a dict along with `Embed.from_dict`. Given the embed's simplicity, using a few constants would probably be better.

PR #951, opened 2020-05-17
Allow commands in #seasonalbot-commands

Akarys42:
While chatting in #seasonalbot-commands yesterday, I tried to set a reminder to open an issue about the bot, but this channel is marked as allowed in the bot code. This PR allows commands to be executed everywhere #bot-commands was already allowed.


SebastiaanZ:
We made a conscious decision not to allow bot-commands in seasonalbot-commands. The seasonalbot-commands channel is the only channel that showcases SeasonalBot and its features and we already have a dedicated bot-commands channel.

PR #952, opened 2020-05-19
Commands for convenient regex testing

decorator-factory:
* create RegularExpressions cog
* create `!regexp search` command


bot/cogs/regular_expressions.py
```diff
@@ -0,0 +1,87 @@
...
+
+class RegularExpressions(Cog):
+    """Commands related to regular expressions."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @group(name='regexp', aliases=('regex', 're'), invoke_without_command=True)
+    async def regexp_group(self, ctx: Context) -> None:
+        """Commands for exploring the misterious worldof regular expressions."""
```
TeamSpen210, line 75: Minor typos here.
```suggestion
        """Commands for exploring the mysterious world of regular expressions."""
```

jos-b:
I'll take a more in depth look at the code soon, but one question I have now is whether executing arbitrary RegExp code is safe?

We used to have an issue on the bot where the RegExp used for code block detection resulted in the bot locking up due something called [catastrophic backtracing](https://www.regular-expressions.info/catastrophic.html). How can we prevent unsafe regular expressions from locking up the bot?

jos-b:
Found the [conversation](https://canary.discordapp.com/channels/267624335836053506/267624335836053506/424307785413689354) for when we discovered the issue in the bot, the solution was to adjust the regular expression to something safe but this still leaves the risk of abuse if people use dangerous regular expressions.

decorator-factory:
@jos-b Why not set a small time limit?

jos-b:
Right, but then we need to isolate the process which would require migrating a regex executor to another platform (which with the new UNIX commands in snekbox and possibility for more things in future is feasible) which would make this implementation obsolete.

TeamSpen210:
Could they use `snekbox` somehow? Format the regex into a template snippet which executes it, then prints to stdout a JSON result. Then parse that in the bot, handling errors.

jos-b:
[Example of catastrophic backtracing from StackOverflow](https://stackoverflow.com/questions/8010005/python-re-infinite-execution)

jos-b:
Snekbox could definitely be used, either by adding a new executor or by basically sending all the processing code as a regular eval to snekbox.

jos-b:
However at that point I think it becomes something where we need to discuss things among the core developers before we can think about implementation detail.

decorator-factory:
The `regex` module seems to handle such regular expressions without any issues. All examples of catastrophic regexes I could find work instantly.

jos-b:
The example from the StackOverflow link I posted above results in 100% CPU when tested locally:
image

decorator-factory:
I meant the third-party [regex](https://pypi.org/project/regex) module:
image

decorator-factory:
A possible issue with `regex` is that it adds some extra regex features, so a regular expression might work with the command, but not in python with `re`, which might confuse people. A simple fix would be to parse the regex using `re` as well and see if it throws an error.

bot/cogs/regular_expressions.py
decorator-factory: Thanks, I'll fix that. I think `format_*` docstrings can be improved as well.

bot/cogs/regular_expressions.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @regexp_group.command(name='search', aliases=('find', 's', '🔍'))
+    async def match_command(self, ctx: Context, pattern: ReRegex, test: str) -> None:
+        """Look for the first match of a pattern in a string."""
+        await ctx.send(match_and_format(pattern, test))
+
+    @regexp_group.command(name='search+', aliases=('find+', 's+', '🔍+'))
+    async def match_plus_command(self, ctx: Context, pattern: RegexRegex, test: str) -> None:
+        """
+        Like `!re search`, but with an extended regex format.
```
MarkKoz, line 116: It's a bad user experience to make them have to invoke the help command again to look at the docstring for `!re search`. Just copy the docstring over to this one.

bot/cogs/regular_expressions.py
```diff
@@ -0,0 +1,126 @@
...
+
+class RegularExpressions(Cog):
+    """Commands related to regular expressions."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @group(name='regexp', aliases=('regex', 're'), invoke_without_command=True)
+    async def regexp_group(self, ctx: Context) -> None:
+        """Commands for exploring the misterious world of regular expressions."""
```
MarkKoz, line 105:
```suggestion
        """Commands for exploring the mysterious world of regular expressions."""
```

bot/cogs/regular_expressions.py
```diff
@@ -0,0 +1,126 @@
...
+        except (regex.error, re.error) as e:
+            error_lines = "\n".join(format_error(e))
+            raise BadArgument(
+                "\nSyntax error in a regular expression: "
+                f"```\n{error_lines}\n```"
+            )
+
+
+ReRegex = ConvertRegex(supports_extended_features=False)
+RegexRegex = ConvertRegex(supports_extended_features=True)
```
MarkKoz, line 94: I think better names would be `Regex` and `ExtendedRegex`.

bot/cogs/regular_expressions.py
```diff
@@ -0,0 +1,126 @@
...
+    Format a match result to display in a response.
+
+    >>> format_match(re.search("(\d\d)+", "hello123456world"))
+    ['    hello123456world', ' 0:      ^^^^^^', ' 1:          ^^']
+
+    Which will look as:
+    |    hello123456world
+    | 0:      ^^^^^^
+    | 1:          ^^
+    """
```
MarkKoz, line 40: What do you think about showing the actual text matched instead of the carets? That may end up being easier to read, especially if there are a lot of groups (eyes won't have to travel far to match the carets to the original string).

bot/cogs/regular_expressions.py
```diff
@@ -0,0 +1,126 @@
...
+        """Commands for exploring the misterious world of regular expressions."""
+        await ctx.invoke(self.bot.get_command("help"), "regexp")
+
+    @regexp_group.command(name='search', aliases=('find', 's', '🔍'))
+    async def match_command(self, ctx: Context, pattern: ReRegex, test: str) -> None:
+        """Look for the first match of a pattern in a string."""
+        await ctx.send(match_and_format(pattern, test))
+
+    @regexp_group.command(name='search+', aliases=('find+', 's+', '🔍+'))
+    async def match_plus_command(self, ctx: Context, pattern: RegexRegex, test: str) -> None:
```
MarkKoz, line 114: `test` is not a clear name for the parameter. Perhaps simply named it `string`?

MarkKoz: This parameter should also be a kw-only arg so it "consumes all" if the string has spaces. This means the pattern needs to be in quotes if it has spaces, but I think this is the best option we have.

decorator-factory: I thought about that. Is there any way to put the pattern or the string inside a code block?

MarkKoz: I am not sure. I think discord.py does not consider backticks to be the same as quotes, which means you will need to parse all arguments yourself from a single string to accomplish this. But don't take my word for it, double check this.

bot/cogs/regular_expressions.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @group(name='regexp', aliases=('regex', 're'), invoke_without_command=True)
+    async def regexp_group(self, ctx: Context) -> None:
+        """Commands for exploring the misterious world of regular expressions."""
+        await ctx.invoke(self.bot.get_command("help"), "regexp")
+
+    @regexp_group.command(name='search', aliases=('find', 's', '🔍'))
```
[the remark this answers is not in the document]
decorator-factory: > On the other hand, if `re` can also handle catastrophic expression, then the `regex` module is redundant.
`re` easily freezes the interpreter if you run the [example](https://stackoverflow.com/questions/8010005/python-re-infinite-execution) linked by @j03b 
So I'm going to use `re` just for compatibility validation.

bot/cogs/regular_expressions.py
decorator-factory: I think this would be better:
```
    |    hello123456world
    | 0:      123456
    | 1:          56
```
just showing the text might be ambiguous if a substring appears multiple times in a string.

However, we still have to figure out how to display zero-width matches :thinking:

j03b:
I've been investigating `regex` a bit more and have found it is still very much susceptible to DoS attacks and should not be used with arbitrary user input.

I'll attach an example:
```py
>>> import regex
>>> regex.match("^(([a-z])+.)+[A-Z]([a-z])+$", "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa!")
... hang ...
```

decorator-factory:
> I've been investigating `regex` a bit more and have found it is still very much susceptible to DoS attacks and should not be used with arbitrary user input.
> 
> I'll attach an example:
> 
> ```python
> >>> import regex
> >>> regex.match("^(([a-z])+.)+[A-Z]([a-z])+$", "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa!")
> ... hang ...
> ```

```py
>>> regex.match("^(([a-z])+.)+[A-Z]([a-z])+$", "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa!", timeout=0.05)
>>> TimeoutError: regex timed out
```

j03b:
Huh!

decorator-factory:
Good point, though, I'll add that as well.

neonsea:
Alternatively, we can use snekbox for execution, leaving DoS and other similar attacks out of the equation. Would require some effort to move the code there, though.

decorator-factory:
Now it looks like this:
image

bot/cogs/regular_expressions.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @group(name='regexp', aliases=('regex', 're'), invoke_without_command=True)
+    async def regexp_group(self, ctx: Context) -> None:
+        """Commands for exploring the mysterious world of regular expressions."""
+        await ctx.invoke(self.bot.get_command("help"), "regexp")
+
+    @regexp_group.command(name='search', aliases=('find', 's'))
+    async def match_command(self, ctx: Context, pattern: Regex, test: str) -> None:
```
ks129, line 118:
```suggestion
    async def match_command(self, ctx: Context, pattern: Regex, *, test: str) -> None:
```

bot/cogs/regular_expressions.py
```diff
@@ -0,0 +1,135 @@
...
+        """Look for the first match of a pattern in a string."""
+        await ctx.send(match_and_format(pattern, test))
+
+    @regexp_group.command(name='search+', aliases=('find+', 's+'))
+    async def match_plus_command(self, ctx: Context, pattern: ExtendedRegex, test: str) -> None:
```
ks129, line 123:
```suggestion
    async def match_plus_command(self, ctx: Context, pattern: ExtendedRegex, *, test: str) -> None:
```

decorator-factory:
TODO:
- [ ] somehow show zero-width groups (right now they just don't show up) 
- [ ] support named groups (right now they show up as anonymous groups)
image
- [ ] possibly move the matching/searching to snekbox

bot/cogs/regular_expressions.py
decorator-factory: Thanks, added now

PR #955, opened 2020-05-20
Source command

### LINKED ISSUE #331: Source command.

ks129:
Closes #331 

Created `!source` command, that supports 3 types of source items: Help command, command, cog. Send embed that has a link to GitHub, help text of the item and when command/cog, are you able to run it in this context.


Numerlor:
Before a proper review,
The embed when not found could use some text instead of just being empty like that.
source_embed

And what's the reason for the two identical links in the title and the footer?

ks129:
@Numerlor Is this better now? I removed the link from the title and made converter raising `BadArgument` instead just sending bot repo when invalid argument provided.

bot/cogs/source.py
```diff
@@ -0,0 +1,99 @@
+import inspect
+import os
+from typing import Union
+
+from discord import Embed
+from discord.ext.commands import BadArgument, Cog, Command, Context, Converter, HelpCommand, command
```
MarkKoz, line 6: This is getting "dangerously" close to a lot of imports, but you don't _have_ to change this. I just want to let you know for the future that we're trying to get away from such long imports, preferring to qualify the names instead. If this only had 2 or 3 things then it wouldn't really matter.

bot/cogs/source.py
```diff
@@ -0,0 +1,99 @@
...
+        return f"{URLs.github_bot_repo}/blob/master/{file_location}#L{first_line_no}-L{first_line_no+len(lines)-1}"
+
+    @staticmethod
+    async def build_embed(link: str, source_object: Union[HelpCommand, Command, Cog], ctx: Context) -> Embed:
+        """Build embed based on source object."""
+        if isinstance(source_object, HelpCommand):
+            title = "Help"
+            description = source_object.__doc__
+        elif isinstance(source_object, Command):
+            title = source_object.qualified_name
```
MarkKoz, line 82: I think it will be a bit clearer to users if the title displays the same command alias that the user used rather than displaying the command's default name.

bot/cogs/source.py
```diff
@@ -0,0 +1,99 @@
...
+            filename = inspect.getsourcefile(src)
+        elif isinstance(source_item, Command):
+            src = source_item.callback.__code__
+            filename = src.co_filename
+        else:
+            src = type(source_item)
+            filename = inspect.getsourcefile(src)
+
+        lines, first_line_no = inspect.getsourcelines(src)
+        file_location = os.path.relpath(filename)
```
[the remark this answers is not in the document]
MarkKoz: Maaaaybe the line number too, but this may end up cluttering the embed. I don't know. Feel free to try it and share some screenshots.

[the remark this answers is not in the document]
ks129: I think maybe is better to add this to footer?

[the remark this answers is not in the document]
MarkKoz: Yes, sounds like that could be better. Could be displayed in the well known format `x/y/z.py:123`, where the number after the colon is the line number.

bot/cogs/source.py
ks129: I added aliases to the title of command source embed, but I can't display name with what user invoked command. Code will be very hard to follow then.

ks129:
These review is now addressed.

bot/cogs/source.py
```diff
@@ -54,15 +54,20 @@ async def source_command(self, ctx: Context, *, source_item: SourceConverter = N
...
             src = type(source_item)
             filename = inspect.getsourcefile(src)
         elif isinstance(source_item, Command):
-            src = source_item.callback.__code__
-            filename = src.co_filename
+            if source_item.cog_name == "Alias":
+                cmd_name = source_item.callback.__name__.replace("_alias", "")
+                cmd = self.bot.get_command(cmd_name.replace("_", " "))
+                src = cmd.callback.__code__
+                filename = src.co_filename
```
MarkKoz, line 71: This is fragile but I think it's a decent solution given what you have to work with. It's more of a problem with the design of aliases than with this code. I did want to look into overhauling aliases (one idea was to use decorators on commands to create aliases), but for now, I think this code is OK.

bot/cogs/source.py
```diff
@@ -10,21 +10,22 @@
...
-        4. Otherwise raise `BadArgument` error
-        """
+    """Convert an argument into a help command, tag, command, or cog."""
+
+    async def convert(self, ctx: Context, argument: str) -> Union[HelpCommand, Command, Cog, str]:
+        """Convert argument into source object."""
         if argument.lower() == "help":
             return ctx.bot.help_command
 
+        tags_cog = ctx.bot.get_cog("Tags")
```
MarkKoz, line 22: You need to account for the cog not being loaded.

bot/cogs/source.py
```diff
@@ -10,21 +10,22 @@
...
+
+        if argument.lower() in tags_cog._cache:
+            tag = argument.lower()
+            if tags_cog._cache[tag]["restricted_to"] != "developers":
+                return f"bot/resources/tags/{tags_cog._cache[tag]['restricted_to']}/{tag}.md"
+            else:
+                return f"bot/resources/tags/{tag}.md"
```
MarkKoz: Instead of this, you could edit the tag cog so it stores the file path in its cache.

bot/cogs/source.py
```diff
@@ -91,6 +99,9 @@ async def build_embed(self, link: str, source_object: Union[HelpCommand, Command
                 description = source_object.help
 
             title = source_object.qualified_name
+        elif isinstance(source_object, str):
+            title = f"Tag: {source_object.split('/')[-1].split('.')[0]}"
```
MarkKoz, line 109: This parsing can be avoided:

1. Implement my other suggestion regarding the tag cache
2. Have the converter simply return the tag name
3. Make `get_source_link` responsible for fetching the file path for the tag
4. In here you will have the original tag name, so no parsing needed!

bot/cogs/source.py
```diff
@@ -94,22 +93,22 @@ async def build_embed(self, link: str, source_object: SourceType, loc: str, firs
...
+                description = source_object.short_doc
 
             title = source_object.qualified_name
         elif isinstance(source_object, str):
             title = f"Tag: {source_object.split('/')[-1].split('.')[0]}"
             description = ""
         else:
             title = source_object.qualified_name
-            description = source_object.description
+            description = source_object.description.splitlines()[0]
```
MarkKoz, line 113: Does this account for the different styles?

```py
"""Hello"""
```

vs.

```py
"""
Hello
"""

File: bot/cogs/source.py [hunk 9]
```diff
@@ -0,0 +1,125 @@
...
+        """Display information and a GitHub link to the source code of a command, tag, or cog."""
+        if not source_item:
+            embed = Embed(title="Bot's GitHub Repository")
+            embed.add_field(name="Repository", value=f"[Go to GitHub]({URLs.github_bot_repo})")
+            embed.set_thumbnail(url="https://avatars1.githubusercontent.com/u/9919")
+            await ctx.send(embed=embed)
+            return
+
+        url, location, first_line = self.get_source_link(source_item)
+        await ctx.send(embed=await self.build_embed(url, source_item, location, first_line))
```
MarkKoz, line 59: At this point, you're relaying a bunch of arguments and it's getting messy. It will be cleaner to have `build_embed` call `get_source_link` directly.

bot/cogs/source.py
```diff
@@ -54,10 +54,10 @@ async def source_command(self, ctx: Context, *, source_item: SourceConverter = N
...
             return
 
-        url = self.get_source_link(source_item)
-        await ctx.send(embed=await self.build_embed(url, source_item))
+        url, location, first_line = self.get_source_link(source_item)
+        await ctx.send(embed=await self.build_embed(url, source_item, location, first_line))
 
-    def get_source_link(self, source_item: SourceType) -> str:
+    def get_source_link(self, source_item: SourceType) -> Tuple[str, str, Optional[int]]:
         """Build GitHub link of source item."""
```
MarkKoz: Update the docstring to describe the tuple it returns.

bot/cogs/source.py
MarkKoz: Then let's not do that. Furthermore, I don't think showing all aliases is valuable information here. I'd rather have it be the way it was with the single command name.

bot/cogs/source.py
ks129: Yes, this account both styles.

bot/cogs/source.py
```diff
@@ -96,8 +93,15 @@ def get_source_link(self, source_item: SourceType) -> Tuple[str, str, Optional[i
...
 
-    async def build_embed(self, link: str, source_object: SourceType, loc: str, first_line: Optional[int]) -> Embed:
+    async def build_embed(self, source_object: SourceType, ctx: commands.Context) -> Optional[Embed]:
         """Build embed based on source object."""
+        url, location, first_line = self.get_source_link(source_object)
+
+        # There is no URL only when bot can't fetch Tags cog
+        if not url:
+            await ctx.send("Unable to get `Tags` cog.")
+            return
```
MarkKoz, line 103: I'm not sure on this idea, but you could let `get_source_link` raise an exception when the cog isn't loaded (though perhaps something more specific than the `AttributeError` on `None` that it would normally raise) and then catch that in `source_command`. What do you think about this? Would that be clearer or just more confusing?

ks129: I refactored this now. I found that converter already makes sure is cog there, so no check is required in function.

bot/cogs/source.py
```diff
@@ -32,6 +27,14 @@ async def convert(self, ctx: commands.Context, argument: str) -> SourceType:
         if cmd:
             return cmd
 
+        tags_cog = ctx.bot.get_cog("Tags")
+
+        if not tags_cog:
+            await ctx.send("Unable to get `Tags` cog.")
+            return commands.ExtensionNotLoaded("bot.cogs.tags")
+        elif argument.lower() in tags_cog._cache:
+            return argument.lower()
```
MarkKoz, line 36: Returning an exception rather than raising it is weird. I don't agree with this approach. I think that if the cog isn't loaded, it should not attempt to get the tag and let the `BadArgument` on like 38 be raised.

bot/cogs/source.py
```diff
@@ -0,0 +1,130 @@
...
+            if source_item.cog_name == "Alias":
+                cmd_name = source_item.callback.__name__.replace("_alias", "")
+                cmd = self.bot.get_command(cmd_name.replace("_", " "))
+                src = cmd.callback.__code__
+                filename = src.co_filename
+            else:
+                src = source_item.callback.__code__
+                filename = src.co_filename
+        elif isinstance(source_item, str):
+            tags_cog = self.bot.get_cog("Tags")
```
MarkKoz, line 79: Technically, the cog could be unloaded between when the converter is used and this part of the code is reached. But this is an edge case that I don't think we should worry about, as it would overcomplicate the code.

bot/cogs/source.py
ks129: Resolved in 51d6816.

bot/cogs/source.py
```diff
@@ -0,0 +1,123 @@
...
+            filename = inspect.getsourcefile(src)
+
+        if not isinstance(source_item, str):
+            lines, first_line_no = inspect.getsourcelines(src)
+            lines_extension = f"#L{first_line_no}-L{first_line_no+len(lines)-1}"
+        else:
+            first_line_no = None
+            lines_extension = ""
+
+        file_location = Path(filename).relative_to("/bot/")
```
MarkKoz, line 85: This only works for the tag paths. It will fail for everything else:

```
ValueError: '/home/mark/repos/python/bot-pydis/bot/cogs/source.py' does not start with '/bot'
```

For everything else, you have to use the following (note this won't work on the tag paths)

```py
Path(filename).relative_to(Path.cwd()).as_posix()
```

The `as_posix()` is added to solve a separate issue - to force forward slashes even on Windows. However, this will still fail for tags cause they are already relative paths

bot/cogs/source.py
```diff
@@ -0,0 +1,128 @@
...
+        cmd = ctx.bot.get_command(argument)
+        if cmd:
+            return cmd
+
+        tags_cog = ctx.bot.get_cog("Tags")
+
+        if tags_cog and argument.lower() in tags_cog._cache:
+            return argument.lower()
+
+        raise commands.BadArgument(f"Unable to convert `{argument}` to valid command, tag, or Cog.")
```
MarkKoz, line 35: Can you exclude "tag" from this message when the cog isn't loaded?

ks129: Added check in caa4210.

PR #956, opened 2020-05-21
Help System: Move dormant faster if help channel is empty

### LINKED ISSUE #939: Shorten time till dormant after claimer deletes their message(s)
Idea behind this issue is to lower the time until a channel becomes dormant/closed when original claimer deletes their messages if there are no replies to their question.
Currently channel will only go dormant after 30 mins of inactivity or when someone closes it manually.

There are cases in which some people will claim a channel and find a solution to their problem before anyone will even reply and due to not knowing they can close channel manually, they might delete their messages which makes channel appear as if it was still available, yet it is claimed and is in `Python Help: Occupied` category, and stays so for 30 minutes of inactivity or until staff member closes it.

Ideally, the bot could check if the messages were deleted and shorten the time till channel closes automatically. Shortening time to 5-10 minutes sounds like a good time in such case as it gives the claimer possibility to send the message again (in case they deleted it to format it better) and if that's not the case, the channel would become dormant sooner, which can be good thing to have in peak hours of help channels usage.

ks129:
Closes #939 

- Implemented `on_message_delete`, that reschedule a task, when `is_empty` return `True` and channel is inside `help in-use` category. 
- `move_idle_channel` check, is channel empty and when it is, this uses correct (shorter) `idle_seconds`.
- `is_empty` request last message and then check, is this bot available message.

### DIFF
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -468,7 +468,7 @@ class HelpChannels(Scheduler, commands.Cog):
         """
         log.trace(f"Handling in-use channel #{channel} ({channel.id}).")
 
-        if not self.is_empty(channel):
+        if not await self.is_empty(channel):
             idle_seconds = constants.HelpChannels.idle_minutes * 60
         else:
             idle_seconds = constants.HelpChannels.deleted_idle_minutes * 60
@@ -731,7 +731,10 @@ class HelpChannels(Scheduler, commands.Cog):
 
         The new time for the dormant task is configured with `HelpChannels.deleted_idle_minutes`.
         """
-        if not self.is_in_category(msg.channel, constants.Categories.help_in_use) or not self.is_empty(msg.channel):
+        if not self.is_in_category(msg.channel, constants.Categories.help_in_use):
+            return
+
+        if not await self.is_empty(msg.channel):
             return
 
         log.info(f"Claimant of #{msg.channel} ({msg.author}) deleted message, channel is empty now. Rescheduling task.")
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -461,7 +461,11 @@ class HelpChannels(Scheduler, commands.Cog):
         """
         log.trace(f"Handling in-use channel #{channel} ({channel.id}).")
 
-        idle_seconds = constants.HelpChannels.idle_minutes * 60
+        if not self.is_empty(channel):
+            idle_seconds = constants.HelpChannels.idle_minutes * 60
+        else:
+            idle_seconds = constants.HelpChannels.deleted_idle_minutes * 60
+
         time_elapsed = await self.get_idle_time(channel)
 
         if time_elapsed is None or time_elapsed >= idle_seconds:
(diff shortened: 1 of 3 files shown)


bot/cogs/help_channels.py
```diff
@@ -713,6 +717,32 @@ async def on_message(self, message: discord.Message) -> None:
         # be put in the queue.
         await self.move_to_available()
 
+    @commands.Cog.listener()
+    async def on_message_delete(self, msg: discord.Message) -> None:
+        """Reschedule dormant when help channel is empty."""
```
MarkKoz, line 722: The documentation should be clearer.
```suggestion
        """
        Reschedule an in-use channel to become dormant sooner if the channel is empty.

        The new time for the dormant task is configured with `HelpChannels.deleted_idle_minutes`. 
        """
```

bot/cogs/help_channels.py
```diff
@@ -713,6 +717,32 @@ async def on_message(self, message: discord.Message) -> None:
...
+        log.info(f"Claimant of #{msg.channel} ({msg.author}) deleted message, channel is empty now. Rescheduling task.")
+
+        # Cancel existing dormant task before scheduling new.
+        self.cancel_task(msg.channel.id)
+
+        task = TaskData(constants.HelpChannels.deleted_idle_minutes * 60, self.move_idle_channel(msg.channel))
+        self.schedule_task(msg.channel.id, task)
+
+    async def is_empty(self, channel: discord.TextChannel) -> bool:
+        """Check is last message bot sent available message."""
```
MarkKoz, line 735:
```suggestion
        """Return True if the most recent message in `channel` is the bot's `AVAILABLE_MSG`."""
```

bot/cogs/help_channels.py
```diff
@@ -713,6 +717,32 @@ async def on_message(self, message: discord.Message) -> None:
...
+        msg = await self.get_last_message(channel)
+        if not msg or not msg.author.bot or not msg.embeds:
+            return False
+
+        embed = msg.embeds[0]
+        if embed.description == AVAILABLE_MSG:
+            return True
+        else:
+            return False
```
MarkKoz, line 744:
```suggestion
        return embed.description == AVAILABLE_MSG
```

bot/cogs/help_channels.py
```diff
@@ -713,6 +717,32 @@ async def on_message(self, message: discord.Message) -> None:
...
+
+        log.info(f"Claimant of #{msg.channel} ({msg.author}) deleted message, channel is empty now. Rescheduling task.")
+
+        # Cancel existing dormant task before scheduling new.
+        self.cancel_task(msg.channel.id)
+
+        task = TaskData(constants.HelpChannels.deleted_idle_minutes * 60, self.move_idle_channel(msg.channel))
+        self.schedule_task(msg.channel.id, task)
+
+    async def is_empty(self, channel: discord.TextChannel) -> bool:
```
MarkKoz, line 734: Take a look at `is_dormant_message`. It strips the messages and also checks that the author is this bot and not some other bot, but basically does the same thing as `is_empty`. You could make `is_dormant_message` a generic function that can compare to any message and then call that in `is_empty`.

PR #961, opened 2020-05-27
Permit moderation commands in ModMail category

lemonsaurus:
This pull request allows commands like `!infr search` inside any channel in the new ModMail category.

It does this by refactoring the `in_whitelist` decorator into an `in_whitelist_check` and making the `in_whitelist` deco simply use this check. Then we also use that check in the `moderation/management.py`'s `cog_check` function, asking it to check that anyone using this cog has a moderator role, is in one of the moderation channels, or is in any channel inside the ModMail category.

There are also a few f-string specifier removals in this PR, which probably shouldn't be there. I would've had to downgrade my flake8 to get rid of them. These are changes that will be made by https://github.com/python-discord/bot/pull/947 anyway, so I guess I don't see the harm in them, but I can go through and clean it up if you guys insist.

Tested on the test server, works like a charm.

### DIFF
--- a/bot/cogs/moderation/management.py
+++ b/bot/cogs/moderation/management.py
@@ -12,7 +12,7 @@ from bot.bot import Bot
 from bot.converters import Expiry, InfractionSearchQuery, allowed_strings, proxy_user
 from bot.pagination import LinePaginator
 from bot.utils import time
-from bot.utils.checks import in_channel_check, with_role_check
+from bot.utils.checks import in_whitelist_check, with_role_check
 from . import utils
 from .infractions import Infractions
 from .modlog import ModLog
@@ -49,8 +49,8 @@ class ModManagement(commands.Cog):
     async def infraction_edit(
         self,
         ctx: Context,
-        infraction_id: t.Union[int, allowed_strings("l", "last", "recent")],
-        duration: t.Union[Expiry, allowed_strings("p", "permanent"), None],
+        infraction_id: t.Union[int, allowed_strings("l", "last", "recent")],  # noqa: F821
+        duration: t.Union[Expiry, allowed_strings("p", "permanent"), None],   # noqa: F821
         *,
         reason: str = None
     ) -> None:
@@ -83,14 +83,14 @@ class ModManagement(commands.Cog):
                 "actor__id": ctx.author.id,
                 "ordering": "-inserted_at"
             }
-            infractions = await self.bot.api_client.get(f"bot/infractions", params=params)
+            infractions = await self.bot.api_client.get("bot/infractions", params=params)
 
             if infractions:
                 old_infraction = infractions[0]
                 infraction_id = old_infraction["id"]
             else:
                 await ctx.send(
-                    f":x: Couldn't find most recent infraction; you have never given an infraction."
+                    ":x: Couldn't find most recent infraction; you have never given an infraction."
                 )
                 return
         else:
@@ -224,7 +224,7 @@ class ModManagement(commands.Cog):
     ) -> None:
         """Send a paginated embed of infractions for the specified user."""
         if not infractions:
-            await ctx.send(f":warning: No infractions could be found for that query.")
+            await ctx.send(":warning: No infractions could be found for that query.")
             return
 
         lines = tuple(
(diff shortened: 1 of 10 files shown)


bot/utils/checks.py
```diff
@@ -1,12 +1,94 @@
...
+    # an iterable to prevent breakage in DM channels (for if we ever decide to enable commands there).
+    if roles and any(r.id in roles for r in getattr(ctx.author, "roles", ())):
+        log.trace(f"{ctx.author} may use the `{ctx.command.name}` command as they have a whitelisted role.")
+        return True
+
+    log.trace(f"{ctx.author} may not use the `{ctx.command.name}` command within this context.")
+
+    # Some commands are secret, and should produce no feedback at all.
+    if not fail_silently:
+        raise InWhitelistCheckFailure(redirect)
```
eivl, line 88: Clever, but do we not log people using secret commands? maybe just to see if it's misused in some way?

lemonsaurus: Not currently. Or, well, there is a trace log a few lines above this but that won't make it to Sentry. 

I believe only actionable programming errors should be in Sentry, so I don't think I want this. it doesn't really matter if regular users try to call this.

SebastiaanZ: `InWhitelistCheckFailure` should never make it to Sentry (the error handler handles them), but they do send a message back to the context in which the command was issued. This check makes sure we don't do that for moderation commands and the like (which is also the current behavior with the `in_channel_check` we are replacing).

lemonsaurus: Yeah - without this, if someone tries to do a `!infr search` in a regular channel, it'll show this in the channel where you ran the command 
image

bot/cogs/moderation/management.py
```diff
@@ -49,8 +49,8 @@ async def infraction_group(self, ctx: Context) -> None:
     async def infraction_edit(
         self,
         ctx: Context,
-        infraction_id: t.Union[int, allowed_strings("l", "last", "recent")],
-        duration: t.Union[Expiry, allowed_strings("p", "permanent"), None],
+        infraction_id: t.Union[int, allowed_strings("l", "last", "recent")],  # noqa: F821
+        duration: t.Union[Expiry, allowed_strings("p", "permanent"), None],   # noqa: F821
```
SebastiaanZ, line 53: I'm probably missing something here, but why did you add these? I'm a bit surprised that this error code would pop-up here (unless PyFlakes does not understand `allowed_strings` with arguments) and I can't reproduce it locally with our linting set-up and these noqa's removed. 

```suggestion
        infraction_id: t.Union[int, allowed_strings("l", "last", "recent")],
        duration: t.Union[Expiry, allowed_strings("p", "permanent"), None],
```

lemonsaurus: It's a new `flake8 3.8` error, that's probably why you can't reproduce it. I agree that it's basically a bug, but it does fail the linter at 3.8.

lemonsaurus:
> Other than that, just a general observation: The "in_whitelist" check/decorator combo is slowly taking over the locations where we currently use in_channel. I know that there was some initial resistance to the idea, but I think we should consider unifying it across the code base to make things consistent.

Ah yeah, I was gonna do this. I think this PR is a good time to do it, too, so I'll commit a fix a little later today removing `in_channel` and replacing all instances of it with `in_whitelist`.

bot/cogs/moderation/management.py
lemonsaurus: Also, I know this isn't strictly relevant to this PR, it's here out of pure laziness because I didn't wanna downgrade my flake8 and pre-commit setup. but if you mind, I could.

SebastiaanZ: We're going to merge a PR that includes a relock soon-ish, right?

It would make more sense to document this in that PR so it's clear why this was changed, but it feels a bit like creating a lot of work (removing it here, adding it back in another PR).

Do you know which PR is merging that version upgrade? It'll be good to keep an eye on if it ran a full lint over the entire code or just the files changed. If the latter, it may create confusing linting errors for future contribs, as linting errors suddenly pop up in files they've edited but not in the parts they actually changed.

lemonsaurus: The full relint is handled in https://github.com/python-discord/bot/pull/947, and yes, it is a full relint.

lemonsaurus:
>  I'll commit a fix a little later today removing in_channel and replacing all instances of it with in_whitelist.

This was handled in db0a384

I also added some tests for the new `in_whitelist_check` in 876fae1.

I think this is probably ready to merge now.

SebastiaanZ:
With four approvals, this should be ready to merge.

PR #964, opened 2020-05-29
Filtering: Implement bad words detection in nicknames

### LINKED ISSUE #856: Detecting bad words in nicknames
## Description
Since the management decided that we don't care about detecting names which are hard to mention
I changed this issue to solve something different - Bad word detection in our members nicknames.

This system will contain some checks that will be triggered on the event `on_message`, the display_name of the message author will be checked, and if it contains bad words, will be reported or logged.

The checks should be fast and simple to prevent long and heavy calculations that are triggered on each message, but we can expand them to fit our needs.

### Solution
**This system should be implemented as part of the Filtering cog, especially considering the overlap between this and the regex filter (formerly word and token filter).**

We will be using REGEX (Can use the same wordlists we have in the regex filters.) to detect bad words in the members `display_name` (If they have a nice nickname and an awful username, we don't care.)

This should be detected in the event `on_message`, but we want a long cooldown so we don't get notified that often - maybe 3 days.

There is no need to ping moderators about this when this filter triggers. This will be a pingless alert.

The correct channel to deliver this to is #mod-alerts (ID: 473092532147060736)

ks129:
This uses the same RegEx filter than message bad words filtering. Every alert has 3 days cooldown, storage this to Redis. Uses `on_message` event. Closes #856


kwzrd:
@ Reviewers:

Discussion has taken place in `#dev-contrib` where we have agreed to use the `on_message` event. The reasoning is that we do not intend to moderate inactive users. Alert spam / repetition should be prevented by internal logic & caching.

This should explain the discrepancy between @ks129's latest comment on the issue & the implementation as seen in this pull request.

[Discord jumplink](https://discordapp.com/channels/267624335836053506/635950537262759947/715593793692237854).

bot/cogs/filtering.py
```diff
@@ -126,6 +132,41 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
             delta = relativedelta(after.edited_at, before.edited_at).microseconds
         await self._filter_message(after, delta)
 
+    @Cog.listener('on_message')
+    async def bad_words_in_name(self, msg: Message) -> None:
+        """Check bad words from user display name. When there is more than 3 days after last alert, send new alert."""
+        if await self.name_alerts.contains(msg.author.id):
+            last_alert = parser.isoparse(await self.name_alerts.get(msg.author.id))
```
kwzrd, line 139: You can save yourself a call to Redis if you just `get` the value and abort if it gives `None`.

Keep in mind that this event will fire for *every* message sent in the guild, so if we can avoid the superfluous Redis call, we probably should. In fact, it would likely make more sense to poll Redis if and only if a `match` is found `(L153)`.

In other words, instead of:

1. See if an alert can be sent for given user (regardless of nickname)
2. If so, analyse author name

We can do:

1. Analyse author name for issues
2. If issues exist, check if alert should be sent

I'm not sure exactly how performant the cache is and whether superfluous calls to it are any problem at all, but given considering that this event fires *very* often, my instinct is to at least consider the potential performance implications it may have.

kwzrd:
Should we consider the use of an `asyncio.Lock`? The event listen is not atomic and I think a race condition can happen where another event will fire before the first one finishes, resulting in an `@everyone` ping-spam in `#mod-alerts`.

ks129:
@kwzrd Is this better now?

bot/cogs/filtering.py
```diff
@@ -126,6 +135,43 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
             delta = relativedelta(after.edited_at, before.edited_at).microseconds
         await self._filter_message(after, delta)
 
+    async def bad_words_in_name(self, msg: Message) -> None:
+        """Check bad words from user display name. When there is more than 3 days after last alert, send new alert."""
+        if not self.name_lock:
+            self.name_lock = asyncio.Lock()
```
TeamSpen210, line 141: Isn't this itself a race condition (though not much of an impact)? :) Why aren't you creating the lock immediately, if there's a reason it should probably be documented...

ks129: I will add comment tomorrow, but lock is not thread safe.

kwzrd: The lock doesn't have to be created in an asynchronous context. It can and should be assigned in the cog's `__init__` method. Then the filter can simply rely on it already existing.

kwzrd: I had a chat with @SebastiaanZ about my comment and I should clarify:

> It can and should be assigned in the cog's `__init__` method.

The lock must bind to the correct event loop, and so *we* can create it in the cog's `__init__` because `discord.py` will initialise the event loop before the cog is instantiated. It is therefore safe (and convenient) for us to do so, but that doesn't mean that locks can always be instantiated as soon as possible.

ks129: Okay, resolved with 0e12ff1.

ks129:
@MarkKoz Addressed reviews.

bot/cogs/filtering.py
```diff
@@ -126,6 +136,53 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
...
+            if match:
+                matches.append(match)
+        return matches
+
+    async def check_send_alert(self, member: Member) -> bool:
+        """When there is less than 3 days after last alert, return `False`, otherwise `True`."""
+        last_alert = await self.name_alerts.get(member.id)
+        if last_alert:
+            last_alert = datetime.fromtimestamp(last_alert)
+            if datetime.now() - timedelta(days=DAYS_BETWEEN_ALERTS) < last_alert:
```
MarkKoz, line 154: Should use UTC for the times: `utcnow` and `utcfromtimestamp`.

bot/cogs/filtering.py
```diff
@@ -126,6 +136,53 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
...
+                return False
+
+        return True
+
+    async def check_is_bad_words_in_name(self, member: Member) -> None:
```
MarkKoz, line 159: Sorry to nitpick, but this is awkward to read since it isn't grammatically correct (should be `are` since `words` is plural). I suggest just dropping the `is` entirely: `check_bad_words_in_name`.

ks129:
@MarkKoz made fixes.

bot/cogs/filtering.py
```diff
@@ -126,6 +136,54 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
...
+                f"**Display Name:** {member.display_name}\n"
+                f"**Bad Matches:** {', '.join(match.group() for match in matches)}"
+            )
+            await self.mod_log.send_log_message(
+                icon_url=Icons.token_removed,
+                colour=Colours.soft_red,
+                title="Username filtering alert",
+                text=log_string,
+                channel_id=Channels.mod_alerts
+            )
```
j03b, line 182: Similar to our other mod alerts, it may be useful to include an image here to stay similar.

The `send_log_message` function supports this in the `thumbnail` argument:
https://github.com/python-discord/bot/blob/b4688f5d9113dafe7e307cc276b27f32f2700475/bot/cogs/moderation/modlog.py#L90

bot/cogs/filtering.py
```diff
@@ -126,6 +136,54 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
...
         await self._filter_message(after, delta)
 
+    @staticmethod
+    def get_name_matches(name: str) -> List[re.Match]:
+        """Check bad words from passed string (name). Return list of matches."""
+        matches = []
+        for pattern in WATCHLIST_PATTERNS:
+            match = pattern.search(name)
+            if match:
+                matches.append(match)
```
j03b, line 146: Seems like a good use case for a walrus (assuming the linter supports them):
```suggestion
            if match := pattern.search(name):
                matches.append(match)
```

bot/cogs/filtering.py
```diff
@@ -52,8 +57,12 @@ def expand_spoilers(text: str) -> str:
 class Filtering(Cog):
     """Filtering out invites, blacklisting domains, and warning us of certain regular expressions."""
 
+    # Redis cache for last bad words in nickname alert sent per user.
```
j03b, line 60: Comment could probably be a little more descriptive:
```suggestion
    # Redis cache mapping a user ID to the last timestamp a bad nickname alert was sent
```

bot/cogs/filtering.py
```diff
@@ -126,6 +136,54 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
...
+                log.trace(f"Last alert was too recent for {member}'s nickname.")
+                return False
+
+        return True
+
+    async def check_bad_words_in_name(self, member: Member) -> None:
+        """Send a mod alert every 3 days if a username still matches a watchlist pattern."""
+        # Use lock to avoid race conditions
+        async with self.name_lock:
+            # Check does nickname have match in filters.
```
j03b, line 164: Better worded comment
```suggestion
            # Check whether the users display name contains any words in our blacklist
```

bot/cogs/filtering.py
j03b, line 182: This could all use a little bit of spacing
```suggestion
            log.info(f"Sending bad nickname alert for '{member.display_name}' ({member.id}).")

            log_string = (
                f"**User:** {member.mention} (`{member.id}`)\n"
                f"**Display Name:** {member.display_name}\n"
                f"**Bad Matches:** {', '.join(match.group() for match in matches)}"
            )

            await self.mod_log.send_log_message(
                icon_url=Icons.token_removed,
                colour=Colours.soft_red,
                title="Username filtering alert",
                text=log_string,
                channel_id=Channels.mod_alerts
            )
```

bot/cogs/filtering.py
```diff
@@ -126,6 +136,54 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
...
+        for pattern in WATCHLIST_PATTERNS:
+            match = pattern.search(name)
+            if match:
+                matches.append(match)
+        return matches
+
+    async def check_send_alert(self, member: Member) -> bool:
+        """When there is less than 3 days after last alert, return `False`, otherwise `True`."""
+        last_alert = await self.name_alerts.get(member.id)
+        if last_alert:
```
j03b, line 152: Another walrus opportunity:
```suggestion
        if last_alert := await self.name_alerts.get(member.id):
```

ks129:
@j03b Reviews addressed

PR #970, opened 2020-05-29
Add staff counts to !server

### LINKED ISSUE #969: Staff count for the !server embed
In Counts, we should have `Staff members` with the number of `Helpers`

lemonsaurus:
A small change to add number of staff channels and number of staff members to !server. These are nice numbers to track, and may be interesting for everyone.

image

Closes #969

### DIFF
--- a/bot/cogs/information.py
+++ b/bot/cogs/information.py
@@ -1,14 +1,13 @@
 import colorsys
-import functools
 import logging
 import pprint
 import textwrap
 from collections import Counter, defaultdict
 from string import Template
-from typing import Any, List, Mapping, Optional, Union
+from typing import Any, Mapping, Optional, Union
 
-import more_itertools
-from discord import Colour, Embed, Member, Message, Role, Status, utils
+from discord import ChannelType, Colour, Embed, Guild, Member, Message, Role, Status, utils
+from discord.abc import GuildChannel
 from discord.ext.commands import BucketType, Cog, Context, Paginator, command, group
 from discord.utils import escape_markdown
 
@@ -29,36 +28,14 @@ class Information(Cog):
         self.bot = bot
 
     @staticmethod
-    def _get_channels_with_role_permission(ctx: Context, role: Role, perm: str, value: Optional[bool]) -> List[int]:
-        """Get a list of channel IDs where one of the specified roles can read."""
-        channel_ids = []
+    def role_can_read(channel: GuildChannel, role: Role) -> bool:
+        """Return True if `role` can read messages in `channel`."""
+        overwrites = channel.overwrites_for(role)
+        return overwrites.read_messages is True
 
-        for channel in ctx.guild.channels:
-            overwrites = channel.overwrites_for(role)
-            if overwrites.is_empty():
-                continue
-
-            for _perm, _value in overwrites:
-                if _perm == perm and _value is value:
-                    channel_ids.append(channel.id)
-
-        return channel_ids
-
-    _get_channels_where_role_can_read = functools.partialmethod(
-        _get_channels_with_role_permission,
-        perm='read_messages',
-        value=True
-    )
-
-    _get_channels_where_role_cannot_read = functools.partialmethod(
-        _get_channels_with_role_permission,
-        perm='read_messages',
-        value=False
-    )
-
-    def _get_staff_channel_count(self, ctx: Context) -> int:
+    def get_staff_channel_count(self, guild: Guild) -> int:
         """
-        Get number of channels that are staff-only.
+        Get the number of channels that are staff-only.
 
         We need to know two things about a channel:
         - Does the @everyone role have explicit read deny permissions?
@@ -66,22 +43,22 @@ class Information(Cog):
 
         If the answer to both of these questions is yes, it's a staff channel.
         """
-        helpers = ctx.guild.get_role(constants.Roles.helpers)
-        moderators = ctx.guild.get_role(constants.Roles.moderators)
-        admins = ctx.guild.get_role(constants.Roles.admins)
-        everyone = ctx.guild.default_role
+        channel_ids = set()
+        for channel in guild.channels:
+            if channel.type is ChannelType.category:
+                continue
 
-        # Let's build some lists of channels.
-        everyone_denied = self._get_channels_where_role_cannot_read(ctx, everyone)
-        staff_allowed = more_itertools.flatten([
-            self._get_channels_where_role_can_read(ctx, admins),       # Admins has explicit read message allow
-            self._get_channels_where_role_can_read(ctx, moderators),   # Moderators has explicit read message allow
-            self._get_channels_where_role_can_read(ctx, helpers),      # Helpers has explicit read message allow
-        ])
+            if channel in channel_ids:
+                continue  # Only one of the roles has to have read permissions, not all
 
-        # Now we need to check which channels are both denied for @everyone and permitted for staff
-        staff_channels = set(cid for cid in staff_allowed if cid in everyone_denied)
-        return len(staff_channels)
+            everyone_can_read = self.role_can_read(channel, guild.default_role)
+
+            for role in constants.STAFF_ROLES:
+                role_can_read = self.role_can_read(channel, guild.get_role(role))
+                if role_can_read and everyone_can_read is False:
+                    channel_ids.add(channel.id)
+
+        return len(channel_ids)
 
     @with_role(*constants.MODERATION_ROLES)
     @command(name="roles")
@@ -176,7 +153,7 @@ class Information(Cog):
 
         # How many staff members and staff channels do we have?
         staff_member_count = len(ctx.guild.get_role(constants.Roles.helpers).members)
-        staff_channel_count = self._get_staff_channel_count(ctx)
+        staff_channel_count = self.get_staff_channel_count(ctx.guild)
 
         # Because channel_counts lacks leading whitespace, it breaks the dedent if it's inserted directly by the
         # f-string. While this is correctly formated by Discord, it makes unit testing difficult. To keep the formatting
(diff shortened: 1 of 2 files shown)


bot/cogs/information.py
```diff
@@ -104,13 +104,35 @@ async def server_info(self, ctx: Context) -> None:
...
+        hidden_channels = 0
+
+        for channel in ctx.guild.channels:
+            overwrites = channel.overwrites_for(everyone_role)
+            if overwrites.is_empty():
+                continue
+
+            for perm, value in overwrites:
+                if perm == 'read_messages' and value is False:
+                    hidden_channels += 1
```
SebastiaanZ, line 127: This will count `#:BIG RED CIRCLE: verification :BIG RED CIRCLE:` as a public channel, which it is, I guess; just not for actual members of our guild. I doubt anyone is going to do a manual count and compare the numbers, though.

On a more serious note: One reason this may fail in the future is if we decide to cherry-pick which channels we want to  showcase in Discovery. When changing the permissions over, I thought about that, but decided to postpone such a discussion. I think just leaving everything "lurkable" is fine.

Another thing is that Code Jam team channels will be displayed as "staff channels", since they won't be publicly visible.

lemonsaurus: This is all true. But the only way to really properly address this would be to check if Helpers, Moderators or Admins had explicit read in these channels, I guess? Even that's not perfect.

MarkKoz: How about considering it a staff channel if the roles with read permissions are a subset of the roles with read permissions in the staff lounge? Maybe something like the event channel would get left out though.

bot/cogs/information.py
```diff
@@ -114,17 +144,27 @@ async def server_info(self, ctx: Context) -> None:
...
+        everyone_denied = self._get_channels_where_role_cannot_read(ctx, everyone)
+        staff_allowed = more_itertools.flatten([
+            self._get_channels_where_role_can_read(ctx, admins),       # Admins has explicit read message allow
+            self._get_channels_where_role_can_read(ctx, moderators),   # Moderators has explicit read message allow
+            self._get_channels_where_role_can_read(ctx, helpers),      # Helpers has explicit read message allow
+        ])
+
+        # Now we need to check which channels are both denied for everyone and permitted for staff
+        staff_channels = [cid for cid in staff_allowed if cid in everyone_denied]
+        staff_channel_count = len(staff_channels)
```
MarkKoz, line 167: Moving this to a separate function like `get_staff_channel_count` would improve readability of the command's function; it's getting a bit large.

bot/cogs/information.py
```diff
@@ -114,17 +144,27 @@ async def server_info(self, ctx: Context) -> None:
...
+        admins = ctx.guild.get_role(constants.Roles.admins)
+        everyone = ctx.guild.roles[0]
+
+        # Let's build some lists of channels.
+        everyone_denied = self._get_channels_where_role_cannot_read(ctx, everyone)
+        staff_allowed = more_itertools.flatten([
+            self._get_channels_where_role_can_read(ctx, admins),       # Admins has explicit read message allow
+            self._get_channels_where_role_can_read(ctx, moderators),   # Moderators has explicit read message allow
+            self._get_channels_where_role_can_read(ctx, helpers),      # Helpers has explicit read message allow
+        ])
```
MarkKoz, line 163: It is probably insignificant, but it strikes me that calling the function 4 times is inefficient. This could all be determined within one pair of iterations, though perhaps at the cost of readability or cleanliness.

MarkKoz, line 163: Isn't it a problem that this could contain duplicate channels?

lemonsaurus: right, shit, I had an "aghh why won't this work" moment where I swapped the two dicts in the list comp below. Originally I was checking for channels in the `everyone_denied` that were also in `staff_allowed`, which doesn't suffer a problem with duplicates. It should be swapped back.

lemonsaurus: It definitely is, I did it for readability reasons. There may be an approach that finds a better balance between readability and optimization

bot/cogs/information.py
lemonsaurus: Agreed.

lemonsaurus:
> I don't understand how this change makes it more precise. Under what circumstances would a channel hidden from `@everyone` not have at least one staff role with read permissions?

This was mostly in reference to what @SebastiaanZ said about us potentially cherry picking which channels are visible in lurk mode. If we did that, we would make them read deny to `@everyone` and read allow to Developers, I imagine. There'd be no reason to have a staff role with read permissions.

This also resolves the Code Jam channels @SebastiaanZ asked about, because these do not have any explicit staff read permits, only admins can see them because admins can see everything. At least, I think so. I skimmed the code in CodeJams to come to that conclusion.

SebastiaanZ:
> @MarkKoz
> 
> > I don't understand how this change makes it more precise. Under what circumstances would a channel hidden from `@everyone` not have at least one staff role with read permissions?
> 
> This was mostly in reference to what @SebastiaanZ said about us potentially cherry picking which channels are visible in lurk mode. If we did that, we would make them read deny to `@everyone` and read allow to Developers, I imagine. There'd be no reason to have a staff role with read permissions.
> 
> This also resolves the Code Jam channels @SebastiaanZ asked about, because these do not have any explicit staff read permits, only admins can see them because admins can see everything. At least, I think so. I skimmed the code in CodeJams to come to that conclusion.

I think I've triggered quite a mess of complexity with that comment. I'm not quite sure if it is really needed; the only downside of not caring about it is that our staff channel count is going to look massive, especially during code jams. Still, the thought of people thinking "What are those people doing in all those channels?" is quite funny.

For the Code Jam, the code is not entirely story there, I think. I know we had a discussion in the mods team about allowing moderators to see the code jam channels around the time the previous jam kicked off, but I don't quite recall the conclusion to that discussion. I *think* we settled on keeping it admins-only, which would make the solution fine. (The new events team will get access, but that role isn't counted as a staff role in our constants.)

I'm glad we're not talking about "send" overwrites, because that would be a mess of a totally different level.

bot/cogs/information.py
lemonsaurus: @SebastiaanZ this has now been addressed, see conversation below.

lemonsaurus:
> I know we had a discussion in the mods team about allowing moderators to see the code jam channels around the time the previous jam kicked off, but I don't quite recall the conclusion to that discussion

Pretty sure we decided against this. It's been fine to handle this in admins only. And yes, this is a silly bit of complexity for such an absolutely inessential feature, but I do think the code has been improved by this, overall.

bot/cogs/information.py
```diff
@@ -26,6 +27,51 @@ class Information(Cog):
...
+                continue
+
+            if channel in channel_ids:
+                continue  # Only one of the roles has to have read permissions, not all
+
+            everyone_can_read = self.role_can_read(channel, guild.default_role)
+
+            for role in constants.STAFF_ROLES:
+                role_can_read = self.role_can_read(channel, guild.get_role(role))
+                if role_can_read and everyone_can_read is False:
```
kwzrd, line 58: Super nit-pick, but since the variable names read so nicely we could do:
```python
if role_can_ready and not everyone_can_ready:
```
Since they are both bools we don't risk accidentally getting something falsey.

bot/cogs/information.py
```diff
@@ -26,6 +27,51 @@ class Information(Cog):
...
+        return len(channel_ids)
+
+    @staticmethod
+    def get_channel_type_counts(guild: Guild) -> str:
+        """Return the total amounts of the various types of channels in `guild`."""
+        channel_counter = Counter(c.type for c in guild.channels)
+        channel_type_list = []
+        for channel in channel_counter:
+            channel_type = str(channel).title()
+            channel_type_list.append(f"{channel_type} channels: {channel_counter[channel]}")
```
kwzrd, line 70: Could go over `channel_counter.items()` to avoid the lookup.

It could also be a list comp, but maybe it's complex enough to warrant a loop.

bot/cogs/information.py
```diff
@@ -26,6 +27,51 @@ class Information(Cog):
...
+
+        channel_type_list = sorted(channel_type_list)
+        return "\n".join(channel_type_list).strip()
```
kwzrd, line 73: Why the strip? I don't see how this could have surrounding whitespace, the strings are made on line 70.

bot/cogs/information.py
MarkKoz: You're right. I was concerned that it may be `None`, but since I explicitly use `is` in the function, it should be always be a bool. In hindsight, it's misleading to return a bool cause a value of `None` doesn't indicate one way or another if a role can read.

bot/cogs/information.py
```diff
@@ -26,6 +27,51 @@ class Information(Cog):
...
+
+        If the answer to both of these questions is yes, it's a staff channel.
+        """
+        channel_ids = set()
+        for channel in guild.channels:
+            if channel.type is ChannelType.category:
+                continue
+
+            if channel in channel_ids:
+                continue  # Only one of the roles has to have read permissions, not all
```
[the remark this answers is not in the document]
MarkKoz: You're right, it should be using the id. It's here because The role loop is nested within the channel loop. Once it finds a role with read permissions (which can be determined by checking if the channel has been added to the set), it doesn't need to check the rest of the roles for that channel.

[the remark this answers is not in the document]
MarkKoz: OK I see where I went wrong. This should be a break and be inside the role loop.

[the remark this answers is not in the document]
MarkKoz: Actually, just break right after adding to the set. Even easier.

lemonsaurus:
> If you're up for it, this could use some tests.

I hate to say this but I'm not really up for this, at least not in the very near future.

If someone else wants to commit some tests for this, you're welcome to. Otherwise, since it's not at all a critical feature and it seems to work, we could probably merge this without tests.

If you insist on tests, I can probably write some in like a week or two when I've freed up some time. Sorry!

MarkKoz:
No, I don't insist. I approved, after all.

PR #971, opened 2020-05-29
Remove the periodic ping from #verification

### LINKED ISSUE #967: Remove periodic ping in #verification
This creates "accept bombs" that may trigger our anti-raid and burst spam filters, and is not really necessary anymore. We have more than enough people joining without having to periodically nag at them.

lemonsaurus:
This creates "accept bombs" that may trigger our anti-raid and burst spam filters, and is not really necessary anymore. We have more than enough people joining without having to periodically nag at them.

Closes #967.

### DIFF
--- a/bot/cogs/verification.py
+++ b/bot/cogs/verification.py
@@ -1,9 +1,7 @@
 import logging
 from contextlib import suppress
-from datetime import datetime
 
 from discord import Colour, Forbidden, Message, NotFound, Object
-from discord.ext import tasks
 from discord.ext.commands import Cog, Context, command
 
 from bot import constants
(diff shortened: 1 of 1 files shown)


bot/cogs/verification.py
```diff
@@ -65,10 +54,7 @@ async def on_message(self, message: Message) -> None:
 
         if message.author.bot:
             # They're a bot, delete their message after the delay.
-            # But not the periodic ping; we like that one.
-            if message.content != PERIODIC_PING:
-                await message.delete(delay=BOT_MESSAGE_DELETE_DELAY)
-            return
```
kwzrd, line 71: Is the deletion of the `return` statement intentional?

lemonsaurus: No, and it's also a bad idea. Nice catch!

PR #974, opened 2020-05-30
Add multichannel !purge

### LINKED ISSUE #968: !purge should take an optional set of channels, for mass-purging multi-channel spam.
some people will spam several channels. Let's make all !purge commands default to current channel only, but optionally take a list of channels to purge.

lemonsaurus:
We can now pass in as many channel mentions as we want after any !purge command - for example `!purge all 5 #python-general #python-language`

Closes #968.

### DIFF
--- a/bot/cogs/clean.py
+++ b/bot/cogs/clean.py
@@ -46,7 +46,6 @@ class Clean(Cog):
         bots_only: bool = False,
         user: User = None,
         regex: Optional[str] = None,
-
     ) -> None:
         """A helper function that does the actual message cleaning."""
         def predicate_bots_only(message: Message) -> bool:
--- a/bot/cogs/clean.py
+++ b/bot/cogs/clean.py
@@ -166,10 +166,8 @@ class Clean(Cog):
             return
 
         # Build the embed and send it
-        if len(channels) > 1:
-            target_channels = ", ".join(channel.mention for channel in channels)
-        else:
-            target_channels = f"<#{channels[0].id}>"
+        target_channels = ", ".join(channel.mention for channel in channels)
+
         message = (
             f"**{len(message_ids)}** messages deleted in {target_channels} by **{ctx.author.name}**\n\n"
             f"A log of the deleted messages can be found [here]({log_url})."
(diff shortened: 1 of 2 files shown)


lemonsaurus:
@MarkKoz should be resolved now.

bot/cogs/clean.py
```diff
@@ -114,43 +113,37 @@ def predicate_regex(message: Message) -> bool:
         if not channels:
             channels = [ctx.channel]
 
+        # Delete the invocation first
+        with self.mod_log.ignore_all():
+            await ctx.message.delete()
```
MarkKoz, line 118: Let's still try to use `ignore()` when feasible. We don't need to search to get the ID for this message, so it's not a problem.

bot/cogs/clean.py
```diff
@@ -114,43 +113,37 @@ def predicate_regex(message: Message) -> bool:
...
+        # This is only done so we can create a log to upload.
         messages = []
         message_ids = []
         self.cleaning = True
-        invocation_deleted = False
 
-        # To account for the invocation message, we index `amount + 1` messages.
         for channel in channels:
-            async for message in channel.history(limit=amount + 1):
+            async for message in channel.history(limit=amount):
```
MarkKoz: Since you added the `ignore_all` decorator, this loop is obsolete.

bot/cogs/moderation/modlog.py
```diff
@@ -81,6 +83,15 @@ def ignore(self, event: Event, *items: int) -> None:
...
                 self._ignored[event].append(item)
 
+    @contextmanager
+    def ignore_all(self) -> None:
+        """Ignore all events while inside this context scope."""
+        self._ignore_all = True
+        try:
+            yield
+        finally:
+            self._ignore_all = False
```
MarkKoz, line 93: This won't work so well if the decorator is being used more than once simultaneously because they all share the same variable and one decorator could exit earlier than another.

bot/cogs/clean.py
lemonsaurus: Looks like channel.purge will return the messages, so you are right - we can just store the messages from the return values here and then use those for the log upload.

bot/cogs/clean.py
lemonsaurus: Alright, fair enough.

lemonsaurus:
Reverted to the old approach now, after our post-staff meeting discussion.

PR #978, opened 2020-06-01
Use Scheduler instead of asyncio.sleep on silence cog

### LINKED ISSUE #977: Use scheduler for silencing
I was looking through the code of [`silencer.py`](https://github.com/python-discord/bot/blob/1d3acb0a222a885c5ef42e91570a00f91976fb5b/bot/cogs/moderation/silence.py) and I found this:

https://github.com/python-discord/bot/blob/1d3acb0a222a885c5ef42e91570a00f91976fb5b/bot/cogs/moderation/silence.py#L93

With this approach, in case the channel was unsilenced and re-silenced back, it might get unsilenced by this timer. This is a relatively rare occurrence as this would have to happen within the 15 minutes limit, but nevertheless it seems more logical to use `Scheduler` in here, which can be canceled when the channel is manually unsilenced, to prevent this.

ItsDrike:
Closes #977 

As described in #977 using `asyncio.sleep` isn't a good way to check if the silence duration has passed, since in some cases, it can cause channels to get prematurely unsilenced.

This solves that issue by using `Scheduler` from `bot.utils.scheduling` which has the ability to cancel on-going scheduled tasks. This also means that there can be a message on `!unsilence` saying that the channel is currently not silenced, which provides much better feedback than not sending anything.

### DIFF
--- a/bot/cogs/moderation/silence.py
+++ b/bot/cogs/moderation/silence.py
@@ -5,16 +5,16 @@ from collections import namedtuple
 from contextlib import suppress
 from typing import Optional
 
+from discord import TextChannel
+from discord.ext import commands, tasks
+from discord.ext.commands import Context
+
 from bot.bot import Bot
 from bot.constants import Channels, Emojis, Guild, MODERATION_ROLES, Roles
 from bot.converters import HushDurationConverter
 from bot.utils import time
 from bot.utils.checks import with_role_check
 from bot.utils.scheduling import Scheduler
-
-from discord import TextChannel
-from discord.ext import commands, tasks
-from discord.ext.commands import Context
 
 log = logging.getLogger(__name__)
 
--- a/bot/cogs/moderation/silence.py
+++ b/bot/cogs/moderation/silence.py
@@ -15,9 +15,12 @@ from bot.utils.scheduling import Scheduler
 
 log = logging.getLogger(__name__)
 
-SilencedChannel = NamedTuple(
-    "SilencedChannel", [("ctx", Context), ("delay", int)]
-)
+
+class TaskData(NamedTuple):
+    """Data for a scheduled task."""
+
+    delay: int
+    ctx: Context
 
 
 class SilenceNotifier(tasks.Loop):
@@ -68,14 +71,14 @@ class Silence(Scheduler, commands.Cog):
         self._get_instance_vars_task = self.bot.loop.create_task(self._get_instance_vars())
         self._get_instance_vars_event = asyncio.Event()
 
-    async def _scheduled_task(self, channel: SilencedChannel) -> None:
+    async def _scheduled_task(self, task: TaskData) -> None:
         """Calls `self.unsilence` on expired silenced channel to unsilence it."""
-        await asyncio.sleep(channel.delay)
+        await asyncio.sleep(task.delay)
         log.info("Unsilencing channel after set delay.")
 
         # Because `self.unsilence` explicitly cancels this scheduled task, it is shielded
         # to avoid prematurely cancelling itself
-        await asyncio.shield(channel.ctx.invoke(self.unsilence))
+        await asyncio.shield(task.ctx.invoke(self.unsilence))
 
     async def _get_instance_vars(self) -> None:
         """Get instance variables after they're available to get from the guild."""
@@ -106,12 +109,12 @@ class Silence(Scheduler, commands.Cog):
 
         await ctx.send(f"{Emojis.check_mark} silenced current channel for {duration} minute(s).")
 
-        channel = SilencedChannel(
-            ctx=ctx,
+        task_data = TaskData(
             delay=duration*60,
+            ctx=ctx
         )
 
-        self.schedule_task(ctx.channel.id, channel)
+        self.schedule_task(ctx.channel.id, task_data)
 
     @commands.command(aliases=("unhush",))
     async def unsilence(self, ctx: Context) -> None:
--- a/bot/cogs/moderation/silence.py
+++ b/bot/cogs/moderation/silence.py
@@ -161,8 +161,8 @@ class Silence(Scheduler, commands.Cog):
         if current_overwrite.send_messages is False:
             await channel.set_permissions(self._verified_role, **dict(current_overwrite, send_messages=None))
             log.info(f"Unsilenced channel #{channel} ({channel.id}).")
-            self.notifier.remove_channel(channel)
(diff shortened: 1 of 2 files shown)


ItsDrike:
> I suggest using Scheduler inside cog, not with other class. One example is reminders.

I thought this was a bit cleaner, but yeah sure, I can do that. Another way would be to add this class to [`moderation/scheduler.py`](https://github.com/python-discord/bot/blob/91facc87d90c18f8cf1fe78d0b92f8bf82264b7c/bot/cogs/moderation/scheduler.py) together with the  `InfractionScheduler` and simply import it from there, so that the code of `silence.py` isn't that long.

So which approach should I take?

ks129:
@ItsDrike `silence.py` is good enough for this. The only use case for this is silence. Inside `scheduler.py` is `InfractionScheduler`, that is used by many files.

bot/cogs/moderation/silence.py
```diff
@@ -53,14 +61,30 @@ async def _notifier(self) -> None:
...
     def __init__(self, bot: Bot):
         self.bot = bot
         self.muted_channels = set()
         self._get_instance_vars_task = self.bot.loop.create_task(self._get_instance_vars())
         self._get_instance_vars_event = asyncio.Event()
+        super().__init__()
+
+    async def schedule_unsilence(self, channel: SilencedChannel) -> None:
+        """Schedule expiration for silenced channels."""
+        await self.bot.wait_until_guild_available()
```
MarkKoz, line 76: This seems ineffective; the guild could become unavailable between being scheduled and when the sleep in the task finishes. It's also redundant; the scheduled task invokes the unsilence command, which already waits for `_get_instance_vars_event`.

Removing this wait, `schedule_unsilence` becomes redundant since it just does a simple log (which is also redundant since the scheduler already logs this) and calls one function.

bot/cogs/moderation/silence.py
```diff
@@ -10,10 +12,16 @@
...
 from bot.converters import HushDurationConverter
+from bot.utils import time
 from bot.utils.checks import with_role_check
+from bot.utils.scheduling import Scheduler
 
 log = logging.getLogger(__name__)
 
+SilencedChannel = namedtuple(
+    "SilencedChannel", ("id", "ctx", "stop")
+)
```
MarkKoz, line 23: This is OK but I would prefer to use `typing.NamedTuple`.

ItsDrike:
@MarkKoz It should be fixed now

bot/cogs/moderation/silence.py
```diff
@@ -11,9 +11,14 @@
...
 from bot.converters import HushDurationConverter
 from bot.utils.checks import with_role_check
+from bot.utils.scheduling import Scheduler
 
 log = logging.getLogger(__name__)
 
+SilencedChannel = NamedTuple(
+    "SilencedChannel", [("ctx", Context), ("delay", int)]
+)
+
```
Numerlor, line 21: This should be a class definition https://github.com/python-discord/bot/blob/dd11ba4e6d6c33005c4eab10f4b8b2882f6f1f3d/bot/cogs/help_channels.py#L60-L64 as an example

bot/cogs/moderation/silence.py
```diff
@@ -103,7 +122,9 @@ async def unsilence(self, ctx: Context) -> None:
         """
         await self._get_instance_vars_event.wait()
         log.debug(f"Unsilencing channel #{ctx.channel} from {ctx.author}'s command.")
-        if await self._unsilence(ctx.channel):
+        if not await self._unsilence(ctx.channel):
+            await ctx.send(f"{Emojis.cross_mark} current channel is not silenced.")
```
Numerlor, line 126: Would `was not` be more fitting here?

bot/cogs/moderation/silence.py
```diff
@@ -141,6 +162,7 @@ async def _unsilence(self, channel: TextChannel) -> bool:
             await channel.set_permissions(self._verified_role, **dict(current_overwrite, send_messages=None))
             log.info(f"Unsilenced channel #{channel} ({channel.id}).")
             self.notifier.remove_channel(channel)
+            self.cancel_task(channel.id)
             self.muted_channels.discard(channel)
```
Numerlor, line 166: I feel like it would look better if the cancel task came before or after the `.remove_channel` and `.discard`, as those do the same job and make sense to be together

bot/cogs/moderation/silence.py
ItsDrike: Why should it be a class? Isn't it just doing the same thing in a longer approach?

Numerlor: It's clearer syntax, and there's no reason to avoid a class definition here

ItsDrike: Alright, I can do that, but I've noticed you replaced `ctx` with `callback`, should I just pass `lambda: ctx.invoke(self.unsilence)` or should I only go with `ctx.invoke` and passing the `self.unsilence` from `_scheduled_task` or did you have something else in mind?

MarkKoz: Pass in `ctx.invoke(self.unsilence)` without awaiting it. `ctx.invoke` is a coroutine function, which means it returns a coroutine object. A coroutine object is awaitable, hence satisfying the annotation `t.Awaitable`. Perhaps `coroutine` is a better name than `callback` but I think it's clear enough either way.

ItsDrike: Yes but that will produce RuntimeWarning if I unsilence prematurely since it won't be awaited

Numerlor: The above was just an example from a different module, not as a direct suggestion.
I think the names are fine with how they are now if you don't want to change the structure, so just need to adjust the syntax and add a docstring

MarkKoz: This is what I did to solve that issue https://github.com/python-discord/bot/blob/master/bot/cogs/help_channels.py#L843

ItsDrike: Oh, that's clever, so should I go with that, or as @Numerlor said, keep it as it was and only turn it into a class?

MarkKoz: Sorry, I forgot what I was looking at for a second 😅. You don't need to pass a coroutine. This would only really be beneficial if you needed to support different coroutines, as was the case in the help channel cog.

PR #25, opened 2018-03-04
Added first / last buttons and max_lines to pagination

runew0lf:

PR #981, opened 2020-06-04
Improve LinePaginator to support long lines

### LINKED ISSUE #926: RuntimeError: Line exceeds maximum page size 998
Sentry Issue: [BOT-3Q](https://sentry.io/organizations/python-discord/issues/1645724137/?referrer=github_integration)

```
RuntimeError: Line exceeds maximum page size 998
  File "bot/pagination.py", line 158, in paginate
    paginator.add_line(line, empty=empty)
  File "bot/pagination.py", line 72, in add_line
    raise RuntimeError('Line exceeds maximum page size %s' % (self.max_size - len(self.prefix) - 2))

Failed to add line to paginator: '===============
            Status: Inactive
            User: XAR#2912 (`627187206083051541`)
            Type: **note**
            Shadow: True
            Reason: discussing hacking in a voice channel and showing off a "hacking tool" they've made [here](https://discordapp.com/channels/267624335836053506/412357430186344448/706680820060127243). there also appears to be some sort of malicious script they've created [here](https://discordapp.com/channels/2676243358360535...
```

aeros:
Closes #926.

See the [issue](https://github.com/python-discord/bot/issues/926) for additional context on the changes.


aeros:
Note that this is still a WIP and not ready for review yet. I intentionally opened the PR as a draft to test against the CI while wrapping it up.

aeros:
I specifically requested a review from @MarkKoz and @lemonsaurus since they were the most involved with the discussion of these changes a few weeks ago.

tests/bot/test_pagination.py
```diff
@@ -8,17 +8,44 @@ class LinePaginatorTests(TestCase):
...
-        with self.assertRaises(RuntimeError, msg=message):
-            self.paginator.add_line('x' * self.paginator.max_size)
+        self.paginator = pagination.LinePaginator(prefix='', suffix='', max_size=30,
+                                                  scale_to_size=50)
 
     def test_add_line_works_on_small_lines(self):
         """`add_line` should allow small lines to be added."""
         self.paginator.add_line('x' * (self.paginator.max_size - 3))
+        # Note that the page isn't added to _pages until it's full.
+        self.assertEqual(len(self.paginator._pages), 0)
```
aeros, line 18: This change isn't directly related to the PR, but I think it's a good idea to ensure the correct number of pages are being created for the pagination tests, as it's relevant to all of them IMO. If needed, I can remove this.

bot/pagination.py
```diff
@@ -62,14 +73,26 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
-            raise RuntimeError('Line exceeds maximum page size %s' % (self.max_size - len(self.prefix) - 2))
+        remaining_words = None
+        if len(line) > (max_chars := self.max_size - len(self.prefix) - 2):
+            if len(line) > self.scale_to_size:
+                line, remaining_words = self._split_remaining_words(line, max_chars)
+                # If line still exceeds scale_to_size, we were unable to split into a second
+                # page without truncating.
+                if len(line) > self.scale_to_size:
+                    raise RuntimeError(f'Line exceeds maximum scale_to_size {self.scale_to_size}'
+                                       ' and could not be split.')
```
aeros, line 95: I'm somewhat undecided about whether it would be better to raise a `RuntimeError` or just truncate the remaining characters if `add_line()` contains a single word in `line` that contains more than `scale_to_size` characters. Since it would not occur from regular usage at all, I'm leaning towards an exception but I could definitely be persuaded otherwise.

bot/pagination.py
```diff
@@ -87,6 +110,41 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
+                    reduced_words.append(word)
+                    reduced_char_count += len(word)
+                else:
+                    is_full = True
+                    remaining_words.append(word)
+            else:
+                remaining_words.append(word)
+
+        return " ".join(reduced_words), " ".join(remaining_words) if len(remaining_words) > 2 \
+               else None
```
MarkKoz, line 146: I don't find using `\` to be particularly readable. I think it's better to do one of the following:

```py
        remaining_words = ...
        return " ".join(reduced_words), remaining_words

        # alternatively...
        return (
            ...,
            ...,
        )
```

bot/pagination.py
```diff
@@ -97,6 +155,7 @@ async def paginate(
         suffix: str = "",
         max_lines: t.Optional[int] = None,
         max_size: int = 500,
+        scale_to_size: int = 2000,
```
MarkKoz, line 158: This isn't actually being used anywhere inside the function. I think it's supposed to be used at line 204.

bot/pagination.py
```diff
@@ -87,6 +110,41 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
+            exceed `max_chars` (rounding down to the nearest word boundary).
+
+        remaining_words: the words in `line` which exceed `max_chars`. This value is None if
+            no words could be split from `line`.
+
+        Return a tuple in the format (reduced_words, remaining_words).
+        """
+        reduced_words = []
+        # "(Continued)" is used on a line by itself to indicate the continuation of last page
+        remaining_words = ["(Continued)\n", "---------------\n"]
```
MarkKoz, line 130: This is OK and simple to implement. However, I do wonder if we can come up with a cleaner way to convey this information e.g. putting it in the footer.

bot/pagination.py
MarkKoz: I think it should be truncated. I don't really see a benefit in raising an error. What if someone _really_ does want to print such a long string? There's no reason we should stop them. It'll also be easier for users of the paginator if they don't have to worry about such cases in the first place.

aeros: Alright, I'll go with truncation then and update the docstring accordingly.

bot/pagination.py
aeros: Hmm, good point. I think I'll go with the first approach then. I personally find the current approach to be decently easy to read, but I agree that it would improve readability with something like this (as suggested above):

```py
remaining_words = " ".join(remaining_words) if len(remaining_words) > 2 else None 
return " ".join(reduced_words), remaining_words
```

bot/pagination.py
aeros: Oops, good catch. I had intended to add it there but forgot to.

aeros:
> From testing with the infraction search command, I found some blank pages. The first page was blank and so was a page between the start of an infraction and its continuation page.

Hmm, that's odd. I didn't pick it up in my initial local server tests of the implementation but after clearing out the infraction history and adding a few pages worth of infraction notes, I did see the missing blank page at the start. I'll have to look further into the cause and see if I can add some additional unit tests to ensure it doesn't happen.

bot/pagination.py
aeros: I'm mostly just not a fan of truncating at specific point in the message and potentially cutting something off, but in the case of a single continuous word that exceeds `scale_to_size` it likely wouldn't cut off anything important.

bot/pagination.py
```diff
@@ -87,6 +110,41 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
+        reduced_char_count = 0
+        is_full = False
+
+        for word in line.split(" "):
+            if not is_full:
+                if len(word) + reduced_char_count <= max_chars:
+                    reduced_words.append(word)
+                    reduced_char_count += len(word)
```
aeros, line 138: I'm just realizing that this has a bit of a logical flaw: it doesn't account for the space characters. I'll have to adjust this approach slightly.

bot/pagination.py
lemonsaurus: I think `\` should be avoided in 99% of cases. I also prefer one of @MarkKoz's suggestions here.

aeros:
Thanks for the reviews Mark and Leon. I should have time to address the feedback in the next few days.

> I'd like to see a fuller PR description (embedding images is great when it's a bot feature)

I'm usually in the habit of keeping the PR description minimal and taking the time to explain features in something like a "What's New" document (since that's more visible, typically). But since we just have the PR descriptions here, I'll try to put more effort into that in the future. 

> I'd like to see commit messages that describe the commits, and I think your first commit here is much bigger than it probably should be. More atomic commits would make it easier to review.

In CPython, my typical workflow is squashing all of the commits into a single one before creating a PR, and using the PR's commit history as a means to track my responses to review feedback rather than showing the full work in progress. Since at the end it gets squashed into a single commit before merging, we typically don't put too much effort into the individual commits that make up the PR.  

But if a more atomic style showing the full history from start to finish is preferred here, I'll try to keep that in mind for future contributions.

(Similar to the approach here, we also advise contributors to avoid force pushing to the PR branch, but that's usually for the sake of showing the progress of the PR based on review feedback rather than having the full history present.)

lemonsaurus:
@aeros 

Makes sense it would be at odds with the cpython workflow.

Yeah, we neither squash, nor keep a changelog document, so commits and PR bodies are our opportunity to present the changes to the reviewers.

You are also welcome to force push to your feature branch here, although don't do it after opening your PR unless you have a very good reason, since it messes with the reviews

aeros:
@MarkKoz Would you mind taking a look over this again and testing it when you get the chance? I believe that I've addressed all of the critical issues. I also ended up having to make some fixes for the way new pages are created, as I found out that some of the existing parts of `LinePaginator.add_line()` were causing issues with the continuation pages.

Also, apologies for the late follow-up on this. I've been rather busy recently, and some of fixes weren't quite as straightforward as I was initially thinking they would be. Pagination can be a bit tricky to get right...

bot/pagination.py
```diff
@@ -62,23 +73,33 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
         """
         Adds a line to the current page.
 
-        If the line exceeds the `self.max_size` then an exception is raised.
+        If the line exceeds `self.max_size`, then `self.max_size` will go up to `scale_to_size` for
+        a single line before creating a new page. If it is still exceeded, the excess characters
+        are stored and placed on the next pages until there are none remaining (by word boundary).
```
lemonsaurus, line 78: I think this still needs a bit of work. It's not immediately clear to me how this works from this explanation. Furthermore, it fails to communicate _why_ we need two variables to determine the length of this line.

bot/pagination.py
```diff
@@ -62,23 +73,33 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
-        if self.max_lines is not None:
-            if self._linecount >= self.max_lines:
-                self._linecount = 0
-                self.close_page()
-
-            self._linecount += 1
-        if self._count + len(line) + 1 > self.max_size:
-            self.close_page()
+        remaining_words = None
+        if len(line) > (max_chars := self.max_size - len(self.prefix) - 2):
```
lemonsaurus, line 88: One of my first real-life encounters with a walrus! I actually quite like this line, nice use of hypermodern Python. :+1:

bot/pagination.py
```diff
@@ -62,23 +73,33 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
+            if len(line) > self.scale_to_size:
+                line, remaining_words = self._split_remaining_words(line, max_chars)
+                if len(line) > self.scale_to_size:
+                    log.debug("Could not continue to next page, truncating line.")
+                    line = line[:self.scale_to_size]
+
+        if self.max_lines is not None and self._linecount >= self.max_lines:
```
lemonsaurus, line 95: Could probably install a block comment for this block and maybe one or two of the others in this section. This code is pretty readable, don't get me wrong, but remember that we're writing software which often serves as an example for beginners and intermediates, and readability should be considered particularly crucial for this repo so that each file can maximize its educational value.

bot/pagination.py
```diff
@@ -87,6 +108,53 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
             self._current_page.append('')
             self._count += 1
 
+        if remaining_words:
+            self._new_page()
+            self.add_line(remaining_words)
+
+    def _new_page(self) -> None:
```
lemonsaurus, line 115: Docstrings, even for private helper methods? Yes please!

bot/pagination.py
```diff
@@ -62,23 +73,33 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
         """
         Adds a line to the current page.
 
-        If the line exceeds the `self.max_size` then an exception is raised.
+        If the line exceeds `self.max_size`, then `self.max_size` will go up to `scale_to_size` for
+        a single line before creating a new page. If it is still exceeded, the excess characters
+        are stored and placed on the next pages until there are none remaining (by word boundary).
+
+        Raises a RuntimeError if `self.max_size` is still exceeded after attempting to continue
+        onto the next page.
```
aeros, line 81: Oops, this still needs to be updated to mention that it truncates the line instead of raising a `RuntimeError`. Heading to sleep now, but I should be able to find the time to make a quick fix tomorrow.

bot/pagination.py
```diff
@@ -87,6 +108,53 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
+            exceed `max_chars` (rounding down to the nearest word boundary).
+
+        remaining_words: the words in `line` which exceed `max_chars`. This value is None if
+            no words could be split from `line`.
+
+        Return a tuple in the format (reduced_words, remaining_words).
+        """
+        reduced_words = []
+        remaining_words = []
+        # "(Continued)" is used on a line by itself to indicate the continuation of last page
```
lemonsaurus, line 134: Nitpick, but block comments trapped between two lines of code feel really claustrophobic to me.
```suggestion

        # "(Continued)" is used on a line by itself to indicate the continuation of last page
```

bot/pagination.py
```diff
@@ -87,6 +108,53 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
+        reduced_char_count = 0
+        is_full = False
+
+        for word in line.split(" "):
+            if not is_full:
+                if len(word) + reduced_char_count <= max_chars:
+                    reduced_words.append(word)
+                    reduced_char_count += len(word) + 1
+                else:
+                    # If reduced_words is empty, we were unable to split the words across pages
```
lemonsaurus, line 145: However, in this case I think a newline on the previous line would look messy. The indentation makes it okay.

bot/pagination.py
```diff
@@ -87,6 +108,53 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
+            no words could be split from `line`.
+
+        Return a tuple in the format (reduced_words, remaining_words).
+        """
+        reduced_words = []
+        remaining_words = []
+        # "(Continued)" is used on a line by itself to indicate the continuation of last page
+        continuation_header = "(Continued)\n-----------\n"
+        reduced_char_count = 0
+        is_full = False
```
lemonsaurus, line 137: I'm on the fence about this boolean. I think naming boolean variables `is_adjective` or `has_noun` or whatever is a nice practice, but on the other hand, if this was just called `full` the code below would read `if not full`, which is lovely.

your call.

bot/pagination.py
```diff
@@ -37,12 +37,19 @@ class LinePaginator(Paginator):
...
+        The maximum amount of characters a single line can scale up to.
     * max_lines: `int`
         The maximum amount of lines allowed in a page.
     """
 
     def __init__(
-        self, prefix: str = '```', suffix: str = '```', max_size: int = 2000, max_lines: int = None
+        self,
+        prefix: str = '```',
+        suffix: str = '```',
```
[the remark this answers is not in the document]
aeros: I personally have no idea, but I can try to take a look at the current places where LinePaginator is used to see if there are any decent use cases for the prefix and suffix parameters. From looking at the git blame, it looks like it was all added at once a couple of years ago: https://github.com/python-discord/bot/pull/25. It could very well be a common case of feature bloat, but I'm not confident enough about the areas in which `LinePaginator` is used by the bot to know if those params are useful.

bot/pagination.py
```diff
@@ -52,6 +59,10 @@ def __init__(
         self.prefix = prefix
         self.suffix = suffix
         self.max_size = max_size - len(suffix)
+        if scale_to_size < max_size:
+            raise ValueError("scale_to_size must be >= max_size.")
```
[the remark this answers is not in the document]
aeros: Sounds good, I'm very rarely against adding additional potentially useful information to exception messages.

bot/pagination.py
aeros: Hmm, okay. I'll try to work on expanding this docstring a bit. In hindsight, I agree that it doesn't really address the reasoning behind having both `max_size` and `scale_to_size`.

bot/pagination.py
aeros: Me as well! I've only recently started to wrap my head around how it can be useful in real-world code. My use cases for it aren't incredibly common, but it's great when you want to avoid repeating yourself for a re-used value from a conditional.

bot/pagination.py
aeros: With the ambiguity that Python has with `bool(obj)`, I put a lot of value is using the `is_`/`has_` prefix to directly communicate to readers (and myself, later down the road) that it's referring to a strictly boolean value. In this case, a name like `full` might be obvious and `is not full` sounds great, but I personally find it to be more explicit.

bot/pagination.py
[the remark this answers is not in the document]
lemonsaurus: well, if it was added a couple of years ago, I suggest we could probably just remove these as arguments and make them constants if (and only if) there are zero usages in the codebase. If we've never used those arguments in 2 years, we're not gonna start using them tomorrow.

Would you mind taking a quick look to see if you can find any usages?

bot/pagination.py
lemonsaurus: then keep it as is. :+1:

bot/pagination.py
lemonsaurus: I may have been a little imprecise here, but this is actually a request for a docstring :grin:

Den4200:
Since this PR will allow the paginator to handle long lines, I think that nomination reasons should not be shortened when looking at someone's nomination history.

You would just have to get rid of the `textwrap.shorten` call in these places:
https://github.com/python-discord/bot/blob/232cc68b0a0ef210027beacb9b4f50ffeeaddd00/bot/cogs/watchchannels/talentpool.py#L227

https://github.com/python-discord/bot/blob/232cc68b0a0ef210027beacb9b4f50ffeeaddd00/bot/cogs/watchchannels/talentpool.py#L240-L243

bot/pagination.py
```diff
@@ -87,6 +108,54 @@ def add_line(self, line: str = '', *, empty: bool = False) -> None:
...
+        if remaining_words:
+            self._new_page()
+            self.add_line(remaining_words)
+
+    def _new_page(self) -> None:
+        self._linecount = 0
+        self._count = len(self.prefix) + 1
+        self.close_page()
+
+    def _split_remaining_words(self, line: str, max_chars: int) -> t.Tuple[str, t.Optional[str]]:
```
MarkKoz, line 120: It would be nice for this to append an ellipses to the reduced words.

bot/pagination.py
aeros: Oh, I see. I'll add a brief docstring to `_new_page` :+1:.

bot/pagination.py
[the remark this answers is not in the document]
aeros: @lemonsaurus

From looking through all of the results in `git grep "LinePaginator"`, I found just one actual usage for *prefix* in the bot repo:

https://github.com/python-discord/bot/blob/232cc68b0a0ef210027beacb9b4f50ffeeaddd00/bot/cogs/help.py#L296-L302

Since this would require some additional investigation to see if the same functionality could be easily implemented without it, I think we should probably consider removal of *prefix* and *suffix* in a separate issue and PR (if at all, since you mentioned that they should only be removed IFF there are zero usages). Arguably, we could just make suffix a constant, but to me it seems like it would be a bit odd to just remove one without the other.

Would this be acceptable? While I think it would be useful to remove these parameters if they're not really being used, I'm definitely leaning towards handling it independently from this PR.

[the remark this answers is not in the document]
MarkKoz: I agree that it is better left for another PR, though I don't have a problem with it just being left as-is.

aeros:
Good catch @MarkKoz. 

> Is it better to require it to be explicitly set it to 2040 or should it automatically set it equal to max_size if unspecified and greater than the default value?

I'm not 100% certain on this. I think for most cases setting `scale_to_size` to `max_size` if it's unspecified and greater than the default of 2,000 would be fine, but it does result in an extra bit of added complexity to the constructor.

On a related note though, I discovered that there's a maximum (presumably on the end of the discord API) embed length, and if exceeded, it results in an `HTTPException`: 

```
In embed.description: Must be 2048 or fewer in length.
```

So perhaps we should also set a hard limit for both `max_size` and `scale_to_size` in the constructor for LinePaginator. I'm a bit concerned that 2040 might be a bit close for comfort for `max_size` with the continuation header being 24 characters. That could potentially put it over the limit with an exceptionally large message. So maybe the limit for `max_size` should be 2,000, and if exceeded, it would raise a `ValueError`? I could rather easily locate anywhere that it goes over (such as for the help command) and lower it accordingly without any real impact to usability.

MarkKoz:
Yeah, a limit of 2000 is reasonable. Then we don't have to worry about changing `scale_to_size` since its default will already be the maximum.

aeros:
Also, thanks for bringing up the nomination reasons @Den4200. I've removed the `textwrap.shorten()`s in a recent commit.

aeros:
I just saw another location that exceeds max_size of 2,000 (also in `help.py`), I'll fix that real quick:

```
bot/cogs/doc.py:371:                await LinePaginator.paginate(lines, ctx, inventory_embed, max_size=400, empty=False)
bot/cogs/extensions.py:162:        await LinePaginator.paginate(lines, ctx, embed, max_size=300, empty=False)
bot/cogs/help.py:302:            max_size=2000,
bot/cogs/help.py:349:        await LinePaginator.paginate(pages, self.context, embed=embed, max_lines=1, max_size=2040)
bot/cogs/moderation/management.py:241:            max_size=1000
bot/cogs/off_topic_names.py:162:            await LinePaginator.paginate(lines, ctx, embed, max_size=400, empty=False)
bot/cogs/off_topic_names.py:181:            await LinePaginator.paginate(lines, ctx, embed, max_size=400, empty=False)
bot/cogs/watchchannels/talentpool.py:141:            max_size=1000
bot/pagination.py:241:        paginator = cls(prefix=prefix, suffix=suffix, max_size=max_size, max_lines=max_lines,
tests/bot/test_pagination.py:11:        self.paginator = pagination.LinePaginator(prefix='', suffix='', max_size=30,
```

bot/pagination.py
```diff
@@ -58,10 +58,20 @@ def __init__(
...
+        if max_size > 2000:
+            raise ValueError(f"max_size must be <= 2,000 characters. ({max_size} > 2000)")
+
         self.max_size = max_size - len(suffix)
+
         if scale_to_size < max_size:
             raise ValueError(f"scale_to_size must be >= max_size. ({scale_to_size} < {max_size})")
 
+        if scale_to_size > 2000:
+            raise ValueError(f"max_size must be <= 2,000 characters. ({scale_to_size} > 2000)")
```
MarkKoz, line 73:
```suggestion
            raise ValueError(f"scale_to_size must be <= 2,000 characters. ({scale_to_size} > 2000)")
```

aeros: Oops, thanks for the catch. Somehow I fixed the f-string but missed the message :-)

aeros:
Thanks for the extensive review and merging @MarkKoz. :-)

bot/pagination.py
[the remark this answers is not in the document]
lemonsaurus: if we're using it we can just keep it.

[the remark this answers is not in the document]
aeros: Alright, sounds good to me.

PR #987, opened 2020-06-05
Add persistence to the help channel system

lemonsaurus:
This PR should ensure that help channel-related data is persisted across bot restarts.

To achieve this, we make two changes.

## RedisCache now supports Booleans as values
For the unanswered cache, we need to know whether or not a channel has been answered, and so it was relevant to store a boolean in the RedisCache. There was no support for this, so I added some. It seems like a feature we might want in the future for other things anyway.

## Refactoring help_channels.py
The rest is pretty straight forward - Replacing the `help_channel_claimants`, `unanswered`, and `claim_times` dicts with RedisCaches, and changing the code that accessed it to do so in a RedisCache compatible way.

### DIFF
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -9,12 +9,14 @@ from contextlib import suppress
 from datetime import datetime
 from pathlib import Path
 
+import dateutil
 import discord
 import discord.abc
 from discord.ext import commands
 
 from bot import constants
 from bot.bot import Bot
+from bot.utils import RedisCache
 from bot.utils.checks import with_role_check
 from bot.utils.scheduling import Scheduler
 
@@ -99,13 +101,24 @@ class HelpChannels(Scheduler, commands.Cog):
     Help channels are named after the chemical elements in `bot/resources/elements.json`.
     """
 
+    # This cache tracks which channels are claimed by which members.
+    # RedisCache[discord.TextChannel.id, t.Union[discord.User.id, discord.Member.id]]
+    help_channel_claimants = RedisCache()
+
+    # This cache maps a help channel to whether it has had any
+    # activity other than the original claimant. True being no other
+    # activity and False being other activity.
+    # RedisCache[discord.TextChannel.id, bool]
+    unanswered = RedisCache()
+
+    # This dictionary maps a help channel to the time it was claimed
+    # RedisCache[discord.TextChannel.id, datetime.datetime]
+    claim_times = RedisCache()
+
     def __init__(self, bot: Bot):
         super().__init__()
 
         self.bot = bot
-        self.help_channel_claimants: (
-            t.Dict[discord.TextChannel, t.Union[discord.Member, discord.User]]
-        ) = {}
 
         # Categories
         self.available_category: discord.CategoryChannel = None
@@ -124,16 +137,6 @@ class HelpChannels(Scheduler, commands.Cog):
         self.ready = asyncio.Event()
         self.on_message_lock = asyncio.Lock()
         self.init_task = self.bot.loop.create_task(self.init_cog())
-
-        # Stats
-
-        # This dictionary maps a help channel to the time it was claimed
-        self.claim_times: t.Dict[int, datetime] = {}
-
-        # This dictionary maps a help channel to whether it has had any
-        # activity other than the original claimant. True being no other
-        # activity and False being other activity.
-        self.unanswered: t.Dict[int, bool] = {}
 
     def cog_unload(self) -> None:
         """Cancel the init task and scheduled tasks when the cog unloads."""
@@ -197,7 +200,7 @@ class HelpChannels(Scheduler, commands.Cog):
 
     async def dormant_check(self, ctx: commands.Context) -> bool:
         """Return True if the user is the help channel claimant or passes the role check."""
-        if self.help_channel_claimants.get(ctx.channel) == ctx.author:
+        if await self.help_channel_claimants.get(ctx.channel.id) == ctx.author.id:
             log.trace(f"{ctx.author} is the help channel claimant, passing the check for dormant.")
             self.bot.stats.incr("help.dormant_invoke.claimant")
             return True
@@ -223,7 +226,7 @@ class HelpChannels(Scheduler, commands.Cog):
         if ctx.channel.category == self.in_use_category:
             if await self.dormant_check(ctx):
                 with suppress(KeyError):
-                    del self.help_channel_claimants[ctx.channel]
+                    await self.help_channel_claimants.delete(ctx.channel.id)
 
                 await self.remove_cooldown_role(ctx.author)
                 # Ignore missing task when cooldown has passed but the channel still isn't dormant.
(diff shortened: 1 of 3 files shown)


bot/utils/redis_cache.py
```diff
@@ -131,6 +134,13 @@ def _from_typestring(key_or_value: Union[bytes, str], prefixes: _PrefixTuple) ->
         # Now we convert our unicode string back into the type it originally was.
         for prefix, _type in prefixes:
             if key_or_value.startswith(prefix):
+
+                # For booleans, we need special handling because bool("False") is True.
+                if prefix == "b|":
+                    value = key_or_value[len(prefix):]
+                    return bool(strtobool(value))
```
MarkKoz, line 141: Is it feasible to store it as 0/1 instead? Then it becomes `bool(int(value))`. I prefer that simplicity over dealing with actual strings and relying on a separate method.

bot/cogs/help_channels.py
```diff
@@ -680,12 +684,12 @@ async def on_message(self, message: discord.Message) -> None:
...
             await self.revoke_send_permissions(message.author)
             # Add user with channel for dormant check.
-            self.help_channel_claimants[channel] = message.author
+            await self.help_channel_claimants.set(channel.id, message.author.id)
 
             self.bot.stats.incr("help.claimed")
 
-            self.claim_times[channel.id] = datetime.now()
-            self.unanswered[channel.id] = True
+            await self.claim_times.set(channel.id, str(datetime.now()))
```
MarkKoz, line 691: Something irks me about storing timestamps as strings. Using a POSIX timestamps (i.e. relative to the epoch) seems more reliable. By the way, should this be in UTC or does it not really matter as long as it's also decoded the same way?

bot/cogs/help_channels.py
```diff
@@ -546,13 +549,14 @@ async def move_to_dormant(self, channel: discord.TextChannel, caller: str) -> No
 
         self.bot.stats.incr(f"help.dormant_calls.{caller}")
 
-        if channel.id in self.claim_times:
-            claimed = self.claim_times[channel.id]
+        if await self.claim_times.contains(channel.id):
+            claimed_datestring = await self.claim_times.get(channel.id)
```
ks129, line 553: Maybe should better to use just `get` and make `None` check instead first `contains` and then `get`.

bot/cogs/help_channels.py
```diff
@@ -546,13 +549,14 @@ async def move_to_dormant(self, channel: discord.TextChannel, caller: str) -> No
...
+            claimed = dateutil.parser.parse(claimed_datestring)
             in_use_time = datetime.now() - claimed
             self.bot.stats.timing("help.in_use_time", in_use_time)
 
-        if channel.id in self.unanswered:
-            if self.unanswered[channel.id]:
+        if await self.unanswered.contains(channel.id):
+            if await self.unanswered.get(channel.id):
```
ks129, line 559: Same as L552-L553

bot/cogs/help_channels.py
```diff
@@ -223,7 +226,7 @@ async def close_command(self, ctx: commands.Context) -> None:
         if ctx.channel.category == self.in_use_category:
             if await self.dormant_check(ctx):
                 with suppress(KeyError):
-                    del self.help_channel_claimants[ctx.channel]
+                    await self.help_channel_claimants.delete(ctx.channel.id)
```
ks129, line 229: I think instead suppressing this should use try-except with some logging.

SebastiaanZ: I disagree. We want to delete a mapping (channel -> claimant) that may not exist, so we either delete (goal accomplished) or we observe that it did not exist in the first place (goal accomplished). The latter situation is less likely now that we're using caching, but we're okay with it happening here. The goal is to not have the mapping anyway.

bot/cogs/help_channels.py
SebastiaanZ: I agree; this is a potential race condition, as both the contains check and the get action are awaited. If another task "runs" after the contains was requested from Redis but before `get` is request from Redis, it could change the state we've just checked for.

bot/cogs/help_channels.py
lemonsaurus: > Using a POSIX timestamps (i.e. relative to the epoch) seems more reliable
I don't really mind either approach, but I'm fine with changing it to do this.

> By the way, should this be in UTC
I can't see that it should matter. Can you think of an edge case where it would?

Akarys42: I think we use UTC based time everywhere in the bot, it would be nice for consistency and in case if debugging.

bot/utils/redis_cache.py
lemonsaurus: sure, but we're still turning that int into a string to store it in redis. But it might remove the reliance on strtobool, so I guess it'd be fine.

bot/cogs/help_channels.py
SebastiaanZ: That's not true; we both use `datetime.datetime.now` and `datetime.datetime.utcnow` and even in this very extension, `datetime.datetime.now` is used. Both return naive datetime objects without tz information and they both currently return the same time. Granted, the latter could change if were to change the local time settings of the VPS.

lemonsaurus: I could change these all to `utcnow` since they are more consistent.

bot/cogs/help_channels.py
lemonsaurus: Yep, agreed.

bot/cogs/help_channels.py
kwzrd: Interesting how different opinions we have on this. My approach would have been `isoformat` and `fromisoformat` using UTC, the string representation requires importing `dateutil` which is fine of course, but it still feels like an extra step (and I personally don't know what `str(dt_instance)` looks like at all).

bot/cogs/help_channels.py
kwzrd: I agree that this is a nice use case for `suppress`, I do not see the need to log whether the key existed or not since it doesn't affect any logic further down the line.

j03b: Correct me if I'm wrong, but why are we using suppress here? If a key that does not exist is requested to be deleted in Redis it should not raise a KeyError, in the RedisCache docs that @lemonsaurus wrote it states that deleting non-existent keys will be silent.

lemonsaurus: yeah @j03b is right. it's not necessary.

lemonsaurus: Addressed in https://github.com/python-discord/bot/pull/987/commits/15dbbcf865dd24f5f8697fd85bd60d53d9450fcf

bot/cogs/help_channels.py
```diff
@@ -546,19 +548,20 @@ async def move_to_dormant(self, channel: discord.TextChannel, caller: str) -> No
...
             self.bot.stats.timing("help.in_use_time", in_use_time)
 
-        if channel.id in self.unanswered:
-            if self.unanswered[channel.id]:
+        unanswered = await self.unanswered.get(channel.id)
+        if unanswered is not None:
+            if unanswered:
                 self.bot.stats.incr("help.sessions.unanswered")
             else:
                 self.bot.stats.incr("help.sessions.answered")
```
MarkKoz, line 569: Thoughts of changing this the following?

```py
if unaswered:
    ...
elif unanswered is not None:
    ...
```

MarkKoz:
I made a change to use the cache to remove expired cooldowns or re-schedule active cooldowns when the bot restarts. See b49f3e5.

bot/cogs/help_channels.py
kosayoda: Don't see why not.

PR #991, opened 2020-06-07
Create cooldown.md

crazygmr101:


bot/resources/tags/cooldown.md
```diff
@@ -0,0 +1,22 @@
+**Cooldowns**
+
+Cooldowns are used in discord.py to rate-limit.
```
j03b, line 3: Rate-limit what? Probably worth mentioning commands here.
```suggestion
Cooldowns are used in discord.py to rate-limit command execution.
```

bot/resources/tags/cooldown.md
```diff
@@ -0,0 +1,22 @@
...
+
+```python
+from discord.ext import commands
+
+class SomeCog(commands.Cog):
+    def __init__(self):
+        self._cd = commands.CooldownMapping.from_cooldown(1.0, 60.0, commands.BucketType.user)
+
+    async def cog_check(self, ctx):
```
j03b, line 12: I'm not incredibly well versed, but why this approach instead of the decorator based approach? For beginners that may come across easier.

crazygmr101: This isn't for specifically ratelimiting commands. It's to manage your own custom cooldowns, whether it be emojis in chat, links, or anything else.

bot/resources/tags/cooldown.md
crazygmr101: True, but the point of this is for custom cooldowns. Would !customcooldown or !customcd be a better tag?

bot/resources/tags/cooldown.md
j03b: Ah gotcha

bot/resources/tags/cooldown.md
j03b: I think `customcooldown` would be a more suitable name (we tend to avoid abbreviations where possible), we can add a cooldown tag in future if we want.

bot/resources/tags/customcooldown.md
```diff
@@ -0,0 +1,20 @@
+**Cooldowns**
```
MarkKoz, line 1: How about this?

```suggestion
**Cooldowns in discord.py**
```

bot/resources/tags/customcooldown.md
```diff
@@ -0,0 +1,20 @@
...
+
+```python
+from discord.ext import commands
+
+message_cooldown = commands.CooldownMapping.from_cooldown(1.0, 60.0, commands.BucketType.user)
+
+@bot.event
+async def on_message(message):
+    bucket = message_cooldown.get_bucket(message)
+    retry_after = bucket.update_rate_limit()
```
MarkKoz, line 13: I think it's worth mentioning what this return value is. It's hard to tell what "retry after" means if one hasn't encountered that term before.

bot/resources/tags/customcooldown.md
```diff
@@ -0,0 +1,20 @@
...
+    if retry_after:
+        await message.channel.send("Slow down! You're sending messages too fast")
+    else:
+        await message.channel.send("Not ratelimited!")
+```
+
+`from_cooldown` takes the amount of `update_rate_limit()`s needed to trigger the cooldown, the time in which the cooldown is triggered, and a [`BucketType`](discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.discord.ext.commands.BucketType).
```
MarkKoz, line 20: Why does the URL have no scheme?

bot/resources/tags/customcooldown.md
```diff
@@ -0,0 +1,22 @@
...
+message_cooldown = commands.CooldownMapping.from_cooldown(1.0, 60.0, commands.BucketType.user)
+
+@bot.event
+async def on_message(message):
+    bucket = message_cooldown.get_bucket(message)
+    # update_rate_limit returns a time you need to wait before
+    # trying again
+    retry_after = bucket.update_rate_limit()
+    if retry_after:
+        await message.channel.send("Slow down! You're sending messages too fast")
```
MarkKoz, line 17: Using `retry_after ` in the message demonstrates what the variable means in a clean, succinct way compared to a comment. The f-string may confuse some, but I think it's a worthy trade-off.
```suggestion
    retry_after = bucket.update_rate_limit()
    if retry_after:
        await message.channel.send(f"Slow down! Try again in {retry_after} seconds.")
```

PR #993, opened 2020-06-08
Implement #incidents channel automation

### LINKED ISSUE #992: Automation for the #incidents channel
Previously tracked here: https://github.com/python-discord/organisation/issues/255

# Basic proposal

Recently, we have introduced a new `#incidents` channel, which now hosts all incident reporting from the `@Helpers+` team. Moderators are encouraged to use 3 status emojis to signal their involvement, if any, with the incident at hand:

image

To make things easier for moderators, we have decided to introduce automation to the emoji signaling system. The idea is to have the bot automatically leave the 3 possible emoji choices under each new message - moderators can then simply click the one they wish to choose. The bot then clears the remaining emojis. Should the moderator un-select their choice, the bot adds the 3 choices again, e.g. to allow a smooth transition from *investigating* to *actioned*.

### Considerations

The desired behaviour in situations where *multiple* moderators react is currently undefined. I haven't put an extensive amount of thought to this, so there may be additional edge cases, but one of the questions we should ask ourselves is what happens if 2 moderators select a reaction, and subsequently, what happens if one un-selects.

The bot can either:
1. Ignore the situation - as long as there is at least *one* react, no bot action is needed
2. Clear *all* reactions, add the 3 options again, and await further input

I'm currently leaning towards option 2, since that will allow, for example, clearing another user's "forgotten" / invalid signal. It would be good to hear others' opinions on this, since we touched on the problem during the staff meet, but afaik haven't come to a definitive answer.

# Advanced features

Additionally, the following functionality (and more) could be supported.

### Incident clearing

Since the bot can now determine the status of any incident, it will also be possible to clear solved incidents (after some period of time, or upon an actioned / unactioned react), like we currently do using internal eval on-demand. It is also possible to relay solved incidents into a read-only channel (a webhook would be nice) for archiving purposes - this channel should probably be visible to `@Helpers`, so that they can review the status of their report despite it being cleared from the active `#incidents` pool.

Implementation-wise, the most important question here is how to schedule the incident clearing. If the incident gets cleared instantly, the first actioned / unactioned react will be definitive. That would actually solve the previously mentioned issues with multi-react, but is that what we want?

### Highlighting unresolved incidents

Incidents that have *not* been resolved (i.e. reacted to with either tick or cross) should be highlighted after some time. The bot could, for example, ping `@Moderators` with a summary of unresolved incidents, and links to them. This could be done in `#mods`, or `#incidents` directly (it could probably be treated as an incident of its own, cleared with a react as usual).

# Afterword

I'm interested in implementing this, but I'd like us to continue the conversation about exactly how we'd like it to behave. Let's make sure we don't introduce a system that we will end up fighting - its primary and perhaps only purpose is to make helpers' and mods' lives easier. I've tried to highlight a few unanswered questions, and I'd be really happy to hear others' opinions on them.

kwzrd:
Closes #992.

# Incidents automation

Although we originally had more ambitious plans (see #992), this PR adds fairly simple automation for the `#incidents` channel. The cog does not currently offer any commands, it merely responds to `on_message` and `on_reaction_add` events in the channel.

The reason why this PR does not add all of the functionality proposed in the issue above (especially @j03b's response) is primarily rate-limits. The inability to use the _clear all reactions_ endpoint to remove the bot's own reactions means that a large amount of _add_ and _remove_ requests must be sent to Discord, and `d.py` will prevent us from doing even the following pattern:

1. Bot adds the 3 emoji
2. User clicks `investigating`
3. Bot removes the 3 emoji from itself, but leaves the user's (requires 3 separate requests)

This will result in 6 requests in quick succession, which overflows `d.py`'s bucket (5) and the last request is not dispatched.

## Behaviour

The cog's behaviour can be summarized as follows.

For each `on_message`, the three reaction emojis will be automatically added (investigating, actioned, unactioned). This is for moderator convenience ~ mods no longer have to search for the emoji, they will simply select one from the available options.

image

Messages starting with `#` will be ignored by the bot:

image

Ignored messages must be deleted manually. The same applies for pinned messages, as those are considered "header", and shouldn't be archived.

---

If a moderator clicks either actioned or unactioned, the message will be relayed to the `#incidents-archive` channel. Other emoji, or emoji from roles below `@Moderators`, will be removed. The original message is removed if the relay was successful.

Now in `#incidents-archive`:

image

Relaying attachments is also possible:

image

This only works for image-like attachments that can be displayed in the embed. Other types of attachments, e.g. a text file, are ignored, as there is no way to display them.

---

On bot start-up, it will automatically crawl the channel to retroactively add missing reactions where appropriate, in case an incident was reported while the bot was offline.

Please see the diff for the cog's docstring, it gives a more detailed description, and references the listeners directly.

### Implementation

- [x] Add incident signaling emojis to the emoji server
- [x] Register new emojis in the config & load them into constants at runtime
- [x] Create an `#incidents` channel in the test server
- [x] Create an `#incidents-archive` channel in the production & test server
- [x] Add `#incidents-archive` channel to config & constants
- [x] Learn how to setup & use a webhook in the channel
- [x] Add webhook to config & constants
- [x] Implement `on_message` listener & internals
- [x] Implement `on_reaction` listener & internals
- [x] Relay incident attachments
- [x] Implement channel crawler for retro-actively adding emoji on missed messages
- [x] Finish documentation

### Tests

**Module coverage**: 100% 🎉

- [x] download_file
- [x] make_embed
- [x] is_incident
- [x] own_reactions
- [x] has_signals
- [x] add_signals
- [x] crawl_incidents
- [x] archive
- [x] make_confirmation_task
- [x] process_event
- [x] resolve_message
- [x] on_raw_reaction_add
- [x] on_message

### After merge

- [ ] Make archive channel visible to all staff (currently `@Core Developers` only)
- [ ] Probably a staff announcement to explain how this works?

## New config

Since I've added a few new config variables (channels & webhook), I will list the values of their counterparts in the test server here for convenience: you will have to add these to your local config files in order to test this feature.
```
channels:
  incidents: 714214212200562749
  incidents_archive: 720668923636351037

webhooks:
  incidents_archive: 720666878367760457
```

## Concerns

~~So far the only concern about my implementation that I have is that since the bot uses and reacts *only* to emoji defined in the `Signal` enum, staff who aren't present in the emoji server will not be able to manually close incidents if the bot is down while a new incident pops up.~~

Solved this by having the bot crawl the channel on start-up and add missing emoji where necessary.


kwzrd:
For a quick status update: I'm now fairly happy with the implementation and have started to move towards writing tests.

config-default.yml
```diff
@@ -251,7 +256,7 @@ guild:
         duck_pond:                      637821475327311927
         dev_log:                        680501655111729222
         python_news:    &PYNEWS_WEBHOOK 704381182279942324
-
+        incidents_archive:              720671599790915702
```
kwzrd, line 259: Should `&PYNEWS_WEBHOOK` be indented further? I didn't want to introduce extra unnecessary diff to the config file, but depending on how you look at it, the alignment is now "broken", although at the same time it's not an issue as everything fits.

tests/bot/cogs/moderation/test_incidents.py
```diff
@@ -0,0 +1,675 @@
...
+        Feed the next item, or raise `StopAsyncIteration`.
+
+        Since we're wrapping a sync iterator, it will communicate that it has been depleted
+        by raising a `StopIteration`. The `async for` construct does not expect it, and we
+        therefore need to substitute it for the appropriate exception type.
+        """
+        try:
+            return next(self.iter_messages)
+        except StopIteration:
+            raise StopAsyncIteration
```
kwzrd, line 57: Is there something else I could have done this to mock an async iterator? This seems like unnecessary boilerplate, but I haven't found anything in `unittest` that would support an `async for` loop out of the box.

tests/bot/cogs/moderation/test_incidents.py
```diff
@@ -0,0 +1,675 @@
...
+        Although we do not need to make any assertions about the task itself while
+        testing the listener, the code will still await it and call the `done` method,
+        and so we must inject something that will not fail on either action.
+
+        Note that this is done in an `asyncSetUp`, which runs after `setUp`.
+        The justification is that creating an actual task requires the event
+        loop to be ready, which is not the case in the `setUp`.
+        """
+        mock_task = asyncio.create_task(AsyncMock()())  # Mock async func, then a coro
+        self.cog_instance.crawl_task = mock_task
```
kwzrd, line 570: I find this super ew, but I just wasn't able to come up with anything better.

kwzrd:
This is now considered ready for review.

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,334 @@
...
+            # First we try to grab the webhook
+            webhook: discord.Webhook = await self.bot.fetch_webhook(Webhooks.incidents_archive)
+
+            # Now relay the incident
+            message: discord.Message = await webhook.send(
+                content=incident.clean_content,  # Clean content will prevent mentions from pinging
+                username=incident.author.name,
+                avatar_url=incident.author.avatar_url,
+                wait=True,  # This makes the method return the sent Message object
+            )
```
kwzrd, line 172: This should apply the clyde cleanse from #1009 once it is merged.

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,334 @@
...
+        not all information was relayed, return False. This signals that the original
+        message is not safe to be deleted, as we will lose some information.
+        """
+        log.debug(f"Archiving incident: {incident.id} with outcome: {outcome}")
+        try:
+            # First we try to grab the webhook
+            webhook: discord.Webhook = await self.bot.fetch_webhook(Webhooks.incidents_archive)
+
+            # Now relay the incident
+            message: discord.Message = await webhook.send(
```
ks129, line 167: Not sure is this type hint required here.

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,334 @@
...
+
+        timeout = 5  # Seconds
+        confirmation_task = self.make_confirmation_task(incident, timeout)
+
+        log.trace("Deleting original message")
+        await incident.delete()
+
+        log.trace(f"Awaiting deletion confirmation: {timeout=} seconds")
+        try:
+            await confirmation_task
```
ks129, line 247: Why not calling `self.make_confirmation_task` here and awaiting same time?

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,334 @@
...
+
+        In an edge-case, it is also possible that the message was already deleted, and
+        the API will respond with a 404. In such a case, None will be returned.
+        This signals that the event for `message_id` should be ignored.
+        """
+        await self.bot.wait_until_guild_available()  # First make sure that the cache is ready
+        log.trace(f"Resolving message for: {message_id=}")
+        message: discord.Message = self.bot._connection._get_message(message_id)
+
+        if message is not None:
```
ks129, line 270:
```suggestion
        if message:
```

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,334 @@
...
+
+def is_incident(message: discord.Message) -> bool:
+    """True if `message` qualifies as an incident, False otherwise."""
+    conditions = (
+        message.channel.id == Channels.incidents,  # Message sent in #incidents
+        not message.author.bot,                    # Not by a bot
+        not message.content.startswith("#"),       # Doesn't start with a hash
+        not message.pinned,                        # And isn't header
+    )
+    return all(conditions)
```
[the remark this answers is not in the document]
kwzrd: The `all` function doesn't take `*args`, it takes a single iterable.

I find both of these noisy and ugly:
```python
return all((
  # conditions
))
```
```python
return (
  (
    # conditions
  )
)
```
Would you prefer either of these? I personally think `return all(conditions)` is really nice.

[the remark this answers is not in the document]
kwzrd: The reason why I went for `conditions = (...)` over just:
```python
return (
  a
  and b
  and c
)
```
Is that I find the `and`/`or`s to also get noise when there's so many of them. The possible short-circuit can be attractive if there are performance implications to consider, but I don't think that's necessarily the case here, so I went for the solution that I found the most beautiful.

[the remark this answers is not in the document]
ks129: When my solution is possible, then I think the current solution is best.

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,334 @@
...
+
+def own_reactions(message: discord.Message) -> t.Set[str]:
+    """Get the set of reactions placed on `message` by the bot itself."""
+    return {str(reaction.emoji) for reaction in message.reactions if reaction.me}
+
+
+def has_signals(message: discord.Message) -> bool:
+    """True if `message` already has all `Signal` reactions, False otherwise."""
+    missing_signals = ALLOWED_EMOJI - own_reactions(message)  # In `ALLOWED_EMOJI` but not in `own_reactions(message)`
+    return not missing_signals
```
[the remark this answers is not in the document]
kwzrd: > These lines should be merged together.

Why? While set operations are neat, they can be difficult to read. I found the intermediate step fairly helpful: both the assignment and `not missing_signals` are quite expressive.

I'd find this a lot more difficult to digest:
```python
return not ALLOWED_EMOJI - own_reactions(message)
```
I'm not saying you're wrong, I'm just wondering why you'd prefer this to be one-lined (especially if the comment won't fit).

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,334 @@
...
+    If the emoji has already been placed on `incident` by the bot, it will be skipped.
+    """
+    existing_reacts = own_reactions(incident)
+
+    for signal_emoji in Signal:
+
+        # This will not raise, but it is a superfluous API call that can be avoided
+        if signal_emoji.value in existing_reacts:
+            log.trace(f"Skipping emoji as it's already been placed: {signal_emoji}")
+
```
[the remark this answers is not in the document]
kwzrd: I'm sometimes quite generous with whitespace ~ I think this should be removed, yeah.

The `archive` method also has whitespace between the `try-except-else` blocks which can probably go. I'll look into it.

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
ks129: This should have a small comment about it. Then 1 line should better. But maybe is better to listen to someone else opinion too.

bot/cogs/moderation/incidents.py
kwzrd: `Webhook.send` will return `None` if unless explicitly `wait=True`. [See here](https://discordpy.readthedocs.io/en/latest/api.html#discord.Webhook.send).

The annotation is there because we have `wait=True`, so we can make a promise of what the type will be. I like these hints because then PyCharm tells me what methods there are, and what params they take. The `d.py` lib isn't annotated, so this cannot be inferred automatically.

bot/cogs/moderation/incidents.py
kwzrd: Because I want the delete listener running concurrently with the delete request.

We do:
1. Launch delete listener task
2. Delete message
3. Await listener finishing (maybe it has received the delete event already)

If we skip step 1. and only begin listening *after* dispatching the delete event, *I think* we risk missing the event (because `d.py` may receive it before we register the listener). For what it's worth, I was not able to replicate that - in the testing server, I always caught the event even if I started listening after the delete dispatch. But conceptually, this feels more correct - we had a brief chat with @MarkKoz about it, and while there just doesn't seem to be a good solution to this problem, this is probably the least bad we've got.

bot/cogs/moderation/incidents.py
ks129: Also, type hint should `Optional[discord.Message]` ad L268

bot/cogs/moderation/incidents.py
ks129: OK, now I understand. Just as a note that dpy have PR opened about adding type hinting.

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,334 @@
...
+        """
+        await self.bot.wait_until_guild_available()
+        incidents: discord.TextChannel = self.bot.get_channel(Channels.incidents)
+
+        # Limit the query at 50 as in practice, there should never be this many messages,
+        # and if there are, something has likely gone very wrong
+        limit = 50
+
+        # Seconds to sleep after adding reactions to a message
+        sleep = 2
```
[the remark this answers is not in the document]
kwzrd: Yeah, maybe, I'm not sure. I placed them here because I thought their purpose would be better understood when defined where they are used, especially since they are used in one place only, and because you don't really need to see these at the top of the module. In fact, the only reason why I defined them here rather than just passing them into `history` and `sleep` directly as literals is that I wanted to log them on line `L131` before starting the loop.

I'm not opposed to putting them at the top, but I'd prefer to hear at least one more opinion on it.

bot/cogs/moderation/incidents.py
kwzrd: I generally prefer checking for `None` explicitly. I want to return if `message` is not `None`, not if `message` is truthy. I don't feel like "Message can never be falsey, so we can get away with this" is a good enough justification - explicit is better than implicit, after all - but yeah, I suppose it's more of a personal preference. If the check was implicit, and I were reviewing this and didn't know whether a `Message` can ever be falsey, I'd have to go searching for it. This way, I don't.

Regarding the annotation, good catch, that is incorrect. I'll adjust it.

kwzrd: Annotation fixed in 8a08ca3.

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
kwzrd: Cleaned up the whitespace in b563063. It was needlessly spaced out considering the first clause only logs (that really doesn't *have* to be there, but I found myself appreciating it when testing).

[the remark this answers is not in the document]
kwzrd: Ended up not reducing the whitespace in `archive` as I feel the extra space makes it considerably more readable.

config-default.yml
MarkKoz: I think it's fine.

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
MarkKoz: I prefer chaining logical operator but I'm not too fussed.

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
MarkKoz: Isn't this the same as `<=` i.e. `ALLOWED_EMOJI` is a subset of `own_reactions`?

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
j03b: I don't think these are necessarily constants worthy, putting them at the top of the file seems fine IMO. Constants are useful for development when testing things like the offensive message remover (we don't want to sit waiting for a week to confirm it works) but I think there is little reason to change these and they are constant... constants. Putting them at the top of the file would be nicer though I think.

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,331 @@
...
+            return
+
+        if reaction not in ALLOWED_EMOJI:
+            log.debug(f"Removing invalid reaction: emoji {reaction} is not a valid signal")
+            await incident.remove_reaction(reaction, member)
+            return
+
+        # If we reach this point, we know that `emoji` is a `Signal` member
+        signal = Signal(reaction)
+        log.trace(f"Received signal: {signal}")
```
MarkKoz, line 225: What do you think about using a try-except instead? "Easier to ask for forgiveness" and whatnot.

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,331 @@
...
+            log.trace("Message was found in cache")
+            return message
+
+        log.trace("Message not found, attempting to fetch")
+        try:
+            message = await self.bot.get_channel(Channels.incidents).fetch_message(message_id)
+        except discord.NotFound:
+            log.trace("Message doesn't exist, it was likely already relayed")
+        except Exception as exc:
+            log.exception("Failed to fetch message!", exc_info=exc)
```
MarkKoz, line 277:
```suggestion
        except Exception:
            log.exception("Failed to fetch message!")
```

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
kwzrd: I made them module-level constants in 6d3a91c. I'm still not entirely convinced that it's better in any way (now they need more documentation), but I don't feel strongly about it.

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
kwzrd: @MarkKoz Using `issubset` is an excellent suggestion, thank you for that. Applied in b8ada89, it's beautiful!

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
kwzrd: I'm going to keep it as is then if no one feels too strongly. Feel free to reopen the conversation.

bot/cogs/moderation/incidents.py
kwzrd: @ks129 Please mark this as resolved if you're happy with the justification.

bot/cogs/moderation/incidents.py
kwzrd: See: 662ca58

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,331 @@
...
+                username=incident.author.name,
+                avatar_url=incident.author.avatar_url,
+                wait=True,  # This makes the method return the sent Message object
+            )
+
+            # Finally add the `outcome` emoji
+            await message.add_reaction(outcome.value)
+
+        except Exception as exc:
+            log.exception("Failed to archive incident to #incidents-archive", exc_info=exc)
```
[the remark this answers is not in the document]
kwzrd: See: 662ca58

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,331 @@
...
+
+        log.debug(f"Crawling messages in #incidents: {limit=}, {sleep=}")
+        async for message in incidents.history(limit=limit):
+
+            if not is_incident(message):
+                log.trace("Skipping message: not an incident")
+                continue
+
+            if has_signals(message):
+                log.trace("Skipping message: already has all signals")
```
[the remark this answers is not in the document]
kwzrd: Yes, you're right, these should log the id. I reviewed my logs and added ids where I felt it was appropriate in 20b27f3.

It's an interesting problem because if you have a stream of `trace`/`debug` logs sprinkled throughout a function, you don't want to be logging the ids in each message. So I usually just log this information in the first one, and then it's clear that "message" refers to the same object. But if they are followed by an >= `info` log, it should probably log the id again, because we usually have logs below that disabled in production. It's also fairly easy to forget to adjust other logs when changing just one.

Hopefully I got it more right this time, let me know what you think.

bot/cogs/moderation/incidents.py
kwzrd: I tried it and ended up being very happy with it. The idea crossed my mind at one point but then I ended up not doing it, because I originally had both the role and emoji check as a single conditional - they both remove the react and abort. Eventually though I felt that it'd be nice to actually log *why* it was removed, and I ended up making them disjoint.

With this change, the `ALLOWED_EMOJI` variable is only used by `has_signals` so the name no longer makes much sense, and so I ended up adjusting that as well. See: 98b8947.

bot/cogs/moderation/incidents.py
kwzrd: Fix from #1009 applied in a8b4e39 with f240a97 providing a simple test for it.

bot/cogs/moderation/incidents.py
ks129: @kwzrd This is OK. But only with write access or PR opener can mark reviews (even own reviews) resolved.

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
Numerlor: Perhaps class constants would be more fitting to give the names a bit more meaning without explaining it all

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,363 @@
...
+
+    username = f"{reported_by.name} | {actioned_by.name}"
+    log.trace(f"Generated webhook username: {username} (length: {len(username)})")
+
+    if len(username) > max_length:
+        stop = max_length - 3
+        username = f"{username[:stop]}..."
+        log.trace(f"Username capped at {max_length=}: {username}")
+
+    return username
```
kwzrd, line 65: So I added this to create the webhook username. It follows this format: `reported_by | actioned_by`. The format is easily adjustable so let me know if that doesn't work (using parens is an alternative).

image

Maximum webhook username length is 80 (request fails if above), so this will cut off anything above:

image

The thing is that both names and display names for regular users are capped at 32 so we should never actually exceed 80, but I did not realize that in time and I guess it doesn't hurt if the logic stays there in case something changes in the future. To trigger the cut-off for the screenshot I manually passed `max_len=40`.

Didn't use `textwrap.shorten` because I didn't want it to respect words / trim whitespace, I just wanted to cap the raw character count, which I believe is easier done manually.

How the `actioned_by` name is relayed was never really agreed upon with others, I failed to get a discussion started twice now, so I decided to just implement it and see how it works for everyone.

j03b:
I'm not sure how I feel about using the username to convey this stuff.

Could we ditch the webhook approach and post with the bot like so:
image

kwzrd:
@j03b Alright, let's use embeds ~ changed back to draft for now until I implement the necessary changes.

kwzrd:
@j03b Incidents are now archived as embeds.

image

See: dd74105, 744aed5. Image-like attachments are also supported: 83544ca.

PR description & tests adjusted as necessary, this should now be ready for review again.

bot/cogs/moderation/incidents.py
kwzrd: Collapsing this as we scrapped this.

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # If the incident had an attachment, we will try to relay it
+        if incident.attachments:
+            attachment = incident.attachments[0]  # User-sent messages can only contain one attachment
+            log.debug(f"Attempting to archive incident attachment: {attachment.filename}")
+
+            attachment_file = await attachment.to_file()  # The file will be sent with the webhook
+            embed.set_image(url=f"attachment://{attachment.filename}")  # Embed displays the attached file
+        else:
+            attachment_file = None
```
kwzrd, line 215: Not clean, but as clean as I could get it. There's something oddly difficult here because we make the embed in `make_embed`, but there also has to be some branching here to determine whether a file should be passed to the `send`.

So if `make_embed` also looks at `incident.attachments` to determine whether `set_image` should be called, and this looks at `incident.attachments` to see whether a file should be passed, we're checking the same thing in two places and it feels ugly & fragile since they could easily go out of sync should one make adjustments to one method or the other.

Using this approach, we can use a single `if-else` to handle both the file and the url, and it feels ok, but it comes at the cost of still mutating the embed instance outside of `make_embed`.

kwzrd:
@Senjan21 
> ... if moderator actions an incident while the bot is offline, bot won't take any actions (archive) on the message ...

The cog is designed to react to events just like any other listener the bot may have, and so it's limited in the same ways: you cannot distribute infractions while the bot is offline, you cannot schedule reminders, the paginator stops working, etc. You don't normally expect the bot to catch up with paginator reaction clicks after it connects; neither does it search for missed command invocations.

Reacting to events which happened while the bot wasn't listening is difficult. It will encounter situations for which we haven't defined the desired behaviour. It can encounter conflicting instructions (for which we don't know which one came first), and it may be unexpected from the user's perspective. How it should work (both UX and internal implementation) was never discussed nor mentioned as part of the issue which this PR solves.

I see it as something well beyond the scope of this PR. If you believe that such functionality is necessary, for this cog in particular, I propose that a separate issue is written, which discusses exactly how it should behave, and once we agree on a solution, a separate PR makes the necessary adjustments. I do not feel like it's fair to block this PR requesting a fairly involved adjustment / rewrite to solve a problem which is seen as acceptable in any other feature of the bot, i.e. having to wait the few secs for it to reconnect *should* it just happen to be unavailable.

As I explained earlier, the purpose of the crawl task is to always ensure that each incident sent in `#incidents` has the 3 Signal reactions added by the bot, to ensure that they can be clicked easily by staff members. It's purpose is *not* to fully sync the channel state: neither the PR description, nor the documentation, make such a claim.

ks129:
Maybe should bot notice in #incidents channel that currently, this is not working (before shutdown)/that this is working again (on startup)?

Senjan21:
In such case, I think I can approve the changes and see how it works in real life scenarios

> If you believe that such functionality is necessary, for this cog in particular, I propose that a separate issue is written.

After thinking about it, I have to agree that it might not be a necessity, I suppose if any moderator will in fact want such functionality, they can make the issue.

kwzrd:
@ks129 
> Maybe should bot notice in #incidents channel that currently, this is not working (before shutdown)/that this is working again (on startup)?

I don't feel like there's anything inherently special about this cog that would require such a thing. The bot does not normally announce that it's going to restart, even if it makes much more critical features unavailable for the time being. In `#mod-log`, we can see new commits on master, which indicate a restart, and we can also see when the bot connects. Staff are familiar with this.

This PR only allows incident archiving. It's not a critical feature that must be available at all times. If you click the reaction and the bot is offline, the only "problem" is that you have to click it again once it connects.

bot/cogs/moderation/incidents.py
```diff
@@ -0,0 +1,381 @@
...
+    All other signals are seen as invalid.
+    """
+
+    ACTIONED = Emojis.incident_actioned
+    NOT_ACTIONED = Emojis.incident_unactioned
+    INVESTIGATING = Emojis.incident_investigating
+
+
+# Reactions from roles not listed here will be removed
+ALLOWED_ROLES: t.Set[int] = {Roles.moderators, Roles.admins, Roles.owners}
```
vivax3794, line 39: could we not use `MODERATION_ROLES` from `constants.py` here instead of listing the roles manually?

kwzrd: Yeah, I think that'd make sense.

tests/bot/cogs/moderation/test_incidents.py
```diff
@@ -312,39 +312,29 @@ async def test_archive_relays_incident(self):
...
+        and that the `archive` method returns True.
         """
-        webhook_message = MockMessage()  # The message that will be returned by the webhook's `send` method
-        webhook = MockAsyncWebhook(send=AsyncMock(return_value=webhook_message))
-
+        webhook = MockAsyncWebhook()
         self.cog_instance.bot.fetch_webhook = AsyncMock(return_value=webhook)  # Patch in our webhook
 
-        # Now we'll pas our own `incident` to `archive` and capture the return value
+        # Define our own `incident` for archivation
```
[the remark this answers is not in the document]
kwzrd: It really is not, but at the time I was struggling to express it in a valid way. I suppose "to be archived" will do.

kwzrd:
@MarkKoz 
> [re: only 1 user-sent attachment] This is true on Desktop, but not on mobile.

Thanks for mentioning this, I didn't think to test on mobile.

> The webhook doesn't use the nickname. Should it?

Good question, I don't know. I think I went with name because we don't otherwise store the incident author identity (only avatar and name at the time or archivation<sup>sorry</sup>), and I figured if someone changes their nickname to something memey temporarily it may end up being difficult to recall who it was after some time. But I don't mind using nickname at all if you or anyone else prefers it, I basically don't have any opinion on it.

> The "actioned"/"rejected" in the footer is a bit too subtle. Can the emoji be fit anywhere in the embed to make it clearer? Maybe it's not too bad since the embeds also change colour.

I think the embed colour will be the primary mean of distinction for most. I see the footer text as more of a detail, it is subtle yes. Admittedly I did not consider at all whether colour-blindness may make it difficult to distinguish them. The easiest way to sneak in the emoji would be to use the footer image field, although that's currently used by the `actioned_by` avatar and I think that looks fairly good. I'll think about it.

bot/cogs/moderation/incidents.py
```diff
@@ -186,22 +186,41 @@ async def archive(self, incident: discord.Message, outcome: Signal, actioned_by:
...
         """
         log.debug(f"Archiving incident: {incident.id} (outcome: {outcome}, actioned by: {actioned_by})")
+        embed = make_embed(incident, outcome, actioned_by)
+
+        # If the incident had an attachment, we will try to relay it
+        if incident.attachments:
+            attachment = incident.attachments[0]  # User-sent messages can only contain one attachment
+            log.debug(f"Attempting to archive incident attachment: {attachment.filename}")
+
+            attachment_file = await attachment.to_file()  # The file will be sent with the webhook
```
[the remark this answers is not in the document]
kwzrd: Good point, this should be addressed. But I'm already anxious thinking about how to write this so that it's not too ugly. I'll ponder upon it.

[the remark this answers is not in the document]
kwzrd: Actually I wonder if this should abort entirely? If this fails for some reason with the attachment being valid & important but we still relay, we will lose potentially important information.

bot/cogs/moderation/incidents.py
kwzrd: Suggestion applied: 918e1b9

tests/bot/cogs/moderation/test_incidents.py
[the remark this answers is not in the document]
kwzrd: See: ddb1f55

config-default.yml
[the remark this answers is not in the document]
kwzrd: Kaizened in 54891b4, also indented the `&PYNEWS_WEBHOOK` further since I was already making changes.

The config files (both the yaml and the `.py`) could use some more refactoring, I noticed that the spacing between classes is inconsistent (1 or 2 empty lines), and some other constants are probably not sorted properly. But I consider those beyond the scope of this PR.

config-default.yml
```diff
@@ -173,6 +177,7 @@ guild:
         organisation:       &ORGANISATION   551789653284356126
         staff_lounge:       &STAFF_LOUNGE   464905259261755392
         incidents:                          714214212200562749
+        incidents_archive:                  720668923636351037
```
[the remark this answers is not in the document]
kwzrd: Sure, that's my bad: 54891b4

kwzrd:
@MarkKoz I've tried a few things to sneak in the Signal emoji, this is an option:

image

The author field requires a name but it'll take `\U0000200B` (zero-width character). We need to resolve the emoji cdn url somehow.

A prettier option would be using the footer icon, but I enjoy having the moderator avatar there. The footer otherwise won't display the emoji (with `<:emojiname:123>` in the text). However I *think* the embed colour is probably sufficient.

j03b:
I'm not sure we need the icon — colour gives away enough imo.

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
MarkKoz: You can fall back to linking the attachment or add a note that there was supposed to be one but it failed.

bot/cogs/moderation/incidents.py
MarkKoz: This is perfectly fine in my opinion.

MarkKoz:
> But I don't mind using nickname at all if you or anyone else prefers it, I basically don't have any opinion on it.

Neither do I. I'm fine with how it is.

> A prettier option would be using the footer icon:

I like it in the footer, but I'm willing to try out the colour only. I will complain later if I find it too subtle in practice 😄

bot/cogs/moderation/incidents.py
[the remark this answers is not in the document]
kwzrd: How would you feel about using the author name field?

image

It can be made to link the attachment proxy url, but I think that stops being functional some time after the original message is deleted.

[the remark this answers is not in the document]
MarkKoz: My first though would be to put it at the end of the description so that the message content "takes priority". However, I am fine with your proposal as well.

> I do log the exception so we'll get a Sentry notif if this happens.

I don't think `NotFound` and any 500's are worth being notified for, though we should probably consider ignoring all 500s at a larger scale rather than here.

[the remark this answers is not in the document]
kwzrd: I preferred the name field because (although it's really not a name) it prevents having to worry about the embed description length, plus I think it looks ok.

I'm still a little bit on the edge though, I feel like it may be more logical to just return if the download fails. I'm considering moving the logic elsewhere, the `archive` function is now branching too much and I'm struggling to adjust the tests. :angry:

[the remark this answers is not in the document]
MarkKoz: >  I feel like it may be more logical to just return if the download fails. 

Yes, if the message has only attachments and no text content. Otherwise, why?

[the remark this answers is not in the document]
kwzrd: Take a peek at c4e9060 and 6e48d66.

I added a method to download the file safely, but it currently error-level logs all exceptions. Additional clauses can be put there to reduce the log level of specific exceptions (maybe the 404? but in practice it really shouldn't happen unless someone deleted the msg manually as it's being relayed).

The latter commit then moves the attachment-handling logic to `make_embed`, I decided to handle it all there and return the file if available. Seems more logical in the end. Keeps `archive` simple.

Let me know how you feel about this.

bot/cogs/moderation/incidents.py
```diff
@@ -42,6 +42,19 @@ class Signal(Enum):
...
+    """
+    Download & return `attachment` file.
+
+    If the download fails, the reason is logged and None will be returned.
+    """
+    log.debug(f"Attempting to download attachment: {attachment.filename}")
+    try:
+        return await attachment.to_file()
+    except Exception:
+        log.exception("Failed to download attachment")
```
MarkKoz, line 59: `NotFound` shouldn't have a high level.

kwzrd: Reduced to `debug` in 6f5fb20, though I still don't think I fully understand why a 404 is in this case different to any other error.

Here's the way I see it: when fetching messages, a 404 is expected in the case of a racing coroutine trying to get its hands on an already relayed message. It signals a certain situation that must be handled differently. An error notification isn't desired, because a) the situation is expected, and b) the behaviour for handling it is defined.

When downloading the attachment, however, I don't see a 404 as something that we should ignore. I cannot think of a way where this fails on a 404 and we don't want to know. I can trigger the 404 by putting a sleep before the download, and manually deleting the original message before `to_file` is called. But if this somehow gives a 404 in production, I feel like we should know?

Either way, it's not too important, just trying to explain my thought process so that you can correct me if I'm making an error somewhere in my thinking.

MarkKoz: If you knew there's a 404, then what exactly would you do with that information? You can't bring back a deleted file. We shouldn't have high logging levels for errors which aren't actionable. I assumed a 404 was expected if a message was manually deleted. Turns out it's a 403, but the same point still applies. You could argue that it'd be better to detect the message delete and abort before trying to get the file, but this may end up being too over-engineered compared to simply ignoring an exception. Even if it's a 403, a 404 is still worthwhile to ignore since, again, it's unactionable. It's just like all the 500s we ignore in Sentry. If a file is gone, then it's gone.

kwzrd: > If you knew there's a 404, then what exactly would you do with that information? You can't bring back a deleted file.

Yes, but it just wasn't clear to me how a 404 is different from any other exception, because the original message gets deleted regardless. So if it's a 404 we lose the file but don't get an alert, if it's a 500 we lose the file but get an alert.

So do you think we need to be ready for the situation where someone actions an incident and deletes it at the same time? I guess I'll make it not log the exception if it's a 403 then too.

MarkKoz: Other exceptions may be the result of buggy code. But a 404 comes from a deleting a message, which is a normal user action.

MarkKoz: > So do you think we need to be ready for the situation where someone actions an incident and deletes it at the same time?

Not really. Things won't break catastrophically will they? In the case it does happen, I would just rather not be pestered by the 404/403 exception.

kwzrd: > Things won't break catastrophically will they?

Depends, it's unfortunate. The thing is that the relay is pretty fast, unless it's downloading a massive file. You'd have to *try* to break it, it's extremely unlikely that it'd happen by accident. But if it does, then:

* Attachment fails to download, we fallback on `proxy_url`
* Let's delete, `confirmation_task` is scheduled
* The call to `message.delete` fails with an unhandled 404
* The `confirmation_task` is already running but isn't awaited in the try block, so the timeout is unhandled after 5 secs

So if someone deletes the message we'll get alerts anyway.

kwzrd: I haven't though about it much, but I suppose the way to handle it would be to catch exceptions from the delete call and if we get something, then cancel `confirmation_task`? But it still feels fragile. I don't think we normally write code to be resilient to intentional malice. For what this cog does, it already feels over-engineered.

MarkKoz: Well, if you want to account for it, feel free. But I'm fine with revisiting this later if we actually do encounter a deleted message in production.

kwzrd: I don't think we'll encounter it personally. I volunteer to handle that Sentry ticket if it ever comes through. :smile: 

403 logs at `debug` level too now: 476d407

PR #1000, opened 2020-06-15
ModLog: fix AttributeError in on_member_update

MarkKoz:
Sentry Issue: [BOT-4R](https://sentry.io/organizations/python-discord/issues/1668251702/?referrer=github_integration)

```
AttributeError: 'Role' object has no attribute 'get'
    await coro(*args, **kwargs)
  File "bot/cogs/moderation/modlog.py", line 511, in on_member_update
    new = value.get("new_value")

Unhandled exception in on_member_update.
```

The `iterable_item_removed` and `iterable_item_added` diffs only contain the actual value added or removed (e.g. a `discord.Role` in the above error) rather than having a `new_value` and an `old_value`. The code was incorrectly trying to access old and new values for the iterable changes. I've completely removed the need for these iterable diffs. The commit message goes into slightly more detail.

I also removed code that checked diffs for `discord.User` attributes. The `on_member_update` event won't be dispatched for those attributes (I got confirmation the old code didn't work by examining old logs). Instead, `on_user_update` should be used. Should this event be implemented, or are logs for the user, discriminator, and avatar not needed?


MarkKoz:
In that case, I'll base a new branch off this one and refactor the cog along with adding `on_user_update`. There's currently a lot of redundant code. I'll leave this PR alone cause it doesn't remove any functionality we didn't already have.

bot/cogs/moderation/modlog.py
```diff
@@ -462,74 +476,27 @@ async def on_member_update(self, before: discord.Member, after: discord.Member)
...
-        for key, value in diff_values.items():
-            if not key:  # Not sure why, but it happens
-                continue
-
-            key = key[5:]  # Remove "root." prefix
-
-            if "[" in key:
-                key = key.split("[", 1)[0]
+        # The regex is a simple way to exclude all sequence and mapping types.
+        diff = DeepDiff(before, after, exclude_regex_paths=r".*\[.*")
```
Numerlor, line 482: I was going to suggest ignoring ABCs here, but it is is only for wherever deepdiff adds the brackets into the paths (didn't find that much in docs on the cases where it uses those). 
I'm not sure if there is a fitting abc and this seems to work nicely for the intended purposes, but the comment was a bit misleading since we want strings here which are also sequences.

PR #1001, opened 2020-06-15
Check infraction reason isn't None before shortening it

MarkKoz:
With the recent changes to truncate reasons, infractions will fail to apply if a reason isn't provided. The fix is to only shorten the reason if it isn't None.

### DIFF
--- a/bot/cogs/moderation/infractions.py
+++ b/bot/cogs/moderation/infractions.py
@@ -75,9 +75,7 @@ class Infractions(InfractionScheduler, commands.Cog):
     # region: Temporary infractions
 
     @command(aliases=["mute"])
-    async def tempmute(
-            self, ctx: Context, user: Member, duration: Expiry, *, reason: t.Optional[str] = None
-    ) -> None:
+    async def tempmute(self, ctx: Context, user: Member, duration: Expiry, *, reason: t.Optional[str] = None) -> None:
         """
         Temporarily mute a user for the given reason and duration.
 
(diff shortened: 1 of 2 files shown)


MarkKoz:
@kosayoda IMO, the optional type annotations are mostly unnecessary and hurt readability. The reason already has a default value of `None`, which should be enough to clue readers in that it's optional. I already have a commit I can push with the annotations removed.

SebastiaanZ:
I like type hints with the explicit `Optional` type hint.

MarkKoz:
OK. Should there be extra indentation for the parameters? PEP 8 suggests this because it argues that they'd otherwise not be distinguishable from the function body. However, they don't put the closing parenthesis and colon on a separate line, which I argue does adequately distinguish parameters from the body along with the mandatory docstring.

SebastiaanZ:
I'm honestly not married on either (the `Optional` type hint) and/or the indentation we choose, but my personal preference is to use the `Optional`, but not add additional indentation.

If you feel like it greatly increases readability by leaving the `Optional` out, I'm fine by that as well; what's the general style throughout the bot? I personally probably always add the `Optional` explicitly.

MarkKoz:
I generally agree that it should be explicit. But I think it's also okay to break consistency to improve readability. Ultimately, these type annotations are supposed to help us, not hurt us. I honestly find most parameter splitting to be ugly, like in this case, so I'd rather avoid it. I also find long function signatures ugly, but what can we do 🤷 .

I'll leave the explicit type hints but remove the extra indentation since it's more consistent with other code. We should consider updating our [style guide](https://pythondiscord.com/pages/contributing/style-guide/#bracket-and-item-arrangement) if this is the consensus.

PR #1002, opened 2020-06-15
Sync: ignore events from other guilds

MarkKoz:
Sentry Issue: [BOT-50](https://sentry.io/organizations/python-discord/issues/1682012885/?referrer=github_integration)

There's no reason to sync users from other guilds like the emoji guilds. It was causing errors because it tried to sync roles specific to those guilds which did not exist in the database.

Showing the 404 warning is debatable (see #779), but I decided to just keep it and add a parameter to ignore it. It's a bit hacky but it will do until the aforementioned issue is resolved.


bot/cogs/sync/cog.py
```diff
@@ -34,18 +34,22 @@ async def sync_guild(self) -> None:
...
             if e.response.status != 404:
                 raise
-            log.warning("Unable to update user, got 404. Assuming race condition from join event.")
+            if not ignore_404:
+                log.warning("Unable to update user, got 404. Assuming race condition from join event.")
 
     @Cog.listener()
     async def on_guild_role_create(self, role: Role) -> None:
         """Adds newly create role to the database table over the API."""
+        if role.guild != constants.Guild.id:
```
ks129, line 50: This is not working. This have to be `role.guild.id != constants.Guild.id`. The current statement is always `True`.

tests/bot/cogs/sync/test_cog.py
```diff
@@ -142,20 +148,32 @@ async def test_sync_cog_on_guild_role_create(self):
             "permissions": 8,
             "position": 23,
         }
-        role = helpers.MockRole(**role_data)
+        role = helpers.MockRole(**role_data, guild=self.guild_id)
```
ks129, line 151: You have to pass `MockGuild` here and adding this ID after fixing cog L50's issue.

bot/cogs/sync/cog.py
MarkKoz: Good catch. Can't rely on the tests when I write the tests incorrectly too 😄

tests/bot/cogs/sync/test_cog.py
```diff
@@ -272,12 +311,15 @@ async def test_sync_cog_on_user_update(self):
 
                     # Don't care if *all* keys are present; only the changed one is required
                     call_args = self.cog.patch_user.call_args
-                    self.assertEqual(call_args[0][0], after_user.id)
-                    self.assertIn("updated_information", call_args[1])
+                    self.assertEqual(call_args.args[0], after_user.id)
+                    self.assertIn("json", call_args.kwargs)
+
+                    self.assertIn("ignore_404", call_args.kwargs)
+                    self.assertTrue(call_args.kwargs["ignore_404"])
```
kwzrd, line 318: I really like the idea of checking membership first. Makes for a much nicer error should the test fail on it not being there, I should do this too.

PR #1009, opened 2020-06-16
Fix 400 when "clyde" is in any webhook username

MarkKoz:
Discord just disallows this name.

The replacement character is [Mathematical Sans-Serif Small E](https://www.compart.com/en/unicode/U+1D5BE), which I felt was the best choice due to its similarity to a normal "e". However, I'm open to changing it to another character, putting it in parenthesis, or whatever else.


bot/utils/messages.py
```diff
@@ -109,8 +110,19 @@ async def send_attachments(
...
                 embed=embed,
-                username=message.author.display_name,
+                username=sub_clyde(message.author.display_name),
                 avatar_url=message.author.avatar_url
             )
 
     return urls
+
+
+def sub_clyde(username: Optional[str]) -> Optional[str]:
```
ks129, line 120: In what case `username` can be `None`? So much how I see, this should pass `str` every time.

MarkKoz: It's optional here though every use of that function passes a name.

https://github.com/python-discord/bot/blob/bcf6993de7de726683e6ca9b0f102b6ad1a732fa/bot/cogs/watchchannels/watchchannel.py#L199-L207

I don't see a problem with leaving it optional anyway.

MarkKoz:
> I believe the cyrillic е would be better here and seems to look the same as a normal e

Good find. The website I used didn't show Cyrillic characters as characters based on "e" (probably only considered latin alphabet). Cyrillic also has an upper case E so I will use that too. I wonder if there's a good way to do a case sensitive substitution. May just be easiest to chain two substitutions.

> Do we also need to sub the names in reddit/python_news? The consistency is nice but having clyde in those is extremely unlikely

It's only one extra line of code, so better safe than sorry. I'd say this should be used for any name that's coming from an external source.

bot/utils/messages.py
kwzrd: If this internally handles the `None` (and it does) then I agree the annotation is appropriate, but why does it handle it? Is it to be "shielded" from callers which may accidentally give a `None`? Won't that just push the error further, since something else that is expecting a string will get the declyded `None`?

Numerlor: None is a valid value for webhook username (acts like an empty string and defaults to the webhook name) that may get passed in so I don't think there's any problem with letting None through.

kwzrd: In that case I agree that this is perfectly fine.

bot/utils/messages.py
```diff
@@ -109,8 +110,23 @@ async def send_attachments(
...
+
+    Discord disallows "clyde" anywhere in the username for webhooks. It will return a 400.
+    Return None only if `username` is None.
+    """
+    def replace_e(match: re.Match) -> str:
+        char = "е" if match[2] == "e" else "Е"
+        return match[1] + char
+
+    if username:
+        return re.sub(r"(clyd)(e)", replace_e, username, flags=re.I)
```
kwzrd, line 132: This will also return `None` if an empty string is given, which I suppose is fine if they are interchangeable in the webhook username field, but the docstring denies that. Explicitly checking for `None` on L131 seems like a better idea than adjusting the docstring though.

PR #1016, opened 2020-06-20
Some bug fixes

### LINKED ISSUE #1006: Catch mute failures if member leaves the guild
Sentry Issue: [BOT-5F](https://sentry.io/organizations/python-discord/issues/1688869272/?referrer=github_integration)

```
NotFound: 404 Not Found (error code: 10007): Unknown Member
  File "bot/cogs/moderation/scheduler.py", line 151, in apply_infraction
    await action_coro
  File "bot/cogs/moderation/infractions.py", line 212, in action
    await user.add_roles(self._muted_role, reason=reason)
  File "discord/member.py", line 641, in add_roles
    await req(guild_id, user_id, role.id, reason=reason)
  File "discord/http.py", line 223, in request
    raise NotFound(r, data)

Failed to apply mute infraction <REDACTED>
```

### LINKED ISSUE #1008: Task result is accessed too early when unloading watch channel cogs
Sentry Issue: [BOT-7C](https://sentry.io/organizations/python-discord/issues/1800298412/?referrer=github_integration)

```
InvalidStateError: Result is not set.
  File "discord/ext/commands/bot.py", line 726, in reload_extension
    self._remove_module_references(lib.__name__)
  File "discord/ext/commands/bot.py", line 554, in _remove_module_references
    self.remove_cog(cogname)
  File "discord/ext/commands/bot.py", line 540, in remove_cog
    cog._eject(self)
  File "discord/ext/commands/cog.py", line 414, in _eject
    self.cog_unload()
  File "bot/cogs/watchchannels/watchchannel.py", line 333, in cog_unload
    self._consume_task.result()

ClientException: Command talentpool is already registered.
(3 additional frame(s) were not displayed)
...
  File "bot/bot.py", line 49, in add_cog
    super().add_cog(cog)
  File "discord/ext/commands/bot.py", line 502, in add_cog
    cog = cog._inject(self)
  File "discord/ext/commands/cog.py", line 379, in _inject
    raise e
  File "discord/ext/commands/cog.py", line 374, in _inject
    bot.add_command(command)
  File "discord/ext/commands/core.py", line 1071, in add_command
    raise discord.ClientException('Command {0.name} is already registered.'.format(command))

Extension 'bot.cogs.watchchannels' failed to reload.
```

When reloading, the cog failed to be unloaded and its commands weren't unregistered. Discord.py tries to revert to the previous state before raising the exception, but it failed to load the cog since it couldn't register the same commands twice.

The reason the unload failed is because `Task.cancel()` doesn't guarantee _when_ the task will be cancelled. Trying to get the task's result too early, as in this case, will raise a `InvalidStateError`.

I think the proper solution is to cancel the task and add a done callback which will retrieve the result. In fact, it seems more appropriate to retrieve the exception rather than the result. This is what the scheduler does in its done callback.

### LINKED ISSUE #1011: Ensure reddit's `revoke_access_token` completes before the aiohttp session closes
Sentry Issue: [BOT-37](https://sentry.io/organizations/python-discord/issues/1608111595/?referrer=github_integration)

```
RuntimeError: Session is closed
  File "bot/cogs/reddit.py", line 106, in revoke_access_token
    response = await self.bot.http_session.post(
  File "aiohttp/client.py", line 357, in _request
    raise RuntimeError('Session is closed')

Task exception was never retrieved
future: <Task finished name='Task-7817977' coro=<Reddit.revoke_access_token() done, defined at /bot/bot/cogs/reddit.py:100> exception=RuntimeError('Session is closed')>
```

`revoke_access_token` is scheduled in `cog_unload`. It's not a coroutine so it cannot block the shutdown process. Therefore, discord.py proceeds and eventually closes the session before `revoke_access_token` finishes.

### LINKED ISSUE #1012: Ensure tasks finish/cancel before the bot shuts down
Sentry Issue: [BOT-54](https://sentry.io/organizations/python-discord/issues/1683129089/?referrer=github_integration)

```
RuntimeError: Session is closed
  File "discord/http.py", line 165, in request
    async with self.__session.request(method, url, **kwargs) as r:
  File "aiohttp/client.py", line 1012, in __aenter__
    self._resp = await self._coro
  File "aiohttp/client.py", line 357, in _request
    raise RuntimeError('Session is closed')

Task exception was never retrieved
future: <Task finished name='Task-1762938' coro=<Messageable.send() done, defined at /usr/local/lib/python3.8/site-packages/discord/abc.py:771> exception=RuntimeError('Session is closed')>
```

It's unclear where this task was scheduled, but it should be cancelled before the bot shuts down and the session closes. May be a similar situation to #1011.

ks129:
- b9d483c: Moved exception logging when cog is being unloaded and messages are still
not consumed from `cog_unload` to `consume_messages` itself in
try-except block to avoid the case when requesting result too early
(before cancel finished). Resolves #1008 
- ac302d3: When user left from guild before bot can add Muted role, then catch this
error and log. Resolves #1006. Also found the issue that when the user is muted and joins and quickly leave, then this fail too (tested in dev server).
- 429cc86: Made to resolve problem with Reddit cog that revoking access token raise
exception because session is closed. To solve this, I made
`Bot.closing_tasks` that bot wait before closing.
Moved all extensions and cogs removing to `remove_extension` what is
called before closing everything else because need to call `cog_unload`. Resolves #1011 **Might need more discussion about implemention**.
- f4004d8: Bot wants to send an alert when channel stay silenced when the bot is unloading cog. Problem was that when the bot was shutting down, the session closed before task had a chance to send the message. Added this task to `closing_tasks` to finish it before shutdown. Resolves #1012 .


bot/cogs/moderation/infractions.py
```diff
@@ -223,10 +223,13 @@ async def apply_mute(self, ctx: Context, user: Member, reason: t.Optional[str],
...
-
-            log.trace(f"Attempting to kick {user} from voice because they've been muted.")
-            await user.move_to(None, reason=reason)
+            try:
+                await user.add_roles(self._muted_role, reason=reason)
+            except discord.NotFound:
+                log.info(f"User {user} ({user.id}) left from guild. Can't give Muted role.")
+            else:
+                log.trace(f"Attempting to kick {user} from voice because they've been muted.")
+                await user.move_to(None, reason=reason)
```
MarkKoz, line 232: According to the documentation, this doesn't raise `NotFound`. However, surely it would still fail if the member left the guild. You need to figure out what _exactly_ it raises and only catch that. You may need to look at `HTTPException.code` if the status is not enough of an indicator.

bot/cogs/watchchannels/watchchannel.py
```diff
@@ -330,10 +336,3 @@ def cog_unload(self) -> None:
         self.log.trace("Unloading the cog")
         if self._consume_task and not self._consume_task.done():
             self._consume_task.cancel()
-            try:
-                self._consume_task.result()
```
MarkKoz, line 334: I don't agree with this. You've completely removed the fetching of the task result. The done callback approach I suggested in the issue is better because it allows for exceptions to be retrieved in addition to checking if the task was cancelled.

bot/bot.py
```diff
@@ -89,9 +92,32 @@ def clear(self) -> None:
         self._recreate()
         super().clear()
 
+    def remove_extensions(self) -> None:
```
MarkKoz, line 95: This should be named as a private method.

bot/bot.py
```diff
@@ -89,9 +92,32 @@ def clear(self) -> None:
...
+        """Remove all extensions and Cog to close bot. Copy from discord.py's own `close` for right closing order."""
```
MarkKoz, line 96:
```suggestion
        """Unload and remove all extensions and Cogs."""
```

bot/bot.py
```diff
@@ -89,9 +92,32 @@ def clear(self) -> None:
...
     async def close(self) -> None:
         """Close the Discord connection and the aiohttp session, connector, statsd client, and resolver."""
-        await super().close()
+        # Remove extensions and cogs before calling super().close() to allow task finish before HTTP session close
+        self.remove_extensions()
+
+        # Wait until all tasks that have to be completed before bot is closing is done
+        for task in self.closing_tasks:
+            log.trace(f"Waiting for task {task.get_name()} before closing.")
+            await task
```
MarkKoz, line 117: Use `asyncio.gather`. No need to wait sequentially.

bot/bot.py
```diff
@@ -89,9 +92,32 @@ def clear(self) -> None:
...
+            try:
+                self.unload_extension(extension)
+            except Exception:
+                pass
+
+        for cog in tuple(self.cogs):
+            try:
+                self.remove_cog(cog)
+            except Exception:
+                pass
```
MarkKoz, line 107: This is _almost_ an exact copy. I'm concerned that if you're copying their code, you must adhere to their license. This would mean including their license and copyright. Here's an example of how I did it for something unrelated in another PR https://github.com/python-discord/bot/blob/de592dc5eb22d061c9b988844e8c7d695a37fa58/LICENSE-THIRD-PARTY

bot/bot.py
```diff
@@ -89,9 +92,32 @@ def clear(self) -> None:
...
+
     async def close(self) -> None:
         """Close the Discord connection and the aiohttp session, connector, statsd client, and resolver."""
-        await super().close()
+        # Remove extensions and cogs before calling super().close() to allow task finish before HTTP session close
```
MarkKoz, line 111: Let's make this more concise. Bonus: it will fit within 100 characters

```suggestion
        # Done before super().close() to allow tasks finish before the HTTP session closes.
```

ks129:
I think manually reloading cogs isn't big problem, because bot still runs and tasks too.

bot/bot.py
```diff
@@ -92,8 +92,8 @@ def clear(self) -> None:
         self._recreate()
         super().clear()
 
-    def remove_extensions(self) -> None:
-        """Remove all extensions and Cog to close bot. Copy from discord.py's own `close` for right closing order."""
+    def _remove_extensions(self) -> None:
+        """Remove all extensions and Cog to close bot. Copyright (c) 2015-2020 Rapptz (discord.py, MIT License)."""
```
MarkKoz, line 96: This does not comply with the license. You need to include the copyright AND the permission notice. So basically, copy pasting the entire license. Again, look at the example I linked.

ks129: I found that this don't actually need remove cogs manually: they get removed automatically by unloading extensions.

MarkKoz: Yes, but cogs aren't restricted to being defined in extensions. Even though (I think) they always are in our bot, the previous code was the proper approach.

bot/bot.py
MarkKoz: You forgot to rename the method call done in `close()`.

bot/bot.py
ks129: This is removing other cogs that is not in extensions later.

MarkKoz: Where does it remove those other cogs? This code was responsible for it, but you removed it

```py
for cog in tuple(self.cogs):
    try:
        self.remove_cog(cog)
    except Exception:
        pass```

Comment (ks129) [reply bot/bot.py]: Now `await super().close()` remove other cogs too.

**Comment by ks129** [comment]
@MarkKoz All feedback addressed

File: bot/bot.py [hunk 9]
```diff
@@ -131,8 +134,26 @@ def clear(self) -> None:
...
+                pass
+
     async def close(self) -> None:
         """Close the Discord connection and the aiohttp session, connector, statsd client, and resolver."""
+        # Done before super().close() to allow tasks finish before the HTTP session closes.
+        self._remove_extensions()
+
+        # Wait until all tasks that have to be completed before bot is closing is done
+        log.trace("Waiting for tasks before closing.")
+        await asyncio.gather(*self.closing_tasks)
```
MarkKoz, line 154: While this approach does seem feasible for solving issues when shutting down, I'm concerned about what will happen when cogs are manually reloaded. For example, could there be a race condition between a closing task and some init task?

MarkKoz (changes requested):

bot/bot.py
ks129: There isn't currently any closing task that really can have dangerous race conditions. This work same way for reloading like this worked before, as this now just add them to list.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -337,10 +358,17 @@ async def deactivate_infraction(
...
+        except discord.NotFound:
+            log.info(f"Can't pardon {infraction['type']} for user {infraction['user']} because user left from guild.")
         except discord.HTTPException as e:
-            log.exception(f"Failed to deactivate infraction #{id_} ({type_})")
-            log_text["Failure"] = f"HTTPException with status {e.status} and code {e.code}."
-            log_content = mod_role.mention
+            if e.code == 10007:
+                log.info(
+                    f"Can't pardon {infraction['type']} for user {infraction['user']} because user left from guild."
+                )
```
MarkKoz, line 367: Since the message is the same, these can be merged by doing `or e.status == 403`.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -337,10 +358,17 @@ async def deactivate_infraction(
...
             log_content = mod_role.mention
+        except discord.NotFound:
+            log.info(f"Can't pardon {infraction['type']} for user {infraction['user']} because user left from guild.")
         except discord.HTTPException as e:
-            log.exception(f"Failed to deactivate infraction #{id_} ({type_})")
-            log_text["Failure"] = f"HTTPException with status {e.status} and code {e.code}."
-            log_content = mod_role.mention
+            if e.code == 10007:
+                log.info(
+                    f"Can't pardon {infraction['type']} for user {infraction['user']} because user left from guild."
```
MarkKoz, line 366:
```suggestion
                    f"Can't pardon {infraction['type']} for user {infraction['user']} because user left the guild."
```

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -337,10 +358,17 @@ async def deactivate_infraction(
...
+                )
+            else:
+                log.exception(f"Failed to deactivate infraction #{id_} ({type_})")
+                log_text["Failure"] = f"HTTPException with status {e.status} and code {e.code}."
+                log_content = mod_role.mention
```
MarkKoz, line 371: A similar thing needs to be done (but with a different message) when the user isn't found, since that is still considered a failure.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -74,8 +74,23 @@ async def reapply_infraction(
...
+        except discord.HTTPException as e:
+            if e.code == 10007:
+                log.info(f"Can't reapply {infraction['type']} to user {infraction['user']} because user left again.")
+            else:
+                log.warning(
+                    (
+                        f"Got unexpected HTTPException (HTTP {e.status}, Discord code {e.code})"
+                        f"when awaiting {infraction['type']} coroutine for {infraction['user']}."
+                    )
+                )
```
MarkKoz, line 91: The second pair of parenthesis are redundant.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -74,8 +74,23 @@ async def reapply_infraction(
...
+        try:
+            await apply_coro
+        except discord.NotFound:
+            # When user joined and then right after this left again before action completed, this can't add roles
+            log.info(f"Can't reapply {infraction['type']} to user {infraction['user']} because user left again.")
+        except discord.HTTPException as e:
+            if e.code == 10007:
+                log.info(f"Can't reapply {infraction['type']} to user {infraction['user']} because user left again.")
+            else:
+                log.warning(
```
MarkKoz, line 86: I think this should be logged as an exception rather than a warning to be consistent.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -74,8 +74,23 @@ async def reapply_infraction(
...
-        await apply_coro
-        log.info(f"Re-applied {infraction['type']} to user {infraction['user']} upon rejoining.")
+        try:
+            await apply_coro
+        except discord.NotFound:
+            # When user joined and then right after this left again before action completed, this can't add roles
+            log.info(f"Can't reapply {infraction['type']} to user {infraction['user']} because user left again.")
+        except discord.HTTPException as e:
+            if e.code == 10007:
+                log.info(f"Can't reapply {infraction['type']} to user {infraction['user']} because user left again.")
```
MarkKoz, line 84: These can be merged by checking the status as mentioned in another comment.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -155,6 +170,8 @@ async def apply_infraction(
                 if expiry:
                     # Schedule the expiration of the infraction.
                     self.schedule_expiration(infraction)
+            except discord.NotFound:
+                log.info(f"Can't apply {infraction['type']} to user {infraction['user']} because user left from guild.")
```
MarkKoz, line 174: This can be merged with the 10007 check by checking the status as mentioned in another comment.

bot/exts/moderation/infraction/_scheduler.py
MarkKoz, line 84: The message is coupled too closely to its sole use in `on_member_join`. However, this function isn't technically restricted to only being used there.
```suggestion
                log.info(f"Can't reapply {infraction['type']} to user {infraction['user']} because user the guild.")
```

bot/bot.py
```diff
@@ -170,6 +174,16 @@ def clear(self) -> None:
 
     async def close(self) -> None:
         """Close the Discord connection and the aiohttp session, connector, statsd client, and resolver."""
+        # Done before super().close() to allow tasks finish before the HTTP session closes.
+        with suppress(Exception):
+            [self.unload_extension(ext) for ext in tuple(self.extensions)]
+            [self.remove_cog(cog) for cog in tuple(self.cogs)]
```
MarkKoz, line 180: `suppress()` will suppress the first exception is encounters and then resume execution on the line following the end of the with statement (i.e. line 183). This effectively prematurely ends the unloading/removal operation. 

Each call needs to be surrounded in a try-except.  Basically, it needs to be similar to how discord.py does it, which you used to follow but changed for some reason.

MarkKoz (changes requested):

MarkKoz (approved):

bot/exts/moderation/watchchannels/_watchchannel.py
```diff
@@ -342,11 +342,14 @@ def cog_unload(self) -> None:
         """Takes care of unloading the cog and canceling the consumption task."""
         self.log.trace("Unloading the cog")
         if self._consume_task and not self._consume_task.done():
+            def done_callback(task: asyncio.Task) -> None:
+                """Send exception when consuming task have been cancelled."""
+                try:
+                    task.result()
+                except asyncio.CancelledError:
+                    self.log.error(
```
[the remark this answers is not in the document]
ks129: Changed this log from error to info.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -352,9 +369,16 @@ async def deactivate_infraction(
             log_text["Failure"] = "The bot lacks permissions to do this (role hierarchy?)"
             log_content = mod_role.mention
         except discord.HTTPException as e:
-            log.exception(f"Failed to deactivate infraction #{id_} ({type_})")
-            log_text["Failure"] = f"HTTPException with status {e.status} and code {e.code}."
-            log_content = mod_role.mention
+            if e.code == 10007 or e.status == 404:
```
Xithrius, line 372: What does error code `10007` imply?

ks129: 10007 is Unknown Member JSON error code.

Xithrius: Thank you for this information.

Xithrius (approved):
LGTM!

PR #1021, opened 2020-07-01
Slowmode command

### LINKED ISSUE #1019: Introduce command to control slowmode
**Context**
High profile channels such as `#python-general` have slowmode enabled to make chat traffic manageable. The amount of delay is adjusted based on current traffic in the channel.

**Problem**
The Discord GUI only allows `@Admins` to choose from a limited amount of options.

image

The API, however, [allows much more granular control](https://discordpy.readthedocs.io/en/latest/api.html#discord.TextChannel.edit), taking ints in the range `[0, 21600]`. As such, we often use int eval to set custom values, commonly 2 seconds during low traffic times.

**Proposal**
The proposal is to introduce a bot command which will take a channel and a delay, and forward them to the API.

The advantages include:
* Convenience: we no longer need to int eval custom values
* The command can be made available to mods, which otherwise do not have access to the channel control panel

The implementation should be fairly trivial. Command signature takes a text channel (a `d.py` convertor makes this convenient), and an `int` delay. The delay is checked to ensure it fits into the correct range, with an informative message returned otherwise. If the delay is valid, it is passed to the `TextChannel.edit` method (linked above).

Den4200:
## Description
This new command will allow moderators+ to set the slowmode delay for a given text channel. Currently, only admins can control the slowmode delay, but only through int eval when we need more granular control. This command will make it much more convenient for the admins, and will actually allow the mods to control slowmode.

Normal members can get the slowmode delay for a given text channel.

Closes #1019.

## Command Showcase
image

#### Command Aliases
- `!slowmode` => `!sm`
- `!slowmode get` => `!sm g`
- `!slowmode set` => `!sm s`
- `!slowmode reset` => `!sm r`

#### Additional Details
If any of these commands are invoked without a given channel, the channel that the command was invoked in will be used instead.


bot/cogs/slowmode.py
```diff
@@ -0,0 +1,85 @@
...
+        """Get the slowmode delay for a given text channel."""
+        delay = relativedelta(seconds=channel.slowmode_delay)
+
+        try:
+            humanized_delay = time.humanize_delta(delay, precision=3)
+
+        except TypeError:
+            # The slowmode delay is 0 seconds,
+            # which causes `time.humanize_delta` to raise a TypeError
+            humanized_delay = '0 seconds'
```
kwzrd, line 39: Looking over the `humanize_delta` source, it wants a string for `precision`.

It will be matched against these:

```python
units = (
    ("years", delta.years),
    ("months", delta.months),
    ("days", delta.days),
    ("hours", delta.hours),
    ("minutes", delta.minutes),
    ("seconds", delta.seconds),
)
```

The `TypeError` comes from here and is a result of an incorrect type being passed in:

https://github.com/python-discord/bot/blob/74de316657d5fed9d5474fa23d05cbe09d9ec470/bot/utils/time.py#L25-L26

bot/cogs/slowmode.py
```diff
@@ -0,0 +1,85 @@
...
+
+        humanized_delay = time.humanize_delta(delay, precision=3)
+
+        if 0 <= slowmode_delay <= 21600:
+            await channel.edit(slowmode_delay=slowmode_delay)
+            await ctx.send(
+                f'{Emojis.check_mark} The slowmode delay for {channel.mention} is now {humanized_delay}.'
+            )
+
+            log.info(f'{ctx.author} set the slowmode delay for #{channel} to {humanized_delay}.')
```
kwzrd, line 61: Generally better to log before the operation rather than after - the log won't be made if something goes wrong above it.

Numerlor:
> ... Regular users can see the delay in the Discord GUI, I'm not sure whether it's at all needed to allow non-mods to use this cog.

Is the get command needed at all? Checking through the GUI will be more convenient and faster in most cases

j03b:
It is needed, yes.

While the UI does present this it only allows it to be locked to the increments permitted through the UI, not what we want.

For example, consider I put a channel on a 2 second slowmode:
image

We can verify slowmode is enabled by looking at the bottom right of the client:
Screenshot 2020-07-03 at 13 08 03

However there is no way to find this duration, if you open up the channel settings:
image

However I agree there is no need for non-mods to be able to use this command.

kwzrd:
It should show including custom values on hover.

image

j03b:
TIL hah. In that case I guess it is not needed.

bot/cogs/slowmode.py
Den4200: I don't know how I missed this. I'll fix that up 👍

Den4200:
I don't mind locking the whole cog to mods, sounds good to me.

As for the get command, I added that so the cog would have a complete pair of get and set. I can get get rid of it though, but I don't see any harm in having it either.

Numerlor:
Forgot about it on the previous comment, this cog would be more fitting in the moderation subpackage

Den4200:
I don't particularly think this fits in the moderation subpackage. The cogs should really be organized as a whole into separate subpackages, similar to as stated here: https://github.com/python-discord/bot/issues/160#issuecomment-643844893.

Then, I think this cog would be more fitting under the `utils` subpackage.

bot/cogs/slowmode.py
```diff
@@ -0,0 +1,86 @@
...
+    async def slowmode_group(self, ctx: Context) -> None:
+        """Get and set the slowmode delay for a given text channel."""
+        await ctx.send_help(ctx.command)
+
+    @slowmode_group.command(name='get', aliases=['g'])
+    async def get_slowmode(self, ctx: Context, channel: TextChannel) -> None:
+        """Get the slowmode delay for a given text channel."""
+        delay = relativedelta(seconds=channel.slowmode_delay)
+
+        # Say "0 seconds" instead of "less than a second"
```
tagptroll1, line 33: Sort of doesn't add much?
The code itself explains what it does, but the comment could mention why we don't give 0 to `time.humanize_delta(delay)`

bot/cogs/slowmode.py
```diff
@@ -0,0 +1,86 @@
...
+    async def set_slowmode(self, ctx: Context, channel: TextChannel, delay: DurationDelta) -> None:
+        """Set the slowmode delay for a given text channel."""
+        # Convert `dateutil.relativedelta.relativedelta` to `datetime.timedelta`
+        # Must do this to get the delta in a particular unit of time
+        utcnow = datetime.utcnow()
+        slowmode_delay = (utcnow + delay - utcnow).total_seconds()
+
+        humanized_delay = time.humanize_delta(delay)
+
+        if 0 <= slowmode_delay <= 21600:
```
tagptroll1, line 51: Possible to comment or create a variable for what 0 and 21600 is?

bot/cogs/slowmode.py
```diff
@@ -0,0 +1,86 @@
...
+
+        else:
+            log.info(
+                f'{ctx.author} tried to set the slowmode delay of #{channel} to {humanized_delay}, '
+                'which is not between 0 and 6 hours.'
+            )
+
+            await ctx.send(
+                f'{Emojis.cross_mark} The slowmode delay must be between 0 and 6 hours.'
+            )
```
tagptroll1, line 67: This could be moved up as a fail early check that returns, instead of if/else

bot/cogs/slowmode.py
```diff
@@ -0,0 +1,86 @@
...
+    @slowmode_group.command(name='get', aliases=['g'])
+    async def get_slowmode(self, ctx: Context, channel: TextChannel) -> None:
+        """Get the slowmode delay for a given text channel."""
+        delay = relativedelta(seconds=channel.slowmode_delay)
+
+        # Say "0 seconds" instead of "less than a second"
+        if channel.slowmode_delay == 0:
+            humanized_delay = '0 seconds'
+        else:
+            humanized_delay = time.humanize_delta(delay)
```
lemonsaurus, line 37: We shouldn't monkeypatch the functionality in our own utils inside of unrelated cogs. This is a code smell.

The change is fine, but if you're going to make this change it should be made inside of `utils.time`, not here.

bot/cogs/slowmode.py
lemonsaurus, line 51: It's unclear what this hardcoded `21600` represents. Either make a constant with a name that makes sense (`SLOWMODE_MAX_DELAY`?) , or add a block comment to explain what you're doing.

bot/cogs/slowmode.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self, bot: Bot) -> None:
+        self.bot = bot
+
+    @group(name='slowmode', aliases=['sm'], invoke_without_command=True)
+    async def slowmode_group(self, ctx: Context) -> None:
+        """Get and set the slowmode delay for a given text channel."""
+        await ctx.send_help(ctx.command)
+
+    @slowmode_group.command(name='get', aliases=['g'])
+    async def get_slowmode(self, ctx: Context, channel: TextChannel) -> None:
```
lemonsaurus, line 29: Maybe this can default to `ctx.channel` if no `channel` is provided?

bot/cogs/slowmode.py
```diff
@@ -0,0 +1,86 @@
...
+        if channel.slowmode_delay == 0:
+            humanized_delay = '0 seconds'
+        else:
+            humanized_delay = time.humanize_delta(delay)
+
+        await ctx.send(f'The slowmode delay for {channel.mention} is {humanized_delay}.')
+
+    @slowmode_group.command(name='set', aliases=['s'])
+    async def set_slowmode(self, ctx: Context, channel: TextChannel, delay: DurationDelta) -> None:
+        """Set the slowmode delay for a given text channel."""
```
lemonsaurus, line 43: We definitely should not have to provide the channel for this command. Use a greedy converter to get the channel optionally. If a channel is provided as the first argument, target that channel. Otherwise, use `ctx.channel`.

If I'm in #python-general, I want to be able to do `!slowmode set 3S`, but I also want the option of being able to do `!slowmode set #python-general 3S` from #mods if desired.

lemonsaurus: For an example of a greedy converter, see the `!clean` command.

bot/cogs/slowmode.py
```diff
@@ -0,0 +1,86 @@
...
+                'which is not between 0 and 6 hours.'
+            )
+
+            await ctx.send(
+                f'{Emojis.cross_mark} The slowmode delay must be between 0 and 6 hours.'
+            )
+
+    @slowmode_group.command(name='reset', aliases=['r'])
+    async def reset_slowmode(self, ctx: Context, channel: TextChannel) -> None:
+        """Reset the slowmode delay for a given text channel to 0 seconds."""
```
lemonsaurus, line 71: Again - optional channel, please.

bot/cogs/slowmode.py
Den4200: Yeah, that would make sense. Thanks for pointing this out.

bot/cogs/slowmode.py
Den4200: I removed the monkeypatch there and applied the changes needed in `bot.utils.time`, as mentioned here: https://github.com/python-discord/bot/pull/1021#discussion_r449915881. There is no longer a need for this comment.

bot/cogs/slowmode.py
Den4200: I think that it would be clearer to leave as an if/else in this situation. Does anyone else have thoughts on this?

bot/cogs/slowmode.py
Den4200: I don't think the greedy converter is necessary here. `typing.Optional` should work fine, since we only want one channel.

bot/cogs/slowmode.py
lemonsaurus: ah, yeah you're right. greedy would make sense if you wanted multiple, which doesn't make sense. Sorry, it was late.

bot/cogs/slowmode.py
kwzrd: Does naming the constant `SLOWMODE_MAX_DELAY` add any kind of information? It feels like that's the one bit of information that is obvious from the conditional - it being the max delay.

Numerlor:
> I don't particularly think this fits in the moderation subpackage. The cogs should really be all organized into separate subpackages, similar to as stated here: [#160 (comment)](https://github.com/python-discord/bot/issues/160#issuecomment-643844893).

Why do you feel like it wouldn't fit there? The slow mode is a moderation tool and the cog is restricted to moderators only. Maybe some others could weigh in here?
I agree that #160 needs to be realized in some way to declutter the cogs but I also don't particularly see this as an util feature with that structure

bot/cogs/slowmode.py
```diff
@@ -0,0 +1,97 @@
...
+
+        # Convert `dateutil.relativedelta.relativedelta` to `datetime.timedelta`
+        # Must do this to get the delta in a particular unit of time
+        utcnow = datetime.utcnow()
+        slowmode_delay = (utcnow + delay - utcnow).total_seconds()
+
+        humanized_delay = time.humanize_delta(delay)
+
+        # Ensure the delay is within discord's limits
+        if slowmode_delay <= SLOWMODE_MAX_DELAY:
```
tagptroll1, line 58: What happens if we provide negative numbers, and should there be a test for that?

Den4200:
@Numerlor After some discussion in `#dev-core`, it would make sense to have it in the `moderation` directory. Also, waiting for #160 to happen will probably be quite a while from now.

Numerlor:
> Also, waiting for #160 to happen will probably be quite a while from now.

Yeah, not trying to suggest that since this also partly blocks it

PR #1028, opened 2020-07-05
Cog for code snippet links

dolphingarlic:
I thought that it would be pretty cool if Discord could show the contents of a snippet link instead of just the link to the snippet, so I made a [bot for that](https://github.com/dolphingarlic/git-the-lines).

However, since adding third-party bots to the Python server was out of the question, I decided to add that functionality to the official server bot instead.

Proof


ks129:
@dolphingarlic I don't think that this should good feature. Sending something every time when someone posts GitHub link is not a good idea. And all PRs have to have approved the linked issue, what this PR definitely don't have. Make issue first, then staff/core devs will discuss it and then decide will this come to bot or not.

dolphingarlic:
Sorry, I didn't know that I should've created an issue first. (I don't think `CONTRIBUTING.md` mentioned that).

I think that printing out snippets would be quite useful since it means that people wouldn't have to open the links each time anymore. Sending embeds with details about repos basically just improves what Discord already does with repo links but with more details.

aadibajpai:
> @dolphingarlic I don't think that this should good feature. Sending something every time when someone posts GitHub link is not a good idea. And all PRs have to have approved the linked issue, what this PR definitely don't have. Make issue first, then staff/core devs will discuss it and then decide will this come to bot or not.

I *think* this was discussed in the #meta channel over at PyDis and there was some merit in the linked idea. Personally I think it is useful, especially when getting help in any of the help channels because you either copy paste your code standalone or link to it which would mean for the other person to switch back and forth.

sco1:
> Sorry, I didn't know that I should've created an issue first. (I don't think CONTRIBUTING.md mentioned that).

It's not mentioned because it's only a hard requirement for Seasonalbot. It's generally a good idea to make an issue first so you don't "waste" your time working on something that might not be merged, but since this seems to be something you've made for other reasons already it's a bit of a moo point.

I'm a bit meh on the repo link prettification, but the snippet formatting is deinitely interesting. We'll have to get some more opinions before we review formally, so please stand by. Is this something you've discussed in the server previously?

dolphingarlic:
Yes - I've discussed it once in `#meta` a while ago

dolphingarlic:
I can remove the repo link prettification if you'd like, but here's an example of when the snippet link feature is very useful:

git

Here, seeing just the single line linked in the snippet instead of having to open the link in a new tab is much more convenient. It would make debugging a lot faster.

SebastiaanZ:
As Discord already embeds a link preview by default for GitHub-links, I'm not sure if I'm in favor of having the bot respond with a second embed containing some details about the repository. If we were to have "repository information" feature, I think it would be better as an explicitly invoked command. I'm just not sure whether or not such a command fits the scope of our community utility bot. (I really mean: "not sure"; it's not a euphemism for "I don't think it fits".)

I do like the snippet feature, although I'm a bit concerned of "surprising" vertical space that's get inserted in a conversation. Maybe we could limit it to snippets that are only a few lines long? I wouldn't be surprised if people pasting a link to a snippet of code don't expect the bot to respond with a 15-line code snippet in the middle of a conversation. Alternatively, this could also have a command interface, which ensures that it's explicitly invoked.

Numerlor:
If this were to be included, I'd prefer it to be a command or limited to maybe 3 lines without directly invoking it instead of looking up links in messages. Most github permalinks that I've seen get posted in discussion channels like python-language where this behavior wouldn't be as desired
While the repo information is nice I don't think the additional info like stars is particularly useful in most situations

ks129:
One more thing: When we do this, this should get filtering too. Maybe someone wants to write a message that gets deleted by filters, but then add this GitHub and post a link. And however, this should have some text filter.

dolphingarlic:
So since nobody seems to like the repo widget prettification feature, I removed that. To address the issue of people not wanting the bot to interrupt their conversations, I added the option for the author of the message to react with an ❌ emoji to delete the bot's message.

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+                        )
+                        message_to_send += await snippet_to_embed(d, file_contents)
+                        break
+
+            for gl in GITLAB_RE.finditer(message.content):
+                d = gl.groupdict()
+                await orig_to_encode(d)
+                headers = {}
+                if 'GITLAB_TOKEN' in os.environ:
+                    headers['PRIVATE-TOKEN'] = os.environ["GITLAB_TOKEN"]
```
MarkKoz, line 148: Is this optional just for higher rate limits? Does anyone think this key needs to be added to the config?

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+                            self.session,
+                            gist_json['files'][f]['raw_url'],
+                            'text',
+                        )
+                        message_to_send += await snippet_to_embed(d, file_contents)
+                        break
+
+            for gl in GITLAB_RE.finditer(message.content):
+                d = gl.groupdict()
+                await orig_to_encode(d)
```
MarkKoz, line 145: Doesn't aiohttp already encode the URLs properly?

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+                if 'GITLAB_TOKEN' in os.environ:
+                    headers['PRIVATE-TOKEN'] = os.environ["GITLAB_TOKEN"]
+                file_contents = await fetch_http(
+                    self.session,
+                    f'https://gitlab.com/api/v4/projects/{d["repo"]}/'
+                    + f'repository/files/{d["file_path"]}/raw?ref={d["branch"]}',
+                    'text',
+                    headers=headers,
+                )
+                await revert_to_orig(d)
```
MarkKoz, line 156: If you really do need to encode it yourself, then use a copy of the dictionary to avoid needing to revert it.

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+async def orig_to_encode(d: dict) -> dict:
+    """Encode URL Parameters."""
+    for obj in d:
+        if d[obj] is not None:
+            d[obj] = d[obj].replace('/', '%2F').replace('.', '%2E')
+
+
+async def snippet_to_embed(d: dict, file_contents: str) -> str:
```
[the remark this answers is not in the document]
MarkKoz: Some comments in the function body would be nice to see.

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+    required = '\n'.join(split_file_contents[start_line - 1:end_line])
+    required = textwrap.dedent(required).rstrip().replace('`', '`\u200b')
+
+    language = d['file_path'].split('/')[-1].split('.')[-1]
+    if not language.replace('-', '').replace('+', '').replace('_', '').isalnum():
+        language = ''
+
+    if len(required) != 0:
+        return f'```{language}\n{required}```\n'
+    return '``` ```\n'
```
MarkKoz, line 62: If it's empty, I feel like sending an empty code block is redundant.

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+        if d[obj] is not None:
+            d[obj] = d[obj].replace('%2F', '/').replace('%2E', '.')
+
+
+async def orig_to_encode(d: dict) -> dict:
+    """Encode URL Parameters."""
+    for obj in d:
+        if d[obj] is not None:
+            d[obj] = d[obj].replace('/', '%2F').replace('.', '%2E')
+
```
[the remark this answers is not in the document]
spyoungtech: I think `urllib.parse.quote` and `urllib.prase.parse_qs` would do the trick. Technically, `/` is safe as part of URL parameters, but you can pass `safe=''` to `quote` if you want slashes to be encoded.

dolphingarlic:
I've made all the changes requested besides for the regex ones. (Sorry I neglected this PR - I've had exams the past few weeks)

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+
+GITHUB_RE = re.compile(
+    r'https://github\.com/(?P<repo>.+?)/blob/(?P<branch>.+?)/'
+    + r'(?P<file_path>.+?)#L(?P<start_line>\d+)([-~]L(?P<end_line>\d+))?\b'
+)
+
+GITHUB_GIST_RE = re.compile(
+    r'https://gist\.github\.com/([^/]*)/(?P<gist_id>[0-9a-zA-Z]+)/*'
+    + r'(?P<revision>[0-9a-zA-Z]*)/*#file-(?P<file_path>.+?)'
```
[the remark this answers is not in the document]
dolphingarlic: I don't know why, but changing `/*` to `/+` seems to break the regex. I think it might be because the hash is optional in gist snippet links

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+
+    if start_line > end_line:
+        start_line, end_line = end_line, start_line
+    if start_line > len(split_file_contents) or end_line < 1:
+        return ''
+    start_line = max(1, start_line)
+    end_line = min(len(split_file_contents), end_line)
+
+    required = '\n'.join(split_file_contents[start_line - 1:end_line])
+    required = textwrap.dedent(required).rstrip().replace('`', '`\u200b')
```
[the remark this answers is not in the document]
dolphingarlic: Escaping the \` characters doesn't work in Discord for some reason, so I feel like this is the best option. (I don't think people will normally have \`\`\` in their code anyway)

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+import aiohttp
+from discord import Message, Reaction, User
+from discord.ext.commands import Cog
+
+from bot.bot import Bot
+
+
+async def fetch_http(session: aiohttp.ClientSession, url: str, response_format: str, **kwargs) -> str:
+    """Uses aiohttp to make http GET requests."""
+    async with session.get(url, **kwargs) as response:
```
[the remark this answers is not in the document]
dolphingarlic: Should I use the built-in Python logging or is there something else I should use exclusive to this bot?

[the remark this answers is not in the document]
lemonsaurus: we use regular logging - just take a look at pretty much any other cog and see how it's done there.

The only thing special about our logging is that we have added a TRACE level, where you can put logging that will only appear when running the bot locally, stuff that might be useful when testing.

bot/cogs/code_snippets.py
```diff
@@ -0,0 +1,216 @@
...
+                message_to_send += await fetch_github_snippet(self.bot.http_session, **gh.groupdict())
+
+            for gh_gist in GITHUB_GIST_RE.finditer(message.content):
+                message_to_send += await fetch_github_gist_snippet(self.bot.http_session, **gh_gist.groupdict())
+
+            for gl in GITLAB_RE.finditer(message.content):
+                message_to_send += await fetch_gitlab_snippet(self.bot.http_session, **gl.groupdict())
+
+            for bb in BITBUCKET_RE.finditer(message.content):
+                message_to_send += await fetch_bitbucket_snippet(self.bot.http_session, **bb.groupdict())
```
lemonsaurus, line 203: I feel like there's a DRYer solution available here.

Maybe make a list of tuples, something like this:
```py
pattern_handlers = [
    (GITHUB_RE, fetch_github_snippet),
    (GITHUB_GIST_RE, fetch_github_gist_snippet),
    ...
]
```
And then you can just iterate this like
```py
for pattern, handler in pattern_handlers:
    for match in pattern.finditer(message.content):
        message_to_send += await handler(self.bot.http_session, **match.groupdict())
```

This is a little oversimplified (you still need to do the search), but you get the idea.

bot/cogs/code_snippets.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+async def fetch_github_snippet(session: ClientSession, repo: str,
+                               path: str, start_line: str, end_line: str) -> str:
+    """Fetches a snippet from a GitHub repo."""
+    headers = {'Accept': 'application/vnd.github.v3.raw'}
+
+    # Search the GitHub API for the specified branch
+    refs = (await fetch_http(session, f'https://api.github.com/repos/{repo}/branches', 'json', headers=headers)
+            + await fetch_http(session, f'https://api.github.com/repos/{repo}/tags', 'json', headers=headers))
```
MarkKoz, line 29:
```suggestion
    branches = await fetch_http(session, f'https://api.github.com/repos/{repo}/branches', 'json', headers=headers)
    tags = await fetch_http(session, f'https://api.github.com/repos/{repo}/tags', 'json', headers=headers)
    refs = branches + tags
```

bot/cogs/code_snippets.py
```diff
@@ -0,0 +1,216 @@
...
+        session,
+        f'https://api.github.com/repos/{repo}/contents/{file_path}?ref={ref}',
+        'text',
+        headers=headers,
+    )
+
+    return await snippet_to_md(file_contents, file_path, start_line, end_line)
+
+
+async def fetch_github_gist_snippet(session: ClientSession, gist_id: str, revision: str,
```
MarkKoz, line 49: If you kept everything inside the cog as methods, then you wouldn't need to keep passing around the session. But whatever, this is okay as-is.

bot/cogs/print_snippets.py
MarkKoz: It's good that you switched your own methods out, but you never answered this question despite resolving it. Can you explicitly confirm whether or not escaping yourself is necessary?

bot/cogs/print_snippets.py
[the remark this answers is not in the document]
MarkKoz: You're right. Leaving it at `/*` is fine. The slash before the fragment is optional, but valid. The slash before the revision is mandatory if the revision exists, but in practice `/*` works out for it anyway. The alternative of `(?:/(?P<revision>[^\W_]+))?/?` would be more "correct" but ugly and complicated.

bot/cogs/code_snippets.py
lemonsaurus: I also would prefer these inside the cog. I think module-level functions in cogs should be avoided.

MarkKoz:
For future changes, please don't push them all in a single commit. It makes it hard for reviewers. See point № 4 in our contributing guidelines https://github.com/python-discord/bot/blob/master/CONTRIBUTING.md

kwzrd:
Hi @dolphingarlic, are you still interested in working on this?

dolphingarlic:
Hi, I'm still interested in working on this. I've just been overwhelmed by finals/IOI/IMO lately and haven't found the time or energy to address the reviews :grimacing: 

I'll definitely be able to work on this in a few (~3) weeks though

kwzrd:
Ok, that is fine, no worries. Just making sure the PR hasn't been abandoned.

bot/cogs/print_snippets.py
dolphingarlic: I do need to escape it myself. I get a `404: Not Found` error without it

bot/cogs/code_snippets.py
dolphingarlic: I kept the regex as global variables because I saw that other cogs did the same. Is everything else that I did regarding this fine?

bot/cogs/code_snippets.py
dolphingarlic: I think the search is actually unnecessary, so this alone works fine

dolphingarlic:
Also, should I add tests for this cog? If so, how?

bot/exts/info/code_snippets.py
```diff
@@ -0,0 +1,226 @@
...
+            if response_format == 'text':
+                return await response.text()
+            elif response_format == 'json':
+                return await response.json()
+
+    def _find_ref(self, path: str, refs: tuple) -> tuple:
+        """Loops through all branches and tags to find the required ref."""
+        # Base case: there is no slash in the branch name
+        ref = path.split('/')[0]
+        file_path = '/'.join(path.split('/')[1:])
```
MarkKoz, line 53: Make use of the maxsplit argument
```suggestion
        ref, file_path  = path.split('/', 1)
```

MarkKoz (changes requested):
There are still a few unresolved comments from previous reviews. I also have more concerns about the regular expressions. I analysed the GitHub one and made some points below, but the other regular expressions likely have similar problems.

1. A repo may be called "blob", or the organisation/user. This can be fixed by using a more limited character set. Specifically, problems arise because it's currently matching `/`. A `/` is not even a valid character in a repository name (nor is it presumably a valid organisation/user name) according to [this](https://stackoverflow.com/questions/21262997/what-is-the-format-for-valid-git-repo-names), though I think just ignoring slashes will be good enough.
2. Account for the branch name being "blob". Make sure it doesn't mess up the regex.

MarkKoz:
> Also, should I add tests for this cog? If so, how?

No, you don't have to.

bot/cogs/code_snippets.py
MarkKoz: Yes, looks good.

bot/exts/info/code_snippets.py
```diff
@@ -0,0 +1,237 @@
...
+
+    async def _fetch_response(self, url: str, response_format: str, **kwargs) -> str:
+        """Makes http requests using aiohttp."""
+        try:
+            async with self.bot.http_session.get(url, **kwargs) as response:
+                if response_format == 'text':
+                    return await response.text()
+                elif response_format == 'json':
+                    return await response.json()
+        except Exception:
```
ks129, line 51: You should avoid bare exceptions.

bot/exts/info/code_snippets.py
```diff
@@ -0,0 +1,237 @@
...
+            log.exception(f'Failed to fetch code snippet from {url}.')
```
ks129, line 52: Use `log.error` instead.

bot/exts/info/code_snippets.py
```diff
@@ -0,0 +1,237 @@
...
+
+    async def _fetch_github_snippet(
+        self,
+        repo: str,
+        path: str,
+        start_line: str,
+        end_line: str
+    ) -> str:
+        """Fetches a snippet from a GitHub repo."""
+        headers = {'Accept': 'application/vnd.github.v3.raw'}
```
ks129, line 74: This should be constant.

bot/exts/info/code_snippets.py
```diff
@@ -0,0 +1,237 @@
...
+    async def _fetch_github_gist_snippet(
+        self,
+        gist_id: str,
+        revision: str,
+        file_path: str,
+        start_line: str,
+        end_line: str
+    ) -> str:
+        """Fetches a snippet from a GitHub gist."""
+        headers = {'Accept': 'application/vnd.github.v3.raw'}
```
ks129, line 98: This should be too constant

bot/exts/info/code_snippets.py
```diff
@@ -0,0 +1,237 @@
...
+    def __init__(self, bot: Bot):
+        """Initializes the cog's bot."""
+        self.bot = bot
+
+    @Cog.listener()
+    async def on_message(self, message: Message) -> None:
+        """Checks if the message has a snippet link, removes the embed, then sends the snippet contents."""
+        if not message.author.bot:
+            message_to_send = ''
+            pattern_handlers = [
```
ks129, line 215: This should be created in `__init__.py`.

ks129 (changes requested):
Few comments

bot/exts/info/code_snippets.py
MarkKoz: Strictly speaking, a bare exception is a term attributed to `except:`. `except Exception:` is relatively much more justifiable. Anyway, I agree that the exception should be more specific in this case. In fact, the `get` should be using `raise_for_status=True` so that an exception is raised in the first place.

bot/exts/info/code_snippets.py
MarkKoz: The log should also contain the HTTP status code from the response (which you can get by catching the correct error as mentioned in the other comment).

dolphingarlic: Is what I did in https://github.com/python-discord/bot/pull/1028/commits/08b793024f271de009aab2391cd85576af5313cf fine? I feel like the error message may get a bit long, so should I split it up into two lines instead?

bot/cogs/print_snippets.py
```diff
@@ -0,0 +1,190 @@
...
+)
+
+GITLAB_RE = re.compile(
+    r'https://gitlab\.com/(?P<repo>.+?)/\-/blob/(?P<branch>.+?)/'
+    + r'(?P<file_path>.+?)#L(?P<start_line>\d+)([-~](?P<end_line>\d+))?\b'
+)
+
+BITBUCKET_RE = re.compile(
+    r'https://bitbucket\.org/(?P<repo>.+?)/src/(?P<branch>.+?)/'
+    + r'(?P<file_path>.+?)#lines-(?P<start_line>\d+)(:(?P<end_line>\d+))?\b'
```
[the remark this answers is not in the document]
MarkKoz: @dolphingarlic this is still unresolved. Here is a live example of such URL if you need one:

https://bitbucket.org/m2m/cumulocity-examples/src/527d84ff48dd8488b248a312955f40a2fdd26691/snmp/integration-test/license.txt?at=develop#lines-10:13

It seems to add that query parameter when selecting a commit from the commit history drop down.

MarkKoz (changes requested):
Regex still isn't quite right. Again, I'm testing only for GitHub. You'll have to determine yourself what the behaviour for the other sites is. I think the blob support is the most important thing. If you can't figure out the rest then I'd be fine to merge without the other things fixed as they're edge cases.

The `repo` group still doesn't account for a name being "blob". Here is a regex that fixes that:

```regex
(?P<repo>[a-zA-Z0-9-]+/[\w.-]+)
```

With that change, it seems the GitHub regex will finally correctly account for account names, repository names, and branch names using "blob". Here are a bunch of test cases if you want to try them out:

```
https://github.com/repo-name/project-name/blob/branch/sub-branch/directory/file.py#L13-L14
https://github.com/repo-name/blob/blob/branch/sub-branch/directory/file.py#L13-L14
https://github.com/blob/project-name/blob/branch/sub-branch/directory/file.py#L13-L14
https://github.com/blob/blob/blob/branch/sub-branch/directory/file.py#L13-L14


https://github.com/repo-name/project-name/blob/blob/sub-branch/directory/file.py#L13-L14
https://github.com/repo-name/project-name/blob/branch/blob/directory/file.py#L13-L14

https://github.com/repo-name/blob/blob/blob/sub-branch/directory/file.py#L13-L14
https://github.com/repo-name/blob/blob/branch/blob/directory/file.py#L13-L14
```

Keep in mind that regex is specific to GitHub. Other sites may allow different characters for account or repository names. In fact, that regex isn't totally accurate for GitHub. For account names, it doesn't allow a hyphen at the start or end nor does it allow consecutive hyphens. For fun, I wrote a regex below that is more accurate but I don't think this feature needs to take it that far.

```regex
(?P<repo>(?!-)(?:[^\W_]|-(?!-))+(?<!-)/[\w.-]+)
```

---

The regex needs to only match the closing `>` if it already matched an opening `<` before the URL. Also keep in mind Discord doesn't allow spaces between either angle bracket and the enclosed URL. Currently, the regex behaves incorrectly if there are adjacent URLs in brackets without spaces between e.g. `<url><url>`.

---

While a space is valid in file names and GitHub can resolve a URL with a literal space (i.e. not URL-encoded), I think you have to compromise and not support non-encoded spaces in the path. Otherwise, it'd be ambiguous where the URL ends (since Linux is very permissive with what a file can be named).

---

Given a file name with one of the following characters, GitHub seems to not resolve a URL if `?`, `=`, or `#` are not URL-encoded. This makes it easier to deal with matching URL fragments and query parameters, since you don't have to worry about those characters literally being in the path.

---

While I haven't really seen GitHub add query parameters to URLs, they do support them and they do ignore any invalid parameters. The only exception I've seen is in raw links for private repositories, which have a `token` query parameter. I'm not sure if the regex should account for query parameters and just ignore them.

dolphingarlic:
Do you think I should just copy what I did for BitBucket query parameters for the other regexes? Since GitHub doesn't support straight-up `?` characters in the URL, I don't think this will break anything

bot/exts/info/code_snippets.py
```diff
@@ -18,22 +18,21 @@
...
+    r'(?P<revision>[a-zA-Z0-9-]*)/*#file-(?P<file_path>[^#>]+?)'
     r'(-L(?P<start_line>\d+)([-~:]L(?P<end_line>\d+))?)'
 )
 
 GITHUB_HEADERS = {'Accept': 'application/vnd.github.v3.raw'}
 
 GITLAB_RE = re.compile(
-    r'https://gitlab\.com/(?P<repo>\S+?)/\-/blob/(?P<path>\S+/[^\s#,>]+)'
-    r'(#L(?P<start_line>\d+)([-](?P<end_line>\d+))?)'
+    r'https://gitlab\.com/(?P<repo>[a-zA-Z0-9-]+?)/\-/blob/(?P<path>[^#>]+/{0,1})'
```
MarkKoz, line 29: The repo part is incorrect and completely breaks the regex. It must match both the username and the project name `(?P<repo>[\w.-]+/[\w.-]+)`

MarkKoz, line 29: What's the reason for doing `/{0,1}`? None of the other URLs seem to need it. Does GitLab do something extra which I'm not yet seeing?

bot/exts/info/code_snippets.py
```diff
@@ -18,22 +18,21 @@
 )
 
 GITHUB_GIST_RE = re.compile(
-    r'https://gist\.github\.com/([^/]+)/(?P<gist_id>[^\W_]+)/*'
-    r'(?P<revision>[^\W_]*)/*#file-(?P<file_path>\S+?)'
+    r'https://gist\.github\.com/([^/]+)/(?P<gist_id>[a-zA-Z0-9]+)/*'
```
MarkKoz, line 21: This should use the same regex for the username that the normal GitHub regex does. The current regex you use is too permissive.

bot/exts/info/code_snippets.py
```diff
@@ -18,22 +18,21 @@
...
+    r'(?P<revision>[a-zA-Z0-9-]*)/*#file-(?P<file_path>[^#>]+?)'
```
MarkKoz, line 22: Does it really allow dashes in the revision?

MarkKoz (changes requested):

MarkKoz:
> 
> 
> Do you think I should just copy what I did for BitBucket query parameters for the other regexes? Since GitHub doesn't support straight-up `?` characters in the URL, I don't think this will break anything

Yup, that sounds fine. I also think it's safe.

bot/exts/info/code_snippets.py
dolphingarlic: It's so that it matches `<part of the URL>/` for all parts of the path except the last, in which case it matches `<part of the URL>` only. I also do this in the GitHub regex. Is this part unnecessary?

MarkKoz: Sorry for getting back to you a bit late. I still don't understand what you mean. I thought you meant that it avoids matching a trailing slash in the path, but the regex still matches it. Here is a real example:

```
https://gitlab.com/gitlab-org/gitlab/-/blob/master/scripts/create_postgres_user.sh/#L3-4
```

Furthermore, GitLab resolves that URL just fine in my browser (though maybe it's cause my browser is silently stripping it). If the trailing slash is really a problem when you need to fetch the content, then using `rstrip("/")` in Python is a simple solution.

If you were not referring to a trailing slash, then please clarify.

dolphingarlic: Oops, I just misunderstood what my regex did. You're right - the `/{0,1}` is unnecessary because the trailing slash is already matched in the preceding regex. I tested it and the slash isn't a problem when I fetch the content

MarkKoz (changes requested):
Running the code reveals errors:

`https://gitlab.com/gitlab-org/gitlab/-/blob/master/scripts/create_postgres_user.sh#L3-5` leads to 
```
2021-01-30 12:12:37 | bot.exts.info.code_snippets | ERROR | Failed to fetch code snippet from https://api.github.com/repos/gitlab-org/gitlab/branches. HTTP Status: 404. Message: 404, message='Not Found', url=URL('https://api.github.com/repos/gitlab-org/gitlab/branches').
2021-01-30 12:12:37 | bot.exts.info.code_snippets | ERROR | Failed to fetch code snippet from https://api.github.com/repos/gitlab-org/gitlab/tags. HTTP Status: 404. Message: 404, message='Not Found', url=URL('https://api.github.com/repos/gitlab-org/gitlab/tags').
2021-01-30 12:12:37 | bot | ERROR | Unhandled exception in on_message.
  File "/home/mark/repos/python/bot-pydis/.venv/lib/python3.8/site-packages/discord/client.py", line 343, in _run_event
    await coro(*args, **kwargs)
  File "/home/mark/repos/python/bot-pydis/bot/exts/info/code_snippets.py", line 228, in on_message
    message_to_send += await handler(**match.groupdict())
  File "/home/mark/repos/python/bot-pydis/bot/exts/info/code_snippets.py", line 132, in _fetch_gitlab_snippet
    refs = branches + tags
TypeError: unsupported operand type(s) for +: 'NoneType' and 'NoneType'
```

Looks like it's trying to use the GitHub API to fetch a GitLab file. Perhaps you need some None checks too.

---

Any input (except for when the above happens) will lead to

```
2021-01-30 12:12:32 | bot | ERROR | Unhandled exception in on_message.
  File "/home/mark/repos/python/bot-pydis/.venv/lib/python3.8/site-packages/discord/client.py", line 343, in _run_event
    await coro(*args, **kwargs)
  File "/home/mark/repos/python/bot-pydis/bot/exts/info/code_snippets.py", line 232, in on_message
    await wait_for_deletion(
TypeError: wait_for_deletion() got an unexpected keyword argument 'client'
```

---

It'd be nicer to truncate code rather than showing nothing if it's too long. Right now it isn't immediately obvious why some links don't get code blocks from the bot. However, this isn't a requirement from me and I'd approve the PR regardless.

dolphingarlic:
Truncating might be a bit tricky when there are multiple snippet links in a single message. Do you think I should do something like truncate all code blocks by ceil((len(message) - 2000) / (number of code blocks)) chars if needed? Also, what about the number of lines allowed per message?

MarkKoz:
I imagine that trying to evenly distribute truncation would result in all code blocks being too heavily truncated to still be meaningful. A simpler strategy of truncating the first blocks and omitting the rest sounds fine.

Line limit should still apply to the final line count and will truncate at line boundaries. The limit could be raised to 20. Make sure you don't include "bare" file paths that lack code blocks because they got truncated. In other words, if there isn't enough space for the entire file path and at least some of the code block, then omit the file path too.

I think what I'm describing is pretty similar to the extant truncation logic for the snekbox eval command. It's worth investigating how much of that code can be shared between the two features.

Xithrius:
@dolphingarlic Greetings. Will you be finishing up this PR?

dolphingarlic:
I'm still waiting for my code to be reviewed again. Sorry about the delay

Xithrius:
Thanks for the quick reply! I'll poke some people for reviews.

bot/exts/info/code_snippets.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @Cog.listener()
+    async def on_message(self, message: Message) -> None:
+        """Checks if the message has a snippet link, removes the embed, then sends the snippet contents."""
+        if not message.author.bot:
+            message_to_send = ''
+
+            for pattern, handler in self.pattern_handlers:
+                for match in pattern.finditer(message.content):
+                    message_to_send += await handler(**match.groupdict())
```
MarkKoz, line 231: Ideally the matches would be sorted by their starting positions so that the output code blocks match the order of the URLs in the original message.

MarkKoz (approved):

vcokltfre (approved):
Looking good! Apologies I wasn't able to review this earlier I just got the notification that xith requested my review

bot/exts/info/code_snippets.py
```diff
@@ -224,13 +224,18 @@ def __init__(self, bot: Bot):
...
 
             for pattern, handler in self.pattern_handlers:
                 for match in pattern.finditer(message.content):
-                    message_to_send += await handler(**match.groupdict())
+                    snippet = await handler(**match.groupdict())
+                    all_snippets.append((match.start(), snippet))
 
-            if 0 < len(message_to_send) <= 2000 and message_to_send.count('\n') <= 15:
+            # Sorts the list of snippets by their match index and joins them into
+            # a single message
```
Xithrius, line 235:
```suggestion
            # Sorts the list of snippets by their match index and joins them into a single message
```
If I'm not mistaken, this should be fine given the config line below for a max line length of 120.

https://github.com/python-discord/bot/blob/main/tox.ini#L2

PR #1032, opened 2020-07-08
Sanitize bot output

### LINKED ISSUE #1031: Sanitize bot output
## The problem
Right now, it is possible to exploit certain parts of the bot in order to make the bot post things a regular member would normally not be allowed to post.

These techniques can be used to circumvent our filters, which is somewhat problematic.

For example, by using eval to build a string containing a discord guild invite, you can circumvent our guild invite filter.
image

You can also do this by exploiting the fact that our custom exception format will repeat the input
image

## Running filters over eval output
To fix the exploit involving `!eval`, we should change the `!eval` command so that it will not return anything that triggers a filter.

Let's change the Snekbox cog so that we will do the following:
- Call snekbox and get the output
- **Run the output through all the active filters.**
- If a filter is triggered, do not post the output. Instead post a message that says something like `Attempt to circumvent filter detected. Moderator team has been alerted.`

To do this, we will probably also need to make some changes to the Filtering cog so that we can run filters over an arbitrary string.

## Changing the custom exception output
Currently when we return our custom exceptions, we repeat the input that failed. The simplest solution is not to do this.

Let's simply change this code so that the exception output does not repeat the input, but otherwise looks more or less the same.

slushiegoose:
Fixes #1031 

This removes the argument from the BadArgument command and runs filters over `!eval` output, as requested.

### DIFF
--- a/bot/cogs/error_handler.py
+++ b/bot/cogs/error_handler.py
@@ -170,7 +170,7 @@ class ErrorHandler(Cog):
             await prepared_help_command
             self.bot.stats.incr("errors.too_many_arguments")
         elif isinstance(e, errors.BadArgument):
-            await ctx.send("Bad argument: Please double check your input arguments and try again.\n")
+            await ctx.send("Bad argument: Please double-check your input arguments and try again.\n")
             await prepared_help_command
             self.bot.stats.incr("errors.bad_argument")
         elif isinstance(e, errors.BadUnionArgument):
--- a/bot/cogs/error_handler.py
+++ b/bot/cogs/error_handler.py
@@ -170,7 +170,7 @@ class ErrorHandler(Cog):
             await prepared_help_command
             self.bot.stats.incr("errors.too_many_arguments")
         elif isinstance(e, errors.BadArgument):
-            await ctx.send(f"Bad argument: {e}\n")
+            await ctx.send("Bad argument: Please double check your input arguments and try again.\n")
             await prepared_help_command
             self.bot.stats.incr("errors.bad_argument")
         elif isinstance(e, errors.BadUnionArgument):
--- a/bot/cogs/filtering.py
+++ b/bot/cogs/filtering.py
@@ -207,6 +207,7 @@ class Filtering(Cog, Scheduler):
         Also requires the original message, to check whether to filter and for mod logs.
         Returns whether a filter was triggered or not.
         """
+        filter_triggered = False
         # Should we filter this message?
         if self._check_filter(msg):
             for filter_name, _filter in self.filters.items():
--- a/bot/cogs/filtering.py
+++ b/bot/cogs/filtering.py
@@ -217,43 +217,43 @@ class Filtering(Cog, Scheduler):
                 if _filter["enabled"] and _filter["content_only"]:
                     match = await _filter["function"](result)
 
-                if match:
-                    # If this is a filter (not a watchlist), we set the variable so we know
-                    # that it has been triggered
-                    if _filter["type"] == "filter":
-                        filter_triggered = True
+                    if match:
+                        # If this is a filter (not a watchlist), we set the variable so we know
+                        # that it has been triggered
+                        if _filter["type"] == "filter":
+                            filter_triggered = True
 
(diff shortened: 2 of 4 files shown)


slushiegoose:
> This only fixes one possible path in the error handler, not all of them, 

this is the only path that repeats the argument

> also has missed the docs cog.

lemon mentioned in https://github.com/python-discord/bot/issues/1031#issuecomment-655702470 that the docs don't actually circumvent it very efficiently so it isn't really a problem

bot/cogs/error_handler.py
```diff
@@ -170,7 +170,7 @@ async def handle_user_input_error(self, ctx: Context, e: errors.UserInputError)
             await prepared_help_command
             self.bot.stats.incr("errors.too_many_arguments")
         elif isinstance(e, errors.BadArgument):
-            await ctx.send(f"Bad argument: {e}\n")
+            await ctx.send("Bad argument: Please double check your input arguments and try again.\n")
```
lemonsaurus, line 173:
```suggestion
            await ctx.send("Bad argument: Please double-check your input arguments and try again.\n")
```
It seems the more correct form is to hyphenate here.

lemonsaurus:
@slushiegoose Can you sort out the failing test?

bot/cogs/filtering.py
```diff
@@ -200,24 +200,67 @@ async def check_bad_words_in_name(self, member: Member) -> None:
             # Update time when alert sent
             await self.name_alerts.set(member.id, datetime.utcnow().timestamp())
 
+    async def _filter_eval(self, result: str, msg: Message) -> bool:
```
MarkKoz, line 203: It's used by snekbox, so don't name it like a private function.

bot/cogs/filtering.py
```diff
@@ -437,7 +506,7 @@ async def _has_rich_embed(msg: Message) -> bool:
                     if not embed.url or embed.url not in urls:
                         # If `embed.url` does not exist or if `embed.url` is not part of the content
                         # of the message, it's unlikely to be an auto-generated embed by Discord.
-                        return True
+                        return msg.embeds
```
MarkKoz, line 509: This change breaks the following check:

```py
if name == "filter_invites" and match is not True:
```

It will also break the following lines since trying to use `.items()` will fail on a list.

```py
for invite, data in match.items():
```

bot/cogs/snekbox.py
```diff
@@ -212,7 +212,12 @@ async def send_eval(self, ctx: Context, code: str) -> Message:
             else:
                 self.bot.stats.incr("snekbox.python.success")
 
-            response = await ctx.send(msg)
+            filter_cog = self.bot.get_cog("Filtering")
+            filter_triggered = await filter_cog._filter_eval(msg, ctx.message)
```
MarkKoz, line 216: You've misunderstood the issue. The bot's output needs to be filtered, not the invocation message. The latter was already getting caught by the filter.

slushiegoose: The invocation message is only passed to check for whether to filter the message (as Staff messages aren't filtered) and for user information for when a filter is triggered. The first parameter (`msg`) is the bot's output that's checked before it's sent

bot/cogs/filtering.py
slushiegoose: That change applies to the `watch_rich_embeds` filter so the first check will always fail since it is not the `filter_invites` check, which remains the same

bot/cogs/filtering.py
```diff
@@ -200,24 +200,67 @@ async def check_bad_words_in_name(self, member: Member) -> None:
...
+        # Should we filter this message?
+        if self._check_filter(msg):
+            for filter_name, _filter in self.filters.items():
+                # Is this specific filter enabled in the config?
+                # We also do not need to worry about filters that take the full message,
+                # since all we have is an arbitrary string.
+                if _filter["enabled"] and _filter["content_only"]:
+                    match = await _filter["function"](result)
+
+                if match:
```
MarkKoz, line 220: This causes an error

```
UnboundLocalError: local variable 'match' referenced before assignment
```

PR #1035, opened 2020-07-09
Add range-len tag

swfarnsworth:
This would be a message we could whip out when we see `range(len(...))` for loops during help sessions. I intentionally left out a lot of interesting details to keep the message concise.

I recall there being a RealPython article about "pythonic loops" and was hoping to link to it, but it looks like it's now paid content. I saw that our guides are under construction and I'd be happy to write a guide on loops with an emphasis on pythonicness.

I don't know that linking to the official Python docs is a digestible resource for some learners, especially those who don't have experience in other languages, so I'm very much open to alternatives for those.

### DIFF
--- a/bot/resources/tags/range-len.md
+++ b/bot/resources/tags/range-len.md
@@ -1,19 +1,15 @@
-Iterating over `range(len(...))` is a common approach to accessing each item
-in an ordered collection.
+Iterating over `range(len(...))` is a common approach to accessing each item in an ordered collection.
 
 ```py
 for i in range(len(my_list)):
     do_something(my_list[i])
 ```
 
-The pythonic syntax is much simpler, and is
-guaranteed to produce elements in the same order:
+The pythonic syntax is much simpler, and is guaranteed to produce elements in the same order:
 
 ```py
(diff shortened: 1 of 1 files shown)


swfarnsworth:
@kwzrd I see what you're saying about displaying the REPL output. I think people who use `range(len(...))` are familiar with iteration in general (and probably use it that way in a C-style language); if the concept of iteration itself is where they're struggling, they probably don't know why they have `range(len(...))` in their code, and a tailored approach would be needed to teach them that concept. If we find that the current message isn't effective, we could see if a REPL display is what would trigger the aha moment for them.

Den4200:
Looks good! However, since we do have `!zip` and `!enumerate` tags, maybe they could be mentioned in this tag?

swfarnsworth:
@Den4200 Do we have any other messages that suggest that you run a command to get another message? The idea sounds a bit weird to me, but our messages would probably be more informative for our intended audience than the Python docs.

Den4200:
After some discussion in `#organisation`, it looks like the tag is ready to go.

PR #1039, opened 2020-07-12
Allow role mentions in specific areas

### LINKED ISSUE #1038: Allow role mentions in specific areas
## Description
Since we do not allow the bot to mention roles anymore (b1c0177) except for moderators+ (36330b1) after the [incident](https://seph.club/who-pinged/?utm_medium=social&utm_source=discord&utm_campaign=whopinged), this has created a new issue where the bot cannot ping roles where it is actually needed in certain places. This needs to be fixed.

Some of these places include:
- Mod alerts (pings `@everyone`)
- Mod log (pings `@everyone`)
- The help channel system (pings `@Helpers`)
- Syncers (pings `@PyDis Core Developers`)

There may be some other areas I missed, so please look over the codebase carefully and explicitly allow role mentions where needed.

## Implementation
We can use [`discord.AllowedMentions`](https://discordpy.readthedocs.io/en/latest/api.html?highlight=allowedmentions#discord.AllowedMentions) when sending messages where non-whitelisted role mentions are needed, which will override the global setting. This object can be passed using the [`allowed_mentions`](https://discordpy.readthedocs.io/en/latest/api.html?highlight=allowedmentions#discord.abc.Messageable.send) kwarg.

lemonsaurus:
This PR makes two changes:
- Make the bot ping `@Moderators` instead of `@everyone` in #mod-alerts and #mod-logs
- Allow role pings in the Syncers and in help_channels.py by adding the roles to allowed_mentions.

This is necessary for pings to come through after we changed to Discord.py 1.4.0a

Closes #1038.

### DIFF
--- a/config-default.yml
+++ b/config-default.yml
@@ -269,7 +269,7 @@ filter:
     notify_user_domains:     false
 
     # Filter configuration
-    ping_moderators:             true  # Ping @everyone when we send a mod-alert?
+    ping_moderators:             true
     offensive_msg_delete_days: 7     # How many days before deleting an offensive message?
 
     guild_invite_whitelist:
--- a/config-default.yml
+++ b/config-default.yml
@@ -269,7 +269,7 @@ filter:
     notify_user_domains:     false
 
     # Filter configuration
-    ping_everyone:             true  # Ping @everyone when we send a mod-alert?
+    ping_moderators:             true  # Ping @everyone when we send a mod-alert?
     offensive_msg_delete_days: 7     # How many days before deleting an offensive message?
 
     guild_invite_whitelist:
@@ -428,7 +428,7 @@ urls:
 anti_spam:
     # Clean messages that violate a rule.
     clean_offending: true
-    ping_everyone: true
+    ping_moderators: true
 
     punishment:
         role_id: *MUTED_ROLE
--- a/config-default.yml
+++ b/config-default.yml
@@ -269,7 +269,7 @@ filter:
     notify_user_domains:     false
 
     # Filter configuration
-    ping_moderators:             true
+    ping_everyone:             true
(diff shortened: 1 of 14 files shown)


config-default.yml
```diff
@@ -269,7 +269,7 @@ filter:
     notify_user_domains:     false
 
     # Filter configuration
-    ping_everyone:             true  # Ping @everyone when we send a mod-alert?
+    ping_moderators:             true  # Ping @everyone when we send a mod-alert?
```
SebastiaanZ, line 272: The comment does not reflect the actual value; the comment still mentions `@everyone`.

```suggestion
    ping_moderators:             true  # Ping @Moderators when we send a mod-alert?
```

Or we could simply remove this unnecessary comment to prevent these kind of "left behind" comments during refactor

```suggestion
    ping_moderators:             true
```

vivax3794, line 272: I guess the comment should be changed as well.
```suggestion
    ping_moderators:             true  # Ping @moderators when we send a mod-alert?
```

lemonsaurus:
@vivax3794 @SebastiaanZ Good catch. I've removed that comment since it violates the DRY principle.

lemonsaurus:
Consensus among core devs appears to be that we should continue to use `@everyone` pings.

The reason for this is that it allows the #mod-alerts channel, which typically has a low signal-to-noise ratio, to be muted. The trick to it is to surpress `@here` and `@everyone` for the entire server and then mute the channel. Changing it to a `@Moderators` ping makes this trick impossible.

Ideally, I think Discord should address this in the UI, but I suppose I don't really care whether we use one or the other, so I'll change it back and add the `@everyone` to the `allowed_mentions` for those sends. Don't merge this yet.

lemonsaurus:
This should be good now, I think.

PR #1041, opened 2020-07-12
Relay all DMs sent to the bot to #dm_log

### LINKED ISSUE #667: Relay all bot DM's to log channel
# Abstract

Currently we're only able to read messages that were sent as a DM to our bot that trip our filters. We should relay **all** messages that are sent to the bot, and who sent it.

Depends on [#174](https://github.com/python-discord/organisation/issues/174) being actioned.

# Specification

This should probably be discussed, but I propose we make it work the same as `#big-brother` does. I'm not certain on the specifics, so if you are, please edit this issue. Essentially I believe it uses webhooks.

# Motivation

ELA thought it'd be interesting. Everyone agreed with that.

lemonsaurus:
This PR adds a feature which will relay DMs sent to the bot through a simple webhook relay system. It also provides a mechanism for responding to these DMs, should you need to do so.

The mechanism for responding will by default respond to the last person who DMd the bot, but can optionally take a member parameter to reply to a specific member.

image

## Refactoring Duck Pond and Python News
To do this in a DRY way, I first refactored the send_webhook helper used in Duck Pond into a generic util, which could be used for both Duck Pond and the DM Relay cog.

As it turns out, we had a similar mechanism in use over in the Python News handler as well, so I had to refactor that as well to use this util.

## Fixing the DuckPond tests
Holy shit, DuckPond has ten million tests, and half of them broke when I made the first change. So I had to fix some of those broken tests.

## Add the DMRelay cog
Next, we add a cog that will just relay anything that is a DM sent by a human to the right channel. The cog also adds the command to allow you to respond. Easy peasy.

Closes #667.


bot/utils/webhooks.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+async def send_webhook(
+        webhook: discord.Webhook,
+        content: Optional[str] = None,
+        username: Optional[str] = None,
+        avatar_url: Optional[str] = None,
+        embed: Optional[Embed] = None,
+        wait: Optional[bool] = False
+) -> discord.Message:
```
MarkKoz, line 19: I feel like this may as well just take `**kwargs` at this point.

bot/cogs/dm_relay.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.webhook_id = constants.Webhooks.dm_log
+        self.webhook = None
+        self.bot.loop.create_task(self.fetch_webhook())
+        self.last_dm_user = None
+
+    @commands.command(aliases=("reply",))
+    async def send_dm(self, ctx: commands.Context, member: Optional[discord.Member], *, message: str) -> None:
```
MarkKoz, line 28: Are you intentionally allowing everyone to use this?

bot/cogs/dm_relay.py
```diff
@@ -0,0 +1,99 @@
...
+            await ctx.message.add_reaction("❌")
+
+    async def fetch_webhook(self) -> None:
+        """Fetches the webhook object, so we can post to it."""
+        await self.bot.wait_until_guild_available()
+
+        try:
+            self.webhook = await self.bot.fetch_webhook(self.webhook_id)
+        except discord.HTTPException:
+            log.exception(f"Failed to fetch webhook with id `{self.webhook_id}`")
```
MarkKoz, line 59: At this point, `on_message` should be disabled somehow. Otherwise, it will raise another exception every time it tries to use `None` as the webhook.

bot/cogs/dm_relay.py
```diff
@@ -0,0 +1,99 @@
...
+            return
+
+        clean_content = message.clean_content
+        if clean_content:
+            await send_webhook(
+                webhook=self.webhook,
+                content=message.clean_content,
+                username=message.author.display_name,
+                avatar_url=message.author.avatar_url
+            )
```
Numerlor, line 75: The `clean_content` vars here and in duck pond feel unnecessary to me, or should at least be passed to the content kwarg

bot/utils/webhooks.py
lemonsaurus: hmm, maybe. but I don't feel like that would be much of a readability improvement.

bot/cogs/dm_relay.py
lemonsaurus: haha, absolutely not. good point.

bot/cogs/dm_relay.py
```diff
@@ -0,0 +1,104 @@
...
+            await member.send(message)
+            await ctx.message.add_reaction("✅")
+            return
+        elif self.last_dm_user:
+            await self.last_dm_user.send(message)
+            await ctx.message.add_reaction("✅")
+            return
+        else:
+            log.debug("Unable to send a DM to the user.")
+            await ctx.message.add_reaction("❌")
```
MarkKoz, line 52: These may raise exceptions if the user has DMs disabled. Let's catch exceptions and add an ❌ reaction.

lemonsaurus: Solved in the latest commit. I also added caching of user IDs so we can reply to the latest DM sender through restarts.

lemonsaurus:
Hmm, good question. We might want to contain that in the #dm-log channel so that all the dm traffic (both in and out) is in one place. I'll make that change.

Numerlor:
Should the cog make some attempts to mitigate issues when the !reply command is used for the last user, but an another DM comes in while the invokation message is being sent and processed?

lemonsaurus:
> Should the cog make some attempts to mitigate issues when the !reply command is used for the last user, but an another DM comes in while the invokation message is being sent and processed?

I did consider this, and I wonder how much of a problem it's really going to be.

How would you suggest we could go about mitigating this anyway?

lemonsaurus:
the easiest approach would be to just remove the whole last user feature altogether and always require a member to be specified. More technical approach would be some sort of locking mechanism, but I'm not really keen to complicate this. This feature is absolutely not meant to be used for anything serious, so it feels like even if the DM gets delivered to the wrong person, it shouldn't really do any damage.

Numerlor:
Perhaps some sort of a delay on the setting the user would be enough to give time for the person writing it to notice the change without influencing other things.
 Are there any stats on how many dms the bot gets? If it's something relatively rare and the command is supposed to be used mostly for some shenanigans then the current approach ought to be fine. But if it's also supposed to be used sparingly then making the user a required argument shouldn't be that much of a problem for the people replying.

lemonsaurus:
Yeah, I'm leaning towards just simplifying and requiring the user to prevent this kind of race condition issue.

We don't have any such stats from what I'm aware. I'm curious myself. Maybe this PR should add stats for that, too.

lemonsaurus:
oh shit, that last commit was supposed to be two commits, but anyway I've made the changes requested by @kosayoda and @Numerlor now:
- It is no longer possible to use `.reply` outside the `#dm-log` channel.
- `.reply` now requires a user to be specified for each call. This is to prevent situations where you would reply to the wrong person. It also has the added benefit of removing a bit of complexity (to do with caching)

PR #1045, opened 2020-07-14
Invoke fuzzywuzzy's processor before matching

### LINKED ISSUE #980: Providing a single apostrophe-like character behind `!help` raises warning
Issuing an help command like `!help '` (a single single quote character as the help query) results in the following warning:

```py
2020-06-02 21:52:38 | root | WARNING | Applied processor reduces input query to empty string, all comparisons will have score 0. [Query: ''']
```

The same goes for at least  the queries:

```
[Query: '´']
[Query: '"']
[Query: '`']
```

The warning message hints at that such a "query" will be reduces to an empty string and, in turn, to a comparison score of `0`. The latter suggests that this warning is probably given by something in our fuzzy matching procedure. (But I haven't taken steps to confirm this yet.)

I think the best solution is to just show the normal help embed for such queries, but what the most robust way to do this is needs to be determined by the one who takes on this issue.

Sentry Issue: [BOT-5W](https://sentry.io/organizations/python-discord/issues/1705336367/?referrer=github_integration)

kosayoda:
Fixes #980 

Trying to match a string with only non-alphanumeric characters results in a warning by fuzzywuzzy.
Processing the string before matching lets us avoid the warning, which uses the root logger and thus isn't supressible.

### DIFF
--- a/bot/cogs/help.py
+++ b/bot/cogs/help.py
@@ -150,8 +150,8 @@ class CustomHelpCommand(HelpCommand):
 
         # Run fuzzywuzzy's processor beforehand, and avoid matching if processed string is empty
         # This avoids fuzzywuzzy from raising a warning on inputs with only non-alphanumeric characters
-        if full_process(string):
-            result = process.extractBests(string, choices, scorer=fuzz.ratio, score_cutoff=60, processor=None)
+        if (processed := full_process(string)):
+            result = process.extractBests(processed, choices, scorer=fuzz.ratio, score_cutoff=60, processor=None)
         else:
             result = []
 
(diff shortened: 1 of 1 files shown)


bot/cogs/help.py
```diff
@@ -146,7 +147,13 @@ async def command_not_found(self, string: str) -> "HelpQueryNotFound":
         Will return an instance of the `HelpQueryNotFound` exception with the error message and possible matches.
         """
         choices = await self.get_all_help_choices()
-        result = process.extractBests(string, choices, scorer=fuzz.ratio, score_cutoff=60)
+
+        # Run fuzzywuzzy's processor beforehand, and avoid matching if processed string is empty
+        # This avoids fuzzywuzzy from raising a warning on inputs with only non-alphanumeric characters
+        if full_process(string):
```
MarkKoz, line 153: This returns a new string. You should pass it to `extractBests`. Good opportunity for a walrus here.

kosayoda: Good catch.

PR #1046, opened 2020-07-14
Update or-gotcha.md

ItsCinnabar:
Adjust description and include link to docs

### DIFF
--- a/bot/resources/tags/or-gotcha.md
+++ b/bot/resources/tags/or-gotcha.md
@@ -3,7 +3,7 @@ When checking if something is equal to one thing or another, you might think tha
 if favorite_fruit == 'grapefruit' or 'lemon':
     print("That's a weird favorite fruit to have.")
 ```
-While this makes sense in English, it may not behave the way you would expect. In Python, you should have _complete instructions on both sides of the logical operator_.
+While this makes sense in English, it may not behave the way you would expect. [In Python, you should have _complete instructions on both sides of the logical operator_.](https://docs.python.org/3/reference/expressions.html#boolean-operations)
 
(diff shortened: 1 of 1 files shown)


Numerlor:
Discord mobile markdown is a bit messed up and in text links, the formatting gets ignored.

Would it make sense to only use "In Python" as the link portion?

SebastiaanZ:
I can't replicate the formatting issue on android, which means I can't experiment with it. I'm wondering if this would give better results:

```md
In Python, you should have _[complete instructions on both sides of the logical operator](https://docs.python.org/3/reference/expressions.html#boolean-operations)_.
```

bot/resources/tags/or-gotcha.md
```diff
@@ -3,7 +3,7 @@ When checking if something is equal to one thing or another, you might think tha
 if favorite_fruit == 'grapefruit' or 'lemon':
     print("That's a weird favorite fruit to have.")
 ```
-After all, that's how you would normally phrase it in plain English. In Python, however, you have to have _complete instructions on both sides of the logical operator_.
+While this makes sense in English, it may not behave the way you would expect. [In Python, you should have _complete instructions on both sides of the logical operator_.](https://docs.python.org/3/reference/expressions.html#boolean-operations)
```
SebastiaanZ, line 6: Apparently, using formatting markdown inside of links causes issues on Discord's iOS client. What does work is wrapping the italics markdown around the link markdown:

```suggestion
While this makes sense in English, it may not behave the way you would expect. In Python, you should have _[complete instructions on both sides of the logical operator](https://docs.python.org/3/reference/expressions.html#boolean-operations)_.
```

PR #1049, opened 2020-07-14
FIXED YOUR STUPID BOT

PythonLover99:
it's way more efficient and it's 1000x better than the OG bot is.


Senjan21:
This is not a place to troll. Please don't waste our time as well as your own

PR #1054, opened 2020-07-15
DM relay - minor fixes

lemonsaurus:
This hotfix PR makes three minor changes - We restore user caching, add the user ID to the author name, and add some stats.

## User caching
We previously had a feature to allow a `!reply` without specifying the target user. This was removed because we were afraid that the number of DMs would be so great that the risk of replying to the wrong user was high.

As it turns out, this traffic is much lower than anticipated and the risk of replying to the wrong user appears to be low. That, coupled with the fact that we shouldn't really be using `!reply` for anything serious, has made me decide to restore this functionality.

## Adding the user ID to the username
When we get a DM, there's no actual user mention since it's relayed via a webhook. That means that a moderator would have to search for that user and figure out what their ID is in order to do stuff like infraction searches.

To make this easier, I'm adding the user ID to the author name:
image

## Stats!
We're now incrementing `dm_relay.dm_sent` whenever we send a DM, and `dm_relay.dm_received` whenever we receive one. This should allow us to track how many DMs the bot sends and receives.

### DIFF
--- a/bot/cogs/dm_relay.py
+++ b/bot/cogs/dm_relay.py
@@ -41,23 +41,24 @@ class DMRelay(Cog):
 
         NOTE: This feature will be removed if it is overused.
         """
-        user_id = await self.dm_cache.get("last_user")
-        last_dm_user = ctx.guild.get_member(user_id) if user_id else None
+        if not member:
+            user_id = await self.dm_cache.get("last_user")
+            member = ctx.guild.get_member(user_id) if user_id else None
+
+        # If we still don't have a Member at this point, give up
+        if not member:
+            log.debug("This bot has never gotten a DM, or the RedisCache has been cleared.")
+            await ctx.message.add_reaction("❌")
+            return
 
         try:
-            if member:
-                await member.send(message)
-                await ctx.message.add_reaction("✅")
-                self.bot.stats.incr("dm_relay.dm_sent")
-                return
-            elif last_dm_user:
-                await last_dm_user.send(message)
-                await ctx.message.add_reaction("✅")
-                self.bot.stats.incr("dm_relay.dm_sent")
-                return
-            else:
-                log.debug("This bot has never gotten a DM, or the RedisCache has been cleared.")
-                await ctx.message.add_reaction("❌")
+            await member.send(message)
+        except discord.errors.Forbidden:
+            log.debug("User has disabled DMs.")
+            await ctx.message.add_reaction("❌")
+        else:
+            await ctx.message.add_reaction("✅")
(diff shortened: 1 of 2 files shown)


bot/cogs/dm_relay.py
```diff
@@ -17,30 +19,45 @@
...
+                self.bot.stats.incr("dm_relay.dm_sent")
+                return
+            elif last_dm_user:
+                await last_dm_user.send(message)
+                await ctx.message.add_reaction("✅")
+                self.bot.stats.incr("dm_relay.dm_sent")
+                return
+            else:
+                log.debug("This bot has never gotten a DM, or the RedisCache has been cleared.")
+                await ctx.message.add_reaction("❌")
```
SebastiaanZ, line 60: Why don't we do something like:

```suggestion
        if not member:
            user_id = await self.dm_cache.get("last_user")
            member = ctx.guild.get_member(user_id) if user_id else None

        # If we still don't have a Member at this point, give up
        if not member:
            log.debug("This bot has never gotten a DM, or the RedisCache has been cleared.")
            await ctx.message.add_reaction("❌")
            return

        try:
            await member.send(message)
        except discord.errors.Forbidden:
            log.debug("User has disabled DMs.")
            await ctx.message.add_reaction("❌")
        else:
            await ctx.message.add_reaction("✅")
            self.bot.stats.incr("dm_relay.dm_sent")
```

bot/cogs/dm_relay.py
```diff
@@ -17,30 +19,45 @@
...
+
     def __init__(self, bot: Bot):
         self.bot = bot
         self.webhook_id = constants.Webhooks.dm_log
         self.webhook = None
         self.bot.loop.create_task(self.fetch_webhook())
 
     @commands.command(aliases=("reply",))
-    async def send_dm(self, ctx: commands.Context, member: discord.Member, *, message: str) -> None:
+    async def send_dm(self, ctx: commands.Context, member: Optional[discord.Member], *, message: str) -> None:
```
MarkKoz, line 32: This converter will convert by name and nickname too. So if someone happens to have a common word as a name, you'll end up DMing them.

lemonsaurus: That's a very good point. What I really want is a converter that only does mentions and userIDs. Maybe I'll need to make a custom one?

bot/cogs/dm_relay.py
lemonsaurus: Sure.

bot/cogs/dm_relay.py
lemonsaurus: Fixed in 403572b

Issue #1024, opened 2020-07-03
Purge messages up to given message


wookie184:
**Problem:**
Currently the purge/clean command takes the number of messages to clean as the only way of specifying how far to clean, this can cause some difficulties when trying to clean lots of messages as you either have to guess roughly how many messages were sent or count them manually, and also makes it easy to go too far and delete messages you didn't want to.

**Proposed Solution:**
We could allow a message ID or link to be given as an argument instead of the number of messages. The bot would then delete all messages up to that message. This would speed up the process of removing messages and reduce the possibility of human error.

I'm not sure what the best way to modify the command to accept the ID as an argument would be so that would need to be discussed.

Senjan21:
I can work on that issue if that's possible.

Senjan21:
Implemented in #1055

PR #1055, opened 2020-07-15
Added command&system to purge all messages up to given message

Senjan21:
This PR aims to close issue #1024 by implementing purging system that purges messages up until certain message specified by moderator.
The command input supports:
- Message ID (Will only work if message is in the channel in which command was invoked)
- ChannelID-MessageID (obtainable via shift-clicking on “Copy ID”)
- jump url to the message

However, the system of deletion is completely different from the original one, due to the fact that `TextChannel.purge` wouldn't otherwise support that system well enough.
I'd like to hear opinions on the PR.

### DIFF
--- a/bot/cogs/clean.py
+++ b/bot/cogs/clean.py
@@ -45,6 +45,7 @@ class Clean(Cog):
         bots_only: bool = False,
         user: User = None,
         regex: Optional[str] = None,
+        until_message: Optional[Message] = None,
     ) -> None:
         """A helper function that does the actual message cleaning."""
         def predicate_bots_only(message: Message) -> bool:
@@ -129,6 +130,25 @@ class Clean(Cog):
                 if not self.cleaning:
                     return
 
+                # If we are looking for specific message.
+                if until_message:
+                    # Since we will be using `delete_messages` method
+                    # of a TextChannel
+                    # and we need message objects to use it
+                    # as well as to send logs
+                    # we will start appending messages here
+                    # instead adding them from purge.
+                    messages.append(message)
+                    # we could use ID's here however
+                    # in case if the message we are looking for
+                    # gets deleted, we won't have a way to figure that out
+                    # thus checking for datetime should be more reliable
+                    if message.created_at <= until_message.created_at:
+                        # means we have found the message until which
+                        # we were supposed to be deleting.
+                        message_ids.append(message.id)
+                        break
+
                 # If the message passes predicate, let's save it.
                 if predicate is None or predicate(message):
                     message_ids.append(message.id)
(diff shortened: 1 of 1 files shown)


bot/cogs/clean.py
```diff
@@ -138,7 +152,14 @@ def predicate_regex(message: Message) -> bool:
...
         for channel in channels:
-            messages += await channel.purge(limit=amount, check=predicate)
+            if until_message:
+                for i in range(0, len(messages), 100):
+                    # while purge automatically handles the amount of messages
+                    # delete_messages only allows for up to 100 messages at once
+                    # thus we need to paginate the amount to always be <= 100
+                    await channel.delete_messages(messages[i:i + 100])
+            else:
+                messages += await channel.purge(limit=amount, check=predicate)
```
MarkKoz, line 162: Since you've essentially written all the code needed, can we get rid of `purge`? It redundantly iterates message history. We already have message objects from our own history search, so let's use those. That being said, one strange thing `purge` does is use a single delete strategy if messages are older than 2 weeks. I don't know if that's important.

Senjan21: Reason why it does single deletes is because you can't delete messages older than 14 days via bulk delete endpoint, I didn't include that edge case in the code as (at least to me) it appears unlikely for moderators to attempt and purge messages older than that.
That being said attempts of deleting the messages older than 14 days via `delete_messages` will raise following error:
```python
discord.errors.HTTPException: 400 Bad Request (error code: 50034): You can only bulk delete messages that are under 14 days old.
```

MarkKoz: Do you think we can do away with `purge` though?

Senjan21: I don't think `purge` is necessary considering we iterate the `history` either way,  I am all up to rewriting current system so it doesn't use `purge` in separate PR as it isn't exactly the point of this PR

bot/cogs/clean.py
```diff
@@ -129,6 +130,19 @@ def predicate_regex(message: Message) -> bool:
...
+                    # Since we will be using `delete_messages` method of a TextChannel and we need message objects to
+                    # use it as well as to send logs we will start appending messages here instead adding them from
+                    # purge.
+                    messages.append(message)
+                    # we could use ID's here however in case if the message we are looking for gets deleted,
+                    # we won't have a way to figure that out thus checking for datetime should be more reliable
+                    if message.created_at <= until_message.created_at:
+                        # means we have found the message until which we were supposed to be deleting.
+                        message_ids.append(message.id)
+                        break
```
MarkKoz, line 144: If the current message is older than `until_message`, then the current message should not be appended and deleted, as it's too far back. Something like this, but maybe you can come up with something more elegant:

```py
if message.created_at < until_message.created_at:
    break

messages.append(message)

if message.id == until_message.id:
    message_ids.append(message.id)
    break
```

PR #1058, opened 2020-07-19
FilterLists: Manage whitelisting and blacklisting via the bot

lemonsaurus:
Up until now, we have managed various whitelists and blacklists via adding constants to our `default-config.yml` file. However, it has often been suggested that we should instead be managing these via a bot command, and storing them in the database.

Related issues:
- Closes https://github.com/python-discord/bot/issues/639
- https://github.com/python-discord/bot/issues/126
- https://github.com/python-discord/site/issues/305 (pull request open at https://github.com/python-discord/site/pull/371)
- https://github.com/python-discord/organisation/issues/261

**NOTE: This pull request must be merged _after_ https://github.com/python-discord/site/issues/305!**

This is a big pull request, best reviewed commit-by-commit, which makes a number of changes to filtering, antimalware, constants, and implements a new cog for managing blacklists and whitelists.

## Cache blacklists and whitelists, to minimize API requests.
We don't want to be making API request for every single message, but we need to scan all messages with our filters and see if they contain blacklisted words, or server invites. The solution is caching. This pull requests makes the bot cache all the AllowDenyList data from the site when the bot starts, and then updates this cache whenever the data changes.

## Changes to the exception handler
Currently, we're not returning the error message whenever a BadArgument is raised from a command. This was done because some BadArguments would contain the input data, and we would then post that input data directly into the response from the bot when the exception was raised. This allowed users to exploit the `!raw` command with something like `!raw @Developers` in order to have the bot ping 65 000 users in our guild and create absolute chaos.

So, to prevent that, we just made BadArgument not return the exception message at all. But this isn't a great solution, because the exception message often contains valuable information that our users need in order to figure out how to pass a valid argument. For the cog I was making for the whitelist and blacklist, I felt I needed to provide this kind of information to the users, particularly to tell them which kinds of blacklists and whitelists we support.

This PR changes the exception handler so that all types of exceptions handled there will return an embed containing the full exception message. This strikes a nice balance between usability and security, since pings do not work from inside embeds.
image

## New cog: FilterLists
The main attraction of this pull request is of course the new FilterLists cog. This cog allows us to add or remove items in the whitelist or the blacklist, and to show all the items that are currently there.

**Adding file formats:**
image

**Adding guild invites:**
image

**Removing stuff:**
image

**Listing guild invites:**
image

**Listing file formats:**
image

## Changes to AntiMalWare and Filtering
These two existing cogs have gotten some kaizen, and have been refactored to utilize the blacklists and whitelists in the cache instead of relying on constants.

The constants they used before have been removed.

We no longer differentiate between `word_watchlist` and `token_watchlist` - if you need word boundaries before and after your regex, you just add a `\b`. This new system, simply called `filter_tokens`, allows you then to input any regex you want with no implicit magic.

## New converters
In order to validate list types, I've created a converter called `ValidAllowDenyListType`, which will make an API request to figure out exactly which kinds of blacklists and whitelists we support. This means that the available blacklists and whitelists is managed in realtime from a single location - an enum inside the model definition on the site. This is the DRYest way I could think of to do this.

In order to validate guild invites, I've created a converter called `ValidDiscordServerInvite`, which does the following:
- Checks the string against a compiled regex pattern which contains all legal variations of guild invites.
- If it gets a match, it uses the Discord API to look up the invite and get information about the guild. The converter then returns this information as a dictionary, including stuff like the guild name. **We always attach the guild name as a comment when adding a new guild invite to the blacklist or whitelist, so that you can see which Guild ID's belong to which guilds.**
- If the string contains a Guild ID, it raises a BadArgument saying Guild ID's can't be resolved.
- Otherwise, it raises a BadArgument saying the invite is invalid.

## New guild filtering logic
As discussed in https://github.com/python-discord/organisation/issues/261, we wanted a slightly more nuanced approach to filtering guild invites. This PR makes these changes, so that the following is true:
- A blacklisted guild invite will **always** trigger an alert
- A whitelisted guild invite will **never** trigger an alert
- Partnered and Verified guild invites only trigger an alert if they are blacklisted.
- All other guild invites will trigger an alert.


bot/utils/regex.py
```diff
@@ -0,0 +1,12 @@
+import re
+
+INVITE_RE = re.compile(
```
MarkKoz, line 3: Liking the initiative to move more stuff to util files! There are certainly a few more expressions I can think of putting in here later.

bot/cogs/allow_deny_lists.py
```diff
@@ -31,6 +31,24 @@ async def _add_data(
         """Add an item to an allow or denylist."""
         allow_type = "whitelist" if allowed else "blacklist"
 
+        # If this is a server invite, we gotta validate it.
+        if list_type == "GUILD_INVITE":
+            log.trace(f"{content} is a guild invite, attempting to validate.")
+            validator = ValidDiscordServerInvite()
```
MarkKoz, line 37: If I understand this correctly, you only allow invite links, but you end up storing the guild ID in the DB? Why didn't you accept guild IDs directly? Is it because you wouldn't be able to auto-generate a comment?

bot/cogs/filtering.py
```diff
@@ -478,7 +468,7 @@ async def _has_invites(self, text: str) -> Union[dict, bool]:
                 return True
 
             guild_id = guild.get("id")
```
MarkKoz, line 470: Would it make sense to use the invite converter here too? It's pretty similar code besides the `with_counts`.

bot/cogs/allow_deny_lists.py
```diff
@@ -95,9 +95,21 @@ async def _delete_data(self, ctx: Context, allowed: bool, list_type: ValidAllowD
...
+        # If this is a server invite, we need to convert it.
+        if list_type == "GUILD_INVITE" and not id_converter._get_id_match(content):
+            log.trace(f"{content} is a guild invite, attempting to validate.")
+            validator = ValidDiscordServerInvite()
+            guild_data = await validator.convert(ctx, content)
+
+            # If we make it this far without raising a BadArgument, the invite is
+            # valid. Let's convert the content to an ID.
+            log.trace(f"{content} validated as server invite. Converting to ID.")
+            content = guild_data.get("id")
```
MarkKoz: This is pretty similar to the `_add_data` code. Does it make sense to refactor it into a separate function?

bot/cogs/filtering.py
```diff
@@ -468,9 +468,21 @@ async def _has_invites(self, text: str) -> Union[dict, bool]:
...
-            guild_invite_whitelist = self._get_allowlist_items(True, "guild_invite")
+            guild_invite_whitelist = self._get_allowlist_items("guild_invite", allowed=True)
+            guild_invite_blacklist = self._get_allowlist_items("guild_invite", allowed=False)
 
-            if guild_id not in guild_invite_whitelist:
+            # Is this invite allowed?
+            guild_partnered_or_verified = (
+                'PARTNERED' in guild.get("features")
+                or 'VERIFIED' in guild.get("features")
```
MarkKoz, line 477: Is using `get` necessary here? If it is, keep in mind `x in None` will raise an error, so you'd need a more suitable default value.

bot/cogs/filtering.py
```diff
@@ -99,9 +99,9 @@ def __init__(self, bot: Bot):
 
         self.bot.loop.create_task(self.reschedule_offensive_msg_deletion())
 
-    def _get_allowlist_items(self, allow: bool, list_type: str, compiled: Optional[bool] = False) -> list:
+    def _get_allowlist_items(self, list_type: str, *, allowed: bool, compiled: Optional[bool] = False) -> list:
```
MarkKoz, line 102: Good call to make these keyword-only arguments.

bot/cogs/allow_deny_lists.py
```diff
@@ -95,9 +95,21 @@ async def _delete_data(self, ctx: Context, allowed: bool, list_type: ValidAllowD
         """Remove an item from an allow or denylist."""
         item = None
         allow_type = "whitelist" if allowed else "blacklist"
+        id_converter = IDConverter()
 
-        log.trace(f"Trying to delete the {content} item from the {list_type} {allow_type}")
+        # If this is a server invite, we need to convert it.
+        if list_type == "GUILD_INVITE" and not id_converter._get_id_match(content):
```
MarkKoz, line 101: To be clear, `_get_id_match` is checked because unlike `_add_data`,  you also want to permit deleting directly by ID?

bot/bot.py
```diff
@@ -49,6 +50,22 @@ def __init__(self, *args, **kwargs):
...
+        for item in full_cache:
+            type_ = item.get("type")
+            allowed = item.get("allowed")
+            metadata = {
+                "content": item.get("content"),
+                "comment": item.get("comment"),
+                "id": item.get("id"),
+                "created_at": item.get("created_at"),
+                "updated_at": item.get("updated_at"),
+            }
```
MarkKoz, line 66: Can the API really return an empty dict for records? Are all these `get`s necessary?

bot/cogs/antimalware.py
```diff
@@ -38,6 +38,16 @@ class AntiMalware(Cog):
     def __init__(self, bot: Bot):
         self.bot = bot
 
+    def _get_whitelisted_file_formats(self) -> list:
+        """Get the file formats currently on the whitelist."""
+        return [item.get('content') for item in self.bot.allow_deny_list_cache['file_format.True']]
```
MarkKoz, line 43: It's slightly unfortunate you have to use a comprehension every time you need a list of contents. Can the data be restructured to avoid this? In practice, you only need the content and the comments, so caching them in a dict with the former as keys and latter as values may work. Admittedly, it does sound a bit strange, but the contents are guaranteed to be unique already.

bot/cogs/filtering.py
```diff
@@ -125,6 +99,23 @@ def __init__(self, bot: Bot):
 
         self.bot.loop.create_task(self.reschedule_offensive_msg_deletion())
 
+    def _get_allowlist_items(self, list_type: str, *, allowed: bool, compiled: Optional[bool] = False) -> list:
+        """Fetch items from the allow_deny_list_cache."""
+        items = self.bot.allow_deny_list_cache.get(f"{list_type.upper()}.{allowed}", [])
+
+        if compiled:
+            return [re.compile(fr'{item.get("content")}', flags=re.IGNORECASE) for item in items]
```
MarkKoz, line 107: Would've been a bit nicer to cache the compiled regular expressions like they were before, but I think doing that now would get kinda ugly. I don't think compiling them serves much benefit if you discard them after every use.

bot/converters.py
```diff
@@ -34,6 +34,32 @@ def converter(arg: str) -> str:
...
+
+    Raises `BadArgument` if the argument is not a valid AllowDenyList type, and simply
+    passes through the given argument otherwise.
+    """
+
+    async def convert(self, ctx: Context, list_type: str) -> str:
+        """Checks whether the given string is a valid AllowDenyList type."""
+        try:
+            valid_types = await ctx.bot.api_client.get('bot/allow_deny_lists/get_types')
+        except ContentTypeError:
```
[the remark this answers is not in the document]
lemonsaurus: Yeah, you're right.

bot/cogs/error_handler.py
```diff
@@ -20,6 +21,14 @@ class ErrorHandler(Cog):
     def __init__(self, bot: Bot):
         self.bot = bot
 
+    def _get_error_embed(self, title: str, body: str) -> Embed:
+        """Return an embed that contains the exception."""
+        return Embed(
+            title=title,
+            colour=Colours.soft_red,
+            description=body
+        )
```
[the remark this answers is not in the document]
lemonsaurus: So, I've always hated the plain messages, I won't deny there's a little bit of stylistic preference there. And yes, `allowed_mentions` solves this particular case - but it just seems more sane to me to put user input inside an embed, more future-proof. We don't know in which ways our members will be able to exploit our lack of sanitization in the future, but we can assume that sticking their input inside an embed will probably make it harder for them to break anything.

This was discussed briefly in #dev-core, and there seemed to be a small consensus that we should just stick stuff in embeds to be on the safe side, and because it looks more professional.

bot/converters.py
```diff
@@ -50,12 +50,13 @@ async def convert(self, ctx: Context, list_type: str) -> str:
             raise BadArgument("Cannot validate list_type: Unable to fetch valid types from API.")
 
         valid_types = [enum for enum, classname in valid_types]
+        valid_types_lower = [type_.lower() for type_ in valid_types]
         list_type = list_type.upper()
```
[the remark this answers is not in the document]
lemonsaurus: The list comprehension is slightly more involved in the latest version:

```py
valid_types = [enum for enum, classname in valid_types]
valid_types_list = '\n'.join([f"• {type_.lower()}" for type_ in valid_types])
```

and `list_type` has to be upper, to work with the site.

bot/cogs/allow_deny_lists.py
```diff
@@ -0,0 +1,144 @@
...
+        # Insert the item into the cache
+        type_ = item.get("type")
+        allowed = item.get("allowed")
+        metadata = {
+            "content": item.get("content"),
+            "id": item.get("id"),
+            "created_at": item.get("created_at"),
+            "updated_at": item.get("updated_at"),
+        }
+        self.bot.allow_deny_list_cache.setdefault(f"{type_}.{allowed}", []).append(metadata)
```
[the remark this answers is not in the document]
lemonsaurus: yes, that's true. I guess I can make a helper to parse it?

bot/cogs/allow_deny_lists.py
lemonsaurus: Yes. Guild IDs cannot be looked up in order to get any kind of guild information (for data privacy reasons). We can't just store the invites either, because we want to whitelist an entire guild, not just a specific invite.

bot/cogs/allow_deny_lists.py
lemonsaurus: That is correct. There's no reason we can't allow deletion by ID, and in my opinion, rejecting deletion by ID would be worse UX. When you list all the guild invites, you get a list of IDs. It's quite natural to then try to remove one of those IDs, and kind of a pain in the ass to have to figure out what the corresponding invite is.

bot/bot.py
lemonsaurus: I like `get`'s. I never considered whether they were necessary. If you insist, I guess I'll go over them all and do that.

bot/cogs/allow_deny_lists.py
```diff
@@ -0,0 +1,144 @@
...
+from bot.api import ResponseCodeError
+from bot.bot import Bot
+from bot.converters import ValidAllowDenyListType
+from bot.pagination import LinePaginator
+from bot.utils.checks import with_role_check
+
+log = logging.getLogger(__name__)
+
+
+class AllowDenyLists(Cog):
```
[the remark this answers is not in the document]
lemonsaurus: I don't hate it.

bot/cogs/allow_deny_lists.py
```diff
@@ -0,0 +1,144 @@
...
+        allow_type = "whitelist" if allowed else "blacklist"
+
+        # Try to add the item to the database
+        try:
+            item = await self.bot.api_client.post(
+                "bot/allow_deny_lists",
+                json=payload
+            )
+        except ResponseCodeError as e:
+            if e.status == 500:
```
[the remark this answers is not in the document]
lemonsaurus: So would I, but that's not what it does. The constraint violations produce 500s. This might be solvable by making some change on the Django side, but I couldn't figure out what that change would have to be.

bot/cogs/allow_deny_lists.py
```diff
@@ -0,0 +1,144 @@
...
+            "updated_at": item.get("updated_at"),
+        }
+        self.bot.allow_deny_list_cache.setdefault(f"{type_}.{allowed}", []).append(metadata)
+        await ctx.message.add_reaction("✅")
+
+    async def _delete_data(self, ctx: Context, allowed: bool, list_type: ValidAllowDenyListType, content: str) -> None:
+        """Remove an item from an allow or denylist."""
+        item = None
+
+        for allow_list in self.bot.allow_deny_list_cache.get(f"{list_type}.{allowed}", []):
```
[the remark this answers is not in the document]
lemonsaurus: This sounds like a pretty big rewrite for very little added convenience to me. If you insist, I guess I'll do it when I'm back from vacation.

bot/cogs/antimalware.py
lemonsaurus: Hmm, you've got a point. I mean, this is cheap since this list will always be really really short, but we are running this list comprehension for every single message that contains an attachment and isn't sent by a staff member.

I suppose caching may be possible here.

bot/cogs/filtering.py
lemonsaurus: yeah, I should probably refactor this to just use non-compiled expressions. will do.

bot/cogs/filtering.py
lemonsaurus: I experimented with this, but the problem with doing it is that the error messages raised by that converter make no sense in this context. For that reason I decided not to do it.

If you think I should still do it, I need to introduce some sort of complexity to handle that in a way that makes sense in a more generic context.

bot/cogs/filtering.py
lemonsaurus: I guess the answer is, I don't know if it's necessary. I don't know the Discord API well enough to say whether it guarantees that `features` is going to be in the response.

You're absolutely right about the other thing though, refactoring to return empty list.

bot/cogs/allow_deny_lists.py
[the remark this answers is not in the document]
MarkKoz: You'd need to change all `cache.get(..., [])` to `cache[...]`. It doesn't seem like a big rewrite.

bot/cogs/allow_deny_lists.py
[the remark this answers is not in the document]
MarkKoz: Yeah, this needs to be figured out, I think. It doesn't sound normal. If you get 500s it means DRF isn't handling the constraint for some reason.

bot/cogs/allow_deny_lists.py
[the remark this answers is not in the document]
MarkKoz: Would you mind renaming the cog and extension to that then? The current name is precise, but kind of... awkward?

bot/cogs/allow_deny_lists.py
MarkKoz: Sounds reasonable, thanks.

bot/cogs/filtering.py
MarkKoz: That's fine. I've wanted to refactor the filtering cog eventually so I may take care of this when I get to that.

bot/cogs/filtering.py
MarkKoz: You're right. Better to be safe.

bot/cogs/allow_deny_lists.py
MarkKoz: Thanks for confirming.

bot/bot.py
MarkKoz: I suppose that's a good habit. Kind of like using `"${}"` in shell scripts event when it's not necessary so I don't have to think about all the complicated rules for when it is/isn't necessary. I don't insist, you can leave them. I just think the `[]` is preferable when possible since it's more concise.

bot/cogs/antimalware.py
MarkKoz: By "caching", I meant modifying the structure of the cache you've implemented rather than creating a separate cache.

bot/cogs/allow_deny_lists.py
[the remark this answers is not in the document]
MarkKoz: Yes, that makes a lot of sense to me.

bot/cogs/allow_deny_lists.py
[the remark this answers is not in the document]
lemonsaurus: Siteside changes at https://github.com/python-discord/site/pull/371/commits/285c81bc13e996246ad9463951853fc12903d766

Botside changes at https://github.com/python-discord/bot/pull/1058/commits/3d5faa421756fadb42590db92e8fee64578390d4

bot/cogs/allow_deny_lists.py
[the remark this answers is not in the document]
lemonsaurus: Okay, looks like this is happening because I am setting this constraint in a Django vanilla way, not in a DRF way. 

Fixed on the site in https://github.com/python-discord/site/pull/371/commits/0d7c95a80692f15c36b36152034e3bbe54e3c37c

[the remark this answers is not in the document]
lemonsaurus: Fixed on the bot in 1b6be86

bot/cogs/allow_deny_lists.py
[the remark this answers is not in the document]
lemonsaurus: Fixed in 222cdce0b9771b0c121da39b5f38363baf8bce09

bot/cogs/allow_deny_lists.py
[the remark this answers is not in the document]
lemonsaurus: Fixed in a23b273734ddee5fd082bda4fa14aebfff1317ca

bot/bot.py
lemonsaurus: Cleaned up in a23b273734ddee5fd082bda4fa14aebfff1317ca

bot/cogs/allow_deny_lists.py
lemonsaurus: Yep! fixed in a7a3e29ca901b84570e5a1ff1e4c2bcf22b86552

bot/cogs/antimalware.py
lemonsaurus: Sorry, I have no idea how you propose we do that.

bot/cogs/filtering.py
lemonsaurus: Fixed in 4e1609695762524bc707b6a8d39e88c2710cff6b

bot/cogs/filter_lists.py
```diff
@@ -88,16 +88,16 @@ async def _delete_data(self, ctx: Context, allowed: bool, list_type: ValidFilter
 
         # Find the content and delete it.
         log.trace(f"Trying to delete the {content} item from the {list_type} {allow_type}")
-        for allow_list in self.bot.filter_list_cache[f"{list_type}.{allowed}"]:
-            if content == allow_list.get("content"):
-                item = allow_list
+        for allow_list, metadata in self.bot.filter_list_cache[f"{list_type}.{allowed}"].items():
+            if content == allow_list:
+                item = metadata
                 break
```
MarkKoz:
```suggestion
        item = self.bot.filter_list_cache[f"{list_type}.{allowed}"].get(content)
```

bot/cogs/filter_lists.py
```diff
@@ -0,0 +1,250 @@
...
+            content = f".{content}"
+
+        # Try to add the item to the database
+        log.trace(f"Trying to add the {content} item to the {list_type} {allow_type}")
+        payload = {
+            'allowed': allowed,
+            'type': list_type,
+            'content': content,
+            'comment': comment,
+        }
```
kwzrd, line 81: I noticed that you mix single and double quotes. Is there a system to it?

bot/cogs/filter_lists.py
```diff
@@ -0,0 +1,250 @@
...
+    async def allow_add(
+            self,
+            ctx: Context,
+            list_type: ValidFilterListType,
+            content: str,
+            *,
+            comment: Optional[str] = None,
+    ) -> None:
+        """Add an item to the specified allowlist."""
+        await self._add_data(ctx, True, list_type, content, comment)
```
kwzrd, line 209: I think prevailing code-style is to indent at 4 spaces always.

MarkKoz: On one hand, that is true (though there is also some existing code with 8 spaces) and my personal preference. On the other hand, both PEP 8 and our own style guide suggest using 8 spaces.

bot/cogs/filter_lists.py
lemonsaurus: I'm afraid not - I mix them because single quotes are the prevailing style in all the repos at my job, and double quotes are the prevailing style here at pydis. Sometimes my fingers forget where I am.

I'll change these to double quotes.

bot/cogs/filter_lists.py
lemonsaurus: I can change it - this was probably changed by my IDE, not a conscious choice - but I do think both approaches have clear advantages, and that there's no clear winner. This style makes a clearer visual separation between the paramters and the function body, and yeah, PEP8 does suggest it. On the other hand, PEP8 also says "adhere to prevailing style", so no clear winner.

lemonsaurus: ugh. I changed it and I really think it makes it worse. but oh well.

PR #1062, opened 2020-07-21
Truncate charinfo results

### LINKED ISSUE #897: Truncate the charinfo embed to fit within 2048 characters
Sentry Issue: [BOT-3D](https://sentry.io/organizations/python-discord/issues/1621405078/?referrer=github_integration)

```
In embed.description: Must be 2048 or fewer in length.
  File "bot/cogs/utils.py", line 162, in charinfo
    await ctx.send(embed=embed)
  File "discord/http.py", line 225, in request
    raise HTTPException(r, data)

Error executing command invoked by <REDACTED>: !charinfo ﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹﯹ
```

MarkKoz:
Fixes #897 

Just uses the `LinePaginator`. The `raw` field is present and has the exact same value on every page. I'm concerned it may be unclear whether the value is only for the current page or for the entire input (the truth is the latter).

bild

I also updated the command to use the new `send_denial` utility function since the existing code was quite similar to what the helper does. The difference is that `send_denial` has those comedic titles.

bild

bild

### DIFF
--- a/bot/cogs/utils.py
+++ b/bot/cogs/utils.py
@@ -12,6 +12,7 @@ from discord.ext.commands import BadArgument, Cog, Context, command
 from bot.bot import Bot
 from bot.constants import Channels, MODERATION_ROLES, STAFF_ROLES
 from bot.decorators import in_whitelist, with_role
+from bot.pagination import LinePaginator
 from bot.utils import messages
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


MarkKoz:
>  could we use a bigger char limit

> I don't mind it being 10 lines per page, with a 50 character max.

I was tempted but I ended up waiting for someone else to suggest it 👍

bot/cogs/utils.py
```diff
@@ -117,25 +119,18 @@ async def pep_command(self, ctx: Context, pep_number: str) -> None:
...
-            await ctx.send(embed=embed)
-            return
 
-        if len(characters) > 25:
-            embed = Embed(title=f"Too many characters ({len(characters)}/25)")
-            embed.colour = Colour.red()
-            await ctx.send(embed=embed)
-            return
+        if len(characters) > 50:
+            return await messages.send_denial(ctx, f"Too many characters ({len(characters)}/25)")
```
kosayoda, line 133: Missed the change here. We could have a constant for that in the function, but I don't think it helps much given we only use it in two adjacent lines.
```suggestion
            return await messages.send_denial(ctx, f"Too many characters ({len(characters)}/50)")
```

PR #1077, opened 2020-08-02
Take steps to increase verification rate

kwzrd:
This PR closes https://github.com/python-discord/organisation/issues/276. If you cannot see this issue, a summary is below.

# Increase verification rate

The aim is to reduce the amount of dormant users present in the guild without the `@Developers` role. This problem has become more pronounced after we allowed non-verified users to read all public channels (the justification is to allow users lurking from Discovery to see more than the verification channel). As an unfortunate side-effect, we're now seeing large numbers of users who never verify, as they aren't locked into the verification channel.

The steps will will take is:
1. [x] Send a welcome DM to all new-joiners, linking them to the verification channel
2. [x] Create a new `@Unverified` role with restricted read permissions for public channels
3. [x] Assign the `@Unverified` role to all users who fail to verify within `3` (configurable) days after joining
4. [x] Ping the `@Unverified` role once every `28` (configurable) hours
5. [x] Kick unverified members after `30` (configurable) days if they fail to verify

There are a few side quests to complete:
- [x] Remove `@Unverified` from users upon verification
- [x] Disable burst shared anti-spam rule in `#verification`
- [x] Figure out a sane value for `KICK_CONFIRMATION_THRESHOLD`
- [x] Probably send a DM to kicked users
- [x] Collect stats
- [x] Make the `@Unverified` role pingable
- [x] Add a command group for management of the new internal tasks
- [x] Persist whether tasks should start on cog load
 
New config for the test server - add this to your local config files to test this feature.
```
guild:
  roles:
    unverified: 739798569879470172  # Users that have failed to verify
```

The following don't need to be set explicitly because production values will work on the test server:
```
verification:
    unverified_after: ...
    kicked_after: ...
    reminder_frequency: ...
    bot_message_delete_delay: ...
    kick_confirmation_threshold: ...
```

## Showcase
Each new-joiner will receive a new DM right after joining the guild:
image

Every `15` (configurable) minutes, a background task will do the following:
1. Add `@Unverified` to unverified users having joined more than `3` days ago
2. Collect members who have failed to verify after `30` days (does not consider whether they have `@Unverified` or not)

**An unverified user is a user who has no roles apart from `@everyone` and optionally `@Unverified`.**

If this is a significant portion of the guild, core devs will be asked to confirm or abort the operation:
image

We determine this by looking at the % of the guild's population to be kicked. If it exceeds `KICK_CONFIRMATION_THRESHOLD`, a confirmation is necessary. ~~The threshold is currently set to `0` for development purposes, which means that a manual approval is necessary for *any* amount of kicking.~~

If the operation is authorized, or if the amount of members is below the threshold, the bot proceeds with kicking all unverified members. Each kicked member with open DMs receives this message:
image

The `0` is because I changed my config to test this. Sadly there's not much we can do with members with closed DMs, *but* they were getting pung in a public channel for 30 days so they must have known.

After the task finishes, a report is sent to `#mod-log`:
image

The numbers follow the format `n_successes`/`n_attempts`, if these two numbers are not equal it means that some requests have failed (bad status from API). The statuses are logged at info level so we can look at logs to see why it's failing. But in practice it shouldn't fail unless a stray 500 happens or something.

In the verification channel, we re-send this message every `28` (configurable) hours.
image

Justification for `28` is that we get a ping out at reasonable times in all time zones once in a while. It's easily adjustable if necessary, feel free to suggest changes.

Additionally, the `@Unverified` role has some extra permission overrides in channels, making unverified users only see the following channels:
image

## Anti-spam
The `@Unverified` pings may cause surges of activity, so we're disabling the burst shared filter in the verification channel. 

See: b1c800c623f90f46c4ecaff8da2269efcd04ee05

An ideal solution would allow this to be configurable from the outside, but internally we agreed that this is fine, at least for now.

## Stats
With @SebastiaanZ we've agreed to collect statistics by upping counters in the following categories:
- [x] `verification.accepted_on_day_one`: Verified on join day
- [x] `verification.accepted_before_unverified`: Verified before `@Unverified` role given (before day 3)
- [x] `verification.accepted_after_unverified: `Verified after role was given (after day 3, before day 30)
- [x] `verification.kicked`: Kicked (never verified)

## Notes
* Currently more than `10%` of the guild's population satisfies the requirements for a kick. This means that once the bot connects for the first time, we'll get a ping to confirm this. After that, we probably won't be getting pings, unless something weird happens.


kwzrd:
This will need #1090 to be fixed. I'm leaning towards fixing it in a separate PR.

bot/cogs/verification.py
```diff
@@ -32,26 +59,325 @@
...
+
+    @ping_unverified.before_loop
+    async def _before_first_ping(self) -> None:
+        """
+        Sleep until `REMINDER_MESSAGE` should be sent again.
+
+        If latest reminder is not cached, exit instantly. Otherwise, wait wait until the
+        configured `REMINDER_FREQUENCY` has passed.
+        """
+        last_reminder: t.Optional[int] = await self.reminder_cache.get("last_reminder")
```
kwzrd, line 342: After #1091 this should now be safe - it will wait for `redis_session` inside.

bot/cogs/verification.py
```diff
@@ -32,26 +59,325 @@
...
+        for option in options:
+            await confirmation_msg.add_reaction(option)
+
+        def check(reaction: discord.Reaction, user: discord.User) -> bool:
+            """Check whether `reaction` is a valid reaction to `confirmation_msg`."""
+            return (
+                reaction.message.id == confirmation_msg.id  # Reacted to `confirmation_msg`
+                and str(reaction.emoji) in options  # With one of `options`
+                and not user.bot  # By a human
+            )
```
kwzrd, line 157: This will take a reaction from *any* user. Is this ok because the message is sent in a private channel, or should we add a sanity check for `user`'s roles in `pydis` too?

bot/cogs/verification.py
```diff
@@ -32,26 +59,325 @@
...
+"""
+
+REMINDER_FREQUENCY = 28  # Hours to wait between sending `REMINDER_MESSAGE`
+
+MENTION_CORE_DEVS = discord.AllowedMentions(
+    everyone=False, roles=[discord.Object(constants.Roles.core_developers)]
+)
+MENTION_UNVERIFIED = discord.AllowedMentions(
+    everyone=False, roles=[discord.Object(constants.Roles.unverified)]
+)
```
kwzrd, line 85: I think it's probably ok to have these as module-level constants. Didn't want to pollute the actual logic with this much boilerplate code.

bot/cogs/verification.py
```diff
@@ -32,26 +59,325 @@
...
+            await member.send(ON_JOIN_MESSAGE)
+
+    @Cog.listener()
+    async def on_message(self, message: discord.Message) -> None:
         """Check new message event for messages to the checkpoint channel & process."""
         if message.channel.id != constants.Channels.verification:
             return  # Only listen for #checkpoint messages
 
+        if message.content == REMINDER_MESSAGE.strip():
+            return  # Ignore bots own verification reminder
```
kwzrd, line 379: The way we're creating the message constants puts a newline at the top, which is stripped in Discord. I decided to solve it here so that I wouldn't have to adjust all the messages, but maybe it would be better to handle it there?

kwzrd:
Let's wait for #1093 because `Loop.is_running()` will make my life a little bit easier here.

kwzrd:
I think this should be ready for review now. Decided to add a super simple command interface for the internal tasks, it offers the following commands:

**`!verification status`**: check whether internal tasks are running
**`!verification start`**: start both internal tasks, if not already running (no-op otherwise)
**`!verification stop`**: stop both tasks if running, no-op otherwise

I don't think we need granular, per-task control. It's mostly there so that we can turn off the tasks if something goes wrong or we decide we don't like them, without unloading the cog. The settings are persisted with Redis so if the tasks are turned off, they will not start up on bot restart. It also helps during testing.

---

On the topic of Redis, because of #1095 I'm using ints `1` and `0` instead of bools for this, which is fine. The issue is pretty much insignificant and shouldn't block this PR. I will probably look into #1095 anyway, but I'd prefer to do it in a separate PR. If it gets fixed before this is approved, the fix can be used here, otherwise it doesn't matter.

I'm starting to make use of the d.py tasks framework so if anyone has any experience with it, I'd love to hear whether I used it somewhat reasonably.

kwzrd:
Testing is a little bit tricky but you will want to manually change the values of:
`UNVERIFIED_AFTER`
`KICKED_AFTER`

And probably use an alt in the testing server to see how the bot reacts to it. I'm not sure how this could be tested at scale. Please pay special attention to the way members for kicking are chosen, and the authorization process in `_verify_kick`.

I could do some unit tests if reviewers consider it necessary but I'm still totally burnt out from the testing suite for #993. If some adventurous soul want to contribute a few unit tests to this PR, I will buy you a pizza.

bot/cogs/verification.py
```diff
@@ -32,26 +59,372 @@
...
+            kick_report = "Found no users to be kicked."
+        elif not await self._verify_kick(len(for_kick)):
+            kick_report = f"Not authorized to kick `{len(for_kick)}` members."
+        else:
+            n_kicks = await self._kick_members(for_kick)
+            kick_report = f"Kicked `{n_kicks}`/`{len(for_kick)}` members from the guild."
+
+        await self.mod_log.send_log_message(
+            icon_url=self.bot.user.avatar_url,
+            colour=discord.Colour.blurple(),
```
MarkKoz, line 350: Make it red if unauthorised?

bot/cogs/verification.py
```diff
@@ -32,26 +59,372 @@
...
+            with suppress(discord.HTTPException):  # If something goes wrong, just ignore it
+                await self.bot.http.delete_message(verification.id, last_reminder)
+
+        log.trace("Sending verification reminder")
+        new_reminder = await verification.send(REMINDER_MESSAGE, allowed_mentions=MENTION_UNVERIFIED)
+
+        await self.task_cache.set("last_reminder", new_reminder.id)
+
+    @ping_unverified.before_loop
+    async def _before_first_ping(self) -> None:
```
MarkKoz, line 382: Is this basically a way to offset the loop's start relative to when the last message was sent?

bot/cogs/verification.py
MarkKoz: This is fine, but I suppose it's slightly better to do the strip where you define the constant, since that will only need to be done once.

bot/cogs/verification.py
kwzrd: Yes, this will run before the loop starts, and the loop starts only after this exits. So we're offsetting the loop's start, exactly as you say. If the last ping was sent 5 hours ago and the frequency is 28, this will sleep for 23 hours before letting the loop begin.

bot/cogs/verification.py
kwzrd: I do usually colour embeds but in this case it wasn't clear to me exactly what would determine the colours. There's a lot that can happen, no members are found, some are found but some requests fail, all requests fail, request isn't authorized, etc. For this reason I decided to just colour it blurple so that it looks more appealing but doesn't mislead. Colouring the embed red if the request isn't authorized but the roles are still assigned feels off to me ~ technically nothing has gone wrong, it's working as intended.

bot/cogs/verification.py
```diff
@@ -32,26 +59,372 @@
...
+                n_kicked += 1
+
+        self.bot.stats.incr("verification.kicked", count=n_kicked)
+
+        if bad_statuses:
+            log.info(f"Failed to kick {len(members) - n_kicked} members due to following statuses: {bad_statuses}")
+
+        return n_kicked
+
+    async def _give_role(self, members: t.Set[discord.Member], role: discord.Role) -> int:
```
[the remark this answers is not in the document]
kwzrd: Hmm I was playing with the idea but I felt like a) I wasn't actually saving much code, and b) it led to a little bit of indirection with what's happening which I really wanted to avoid for this feature. I will give it another try.

bot/cogs/verification.py
```diff
@@ -32,26 +59,372 @@
...
+    Check whether `member` is considered verified.
+
+    Members are considered verified if they have at least 1 role other than
+    the default role (@everyone) and the @Unverified role.
+    """
+    unverified_roles = {
+        member.guild.get_role(constants.Roles.unverified),
+        member.guild.default_role,
+    }
+    return bool(set(member.roles) - unverified_roles)
```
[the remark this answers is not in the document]
kwzrd: I agree that's better: 553b6a9

bot/cogs/verification.py
kwzrd: Replaced the bot check with a role id check in f3c16f7.

bot/cogs/verification.py
```diff
@@ -32,26 +59,372 @@
...
+        result = str(choice) == constants.Emojis.incident_actioned
+        log.debug(f"Received answer: {choice}, result: {result}")
+
+        # Edit the prompt message to reflect the final choice
+        await confirmation_msg.edit(
+            content=f"Request to kick `{n_members}` members was {'authorized' if result else 'denied'}!"
+        )
+        return result
+
+    async def _kick_members(self, members: t.Set[discord.Member]) -> int:
```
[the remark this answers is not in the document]
kwzrd: Yeah, it makes sense: 9b91847

bot/cogs/verification.py
kwzrd: Stripping at definition time in 174796a.

bot/cogs/verification.py
```diff
@@ -32,26 +59,372 @@
...
+            return False
+        finally:
+            await confirmation_msg.clear_reactions()
+
+        result = str(choice) == constants.Emojis.incident_actioned
+        log.debug(f"Received answer: {choice}, result: {result}")
+
+        # Edit the prompt message to reflect the final choice
+        await confirmation_msg.edit(
+            content=f"Request to kick `{n_members}` members was {'authorized' if result else 'denied'}!"
```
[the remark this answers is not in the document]
kwzrd: Added the emoji in 286cdcc. I used :warning: to keep consistency with the syncer confirmation.

Do you think that the method is a little too long now? It barely fits on my screen. We may have to break it up, although it doesn't do anything too complex. It's just the complexity of communicating with pesky humans. :angry:

bot/cogs/verification.py
```diff
@@ -32,26 +59,372 @@
...
+            )
+
+        timeout = 60 * 5  # Seconds, i.e. 5 minutes
+        try:
+            choice, _ = await self.bot.wait_for("reaction_add", check=check, timeout=timeout)
+        except asyncio.TimeoutError:
+            log.debug("Staff prompt not answered, aborting operation")
+            return False
+        finally:
+            await confirmation_msg.clear_reactions()
```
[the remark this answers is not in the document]
kwzrd: I suppose it doesn't hurt to suppress evil doings. See: 286cdcc

I think this is fine now but could you please double check. Maybe there's a way in which it could enter a weird state. But I think it's probably fine.

bot/cogs/verification.py
```diff
@@ -32,26 +59,372 @@
...
+
+        for member in members:
+            if is_verified(member):  # Member could have verified in the meantime
+                continue
+            with suppress(discord.Forbidden):
+                await member.send(KICKED_MESSAGE)  # Send message while user is still in guild
+            try:
+                await member.kick(reason=f"User has not verified in {KICKED_AFTER} days")
+            except discord.HTTPException as http_exc:
+                bad_statuses.add(http_exc.status)
```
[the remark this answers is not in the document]
kwzrd: I'm not sure about the `gather`. The first time this runs, it will try to kick about 10k people. I honestly don't know what will happen, and whather `gather`ing the coros would make things any different. I was actually considering putting a short sleep after each request, too ~ but that wouldn't work with `gather`, I guess, since they'd sleep concurrently.

bot/cogs/verification.py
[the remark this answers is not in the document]
MarkKoz: Looks fine.

bot/cogs/verification.py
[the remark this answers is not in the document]
MarkKoz: It's getting there, but it's okay for now. I can't think of a good way to break it up. If you can, then go for it.

bot/cogs/verification.py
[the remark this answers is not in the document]
kwzrd: I added a generic executor in 1958978 and then adjusted the two methods to use it in dc70a01. The idea is similar to what I was showing to you earlier, but that solution was flawed due to the fact the the built coros weren't able to communicate whether they aborted due to `is_verified(member)` giving `True`. This should no longer be a problem with this new approach, as I was able to move the check outside of the coro.

kwzrd:
@MarkKoz I think the idea we had with only letting `_check_members` collect up to 100 members for each category doesn't work, because it entirely defeats the purpose of `_verify_kick`. Now that we have a generic request executor, we can probably figure out some logic to batch the members and sleep after each batch.

kwzrd:
689e520 is an attempt at throttling the requests so that the bot doesn't drown in ratelimits.

The chosen values are mostly semi-educated guesses based on the following:
```
REST:
        POST Message |  5/5s    | per-channel
      DELETE Message |  5/1s    | per-channel
 PUT/DELETE Reaction |  1/0.25s | per-channel
        PATCH Member |  10/10s  | per-guild
   PATCH Member Nick |  1/1s    | per-guild
      PATCH Username |  2/3600s | per-account
      |All Requests| |  50/1s   | per-account
WS:
     Gateway Connect |   1/5s   | per-account
     Presence Update |   5/60s  | per-session
 |All Sent Messages| | 120/60s  | per-session
```

bot/cogs/verification.py
```diff
@@ -32,26 +59,408 @@
...
+            try:
+                await request(member)
+            except discord.HTTPException as http_exc:
+                bad_statuses.add(http_exc.status)
+            else:
+                n_success += 1
+
+            if progress % limit.batch_size == 0:
+                log.trace(f"Processed {progress} requests, pausing for {limit.sleep_secs} seconds")
+                await asyncio.sleep(limit.sleep_secs)
```
kwzrd, line 270: We could bump this log to info level for the initial pass, to keep an eye on the bot.

bot/cogs/verification.py
```diff
@@ -88,6 +88,13 @@
 Request = t.Callable[[discord.Member], t.Awaitable]
 
 
+class Limit(t.NamedTuple):
+    """Composition over config for throttling requests."""
+
+    batch_size: int  # Amount of requests after which to pause
+    sleep_secs: int  # Sleep this many seconds after each batch
```
MarkKoz, line 95: What's the justification for this data structure? It's only 2 arguments; they could easily be passed directly to the function.

kwzrd: I'm not sure, I think I just felt like both ints configure one thing, so it made sense to me to group them together. It also keeps the function signature a little bit cleaner & shorter. Instead of documenting them in the `_send_request` docstring I can explain what the composition is, and then just refer to the type. They can be separate if you prefer.

I had originally created this:
```py
ThrottledRequest = t.Tuple[Request, Limit]
```
... but eventually abandoned the idea.

kwzrd: The function does not currently support *not* receiving a `limit` but if it did then I think it'd really make sense to group them. Then it enforces that either both are provided or neither. Perhaps I was originally considering making it optional, but since both uses now are throttled it'd probably just be unnecessary.

MarkKoz: I don't feel strongly about it.

bot/cogs/verification.py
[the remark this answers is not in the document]
MarkKoz: Looks good now.

bot/cogs/verification.py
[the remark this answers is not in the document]
MarkKoz: We concluded a gather doesn't make sense due to rate limits. It probably isn't smart enough to know a separate request has triggered a rate limit, so it'll keep spamming the rest of the requests and rate limit the bot even harder. However, I haven't confirmed that. Regardless, we're going to be waiting a long time to finish all the requests so it doesn't make a difference if they are all scheduled at the start of an interval or spread out over the interval.

kwzrd:
@MarkKoz Thanks for the review!

Out of interest, how did you reach this state? I think it shouldn't be possible unless the first task fails.

screenshot

MarkKoz:
I don't have an exact answer but basically I changed durations from hours/days to smaller units and forgot to change a couple spots.

kwzrd:
Alright, that's fine then. I noticed a word was missing from that message; I pushed a commit to amend that.

jb3:
Some notes on a not getting banned from Discord front:
- If we trip the anti-spam system we will receive an error similar to `discord.errors.Forbidden: 403 Forbidden (error code: 20026): Your bot has been flagged by our anti-spam system for abusive behavior. Please reach out to our team by going to https://dis.gd/report and selecting "Appeal an Action Trust and Safety took on my bot" as the Report Type.`
- This should cancel the DM process and alert the entire administration team since it will lock the bots functionality to DM members (I have systems ready to bring up a backup bot should we experience this issue).
- We do not know that this is the only error code in use for anti-spam violations, instead we should just whitelist the 50007 code used for failed DM and everything else should probably cancel the process.

bot/cogs/verification.py
```diff
@@ -32,26 +59,408 @@
 <#{constants.Channels.bot_commands}>.
 """
 
-BOT_MESSAGE_DELETE_DELAY = 10
+# Sent via DMs to users kicked for failing to verify
+KICKED_MESSAGE = f"""
+Hi! You have been automatically kicked from Python Discord as you have failed to accept our rules \
+within `{KICKED_AFTER}` days. If this was an accident, please feel free to join again.
+"""
```
jb3, line 66: I think we should try make it as easy as possible for users to re-join in this message.

```suggestion
KICKED_MESSAGE = f"""
Hi! You have been automatically kicked from Python Discord as you have failed to accept our rules \
within `{KICKED_AFTER}` days. If this was an accident, please feel free to join us again at https://discord.gg/python.
"""
```

jb3 (changes requested):
Quick code comment, and treat the above comment as requested changes as well.

bot/cogs/verification.py
kwzrd: Yea, good idea. I'm tempted to define it in the config, probably under `Guild.invite`.

bot/cogs/verification.py
```diff
@@ -32,26 +59,408 @@
...
+You will be kicked if you don't verify within `{KICKED_AFTER}` days.
+""".strip()
+
+REMINDER_FREQUENCY = 28  # Hours to wait between sending `REMINDER_MESSAGE`
+
+MENTION_CORE_DEVS = discord.AllowedMentions(
+    everyone=False, roles=[discord.Object(constants.Roles.core_developers)]
+)
+MENTION_UNVERIFIED = discord.AllowedMentions(
+    everyone=False, roles=[discord.Object(constants.Roles.unverified)]
```
Senjan21, line 84: Doesn't seem like there is any use in specifying `everyone` to `False` since the messages won't be controlled by users nor will they actually ping the `@everyone` role

bot/cogs/verification.py
```diff
@@ -32,26 +59,408 @@
...
+        log.debug(f"Received answer: {choice}, result: {result}")
+
+        # Edit the prompt message to reflect the final choice
+        if result is True:
+            result_msg = f":ok_hand: Request to kick `{n_members}` members was authorized!"
+        else:
+            result_msg = f":warning: Request to kick `{n_members}` members was denied!"
+
+        with suppress(discord.HTTPException):
+            await confirmation_msg.edit(content=result_msg)
```
Senjan21, line 239: By doing that, we are editing out the ping to our core developers, which could result in confusion as to why a ping occurred. Not editing the ping out would be clearer as well as more consistent considering our sync check doesn't edit out the ping either.

Senjan21 (changes requested):

bot/cogs/verification.py
kwzrd: Added invite in 53189e8, but I had to put it on the next line. The invite does not embed properly if followed by a dot.
image

bot/cogs/verification.py
```diff
@@ -32,26 +59,409 @@
 <#{constants.Channels.bot_commands}>.
 """
 
-BOT_MESSAGE_DELETE_DELAY = 10
+# Sent via DMs to users kicked for failing to verify
+KICKED_MESSAGE = f"""
+Hi! You have been automatically kicked from Python Discord as you have failed to accept our rules \
+within `{KICKED_AFTER}` days. If this was an accident, please feel free to join us again!
+{constants.Guild.invite}
+"""
```
jb3, line 67: I think an extra newline may look tidier.

```suggestion
KICKED_MESSAGE = f"""
Hi! You have been automatically kicked from Python Discord as you have failed to accept our rules \
within `{KICKED_AFTER}` days. If this was an accident, please feel free to join us again!

{constants.Guild.invite}
"""
```

jb3 (changes requested):

bot/cogs/verification.py
kwzrd: Newlinered in 3ea04e3.

john_newlineman

bot/cogs/verification.py
kwzrd: See: eff2f75

bot/cogs/verification.py
kwzrd: Mention will be retained in 44aae45.

image

kwzrd:
@jb3 We now stop kicking users on suspicious 403 and alert admins:

image

Screenshot is misleading ~ it will happen for 403s that are *not* `50_007`, but I didn't know how to produce those in testing.

See 1d6d845 for implementation details. It may have to be adjusted, but I just don't know how to do it better.

kwzrd:
Currently there are two internal tasks: one that manages users periodically (give role or kick), and one that pings `@Unverified` every n hours. When I was designing this initially I wanted there to be a command which will allow us to start or stop the tasks, and persist this setting in Redis, so that they will remember whether to run or not if the bot restarts. However, at the time it felt like we didn't need to keep track of these settings on per-task basis, either both tasks are running (most commonly) or something goes wrong and we disable both tasks while we're resolving the issue that occurred.

Now that the first task will abort on suspicious 403s, however, it would also make sense for it to un-schedule itself, so that it doesn't restart in 30 minutes. This is easy, but then the cog enters a state where one task is running and the other is not, and given the current setup we cannot persist this state in Redis (because there is only one control variable).

There are a few things that we can do here:
* Do nothing ~ if something goes wrong, the bot pings `@Admins` which will then review the situation and use `!verification stop` to kill both tasks if appropriate (if the first task isn't running, it won't be assigning the `@Unverified` role anyway)
* If the suspicious 403 occurs, ping `@Admins` and automatically kill both tasks, and store a `0` in Redis meaning that if the bot restarts, the tasks should remain dead
* If the suspicious 403 occurs, kill the first task only, add commands to manage each task separately, and keep track of each task status separately in Redis (some extra work and complexity, but not too horrible...)

@MarkKoz @Senjan21 @jb3 or anyone else, how do we feel about this?

MarkKoz:
I prefer the second option (kill both tasks) because it's the simplest. The role task just adds a role, which is useless on its own (i.e. when the kick task isn't also running). Once the role task is restarted in the future, it'll process all the users it missed while it was off, so there's no benefit to keeping it running while the kick task is off.

Senjan21:
I agree with Mark in that case, second option would be ideal.

kwzrd:
@MarkKoz @Senjan21 I agree & went with option 2 but for clarity: kicking users and adding the `@Unverified` role is handled by the same task (above referred to task task `1`). Task `2` only pings.

bot/cogs/verification.py
```diff
@@ -32,26 +59,464 @@
...
+
+        If latest reminder is not cached, exit instantly. Otherwise, wait wait until the
+        configured `REMINDER_FREQUENCY` has passed.
+        """
+        last_reminder: t.Optional[int] = await self.task_cache.get("last_reminder")
+
+        if last_reminder is None:
+            log.trace("Latest verification reminder message not cached, task will not wait")
+            return
+
```
MrHemlock, line 486: Will this be an issue if the bot is forced to restart?  Is the task_cache stored somewhere persistent?

MrHemlock (changes requested):
Overall I like what I see, but I do have one question and one suggestion that I'd like to see addressed.

bot/cogs/verification.py
SebastiaanZ: `task_cache` is a `RedisCache`; it will persist in Redis.

kwzrd: Yea, think of `task_cache` as a dict that persists across restarts.

This condition should only be True the first time the bot connects. After that, the last sent reminder should always be cached, but nothing bad happens if it isn't ~ the bot will just send another one, and then cache it.

bot/cogs/verification.py
```diff
@@ -1,19 +1,46 @@
...
-Hello! Welcome to the server, and thanks for verifying yourself!
+UNVERIFIED_AFTER = 3  # Amount of days after which non-Developers receive the @Unverified role
+KICKED_AFTER = 30  # Amount of days after which non-Developers get kicked from the guild
+
+# Number in range [0, 1] determining the percentage of unverified users that are safe
+# to be kicked from the guild in one batch, any larger amount will require staff confirmation,
+# set this to 0 to require explicit approval for batches of any size
+KICK_CONFIRMATION_THRESHOLD = 0.01  # 1%
+
+BOT_MESSAGE_DELETE_DELAY = 10
```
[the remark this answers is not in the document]
kwzrd: I'm not sure. I usually try to advocate keeping the definitions close to where they are used.

Let's imagine that these settings are in the `config.yml`. If you're reading the code and you wonder what the value is, you have to:
* Follow the import to `constants.py`
* Look at from where the value is being loaded
* Navigate to the YAML and find the value

IDE generally cannot aid you in this process, as the value only gets bound at runtime. In my opinion, this creates a significant level of indirection between the developer and the code. I think it's fair to say that these values have no use outside of this module, and so making them more accessible globally has no benefit.

On the other hand, having the constants here, your IDE can probably tell you what the value is if you just hover the reference, and if you want to change the values locally for debugging purposes, you can do that just as easily.

These are my thoughts on the topic, but I know that many would disagree. I wouldn't mind moving them to the config file, but I'd like to hear more of a justification for it. Because personally I cannot help but feel like I'm only making things more difficult for myself.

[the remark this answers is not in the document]
kwzrd: Another downside that I notice with creating move distance between the value and its use, is that it increases the amount of necessary documentation. The names of these variables would likely need to be longer, and be more annotated, so that it's obvious what they represent.

bot/cogs/verification.py
MrHemlock: Excellent, that takes care of my worry on that then.

bot/cogs/verification.py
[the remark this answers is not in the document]
SebastiaanZ: One additional factor to consider is if these values need to be changed for a testing environment. If so, using the config system ensures that people do not have the modify any of the files checked into our version control system, as they only have to configure `config.yaml` to their liking to overwrite the default in `config-default.yaml`. This prevents debug values from being accidentally committed to the repository.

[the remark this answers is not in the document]
MrHemlock: I'm honestly kind of torn about whether we should move it.  On one hand you're right, there's all the distance between the values and the and where they're used that causes potential issues sifting through the code, as I've certainly had in the past.  

I also think there's an argument to be made that the anti-spam cog needs it to be in the config file just due to the number of rules it touches.  But I think my main concern is when we actually need to tweak the configuration from a moderation standpoint.  Most of the other things that can be tweaked are in the config and are grouped together and easy enough to navigate just by using CTRL+F to find the right key word for the cog.  

If we do decide to keep them in the cog, I think we at the very least move any of the numerical constants to the top so that they're grouped together, as they're the ones we'd be more likely to need to tweak.   I'd be curious to hear the opinions of some of the others on this, though.  

And if nothing else it's something we can move at a later date should we decide to go that route.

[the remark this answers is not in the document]
kwzrd: > This prevents debug values from being accidentally committed to the repository.

@SebastiaanZ: This is true, although it doesn't feel like a very strong argument to me. I don't think it's difficult to review the diff before commiting it, and it should definitely be done before pushing. Unintended diff being commited & pushed sadly happens often even with these measures in place. I actually probably find it easier to have them version controlled, it forces you to keep your settings tidy & you will always know when and how your configured behaviour differs from the production one.

With regard to debug values, keeping the values here will reload them if you reload the extension. This means that you don't have to restart the whole bot to tweak them. But that's more of a `d.py` detail. It is extremely useful when testing the feature though.

> But I think my main concern is when we actually need to tweak the configuration from a moderation standpoint.

I'd be perfectly fine with moving them if it makes things easier for others. For the reasons outlined in the paragraph above, I don't like working with the config yaml too much, but once the code hits production it will be a different situation too. From a maintenance perspective it may just make more sense.

I'm happy that we're talking about this though. Let's wait a little longer to see what others think.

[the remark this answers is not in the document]
SebastiaanZ: > This is true, although it doesn't feel like a very strong argument to me. I don't think it's difficult to review the diff before commiting it, and it should definitely be done before pushing. Unintended diff being commited & pushed sadly happens often even with these measures in place. I actually probably find it easier to have them version controlled, it forces you to keep your settings tidy & you will always know when and how your configured behaviour differs from the production one.

To me, personally, it's one of the biggest reasons why we use the configuration system we do. You can easily overwrite constants without having to touch files that actually have to be checked in. This means that you can keep your local debugging setup without having to revert values and us risking a debug value making it past review. I'd rather have a foolproof method than one that relies on human review, because we all make mistakes during reviews and sometimes features get merged with fairly minimal reviews.

I find the opposite approach tidy: You don't have to mess with checked in files, you don't have to be careful with committing changes to avoid committing lines then need to be reverted to keep a clean blame/commit history, you can persist your local debug setup independent of check in files.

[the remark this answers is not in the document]
MarkKoz: Ideally these would be configurable. It's just that the current config system sucks and we haven't gotten far with coming up with an alternative.

AmerMarkisic:
CodeJam is bullying us :(

bot/cogs/verification.py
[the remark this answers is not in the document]
MrHemlock: I think for now, we should align with the current configuration standard and move the things we would need to tweak on the fly (the constants I mentioned up above) to the config-default.yaml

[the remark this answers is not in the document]
kwzrd: Moved constants to the YAML file in 1b3a5c5 including `BOT_MESSAGE_DELETE_DELAY` which was an in-module constant before this PR. I'm not sure, but it doesn't look like the top-level definitions in neither `constants.py` nor the YAML are ordered lexicographically, so I just put mine below `PythonNews` in both.

MrHemlock (approved):
Looks good to me, let's get the ball rolling.

PR #1086, opened 2020-08-04
Help channel message pin fixes

### LINKED ISSUE #1082: Don't unpin message if the message ID is None
Sentry Issue: [BOT-7G](https://sentry.io/organizations/python-discord/issues/1812744317/?referrer=github_integration)

If a message for the channel cannot be retrieved from the cache, don't try to unpin the message. It currently results in a warning.

https://github.com/python-discord/bot/blob/3649c2a03dc158a25f2c3be98db8691f903a1953/bot/cogs/help_channels.py#L554-L562

Also, sneak in a minor change to the warning log message to include the HTTP status as well as the code.

MarkKoz:
Fixes #1082 - don't unpin message if the message ID is `None`. This only really happened when the pin feature was added and the cache was empty, but it doesn't hurt to check anyway.

The check for an empty channel was not accounting for the bot's `pins_add` message and therefore did not re-schedule the channel to be made dormant earlier when messages were deleted. It has been fixed by searching for the bot's available message embed within the last 100 messages and ensuring every message leading up to the embed was also a bot message. This means all bot messages are ignored, such as command invocations and the code block format embed.

Lastly, the pin and unpin code was refactored into separate functions. The refactor also makes them share error handling code, which reduces redundancy.

### DIFF
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -738,6 +738,7 @@ class HelpChannels(commands.Cog):
 
     async def is_empty(self, channel: discord.TextChannel) -> bool:
         """Return True if there's an AVAILABLE_MSG and the messages leading up are bot messages."""
+        log.trace(f"Checking if #{channel} ({channel.id}) is empty.")
         found = False
 
         # A limit of 100 results in a single API call.
@@ -745,9 +746,11 @@ class HelpChannels(commands.Cog):
         # Not gonna do an extensive search for it cause it's too expensive.
         async for msg in channel.history(limit=100):
             if not msg.author.bot:
+                log.trace(f"#{channel} ({channel.id}) has a non-bot message.")
                 return False
 
             if self.match_bot_embed(msg, AVAILABLE_MSG):
+                log.trace(f"#{channel} ({channel.id}) has the available message embed.")
                 found = True
                 break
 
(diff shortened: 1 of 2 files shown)


bot/cogs/help_channels.py
```diff
@@ -754,9 +737,24 @@ async def on_message_delete(self, msg: discord.Message) -> None:
...
+            if not msg.author.bot:
+                log.trace(f"#{channel} ({channel.id}) has a non-bot message.")
+                return False
+
+            if self.match_bot_embed(msg, AVAILABLE_MSG):
+                log.trace(f"#{channel} ({channel.id}) has the available message embed.")
+                found = True
+                break
+
+        return found
```
kwzrd, line 757: Usually when I see this pattern I try to refactor it into just returning directly, i.e. instead of:
```python
found = False

for x in y:
  if z:
    found = True
    break

return found
```
I think it's nicer to just:
```python
for x in y:
  if z:
    return True

return False
```

MarkKoz: There was a reason for this before but I don't remember anymore. Probably a previous version of the code.

bot/cogs/help_channels.py
```diff
@@ -863,6 +861,47 @@ async def try_get_channel(self, channel_id: int) -> discord.abc.GuildChannel:
...
+
+        try:
+            await func(channel.id, msg_id)
+        except discord.HTTPException as e:
+            if e.code == 10008:
+                log.debug(f"Message {msg_id} in {channel_str} doesn't exist; can't {verb}.")
+            else:
+                log.exception(
+                    f"Error {verb}ning message {msg_id} in {channel_str}: {e.status} ({e.code})"
+                )
```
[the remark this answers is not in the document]
MarkKoz: I try to use the 100 col margin if it's not too inconvenient.

PR #1089, opened 2020-08-05
HelpChannels: use more reliable check for claimed channel

### LINKED ISSUE #1074: API lag may lead to more than 2 available help channels
`on_message` uses a lock to ensure messages sent in available channels are processed sequentially. After it acquires the lock, it checks if the channel has already been moved out of available. If it has, then it aborts further processing. The problem is that this relies on Discord moving the channel in a reasonable amount of time. If it happens slowly, then the next message that acquires the lock will still think the channel can be claimed. 

This is because only the request to move the channel is awaited. It can be sure the request has been sent, but not that Discord has processed it and moved the channel. To do that, it'd have to wait for a guild channel update event. That's just one solution. Another idea I have is to check if the channel has a claimant associated with it, since that is already tracked for other purposes.

MarkKoz:
Using the channel's category isn't reliable since it may take Discord a while to actually move the channel once it's received a request from the bot. I suppose using redis technically has the same problem, but it should be much faster and less susceptible to lag than Discord. If it's still a problem down the road, we can consider waiting for the channel update event from Discord. I didn't go for that solution right away since it's relatively more complicated and the gateway cannot exactly always be relied upon either.

Fixes #1074

### DIFF
--- a/bot/cogs/help_channels.py
+++ b/bot/cogs/help_channels.py
@@ -694,7 +694,7 @@ class HelpChannels(commands.Cog):
         async with self.on_message_lock:
             log.trace(f"on_message lock acquired for {message.id}.")
 
-            if not self.is_in_category(channel, constants.Categories.help_available):
+            if await self.help_channel_claimants.contains(channel.id):
                 log.debug(
                     f"Message {message.id} will not make #{channel} ({channel.id}) in-use "
                     f"because another message in the channel already triggered that."


kwzrd:
Apologies, reverted in 5a7ca92cf5d5ae7c7d4aa7ba086237586832af1a.

The mappings are only being removed from `help_channel_claimants` if the `!close` command is used. When `move_to_dormant` is called from elsewhere, the mapping persists. This causes the bot to think that the channel was already claimed by a previous message. Definitely should have checked for that, sorry.

PR #1103, opened 2020-08-14
Move cogs to subdirectories

### LINKED ISSUE #160: Introduce subfolders into the cog folder so that we can organize the cogs better.
**Originally posted by Leon Sandøy:**

We should consider either consolidating similar cogs into a single cog with more features (like an Admin cog that had big brother, clean features, and defcon), or to create subfolders in the cogs folder so we can at least group the similar cogs together. The folder is growing unruly.

MarkKoz:
Resolves #160 

## Restructure

This restructure helps with organisation and declutters the `cogs` directory. The subdirectories roughly correspond to how issues and PRs were being labelled here on GitHub. Tests have been restructured to mirror this. `help_channels` was left uncategorised, but it'll likely get its own subpackage in the future once it's refactored into separate modules.

```
.
├── backend
│   ├── sync
│   │   ├── __init__.py
│   │   ├── _cog.py
│   │   └── _syncers.py
│   ├── __init__.py
│   ├── alias.py
│   ├── config_verifier.py
│   ├── error_handler.py
│   └── logging.py
├── filters
│   ├── __init__.py
│   ├── antimalware.py
│   ├── antispam.py
│   ├── filtering.py
│   ├── filter_lists.py
│   ├── security.py
│   ├── token_remover.py
│   └── webhook_remover.py
├── fun
│   ├── __init__.py
│   ├── duck_pond.py
│   └── off_topic_names.py
├── info
│   ├── __init__.py
│   ├── doc.py
│   ├── help.py
│   ├── information.py
│   ├── python_news.py
│   ├── reddit.py
│   ├── site.py
│   ├── source.py
│   ├── stats.py
│   ├── tags.py
│   └── wolfram.py
├── moderation
│   ├── infraction
│   │   ├── __init__.py
│   │   ├── _scheduler.py
│   │   ├── _utils.py
│   │   ├── infractions.py
│   │   ├── management.py
│   │   └── superstarify.py
│   ├── watchchannels
│   │   ├── __init__.py
│   │   ├── _watchchannel.py
│   │   ├── bigbrother.py
│   │   └── talentpool.py
│   ├── __init__.py
│   ├── defcon.py
│   ├── dm_relay.py
│   ├── incidents.py
│   ├── modlog.py
│   ├── silence.py
│   ├── slowmode.py
│   └── verification.py
├── utils
│   ├── __init__.py
│   ├── bot.py
│   ├── clean.py
│   ├── eval.py
│   ├── extensions.py
│   ├── jams.py
│   ├── reminders.py
│   ├── snekbox.py
│   └── utils.py
├── __init__.py
└── help_channels.py
```

## Renaming `cogs` to `exts`

The directory contains modules, which are extensions. It only indirectly contains cogs through the extensions. Therefore, a technically more accurate name is "extensions", or "exts" when abbreviated. Furthermore, "exts" is consistent with SeasonalBot. This was discussed in the comment in #160.

## Finer-grained extensions

The "meta" extensions have also been removed (such as `moderation`); now each cog has its own extension. As a result, there is finer-grained control over which cogs to load. I didn't come across a case in which it made more sense to keep cogs grouped into a single extension. Besides, I think a better solution would be partial wildcard support by the `!extensions` command e.g. `!extensions load bot.cogs.moderation.*`. Such feature can be considered in the future.

## Extension discovery

From now on, all non-extension modules should be prefixed with an underscore. This will indicate to extension-discovering algorithms (used by the `!extensions` command and command-shadowing tests) that those modules should be skipped. However, this does not apply to package names. Instead, packages will be imported to determined if they are extensions (by checking for the presence of `setup()`). This method was chosen for the following reasons:

1. It's automated - no need to manually manage lists of valid extensions
2. It's clean - imposes nothing on how the packages should be named
3. Imports are done anyway - `pkgutil.walk_packages` imports packages to get the `__path__` attribute (apparently to support namespace packages), so a second import by us isn't going to have any additional impact. This could have been avoided by writing a custom `walk_packages` which does away with namespace package support, but given that inconvenience and the two previous reasons, I am okay with `pkgutil`'s behaviour.

I recommend to avoid writing code in `__init__.py`s that will have (significant) side effects upon import. What I've done so far to alleviate this is move imports into `setup()` where applicable (currently only for the `sync` extension).

#### List

`!extensions list` now groups extensions by their category:

extensions list

The embed is normally paginated; I just had a large character limit for debugging purposes. The pagination is done in a way that avoids cutting off a page in the middle of a category. A category will always be shown in full on a single page, unless it's at Discord's description character cap of 2000.

#### Extension arguments

The commands still support unqualified extension names. However, if there is an ambiguous name, an error will ask the user to use a fully-qualified, unambiguous name. An ambiguous name would result from two modules having the same name but being in different subpackages.

ambiguous extensions error


bot/exts/backend/sync/__init__.py
```diff
@@ -1,7 +1,7 @@
 from bot.bot import Bot
-from .cog import Sync
 
 
 def setup(bot: Bot) -> None:
     """Load the Sync cog."""
+    from ._cog import Sync
```
lemonsaurus, line 6: Might be worth adding a comment to explain why this import is now here. It's not immediately intuitive.

bot/exts/backend/sync/_cog.py
```diff
@@ -8,7 +8,7 @@
 from bot import constants
 from bot.api import ResponseCodeError
 from bot.bot import Bot
-from bot.cogs.sync import syncers
+from . import _syncers
```
lemonsaurus, line 11: Why is this necessary? I really dislike dot imports.

bot/exts/utils/extensions.py
```diff
@@ -1,26 +1,47 @@
...
-)
+
+def walk_extensions() -> t.Iterator[str]:
+    """Yield extension names from the bot.exts subpackage."""
+
+    def on_error(name: str) -> t.NoReturn:
+        raise ImportError(name=name)  # pragma: no cover
+
+    for module in pkgutil.walk_packages(exts.__path__, f"{exts.__name__}.", onerror=on_error):
+        if module.name.rsplit(".", maxsplit=1)[-1].startswith("_"):
```
lemonsaurus, line 29: kinda gross, but I'm not sure turning it into more lines would really help readability much.

lemonsaurus (changes requested):
It mostly looks fine, but I have a few changes to request.
- Please justify or change the frequent use of explicit relative imports. I don't think we have enough complexity in our package layout for the imports to end up being that verbose, and PEP8 recommends that we use absolute imports as much as possible in cases where this is true.
- Maybe we should default to just importing all the extensions instead of maintaining this long list of load_extensions calls in `__main__`. For extensions that should be conditionally loaded (like HelpChannels), we could either conditionally _unload_ them, or add them to some sort of list of extensions that should not be loaded and then handle the loading afterwards.
- I think we should find a home for the root folder extensions. For example, dm_relay seems like it could probably fit into the moderation folder, and alias might be a good fit for backend. Let's discuss this in #dev-core and try to find a suitable home for the remaining extensions before we merge this?

bot/exts/backend/sync/__init__.py
MarkKoz: Addressed by a66df29.

bot/exts/backend/sync/_cog.py
MarkKoz: Addressed by f150e69.

I don't feel strongly about it, but there are some nice things about relative imports:

1. They tend to be shorter
    * Easier to read (contentious, I guess)
    * Easier to write (mentioning this for completeness sake; it's not really a valid argument on its own)
    * More room to import many things without breaking into multiple lines
2. Sorting imports puts relative ones at the end. This makes it more convenient to find the most relevant objects to the current module, on the assumption that code in the same package is more relevant than objects in external packages.

They weren't importing anything deeply nested, so I didn't mind them.

bot/exts/utils/extensions.py
MarkKoz: Addressed by 57f5f10.

MrHemlock (approved):
Everything looks clean to me.

kosayoda (approved):
All aboard the merge train choo choo~

Issue #1098, opened 2020-08-10
Add option to delete docs embed


gustavwilliam:
When `!docs print` is run, python returns the documentation for the print function.

Often times, the docs returned may not be what was expected or the user simply wanted to do a quick check and then get rid of the large embed. In both of those cases, the option to delete the response would be perfect.

This is already implemented for commands such as `!eval`, where a 🗑 reaction can be clicked to delete the response. The logic could simply be carried over.

AtieP:
I can do this

MarkKoz:
Yeah, this is a good idea. We already do this for a lot of other command outputs, so I don't see a reason to leave docs out. Let's do this.

AtieP:
This issue has been resolved.

PR #1106, opened 2020-08-15
Add doc cleanup

AtieP:
Closes issue #1098. This adds a reaction to the documentation embed and waits for a trashcan reaction by the message author, so the embed will delete itself. Note that there is a 300 seconds timeout, like the help embed.

### DIFF
--- a/bot/cogs/doc.py
+++ b/bot/cogs/doc.py
@@ -19,7 +19,7 @@ from sphinx.ext import intersphinx
 from urllib3.exceptions import ProtocolError
 
 from bot.bot import Bot
-from bot.constants import MODERATION_ROLES, RedirectOutput
+from bot.constants import MODERATION_ROLES, RedirectOutput, Emojis
 from bot.converters import ValidPythonIdentifier, ValidURL
 from bot.decorators import with_role
 from bot.pagination import LinePaginator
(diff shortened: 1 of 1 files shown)


AtieP:
Preview: https://streamable.com/eqhboz

AtieP:
Hm, I didn't know that. I'm at the moment away from my computer, so feel free to change my code somehow.

bot/cogs/doc.py
```diff
@@ -391,7 +392,8 @@ async def get_command(self, ctx: commands.Context, symbol: commands.clean_conten
                     await error_message.delete(delay=NOT_FOUND_DELETE_DELAY)
                     await ctx.message.delete(delay=NOT_FOUND_DELETE_DELAY)
             else:
-                await ctx.send(embed=doc_embed)
+                doc_embed = await ctx.send(embed=doc_embed)
+                await wait_for_deletion(doc_embed, (ctx.author.id,), client=self.bot)
```
ks129, line 396: No need to assign message to variable. Put send directly to call. Also no need to pass custom client.

Numerlor: I don't agree with this, using a name makes this more readable compared to doing the whole await expression within an argument. Although an another name may be more fitting since `doc_embed` is already used to describe the embed object in the same scope.

AtieP: What names do you suggest? I kept it the same because it's still an embed... but not technically.

ks129: I'd suggest `msg`.

PR #1107, opened 2020-08-15
Don't patch ctx.message.author in antispam

### LINKED ISSUE #1005: Author of a filtered message somehow becomes the bot itself
Sentry Issue: [BOT-7D](https://sentry.io/organizations/python-discord/issues/1801394207/?referrer=github_integration)

```
AttributeError: 'ClientUser' object has no attribute 'send'
    await coro(*args, **kwargs)
  File "bot/cogs/filtering.py", line 114, in on_message
    await self._filter_message(msg)
  File "bot/cogs/filtering.py", line 184, in _filter_message
    await self.notify_member(msg.author, _filter["notification_msg"], msg.channel)
  File "bot/cogs/filtering.py", line 366, in notify_member
    await filtered_member.send(reason)

```

This is a very strange error. It was triggered by an invite link filter. Apparently the user left the guild quickly after sending the invite.

Unfortunately, the message object itself was filtered out by Sentry. Sentry shows that the match indeed had an invite link and that the author only had the `Developers` role. Furthermore, `_filter_messages` will exit early if the message's author is a bot. This means the bot definitely wasn't the author of the message. Yet somehow, the bot became the author when it tried to send a notification.

I suspect this is some esoteric discord.py cache or Discord API bug. Perhaps we should just check catch the `AttributeError` and move on. I really don't see anything wrong with our code. Does anyone else have insight into this?

MarkKoz:
The modification propagated across all code that is using the same `Message` object, including all other `on_message` listeners. This caused weird bugs e.g. the filtering cog thinking the bot authored a message that triggered a filter.

Patching only `ctx.author` means the implementation is more fragile. Infraction code must ensure it only retrieves the author via `ctx.author` and not through `ctx.message`.

How to reproduce the bug:

1. Put 3 second sleep before this line https://github.com/python-discord/bot/blob/b27c286f2ce648d021eed0c8e07476066b86dd98/bot/cogs/filtering.py#L292
2. Send a message with 15 _custom_ Discord emotes
3. Send another message with 15 emotes but also include an invite link that will trigger the filter
4. Observe the same exception as in #1005

You can follow the above steps this PR branch and see the exception no longer gets raised.

Fixes #1005
Fixes BOT-7D

### DIFF
--- a/bot/cogs/antispam.py
+++ b/bot/cogs/antispam.py
@@ -219,7 +219,6 @@ class AntiSpam(Cog):
             # Get context and make sure the bot becomes the actor of infraction by patching the `author` attributes
             context = await self.bot.get_context(msg)
             context.author = self.bot.user
-            context.message.author = self.bot.user
 
             # Since we're going to invoke the tempmute command directly, we need to manually call the converter.
             dt_remove_role_after = await self.expiration_date_converter.convert(context, f"{remove_role_after}S")
--- a/bot/cogs/moderation/modlog.py
+++ b/bot/cogs/moderation/modlog.py
@@ -120,6 +120,10 @@ class ModLog(Cog, name="ModLog"):
             else:
                 content = "@everyone"
 
+        # Truncate content to 2000 characters and append an ellipsis.
+        if content and len(content) > 2000:
+            content = content[:2000 - 3] + "..."
+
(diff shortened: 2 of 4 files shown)


Senjan21 (approved):
LGTM

lemonsaurus (approved):
looks excellent, nice solve.

PR #1109, opened 2020-08-15
Unnominate banned users from the talent pool

### LINKED ISSUE #1065: Unnominate user when banned
Currently, if a user is nominated and then banned, they are still left watched in the talent pool. Instead, they should be unnominated and removed from the talent pool.

MarkKoz:
Fixes #1065

Example history for a banned user:

bild

### DIFF
--- a/bot/cogs/watchchannels/talentpool.py
+++ b/bot/cogs/watchchannels/talentpool.py
@@ -1,8 +1,9 @@
 import logging
 import textwrap
 from collections import ChainMap
+from typing import Union
 
-from discord import Color, Embed, Member
+from discord import Color, Embed, Member, User
 from discord.ext.commands import Cog, Context, group
 
 from bot.api import ResponseCodeError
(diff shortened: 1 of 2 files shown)


bot/cogs/watchchannels/talentpool.py
```diff
@@ -220,6 +206,36 @@ async def edit_reason_command(self, ctx: Context, nomination_id: int, *, reason:
...
+        """Remove `user` from the talent pool after they are banned."""
+        await self.unwatch(user.id, "User was banned.")
+
+    async def unwatch(self, user_id: int, reason: str) -> bool:
+        """End the active nomination of a user with the given reason and return True on success."""
+        active_nomination = await self.bot.api_client.get(
+            self.api_endpoint,
+            params=ChainMap(
+                self.api_default_params,
+                {"user__id": str(user_id)}
```
SebastiaanZ, line 220: Reading this code again (I think I wrote it), I wonder why I put the defaults *before* the `user__id` dict in the `ChainMap`. The lookup order is from left to right, not right to left, meaning that the default values in `api_default_params` now take precedence over the specific values inserted as params.

It may not matter too much in this version, as I don't think our defaults will ever include a `user__id` field, but if we ever add something to the second dictionary hoping to "override" a default, it won't work. Not really in the scope of your PR, but my personal feeling is that I should have picked the reverse order in the original.

lemonsaurus: Seems reasonable to me. I don't really see any reason not to swap these?

lemonsaurus, line 220:
```suggestion
                {"user__id": str(user_id)},
                self.api_default_params,
```

bot/cogs/watchchannels/talentpool.py
```diff
@@ -220,6 +206,36 @@ async def edit_reason_command(self, ctx: Context, nomination_id: int, *, reason:
...
+            )
+        )
+
+        if not active_nomination:
+            log.debug(f"No active nominate exists for {user_id=}")
+            return False
+
+        log.info(f"Ending nomination: {user_id=} {reason=}")
+
+        [nomination] = active_nomination
```
lemonsaurus, line 230: This line triggers me a little bit. I don't really see why you have to do this. `active_nomination['id']` in the patch call wouldn't really be that bad, and the `[x] = y` syntax just looks deeply wrong to me no matter how many times I come across it.

lemonsaurus (approved):
Minor nitpicks, but overall this looks great to me. Feel free to address whatever you agree with, and leave the rest as-is.

PR #1113, opened 2020-08-18
Cache silences to enable rescheduling and revoking reaction permissions

### LINKED ISSUE #915: Forbid emoji reaction when a channel is shhhh-ed
We recently had an issue with users reaction spamming the `!silence` message, in my opinion it would be smart to also disable emojis when a channel is silenced. The following lines might be of interest:
https://github.com/python-discord/bot/blob/fe325adc66e9621ea970abd83db58df4b792caa5/bot/cogs/moderation/silence.py#L121
https://github.com/python-discord/bot/blob/fe325adc66e9621ea970abd83db58df4b792caa5/bot/cogs/moderation/silence.py#L141

### LINKED ISSUE #1053: Unsilence may try to cancel a task that doesn't exist due to a bot restart
Tasks aren't re-scheduled upon a restart, so they won't always exist. To avoid the warning, we can either add a simple check that the tasks exists before cancelling. Alternatively, we can cache the silenced channels in redis and re-schedule tasks when the bot restarts. A notification is supposed to be sent if there are active silenced channels during shutdown, but this mechanism is a victim to #1012.

I prefer the re-scheduling option. Does anyone else have thoughts on which way to go?

MarkKoz:
Fixes #915 
Fixes #1053

As discussed in #915, it isn't safe to assume default values for a channel's previous overwrites. Therefore, their previous values must be cached and restored once unsilenced. I decided to store both overwrites in a single cache via JSON (de)serialisation. This feels nicer than dealing with multiple caches and also removes worries of them falling out of sync.

Since I was already working on caching, I implemented caching of the unsilence times too. This means moderators no longer need to worry about manually  unsilencing channels after the bot restarts. Therefore, I've removed the notification on cog unload which was plagued by a race condition issue (#1053).

Most of the existing tests for the cog were broken by these changes. After fixing them, I realised there's some room for refactoring and took it upon myself to carry that out. Tests have been split into separate cases, which is more organised and allows for more effective use of the `setUp` fixture. There's only one thing I left out from tests because it was too annoying to test (somehow I found it more annoying than `test_rescheduled_active`).

Lastly, I updated the `autospec` decorator with a feature to prevent passing the mocks as arguments to the decorated function. This took longer than it seems to get to work correctly. Unfortunately, the implementation required monkey-patching `unittest.mock`, but I think the trade-off is worthwhile.


tests/_autospec.py
```diff
@@ -0,0 +1,64 @@
...
+            arg = exit_stack.enter_context(patching)
+            if not getattr(patching, "dont_pass", False):
+                # Only add the patching as an arg if dont_pass is False.
+                if patching.attribute_name is not None:
+                    keywargs.update(arg)
+                elif patching.new is unittest.mock.DEFAULT:
+                    extra_args.append(arg)
+
+        args += tuple(extra_args)
+        yield args, keywargs
```
MarkKoz, line 24: Since this is mostly copied, whatever license its under should be probably followed. [This license](https://github.com/python/cpython/blob/master/Lib/unittest/__init__.py#L29-L44) is for `unittest`. The mock module does not have its own license, only a comment that says its maintained by Michael Foord. Maybe the [PSF license](https://docs.python.org/3/license.html#psf-license-agreement-for-python-release) should be followed. I don't know, this is not straight forward.

bot/cogs/moderation/silence.py
```diff
@@ -10,11 +11,23 @@
...
 from bot.converters import HushDurationConverter
+from bot.utils import RedisCache
 from bot.utils.checks import with_role_check
 from bot.utils.scheduling import Scheduler
 
 log = logging.getLogger(__name__)
 
+MSG_SILENCE_FAIL = f"{Emojis.cross_mark} current channel is already silenced."
+MSG_SILENCE_PERMANENT = f"{Emojis.check_mark} silenced current channel indefinitely."
+MSG_SILENCE_SUCCESS = Emojis.check_mark + " silenced current channel for {duration} minute(s)."
```
Numerlor, line 22:
```suggestion
MSG_SILENCE_SUCCESS = f"{Emojis.check_mark} silenced current channel for {{duration}} minute(s)."
```

bot/cogs/moderation/silence.py
```diff
@@ -83,18 +103,23 @@ async def silence(self, ctx: Context, duration: HushDurationConverter = 10) -> N
...
-            await ctx.send(f"{Emojis.check_mark} silenced current channel indefinitely.")
+            await ctx.send(MSG_SILENCE_PERMANENT)
+            await self.unsilence_timestamps.set(ctx.channel.id, -1)
             return
 
-        await ctx.send(f"{Emojis.check_mark} silenced current channel for {duration} minute(s).")
+        await ctx.send(MSG_SILENCE_SUCCESS.format(duration=duration))
 
         self.scheduler.schedule_later(duration * 60, ctx.channel.id, ctx.invoke(self.unsilence))
+        unsilence_time = (datetime.now(tz=timezone.utc) + timedelta(minutes=duration))
```
Numerlor, line 121: Any reason for the parentheses here? Can also leave out the tz since it only gets converted to the timestamp

bot/cogs/moderation/silence.py
```diff
@@ -136,28 +174,70 @@ async def _unsilence(self, channel: TextChannel) -> bool:
...
+                log.info(f"Can't reschedule silence for {channel_id}: channel not found.")
+                continue
+
+            if timestamp == -1:
+                log.info(f"Adding permanent silence for #{channel} ({channel.id}) to the notifier.")
+                self.notifier.add_channel(channel)
+                continue
+
+            dt = datetime.fromtimestamp(timestamp, tz=timezone.utc)
+            delta = (dt - datetime.now(tz=timezone.utc)).total_seconds()
```
Numerlor, line 226: Same here with the timezones, not strictly necessary

bot/cogs/moderation/silence.py
```diff
@@ -10,11 +11,23 @@
...
 
+MSG_SILENCE_FAIL = f"{Emojis.cross_mark} current channel is already silenced."
+MSG_SILENCE_PERMANENT = f"{Emojis.check_mark} silenced current channel indefinitely."
+MSG_SILENCE_SUCCESS = Emojis.check_mark + " silenced current channel for {duration} minute(s)."
+
+MSG_UNSILENCE_FAIL = f"{Emojis.cross_mark} current channel was not silenced."
+MSG_UNSILENCE_MANUAL = (
+    f"{Emojis.cross_mark} current channel was not unsilenced because the current "
+    f"overwrites were set manually. Please edit them manually to unsilence."
+)
```
Numerlor, line 28: This will be sent when the redis cache loses the data on a currently silenced channel, would it be worthwhile to add to the wording?

Numerlor (commented):
Code looks good so far, still need to look through the tests.

Do manual per channel overwrites for the roles happen at all to be worth accounting for in the command and not resetting them to None? Although still fairly unlikely, the behaviour will also trigger on cache loss which means an admin will need to do it manually instead

bot/cogs/moderation/silence.py
MarkKoz: See the commit message

> `datetime.timestamp()` assumes naïve `datetime`s are in local time, so
getting POSIX timestamps in UTC isn't easy for naïve ones. Technically,
the timestamp's timezone doesn't matter if all code is on the same page
and parsing it with the same timezone. Keeping it in the local timezone
would be okay then, but I feel safer locking it to UTC explicitly.

> Any reason for the parentheses here?

No, they can be removed.

MarkKoz:
> Do manual per channel overwrites for the roles happen at all to be worth accounting for in the command and not resetting them to None? 

I'm not sure, but I accounted for it for the sake of thoroughness. My thinking is that if it is set manually, the intention may be more permanent than what silence/unsilence does. Therefore, I didn't want to give a false impression by allowing `!unsilence` to undo such case.

bot/cogs/moderation/silence.py
MarkKoz: Yes, I agree.

bot/cogs/moderation/silence.py
Numerlor: I'm not sure I understand the implications of the different computation methods here, but the timestamp is inherently timezone agnostic and generated on the same machine so they should be fine no? In any case it's not a lot of additional code so I'm not against it, but don't entirely understand the reason for it

MarkKoz: That's incorrect. A timestamp is _not_ timezone agnostic. For naïve datetimes, the timestamp will be relative to the system's timezone. I did point out that it's okay since it's on the same machine, but I preferred the explicitness.

MarkKoz:
I'd actually welcome the simplification if a majority thinks it's unnecessary.

bot/cogs/moderation/silence.py
```diff
@@ -57,23 +70,30 @@ async def _notifier(self) -> None:
...
-        self.muted_channels = set()
 
-        self._get_instance_vars_task = self.bot.loop.create_task(self._get_instance_vars())
-        self._get_instance_vars_event = asyncio.Event()
+        self._init_task = self.bot.loop.create_task(self._init_cog())
 
-    async def _get_instance_vars(self) -> None:
-        """Get instance variables after they're available to get from the guild."""
+    async def _init_cog(self) -> None:
+        """Set instance attributes once the guild is available and reschedule unsilences."""
```
kwzrd, line 88: How would you feel about naming this `_async_init` or similar? I feel like that reflects the purpose better.

bot/cogs/moderation/silence.py
```diff
@@ -57,23 +70,30 @@ async def _notifier(self) -> None:
 class Silence(commands.Cog):
     """Commands for stopping channel messages for `verified` role in a channel."""
 
+    # Maps muted channel IDs to their previous overwrites for send_message and add_reactions.
+    # Overwrites are stored as JSON.
+    previous_overwrites = RedisCache()
+
+    # Maps muted channel IDs to POSIX timestamps of when they'll be unsilenced.
+    # A timestamp equal to -1 means it's indefinite.
+    unsilence_timestamps = RedisCache()
```
kwzrd, line 79: Have you considered using just a single cache?

Since we're already serializing the perms as JSON, it could be something like:
```
{
  permissions: { ... },
  timestamp: 123
}
```
For the sake of "safety" I'd probably feel tempted to write a class `Silence`, instances of which can serialize themselves into such a JSON, and can be created from one. That would create more of a "protocol" for what is allowed to exist in the cache, but I realize that I'm basically reinventing a table here with permissions and timestamps being columns. It'd also make it so that if we wanted to add more data, we'd have to "migrate" existing cache entries. Probably an awful idea, but it makes me feel a little uneasy how the caches are written to from different methods and there is no guarantee that a key exists in both.

bot/cogs/moderation/silence.py
```diff
@@ -118,12 +151,17 @@ async def _silence(self, channel: TextChannel, persistent: bool, duration: Optio
         `duration` is only used for logging; if None is passed `persistent` should be True to not log None.
         Return `True` if channel permissions were changed, `False` otherwise.
```
kwzrd, line 152: > `duration` is only used for logging

I feel like the log could be done on the outside, so that the param is unnecessary (we return the bool to signal success).

Additionally, I feel like the information represented by `persistent` and `duration` can be expressed by one value, since `duration` is already optional; if it's `None` we know that the silence is indefinite, right? I also find `persistent` to be an awkward name - on the surface it gives me the impression of telling whether the silence should be persisted.

I understand that this is not part of this PR's diff, but figured I'd mention it anyway.

bot/cogs/moderation/silence.py
kwzrd: Since the POSIX timestamp comes out as a float, maybe we could write a `-1.` for consistent types, but eh nitpick.

kwzrd (commented):
I'll try to test as soon as I can, for now I've left a few thoughts (really more thoughts than anything) from an initial read.

bot/cogs/moderation/silence.py
kwzrd: My comment here is mainly motivated by the fact that one param restricts the valid values of the other, i.e.:
> if None is passed `persistent` should be True to not log None

I generally try to avoid that and see if I can find a way to represent the information using a single value that cannot exist in an invalid state. Here I think the solution is fairly easy, unless I'm missing something.

MarkKoz: I refactored it in cf2c032

bot/cogs/moderation/silence.py
MarkKoz: Sure. See f218c7b

bot/cogs/moderation/silence.py
MarkKoz: Keeping them separate makes it easier to separate permissions code from scheduling code. The caches are not intended to be used publicly so I'm not keen on overcomplicating serialisation for some extra safety. I could add some asserts if you'd like.

MarkKoz:
There's a slightly annoying issue I need to resolve. The `Scheduler` is mocked, so when (mock) coroutines are scheduled, they don't get awaited and warnings are issued (though tests still pass).

bot/cogs/moderation/silence.py
kwzrd: Yes, nice. I think that's much better.

bot/cogs/moderation/silence.py
kwzrd: It's fine, I agree that keeping it simple probably makes more sense.

bot/cogs/moderation/silence.py
kwzrd: I was always surprised by the parens. Do you think it reads better with them?

bot/cogs/moderation/silence.py
```diff
@@ -103,61 +130,112 @@ async def unsilence(self, ctx: Context) -> None:
...
+
+        return True
+
+    async def _reschedule(self) -> None:
+        """Reschedule unsilencing of active silences and add permanent ones to the notifier."""
+        for channel_id, timestamp in await self.unsilence_timestamps.items():
+            channel = self.bot.get_channel(channel_id)
+            if channel is None:
+                log.info(f"Can't reschedule silence for {channel_id}: channel not found.")
+                continue
```
kwzrd, line 216: Should we pop the `channel_id` from the cache? If there's an invalid cache entry, it will remain in the cache forever. Although if we just fail to get the channel for some reason temporarily, losing the mapping is not great either. :man_shrugging:

bot/cogs/moderation/silence.py
```diff
@@ -103,61 +130,112 @@ async def unsilence(self, ctx: Context) -> None:
...
-            channels_string = ''.join(channel.mention for channel in self.muted_channels)
-            message = f"<@&{Roles.moderators}> channels left silenced on cog unload: {channels_string}"
-            asyncio.create_task(self._mod_alerts_channel.send(message))
+        """Cancel the init task and scheduled tasks."""
+        # It's important to wait for _init_task (specifically for _reschedule) to be cancelled
+        # before cancelling scheduled tasks. Otherwise, it's possible for _reschedule to schedule
+        # more tasks after cancel_all has finished, despite _init_task.cancel being called first.
+        # This is cause cancel() on its own doesn't block until the task is cancelled.
+        self._init_task.cancel()
+        self._init_task.add_done_callback(lambda _: self.scheduler.cancel_all)
```
kwzrd, line 238: I'm not sure I follow, what do you mean by this?

> This is cause cancel() on its own doesn't block until the task is cancelled.

Here the call to `cancel` will block until the task is guaranteed to be dead right?

bot/cogs/moderation/silence.py
```diff
@@ -57,23 +70,30 @@ async def _notifier(self) -> None:
...
+        """Set instance attributes once the guild is available and reschedule unsilences."""
         await self.bot.wait_until_guild_available()
+
         guild = self.bot.get_guild(Guild.id)
         self._verified_role = guild.get_role(Roles.verified)
         self._mod_alerts_channel = self.bot.get_channel(Channels.mod_alerts)
         self._mod_log_channel = self.bot.get_channel(Channels.mod_log)
         self.notifier = SilenceNotifier(self._mod_log_channel)
-        self._get_instance_vars_event.set()
+        await self._reschedule()
```
kwzrd, line 96: We could get rid of the `self._mod_log_channel` since it's not being accessed anywhere.

kwzrd (changes requested):

bot/cogs/moderation/silence.py
MarkKoz: No, that's a misconception. It only requests for the task to be cancelled. It stops blocking once it has arranged for the `CancelledError` to be raised. IT doesn't get raised immediately and the task can choose to catch the exception to delay or even suppress cancellation.

bot/cogs/moderation/silence.py
MarkKoz: I'd rather keep it since it may be a temporary issue. It doesn't cause any problems to leave it in there. In other places the cache is used, the channel is supplied by the channel of the context. Therefore, rescheduling is the only time the channel's validity needs to be checked.

bot/cogs/moderation/silence.py
MarkKoz: It's being passed to `SilenceNotifier` on the next line.

bot/cogs/moderation/silence.py
```diff
@@ -83,18 +103,25 @@ async def silence(self, ctx: Context, duration: HushDurationConverter = 10) -> N
...
+        await self._schedule_unsilence(ctx, duration)
 
-        self.scheduler.schedule_later(duration * 60, ctx.channel.id, ctx.invoke(self.unsilence))
+        if duration is None:
+            log.info(f"Silenced {channel_info} indefinitely.")
+            await ctx.send(MSG_SILENCE_PERMANENT)
+        else:
+            self.notifier.add_channel(ctx.channel)
+            log.info(f"Silenced {channel_info} for {duration} minute(s).")
+            await ctx.send(MSG_SILENCE_SUCCESS.format(duration=duration))
```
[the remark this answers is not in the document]
MarkKoz: I like how the refactor separated the responsibilities. One function sets permissions, one schedules an unsilence, and the command function, for the most part, just sends messages/logs. Of course, it also calls the other two functions. Considering this, the `notifier.add_channel` isn't really that out of place either.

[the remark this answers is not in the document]
MarkKoz: Fixed in d0c3990

bot/cogs/moderation/silence.py
MarkKoz: No. This code just went through several iterations. There used to be a function call somewhere in between.

bot/cogs/moderation/silence.py
kwzrd: I meant it being assigned as an attr. It's only used within the function.

MarkKoz: You're right. Removed in 5b87a27.

bot/cogs/moderation/silence.py
kwzrd: Thanks for the explanation. It also make sense, as the docs explain, that the exc will not be raised until the next cycle of the event loop, so we need to hit an `await`. I understand the code now - we add a done callback precisely because we want the `cancel_all` to be called only after the task actually dies.

bot/cogs/moderation/silence.py
kwzrd: Yea, makes sense. In practice it's very unlikely to happen anyway, unless a silenced channel is deleted or something.

bot/cogs/moderation/silence.py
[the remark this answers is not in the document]
kwzrd: Sure, it's fine and it makes sense to split it up. I just think it adds a little bit of mental overhead, having to follow the same conditional in two places, but I'm cool with it.

kwzrd (approved):
Looks good, works well. Again sorry for the time it took to get this reviewed. :disappointed: 

I've done a code review and tested the cog, I did not spend too much time on the tests, hope that's ok. For the `autospec`, it's difficult for me to give a review there, but as far as I can tell the way we patch should be safe (i.e. it won't make a test that should fail pass instead). I agree that the new behaviour should be more convenient and have no issues with the monkeypatch as long as it doesn't break tests.

Regarding the license, I'm sadly not qualified to say whether we've attributed correctly, sorry. It looks alright to my uneducated eye.

MarkKoz:
@kwzrd Thanks. Do you have any thoughts on https://github.com/python-discord/bot/pull/1113#issuecomment-696989906 ?

bot/cogs/moderation/silence.py
MarkKoz: I'm not making this change yet. I'm still waiting for thoughts on whether forcing manual unsilences is desirable.

MarkKoz:
Since I already implemented it, I'm inclined to leave it at this point. I think it will be such a rare occurrence that no one will even notice either way. And if they do, they can complain and I'll remove it.

bot/exts/moderation/silence.py
```diff
@@ -102,61 +128,112 @@ async def unsilence(self, ctx: Context) -> None:
...
         log.debug(f"Unsilencing channel #{ctx.channel} from {ctx.author}'s command.")
-        if not await self._unsilence(ctx.channel):
-            await ctx.send(f"{Emojis.cross_mark} current channel was not silenced.")
+        await self._unsilence_wrapper(ctx.channel)
+
+    async def _unsilence_wrapper(self, channel: TextChannel) -> None:
+        """Unsilence `channel` and send a success/failure message."""
+        if not await self._unsilence(channel):
+            overwrite = channel.overwrites_for(self._verified_role)
+            if overwrite.send_messages is False and overwrite.add_reactions is False:
```
kwzrd, line 139: How would you feel about making this an `or`? To get the manual message out if either was disabled manually. I think when mods silence manually, they may not always disable reactions.

MarkKoz: Changed in bfab492

bot/exts/moderation/silence.py
```diff
@@ -102,61 +129,112 @@ async def unsilence(self, ctx: Context) -> None:
...
+            else:
+                await channel.send(MSG_UNSILENCE_FAIL)
         else:
-            await ctx.send(f"{Emojis.check_mark} unsilenced current channel.")
+            await channel.send(MSG_UNSILENCE_SUCCESS)
 
-    async def _silence(self, channel: TextChannel, persistent: bool, duration: Optional[int]) -> bool:
-        """
-        Silence `channel` for `self._verified_role`.
+    async def _silence_overwrites(self, channel: TextChannel) -> bool:
```
Numerlor, line 147:
```suggestion
    async def _set_silence_overwrites(self, channel: TextChannel) -> bool:
```
To me, silence_overwrites either sounds like a variable that contains the overwrites for silences or a method that silences overwrites; not sure if adding the set or an another name would be a better choice

Numerlor (changes requested):
Currently, if two users invoke the silence command at the same time, the value of the second silence will go to the redis cache while the first value will be used in the scheduler, causing some very weird behaviour, apart from the unexpected two silences going through. 
We should use a lock on the command for the channel it was invoked in. There shouldn't be any issues from multiple unsilences as far as I can see but also using a lock there doesn't seem like a bad idea

bot/exts/moderation/silence.py
MarkKoz: See dc2797b

MarkKoz:
@Numerlor added locks in a03fe44. It's actually a good idea to have a lock on the unsilence too because I foresee race conditions between a silence and unsilence for the same channel.

Numerlor (approved):
There are some outdated docstrings in the `SilenceNotifierTests` that were missed in the last PR and still reference FirstHash (`add_channel` and `remove_channel` specifically). Seems to look good otherwise

PR #1117, opened 2020-08-20
Make client parameter mandatory for wait_for_deletion

### LINKED ISSUE #1112: `wait_for_deletion` from `bot/utils/messages.py` raises an exception if the `client` parameter is not specified, although it's optional
If the `client` parameter of `wait_for_deletion` is not specified, the function itself will set the `client` parameter, but as an instance of `discord.Member`, rather than an instance of `discord.Client` or a subclass of it. What's the problem with this? `discord.Member` object has no attribute `wait_for`, and this attribute is actually used in the function, no matter what.

Points of interest:
https://github.com/python-discord/bot/blob/4df3089d8d03f54cdbd14d7683149ae7931036c1/bot/utils/messages.py#L25 - The `client` parameter
https://github.com/python-discord/bot/blob/4df3089d8d03f54cdbd14d7683149ae7931036c1/bot/utils/messages.py#L39 - How the `client` parameter is set if `client` is `None`

MarkKoz:
A client instance is necessary for the core feature of this function. There is no way to obtain it from the other arguments. The previous code was wrong to think `discord.Guild.me` is an equivalent.

Fixes #1112

### DIFF
--- a/bot/cogs/bot.py
+++ b/bot/cogs/bot.py
@@ -337,7 +337,7 @@ class BotCog(Cog, name="Bot"):
                         self.codeblock_message_ids[msg.id] = bot_message.id
 
                         self.bot.loop.create_task(
-                            wait_for_deletion(bot_message, user_ids=(msg.author.id,), client=self.bot)
+                            wait_for_deletion(bot_message, (msg.author.id,), self.bot)
                         )
                     else:
                         return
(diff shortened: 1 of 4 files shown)


Numerlor (approved):

AtieP (approved):

Numerlor:
Works fine but figured I'd mention it since you changed it in all the other places, this usage still passes the client as a kwarg:
https://github.com/python-discord/bot/blob/47521608d573c97597df7b97bf42b0142f79e98c/bot/cogs/doc.py#L396

MarkKoz:
> Works fine but figured I'd mention it since you changed it in all the other places, this usage still passes the client as a kwarg:

Oh well.

PR #1118, opened 2020-08-20
Censor watchchannel messages when they contain leaked tokens

### LINKED ISSUE #1094: Leaked bot token still relayed to big-brother if user is watched
If a user pastes a bot token and the user is currently being watched (big-brother or nomination) then it will relay the message to the watch channel, which is not ideal. The same thing would apply for webhook URLs.
Ideally it would either send nothing, or possibly a message saying that it was removed, to the watch channels.

MarkKoz:
Fixes #1094 

Also added an ignore for webhook tokens in the code block help embed thingy. It was already ignoring bot tokens.

### DIFF
--- a/bot/cogs/bot.py
+++ b/bot/cogs/bot.py
@@ -9,6 +9,7 @@ from discord.ext.commands import Cog, Context, command, group
 
 from bot.bot import Bot
 from bot.cogs.token_remover import TokenRemover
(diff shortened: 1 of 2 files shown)


ikuyarihS (approved):
Looks good to me.

Den4200:
Thanks for the review, @Aniperk!

PR #1124, opened 2020-08-23
Support specifying top-level aliases in the command decorator

### LINKED ISSUE #707: Move aliases to their respective cogs
I think it's easier to maintain aliases if they are located near the actual commands. Furthermore, creating separate functions for aliases feels redundant and harder to maintain (particularly the parameters and their type annotations). I wonder if it's possible to create a decorator for commands which can add an alias for a command. Perhaps the command resolution of discord.py can be manipulated rather than having to create "dummy" commands for aliases (this may prove to be complicated for subcommands/groups particularly).

The point of this issue right now is to have some discussion on what a good design would be and some ideas for implementations.

MarkKoz:
Closes #707 

## Feature overview

A new keyword argument has been added to the `discord.ext.commands.GroupMixin.command` decorator. The kwarg `root_aliases` allows a sequence of alias names to be specified which are meant to act as top-level commands. For example,

```py
@site_group.command(name="rules", aliases=("r", "rule"), root_aliases=("rules", "rule"))
```

### Help command

These aliases also show up in the help command (they're sorted too)

bild

### Implementation

There is no separate command created. They're implemented quite similarly to normal aliases, which are mappings of several names to the same command object.

## Remaining aliases

There are a couple of the old aliases left:

1. A tag alias. It could easily be implemented inside the tags cog as a mapping that the command checks before looking up the tag.
2. Reverse aliases for `docs get` and `tags get`: `get docs` and `get tags`. I played around with the idea of removing the `get` aliases, but they have some complications. 

I'd really like to get rid of them. Perhaps they should be removed without any replacement, as I doubt these aliases are used often. Tags are more convenient to fetch directly with the fuzzy search. Docs are more convenient through the `!d <symbol name>` alias.

### DIFF
--- a/bot/__init__.py
+++ b/bot/__init__.py
@@ -2,10 +2,14 @@ import asyncio
 import logging
 import os
 import sys
+from functools import partial, partialmethod
 from logging import Logger, handlers
 from pathlib import Path
 
 import coloredlogs
+from discord.ext import commands
+
+from bot.command import Command
 
 TRACE_LEVEL = logging.TRACE = 5
 logging.addLevelName(TRACE_LEVEL, "TRACE")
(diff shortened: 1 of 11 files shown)


Senjan21 (approved):
Looks good :+1:

ikuyarihS (approved):
Looks good to me!

PR #1126, opened 2020-08-25
Whitelist the modmail category for displaying verbose infraction info from the user command

### LINKED ISSUE #1125: Whitelist modmail threads as mod channels
Currently Modmail threads aren't marked as mod channels. This means that invoking the `!user` command does now show any existing notes on a user. This means that the moderator then either has to make an explicit infraction search just to make sure, or to move to a mod channel and invoke the command there (which is less desirable as we usually keep all relevant actions inside the Modmail channel).

The suggestion is to therefore whitelist Modmail channels as mod channels. As these channels are dynamic there is no static way to do so. A solution would then be to add a functionality which allows whitelisting an entire category.
This, however, will have the adverse affect of including the incidents channels which are open to Helpers. It would then be necessary to be able to exclude specific channels in the category whitelisting.

In all the proposed solution is to allow whitelisting of entire categories with the exclusion of specific channels.

MarkKoz:
Fixes #1125

I also did a small refactor that switches out the command's redirect code with a `in_whitelist_check`.


MarkKoz:
The staff channels constant is no longer used anywhere. Should it be deleted?

lemonsaurus:
> The staff channels constant is no longer used anywhere. Should it be deleted?

I don't see why not. No point in keeping stuff that we're not using.

kosayoda (approved):
Works good.

Numerlor (approved):

PR #1135, opened 2020-08-29
Auto-respond to everyone pings.

### LINKED ISSUE #1131: Auto-respond to non-mod everyone pings.
# The issue:
Currently, if someone puts down an `@everyone` ping, a mod has to come and respond to it to tell the person not to ping 95k+ people. Now, this usually works fine, but accidents happen, and things can end up happening like lemon accidentally pinging everyone, because android doesn't have a check for that (we're not upset lemon!).

# Proposed solution:
We add a auto-respond to the bot, sort of like in the attachment auto-respond, that contacts the user to explain why they shouldn't ping 95k people, and maybe even temp-mutes them and/or deletes the message. Here is an example message:
```
Sorry @user, but please don't try to ping 95k people. It will not have good results.
If you want to see what can happen, take a look at this pin: 
https://discordapp.com/channels/267624335836053506/267624335836053506/748352909904969740
```
It should only respond to non-staff too.

MrAwesomeRocks:
# Everyone Ping Auto-Responder
### Closes #1131 

The result of sending an `@everyone` ping would look like this: 
result of ping in server
The user would also get DMed the default message for an anti-spam rule:
result of ping in DM

Also, this filter detects ***all*** cases of `@everyone` pings.

## Notes:
Set the `max` value in the config file to zero for this to work.

The server member count updates automatically.


bot/rules/everyone_ping.py
```diff
@@ -0,0 +1,37 @@
...
+
+async def apply(
+    last_message: Message,
+    recent_messages: List[Message],
+    config: Dict[str, int],
+) -> Optional[Tuple[str, Iterable[Member], Iterable[Message]]]:
+    """Detects if a user has sent an '@everyone' ping."""
+    relevant_messages = tuple(msg for msg in recent_messages if msg.author == last_message.author)
+
+    ev_msgs_ct = 0
```
lemonsaurus, line 17: Please don't use cryptic, abbreviated variable names like this.

MrAwesomeRocks:
Changed the message to show the full member count, not just thousands: 
ping response

### Notes:
- This value will be the actual member count.
- The number will also have a comma every three digits.

bot/rules/everyone_ping.py
```diff
@@ -0,0 +1,37 @@
...
+    for msg in relevant_messages:
+        if "@everyone" in msg.content:
+            ev_msgs_ct += 1
+
+    if ev_msgs_ct > config["max"]:
+        # Send the user an embed giving them more info:
+        embed_text = textwrap.dedent(
+            f"""
+            Please don't try to ping {last_message.guild.member_count:,} people.
+            **It will not have good results.**
```
lemonsaurus, line 27: I don't like the `It will not have good results` part of this, that feels like we're saying it's okay to ping 95000 people but it won't get you the results you were hoping for.

Instead of this second line, maybe we can give this embed a headline. For example, you could use a random headline from `bot.constants.NEGATIVE_REPLIES`?

bot/rules/everyone_ping.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        )
+        embed = Embed(description=embed_text, colour=Colours.soft_red)
+        await last_message.channel.send(f"Hey {last_message.author.mention}!", embed=embed)
```
lemonsaurus, line 31: Feels weird to me that we're pinging the user before the embed. It makes the embed sandwiched between two plaintext lines, and also makes us ping the user twice instead of just once. I don't think we need it.
```suggestion
        await last_message.channel.send(embed=embed)
```

bot/rules/everyone_ping.py
lemonsaurus: With the second line removed, this can just be a regular f string. There's no need for textwrap dedent.

lemonsaurus (changes requested):
I have a few minor suggestions here.

bot/rules/everyone_ping.py
MrAwesomeRocks: By `It will not have good results`, I was trying to say that people would get upset if you pinged them.

I'll look at `bot.constants.NEGATIVE_REPLIES`, didn't know that existed.

MrAwesomeRocks:
The bot now responds like this:
image

bot/rules/everyone_ping.py
```diff
@@ -0,0 +1,42 @@
...
+    everyone_messages_count = 0
+    for msg in relevant_messages:
+        if "@everyone" in msg.content:
+            everyone_messages_count += 1
+
+    if everyone_messages_count > config["max"]:
+        # Send the user an embed giving them more info:
+        embed_text = textwrap.dedent(
+            f"""
+            **{random.choice(NEGATIVE_REPLIES)}**
```
ks129, line 27: I'd suggest to make this as embed title

MrAwesomeRocks:
`NEGATIVE_REPLIES` now moved to title of embed instead of bold.
ping reply embed

lemonsaurus:
@MrAwesomeRocks 
Remember to resolve conversations in this thread when you've addressed them.

MrAwesomeRocks:
@lemonsaurus 
Didn't know about that. All the conversations have been resolved.

lemonsaurus (changes requested):
You've implemented this as a frequency filter, which I find a bit weird. It's not really relevant for our community to be able to configure this filter to allow only 2 `@everyone` per 10 seconds, for example. I can't see why anyone would ever need that.

Perhaps we can just simplify this pull request by removing the frequency aspect completely from this filter, and removing the corresponding configuration option, too.

MrAwesomeRocks:
@lemonsaurus 
The issue is, to implement this as an anti-spam rule, in anitspam.py, there is this code that checks for those two config options: 
```python
for required_key in ('interval', 'max'):
    if required_key not in config:
        log.error(
            f"`{required_key}` is required but was not "
                f"set in rule `{name}`'s configuration."
            )
            validation_errors[name] = f"Key `{required_key}` is required but not set for rule `{name}`"
```
And the `interval` rule is used to determine how often to check messages: 
```python
# Fetch the rule configuration with the highest rule interval.
max_interval_config = max(
    AntiSpamConfig.rules.values(),
    key=itemgetter('interval')
)
max_interval = max_interval_config['interval']
```
Which is why I have this implemented as a frequency filter.

bot/rules/everyone_ping.py
```diff
@@ -0,0 +1,36 @@
...
+
+    if everyone_messages_count > config["max"]:
+        # Send the user an embed giving them more info:
+        embed_text = f"Please don't try to ping {last_message.guild.member_count:,} people."
+
+        # Make embed:
+        embed = Embed(title=random.choice(NEGATIVE_REPLIES), description=embed_text, colour=Colours.soft_red)
+
+        # Send embed:
+        await last_message.channel.send(embed=embed)
```
kwzrd, line 30: I do not believe this comment is necessary, it just repeats what the code says. Similar above with "Make embed".

kwzrd (commented):
The current solution will punish me if I send a message containing `@everyone` as a markdown codeblock, which is unfortunate because there is clearly no intention to ping, e.g. when discussing this feature.

I also believe this will not react to pings constructed as `<@&{guild_id}>`.

Edit: It looks like Discord may not be allowing pinging the everyone role with this syntax, although it almost looks a bit like a bug? Either way, if someone does send that string, I'd argue that there is a clear intention to ping.

lemonsaurus:
okay, after looking at antispam.py a bit, I guess refactoring to allow non-frequency spam filters seems like it was would be out of scope.

I guess we will go with this solution.

MrAwesomeRocks:
- Switched to regex to avoid punishing users for codeblocks.
- Also now catches `<@&{guild_id}>` style pings.
- Deleted the redundant comments.
- Changed the message to be less like a frequency anti-spam rule: 
everyone ping message reply

MrAwesomeRocks:
My regex was catching some incorrect strings, such as ```markdown

MrAwesomeRocks:
My regex was catching some incorrect strings, such as 
```markdown
```@everyone
here is some text
here is more text```
``` 
Now the regex doesn't catch those incorrect strings.

bot/cogs/antispam.py
```diff
@@ -34,7 +34,8 @@
     'links': rules.apply_links,
     'mentions': rules.apply_mentions,
     'newlines': rules.apply_newlines,
-    'role_mentions': rules.apply_role_mentions
+    'role_mentions': rules.apply_role_mentions,
+    'everyone_ping': rules.apply_everyone_ping,
```
Aniperk, line 38: line 38: at the end im pretty sure there doesn't need to be a comma, probably just nit picking here :)

kwzrd: This diff is a good example of how the trailing comma is useful. Because line `37` did not have the trailing comma, it had to be changed in order to add another element.

lemonsaurus: Yep, leave this trailing comma in - in fact, it's encouraged by PEP8 as well.

Aniperk: Alright, i was just thinking it might not need to be there, im just someone who works in environments where using as least space as possible is the goal. :)

MrAwesomeRocks:
All of kwzrd's concerns about punishing users who put `@everyone` in codeblocks should be fixed now.

MrAwesomeRocks:
Wait I need to fix some regex.

MrAwesomeRocks:
Now it catches `@everyone` with text next to it, such as `hello@everyone`.

lemonsaurus:
@MrAwesomeRocks 

How will this affect staff members? They should be immune to it. Can you confirm that they will be? I think they're immune to all anti-spam, but I'd just like a confirmation.

LordHemlock (approved):
Seems to work as intended in testing.  Giving it the green light.

lemonsaurus (approved):
Alright, seems fine now. let's get this merged.

PR #1144, opened 2020-09-04
Move bolding markdown outside of text link in help channel messages.

Numerlor:
On some clients (android), the markdown is commonly not rendered properly within text links, in this case the bolding didn't happen and the markdown asterisks remainder as plaintext.
Moving them outside of the link text fixes the issue


MarkKoz (approved):
For what it's worth, both on Windows and Android, it didn't render as bold before your fix. However, unlike the screenshot, the asterisk weren't visible either.

Numerlor:
> For what it's worth, both on Windows and Android, it didn't render as bold before your fix.

Hmm, it does render properly for me on windows; I guess it's inconsistent all over.

ikuyarihS (approved):
LGTM

PR #1145, opened 2020-09-05
Fix @everyone ping rule.

### LINKED ISSUE #1142: Everyone Ping rule picking up false positives
The everyone ping rule has been picking up some false positives, such as 
```markdown
`a`@everyone`b`
```
or 
```markdown
Hello`@everyone`
```
The regex needs to be redone.

kosayoda:
This PR modifies the regex in the everyone ping rule to avoid false positives and negatives. Test cases have also been added to enforce correctness.

### Implementation
The check now removes all codeblocks, inline or multiline from the message before detecting pings. As it uses `re.sub`, there may be a performance concern. If that turns out to be the case, we can do a basic ping inclusion check, then strip codeblocks if the first filter passes.

### Tests
~~There may be more edge cases that aren't checked, feel free to add/modify the tests if necessary.~~
Tests have been removed as the rule has been moved to the Filtering cog, which will be split up and overhauled after this PR is merged.

### Reviewing
Functionality/correctness should have priority, code quality less so as the there will be a Filtering rework soon.

Fixes #1142.


fadedmax (approved):
LGTM

MrAwesomeRocks (changes requested):
I would add a check for `@everyone` in messages before `re.sub`, just to make this as robust as possible, because we can't really test this in a server of 98k+ without deploying it, and it would be bad if the bot got lagged out during a raid or something.

lemonsaurus (changes requested):
As we discussed during the admin meeting today, we'd like to see the automod stuff completely removed from the feature, keep the mod-alert, and make it work for @here too.

kosayoda:
Going by that, does it make sense to move it from `antispam` to `filtering`? The former is rigid on pinging/infracting, and I think it suits the latter more.

lemonsaurus:
> does it make sense to move it from antispam to filtering?

Yes I think it does.

MrAwesomeRocks:
Looking at `filtering`, that file is getting very long (583 lines) because of all the different rules. The `filtering` rules also seem to be organized similarly to the `antispam` rules.

Would it make sense to move the rules out of `filtering` and into a `rules` directory somewhere, sort of like in `antispam`? It would make it simpler to add rules later on. 
It could be out of the scope of this PR to change `filtering` that much, though.

lemonsaurus:
> Would it make sense to move the rules out of filtering and into a rules directory somewhere, sort of like in antispam?

Yeah, I think we should probably consider doing that - maybe `rules` could have a subfolder for `filtering` and one for `antispam`. But yeah it's probably out of scope for this PR. You could make an issue for it if you want, @MrAwesomeRocks ?

MrHemlock (approved):
Everything looks solid to me.

lemonsaurus (approved):
I feel like it's a bit dry without the "don't try to ping 95000 people" line, but I suppose that was removed because it no longer makes sense with `@here`?

Anyway, not a showstopper for me, looks good to me.

PR #1148, opened 2020-09-06
New guilds tag

Jack92829:
A new tag named **guilds** that directs people towards the [awesome list](https://github.com/mhxion/awesome-discord-communities) found under the [communities](https://pythondiscord.com/pages/resources/communities/) section of the python discord's site and to partners of the python discord. I believe this will come in handy both when people ask for help regarding other languages and fields and for when people ask if there is any guilds similar to the python discord but that are focused on other languages, tools etc.


AtieP (approved):

Akarys42 (approved):
This tag will come in handy, thanks!

SebastiaanZ:
This tag is different from the existing tags we have. All of our other tags are either directly related to the community (e.g. `off-topic` and `modmail`) or touch upon Python concepts. The way this tag is worded feels a bit out of scope for our tag system.

If we want to have a tag that is specifically meant to be used when people ask for programming servers in other languages, I'd focus the tag on that instead of presenting it as a more disconnected "looking to join more servers? look here" kind of thing.

Jack92829:
I got a few ideas, I'll see what I can come up with

Den4200:
I think this tag is too long and over-explained. There is no need to point to different sections of the awesome discord communities list and I don't think we need to explain that our community is dedicated to Python. I believe that something short and concise, like our `!resources` or `!projects` tags would be more desirable.

image

image

For reference, this is what your tag currently looks like in an embed:
image

MrAwesomeRocks (approved):
This PR looks good to me! My only suggestion would be to possibly rename the tag to `communities` instead of `guilds`, but either way this tag looks good.

Den4200 (commented):
The word `communities` is used too repetitively, but I'm not sure what word could replace it.

mbaruh (commented):
You could avoid the extra `communities` if you reword it to be something like

**Communities**

The [communities page](https://pythondiscord.com/pages/resources/communities/) on our website contains a number of communities relating to programming and technology, including those we have partnered with as well as a [curated list](https://github.com/mhxion/awesome-discord-communities) of many others.

PR #1150, opened 2020-09-07
Implement with(out)_role checks and decorators using has_any_role

### LINKED ISSUE #217: remove with_roles decorator and replace with has_any_role

MarkKoz:
Resolves #217

This simplifies the implementations by relying on the function discord.py provides. `with_role` is entirely replaced by `has_any_role` since it was identical. The rest had their implementations changed. Consequently, they now support specifying roles as names in addition to IDs. They were also renamed to be more consistent with `has_any_role`'s name.


Senjan21 (approved):
Code looks very solid :+1:

Numerlor (approved):

PR #1151, opened 2020-09-07
Sync: remove confirmation

### LINKED ISSUE #1075: Should the syncer confirmation be removed?
The current threshold is way too low, but some of us have doubts on its usefulness. The `wait_until_guild_available` solution should prevent the cache from ever being empty. It even explicitly checks that the cache isn't empty and avoids setting the event. That being said, there may be some strange situation in which the bot receives a partial cache from Discord.

If it's not removed, then the thresholds should at least be increased. I also had the idea of only applying the size check to the deleted diffs and not the created or updated ones. What would be good thresholds for roles and users? I think it's currently set to 10 for both. 10 is probably good for deleted roles, but for users, it'd need to be somewhere in the thousands considering we have 70k members.

MarkKoz:
The confirmation was intended to be a safe guard against cache issues that would cause a huge number of roles/users to deleted after syncing. With `wait_until_guild_available`, such cache issue shouldn't arise. Therefore, this feature is obsolete.

Resolve #1075


bot/cogs/sync/syncers.py
```diff
@@ -217,11 +60,14 @@ async def sync(self, guild: Guild, ctx: t.Optional[Context] = None) -> None:
 
             # Don't show response text because it's probably some really long HTML.
             results = f"status {e.status}\n```{e.response_json or 'See log output for details'}```"
-            content = f":x: {mention}Synchronisation of {self.name}s failed: {results}"
+            content = f":x: Synchronisation of {self.name}s failed: {results}"
         else:
-            results = ", ".join(f"{name} `{total}`" for name, total in totals.items())
+            diff_dict = diff._asdict()  # Ugly method for transforming the NamedTuple into a dict
```
Numerlor, line 65: The comment doesn't add much here, I think we should just remove it.

bot/cogs/sync/syncers.py
```diff
@@ -191,24 +50,8 @@ async def sync(self, guild: Guild, ctx: t.Optional[Context] = None) -> None:
         """
         log.info(f"Starting {self.name} syncer.")
 
-        message = None
-        author = self.bot.user
-        if ctx:
-            message = await ctx.send(f"📊 Synchronising {self.name}s.")
-            author = ctx.author
-
+        message = await ctx.send(f"📊 Synchronising {self.name}s.") if ctx else None
```
Numerlor, line 53: Personal opinion here so I'd like to hear other opinions, but I don't like the send in a ternary too much; with something like this or message initialized before the check to None being clearer to me
```suggestion
        if ctx:
            message = await ctx.send(f"📊 Synchronising {self.name}s.")
        else:
            message = None
```

Numerlor (commented):
Looks good, just a couple of style points. While we're in the syncers file, would it be worth changing the `namedtuple`s to `NamedTuples`?

bot/cogs/sync/syncers.py
MarkKoz: The suggestion is just as clear, if not clearer, so I don't think this needs much discussion.

bot/cogs/sync/syncers.py
MarkKoz: I recall it was requested by @lemonsaurus 🤷

MarkKoz:
> would it be worth changing the namedtuples to NamedTuples?

I don't think so. It's internal and the individual attributes aren't really used directly besides the ID. `_Diff` would have to be generic if done properly, which significantly complicates things.

bot/cogs/sync/syncers.py
Numerlor: Don't think there's anything ugly about it with since it is a properly documented method with a clear name https://docs.python.org/3.8/library/collections.html#collections.somenamedtuple._asdict

Numerlor (approved):

bot/cogs/sync/syncers.py
lemonsaurus: I agree with Numerlor. Can't really remember talking about this with you, but I forget a lot of stuff like this so maybe. Maybe it's because of the leading underscore? It is a bit ugly but I also think it's crystal clear.

MarkKoz: https://github.com/python-discord/bot/pull/711#discussion_r383047221

Anyway, I'll remove it

kwzrd (changes requested):
The `sync` docstring hasn't been updated, we can probably just keep the summary:

https://github.com/python-discord/bot/blob/9a90d6b16e0ec61c023a916ec58d05b6142a6e2d/bot/exts/backend/sync/_syncers.py#L43-L50

Otherwise looks good.

kwzrd (approved):

PR #1157, opened 2020-09-15
Reddit: restrict nsfw subreddit(s)

### LINKED ISSUE #1153: !reddit command doesn’t filter nsfw subreddits.
So yeah, as the title says `!reddit` command doesn’t filter nsfw subreddits ._.

doing: `!reddit top nsfw` displays the direct URL link to the nsfw subreddit.

RohanJnr:
closes #1153 

### Changes made to Reddit Cog:
- Check if the subreddit is NSFW or similar by using the `over_18` boolean value got with the API response.
 - if True, then send an error message.
 - if False, the fetched data is sent to the user.

- A small change to the return format of the `fetch_posts()` method, instead of returning an empty `list` in case of an error, it now returns a list, with a `dict` containing the error message. This is done in order to distinguish between the NSFW restriction error message and network/connection/failed to connect to API error.

### Pending

- Unfortunately, I cannot test this as my account is restricted to view any NSFW subreddits due to age and country (whenever I send a request to an NSFW subreddit, I get a 451 status code - Unavailable For Legal Reasons).
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/451
- If anyone above 18 years of age, having a  Reddit account, and living in a country that allows users to view NSFW content on Reddit, can test this feature, it will be helpful!

**Edit: this should be working.**


Numerlor:
I believe filtering posts and only sending the NSFW error message when no posts pass the filter would be a better option here, currently if the first fetched post happens to be NSFW while none of the others are, the user will get no output.

RohanJnr:
I think a better approach will be to check if any 1 of the posts does not pass the filter, then send nothing.
This is because, if we query for NSFW subreddit posts from the bot, then we don't want to send any posts, even tho is post is not entirely NSFW, we want to block the subreddit.

kosayoda (changes requested):
Thanks for the pull request!

1. Regarding NSFW subreddits: Subreddits marked as over 18 will have all the posts marked over 18, so it shouldn't be an issue.
2. Rather than telling the user about the subreddit being out of scope, the error message when no posts are found should just be `... couldn't find any SFW posts from that subreddit...`. That way it is clearer and easier to handle in the code.
3. Error handling should be in `get_top_posts` rather than `fetch_posts`, which would basically be what it is originally.

bot/exts/info/reddit.py
```diff
@@ -163,12 +166,11 @@ async def get_top_posts(self, subreddit: Subreddit, time: str = "all", amount: i
             amount=amount,
             params={"t": time}
         )
-
```
Numerlor, line 166: Any reason for the removal of the newline here?

kosayoda (approved):
Works well in testing now, awesomesauce.

kwzrd (approved):
Works great, thanks for the fix. There is a good opportunity to test against [r/test](https://old.reddit.com/r/test)'s top weekly posts:

image

Also works against NSFW-only subs.

kwzrd:
Since this is somewhat critical I decided to merge it even with an unresolved comment from @Numerlor. It only applies to code style and can be addressed separately if necessary. Feel free to continue the conversation.

Issue #673, opened 2019-11-30
Duck Pond improvements


lemonsaurus:
After merging the duck pond, a few minor bugs and potential for improvements have been uncovered. Please make a PR implementing the following changes:

- [x] A blacklist for channels that duck pond should ignore. This blacklist must contain #duck-pond itself, #announcements, #staff-announcements and #user-event-announcements.
- [x] A command for forcing a message into the duck pond. something like `!duckpond`, alias `!pondify`, `!duckpondify` and `!duckify`
- [ ] New test for making sure it returns early if in a blacklisted channel.
- [x] New test for the command.
- [x] Restrict the `on_raw_reaction_add` handler it so only messages posted by staff members will be duckpondified.
- [x] Prevent the duck pond from relaying the same message more than once during lag by adding a lock.

sco1:
At what level should the new command be role locked to?

SebastiaanZ:
>  A command for forcing a message into the duck pond. something like !duckpond, alias !pondify, !duckpondify and !duckify

What's the use of this? We discussed such a command before, but the argument was that it was only going to be used for historical messages which can easily be relayed using internal eval on the existing `relay_message` method of the class.

lemonsaurus:
I guess the use is to not have to spend time figuring out how to internal eval it.

and role lock would be admins+ imo

SebastiaanZ:
I've added an action point to implement a lock for the message relay.

The reason is that, on occasion, the same message gets relayed more than once due to a race condition: If, for some reason, multiple reaction events fire when we go over the ducking threshold and all of them make it past the "green checkmark check" before a relay is successful, all of them will relay the message as none of them are "aware" that the message is already going to be relayed by one of the other near-simultaneous events.

The solution is straightforward: Add an [`asyncio.Lock`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock) and have the event handler acquire it before it checks for the green checkmark and after it has successfully added the greencheck mark after relaying the message. That way, any additional events have to wait with their green checkmark check until after the previous events has potentially added it.

MarkKoz:
@SebastiaanZ Any updates on this?

SebastiaanZ:
This has been partially implemented. The "New test for making sure it returns early if in a blacklisted channel." was ignored; instead, I removed all tests.

PR #1164, opened 2020-09-20
Several improvements to duckpond

SebastiaanZ:
This PR brings several improvements to our duckpond, including those suggested in #673. It also does away with the over-complicated tests we'd written for the duckpond that only really served to slow down development. I suggest reviewing the PR commit by commit, as I tried to document the changes in each commit message. I've also summarized the changed below.

### Changes
- The duckpond should now longer suffer from a relay race condition. Say goodbye to duplicate messages in our pond!
- We no longer have to maintain a list of ducky emojis: We will now count every emoji that starts with `ducky_`.
- There's now an admins-only command to force-relay a message to the pond: `!duckify <message>`.
- The duckpond will now ignore non-staff messages for privacy reasons.
- The duckpond threshold is experimentally lowered to four to increase activity: hype!

### Considerations
- All unit tests were removed. If we want to introduce new tests in the future, my recommendation is to include blackbox tests that only test input/output/result instead of the tests that focused on implementation details within the functions.


Senjan21 (approved):
Code looks solid, very nice.

Den4200 (approved):
Excellent commits and code! I really enjoyed reviewing this one. Everything looks good to me.

PR #1165, opened 2020-09-21
Smart syncing of Users

### LINKED ISSUE #1119: Smart syncing of users to the site
Currently when we sync the users to the site, we sync all 90,000 users one at a time. This is pretty much the single worst way we could possibly do this, so let's make it smarter.

See the related site-side changes required for this here: https://github.com/python-discord/site/issues/375

## Paginated retrievals
In https://github.com/python-discord/site/issues/375, I propose adding pagination to the user endpoint via DRFs built-in pagination feature. This means we'll need to make a smarter way to handle this on the bot side so that we will make one request per page.

This will be more scalable - right now there's no limit to how large the response from a request for all the users will be, and we don't want to run into issues where it's too big to handle.

## Only update the diff
Currently, we just sync all the users. We don't need to do this, if we just figure out the diff between the users we've retrieved and the users in our cache, we can just update the users who have actually changed. To do this, we can probably use 
[difflib.SequenceMatcher](https://docs.python.org/3/library/difflib.html#difflib.SequenceMatcher) to find the diff between the two and then send the diff to the API.

## Send updates in batches
Instead of updating users one at a time, we should send them in batches. In https://github.com/python-discord/site/issues/375, I propose allowing updates in batches using DRFs built in `many` option for the serializers. This should then allow us to POST big batches of users at a time. I suggest we limit these batches to 10000 at a time, so as not to overwhelm the endpoint.

RohanJnr:
Closes #1119 

## Implementation

- [x] Refactor Code to support paginated response.
- [x] Bulk Update Users whose properties have changed.
- [x] Update Users via PATCH method by only sending user ID and modified user details.

## Details
- This PR is directly linked with the Site repo PR https://github.com/python-discord/site/pull/378

- This PR uses the changes made in the above-mentioned PR.

**Only merge this PR after the Site PR is merged.**


RohanJnr:
This PR will be ready for review when the related site PR has been merged.
https://github.com/python-discord/site/pull/378

bot/exts/backend/sync/_syncers.py
```diff
@@ -336,12 +353,47 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
+
+        # The `response` is paginated, hence check if next page exists.
+        if (next_page_url := response["next"]) is not None:
+            next_endpoint, query_params = self.get_endpoint(next_page_url)
+            users.extend(await self._get_users(next_endpoint, query_params))
+
+        return users
+
+    @staticmethod
+    def get_endpoint(url: str) -> tuple:
```
ks129, line 370: I think it's better to use `typing.Tuple` for this.

bot/exts/backend/sync/_syncers.py
```diff
@@ -336,12 +353,47 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
         """Synchronise the database with the user cache of `guild`."""
         log.trace("Syncing created users...")
-        for user in diff.created:
-            await self.bot.api_client.post('bot/users', json=user._asdict())
-
-        log.trace("Syncing updated users...")
-        for user in diff.updated:
-            await self.bot.api_client.put(f'bot/users/{user.id}', json=user._asdict())
+        if diff.created:
+            created: list = [user._asdict() for user in diff.created]
```
ks129, line 395: This have already [] and this indicates that this is list. I think type hint in this variable is unnecessary.

bot/exts/backend/sync/_syncers.py
```diff
@@ -336,12 +353,47 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
+            await self.bot.api_client.post("bot/users", json=created)
+        if diff.updated:
+            updated: list = [self.patch_dict(user) for user in diff.updated]
```
ks129, line 398: Same as L395

bot/exts/backend/sync/_syncers.py
```diff
@@ -336,12 +353,47 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
-        for user in diff.created:
-            await self.bot.api_client.post('bot/users', json=user._asdict())
-
-        log.trace("Syncing updated users...")
-        for user in diff.updated:
-            await self.bot.api_client.put(f'bot/users/{user.id}', json=user._asdict())
+        if diff.created:
+            created: list = [user._asdict() for user in diff.created]
+            await self.bot.api_client.post("bot/users", json=created)
+        if diff.updated:
```
ks129, line 397: Looks like you removed log of users updating. I think it's better to restore it.

bot/exts/backend/sync/_syncers.py
```diff
@@ -336,12 +353,47 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
+
+        # Query params.
+        params = parse_qsl(url.query)
+
+        return endpoint, params
+
+    @staticmethod
+    def patch_dict(user: _User) -> dict:
+        """Convert namedtuple to dict by omitting None values."""
+        user_dict: dict = {}
```
ks129, line 385: `{}` already indicates that this is dict, no need for type hint.

bot/exts/backend/sync/_syncers.py
```diff
@@ -336,12 +353,47 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
+        # Do not include starting `/` for endpoint.
+        endpoint = url.path[1:]
+
+        # Query params.
+        params = parse_qsl(url.query)
+
+        return endpoint, params
+
+    @staticmethod
+    def patch_dict(user: _User) -> dict:
```
ks129, line 383: Use `typing.Dict`and provide key-value too.

ks129 (changes requested):
Currently some code style/small changes, but there is one big issue: Type hinting variables that have very clear type is pointless.

bot/exts/backend/sync/_syncers.py
```diff
@@ -183,12 +200,47 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
+        """Extract the API endpoint and query params from a URL."""
+        url = urlparse(url)
+
+        # Do not include starting `/` for endpoint.
+        endpoint = url.path[1:]
+
+        # Query params.
+        params = parse_qsl(url.query)
+
+        return endpoint, params
```
MarkKoz, line 226: Considering customising the pagination style on the API side to avoid having to parse URLs. An ideal style would be one that only returns page numbers rather than entire links.

bot/exts/backend/sync/_syncers.py
```diff
@@ -183,12 +200,47 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
+
     async def _sync(self, diff: _Diff) -> None:
         """Synchronise the database with the user cache of `guild`."""
         log.trace("Syncing created users...")
-        for user in diff.created:
-            await self.bot.api_client.post('bot/users', json=user._asdict())
-
+        if diff.created:
+            created = [user._asdict() for user in diff.created]
+            await self.bot.api_client.post("bot/users", json=created)
```
MarkKoz, line 242: Leave some space between these to make it more readable.

```suggestion
            await self.bot.api_client.post("bot/users", json=created)
            
```

bot/exts/backend/sync/_syncers.py
```diff
@@ -161,17 +163,32 @@ async def _get_diff(self, guild: Guild) -> _Diff:
 
         for db_user in db_users.values():
             guild_user = guild_users.get(db_user.id)
+
             if guild_user is not None:
                 if db_user != guild_user:
```
MarkKoz, line 168: This is redundant and inefficient since you'll compare each attribute again when this is true.

bot/exts/backend/sync/_syncers.py
```diff
@@ -161,17 +163,32 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
-                    users_to_update.add(guild_user)
+                    fields_to_none: dict = {}
+
+                    for field in _User._fields:
+                        # Set un-changed values to None except ID to speed up API PATCH method.
+                        if getattr(db_user, field) == getattr(guild_user, field) and field != "id":
+                            fields_to_none[field] = None
+
+                    new_api_user = guild_user._replace(**fields_to_none)
+                    users_to_update.add(new_api_user)
```
MarkKoz, line 177: It's clear that a namedtuple is no longer a sufficient data structure for updates. Setting attributes to None only to remove them later is a poor design. As a comment states, namedtuples were used to facilitate easy comparison between guild and DB users. However, since you now compare attributes individually, you no longer benefit from the `tuple != other_tuple` convenience.

I don't believe the namedtuples have any advantages outside this function, so you could return a `_Diff` containing dictionaries rather than namedtuples (for roles too, for consistency).

MarkKoz: I was too tempted to come up with something more efficient, so here it is

```py
users_to_create = set()
users_to_update = set()
seen_guild_users = set()

async for db_user in self._get_users():
    updated_fields = {}

    def maybe_update(db_field, guild_value):
        if db_user[db_field] != guild_value:
            updated_fields[db_field] = guild_value

    if (guild_user := guild.get_member(db_user["id"])):
        seen_guild_users.add(guild_user.id)

        maybe_update("name", guild_user.name)
        maybe_update("discriminator", int(guild_user.discriminator))
        maybe_update("in_guild", True)

        guild_roles = [role.id for role in guild_user.roles]
        if set(user_dict["roles"]) != set(guild_roles):
            updated_fields["roles"] = guild_roles
    elif db_user["in_guild"]:
        updated_fields["in_guild"] = False
    
    users_to_update.add(updated_fields)

for member in guild.members:
    if member.id not in seen_guild_users:
        new_user = dict(
            id=member.id,
            name=member.name,
            discriminator=int(member.discriminator),
            roles=[role.id for role in member.roles],
            in_guild=True            
        )
        users_to_create.add(new_user)
```

Uses way less memory (not a big deal, but still nice) and has less loops.

MarkKoz (changes requested):

bot/exts/backend/sync/_syncers.py
RohanJnr: Sets cannot take in dicts as they are hashable, so I could use a list and check before appending or I convert the dict to a tuple to storing it in the set().

bot/exts/backend/sync/_syncers.py
```diff
@@ -136,111 +130,67 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
+        for member in guild.members:
+            if member.id not in seen_guild_users:
+                new_user = {
+                    "id": member.id,
+                    "name": member.name,
+                    "discriminator": int(member.discriminator),
+                    "roles": [role.id for role in member.roles],
+                    "in_guild": True
+                }
+                if new_user not in users_to_create:
```
MarkKoz, line 171: You don't need to check this. I'm pretty sure we can safely assume `guild.members` will not contain duplicates.

bot/exts/backend/sync/_syncers.py
```diff
@@ -136,111 +130,67 @@ async def _get_diff(self, guild: Guild) -> _Diff:
...
-            users.extend(await self._get_users(next_endpoint, query_params))
-        return users
+            elif db_user["in_guild"]:
+                updated_fields["in_guild"] = False
 
-    @staticmethod
-    def get_endpoint(url: str) -> t.Tuple[str, t.List[tuple]]:
-        """Extract the API endpoint and query params from a URL."""
-        url = urlparse(url)
+            if updated_fields and updated_fields not in users_to_update:
```
MarkKoz, line 158: You don't need to check `updated_fields not in users_to_update`. It's not possible for the data returned from the DB to contain duplicates.

bot/exts/backend/sync/_syncers.py
```diff
@@ -42,11 +40,7 @@ async def _sync(self, diff: _Diff) -> None:
         raise NotImplementedError  # pragma: no cover
 
     async def sync(self, guild: Guild, ctx: t.Optional[Context] = None) -> None:
-        """
-        Synchronise the database with the cache of `guild`.
-
-        If `ctx` is given, send a message with the results.
-        """
+        """If `ctx` is given, send a message with the results."""
         log.info(f"Starting {self.name} syncer.")
```
MarkKoz, line 44: Why did you change this? The original docstring was better.

MarkKoz (changes requested):
This is almost there.

MarkKoz (changes requested):
Almost forgot: the code in my comment omitted the comments only for the sake of example. I think the comments that existed in `_get_diff` were useful and you should keep them in one form or another. They may not make 100% sense any more due to changes, so you may need to revise them a bit.

bot/exts/backend/sync/_syncers.py
RohanJnr: really confused about how that changed, will add the old one back.

MarkKoz (approved):

ks129 (approved):
Looks good for me

Issue #918, opened 2020-04-29
Upload snekbox output with code block escapes to hastebin


Numerlor:
When the snekbox output contains a pattern from the ESCAPE_REGEX ``[`\u202E\u200B]{3,}``, the otput is not displayed (and unnecessary work is done)
https://github.com/python-discord/bot/blob/fe325adc66e9621ea970abd83db58df4b792caa5/bot/cogs/snekbox.py#L160-L161
The output can still be uploaded to the paste service, as the contents won't affect anything on there, to provide it in some form to the user.

The `findall` method used can also be changed to `search` since we don't care about the contents.

As a bit of an unrelated question, what's the reason for forbidding zero width spaces?

MarkKoz:
> As a bit of an unrelated question, what's the reason for forbidding zero width spaces?
Dunno, it's always been there so I assumed it's important.

Akarys42:
I assumed the same thing during my sneak box refactoring, but according to a git blame, @gdude2002 added this in the commit 766faffa607ea973ec83590e0a28eb49be29bdb0, surely he would know why 😁

gdude2002:
Nope, can't remember. All I know is there was a good reason for it. 

This is a two year old commit after all!

Numerlor:
Well, if there was a good reason then perhaps adding `\u200c` (ZERO WIDTH NON-JOINER)  and `\u200d` (ZERO WIDTH JOINER) to the characters would also make sense

PR #1167, opened 2020-09-21
Upload eval output with codeblock escapes to pastebin

Numerlor:
closes: #918 

Attempts to upload eval output to the paste service to give some output to the user in case the escape filter is triggered


Den4200 (approved):
Looks good to me, thanks!

MarkKoz (approved):

PR #1191, opened 2020-09-27
Verification: broaden 50_007 error code handling to all Discord exception types

### LINKED ISSUE #1179: Error for DMing a user is uncaught in verification
Sentry Issue: [BOT-8Z](https://sentry.io/organizations/python-discord/issues/1901783817/?referrer=github_integration)

As can be seen below, Discord returns a 400 with a specific code rather than a 403 forbidden. Currently, verification attempts to catch this by suppressing a 403 rather than a 400.

```
HTTPException: 400 Bad Request (error code: 50007): Cannot send messages to this user
...
  File "bot/cogs/verification.py", line 503, in on_member_join
    await member.send(ON_JOIN_MESSAGE)
  File "discord/abc.py", line 850, in send
    channel = await self._get_channel()
  File "discord/member.py", line 243, in _get_channel
    ch = await self.create_dm()
  File "discord/user.py", line 715, in create_dm
    data = await state.http.start_private_message(self.id)
  File "discord/http.py", line 245, in request
    raise HTTPException(r, data)

Unhandled exception in on_member_join.
```

kwzrd:
Fixes #1179.

# Context

When sending DMs to Discord users, it is necessary to consider that they may not be accepting them. The canonical way to respond to such situations is to either suppress `discord.Forbidden` (`403`) or to catch & handle it in some way.

However, we've come to learn 2 things:
* A `403` response may potentially indicate something worse than just disabled DMs (@jb3)
* It's possible to receive a `400` status instead of a `403` when DMing a user with closed DMs (#1179)

Discord exceptions carry not only the status, but also a Discord-specific error code. The code `50_007` indicates disabled DMs, and as per #1179 can appear on non-`403` statuses as well.

# Changes

Equipped with this new knowledge, I've made the following adjustments. All DM dispatches in the Verification cog are gated through a new helper `safe_dm`, which will:
* Catch any `discord.HTTPException`
* Suppress the exception if the code was `50_0007`
* Re-raise the exception if the code is different

When sending the "you have been kicked" notification from the verification task, we will now:
* Suppress any Discord exception with the `50_007` error code
* Stop the task and alert `@Admins` on *any* non-`50_007` exception (regardless of status)

**This is different to the previous behaviour**, which would *only* stop the task on non-`50_007` `403`s, and ignore all other exceptions.

For DMs happening outside of the task (on-join DM & rules-accepted DM), we suppress *any* `50_007` Discord exception and log *any* other at error level.

### DIFF
--- a/bot/exts/moderation/verification.py
+++ b/bot/exts/moderation/verification.py
@@ -520,8 +520,10 @@ class Verification(Cog):
             return  # Only listen for PyDis events
 
         log.trace(f"Sending on join message to new member: {member.id}")
-        with suppress(discord.Forbidden):
-            await member.send(ON_JOIN_MESSAGE)
+        try:
+            await safe_dm(member.send(ON_JOIN_MESSAGE))
+        except discord.HTTPException:
+            log.exception("DM dispatch failed on unexpected error code")
 
     @Cog.listener()
     async def on_message(self, message: discord.Message) -> None:
@@ -688,9 +690,9 @@ class Verification(Cog):
             await ctx.author.remove_roles(discord.Object(constants.Roles.unverified))
 
         try:
-            await ctx.author.send(VERIFIED_MESSAGE)
-        except discord.Forbidden:
-            log.info(f"Sending welcome message failed for {ctx.author}.")
+            await safe_dm(ctx.author.send(VERIFIED_MESSAGE))
+        except discord.HTTPException:
+            log.exception(f"Sending welcome message failed for {ctx.author}.")
         finally:
             log.trace(f"Deleting accept message by {ctx.author}.")
             with suppress(discord.NotFound):
--- a/bot/exts/moderation/verification.py
+++ b/bot/exts/moderation/verification.py
@@ -109,6 +109,25 @@ def is_verified(member: discord.Member) -> bool:
     return len(set(member.roles) - unverified_roles) > 0
 
 
+async def safe_dm(coro: t.Coroutine) -> None:
+    """
+    Execute `coro` ignoring disabled DM warnings.
+
+    The 50_0007 error code indicates that the target user does not accept DMs.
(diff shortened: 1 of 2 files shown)


jb3 (approved):

kwzrd:
Squeezed in a tiny fix for a Sentry issue in 27b666b. After discussion with @MarkKoz, we agreed that semantically it makes sense for the log to be at warning level, but we don't want Sentry to notify us about it as there's nothing we can do to prevent it from happening.

bot/exts/moderation/verification.py
```diff
@@ -109,6 +109,25 @@ def is_verified(member: discord.Member) -> bool:
     return len(set(member.roles) - unverified_roles) > 0
 
 
+async def safe_dm(coro: t.Coroutine) -> None:
```
MarkKoz, line 112: This seems like it'll be useful elsewhere. Therefore, I think its home should be in utils/messages or whatever it's named. However, the move could be saved for when someone goes through all other cases to which this function could be applied.

MarkKoz: I feel like this would be better as a context manager, but a function is okay.

MarkKoz (approved):

bot/exts/moderation/verification.py
kwzrd: You're right, it would make sense for it to be a context manager. Although it adds an extra level of indentation if you're still catching other errors coming out of it, but that's a weak excuse.

I'd say let's leave it as a func for now and start thinking of how we can design a more generic suppressor that can eventually replace this and be used elsewhere as well. It may be useful to have the error code configurable, e.g.:

```python
with suppress_error_code(Codes.DISABLED_DM):
  ...
```

I'm not sure whether we'd need the error code configurable though, it may just be the case that `50_007` is the only one we realistically need to handle in a specific way.

PR #1196, opened 2020-09-30
Introduction of discord.py 1.5.x

Senjan21:
As of October 7th, discord will require bots to use intents.

This PR aims to prepare the bot for that by introducing new version of `discord.py` (1.5) and intents set up.

Additionally, `discord.py` has allowed us to disable some events from occurring, thus some events were additionally disabled in this PR.

Due to no need of Presence intent in our bot, `server` and `user` commands were updated to reflect those changes.


thomaspet (approved):
I aproof this, even though i don't know what the integration intent is.

ks129:
@thomaspet Next time when you don't understand thing fully, don't approve it, please.

thomaspet:
> 
> 
> @thomaspet Next time when you don't understand thing fully, don't approve it, please.

We don't use `on_guild_integrations_update`, so I don't need to know what it does. The comment is a slight joke.

MarkKoz (changes requested):
The selection of intents you've disabled looks reasonable.

I've looked for `status`, `raw_status`, and `activities` attribute access and couldn't find anything else, so you seem to have it covered for the most part. However, you need to adjust `on_member_update` in the stats cog since it is keeping track of presences.

Senjan21:
How do you propose I do it @MarkKoz?
Remove it completely? Fetch invite and just update online/offline?

MarkKoz:
Probably has to be gutted. I don't think the event will even be dispatched for status updates if the intent is disabled. Besides, doing an API request (to fetch an invite) so frequently doesn't sound good. That being said, you may be able to fetch the invite and poll periodically. @jb3 may have more to say.

jb3:
The stat breaks anyway since the reporting is often incorrect.

I think it is a nice stat to have, so periodically polling the invite every 10 minutes or something feels pretty okay and is well within the ratelimits.

Senjan21:
Adjusted the cog in 93ce90d @MarkKoz 
Is there anything else that should be adjusted?

tests/bot/exts/info/test_information.py
```diff
@@ -465,11 +393,6 @@ def test_create_user_embed_basic_information_outside_of_moderation_channels(self
             embed.fields[1].value
         )
 
-        self.assertEqual(
-            "basic infractions info",
-            embed.fields[3].value
-        )
```
MarkKoz, line 471: I don't see why this would fail considering you never touched the infraction info code.

Senjan21: It was due to the fact that I have removed 3rd field which had informations about user's status, I have since added the test back and changed the index to `2` in 764f35f as currently, infractions will be in 3rd field and not 4th

kwzrd (changes requested):
With this change, it will now be necessary for developers to enable the members priviledged intent in the dev portal for their bots, otherwise the devbot exits on start-up:

```
discord.errors.PrivilegedIntentsRequired: Shard ID None is requesting privileged intents that have not been explicitly enabled in the developer portal. It is recommended to go to https://discord.com/developers/applications/ and explicitly enable the privileged intents within your application's page. If this is not possible, then consider disabling the privileged intents instead.
```

As this is something that both current and new contributors will need to be aware of in order to run the bot, I propose we add a quick guide on how to do that to the [contributing guide on site](https://pythondiscord.com/pages/contributing/bot/) and perhaps also a quick mention in the readme.

bot/exts/info/information.py
```diff
@@ -153,7 +152,9 @@ async def server_info(self, ctx: Context) -> None:
         channel_counts = self.get_channel_type_counts(ctx.guild)
 
         # How many of each user status?
-        statuses = Counter(member.status for member in ctx.guild.members)
+        py_invite = await self.bot.fetch_invite("python")
+        online_presences = py_invite.approximate_presence_count
+        offline_presences = ctx.guild.member_count - online_presences
```
kwzrd, line 157: Since the `!server` cmd always fetches the main guild invite, the presence counts will be very off elsewhere.

image

Do you think there's something we could do about this, or is it not worth worrying about? It would be nice to also use the prod guild for the `member_count`, but then devbots cannot fetch it because they're not in it.

kwzrd: Maybe it could just say, e.g.:
> :green_circle: Approximately `23,687` members online

jb3: We could iterate through guild invites to generate that section.

jb3: Actually, to expand on that we could iterate on *startup* to find a permanent invite..

Senjan21: I don't see sending yet another request to the API just for some stats a nice thing to do.

We can instead make the invite link a config variable and use both approx presence count and approx member count rather than currently having hardcoded invite link, approx presence count and actual server size.

However as we grow bigger, approx member count will be more out of sync with reality I'd assume, however if we need such modularity it sounds like a good trade.

kwzrd: > use both approx presence count and approx member count

Yeah, I wasn't aware this was possible, I assumed the current solution uses the `ctx` guild because it cannot get the info from the invite. I'm in favour of doing this then, as long as the invite gives us all we need.

MarkKoz (approved):

bot/__main__.py
```diff
@@ -47,14 +47,22 @@
...
 bot = Bot(
     redis_session=redis_session,
     loop=loop,
     command_prefix=when_mentioned_or(constants.Bot.prefix),
     activity=discord.Game(name="Commands: !help"),
     case_insensitive=True,
     max_messages=10_000,
-    allowed_mentions=discord.AllowedMentions(everyone=False, roles=allowed_roles)
+    allowed_mentions=discord.AllowedMentions(everyone=False, roles=allowed_roles),
+    intents=intents
```
MrAwesomeRocks, line 65:
```suggestion
    intents=intents,
```
Can you add a comma here? It doesn't change the execution of the code, but it's recommended by PEP 8, and leads to nicer diffs in GitHub.

MrAwesomeRocks, line 65:
```suggestion
    intents=intents,
```
Can you add a comma here? It doesn't change the execution of the code, but it's recommended by PEP 8, and leads to nicer diffs in GitHub.

MrAwesomeRocks (changes requested):

bot/__main__.py
MrAwesomeRocks, line 65:
```suggestion
    intents=intents,
```
Can you add a comma here? It doesn't change the execution of the code, but it's recommended by PEP 8, and leads to nicer diffs in GitHub.

MrAwesomeRocks: Sorry for posting that three times, GitHub has been acting up for me.

Senjan21: That's a good idea I suppose, will do so.

bot/exts/info/information.py
Senjan21: With 397d29a it should now display stats for the configurated invite link.

MrAwesomeRocks (approved):
This PR looks good to me!

kwzrd (approved):
This all looks ok to me. I've tested the bot's core functionality and everything seems to work on `1.5.0`. The intents seem to be setup correctly and I cannot think of anything that we missed ~ if we did, then it shouldn't be anything critical.

Regarding the documentation, I have added a short section to the [bot contrib guidelines](https://pythondiscord.com/pages/contributing/bot/#privileged-intents). It may make more sense to have it on the [bot account setup page](https://pythondiscord.com/pages/contributing/setting-test-server-and-bot-account/#setting-up-a-bot-account), but that is also linked from seasonalbot's setup guide and we haven't yet fully agreed as to whether we'd require the member intent for seasonalbot too; it will require some more core dev discussion.

Thanks for taking care of it!

PR #1197, opened 2020-09-30
Infraction subcommand for amending infraction reasons

### LINKED ISSUE #1170: Infraction subcommand for amending infraction reasons
## Motivation
Currently, the only way to amend an infraction reason is to search and copy the original reason before invoking
```
!infr edit (last|infr_id) {old_reason} (Edit|Context): {amended_reason}
```
An `infr` subcommand that abstracts this action would allow for a nicer interface, at least in my opinion, especially for amending infractions that were needed to be handed out quickly.

## Proposed Implementation
A subcommand that's similar in form to `infraction_edit` to be called `infraction_amend` (à la `git commit --amend`) first looks up the original reason for the infraction (possibly using [this routine](<https://github.com/python-discord/bot/blob/1ae971423ae5fcf3f8a68bcacbf1591c2ccf0aac/bot/exts/moderation/infraction/management.py#L81-L97>)), amending the new reason prefixed with either `Edit: ` or `Context: `, and then invoking  `infraction_edit`. The infraction duration can also be edited by passing the appropriate parameters and the new duration could also possibly be amended automatically to the message.

### Example Usage:
```
> !infr amend last This member had `...`
] Spamming in #python-general. Edit: This member had `...`

> !infr amend 512 1h This member had ...
] Spamming in #python-general. Edit: This member had `...`. Infraction incremented/reduced to `...` from `...`.
```

PureFunctor:
Closes #1170.

Rudimentary implementation example

Screenshot_2020-10-01_04-13-15

Since `amend` invokes `edit` I've also written a helper function (`get_latest_infraction`) to prevent duplication of the routine needed to obtain the latest infraction by a specific actor although I'm still not sure what to make of
```python
        if isinstance(infraction_id, str):
            old_infraction = await self.get_latest_infraction(ctx.author.id)

            if old_infraction is None:
                await ctx.send(
                    ":x: Couldn't find most recent infraction; you have never given an infraction."
                )
                return

            infraction_id = old_infraction["id"]

        else:
            old_infraction = await self.bot.api_client.get(f"bot/infractions/{infraction_id}")
```


bot/exts/moderation/infraction/management.py
```diff
@@ -45,6 +45,52 @@ async def infraction_group(self, ctx: Context) -> None:
...
+            if old_infraction is None:
+                await ctx.send(
+                    ":x: Couldn't find most recent infraction; you have never given an infraction."
+                )
+                return
+
+            infraction_id = old_infraction["id"]
+
+        else:
+            old_infraction = await self.bot.api_client.get(f"bot/infractions/{infraction_id}")
```
MarkKoz, line 88: Since this is all repeated in the other command, it should all be in a separate function. At this point, I think you should make it a converter. It'll allow you to get rid of `allowed_strings("l", "last", "recent")` since you'll be able to internalise that into the converter.

bot/exts/moderation/infraction/management.py
```diff
@@ -45,6 +45,52 @@ async def infraction_group(self, ctx: Context) -> None:
...
+
+        reason = fr"{old_infraction['reason']} **\|\|** {reason}"
+
+        await ctx.invoke(self.infraction_edit, infraction_id=infraction_id, duration=duration, reason=reason)
```
MarkKoz, line 92: Since d.py 1.3, commands can be invoked by calling the function e.g. `await self.infraction_edit(ctx, ...)`

MarkKoz (changes requested):

bot/exts/moderation/infraction/management.py
PureFunctor: Should the converter return the infraction id or the infraction dictionary itself? I'm thinking of something like
```python
class InfractionLookup(Converter):
    async def convert(self, ctx: Context, arg: str) -> t.Optional[dict]:
```
although I'm not sure whether the `infraction_id` parameter reflects its purpose concisely.

MarkKoz: Return the full infraction data as a dictionary. Otherwise, you'd need to make a redundant request to get the data within the command. This is cause to get the ID, the converter has to make an API request in the case that the user wants to use the most recent infraction.

bot/exts/moderation/infraction/management.py
```diff
@@ -49,7 +49,7 @@ async def infraction_group(self, ctx: Context) -> None:
     async def infraction_append(
         self,
         ctx: Context,
-        infraction_id: t.Union[int, allowed_strings("l", "last", "recent")],  # noqa: F821
+        infraction: Infraction,  # noqa: F821
```
MarkKoz, line 52: This noqa isn't needed. It was here because flake8 doesn't handle function calls with strings (allowed_strings) very well.

bot/exts/moderation/infraction/management.py
```diff
@@ -73,29 +73,21 @@ async def infraction_append(
...
+            duration=duration,
+            reason=fr"{infraction['reason']} **\|\|** {reason}",
+        )
 
     @infraction_group.command(name='edit')
     async def infraction_edit(
         self,
         ctx: Context,
-        infraction_id: t.Union[int, allowed_strings("l", "last", "recent")],  # noqa: F821
+        infraction: Infraction,  # noqa: F821
```
MarkKoz, line 90: Noqa isn't needed.

bot/converters.py
```diff
@@ -549,6 +549,36 @@ def _snowflake_from_regex(pattern: t.Pattern, arg: str) -> int:
...
+                "ordering": "-inserted_at"
+            }
+
+            infractions = await ctx.bot.api_client.get("bot/infractions", params=params)
+
+            if not infractions:
+                await ctx.send(
+                    ":x: Couldn't find most recent infraction; you have never given an infraction."
+                )
+                return None
```
MarkKoz, line 574: You should raise an exception (probably `BadArgument`). This makes more sense for a converter. It also means you won't need to do `if infraction is None: return` in the command later.

bot/exts/moderation/infraction/management.py
```diff
@@ -73,29 +73,21 @@ async def infraction_append(
...
+        if not infraction:
+            return
 
-        await self.infraction_edit(infraction_id=infraction_id, duration=duration, reason=reason)
+        await self.infraction_edit(
+            ctx=ctx,
+            infraction=infraction,
+            duration=duration,
+            reason=fr"{infraction['reason']} **\|\|** {reason}",
+        )
```
MarkKoz, line 84: You could have kept this as 2 lines rather than 6 (only the `reason` needs to be a kwarg). But it's fine.

bot/exts/moderation/infraction/management.py
```diff
@@ -123,20 +115,11 @@ async def infraction_edit(
...
-                )
-                return
-
-            infraction_id = old_infraction["id"]
+        if not infraction:
+            return
 
-        else:
-            old_infraction = await self.bot.api_client.get(f"bot/infractions/{infraction_id}")
+        old_infraction = infraction
```
MarkKoz, line 121: This is redundant. Just change the following code to use `infraction` instead.

MarkKoz (changes requested):

bot/converters.py
```diff
@@ -549,6 +549,35 @@ def _snowflake_from_regex(pattern: t.Pattern, arg: str) -> int:
...
+
+            if not infractions:
+                raise BadArgument(
+                    "Couldn't find most recent infraction; you have never given an infraction."
+                )
+            else:
+                return infractions[0]
+
+        else:
+            return ctx.bot.api_client.get(f"bot/infractions/{arg}")
```
mbaruh, line 578: This needs to be awaited, otherwise you'll be trying to parse a coroutine object.

bot/exts/moderation/infraction/management.py
```diff
@@ -45,11 +46,50 @@ async def infraction_group(self, ctx: Context) -> None:
...
+        Use "l", "last", or "recent" as the infraction ID to specify that the most recent infraction
+        authored by the command invoker should be edited.
+
+        Use "p" or "permanent" to mark the infraction as permanent. Alternatively, an ISO 8601
+        timestamp can be provided for the duration.
+
+        If a previous infraction reason does not end with an ending punctuation mark, this automatically
+        adds a period before the amended reason.
+        """
+        add_period = not END_PUNCTUATION_RE.match(infraction["reason"])
```
mbaruh, line 80: In the case where no reason is provided (for example bans), `infraction["reason"]` is None, and this line will freak out.

bot/utils/regex.py
```diff
@@ -10,3 +10,5 @@
     r"([a-zA-Z0-9\-]+)",                              # the invite code itself
     flags=re.IGNORECASE
 )
+
+END_PUNCTUATION_RE = re.compile("^.+?[.?!]$")
```
mbaruh, line 14: I'm not requiring it, but if it were me I would just have this constant in the file where it's used, since it only serves one function currently. It might be generic enough to be useful in the future for something else, but not right now.

mbaruh (changes requested):

bot/utils/regex.py
MarkKoz, line 14: I suggest avoiding a regex for this. Use `str.endswith((".", "?", "!"))`; it's simpler and therefore easier to understand.

bot/exts/moderation/infraction/management.py
```diff
@@ -72,13 +73,17 @@ async def infraction_append(
...
-            ctx=ctx,
-            infraction=infraction,
-            duration=duration,
-            reason=fr"{infraction['reason']} **\|\|** {reason}",
-        )
+        add_period = not END_PUNCTUATION_RE.match(infraction["reason"])
+
+        new_reason = "".join((
+            infraction["reason"], ". " if add_period else " ", reason,
+        ))
```
MarkKoz, line 84: Strange to see `join` used here. I find it confusing to follow at first glance, since I associate a `join` with doing something a bit more complicated typically. I suggest you simply use the `+` operator.

MarkKoz (changes requested):

bot/exts/moderation/infraction/management.py
```diff
@@ -288,6 +310,20 @@ def infraction_to_string(self, infraction: t.Dict[str, t.Any]) -> str:
...
+            "ordering": "-inserted_at"
+        }
+
+        infractions = await self.bot.api_client.get("bot/infractions", params=params)
+
+        if infractions:
+            return infractions[0]
+
+        return None
+
```
mbaruh, line 326: This doesn't seem to be used anywhere anymore

mbaruh (changes requested):

mbaruh (approved):

MarkKoz (approved):

Issue #1195, opened 2020-09-29
Source command bug


Anubhav1603:
source command is erroring out when i searched for `get` command and thumbnail is also not working
![b](https://user-images.githubusercontent.com/57266248/94529390-4bad7000-0257-11eb-8037-ced39cc32a48.PNG)

Den4200:
This happens because of our configured aliases.

![image](https://user-images.githubusercontent.com/52584399/94641242-b76ef780-02ae-11eb-8c41-43a1d955b9c6.png)

`!exception` and `!get` will fail because they are not actual commands, groups, or cogs - just aliases. On the other hand, `!docs` and `!tags` are both command groups while also being aliases, so those do not have this issue.

Anubhav1603:
Ok tnx for info :)

Den4200:
Well, this bug still needs fixing. I was just explaining why the error occurs.

Anubhav1603:
Ok

PR #1198, opened 2020-09-30
Remove alias cog

### LINKED ISSUE #1159: Remove alias cog
Since the merging of #1124, only 3 alias commands remain, listed below. All usage statistics are since stats have been tracked, which is around April of 2020.

1. `!get docs` - alias for `!docs get`
    * Usage: 4 times
    * Usage of `!docs get`: 793 times
    * Usage of `!docs`: 9240 times (includes all the above and empty arg invocations)
2. `!get tags` - alias for `!tags get`
    * Usage: 0 times
    * Usage of `!tags get`: 4 times
    * Usage of `!tags`: 720 times (includes all the above and empty arg invocations)
3. `!extension` - alias for `!tags get traceback`
    * Usage: 12 times
    * Usage of the `traceback` tag: 194 times

The usages compared to other ways to invoke the commands are ~0.04%, 0%, and 6.19% respectively. Quite low numbers. Regarding the `traceback` alias, the tag search functionality will likely pick up the `exception` keyword and show the traceback tag anyway.

Removing this cog would reduce maintenance costs (granted, not by much). It would also just clean things up a bit. I find the existence of this code to be an anomaly, for lack of a better word, considering #1124. Lastly, it means #844 and #847 would be closed.

Is anyone on board with this proposal?

MarkKoz:
Last few aliases are an anomaly since #1124 was merged. The remaining aliases are seldom used. The code isn't exactly clean and it has some maintenance costs.

Resolves #1159


Den4200:
This also resolves #1195.

Den4200 (approved):

kwzrd (commented):
The BotSource cog has some special handling for Alias that won't be needed anymore.

https://github.com/python-discord/bot/blob/e80bfe4b4f6901d8efb31c24e4af23b4b4317193/bot/exts/info/source.py#L69-L73

https://github.com/python-discord/bot/blob/e80bfe4b4f6901d8efb31c24e4af23b4b4317193/bot/exts/info/source.py#L116-L119

Should we remove it too?

kwzrd (approved):

PR #1201, opened 2020-10-01
Duck pond: ignore reactions in DMs

### LINKED ISSUE #1183: Duck pond may fail to fetch reaction message due to a missing channel
Sentry Issue: [BOT-8T](https://sentry.io/organizations/python-discord/issues/1892623782/?referrer=github_integration)

Possibly a cache issue. I believe that a channel object isn't really needed. There's likely a direct way to get the message via a method in `bot.http`. After all, the actual endpoint only needs a channel ID and a message ID.

```
AttributeError: 'NoneType' object has no attribute 'fetch_message'
    await coro(*args, **kwargs)
  File "bot/cogs/duck_pond.py", line 158, in on_raw_reaction_remove

Unhandled exception in on_raw_reaction_remove.
```

MarkKoz:
Also handle the channel not being found, which may be due to a cache issue or because it got deleted.

Fixes #1183
Fixes BOT-8T

### DIFF
--- a/bot/exts/fun/duck_pond.py
+++ b/bot/exts/fun/duck_pond.py
@@ -145,8 +145,8 @@ class DuckPond(Cog):
         amount of ducks specified in the config under duck_pond/threshold, it will
         send the message off to the duck pond.
         """
-        # Ignore DMs.
-        if payload.guild_id is None:
+        # Ignore other guilds and DMs.
+        if payload.guild_id != constants.Guild.id:
             return
 
         # Was this reaction issued in a blacklisted channel?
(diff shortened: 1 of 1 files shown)


Den4200 (approved):

kwzrd (approved):

bot/exts/fun/duck_pond.py
```diff
@@ -145,6 +145,10 @@ async def on_raw_reaction_add(self, payload: RawReactionActionEvent) -> None:
         amount of ducks specified in the config under duck_pond/threshold, it will
         send the message off to the duck pond.
         """
+        # Ignore DMs.
+        if payload.guild_id is None:
```
kwzrd, line 149: I wanted to say that it may make more sense to check whether the ID matches the configured PyDis ID, but in practice it probably makes no difference.

Issue #905, opened 2020-04-24
Identify and fix potential forbidden errors due to role hierarchy


MarkKoz:
Sometimes `discord.Forbidden` errors are raised due to the bot being lower in the role hierarchy than the user they are trying to edit. An example of this is joe trying to superstarify himself (see [BOT-3C](https://sentry.io/organizations/python-discord/issues/1621105430/?referrer=github_integration)). 

Depending on which groups of users this affects, we should try to gracefully handle such error. If this only affects the server owner (as in, true ownership, not just the role we have), the we could ignore it. If it affects roles like admins and mods, then I think we should address this error.

Unfortunately, I don't think Discord's API distinguishes between someone being too low in the hierarchy and the bot not having permissions granted. If we want to log the latter as much as possible, then we couldn't do a blanket ignore (or lower log level) on all `discord.Forbidden` exceptions. However, if we think it's safe to assume the bot will always have the right permissions, then maybe such an ignore is a valid solution. Otherwise, potential errors specifically do to hierarchy will need to be handled individually. For those errors, we'd have to assume that it's due to role hierarchy and not the bot missing permissions.

mbaruh:
The PR is merged and now it just fails to apply the infraction, which is something we handle in `apply_infraction`.

PR #1204, opened 2020-10-01
Modify `!superstar` to use `apply_infraction`.

kosayoda:
Using `apply_infraction` from `InfractionScheduler` rather than doing it manually allows us to handle HTTP errors while reducing code duplication.

Specifically, `discord.Forbidden` is handled when the bot tries to superstar someone they do not have permissions to.

This *should* address #905. There's only one instance of it unhandled on Sentry (which this PR fixes), and other potential places with `discord.Forbidden` is also handled in their respective cogs. Therefore, there is no need to implement a blanket ignore just so we can catch errors like this as they occur clear and quick.

If the bot reschedules the superstar, the issue will occur again. This should be handled manually, ie. by `!unstar`ing that specific member.

Resolves BOT-5Q.

PS: To reproduce `BOT-5Q`, superstar an admin or higher.


bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -81,15 +81,29 @@ async def apply_infraction(
...
         icon = _utils.INFRACTION_ICONS[infr_type][0]
         reason = infraction["reason"]
         expiry = time.format_infraction_with_duration(infraction["expires_at"])
         id_ = infraction['id']
 
+        if reason_override is not None:
+            reason_override = reason
+
+        if additional_info is not None:
+            additional_info = ""
```
kwzrd, line 105: Logically this does not look right to me. How will any `reason_override` or `additional_info` show anywhere?

Additionally, an empty string seems like a better default for `additional_info` than `None` due to the way it's used.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -125,7 +139,7 @@ async def apply_infraction(
                 log.error(f"Failed to DM {user.id}: could not fetch user (status {e.status})")
             else:
                 # Accordingly display whether the user was successfully notified via DM.
-                if await _utils.notify_infraction(user, infr_type, expiry, reason, icon):
+                if await _utils.notify_infraction(user, infr_type, expiry, reason_override, icon):
```
kwzrd, line 142: Should this be passed `reason` still and instead have `reason_override` assigned to `reason` if provided?

kwzrd (changes requested):

bot/exts/moderation/infraction/_scheduler.py
MrAwesomeRocks: Shouldn't this be checking if `reason_override` or `additional_info` are `None`? Otherwise it will send `None` to the user.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -81,15 +81,29 @@ async def apply_infraction(
...
+        """
+        Apply an infraction to the user, log the infraction, and optionally notify the user.
+
+        `reason_override`, if provided, will be sent to the user in place of the infraction reason.
+        `additional_info` will be attached to the text field in the mod-log embed.
+        Returns whether or not the infraction succeeded.
+        """
         infr_type = infraction["type"]
         icon = _utils.INFRACTION_ICONS[infr_type][0]
         reason = infraction["reason"]
```
MrAwesomeRocks, line 97:
```suggestion
        reason = infraction["reason"] if reason_override is not None else reason_override
```
I think something along the lines of this would be a better way of setting the reason for the infraction.

bot/exts/moderation/infraction/_scheduler.py
kosayoda: I bad :(

Resolved in 54a0ad2

bot/exts/moderation/infraction/_scheduler.py
kosayoda: Perhaps the name `reason_override` is a bit misleading, it's for having a different message be sent to the user as the `reason` for the infraction rather than the reason sent to `#mod-log`. I've renamed the parameter to hopefully make it clearer.

bot/exts/moderation/infraction/_scheduler.py
kosayoda: Perhaps the name `reason_override` is a bit misleading, it's for having a different message be sent to the user as the `reason` for the infraction rather than the reason sent to `#mod-log`. I've renamed the parameter to hopefully make it clearer.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -81,15 +81,27 @@ async def apply_infraction(
         ctx: Context,
         infraction: _utils.Infraction,
         user: UserSnowflake,
-        action_coro: t.Optional[t.Awaitable] = None
-    ) -> None:
-        """Apply an infraction to the user, log the infraction, and optionally notify the user."""
+        action_coro: t.Optional[t.Awaitable] = None,
+        user_reason: t.Optional[str] = None,
+        additional_info: t.Optional[str] = "",
```
kwzrd, line 86: I think it no longer makes sense for `additional_info` to be optional. The default is a string and we wouldn't want to get a `None` as its repr would show in the modlog.

bot/exts/moderation/infraction/_scheduler.py
kwzrd: I agree that the new name is better.

bot/exts/moderation/infraction/_scheduler.py
kwzrd: Thanks, it's better now.

bot/exts/moderation/infraction/_scheduler.py
kosayoda: Addressed in acb6d73

Senjan21 (approved):
Code itself looks good, I haven't tested it yet although I doubt it will break anyhow now.

mbaruh (approved):

PR #1207, opened 2020-10-02
Fix typos in bot/ docstrings and comments

hedythedev:
for `bot/exts/info/information.py`, flake8 said that the line length was too long so I had to re-wrap the lines because of the spelling change formated => formatted.

I hope having the commits on master was fine because they are really just minor changes


MarkKoz (approved):
Thanks. It's fine if the source branch is master; doesn't matter on our end.

Den4200 (approved):
Thanks!

PR #1208, opened 2020-10-02
Reminder: no feedback message when no mention is used

Akarys42:
Remove the `and will mention x other(s)` message when no mention is attached to the reminder.


MarkKoz (approved):
I remember making this suggestion in the original PR but approving anyway. Nice to see it come around eventually.

kwzrd (approved):
I'd prefer Scrags' as per `#dev-core`, but this is fine.

PR #1209, opened 2020-10-02
Send attachments improvements

### LINKED ISSUE #909: 415 error for attachment re-upload
Sentry issue: [BOT-6V](https://sentry.io/organizations/python-discord/issues/1780462718/?referrer=github_integration)

Strange issue that has happened 6 times in which a attachment fails to be uploaded due to a 415 (unsupported media type) error.

My suspicion is that it's not due to a misconfigured `Content-Type` header as one would expect but rather Discord's API using that error to reject certain files. I don't know how `Content-Type` could possibly be wrong.

This has happened for both `.py` and `.txt` files so far.

### LINKED ISSUE #1137: ID not displayed in dm_relay when attachment sent
Currently the dm-relay message username will not include the user ID if only an attachment is sent, so it looks like this:
image
Whereas it should look like this:
image

This is because it uses [`bot.utils.messages.send_attachments`](https://github.com/python-discord/bot/blob/1bd5c80e7b2aa90162639b86960c66f6a218022f/bot/utils/messages.py#L53), which by default just sends the message with the ID. Relevant line in `dm_relay` cog [here](https://github.com/python-discord/bot/blob/1bd5c80e7b2aa90162639b86960c66f6a218022f/bot/cogs/dm_relay.py#L93)

Simple solution here would be to add a `username` keyword argument to `send_attachments`. I would be happy to make this change if this all sounds good.

### LINKED ISSUE #1139: Some attachments being ignored in dm-log
### Bug:
If you DM an attachment of a certain type (seems to only happen for smaller files or non image/video files, not exactly sure) to the bot, instead of relaying it to the dm-log channel as expected, the message is seemingly ignored.
### Cause:
I did some detective work and found out this error was being raised by [this](https://github.com/python-discord/bot/blob/a61d0321c4b7a6e137ccb59d8a3af0428838778c/bot/utils/messages.py#L53) function when the attachment was DMed to the bot:
```
bot_1       | 2020-08-31 17:46:29 | bot.utils.messages | WARNING | Failed to re-upload attachment test.txt from message 750048883488325682 with status 415.
bot_1       | Traceback (most recent call last):
bot_1       |   File "/bot/bot/utils/messages.py", line 79, in send_attachments
bot_1       |     await attachment.save(file, use_cached=True)
bot_1       |   File "/usr/local/lib/python3.8/site-packages/discord/message.py", line 124, in save
bot_1       |     data = await self.read(use_cached=use_cached)
bot_1       |   File "/usr/local/lib/python3.8/site-packages/discord/message.py", line 166, in read
bot_1       |     data = await self._http.get_from_cdn(url)
bot_1       |   File "/usr/local/lib/python3.8/site-packages/discord/http.py", line 239, in get_from_cdn
bot_1       |     raise HTTPException(resp, 'failed to get asset')
bot_1       | discord.errors.HTTPException: 415 Unsupported Media Type (error code: 0): failed to get asset
```
Some further investigations have lead me to believe that the problem is caused by using `use_cached=True` on [this](https://github.com/python-discord/bot/blob/a61d0321c4b7a6e137ccb59d8a3af0428838778c/bot/utils/messages.py#L78) line, and indeed, looking at the [docs for attachment.save](https://discordpy.readthedocs.io/en/latest/api.html?highlight=attachment#discord.Attachment.save) it says `Note that this can still fail to download deleted attachments if too much time has passed and it does not work on some types of attachments.`, which seems to fit.

### Solution:
The simplest option would just be to removed `use_cached` (as it defaults to `False`). I'm not sure on the reasoning for setting it to `True` though, in the docs it says `This will allow attachments to be saved after deletion more often, compared to the regular URL which is generally deleted right after the message is deleted`, although I'm not sure how necessary this is.

I would be happy to make whatever change is necessary for this, if somebody can confirm what I am observing is correct and the solution is correct. Let me know here or on discord if you are struggling to reproduce this, thanks.

wookie184:
Closes #1137 and closes #1139 
I think this also closes #909 

I just changed the default for `use_cached` to False, and didn't change it to set to True manually anywhere as I couldn't find anywhere that it was needed, antispam was the main one I checked but doesn't seem to try to fetch deleted messages from what I can see. Would be nice if somebody could confirm this

### DIFF
--- a/bot/exts/moderation/dm_relay.py
+++ b/bot/exts/moderation/dm_relay.py
@@ -90,7 +90,11 @@ class DMRelay(Cog):
         # Handle any attachments
         if message.attachments:
             try:
-                await send_attachments(message, self.webhook)
+                await send_attachments(
+                    message,
+                    self.webhook,
+                    username=f"{message.author.display_name} ({message.author.id})"
+                )
             except (discord.errors.Forbidden, discord.errors.NotFound):
                 e = discord.Embed(
                     description=":x: **This message contained an attachment, but it could not be retrieved**",
--- a/bot/utils/messages.py
+++ b/bot/utils/messages.py
@@ -56,15 +56,22 @@ async def wait_for_deletion(
 async def send_attachments(
     message: discord.Message,
     destination: Union[discord.TextChannel, discord.Webhook],
-    link_large: bool = True
+    link_large: bool = True,
+    **kwargs
 ) -> List[str]:
     """
     Re-upload the message's attachments to the destination and return a list of their new URLs.
 
     Each attachment is sent as a separate message to more easily comply with the request/file size
     limit. If link_large is True, attachments which are too large are instead grouped into a single
-    embed which links to them.
+    embed which links to them. Extra kwargs will be passed to send() when sending the attachment.
     """
+    webhook_send_kwargs = {
+        'username': sub_clyde(message.author.display_name),
+        'avatar_url': message.author.avatar_url,
+    }
+    webhook_send_kwargs.update(kwargs)
+
     large = []
     urls = []
     for attachment in message.attachments:
@@ -82,14 +89,10 @@ async def send_attachments(
                     attachment_file = discord.File(file, filename=attachment.filename)
 
                     if isinstance(destination, discord.TextChannel):
-                        msg = await destination.send(file=attachment_file)
+                        msg = await destination.send(file=attachment_file, **kwargs)
                         urls.append(msg.attachments[0].url)
                     else:
-                        await destination.send(
-                            file=attachment_file,
-                            username=sub_clyde(message.author.display_name),
-                            avatar_url=message.author.avatar_url
-                        )
+                        await destination.send(file=attachment_file, **webhook_send_kwargs)
             elif link_large:
                 large.append(attachment)
             else:
@@ -106,13 +109,9 @@ async def send_attachments(
         embed.set_footer(text="Attachments exceed upload size limit.")
 
         if isinstance(destination, discord.TextChannel):
-            await destination.send(embed=embed)
+            await destination.send(embed=embed, **kwargs)
         else:
-            await destination.send(
-                embed=embed,
-                username=sub_clyde(message.author.display_name),
-                avatar_url=message.author.avatar_url
-            )
+            await destination.send(embed=embed, **webhook_send_kwargs)
 
     return urls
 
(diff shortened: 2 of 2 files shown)


bot/exts/moderation/dm_relay.py
```diff
@@ -90,7 +90,11 @@ async def on_message(self, message: discord.Message) -> None:
         # Handle any attachments
         if message.attachments:
             try:
-                await send_attachments(message, self.webhook)
+                await send_attachments(
+                    message,
+                    self.webhook,
+                    username=f"{message.author.display_name} ({message.author.id})"
```
MarkKoz, line 96: You must use `sub_clyde` for the username.

MarkKoz (changes requested):
You're correct that antispam sends the attachments before deleting the messages. However, setting `use_cached` to false decreases the likelihood of successfully sending deleted attachments for all callers of the function. This is because a message can be deleted by the author while the bot is trying to process/handle the message. That being said, I don't know if it's really worth potentially making two requests per attachment via the idea to try the proxy and then the normal if the proxy fails.

bot/exts/moderation/dm_relay.py
wookie184: > You're correct that antispam sends the attachments before deleting the messages. However, setting `use_cached` to false decreases the likelihood of successfully sending deleted attachments for all callers of the function. This is because a message can be deleted by the author while the bot is trying to process/handle the message. That being said, I don't know if it's really worth potentially making two requests per attachment via the idea to try the proxy and then the normal if the proxy fails.

I see, I don't think it would be really be worth trying with `use_cached` and then without, as without should work the vast majority of the time. Trying without `use_cached` and with it if that fails would probably be best if we thought deleted messages could be a problem, but I'm not sure it's that important.

MarkKoz: It depends on the kind of accuracy the mod team wants, but yes, I tend to agree that it isn't worth using. Anyway, you still need to fix the sub_clyde issue. Now that I think about it, it'd be more convenient to do it within `send_attachments` i.e. 

```py
    webhook_send_kwargs = {
        'username': message.author.display_name,
        'avatar_url': message.author.avatar_url,
    }
    webhook_send_kwargs.update(kwargs)
    webhook_send_kwargs['username'] = sub_clyde(webhook_send_kwargs['username'])
```

MarkKoz (approved):

kwzrd (approved):
Looks alright. I think it's a little risky to forward `**kwargs` because of the `destination` union type, but I guess it would be the caller's responsibility to make sure they don't pass an invalid kwarg.

PR #1210, opened 2020-10-03
Add a socket stats command

jb3:
This adds a command which can be used by Core Developers to observe the types of events received by the bot.

It will present the number of events received per second and the number of events for each `t` value in Discord dispatch payloads.

Screenshot 2020-10-03 at 11 37 16


Senjan21 (approved):
Looks rock solid.

Akarys42 (approved):
Perfect 👌

PR #1211, opened 2020-10-04
Improvements for the rules command

Den4200:
## Description
This PR has a few improvements for the `!rules` command. Instead of using the splat operator (`*`), we are now using the `Greedy` converter to allow members to add extra text after the rule indices. I've also removed duplicate rule indices by using sets, and then sorting them afterwards.

## Screenshots
Text after rule indices:
image

Before duplication removal (note the 357 pages):
image

After duplication removal (note the 3 pages):
image

### DIFF
--- a/bot/exts/info/site.py
+++ b/bot/exts/info/site.py
@@ -122,10 +122,14 @@ class Site(Cog):
             return
 
         full_rules = await self.bot.api_client.get('rules', params={'link_format': 'md'})
-        invalid_indices = tuple(
-            pick
-            for pick in rules
-            if pick < 1 or pick > len(full_rules)
+
+        # Remove duplicates and sort the invalid rule indices
+        invalid_indices = sorted(
+            set(
+                pick
+                for pick in rules
+                if pick < 1 or pick > len(full_rules)
+            )
         )
 
         if invalid_indices:
@@ -135,6 +139,9 @@ class Site(Cog):
 
         for rule in rules:
             self.bot.stats.incr(f"rule_uses.{rule}")
+
+        # Remove duplicates and sort the rule indices
+        rules = sorted(set(rules))
 
         final_rules = tuple(f"**{pick}.** {full_rules[pick - 1]}" for pick in rules)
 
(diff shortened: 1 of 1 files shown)


bot/exts/info/site.py
```diff
@@ -122,20 +121,27 @@ async def site_rules(self, ctx: Context, *rules: int) -> None:
...
-            indices = ', '.join(map(str, invalid_indices))
+            indices = ', '.join(str(index) for index in invalid_indices)
             await ctx.send(f":x: Invalid rule indices: {indices}")
             return
 
         for rule in rules:
             self.bot.stats.incr(f"rule_uses.{rule}")
 
+        # Remove duplicates and sort the rule indices
+        rules = sorted(set(rules))
```
MarkKoz, line 143: Do this before getting invalid indices to avoid having to iterate through duplicates during the search and avoid sorting again. Example

```py
rules = sorted(set(rules))
invalid = ', '.join(str(i) for i in rules if i < 1 or i > len(full_rules))

if invalid:
    ...
```

MarkKoz (changes requested):
Convert

bot/exts/info/site.py
Den4200: Good catch, addressed in 2553a1d

kwzrd (approved):
Look alright to me. I think the variable naming in the method could be overall improved, it's a little bit all over the place (perhaps better to use either `index` or `pick` in both genexps, `full_rules` could be `fetched_rules` and `rules` could be `picks` or `picked_rules`) but that's minor.

I'm also triggered by the way this string is formatted. :rage:
https://github.com/python-discord/bot/blob/ff7c34aee21493efbc78a9b6521dd9c0f48d4a72/bot/exts/info/site.py#L116-L118

kwzrd (commented):
I also noticed that the annotation for `LinePaginator.paginate`'s first arg is violated - it wants a list but we give it a tuple. That's not caused by this PR's diff though, and the annotation is probably needlessly strict.

MarkKoz (approved):

PR #1212, opened 2020-10-04
Added interleaving text in code blocks option

### LINKED ISSUE #1104: Allow text to be inserted before and after eval codeblocks
## Rationale

During an eval demonstration, it'd be nice to be able to insert text before and after the codeblock to comment it, instead of putting actual comments in the code, or sending two messages.

## Specifications

Here is the current regex used to extract the code of a message is this one : 

https://github.com/python-discord/bot/blob/b27c286f2ce648d021eed0c8e07476066b86dd98/bot/cogs/snekbox.py#L22-L32

We could simply insert a `.*` before and after to make it work, as in [this example](https://regex101.com/r/Xac4It/1)

mbaruh:
Closes #1104 

If the message contains both plaintext and code blocks, the text will be ignored.
If several code blocks are present, they are concatenated.

#### Implementation rationale
1. The issue with inline code is that it's used often inside an explanation to mark certain keywords.
For example:
> You can use the `join` function to concatenate a collection of strings:
> ```py
> strings = ('abc', 'xyz', 'asdf')
> print(' '.join(strings))

If we allow text to be added around the code, it wouldn't make sense to take those inline code segments into account.
We could decide to take into account only those that appear at the beginning of a line, but it would still easily trip users. I could just as easily write: 
>  `join` is a function to concatenate a collection of strings: ...

2. Therefore the idea is to make sure that if someone wants to add text, they should state their intention clearly for what part of the message they want to be the code, therefore allowing only code blocks in that case. Hence the separate regex.

Since it has separate regex, the pattern has to be matched separately once we know that not the whole message is formatted as code.

3. As per what was discussed on the issue, I added the option for separate code blocks.
Since we allow several code blocks, we need to make sure that the first regex doesn't result in a full match.
The wildcard for the code is lazy, but the pattern requires that it is followed from the beginning of the string to its end (^, $). therefore it will match the whole string:
"""\```py
print('hi')
\```
hello
\```
print('bye')
\```"""

Instead of just the first block, and will prevent the part of the code that handles multiple blocks from being reached, resulting in an error.
To fix it, I added a condition that makes sure that the delimiter doesn't appear inside the code:

[`if match and match.group("delim") not in match.group("code"):`](https://github.com/mbaruh/bot/blob/20c5a6946a140ef9e79f8a7c4edb60e2d5372298/bot/exts/utils/snekbox.py#L92-L93)

I considered at first amending `FORMATTED_CODE_REGEX`, but that would over-complicate it and could create bugs with the original function of the command.
There are other conditions that could be made, and I could switch the order of what's checked first, but I didn't see any specific advantage.

### DIFF
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -88,7 +88,7 @@ class Snekbox(Cog):
         """Extract code from the Markdown, format it, and insert it into the code template."""
         match = FORMATTED_CODE_REGEX.fullmatch(code)
 
-        # Despite the wildcard being lazy, this is a fullmatch so we need to check the presence of the delim explicitly.
+        # Despite the wildcard being lazy, the pattern is from start to end and will eat any delimiters in the middle.
         if match and match.group("delim") not in match.group("code"):
             code, block, lang, delim = match.group("code", "block", "lang", "delim")
             code = textwrap.dedent(code)
@@ -100,7 +100,7 @@ class Snekbox(Cog):
 
         else:
             code_parts = CODE_BLOCK_REGEX.finditer(code)
-            merge = '\n'.join(map(lambda part: part.group("code"), code_parts))
+            merge = '\n'.join(part.group("code") for part in code_parts)
             if merge:
                 code = textwrap.dedent(merge)
                 log.trace(f"Merged one or more code blocks from text combined with code:\n{code}")
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -21,23 +21,12 @@ log = logging.getLogger(__name__)
 
 ESCAPE_REGEX = re.compile("[`\u202E\u200B]{3,}")
 FORMATTED_CODE_REGEX = re.compile(
-    r"^\s*"                                 # any leading whitespace from the beginning of the string
     r"(?P<delim>(?P<block>```)|``?)"        # code delimiter: 1-3 backticks; (?P=block) only matches if it's a block
     r"(?(block)(?:(?P<lang>[a-z]+)\n)?)"    # if we're in a block, match optional language (only letters plus newline)
     r"(?:[ \t]*\n)*"                        # any blank (empty or tabs/spaces only) lines before the code
     r"(?P<code>.*?)"                        # extract all code inside the markup
     r"\s*"                                  # any more whitespace before the end of the code markup
-    r"(?P=delim)"                           # match the exact same delimiter from the start again
-    r"\s*$",                                # any trailing whitespace until the end of the string
-    re.DOTALL | re.IGNORECASE               # "." also matches newlines, case insensitive
-)
-CODE_BLOCK_REGEX = re.compile(
-    r"```"                                  # code block delimiter: 3 batckticks
-    r"([a-z]+\n)?"                          # match optional language (only letters plus newline)
-    r"(?:[ \t]*\n)*"                        # any blank (empty or tabs/spaces only) lines before the code
-    r"(?P<code>.*?)"                        # extract all code inside the markup
-    r"\s*"                                  # any more whitespace before the end of the code markup
-    r"```",                                 # code block end
+    r"(?P=delim)",                          # match the exact same delimiter from the start again
     re.DOTALL | re.IGNORECASE               # "." also matches newlines, case insensitive
 )
 RAW_CODE_REGEX = re.compile(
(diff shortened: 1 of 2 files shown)


MarkKoz (commented):
> I considered at first amending FORMATTED_CODE_REGEX, but that would over-complicate it and could create bugs with the original function of the command.

I don't see what's overcomplicated. You can just delete the previous matching code and regex and use yours. It looks like it will for just fine for one match too. You may not get all the extra info for logging easily, but that isn't really important and can be discarded in my opinion.

mbaruh:
Not sure  I understand. The pattern I added doesn't deal with inline code at all. Meaning if someone were to send:
\`print("hi")\`
as a message, the pattern I added won't catch it.

mbaruh:
Ok so you're saying we shouldn't be sticking to a fullmatch anyway. 
I don't mind, the reason I didn't go that route is that it would slightly repurpose the current logging.

MarkKoz:
Fair enough, but it's just trace logging; not particularly important. What you did is fine, but, if possible (seems like it is), I'd rather not have 2 almost-identical regular expressions in the code.

mbaruh:
This is how it would look like:

```py
FORMATTED_CODE_REGEX = re.compile(
    r"(?P<delim>(?P<block>```)|``?)"        # code delimiter: 1-3 backticks; (?P=block) only matches if it's a block
    r"(?(block)(?:(?P<lang>[a-z]+)\n)?)"    # if we're in a block, match optional language (only letters plus newline)
    r"(?:[ \t]*\n)*"                        # any blank (empty or tabs/spaces only) lines before the code
    r"(?P<code>.*?)"                        # extract all code inside the markup
    r"\s*"                                  # any more whitespace before the end of the code markup
    r"(?P=delim)",                          # match the exact same delimiter from the start again
    re.DOTALL | re.IGNORECASE               # "." also matches newlines, case insensitive
)
RAW_CODE_REGEX = re.compile(
    r"^(?:[ \t]*\n)*"                       # any blank (empty or tabs/spaces only) lines before the code
    r"(?P<code>.*?)"                        # extract all the rest as code
    r"\s*$",                                # any trailing whitespace until the end of the string
    re.DOTALL                               # "." also matches newlines
)

@staticmethod
    def prepare_input(code: str) -> str:
        """Extract code from the Markdown, format it, and insert it into the code template."""
        code_parts = list(FORMATTED_CODE_REGEX.finditer(code))

        if len(code_parts) == 1:
            code, block, lang, delim = code_parts[0].group("code", "block", "lang", "delim")
            code = textwrap.dedent(code)
            if block:
                info = (f"'{lang}' highlighted" if lang else "plain") + " code block"
            else:
                info = f"{delim}-enclosed inline code"
            log.trace(f"Extracted {info} for evaluation:\n{code}")

        elif len(code_parts) > 1:
            code = '\n'.join(part.group("code") for part in code_parts if part.group("block"))
            code = textwrap.dedent(code)
            log.trace(f"Merged several code blocks from text combined with code:\n{code}")

        else:
            code = textwrap.dedent(RAW_CODE_REGEX.fullmatch(code).group("code"))
            log.trace(
                f"Eval message contains unformatted or badly formatted code, "
                f"stripping whitespace only:\n{code}"
            )

        return code
```

MarkKoz:
That's not quite right because you specifically need to check if there is more than 1 code block. Currently you check if theres more than 1 match. A match could be an inline code block too, so you end up with a case of code being blank if the input has two inline code blocks.

MarkKoz:
I believe it needs to be something like this

```py
matches = list(FORMATTED_CODE_REGEX.finditer(code))
blocks = [match for match in matches if match.group("block")]
final = blocks if len(blocks) > 1 else matches[0]
code = textwrap.dedent(', '.join(final))
```

Anyone else have an opinion on if this is better than 2 separate regular expressions?

mbaruh:
Ok, I really overestimated the importance of the logs then

mbaruh:
Had to change the logic a bit, so that it would actually take the code block when there is only one code block, but there is another one inline before it:

```py
@staticmethod
def prepare_input(code: str) -> str:
    """Extract code from the Markdown, format it, and insert it into the code template."""
    if match := list(FORMATTED_CODE_REGEX.finditer(code)):
        blocks = [block for block in match if block.group("block")]

        if len(blocks) > 1:
            code = '\n'.join(block.group("code") for block in blocks)
            info = "several code blocks"
        else:
            match = match[0] if len(blocks) == 0 else blocks[0]
            code, block, lang, delim = match.group("code", "block", "lang", "delim")
            if block:
                info = (f"'{lang}' highlighted" if lang else "plain") + " code block"
            else:
                info = f"{delim}-enclosed inline code"
    else:
        code = RAW_CODE_REGEX.fullmatch(code).group("code")
        info = "unformatted or badly formatted code"

    code = textwrap.dedent(code)
    log.trace(f"Extracted {info} for evaluation:\n{code}")
    return code
```

Also aggregated the dedent and logging

MarkKoz:
`match[0] if len(blocks) == 0 else blocks[0]` could be `blocks[0] if blocks else match[0]` but then len is fine if you prefer the explicitness. I was thinking about this after I posted the code. It's a good feature to prefer code blocks.

mbaruh:
The diff is kind of a mess now

MarkKoz (changes requested):
Everything seems to be in order behaviour-wise. It was a good refactor too. I'd like to see some comments or a description of what it does in the docstring. It's also missing tests but writing them is not a requirement.

mbaruh:
Let me know if the docstring isn't enough. Still trying to wrap my ahead around the coding style in this repo

mbaruh:
This feels somewhat inconsistent to me with our other commands (3 vs 3d), however I don't see an elegant way to deal with it considering you can only specify a number of days.

bot/exts/utils/snekbox.py
```diff
@@ -74,7 +74,13 @@ async def upload_output(self, output: str) -> Optional[str]:
 
     @staticmethod
     def prepare_input(code: str) -> str:
-        """Extract code from the Markdown, format it, and insert it into the code template."""
+        """
+        Extract code from the Markdown, format it, and insert it into the code template.
+
+        If there is Markdown, ignores surrounding text.
+        If there are several Markdown parts in the message, concatenates only the code blocks.
+        If there is inline code but no code blocks, takes the first instance of inline code.
```
MarkKoz, line 82: Calling it "Markdown" is too vague.

```suggestion
        If there is any code block, ignore text outside the code block.
        If there are several fenced code blocks, concatenate only the fenced code blocks.
        Prefer a fenced code block over an inline code block if both are present.
```

MarkKoz (changes requested):

bot/exts/utils/snekbox.py
mbaruh: Hmm, I tried to rely on the terminology in the existing docstring. Would "formatted code" be more accurate?

mbaruh: Ah, missed the suggestion

mbaruh: Your third line doesn't convey the meaning in my third line. I tried to convey the preference by saying "concatenates only the code blocks", though I can add "and no inline code blocks"

MarkKoz: Third line? Did you mean second? It says "concatenate only the fenced code blocks" which makes it clear inline code blocks are excluded, because inline ≠ fenced.

mbaruh: I mean that the diff has a "-" on the third line "If there is inline code but no code blocks, takes the first instance of inline code" and I don't see an alternative in your suggestion

MarkKoz: Sorry, I get it now. I updated the suggestion. I suppose your original third line was okay but I feel mine is more succinct.

MarkKoz (approved):

Numerlor (approved):

Issue #1185, opened 2020-09-26
Gracefully handle statds being down


MarkKoz:
Sentry Issue: [BOT-7R](https://sentry.io/organizations/python-discord/issues/1824773198/?referrer=github_integration)

Since stats aren't critical, the bot shouldn't fatally crash if statsd is down. In this case, it was down due to a temporary outage.

```
gaierror: [Errno -2] Name or service not known
(1 additional frame(s) were not displayed)
...
  File "runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "__main__.py", line 28, in <module>
    bot = Bot(
  File "bot/bot.py", line 52, in __init__
    self.stats = AsyncStatsClient(self.loop, statsd_url, 8125, prefix="bot")
  File "bot/async_stats.py", line 18, in __init__
    family, _, _, _, addr = socket.getaddrinfo(
  File "socket.py", line 918, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
```

thomaspet:
I can probably look at this, if you assign it to @tagptroll1

PR #1213, opened 2020-10-05
fix(statsd): Gracefully handle socket.gaierro in AsyncStatsClient

thomaspet:
Per issue #1185  the bot might go down if the statsd client fails to connect during instantiation. This can be caused by an outage on their part, or network issues.
If this happens getaddrinfo will raise a socket.gaierror.
This PR catched the error, sets self.stats to None for the time being, and handles that elsewhere.
In addition, fallback logic was added to attempt to reconnect, in the off-chance it's a temporary outage

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -46,7 +46,25 @@ class Bot(commands.Bot):
             # will effectively disable stats.
             statsd_url = "127.0.0.1"
 
-        self.stats = AsyncStatsClient(self.loop, statsd_url, 8125, prefix="bot")
+        try:
+            self.stats = AsyncStatsClient(self.loop, statsd_url, 8125, prefix="bot")
+        except socket.gaierror as socket_error:
+            self.stats = None
+            self.loop.call_later(30, self.retry_statsd_connection, statsd_url)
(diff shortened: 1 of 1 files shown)


MarkKoz:
Setting self.stats to None is not a solution because all the code that tries to send stats will get an AttributeError.

thomaspet:
> 
> 
> Setting self.stats to None is not a solution because all the code that tries to send stats will get an AttributeError.
Did i miss anymore usages of it? I did a project search for it, and only found usage in bot.py and a mock in helpers.py

MarkKoz:
Yeah, it's used by many cogs, usually as `self.bot.stats`. For example, see the help channel cog.

thomaspet:
> 
> 
> Yeah, it's used by many cogs, usually as `self.bot.stats`. For example, see the help channel cog.

Jeez, now that's a facepalm from me. I did a search for self.stats, and found it really odd that was all the stats we tracked

tagptroll1:
Noticed I somehow opened this through my work account... Apologies, reopening on correct account

PR #1214, opened 2020-10-05
fix(statsd): Gracefully handle socket.gaierro in AsyncStatsClient

tagptroll1:
Per issue #1185 the bot might go down if the statsd client fails to connect during instantiation. This can be caused by an outage on their part, or network issues.
If this happens getaddrinfo will raise a socket.gaierror.
This PR catched the error, sets self.stats to None for the time being, and handles that elsewhere.
In addition, fallback logic was added to attempt to reconnect, in the off-chance it's a temporary outage


bot/bot.py
```diff
@@ -44,9 +45,27 @@ def __init__(self, *args, redis_session: RedisSession, **kwargs):
...
             # will effectively disable stats.
-            statsd_url = "127.0.0.1"
+            statsd_url = LOCALHOST
+
+        try:
+            self.stats = AsyncStatsClient(self.loop, statsd_url, 8125, prefix="bot")
+        except socket.gaierror as socket_error:
+            self.stats = AsyncStatsClient(self.loop, LOCALHOST)
+            self.loop.call_later(30, self.retry_statsd_connection, statsd_url)
+            log.warning(f"Statsd client failed to instantiate with error:\n{socket_error}")
```
MarkKoz, line 55: This feels redundant. It can be replaced with a call to `retry_statsd_connection` if you do `self.stats = AsyncStatsClient(self.loop, LOCALHOST)` inside the function. Maybe after an `if attempt == 1`. Or you could set it before you call the function.

MarkKoz (changes requested):

bot/bot.py
```diff
@@ -44,9 +45,23 @@ def __init__(self, *args, redis_session: RedisSession, **kwargs):
...
             # For this reason, setting the statsd host to 127.0.0.1 for development
             # will effectively disable stats.
-            statsd_url = "127.0.0.1"
+            statsd_url = LOCALHOST
 
-        self.stats = AsyncStatsClient(self.loop, statsd_url, 8125, prefix="bot")
+        self.stats = AsyncStatsClient(self.loop, LOCALHOST)
+        self.connect_statsd(statsd_url)
+
+    def connect_statsd(self, statsd_url: str, retry_after: int = 2, attempt: int = 1) -> None:
```
MarkKoz, line 53: This should be named as a private method.

bot/bot.py
```diff
@@ -44,9 +45,23 @@ def __init__(self, *args, redis_session: RedisSession, **kwargs):
...
+        if attempt > 5:
+            log.error("Reached 10 attempts trying to reconnect AsyncStatsClient. Aborting")
+            return
+
+        try:
+            self.stats = AsyncStatsClient(self.loop, statsd_url, 8125, prefix="bot")
+        except socket.gaierror:
+            log.warning(f"Statsd client failed to connect (Attempts: {attempt})")
+            # Use a fallback strategy for retrying, up to 5 times.
+            self.loop.call_later(retry_after, self.retry_statsd_connection, statsd_url, retry_after ** 2, attempt + 1)
```
MarkKoz, line 64: You need to change `retry_statsd_connection` since you renamed the function

MarkKoz, line 64: Does this need to be explicitly cancelled if the bot stops, or does asyncio kind of take care of this already?

MarkKoz, line 64: The sixth attempt will wait over 18 hours, which is too excessive. 5 attempts will only be a bit over 4 minutes. I don't think an exponential function is good, because waiting 4 minutes is not that long, but the next best is 18 hours, which is too long.

bot/bot.py
```diff
@@ -44,9 +45,23 @@ def __init__(self, *args, redis_session: RedisSession, **kwargs):
...
+            statsd_url = LOCALHOST
 
-        self.stats = AsyncStatsClient(self.loop, statsd_url, 8125, prefix="bot")
+        self.stats = AsyncStatsClient(self.loop, LOCALHOST)
+        self.connect_statsd(statsd_url)
+
+    def connect_statsd(self, statsd_url: str, retry_after: int = 2, attempt: int = 1) -> None:
+        """Callback used to retry a connection to statsd if it should fail."""
+        if attempt > 5:
+            log.error("Reached 10 attempts trying to reconnect AsyncStatsClient. Aborting")
```
MarkKoz, line 56: This error message has the wrong count. As things currently stand, it should say "6 attempts".

MarkKoz (changes requested):

thomaspet:
I must've been drowzy revisting this. Will fix the remarks later today

tagptroll1:
> 
> 
> I must've been drowzy revisting this. Will fix the remarks later today

Then he died, and revived.

bot/bot.py
tagptroll1: Dont know, went ahead and put up some logic for it.

bot/bot.py
```diff
@@ -44,9 +46,27 @@ def __init__(self, *args, redis_session: RedisSession, **kwargs):
...
+            log.error("Reached 5 attempts trying to reconnect AsyncStatsClient. Aborting")
+            return
+
+        try:
+            self.stats = AsyncStatsClient(self.loop, statsd_url, 8125, prefix="bot")
+        except socket.gaierror:
+            log.warning(f"Statsd client failed to connect (Attempt(s): {attempt})")
+            # Use a fallback strategy for retrying, up to 5 times.
+            self._statsd_timerhandle = self.loop.call_later(
+                retry_after, self._connect_statsd, statsd_url, retry_after * 5, attempt + 1)
```
MarkKoz, line 69: Should follow the style guide. Each argument needs to be on a separate line as well as the closing parenthesis.

bot/bot.py
```diff
@@ -152,6 +172,9 @@ async def close(self) -> None:
         if self.redis_session:
             await self.redis_session.close()
 
+        if self._statsd_timerhandle and not self._statsd_timerhandle.cancelled:
```
MarkKoz, line 175: Checking if it's cancelled is redundant

> If the callback has already been canceled or executed, this method has no effect.

bot/bot.py
```diff
@@ -44,9 +46,27 @@ def __init__(self, *args, redis_session: RedisSession, **kwargs):
...
+            statsd_url = LOCALHOST
 
-        self.stats = AsyncStatsClient(self.loop, statsd_url, 8125, prefix="bot")
+        self.stats = AsyncStatsClient(self.loop, LOCALHOST)
+        self._connect_statsd(statsd_url)
+
+    def _connect_statsd(self, statsd_url: str, retry_after: int = 2, attempt: int = 1) -> None:
+        """Callback used to retry a connection to statsd if it should fail."""
+        if self._statsd_timerhandle and not self._statsd_timerhandle.cancelled:
+            self._statsd_timerhandle.cancel()
```
MarkKoz, line 57: Why is it being cancelled in here too? I don't see how the function can be called again before a previous one is executed.

bot/bot.py
MarkKoz, line 69: I think `retry_after = attempt * 2` would be better along with increasing the maximum attempts to 8. Will end up with the following wait times for each subsequent attempt:

```
1
2
4
8
16
32
64
128
``` 

This total 255 minutes or about 4.5 hours before giving up. This has a nice balance of having short wait times when the issue is brief and long wait times when it's a longer outage.

MarkKoz (changes requested):

bot/bot.py
tagptroll1: Mostly to ensure it wasn't running before i reassigned the attribute

MarkKoz (approved):

ks129 (approved):
Looks good to me

PR #1215, opened 2020-10-05
Support alternate gating methods

jb3:
This PR extends the verification cog to support users joining through alternate gating methods.

It will send a separate DM for any user verifying through a gating method which is not `!accept` and will completely disable the welcome DM.

image

### DIFF
--- a/bot/exts/moderation/verification.py
+++ b/bot/exts/moderation/verification.py
@@ -553,7 +553,7 @@ class Verification(Cog):
             log.exception("DM dispatch failed on unexpected error code")
 
     @Cog.listener()
-    async def on_member_update(self, before: discord.Member, after: discord.Member):
+    async def on_member_update(self, before: discord.Member, after: discord.Member) -> None:
         """Check if we need to send a verification DM to a gated user."""
         before_roles = [r.id for r in before.roles]
         after_roles = [r.id for r in after.roles]
--- a/bot/exts/moderation/verification.py
+++ b/bot/exts/moderation/verification.py
@@ -173,7 +173,7 @@ class Verification(Cog):
     # ]
     task_cache = RedisCache()
 
-    # Cache who needs to receive an alternate verified DM.
+    # Create a cache for storing recipients of the alternate welcome DM.
     member_gating_cache = RedisCache()
 
     def __init__(self, bot: Bot) -> None:
--- a/bot/exts/moderation/verification.py
+++ b/bot/exts/moderation/verification.py
@@ -53,6 +53,23 @@ If you'd like to unsubscribe from the announcement notifications, simply send `!
 <#{constants.Channels.bot_commands}>.
 """
 
+ALTERNATE_VERIFIED_MESSAGE = f"""
+Thanks for accepting our rules!
+
+You can find a copy of our rules for reference at <https://pythondiscord.com/pages/rules>.
+
+Additionally, if you'd like to receive notifications for the announcements \
+we post in <#{constants.Channels.announcements}>
+from time to time, you can send `!subscribe` to <#{constants.Channels.bot_commands}> at any time \
+to assign yourself the **Announcements** role. We'll mention this role every time we make an announcement.
+
+If you'd like to unsubscribe from the announcement notifications, simply send `!unsubscribe` to \
+<#{constants.Channels.bot_commands}>.
+
+To introduce you to our community, we've made the following video:
+https://youtu.be/ZH26PuX3re0
+"""
+
 # Sent via DMs to users kicked for failing to verify
 KICKED_MESSAGE = f"""
 Hi! You have been automatically kicked from Python Discord as you have failed to accept our rules \
@@ -155,6 +172,9 @@ class Verification(Cog):
     #   "last_reminder": int (discord.Message.id),
     # ]
     task_cache = RedisCache()
+
+    # Cache who needs to receive an alternate verified DM.
+    member_gating_cache = RedisCache()
 
     def __init__(self, bot: Bot) -> None:
         """Start internal tasks."""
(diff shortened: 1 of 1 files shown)


bot/exts/moderation/verification.py
```diff
@@ -519,12 +539,34 @@ async def on_member_join(self, member: discord.Member) -> None:
...
+        after_roles = [r.id for r in after.roles]
+
+        if constants.Roles.verified not in before_roles and constants.Roles.verified in after_roles:
+            if await self.member_gating_cache.get(after.id):
+                try:
+                    await safe_dm(after.send(ALTERNATE_VERIFIED_MESSAGE))
+                except discord.HTTPException:
+                    log.exception("DM dispatch failed on unexpected error code")
+                finally:
+                    self.member_gating_cache.pop(after.id)
```
MarkKoz, line 568: Pretty sure this must be awaited.

bot/exts/moderation/verification.py
```diff
@@ -519,12 +539,34 @@ async def on_member_join(self, member: discord.Member) -> None:
...
             log.exception("DM dispatch failed on unexpected error code")
 
+    @Cog.listener()
+    async def on_member_update(self, before: discord.Member, after: discord.Member) -> None:
+        """Check if we need to send a verification DM to a gated user."""
+        before_roles = [r.id for r in before.roles]
+        after_roles = [r.id for r in after.roles]
+
+        if constants.Roles.verified not in before_roles and constants.Roles.verified in after_roles:
+            if await self.member_gating_cache.get(after.id):
```
MarkKoz, line 562: I think you can merge the pop operation into this since it will silently fail if the key doesn't exist. I don't have the docs on hand so please confirm.

MarkKoz (changes requested):

bot/exts/moderation/verification.py
lemonsaurus: Not sure I love using pop as delete. just do cache.delete if you don't need it.

bot/exts/moderation/verification.py
```diff
@@ -519,12 +539,34 @@ async def on_member_join(self, member: discord.Member) -> None:
...
+                try:
+                    await safe_dm(after.send(ALTERNATE_VERIFIED_MESSAGE))
```
lemonsaurus, line 564: feels like a good place for a block comment explaining what happens. this line in particular is a bit opaque.

bot/exts/moderation/verification.py
```diff
@@ -519,12 +539,34 @@ async def on_member_join(self, member: discord.Member) -> None:
...
         log.trace(f"Sending on join message to new member: {member.id}")
         try:
             await safe_dm(member.send(ON_JOIN_MESSAGE))
         except discord.HTTPException:
             log.exception("DM dispatch failed on unexpected error code")
 
+    @Cog.listener()
+    async def on_member_update(self, before: discord.Member, after: discord.Member) -> None:
+        """Check if we need to send a verification DM to a gated user."""
+        before_roles = [r.id for r in before.roles]
```
lemonsaurus, line 558: nitpicking, but I'd prefer just calling this role instead of r

bot/exts/moderation/verification.py
```diff
@@ -519,12 +539,32 @@ async def on_member_join(self, member: discord.Member) -> None:
         if member.guild.id != constants.Guild.id:
             return  # Only listen for PyDis events
 
+        raw_member = await self.bot.http.get_member(member.guild.id, member.id)
+
+        # Only send the message to users going through our gating system
```
lemonsaurus, line 544: a little unclear now that we technically have two gating systems.

bot/exts/moderation/verification.py
jb3: Addressed in 880b936faf83d8fa3d7489e1f9eaab89b93af1b8

bot/exts/moderation/verification.py
jb3: Indeed, you can, addressed in 880b936faf83d8fa3d7489e1f9eaab89b93af1b8

bot/exts/moderation/verification.py
```diff
@@ -156,6 +173,9 @@ class Verification(Cog):
     # ]
     task_cache = RedisCache()
 
+    # Cache who needs to receive an alternate verified DM.
```
lemonsaurus, line 176: the imperative way you phrased this makes it sound like the next line caches something, when it actually just creates a cache.

lemonsaurus (approved):
Looks great overall. Couple nitpicks, but nothing that needs to block merge. fix if you want.

bot/exts/moderation/verification.py
jb3: This was changed in 880b936faf83d8fa3d7489e1f9eaab89b93af1b8, but yep.

bot/exts/moderation/verification.py
```diff
@@ -519,12 +539,32 @@ async def on_member_join(self, member: discord.Member) -> None:
         if member.guild.id != constants.Guild.id:
             return  # Only listen for PyDis events
 
+        raw_member = await self.bot.http.get_member(member.guild.id, member.id)
+
+        # Only send the message to users going through our gating system
+        if raw_member["is_pending"]:
+            await self.member_gating_cache.set(raw_member.id, True)
```
MarkKoz, line 546: This should be `member.id`

```
2020-10-05 16:42:18 | bot | ERROR | Unhandled exception in on_member_join.
  File "/home/mark/repos/python/bot-pydis/.venv/lib/python3.8/site-packages/discord/client.py", line 333, in _run_event
    await coro(*args, **kwargs)
  File "/home/mark/repos/python/bot-pydis/bot/exts/moderation/verification.py", line 546, in on_member_join
    await self.member_gating_cache.set(raw_member.id, True)
AttributeError: 'dict' object has no attribute 'id'
```

MarkKoz (changes requested):

bot/exts/moderation/verification.py
MarkKoz: Or `raw_member["id"]` I suppose.

bot/exts/moderation/verification.py
jb3: Addressed in 8d6d3ef56d29c2eff372c858fa0a228eeefdbfb8

bot/exts/moderation/verification.py
jb3: Addressed in 406da780c06a6797b860d816c4a418def9a3f116

MarkKoz (approved):

bot/exts/moderation/verification.py
jb3: Addressed in ea3217effacc02e06444ea0b21985cd7439a13e7

bot/exts/moderation/verification.py
jb3: Addressed in ceffe46a0d5136308c8f0684c2c406dd34e758fb

bot/exts/moderation/verification.py
jb3: Addressed in 082e26342eb3faf104523334cdb87e07eda03db3

PR #1216, opened 2020-10-06
Fix old nick in superstarify reason

mbaruh:
For some reason the old nick is currently taken after the infraction is applied, meaning that the old nick, taken from the display name, is actually the new nick.

I just changed the point where the variable is made to be before applying the infraction.


Akarys42 (approved):
Looks good to me

kwzrd (approved):

PR #1219, opened 2020-10-06
CI: avoid failing whole job if a cache task fails

MarkKoz:
Restoring from cache is non-critical. The CI can recover if cache tasks fail.

I'm not sure if we can rely on `cacheHitVar` remaining unset when a cache task fails. This variable is used in the condition of pipenv steps:

```yaml
condition: and(succeeded(), ne(variables.PY_ENV_RESTORED, 'true'))
```

If it's not true, then pipenv won't be installed along with dependencies. If the cache task sets the variable before trying to restore (and failing), then this will make the job fail anyway since pipenv won't get installed. That's really no worse than the job failing at the cache task without `continueOnError: true`. Therefore, I don't mind waiting and seeing if this happens. I could be proactive and add a more complicated, ugly condition to the pipenv steps, but I'd rather not.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,7 +21,6 @@ jobs:
       BOT_TOKEN: bar
       REDDIT_CLIENT_ID: spam
       REDDIT_SECRET: ham
-      WOLFRAM_API_KEY: baz
       REDIS_PASSWORD: ''
 
     steps:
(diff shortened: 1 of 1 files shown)


thomaspet:
Sounds fair, but I don't know a lot of azure-pipelines, unfortunately.

kwzrd (approved):
Makes sense to me, let's give it a try.

Senjan21 (approved):
It looks good I agree that we should try it

PR #1220, opened 2020-10-07
Add the ability to purge and ban in one command.

scragly:
Multiple times there's been cases where we've needed to purge a users messages due to spam or unsavoury content after banning via the bot command. As discord do support providing an argument to automatically purge a set number of days worth of messages when banning the user, it could potentially save mods a lot of time and effort.

Relevant links:
[Docs for Guild.ban()](https://discordpy.readthedocs.io/en/neo-docs/api.html#discord.Guild.ban)

The implementation I've gone with is to add a new, separate command to deal with purging on bans, called, hopefully intuitively, `purgeban`.

The command has the following signature:
`!purgeban <user> [purge_days=1] [reason]`

The usage is exactly the same as the original `ban` command, except we can _optionally_ add a number before the reason that represents the days worth of messages to be deleted.

E.g.:
 - `!ban 394529085923262464 This is a ban reason`
   - This would ban the user with ID 394529085923262464, removing 1 day of messages.
 - `!ban 394529085923262464 3 This is a ban reason`
   - This would ban the user with ID 394529085923262464, removing 3 days of messages.

This _does_ introduce a potential issue of reasons that start with numbers causing it to be interpreted as the purge_days value, however this should be minimised with the clearly distinct alternate command usage; if they're remembering to seek out specifically the purgeban command, they'll likely to explicitly remember to not start the reason with a number.

### DIFF
--- a/bot/exts/moderation/infraction/infractions.py
+++ b/bot/exts/moderation/infraction/infractions.py
@@ -71,6 +71,23 @@ class Infractions(InfractionScheduler, commands.Cog):
         """Permanently ban a user for the given reason and stop watching them with Big Brother."""
         await self.apply_ban(ctx, user, reason)
 
+    @command()
+    async def purgeban(
+        self,
+        ctx: Context,
+        user: FetchedMember,
+        purge_days: t.Optional[int] = 1,
+        *,
+        reason: t.Optional[str] = None
+    ) -> None:
+        """
+        Same as ban but removes all their messages for the given number of days, default being 1.
+
+        `purge_days` can only be values between 0 and 7.
+        Anything outside these bounds are automatically adjusted to their respective limits.
+        """
+        await self.apply_ban(ctx, user, reason, max(min(purge_days, 7), 0))
+
     # endregion
     # region: Temporary infractions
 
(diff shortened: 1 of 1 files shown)


bot/exts/moderation/infraction/infractions.py
```diff
@@ -71,6 +71,23 @@ async def ban(self, ctx: Context, user: FetchedMember, *, reason: t.Optional[str
         """Permanently ban a user for the given reason and stop watching them with Big Brother."""
         await self.apply_ban(ctx, user, reason)
 
+    @command()
```
Den4200, line 74: Maybe we could add `pban` as an alias?

Den4200 (approved):
Looks and works great - only one small thing!

MarkKoz (approved):

mbaruh:
This feels somewhat inconsistent to me with our other commands (3 vs 3d), however I don't see an elegant way to deal with it considering you can only specify a number of days.

scragly:
> This feels somewhat inconsistent to me with our other commands (3 vs 3d), however I don't see an elegant way to deal with it considering you can only specify a number of days.

They're different argument types, so I don't consider it a matter of inelegance at all.

bot/exts/moderation/infraction/infractions.py
scragly: Sounds fine to me

PR #1227, opened 2020-10-10
Duckpond: Add a list of already ducked messages

Akarys42:
Previously, race conditions caused the messages to be processed again before knowing the white checkmark reaction got added, this seems to solve it.


scragly (approved):

lemonsaurus (approved):
Yeah, this'll probably work.

PR #1228, opened 2020-10-11
[Critical] Add Metricity DB URL to site (docker-compose.yml)

ks129:
Currently this is impossible to run bot with `docker-compose up` because site will fail start, as now this require metricity DB URL as environment variable.


RohanJnr (approved):

Den4200 (approved):
Thanks, good catch!

PR #1231, opened 2020-10-13
Fixed hardcoded prefix in __main__.py

spitfire-hash:
This is a little commit, but the prefix was hardcoded in [\_\_main\_\_.py](https://github.com/python-discord/bot/blob/b081db8418aedf113bab17e0f44313840b965b8e/bot/__main__.py#L61). This PR should fix it.


kwzrd (approved):
Thanks for the contribution, this is great. It's useful for dev bots.

image

PR #1233, opened 2020-10-13
Verification: add note to kicked members

kwzrd:
We've now had the Verification task kicking dormant unverified users from the guild for some time, and @lemonsaurus has raised the issue of having no easy way to determine whether a specific user was previously kicked by the system, and how many times.

## Solution

In order to amend this issue, we decided to post a note infraction to site on-kick. This will give mods the ability to see that a specific user was kicked in their infraction history:

image

My bot is called `lana del rey` and the alt I kicked is called `lana del rey 2`, sorry that's confusing.

## Implementation details

I decided to write my own small wrapper for the post call, as what's available in the current mod tooling is not compatible. Notes are usually routed through `post_infraction`, which is written with the assumption of being called from a command ~ it wants a `discord.Context` to infer some of the information, and then uses the context to send an error message to should the post fail.

https://github.com/python-discord/bot/blob/eb82da94b5101e8399b0ac168d049cbf2d3ec93b/bot/exts/moderation/infraction/_utils.py#L71-L79

Auto-mod gets around this by building up a ctx locally & then mutating it, I'd rather not do that.

## Considerations

This shouldn't race with the member syncer because the member will not be removed from the DB once the are kicked, *but* they are patched with `in_guild=False` once the guild leave event comes. Should we decide to remove member records from the DB on-leave, this approach will no longer be feasible.

This was discussed, but worth mentioning: because all kicked members will now have infraction records, they will remain in the DB even after the monthly cleanse. We agreed that this is ok.

## Review & testing

Testing this is a little tricky, but I try to give everyone the `@Verified` role in the dev environment to make them immune to kicks, adjust the config to kick on day 0, then kick my alt. It's possible to comment out the `member.kick` call above the post if you want. The payload is logged at trace level so you can see who you've kicked and whose infr history to search. Let me know if you need help orchestrating this.

For review please make sure that I'm constructing the payload correctly, but I copied it over from existing mod tooling & consulted site docs, so I think it's fine. Also please pay extra attention to what effect this may have on the member DB & our data policy, and whether this could race with some other feature.

### DIFF
--- a/bot/exts/moderation/verification.py
+++ b/bot/exts/moderation/verification.py
@@ -367,7 +367,7 @@ class Verification(Cog):
             "actor": self.bot.user.id,  # Bot actions this autonomously
             "expires_at": None,
             "hidden": True,
-            "reason": f"Kicked for not having verified after {constants.Verification.kicked_after} days",
+            "reason": "Verification kick",
             "type": "note",
             "user": member.id,
         }
--- a/bot/exts/moderation/verification.py
+++ b/bot/exts/moderation/verification.py
@@ -396,6 +396,7 @@ class Verification(Cog):
             except discord.HTTPException as suspicious_exception:
                 raise StopExecution(reason=suspicious_exception)
             await member.kick(reason=f"User has not verified in {constants.Verification.kicked_after} days")
+            await self._add_kick_note(member)
 
         n_kicked = await self._send_requests(members, kick_request, Limit(batch_size=2, sleep_secs=1))
         self.bot.stats.incr("verification.kicked", count=n_kicked)
--- a/bot/exts/moderation/verification.py
+++ b/bot/exts/moderation/verification.py
@@ -11,6 +11,7 @@ from discord.ext.commands import Cog, Context, command, group, has_any_role
 from discord.utils import snowflake_time
 
 from bot import constants
+from bot.api import ResponseCodeError
 from bot.bot import Bot
 from bot.decorators import has_no_roles, in_whitelist
 from bot.exts.moderation.modlog import ModLog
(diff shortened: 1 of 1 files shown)


MarkKoz (approved):
I didn't think this would be a useful feature, but if staff needs it then it's fine.

kwzrd:
This may also pollute infraction stats a little bit. Do you think that's ok @jb3? We can probably just filter them out by the reason field, but it adds some complexity.

jb3:
@kwzrd Maybe instead of having a general text note we could just simplify the reason to `verification kick`. I'm not sure we need to know the threshold, since I don't think we'll be changing it, and that would wildly simplify filtering.

kwzrd:
@jb3 Yes, I agree, let's simplify it: 308e6a6.

bot/exts/moderation/verification.py
```diff
@@ -355,6 +356,28 @@ async def _send_requests(self, members: t.Collection[discord.Member], request: R
...
+            "active": False,
+            "actor": self.bot.user.id,  # Bot actions this autonomously
+            "expires_at": None,
+            "hidden": True,
+            "reason": "Verification kick",
+            "type": "note",
+            "user": member.id,
+        }
+
+        log.trace(f"Posting kick note: {payload!r}")
```
Numerlor, line 375: Is logging the whole payload useful here? The only information that really changes is the user id

kwzrd: Could just log the member id if you think it's too noisy. I wanted to log that the request was made and figured I'd also log the payload given that trace logs are usually off in production anyway, but you're right that it's perhaps redundant.

Numerlor: I don't think it's too noisy (unless your terminal line wraps or something like that), but logging the whole payload does push out the user id a bit further, which I think is the thing you'd usually be looking at. Since it's a Member object we could for example log `f"Posting kick note for user {member} ({member.id})"` to expand on the id

kwzrd: It comes with the slight risk that someone changes the payload but forgets to update the log, causing it to lie. But you're right that it obfuscates the important information. I'll adjust it.

kwzrd: Looks better on 54fb163:

image

Numerlor (approved):

Issue #1025, opened 2020-07-03
Command to time code execution


wookie184:
It could be useful to have a [timeit](https://docs.python.org/3/library/timeit.html#timeit.timeit)-like bot command to time code execution. Some ideas of features the command could have are:
- The ability to specify how many times the code should be run. It could also be interesting if this was calculated automatically like how the [timeit Command-Line Interface](https://docs.python.org/3/library/timeit.html#timeit-command-line-interface) does.
- The ability to take in multiple snippets of code to compare
- The output could just be the time taken as `timeit.timeit` gives (e.g `0.3018611848820001`) but it would be nicer if the output was in a more human friendly format (e.g. the timeit CLI outputs more like `10000 loops, best of 5: 30.2 usec per loop`)

I would assume it wouldn't be too difficult to implement as it could borrow from/use most of the same code as `!eval` but I am not sure.

MarkKoz:
This sounds pretty easy. It can generate a `timeit()` code snippet and insert the given code into the first argument. It'd need to have a way to escape quotes. That new code can then be sent to Snekbox.

Supporting multiple code snippets is possible, but adds complexity to the input parser. Furthermore, Snekbox still has a time limit which all snippets would share. I think it's best to not support that; it isn't much trouble for users to invoke a separate eval.

MarkKoz:
That being said, it may make more sense to add another endpoint to Snekbox, since it's pretty simple on that side to just invoke `python -m timeit` rather than `python -c`.

Akarys42:
I think the implementation we choose for this feature would depend on if decide to introduce a mypy endpoint; if we already have two endpoints I think it would make sense to have a third one, otherwise wrapping the code in a timeit snippet makes more sense IMO

davisschenk:
Id like to implement this feature

It would be fairly trivial to wrap the code in a timeit snippet to test the time it takes. Currently Snekbox only has one endpoint and im not sure it would make sense to reimplement eval with `python -m timeit <code>` when its easy to just eval `timeit.main([<code>])`

MarkKoz:
@davisschenk That approach sounds fine, though it is noteworthy that `main()` is undocumented. I'm not sure if this is unintentional or if it implies that it's not part of the public API. In any case, I'll assign you to this issue.

davisschenk:
`main([<args>])` is just the equivalent of `python -m timeit <args>` 

[...]

PR #1234, opened 2020-10-14
Add timeit eval command

davisschenk:
Implemented the timeit command proposed in issue #1025 

This command is very similar to calling `python -m  timeit <code>`, It maintains all of the previous features of the `eval` command with the addition of timing the code using `timeit`.

image

### DIFF
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -217,6 +217,62 @@ class Snekbox(Cog):
             log.info(f"{ctx.author}'s job had a return code of {results['returncode']}")
         return response
 
+    @staticmethod
+    def get_time(results: dict) -> Tuple[str, Optional[str]]:
+        output = results["stdout"].strip("\n")
+
+        if results["returncode"] == 0:
+            time_index = output.rfind("\n")
+            if time_index != -1:
+                return output[:time_index], output[time_index:]
+            return "", output
+
+        return output, None
+
+    async def send_timeit(self, ctx: Context, code: str) -> Message:
+        async with ctx.typing():
+            code = code.replace("'", r"\'")
+            code = f"import timeit\n" \
+                   f"timeit.main(['{code}'])"
+
+            results = await self.post_eval(code)
+            msg, error = self.get_results_message(results)
+            output, time = self.get_time(results)
+            icon = self.get_status_emoji(results)
+
+
+            if error:
+                output, paste_link = error, None
+            else:
+                output, paste_link = await self.format_output(output)
+
+            msg = f"{ctx.author.mention} {icon} {msg}.\n\n```\n{output}\n```"
+
+            if time:
+                msg = f"{msg}\nTime:\n```{time}```"
+
+            if paste_link:
+                msg = f"{msg}\nFull output: {paste_link}"
+
+            if icon == ":x:":
+                self.bot.stats.incr("snekbox.python.fail")
+            else:
+                self.bot.stats.incr("snekbox.python.success")
+
+            filter_cog = self.bot.get_cog("Filtering")
+            filter_triggered = False
+            if filter_cog:
+                filter_triggered = await filter_cog.filter_eval(msg, ctx.message)
+            if filter_triggered:
+                response = await ctx.send("Attempt to circumvent filter detected. Moderator team has been alerted.")
+            else:
+                response = await ctx.send(msg)
+            self.bot.loop.create_task(wait_for_deletion(response, (ctx.author.id,), ctx.bot))
+
+            log.info(f"{ctx.author}'s job had a return code of {results['returncode']}")
+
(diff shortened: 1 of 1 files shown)


bot/exts/utils/snekbox.py
```diff
@@ -21,21 +21,21 @@
...
 )
 RAW_CODE_REGEX = re.compile(
-    r"^(?:[ \t]*\n)*"                       # any blank (empty or tabs/spaces only) lines before the code
-    r"(?P<code>.*?)"                        # extract all the rest as code
-    r"\s*$",                                # any trailing whitespace until the end of the string
-    re.DOTALL                               # "." also matches newlines
+    r"^(?:[ \t]*\n)*"  # any blank (empty or tabs/spaces only) lines before the code
+    r"(?P<code>.*?)"  # extract all the rest as code
+    r"\s*$",  # any trailing whitespace until the end of the string
+    re.DOTALL  # "." also matches newlines
```
MarkKoz, line 38: Removing the spaces before the comments just makes them less readable.

bot/exts/utils/snekbox.py
```diff
@@ -217,6 +217,71 @@ async def send_eval(self, ctx: Context, code: str) -> Message:
...
+    async def send_timeit(self, ctx: Context, code: str) -> Message:
+        """
+        Evaluate code with timing, format it, and send the output to the corresponding channel.
+
+        Return the bot response.
+        """
+        async with ctx.typing():
+            code = code.replace("'", r"\'")
+            code = f"import timeit\n" \
+                   f"timeit.main(['{code}'])"
```
MarkKoz, line 256: I was envisioning this as a command that only returns the execution time. In other words, no output from the actual code. I'm not sure if everyone agrees with this direction.

In any case, it'd be difficult to suppress output when using `timeit.main()` since it also writes to stdout. `contextlib.redirect_stdout` doesn't really work since it's simple to bypass e.g. by writing to `/dev/stdout` directly.

bot/exts/utils/snekbox.py
```diff
@@ -217,6 +217,71 @@ async def send_eval(self, ctx: Context, code: str) -> Message:
...
+
+            if error:
+                output, paste_link = error, None
+            else:
+                output, paste_link = await self.format_output(output)
+
+            msg = f"{ctx.author.mention} {icon} {msg}.\n\n```\n{output}\n```"
+
+            if time:
+                msg = f"{msg}\nTime:\n```{time}```"
```
MarkKoz, line 271: The `output` will contain the time too if it's not truncated. You should remove it from the output.

MarkKoz (changes requested):

MarkKoz:
Can you rebase so that there aren't six commits of you trying to fix CI errors? This is a good example of why the contribution guidelines say that you should lint and test locally.

bot/exts/utils/snekbox.py
mbaruh: I agree, the tool would be used when wanting to compare running times, the output isn't the focus here. For that you have the normal eval. It's also more in line with how standard timing tools work.

Xithrius:
@davisschenk Are you planning to continue this?

Xithrius:
This PR is now available for anyone to take over. Please ping me if you do decide to do so.

Shivansh-007:
Shouldn’t this be part of python-discord/snekbox rather than breaking into the user code and appending few lines to it or making two calls?

bot/exts/utils/snekbox.py
```diff
@@ -217,6 +222,76 @@ async def send_eval(self, ctx: Context, code: str) -> Message:
...
+            if time_index != -1:
+                return output[:time_index], output[time_index:]
+            return "", output
+
+        return output, None
+
+    @staticmethod
+    def escape_code(code: str) -> str:
+        """Escapes quotes and newlines in code string."""
+        return code.translate(ESCAPE_TRANS)
```
ToxicKidz, line 245: This seems to give a `NameError` here.

ToxicKidz:
@Xithrius I'd like to take this.

MarkKoz:
> 
> 
> Shouldn’t this be part of python-discord/snekbox rather than breaking into the user code and appending few lines to it or making two calls?

Either way is fine as far as I am concerned. There was an idea to support supplying arbitrary Python arguments to the API (rather than adding a new endpoint specifically for timeit).

doublevcodes:
Is this PR still being worked on? If not, I'd be happy to complete it.

Shivansh-007:
I think we should wait until @MarkKoz implements the change in snekbox. This would reduce this PR down and won't make the bot break into the user's code to add some `timeit` lines.

ToxicKidz:
Since python-discord/snekbox#108 was merged, I will be implementing this in #1602.

PR #1240, opened 2020-10-16
display inf id actioned in mod channel

mbaruh:
It's often helpful to see the id of the infraction when applying it in a mod channel. This allows for example other mods to edit it without having to pull up the entire infraction history of the user.

With this PR it will look like that:
image


kwzrd (approved):
Why not, although I'd probably format it more as:
```
(infr id: {id_}, n infractions total)
```

scragly (approved):

PR #1242, opened 2020-10-17
Whitelist Core Dev Sprinters from our filters

SebastiaanZ:
To make sure we're not interrupting the Core Devs when they're having their Core Dev Sprint, I've added the `Sprinters` role to our constants and added them to our filter whitelist. To make this easier, I've now made it so all filter features that use a role whitelist use the same role whitelist constant. That makes maintenance a bit easier.

In the process, I've also added a few line breaks in `config-default.yml` to kaizen it up: Now all groups have two line breaks between them.

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -456,6 +456,7 @@ class Roles(metaclass=YAMLGetter):
     owners: int
     partners: int
     python_community: int
+    sprinters: int
     team_leaders: int
     unverified: int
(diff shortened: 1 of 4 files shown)


aeros (approved):
Thanks, looks good to me. Having the sprinters on the whitelist will definitely be convenient for file sharing and command (or chat) cooldowns.

Akarys42 (approved):
Looks good to me

Issue #1127, opened 2020-08-25
mod-alerts channel doesn't display infractions total


mbaruh:
When issuing an infraction in `#mod-alerts`, the bot message doesn't display the infractions total - contrary to how it works in `#mods`. Looking at the behavior of the `!user` command on the channel, it does seem to be considered a mod channel, so it's most likely a bug.

mbaruh:
This ties into #1125 as the same should be there after the category is whitelisted

MarkKoz:
It's because it uses the staff channels constant rather than the mod channels constant.

PR #1243, opened 2020-10-17
Centralized handling of mod channel check

### LINKED ISSUE #1175: Modmail threads don't show total infractions
Related to issues:
https://github.com/python-discord/bot/issues/1127
https://github.com/python-discord/bot/issues/1125

In Modmail threads, notes are listed as necessary, but when issuing an infraction the total number of infractions isn't shown.

mbaruh:
Closes #1175 

The purpose of this PR is to provide one location in the code which is responsible for checking whether a given channel is a mod channel.

Currently this check is performed in the following locations:
* `!user` - If it's a mod channel, it will also display notes and nominations.
* The moderation management cog - only works in mod channels.
* When issuing infractions - mod channels will display infraction number, and total number of infractions for the user.

Issues begin when the definition of what's considered a mod channel changes. The most recent occurrence is the introduction of Modmail, which led to issue https://github.com/python-discord/bot/issues/1125.
However, because those checks are scattered, after that issue was resolved we were still left with https://github.com/python-discord/bot/issues/1175 - which is closed by this PR.

On top of that, despite `#mod-alerts` being whitelisted as a mod channel for quite some time, it didn't display infractions total as that is a separate check, leading to https://github.com/python-discord/bot/issues/1127.

Additionally, I changed all log channels to be considered mod channels, as there were instances where we wanted to run an infraction search in one of the log channels that wasn't `#mod-alerts` but couldn't.

In the future, if we wish to change the definition of what a mod channel is, we can just change the function `is_mod_channel` I'm adding here, and all related commands will work accordingly.

### DIFF
--- a/bot/exts/moderation/infraction/management.py
+++ b/bot/exts/moderation/infraction/management.py
@@ -15,7 +15,7 @@ from bot.exts.moderation.infraction.infractions import Infractions
 from bot.exts.moderation.modlog import ModLog
 from bot.pagination import LinePaginator
 from bot.utils import messages, time
-from bot.utils.checks import in_whitelist_check
+from bot.utils.channel import is_mod_channel
 
 log = logging.getLogger(__name__)
 
@@ -295,13 +295,7 @@ class ModManagement(commands.Cog):
         """Only allow moderators inside moderator channels to invoke the commands in this cog."""
         checks = [
             await commands.has_any_role(*constants.MODERATION_ROLES).predicate(ctx),
-            in_whitelist_check(
-                ctx,
-                channels=constants.MODERATION_CHANNELS,
-                categories=[constants.Categories.modmail],
-                redirect=None,
-                fail_silently=True,
-            )
+            is_mod_channel(ctx.channel)
         ]
         return all(checks)
 
--- a/bot/utils/channel.py
+++ b/bot/utils/channel.py
@@ -2,7 +2,7 @@ import logging
 
 import discord
 
-from bot.constants import Categories
+from bot.constants import Categories, MODERATION_CHANNELS
 
 log = logging.getLogger(__name__)
 
@@ -13,6 +13,14 @@ def is_help_channel(channel: discord.TextChannel) -> bool:
     categories = (Categories.help_available, Categories.help_in_use)
 
     return any(is_in_category(channel, category) for category in categories)
+
+
+def is_mod_channel(channel: discord.TextChannel) -> bool:
+    """Return True if `channel` is one of the moderation channels or in one of the moderation categories."""
+    log.trace(f"Checking if #{channel} is a mod channel.")
+    categories = (Categories.modmail, Categories.logs)
+
+    return channel.id in MODERATION_CHANNELS or any(is_in_category(channel, category) for category in categories)
 
 
 def is_in_category(channel: discord.TextChannel, category_id: int) -> bool:
--- a/bot/utils/channel.py
+++ b/bot/utils/channel.py
@@ -2,7 +2,8 @@ import logging
 
 import discord
 
-from bot.constants import Categories, MODERATION_CHANNELS
+from bot import constants
+from bot.constants import Categories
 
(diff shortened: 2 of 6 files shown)


bot/utils/channel.py
```diff
@@ -15,6 +16,15 @@ def is_help_channel(channel: discord.TextChannel) -> bool:
     return any(is_in_category(channel, category) for category in categories)
 
 
+def is_mod_channel(channel: discord.TextChannel) -> bool:
+    """Return True if `channel` is one of the moderation channels or in one of the moderation categories."""
+    log.trace(f"Checking if #{channel} is a mod channel.")
+    categories = (Categories.modmail, Categories.logs)
+
+    return (channel.id in constants.MODERATION_CHANNELS
+            or any(is_in_category(channel, category) for category in categories))
```
kwzrd, line 25: Prevailing style for a multi-line expression would be:
```python
return (
    a
    or b
)
```

How would you feel about handling each case separately and logging the decision made? For example:
```python
def is_mod_channel(channel: discord.TextChannel) -> bool:
    """True if `channel` is considered a mod channel."""
    if channel.id in constants.MODERATION_CHANNELS:
        log.trace(f"Channel #{channel} is a configured mod channel")
        return True

    mod_categories = Categories.modmail, Categories.logs
    if any(is_in_category(channel, category) for category in mod_categories):
        log.trace(f"Channel #{channel} is in a configured mod category")
        return True

    log.trace(f"Channel #{channel} is not a mod channel")
    return False
```

Looking at one of the callers, the infraction scheduler, it branches based on the return value just so that it can log the decision that was made. One of the clauses could be removed if we just log here. But it's just a thought, I'm not demanding this.

Finally, I wonder if the mod categories should be defined in the config like the mod channels, and then accessed here as e.g. `constants.MODERATION_CATEGORIES`. I don't necessarily think that's the best way to do it, but it'd be more consistent with the way the channel IDs are handled.

mbaruh: Ok, I'll fix the expression.
Yeah, I think adding the constant is a good idea, it will indicate why those categories are needed in the config file.

As for logging, I think it largely depends on the level of granularity the core developers want for it. Handling it outside of the function allows you to add relevant context. I'm fine with whatever you think is better.

bot/exts/moderation/infraction/management.py
```diff
@@ -295,13 +295,7 @@ async def cog_check(self, ctx: Context) -> bool:
...
-            in_whitelist_check(
-                ctx,
-                channels=constants.MODERATION_CHANNELS,
-                categories=[constants.Categories.modmail],
-                redirect=None,
-                fail_silently=True,
-            )
+            is_mod_channel(ctx.channel)
         ]
         return all(checks)
```
kwzrd, line 300: Not part of this PR's diff, but I feel like we could make the first predicate simpler. Checking the intersection of `ctx.author`'s role ids and the constants should work, unless I'm missing something.

kwzrd (approved):
Thanks, I think it's better this way. Logging the exact decision helps a lot during testing. The caller can always log more context ~ both before, and after the check ~ if necessary.

The code is functional and I agree that this is an improvement. Please double-check the configured category IDs. I cannot help verify them as I don't see those channels.

It's good to include the phrase "Closes #1175" ([or a variation thereof](https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references)) in the PR description, it will auto-close the issue once we merge this PR.

bot/exts/moderation/infraction/management.py
mbaruh: I checked the categories, the ID's are correct.
Added the "closes" part

> I feel like we could make the first predicate simpler

Unless there's a matter of efficiency, I think the current line makes it pretty clear what it does.

Numerlor (approved):

lemonsaurus (approved):
This looks good, and is a nice solution. I also really like how thorough the PR description is. Nice work!

PR #1245, opened 2020-10-17
Store the bot instance as a global variable

MarkKoz:
Is **very** convenient when writing utility functions that rely on the bot's state, but aren't in cogs and therefore lack the typical way to access the instance. No more passing around of the instance as an arg! I've gone around the code base and removed any now-unnecessary bot arguments. The changes are nicely in separate commits, so they should be easy to follow.


kwzrd (commented):
I think I prefer SeasonalBot's approach with creating the instance in `bot.bot` and only importing it into the `__main__` to call `run`. Having to do `from bot.bot import bot` is a little eh, but I'd probably prefer it to what's happening in the `__init__` with this approach.

Numerlor:
This could be just be and a consequence of this being new, but when going over it I thought of `bot` as a regular object that held the instance, which made it a bit confusing. I think something like the name `bot_instance` that's imported using a from import could be clearer.

Now there are two methods of accessing the bot: the global instance, and the argument that gets passed to all cogs. Should this be unified to use the global instance? It'd be more consistent within this codebase, but I'm not sure if it's clearer (and would expand the scope of this pr by quite a bit)

MarkKoz:
> Having to do from bot.bot import bot is a little eh,

In such case, the name could be `bot_instance`, which would ease confusion.

> but I'd probably prefer it to what's happening in the __init__ with this approach.

The redis session and intents code would have to be moved over too. I don't like the idea of substantial code execution upon import, even if that may be a more superficial contention in this case. `__main__.py` does _slightly_ feel like a better place for this code, but a good argument could be made for `bot.py` as well.

> I think something like the name bot_instance that's imported using a from import could be clearer.

Perhaps marginally; I don't find it confusing. Why would it even matter to the read what is storing the instance? Whether it's a module or object, it's all the same as far as the reader should be concerned.

> Now there are two methods of accessing the bot: the global instance, and the argument that gets passed to all cogs. Should this be unified to use the global instance

Maybe in the long run, but I don't mind it being an attribute in cogs for now. Right now, I didn't want to deal with large diffs, fixing a bunch of unit tests, and vastly expanding the amount of features I need to test.

kwzrd:
> In such case, the name could be bot_instance, which would ease confusion.

I'd prefer that name a lot.

> I don't like the idea of substantial code execution upon import, even if that may be a more superficial contention in this case.

I think it's fine given the context of our project. It's code that has to execute at init time either way, before the bot connects. The difference is just the module in which it happens. However now that I think about it, moving it to `bot.bot` would cause the instantiation to happen before Sentry is setup (I'm assuming this means that if it fails, we won't get a notif). On the other hand, the Sentry setup could probably be in the `bot.__init__`, which would circumvent this issue, as long as it doesn't import from `bot.bot`.

One more alternative would be having the instantiation where it is, but patching it into `bot.bot` rather than `bot`. That would at least help avoid the circular import caused by the type hint.

MarkKoz:
I've decided to do some refactoring with `__init__.py` and `__main__.py` before trying out moving the instance variable to `bot.py`. But then I realised I somewhat strongly prefer the explicitness of assigning the instance to the variable in `__main__.py` rather than having it be done implicitly upon importing `bot.py`. This left the following arguments:

1. The circular import circumvention for the type annotation is ugly
    * Sure, but I don't think it's so bad; I don't mind. I think it's a decent trade-off for having shorter imports. That will benefit many places, where as removing moderately ugly code in `__init__.py` only benefits that one file which is rarely even worked on.
2. `bot.instance` is confusing; `bot_instance` is better.
    * It could be renamed to that and then the imports could be `from bot import bot_instance`. This doesn't require moving the attribute to `bot.py`. In any case, I don't feel it's worth it since, as I wrote previously, I don't find it confusing. At least, not enough to warrant longer imports. Therefore, it feels redundant to qualify it with a `bot_` prefix when it already has one (the module's name). That being said, I'm not going to fight to the death about a longer import, since it'd still be relatively short and readability is ultimately more important.

kosayoda (approved):
1. I'm fine with the circular import circumvention; I think it's reasonable.

2. I personally prefer `bot_instance`, but either works actually.

lemonsaurus (approved):
No objections from me. This looks good and is a significant improvement over passing around the bot everywhere. Nice work.

PR #1246, opened 2020-10-18
Voice gate dm failed

scragly:
This PR makes two adjustments to the voice gating system:

### In-Channel response fallback.
At the moment, the bot will attempt to DM the verification result for a member which is reliant on privacy settings allowing member DMs. This commit should add a suitable fallback of sending the response in the voice-verification channel instead. As the channel will delete all bot messages automatically after 10 seconds, there was no need to add a delete_after argument to the message sent in-channel.

### Force permission changes to be effective immediately after a voiceban.
A user's permissions will change to not allow speaking when voicebanned, however this permission isn't effective until rejoining. To ensure a voiceban is immediately in effect, the user will be disconnected from any voice channels.


jb3 (approved):
LGTM

PR #1247, opened 2020-10-18
Ensure verified users can see verified message.

scragly:
When verified users get their role, they cannot see the voice-verification channel anymore, so I've added a 3 second delay for granting the role in order to ensure there's some time for them to see the response. 

I've also moved the DM message to only be sent if the DM message succeeds, and to not mention them in-channel to avoid distracting them from the DM notification unnecessarily, as I'm sure they'll see a near-instant response to their command usage in that channel.

I have also been requested to make a grammatical fix as well as to let users know to reconnect to voice to have it apply the change in permissions after verification.

### DIFF
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -22,7 +22,7 @@ FAILED_MESSAGE = (
 )
 
 MESSAGE_FIELD_MAP = {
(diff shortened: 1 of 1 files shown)


jb3 (approved):

gustavwilliam (approved):
LGTM

PR #1248, opened 2020-10-18
Set async-rediscache logging level to warning

Numerlor:
async-rediscache does quite a few debug logs per operation, which drown out the bot logs around them without bringing information of much relevance when developing on the bot, pushing the logging level to only show warnings should keep the logs from the module only to what may matter.

Currently the module only does a couple of error logs and one critical log in places where an exception is raised anyway, but this will make sure that warnings still go through in case they are added in the future.

### DIFF
--- a/bot/__init__.py
+++ b/bot/__init__.py
@@ -65,7 +65,6 @@ logging.getLogger("discord").setLevel(logging.WARNING)
 logging.getLogger("websockets").setLevel(logging.WARNING)
 logging.getLogger("chardet").setLevel(logging.WARNING)
 logging.getLogger("async_rediscache").setLevel(logging.WARNING)
-logging.getLogger(__name__)
 
 
 # On Windows, the selector event loop is required for aiodns.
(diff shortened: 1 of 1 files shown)


bot/__init__.py
```diff
@@ -64,6 +64,7 @@ def monkeypatch_trace(self: logging.Logger, msg: str, *args, **kwargs) -> None:
 logging.getLogger("discord").setLevel(logging.WARNING)
 logging.getLogger("websockets").setLevel(logging.WARNING)
 logging.getLogger("chardet").setLevel(logging.WARNING)
+logging.getLogger("async_rediscache").setLevel(logging.WARNING)
 logging.getLogger(__name__)
```
Numerlor, line 68: Does this call have any use? Didn't want to remove it in case I missed something

SebastiaanZ: Good question. I doubt it's useful, but it may actually do something (as in, create a logger instance there in the logging hierarchy instead of a placeholder object). Not sure why it's there, honestly.

kwzrd: I think it can be removed. The line was added in ebead84ae0eb62ab64b329dfd7fff61ff5de628d and left in this state after e8ad8bb3cf735e9fef09413b8b118b2eff601797.

MarkKoz (approved):

kwzrd (approved):

PR #1249, opened 2020-10-19
Change default API URL to internal docker network.

scragly:
Currently, we're experiencing some issues requesting API data via the public domain, likely caused by some Cloudflare weirdness.

This change sets the internal domain `api.site`, referenceable through the internal Docker DNS only, as the default API domain in config.

This should be **not** merged until https://github.com/python-discord/site/pull/413 and https://github.com/python-discord/salt/pull/21 are both merged and tested manually to resolve correctly.

### DIFF
--- a/bot/api.py
+++ b/bot/api.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 from typing import Optional
 from urllib.parse import quote as quote_url
+from urllib.parse import urlparse
 
 import aiohttp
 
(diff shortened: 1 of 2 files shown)


fadedmax (approved):
LGTM

MarkKoz (approved):

bot/api.py
```diff
@@ -58,7 +59,7 @@ def __init__(self, loop: asyncio.AbstractEventLoop, **kwargs):
 
     @staticmethod
     def _url_for(endpoint: str) -> str:
-        return f"{URLs.site_schema}{URLs.site_api}/{quote_url(endpoint)}"
+        return urlparse(f"{URLs.site_api}/{quote_url(endpoint)}", "http").geturl()
```
MarkKoz, line 62: Not sure why this results in three forward slashes but at least browsers don't seem to care...

jb3:
All our connection issues appear to have halted (win! :tada:) so I'm not sure we need this imminently.

We can revisit internal connections after the Kubernetes mirgation.

Issue #654, opened 2019-11-08
Refactor and improve `information` cog


SebastiaanZ:
In the past couple of months, a number of enhancements have been suggested for the commands in the `information` cog. I decided to make a list of those suggestions (from memory, so please comment the things I've inevitably forgotten) and want to take a stab at implementing those enhancements. In addition, there's also a small bug in the `!roles` command that needs to be fixed, since it's unusable right now.

#### Enhancements

1. If the `!user` command targets a non-moderator in a moderation channel, make the infraction history directly accessible using a reaction. Currently, we almost always combine `!user <user>` with `!infr search <user>` and this change would make the infraction history directly accessible from the `!user` embed while taking up less vertical space.

2. If the `!user` command targets a moderator in a moderation channel, instead of showing a breakdown of their infraction history, show a breakdown of the infractions they've issued. This is mostly for a bit of fun, but I want to make those "actor" infractions accessible with a reaction just like in point 1. That way, you can read back the infractions you've issued recently more easily if you don't remember the exact details.

3. ~~Create a single entry point for the `!roles` command. Currently, there are two commands, `!role` and `!roles`, that look almost identical but do widely different things: One will list all the current roles we have (`!roles`, currently broken) and one will provide a detailed embed for specific roles provided as arguments to the command (`!role <role> [<role> ...]`).~~

    ~~My plan is to create one command, `!role` (with alias `!roles` to keep the API the same), that lists all the roles if no roles were provided as arguments and returns detailed views of roles if roles were provided as arguments. This is fully backwards compatible with the old commands, but you don't have to remember the difference between `!role` and `!roles`)~~ See #532 

4. Add additional information to the `!server` command when issued in a moderation channel. We often look up information like how many members are stuck in checkpoint or how many users are currently being watched. I want to add that information to the `!server` command by default if it was issued in a moderation channel.

5. ~~Make the `!user` command work for users that are currently not a member of the guild. I've noticed that we often try to request the `!user` command on users that are currently not a member of our guild (e.g., when a ban expires). Given that it now gives a nice overview of their infraction history and is going to allow quick access to their infractions (see point 1), it would be useful to make it work for non-members as well. Obviously, we will not have member-specific data, but we can still get general information from the Discord API and historical infractions from our API.~~ See #1250 

#### Bug fixes

[...]

PR #1250, opened 2020-10-21
Add handling of off-server users to user command

mbaruh:
This addresses point 5 in https://github.com/python-discord/bot/issues/654

How it looks like currently:
image

How it will look like:
image

I wonder if it stands out enough. A moderator might miss the fact that they're not on the server.

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -6,7 +6,7 @@ from collections import Counter, defaultdict
 from string import Template
 from typing import Any, Mapping, Optional, Tuple, Union
 
-from discord import ChannelType, Colour, Embed, Guild, Member, Message, Role, Status, utils
+from discord import ChannelType, Colour, Embed, Guild, Member, Message, Role, Status, User, utils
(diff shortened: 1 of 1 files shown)


mbaruh:
Just to make sure, this doesn't close the issue. I will be working on point 1 after a future PR from @scragly is merged

MarkKoz:
This won't be too effective since the user convert only looks at the cache. Chances are that if the user is not in the guild, they won't be in the cache. Maybe briefly after they get banned or leave, but this won't be the case once the bot restarts. There is a custom `FetchedUser` converter which addresses that problem.

mbaruh:
I used it on a user that was never on the server and it seemed to work without a problem

MarkKoz:
After checking out and running it locally, I can confirm it does not work:

bild

mbaruh:
Thanks, should be fixed

MarkKoz (approved):

Numerlor (approved):

PR #1253, opened 2020-10-22
Added voice_chat to eval list

MrHemlock:

### DIFF
--- a/config-default.yml
+++ b/config-default.yml
@@ -195,7 +195,7 @@ guild:
         # Voice
         code_help_voice:                    755154969761677312
         code_help_voice_2:                  766330079135268884
-        voice-chat:                         412357430186344448
+        voice_chat:                         412357430186344448
         admins_voice:       &ADMINS_VOICE   500734494840717332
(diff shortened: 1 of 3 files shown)


config-default.yml
```diff
@@ -195,6 +195,7 @@ guild:
         # Voice
         code_help_voice:                    755154969761677312
         code_help_voice_2:                  766330079135268884
+        voice-chat:                         412357430186344448
```
kwzrd, line 198:
```suggestion
        voice_chat:                         412357430186344448
```

kwzrd (changes requested):

Akarys42:
I think we should allow eval in the whole `VOICE` category instead of manually allowing every channel.

MrHemlock:
> I think we should allow eval in the whole `VOICE` category instead of manually allowing every channel.

Fair point, changing now.

jerbob (approved):
LGTM.

kwzrd (approved):

PR #1257, opened 2020-10-26
Shorter `!code` message

swfarnsworth:
Here's a significantly shorter version of the `!code` message, arrived at by committee in the #organization channel. The goal is to just hit the key point, namely how to format Python code in the here and now, and then get back to the ongoing discussion or help session.

Lemon suggested that we write our own page on where the backtick key is located and we can switch the link out should that ever be completed.

### DIFF
--- a/bot/resources/tags/codeblock.md
+++ b/bot/resources/tags/codeblock.md
@@ -4,4 +4,4 @@ Here's how to format Python code on Discord:
 print('Hello world!')
 \```
 
-**These are backticks, not quotes.** Backticks can usually be found to the left of the `1` key.  
+**These are backticks, not quotes.** See [here](https://superuser.com/questions/254076/how-do-i-type-the-tick-and-backtick-characters-on-windows/254077#254077) if you can't find the backtick key.
--- a/bot/resources/tags/codeblock.md
+++ b/bot/resources/tags/codeblock.md
@@ -4,4 +4,4 @@ Here's how to format Python code on Discord:
 print('Hello world!')
 \```
 
-**These are backticks, not quotes.** See [here](https://superuser.com/questions/254076/how-do-i-type-the-tick-and-backtick-characters-on-windows/254077#254077) if you can't find the backtick key.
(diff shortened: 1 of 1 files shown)


gustavwilliam (approved):
Looks great! It addresses the important things nicely.

bot/resources/tags/codeblock.md
```diff
@@ -1,17 +1,7 @@
...
-Note:  
-• **These are backticks, not quotes.** Backticks can usually be found on the tilde key.  
-• You can also use py as the language instead of python  
-• The language must be on the first line next to the backticks with **no** space between them  
-
-This will result in the following:
-```py
-print('Hello world!')
-```
+**These are backticks, not quotes.** See [here](https://superuser.com/questions/254076/how-do-i-type-the-tick-and-backtick-characters-on-windows/254077#254077) if you can't find the backtick key.
```
MarkKoz, line 7: "See here" sounds awkward to me.

```suggestion
**These are backticks, not quotes.** See [this](https://superuser.com/questions/254076/how-do-i-type-the-tick-and-backtick-characters-on-windows/254077#254077) if you can't find the backtick key.
```

MarkKoz (changes requested):

bot/resources/tags/codeblock.md
swfarnsworth: "See here" sounds more natural to me than "See this" but I don't feel strongly about either.

gustavwilliam: I also think "see here" sounds good. As a third option, something like "See this article" could also be used.

swfarnsworth: "check out" would be a fine alternative to "see", but only if it's followed by "this". Though "see here" is fewer words overall and is still my preferred option. We could also do "read this", but some people might find that slightly abrasive.

gustavwilliam: After thinking about this for a little while now, I’m actually starting to lean toward “check this out.” It sounds quite natural to me and the added word won’t really make much of a difference in length.

swfarnsworth: I'm just trying to figure out how to change it to "check this out" within the GitHub UI--that suggestion works for me.

MarkKoz (approved):

PR #1259, opened 2020-10-27
Added infraction search menu from user embed

mbaruh:
Addresses point 1 in #654. I planned to wait with this one, but we use this combination of commands so often I decided it's worth adding it even if gets partially rewritten by a future PR.

As explained in the issue, we use `!user` and `!infr search` so often we might as well have a reaction leading from one to another.

### Implementation
* The main issue here was to make the information on whether a user has infractions accessible from the `user_info` function. 
   - I wasn't a fan of propagating a boolean value through the helper functions just to that end. 
   - I attempted to rework how the command is divided into its helper function, for example by dividing the part fetching information from the db, and the part that is creating the embed, but the end result was awkward. 
   - Eventually I ended up with accessing the embed fields and checking if the appropriate field has the "No infractions" string. I don't like looking for information I already had access to in one of the inner functions, and it's also volatile to format changes, so I'm open to suggestions.
* ~~I wrote the menu function with extensibility in mind, so that it'll be easy to add more reactions if we want.~~
* I didn't address the part of the issue that says it shouldn't add the reaction for user embeds of moderators. We do make infractions searches on moderators sometimes for whatever reasons, so I don't see the harm in having it there as well.

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -1,12 +1,14 @@
+import asyncio
 import colorsys
 import logging
 import pprint
 import textwrap
 from collections import Counter, defaultdict
+from contextlib import suppress
 from string import Template
 from typing import Any, Mapping, Optional, Tuple, Union
 
-from discord import ChannelType, Colour, Embed, Guild, Message, Role, Status, utils
+from discord import ChannelType, Colour, Embed, Guild, Member, Message, NotFound, Reaction, Role, Status, utils
 from discord.abc import GuildChannel
 from discord.ext.commands import BucketType, Cog, Context, Paginator, command, group, has_any_role
 
@@ -27,6 +29,8 @@ STATUS_EMOTES = {
     Status.dnd: constants.Emojis.status_dnd,
     Status.idle: constants.Emojis.status_idle
 }
+
+INFR_EMOJI = "\u26a0\ufe0f"  # :warning:
 
 
 class Information(Cog):
@@ -206,7 +210,8 @@ class Information(Cog):
         # Will redirect to #bot-commands if it fails.
         if in_whitelist_check(ctx, roles=constants.STAFF_ROLES):
             embed = await self.create_user_embed(ctx, user)
-            await ctx.send(embed=embed)
+            message = await ctx.send(embed=embed)
+            await self.user_info_menu(ctx, message, user)
 
     async def create_user_embed(self, ctx: Context, user: FetchedMember) -> Embed:
         """Creates an embed containing information on the `user`."""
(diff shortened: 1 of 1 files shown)


bot/exts/info/information.py
```diff
@@ -354,6 +359,52 @@ async def user_nomination_counts(self, user: FetchedMember) -> Tuple[str, str]:
...
+                await message.remove_reaction(reaction.emoji, reacter)
+
+                if reaction.emoji == INFR_EMOJI:
+                    search = ctx.bot.get_command("infraction search")
+                    if search:
+                        asyncio.create_task(search(ctx, user.id))
+
+            log.debug("Ending user menu and clearing reactions.")
+            with suppress(NotFound):
+                await message.clear_reactions()
```
jb3, line 406: This seems pretty overcomplicated and could be simplified, there are no other features that have been discussed which would benefit from this.

I think this could be simplified to drop the while loop and instead do something along the lines of:
```py
if is_mod_channel():
    try:
        # wait for reaction
    except asyncio.TimeoutError:
        # No reaction added in time, log
    finally:
        # Remove all reactions and perform infraction search
```

jb3 (changes requested):

bot/exts/info/information.py
mbaruh: Well, I don't want to search in the finally clause because I don't want it pop in when the menu times out, but I'll add it in an else clause.

jb3:
@lemonsaurus has made some points on whether this PR is necessary, so we'll await his views here on that.

SebastiaanZ:
I think this feature is very useful, as we often do use the combo of "user -> infr search". However, my main issue with that combo is the sheer space it takes up. My personal preference would be to have the infraction search embed *replace* the user info embed with the infraction search embed.

That would make the "infraction embed" a menu option of the information embed, instead of an additional embed.

mbaruh:
@SebastiaanZ that would be a fairly simple change, the issue is that information from the user embed can still be helpful even after pulling up the user's infractions to see the whole picture, and this behavior might cause mods to avoid using the feature.

Keep in mind that @scragly is currently working on making the infractions embed a lot more compact, so this issue won't be that severe in the long term. In terms of vertical space this feature already makes a slight improvement.

As I see it, the main gain from this feature is that we won't have to type in the search command every time, and by extension making it more convenient to use on phone.

SebastiaanZ:
I think that the value of user information command mostly goes away after the initial "who is this user" turns into "what are the details of that infractions they have listed in the user embed". If we're not replacing it, I don't think we really need the emoji reaction.

mbaruh:
I don't see it, but if that's what the majority thinks then alright. In fact, I think it's the opposite - if we only want such a feature if it replaces the user embed, then there is no point in having it, as I think we will often find ourselves bringing up the user embed afterwards.

We could make it a paginator where the first page is the user embed and the other pages is the list of infractions, but I'm not sure if that's a good feature.

mbaruh:
This feature will be part of a bigger overhaul of our infraction management system.

PR #1262, opened 2020-10-30
Defcon overhaul

mbaruh:
## Changes
- [x] Ditching the on and off system
  - On an off commands are gone. To turn off auto-kick set days to 0
- [x] Defcon settings change even if writing to db fails
- [x] Use redis instead of postgres
- [x] Failure to load settings will alert moderators in the defcon channel, instead of admins in a log channel. That is because moderators now have access to defcon commands, and if sync fails then it should just be set again manually.

## Additional Features

- [x] A notifier that will send a message once an hour to remind us to turn defcon off.
- [x] A command to shutdown the server (send permissions)
- [x] Status command displays verification level
- [x] More granular control over account age
- [x] Disable time as an optional parameter

## Usage Details
- Defcon days is removed. Instead, there is `defcon threshold` (can be shortened to `dc t`) which takes a join threshold (not limited to days) and an optional duration until expiry:
image
The command can still take an int for the threshold which will be interpreted as days as has been.

- The status command now displays the threshold, the time until expiry, and the server's verification level:
image

- There are now commands to revoke and return send and reaction permissions to everyone (`shutdown` and `unshutdown`).


bot/exts/moderation/defcon.py
```diff
@@ -252,6 +258,21 @@ async def send_defcon_log(self, action: Action, actor: Member, e: Exception = No
...
+            self.defcon_notifier.start()
+
+        elif self.days.days == 0 and self.defcon_notifier.is_running():
+            log.info("DEFCON notifier stopped.")
+            self.defcon_notifier.cancel()
+
+    @tasks.loop(hours=1)
+    async def defcon_notifier(self) -> None:
+        """Routinely notify moderators that DEFCON is active."""
+        await self.channel.send(f"Defcon is on and is set to {self.days.days} day{ngettext('', 's', self.days.days)}.")
```
Akarys42, line 274: Nice usage of `ngettext`, fancy!

Akarys42 (approved):
Looks good to me! (with a bonus point)

bot/exts/moderation/defcon.py
```diff
@@ -153,6 +156,10 @@ async def _defcon_action(self, ctx: Context, days: int, action: Action) -> None:
                     }
                 }
             )
+
+            self.days = timedelta(days=days)
+            self.update_notifier()
```
kwzrd, line 161: This looks like it'd be more appropriate for an `else` clause rather than being in the try block, but it can probably stay.

kwzrd (commented):
I'm struggling to make sense of the cog's overall implementation a little bit so my review may be off lol, feel free to correct me.

The cog enters a weird state when the extension is reloaded:
image

It leaves a "dangling" task still sending notifs, so there should probably be a `cog_unload` listener which will cancel the task if it's running.

mbaruh:
@kwzrd could you describe the series of operations you made that led to the bug? From the screenshot it seems the status command showed that it's set to 0 days, while the notifier still sees the value of 1, even though they're both referring to the same attribute. I didn't manage to replicate such a scenario when I tested it before.
Either way a cog unloader sounds fine to me, but I'm wondering what caused this.

I agree with the other two proposed changes.

mbaruh:
Ignore the review request, it's a byproduct of updating the pr

mbaruh:
This can be reviewed now

Akarys42 (approved):
Still LGTM

mbaruh:
This is ready for review. Don't bother looking at defcon.py's diff, it's practically a new cog.

bot/exts/moderation/defcon.py
```diff
@@ -28,71 +36,75 @@
...
+        try:
+            settings = await self.redis_cache.to_dict()
+            self.threshold = parse_duration_string(settings["threshold"])
+            self.expiry = datetime.fromisoformat(settings["expiry"]) if settings["expiry"] else None
+        except Exception:
             log.exception("Unable to get DEFCON settings!")
-            await self.bot.get_channel(Channels.dev_log).send(
-                f"<@&{Roles.admins}> **WARNING**: Unable to get DEFCON settings!"
-            )
+            await self.channel.send(f"<@&{Roles.moderators}> **WARNING**: Unable to get DEFCON settings!")
```
Akarys42, line 89: I feel like this should also ping the DevOps since they would be more capable of knowing why we can't connect to Redis, and two of our DevOps members aren't mods.

bot/exts/moderation/defcon.py
```diff
@@ -28,71 +36,75 @@
...
 
-            await self.update_channel_topic()
+        self._update_channel_topic()
 
     @Cog.listener()
     async def on_member_join(self, member: Member) -> None:
-        """If DEFCON is enabled, check newly joining users to see if they meet the account age threshold."""
-        if self.enabled and self.days.days > 0:
+        """Check newly joining users to see if they meet the account age threshold."""
+        if self.threshold != relativedelta(days=0):
```
Akarys42, line 103: Couldn't we use some sort of falsy value such as None instead of creating and comparing to a relative delta on each join?

bot/exts/moderation/defcon.py
```diff
@@ -124,134 +136,143 @@ async def defcon_group(self, ctx: Context) -> None:
...
+                **Threshold:** {humanize_delta(self.threshold)}
+                **Expires in:** {humanize_delta(relativedelta(self.expiry, datetime.utcnow())) if self.expiry else "-"}
+                **Verification level:** {ctx.guild.verification_level.name}
+                """
         )
 
         await ctx.send(embed=embed)
 
-    @defcon_group.command(name='days')
+    @defcon_group.command(aliases=('t',))
```
Akarys42, line 154: Maybe we could add the `days` alias, at least for a little bit.

bot/exts/moderation/defcon.py
```diff
@@ -124,134 +136,143 @@ async def defcon_group(self, ctx: Context) -> None:
...
+            self.defcon_notifier.start()
+
+        elif (self.threshold == relativedelta(days=0) or self.expiry is not None) and self.defcon_notifier.is_running():
+            log.info("DEFCON notifier stopped.")
+            self.defcon_notifier.cancel()
+
+    @tasks.loop(hours=1)
+    async def defcon_notifier(self) -> None:
+        """Routinely notify moderators that DEFCON is active."""
+        await self.channel.send(f"Defcon is on and is set to {humanize_delta(self.threshold)}.")
```
Akarys42, line 268: Won't that run even if defcon is disabled?

Akarys42 (changes requested):
Looks pretty good, just a few things. 

What would you say about adding an error handler to relay any failure in defcon to Discord directly, as it is a quite critical cog?

bot/exts/moderation/defcon.py
```diff
@@ -28,71 +36,75 @@
...
+        Icons.defcon_update, Emojis.defcon_update, Colour.blurple(), "**Threshold:** {threshold}\n\n"
+    )
 
 
 class Defcon(Cog):
     """Time-sensitive server defense mechanisms."""
 
-    days = None  # type: timedelta
-    enabled = False  # type: bool
+    redis_cache = RedisCache()
```
ks129, line 57: Please give a better name to it and add a comment what this store.

mbaruh: Hmm, I used the same name we use in voice_gate, but alright lol

bot/exts/moderation/defcon.py
mbaruh: There isn't a concept of disabled anymore. If you mean when it's 0, then no, it will be disabled by _update_notifier

bot/exts/moderation/defcon.py
mbaruh: Not sure, feels like that'd be hiding the fact it's not limited to days anymore

bot/exts/moderation/defcon.py
mbaruh: Agreed

bot/exts/moderation/defcon.py
Akarys42: Alright, good enough for me.

bot/exts/moderation/defcon.py
Akarys42: Also looking back at it, I think the message should contain the traceback.

bot/exts/moderation/defcon.py
Akarys42: The thing is that if it is called with an integer it will assume that the value is in days. I don"t mind it staying like that tbh.

mbaruh:
@Akarys42 I don't think any error at all justifies a ping. If it fails to DM the user it will also throw an error.

bot/exts/moderation/defcon.py
mbaruh: Done

bot/exts/moderation/defcon.py
mbaruh: Doesn't seem right for the defcon channel. Maybe devops? but then again it might be better to just re-raise the exception

Akarys42: I wouldn't mind having it in `#defcon` to be honest. The issue with re-raising is that it might get lost in Sentry because the issue already exists, although the traceback will show in the logs.

bot/exts/moderation/defcon.py
```diff
@@ -28,71 +35,81 @@
...
 
-        try:
-            response = await self.bot.api_client.get('bot/bot-settings/defcon')
-            data = response['data']
+        log.trace("Syncing settings.")
 
-        except Exception:  # Yikes!
+        try:
+            settings = await self.defcon_settings.to_dict()
+            self.threshold = parse_duration_string(settings["threshold"]) if settings["threshold"] else None
```
jb3, line 87: Can we avoid the startup KeyError on first use by using `.get()` here?
```suggestion
            self.threshold = parse_duration_string(settings["threshold"]) if settings.get("threshold") else None
```

bot/exts/moderation/defcon.py
```diff
@@ -28,71 +35,81 @@
...
+            self.expiry = datetime.fromisoformat(settings["expiry"]) if settings["expiry"] else None
```
jb3, line 88: Probably the same comment as above.

jb3 (changes requested):
Minor thoughts, but otherwise all looks good and works locally.

It would be nice to iron out the startup KeyErrors

bot/exts/moderation/defcon.py
mbaruh: Is it possible that we won't be able to read from redis properly? because if there is, a KeyError will allow us to find out about it.

jb3: I'm not aware of such a situation, if we had connection problems then that'd manifest elsewhere.

mbaruh: Right, but then you are also reminded that your defcon settings aren't synced and that you should bring them back if defcon was active. If it's not desirable I don't mind changing it, but that's pretty much the goal of the try clause.

HassanAbouelela: I think the proper solution here is to have a fallback that this cog falls to, not just let everything crash and burn. If we want to add an alert system, that's fine, but it's complementary to the other solution.

mbaruh: It does have fallback, it just sets to the default 0. You can still use defcon, it's just unsynced

HassanAbouelela: I mean we shouldn't error out just because redis is having trouble. The cog should continue without errors, or with proper error handling, and the report should be handled seperately.

mbaruh: It continues, it just lets you know that it wasn't able to sync

mbaruh: I can suppress any errors while updating actually. But other than that defcon is functional even without a connection.

mbaruh: @jb3 yeah so in short the purpose here is to catch that key error or any other error that might indicate things aren't working well. Do you want to remove that?

jb3: So, I haven't tested with redis locally, but if we start up the bot and don't immediately set a defcon will it just alert the moderators on every startup?

I think a better idea is to alert and then actually take action and set the preferences.

mbaruh: No, it will only misfire when the PR is merged because the cache is new. Once we set the cache once it won't do that anymore.

mbaruh: Practically speaking you run `!dc t 0` once and that's it. The issue is that you need to have the keys in the cache.

jb3 (approved):
:shipit::shipit::shipit::shipit:

HassanAbouelela (approved):
I have no real feedback, everything seems fine. The only change I would ask for is for the method of shutting down defcon to be added to the help command.

HassanAbouelela (approved):
Thanks

PR #1265, opened 2020-10-31
Voice Gate: User activity blocks

jb3:
This PR introduces a new check for the voice gate to ensure that a users activity comes from a larger period of time rather than a burst all at once. This should deter spammers.

The way that the check works is that user activity is chunked up into 10 minute blocks which are then counted.

For example:
```
[16:05] User A: Hello!
[16:06] User A: Hello world!
```

This is one activity block because it all happened within the 16:00-16:10 block.

```
[16:05] User B: Hello!
[16:15] User B: Hello world!
```

This is two activity blocks because they fall into the 16:00-16:10 block and then 16:10-16:20 block.

This will all be described in the voice gate channel to help users understand the system.

The current threshold is more of a stab in the dark than the message threshold which is based on actual data, but looking at users we have voice banned it would certainly deter some of them.

Site-side PR: https://github.com/python-discord/site/pull/418

### DIFF
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -60,9 +60,9 @@ class VoiceGate(Cog):
                 embed = discord.Embed(
                     title="Not found",
                     description=(
-                        "We were unable to find user data for you. ",
-                        "Please try again shortly, ",
-                        "if this problem persists please contact the server staff through Modmail.",
+                        "We were unable to find user data for you. "
+                        "Please try again shortly, "
+                        "if this problem persists please contact the server staff through Modmail."
                     ),
                     color=Colour.red()
                 )
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -51,7 +51,7 @@ class VoiceGate(Cog):
         - You must have over a certain number of messages within the Discord server
         - You must have accepted our rules over a certain number of days ago
         - You must not be actively banned from using our voice channels
-        - You must have been active for over a certain number of 10-minute blocks.
+        - You must have been active for over a certain number of 10-minute blocks
         """
         try:
             data = await self.bot.api_client.get(f"bot/users/{ctx.author.id}/metricity_data")
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -25,6 +25,7 @@ MESSAGE_FIELD_MAP = {
     "verified_at": f"have been verified for less than {GateConf.minimum_days_verified} days",
     "voice_banned": "have an active voice ban infraction",
     "total_messages": f"have sent less than {GateConf.minimum_messages} messages",
+    "activity_blocks": f"have been active for less than {GateConf.minimum_activity_blocks} ten-minute blocks"
 }
 
 
@@ -50,6 +51,7 @@ class VoiceGate(Cog):
         - You must have over a certain number of messages within the Discord server
         - You must have accepted our rules over a certain number of days ago
         - You must not be actively banned from using our voice channels
+        - You must have been active for over a certain number of 10-minute blocks.
         """
         try:
             data = await self.bot.api_client.get(f"bot/users/{ctx.author.id}/metricity_data")
@@ -88,7 +90,8 @@ class VoiceGate(Cog):
         checks = {
             "verified_at": data["verified_at"] > datetime.utcnow() - timedelta(days=GateConf.minimum_days_verified),
             "total_messages": data["total_messages"] < GateConf.minimum_messages,
-            "voice_banned": data["voice_banned"]
+            "voice_banned": data["voice_banned"],
+            "activity_blocks": data["activity_blocks"] < GateConf.activity_blocks
         }
         failed = any(checks.values())
         failed_reasons = [MESSAGE_FIELD_MAP[key] for key, value in checks.items() if value is True]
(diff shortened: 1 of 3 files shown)


bot/exts/moderation/voice_gate.py
```diff
@@ -50,6 +51,7 @@ async def voice_verify(self, ctx: Context, *_) -> None:
         - You must have over a certain number of messages within the Discord server
         - You must have accepted our rules over a certain number of days ago
         - You must not be actively banned from using our voice channels
+        - You must have been active for over a certain number of 10-minute blocks.
```
kwzrd, line 54: The other list items above don't end with a dot.

bot/exts/moderation/voice_gate.py
```diff
@@ -25,6 +25,7 @@
     "verified_at": f"have been verified for less than {GateConf.minimum_days_verified} days",
     "voice_banned": "have an active voice ban infraction",
     "total_messages": f"have sent less than {GateConf.minimum_messages} messages",
+    "activity_blocks": f"have been active for less than {GateConf.minimum_activity_blocks} ten-minute blocks"
```
kwzrd, line 28: Keeping the trailing comma is nice.

config-default.yml
```diff
@@ -521,6 +521,7 @@ voice_gate:
     minimum_days_verified: 3  # How many days the user must have been verified for
     minimum_messages: 50  # How many messages a user must have to be eligible for voice
     bot_message_delete_delay: 10  # Seconds before deleting bot's response in Voice Gate
+    minimum_activity_blocks: 3 # Number of 10 minute blocks during which a user must have been active
```
kwzrd, line 524: Although it's not Python code the rest of the config file still uses two spaces before the hash.

kwzrd (commented):
Only style nitpicks sorry. :disappointed: 

Is this ready on the side of Metricity?

bot/exts/moderation/voice_gate.py
kwzrd: Should this be "fewer than"? I don't know lol.

bot/exts/moderation/voice_gate.py
```diff
@@ -88,7 +90,8 @@ async def voice_verify(self, ctx: Context, *_) -> None:
         checks = {
             "verified_at": data["verified_at"] > datetime.utcnow() - timedelta(days=GateConf.minimum_days_verified),
             "total_messages": data["total_messages"] < GateConf.minimum_messages,
-            "voice_banned": data["voice_banned"]
+            "voice_banned": data["voice_banned"],
+            "activity_blocks": data["activity_blocks"] < GateConf.activity_blocks
```
kwzrd, line 94: The attr is named `minimum_activity_blocks` not `activity_blocks`, sorry I missed this the first time.

kwzrd (changes requested):

jb3:
The Metricity PR on site has been merged, so the data is there if you run site locally. There should be some stuff bootstrapped in the PostgreSQL init script with Docker Compose which means that the metricity_data endpoint works.

kwzrd (changes requested):
Okay, there's a bug in the initial implementation. When site returns a 404, the code constructs an error embed for the user, and accidentally passes a tuple into the description field (see the comma at the end of `L63`).

https://github.com/python-discord/bot/blob/c8c58b7283d201a3efb0524b11591b6ed7e4f3c0/bot/exts/moderation/voice_gate.py#L57-L66

This embed then fails to send:

image

Let's remove the culprit comma as part of this PR please.

---

Now for this feature, the code looks fine and I tested it locally, though there were a few steps I had to take that I will explain here so that my methodology can be vibe checked.

Initially, I was running into the issue that the request to site was failing because the 'metricity' database did not exist. @jb3 advised me to exec into my postgres container and run the [boostrap query](https://github.com/python-discord/site/blob/master/postgres/init.sql) to create the database. This worked and solved the issue, but when we run the [total message blocks query](https://github.com/python-discord/site/blob/14d3a7e3cc1b6cc1f25d7d790b42042448383982/pydis_site/apps/api/models/bot/metricity.py#L49), it fails because the bootstrap does not create the `created_at` column:

```sql
...
SELECT
    (floor((extract('epoch' from created_at) / %s )) * %s) AS interval
FROM messages
...
```

I added this column to the create query, added myself to `users`, and inserted a message so my table now looked like this:

```
metricity=# SELECT * FROM messages;
 id  |     author_id      | is_deleted |         created_at         
-----+--------------------+------------+----------------------------
 123 | 144912469071101952 | f          | 2020-11-01 12:39:21.364375
(1 row)
```

This allows the blocks query to execute and everything else works fine. I commented out the other requirements (apart from message blocks) and reduced the required blocks to `2` in my config.

Then, I get denied when I run `!voiceverify`:

image

Adding one row to `messages` with a timestamp pushed into the next block then allows me to pass the gate:

```
metricity=# SELECT * FROM messages;
 id  |     author_id      | is_deleted |         created_at         
-----+--------------------+------------+----------------------------
 123 | 144912469071101952 | f          | 2020-11-01 12:39:21.364375
 124 | 144912469071101952 | f          | 2020-11-01 12:41:21.364375
(2 rows)
```

image

So all in all this seems fine but it may be a good idea to add the `created_at` column to the bootstrap script (although that's in a separate repo). I did not check the site-side implementation much because that was already merged as part of a different PR. I also linked the site PR in your description for the convenience of other reviewers.

This is basically an approval but we should fix the tuple bug explained at the top.

jb3:
Comma has been fixed in 399f286a157b21e860f606b4f4fed11bb29490f2. That column should indeed be added to the bootstrap script.

kwzrd:
@jb3 You added more commas, but the one that was there should have been removed instead. We want to build a string, the trailing comma was making it into a 1-tuple.

kwzrd (approved):
Looks full LGTM to me.

Den4200 (approved):
Looks good to me.

PR #1266, opened 2020-11-01
Updated langs to include python-repl

zachgates:
Amend `codeblock._parsing.PY_LANG_CODES` to include valid `python-repl` lang code.

### DIFF
--- a/bot/exts/info/codeblock/_parsing.py
+++ b/bot/exts/info/codeblock/_parsing.py
@@ -12,7 +12,7 @@ from bot.utils import has_lines
 log = logging.getLogger(__name__)
 
 BACKTICK = "`"
-PY_LANG_CODES = ("python", "python-repl", "pycon", "py")  # Order is important; "py" is last cause it's a subset.
+PY_LANG_CODES = ("python-repl", "python", "pycon", "py")  # Order is important; "py" is last cause it's a subset.
 _TICKS = {
     BACKTICK,
     "'",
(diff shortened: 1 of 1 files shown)


bot/exts/info/codeblock/_parsing.py
```diff
@@ -12,7 +12,7 @@
 log = logging.getLogger(__name__)
 
 BACKTICK = "`"
-PY_LANG_CODES = ("python", "pycon", "py")  # Order is important; "py" is last cause it's a subset.
+PY_LANG_CODES = ("python", "python-repl", "pycon", "py")  # Order is important; "py" is last cause it's a subset.
```
Akarys42, line 15: Seeing the comment at the end of the line, `python-repl` should be before `python`. Have you tested the new regex?

Akarys42 (dismissed):

bot/exts/info/codeblock/_parsing.py
zachgates: I interpreted the bit about "subset" as referring uniquely to `"py"` -- being a subset of the other/prior elements. Reinserting `python-repl` at the beginning of the tuple, however, does not escape the regex, either.

Den4200 (approved):
Looks good to me, thanks!

Numerlor (changes requested):
We should include all aliases from https://github.com/highlightjs/highlight.js/blob/cf63d2fc63c6fa0f7d1ddf99fcdbe8ebf8ee8b93/src/languages/python.js#L249

Senjan21 (approved):
Looks good! Thanks for the PR!

PR #1269, opened 2020-11-02
User Events Management system

### LINKED ISSUE #1172: User Events Management System
I present a design for a full system for managing user events. It will hopefully allow us to accommodate and invite more users to organize their own events, and to properly make use of the new User Events voice channel.

This issue should be implemented alongside the related issue in the site repository https://github.com/python-discord/site/issues/396

## Current State

With the recent changes in the voice channels, we repurposed the `Games` channel to a `User Events` channel which will only be open during a user event. Additionally, we introduced a new channel `#user-event-coordinators`.
Currently, user event coordinators can ask an admin to open the channel for them.

## The New System

We introduce a scheduling system for user events. The system will work as follows:

### Scheduling

 A user who wants to stream some content / organize a small event on the server asks for an admin's permission and is granted the role `User Event Coordinator`, assuming we trust them - nothing different so far.
However, this role only grants the permission to access the `#user-event-coordinators` channel, and to schedule their event.

#### A PR implementing this system will provide a member with the `User Event Coordinators` role with the following commands

`!userevent create <event name>`
* This will allow a coordinator to own a new event, allowing it to be scheduled.
* The event name should be different from any other event created so far.

`!userevent schedule <event name> <start datetime> [duration=3h]` 
* The coordinator will be able to schedule their event by specifying a date and time. The event name should be of an event which already exists (using the `create` command) and is owned by the user invoking the command.
* This will reserve them a time slot so that no two event coordinators accidentally plan an event for the same time slot. When checking whether a time slot is free, it needs to be free from half an hour before the appointed time, until half an hour after the event is scheduled to end.
* A coordinator will be able to only reserve one time slot at a time, for only one of their events.
* The duration of the event is three hours by default, but can be explicitly made to be up to five hours long. It can't be less than half an hour.

`!userevent cancel`
* The coordinator will be able to run the command if need be, and free their timeslot. 
* The event's name is not required as a coordinator can only have up to one event scheduled.

`!userevent delete <event name>`
Deletes the event of that name, assuming it's owned by the user invoking the command, removing it from the DB.

`!userevent sublist <event name>`
Lists the users subscribed to the named event (explained further in the following section).

### Subscription

After an event is created, other users will be able to subscribe to the event, allowing the coordinator to know ahead of time who wants to participate, and to ping them before the event begins (related command detailed in the next section).

#### A PR implementing this system will provide all members of the community with the following commands

`!userevent sub <event name>`
Subscribes the user to the specified event.

`!userevent unsub <event name>`
Unsubscribes the user from the specified event.

`!userevent when <event name>`
Shows the scheduled date and time of the event (assuming it's scheduled).

`!userevent list`
Lists all scheduled userevents and their times.

### The Event

Half an hour before the scheduled time arrives, the bot will ping the coordinator in the `#user-event-coordinators` channel, and will automatically grant the role `User Event: Ongoing`. This is the role which will grant the user streaming permissions.

#### A PR implementing this system will provide a member with the `User Event: Ongoing` role with the following commands

`!userevent open`
Unlocks the `User Events` voice channel.

`!userevent announce [message]`
Pings the users subscribed to the event in the `#user-event-announcements` channel along with the specified message.

`!userevent close`
Locks the voice channel, for example if they don't want any additional people to join or if the event ended ahead of time.

After the scheduled duration, the channel will lock automatically (but not kick them out in case they need a little time to wrap up), and the `User Event: Ongoing` role will be revoked.

___

These commands should be run in either `#user-event-coordinators`, `#user-event-chat` or `#bot-commands`.

Additionally, the relevant roles should be added to the bot's config files.
`User Event Coordinator`: 591999763478609927
`User Event: Ongoing`: 757666374607831221

RohanJnr:
closes #1172 

This PR uses the new API endpoints and features implemented in https://github.com/python-discord/site/pull/410
Hence, this PR should be merged only after the Site PR is merged and deployed.

### Commands/Functionality the Cog provides:
- [x] Create a new user event.
- [x] Delete user event.
- [x] Schedule a user event.
- [x] Cancel scheduled user event.
- [x] Open/Close user events voice channel.
- [x] Announce during event start by tagging all event subscribers.
- [x] Modify user event description.

### How does this work?

- Whenever a new user event is created, it will be listed in a new channel called #user-events-list (name can be flexible). This channel is used to display all the user events and also show their **status** such as:
  - **Scheduled on `DateTime in readable format`**
  - **Not Scheduled**
  - **Live**
image
image

- The event message ID will be stored in the site DB for further usage.
- Users will be able to subscribe to the event by reacting to the :heavy_check_mark: reaction on the message.
- When the user event is scheduled, the status in the event message is edited.
image
image

Also, another message is sent in the #user-event-announcement channel, like the one below:
image

- 30min before the event start time, the organizer is given the *ongoing* role and has access to commands to open/close the user event voice channel and also the announce command.

### Conditions
- 2 Events with the same name cannot exist.
- An organizer cannot have multiple events scheduled.
- No two events can overlap (in terms of time).
- There must be a minimum of 30min gap between 2 scheduled events.

**Note: All the above checks are handled by the site in form of validation checks, the bot just catches the `ResponseCodeError` exception.**

#### Channels used
- #use-events-list(need to be created)
- #user-event-announcements
- #user-event-coordinators(need to be created-not sure if its already present)

#### Roles used
- Developers role(verified role)
- user_event_coordinator
- user_event_ongoing


bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,603 @@
...
+
+    @staticmethod
+    def not_scheduled() -> str:
+        """To indicate user event is not scheduled."""
+        return "Not scheduled"
+
+    @staticmethod
+    def live() -> str:
+        """To indicate user event is live."""
+        return "Live"
```
ks129, line 66: I think it's better to use constants for this.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,603 @@
...
+
+    @property
+    def user_events_list_channel(self) -> TextChannel:
+        """Return #user-events-list channel."""
+        return self.bot.get_channel(Channels.user_event_list)
+
+    @property
+    def user_event_voice_channel(self) -> VoiceChannel:
+        """Return #user-events-voice channel."""
+        return self.bot.get_channel(Channels.user_event_voice)
```
ks129, line 123: I don't know 100% how discord.py caching works, but I think there might be cases when these don't exist in cache.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,603 @@
...
+            )
+        )
+        return embed
+
+    async def fetch_subscribers(self, message_id: int) -> list:
+        """Fetch reacted users to event message as subscribers."""
+        # Fetch the event message
+        message = await self.user_events_list_channel.fetch_message(message_id)
+
+        reaction = message.reactions[0]
```
ks129, line 143: I think we should make sure is this right emoji. I know this isn't something really important, but sometimes something with ratelimits that bot can't add new reaction and somebody else add some other reaction etc.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,603 @@
...
+        # Fetch user event message
+        message = await self.user_events_list_channel.fetch_message(user_event["message_id"])
+
+        embed = self.user_event_embed(
+            user_event["name"],
+            user_event["description"],
+            self.guild.get_member(user_event["organizer"]),
+            status
+        )
+        embed.set_footer(text="React to be notified.")
```
ks129, line 170: I think `React to get notified about event start.` would be better and give more information.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,603 @@
...
+        }
+        await self.bot.api_client.patch(f"bot/user-events/{event_name}", data=patch_data)
+
+        await ctx.send(f"User Event **{event_name}** created.")
+
+    @user_event.command(name="change_desc", aliases=["cd", "desc"])
+    async def change_description(self, ctx: Context, event_name: str, *, event_description: str) -> None:
+        """Change user event description."""
+        # Check if user is event organizer
+        if not self.check_if_user_is_organizer(ctx.author.id, event_name):
```
ks129, line 353: You are not awaiting it and also I think more clean way to do this is do this in decorator.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,603 @@
...
+            await ctx.send("You can only modify your events!")
+            return
+        data = {
```
ks129, line 356: Move this 1 key to call and don't assign it to variable.

ks129 (changes requested):
First batch of reviews

bot/exts/fun/user_events.py
RohanJnr: These commands are rarely used tho, it shouldn't matter much.
If incase its a concern, I guess I could declare all the roles and channels in the `__init__()` method, what do you say?

ks129: I'd say more `self.bot.get_xxxx(yyy) or await self.bot.fetch_xxxx()`, but this mean that you have to make put them to async function. I think yes, you should create them inside task that is created in `__init__` that adds them to variables, but use this `or`.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,607 @@
...
+    1: 'st', 21: 'st', 31: 'st',
+    2: 'nd', 22: 'nd',
+    3: 'rd', 23: 'rd'
+}
+
+NOT_SCHEDULED = "Not scheduled"
+LIVE = "Live"
+
+
+def is_event_organizer(func: Callable) -> Callable:
```
ks129, line 36: I'd implement it as discord.py check decorator: https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html?highlight=checks#checks

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,607 @@
...
+    async def restart_event_reminders(self) -> None:
+        """Restart scheduled event reminders when bot restarts."""
+        await self.bot.wait_until_guild_available()
+        scheduled_events = await self.bot.api_client.get("bot/scheduled-events")
+
+        for event in scheduled_events:
+            # If event is live
+            if datetime.now() > parse(event["start_time"]).replace(tzinfo=None):
+                self.schedule_event_end_reminder(event)
+
```
ks129, line 73: This newline is unnecessary

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,607 @@
...
+    @staticmethod
+    def not_scheduled() -> str:
+        """To indicate user event is not scheduled."""
+        return "Not scheduled"
+
+    @staticmethod
+    def live() -> str:
+        """To indicate user event is live."""
+        return "Live"
```
ks129, line 85: These can be removed now.

bot/exts/fun/user_events.py
ks129: At least when you don't change it, make return type `Optional` and `None` check in code.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,607 @@
...
+
+        # Send message in #user-event-announcements channel
+        subscribers = "".join(sub.mention for sub in subscribers)
+        message = subscribers
+
+        if announcement_message:
+            message += announcement_message
+
+        # Update event status
+        status = LIVE
```
ks129, line 567: Don't assign it to variable.

ks129 (changes requested):
More comments

bot/exts/fun/user_events.py
RohanJnr: I like the idea of loading them inside a task created in __init__ method.

bot/exts/fun/user_events.py
RohanJnr: I don't think you can access command-based arguments if we use the check function, so I will be sticking with the current implementation.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,583 @@
...
+
+        # Load required channels.
+        self.user_event_coord_channel = None
+        self.user_event_announcement_channel = None
+        self.user_events_list_channel = None
+        self.user_event_voice_channel = None
+
+        # Load required roles.
+        self.developers_role = None
+        self.user_event_ongoing_role = None
```
ks129, line 68: This is place where adding type hint to variable is good.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,583 @@
...
+
+        # Load required channels.
+        self.user_event_coord_channel = self.bot.get_channel(Channels.user_event_coordinators)
+        self.user_event_announcement_channel = self.bot.get_channel(Channels.user_event_announcements)
+        self.user_events_list_channel = self.bot.get_channel(Channels.user_event_list)
+        self.user_event_voice_channel = self.bot.get_channel(Channels.user_event_voice)
+
+        # Load required roles.
+        self.developers_role = self.guild.get_role(Roles.verified)
+        self.user_event_ongoing_role = self.guild.get_role(Roles.user_event_ongoing)
```
ks129, line 92: Use here `get or fetch` pattern. When some channel/role is not found, this should unload cog and show warning about it, because everything is required there. Also, use asyncio.Event to wait for items loading.

ks129 (changes requested):

bot/exts/fun/user_events.py
RohanJnr: Discordpy does not provide the fetch method for channels.
I am going for the following pattern:
- using properties like before
- add a custom property decorator that caches the value and unloads the cog if the property value is None.

ks129: discord.py have `fetch_channel` method.

RohanJnr: oh, right, how did I miss that.

bot/exts/fun/user_events.py
```diff
@@ -0,0 +1,583 @@
...
+}
+
+DATE_PREFIX = {
+    1: 'st', 21: 'st', 31: 'st',
+    2: 'nd', 22: 'nd',
+    3: 'rd', 23: 'rd'
+}
+
+NOT_SCHEDULED = "Not scheduled"
+LIVE = "Live"
```
Xithrius, line 33: Why is this a constant, if it's only used once?

ks129: These strings had getter functions before, I asked moving them from functions to constants.

bot/exts/fun/user_events.py
RohanJnr: what if I check if `get_channel` or `get_role` returns None, then I unload the cog with appropriate logging instead of `fetch_channel` raising an error and unloading the cog?

ks129: We must be ready that channels or role(s) is not in cache, so we need to fetch them.

Xithrius:
@RohanJnr are you planning to complete this?

Xithrius:
Current state of this PR: [Organization Issue](<https://github.com/python-discord/organisation/issues/321>).

janine9vn:
I've chatted with @RohanJnr about converting this code for a different but very similar purpose. He'll be using a new branch and PR but has requested this stay open for a bit longer to reference some implementation details.

Xithrius:
@janine9vn Thank you for the update.

Xithrius:
@RohanJnr do you still need this PR to exist?

RohanJnr:
Closing this as a new one will be opened for #1603 .

PR #1271, opened 2020-11-05
```py convention in code block instructions

spacecraft1013:
PR #1257 implemented using \```py rather than \```python in the `!code` message. This PR extends that to the instruction message sent when someone sends incorrectly formatted code.


MarkKoz (approved):
Thanks. I suppose consistency is nice. This will also teach users the shorter, more convenient alias.

Den4200 (approved):
Thanks!

PR #1272, opened 2020-11-05
Voice Gate: notify new unverified users in the verification channel

### LINKED ISSUE #1261: Ping New Voice Chat Users in #voice-verification
Currently as we have the voice-gate features, people are deeply confused as to why they're muted and can't talk in the chat.  I propose we fix that by tracking if a user has joined voice-chat before, and if they haven't, ping them in `#voice-verification`.

The ping message will go away after one of three criteria have been met:
- If 5 minutes have passed then the message will automatically delete
- If the user attempts to verify (successfully or otherwise)
- ~~The user clicks a provided trash can reaction to remove it immediately~~

~~An additional issue and PR would be required in the site repo in order to add another column in order to track if they have visited the voice chat before or not.~~

We'll track if a user has joined the voice channel before through ReDis.  This keeps it all in one place.  We'll likely also need a role check on bot boot to make sure it's up to date and doesn't ping people who have the voice verified role already.

Hopefully this would better inform users about what's going on when they try to talk.

MrHemlock:
This PR's purpose is to ping new users in the voice_verification channel so that they can figure out why they aren't able to talk.  Currently, we have a lot of people confused about the new system. There are multiple people popping in and out of the voice channel thinking that there's something wrong with their connection or thinking they got specifically suppressed.  

By pinging the user in the voice_verification channel (along with a little explanation on how to use it), this should reduce the number of folks asking.

The reason that this is only sending a ping in the channel and not a direct DM is that a fair number of people may have their DMs turned off.  Another reason is that if they are joining a voice channel, they'll already be looking in that section of the channels, and a ping that's right in front of their eyes should get them looking in the right place.

Closes #1261

### DIFF
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -4,8 +4,9 @@ from contextlib import suppress
 from datetime import datetime, timedelta
 
 import discord
+from async_rediscache import RedisCache
 from dateutil import parser
-from discord import Colour, Member, VoiceState
+from discord import Colour, Member
 from discord.ext.commands import Cog, Context, command
 
 from bot.api import ResponseCodeError
@@ -31,8 +32,16 @@ MESSAGE_FIELD_MAP = {
 class VoiceGate(Cog):
     """Voice channels verification management."""
 
-    def __init__(self, bot: Bot):
+    # RedisCache[t.Union[discord.User.id, discord.Member.id], t.Optional[discord.Message.id]]
+    redis_cache = RedisCache()
+
+    def __init__(self, bot: Bot) -> None:
         self.bot = bot
+        self._init_task = self.bot.loop.create_task(self._async_init())
+
+    async def _aysnc_init(self) -> None:
+        await self.bot.wait_until_guild_available()
+        self._voice_verification_channel = self.bot.get_channel(Channels.voice_gate)
 
     @property
     def mod_log(self) -> ModLog:
@@ -158,8 +167,18 @@ class VoiceGate(Cog):
             await message.delete()
 
     @Cog.listener()
-    async def on_voice_state_update(self, member: Member, before: VoiceState, after: VoiceState):
-        pass
+    async def on_voice_state_update(self, member: Member, *_) -> None:
+        """Pings a user if they've never joined the voice chat before and aren't verified"""
+
+        in_cache = await self.redis_cache.get(member.id)
+
+        # member.voice will return None if the user is not in a voice channel
+        if not in_cache and member.voice is not None:
+            log.trace("User not in cache and is in a voice channel")
+            verified = any(Roles.voice_verified == role.id for role in member.roles)
+            if verified:
+                await self.redis_cache.set(member.id, None)
+                return
 
     async def cog_command_error(self, ctx: Context, error: Exception) -> None:
         """Check for & ignore any InWhitelistCheckFailure."""
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -5,7 +5,7 @@ from datetime import datetime, timedelta
 
 import discord
 from dateutil import parser
-from discord import Colour
+from discord import Colour, Member, VoiceState
 from discord.ext.commands import Cog, Context, command
 
 from bot.api import ResponseCodeError
@@ -156,6 +156,10 @@ class VoiceGate(Cog):
 
         with suppress(discord.NotFound):
             await message.delete()
+
+    @Cog.listener()
+    async def on_voice_state_update(self, member: Member, before: VoiceState, after: VoiceState):
+        pass
 
     async def cog_command_error(self, ctx: Context, error: Exception) -> None:
         """Check for & ignore any InWhitelistCheckFailure."""
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -48,8 +48,6 @@ class VoiceGate(Cog):
 
     def __init__(self, bot: Bot) -> None:
         self.bot = bot
-        # voice_verification_channel set to None so that we have a placeholder to get it later in the cog
-        self.voice_verification_channel = None
 
     @property
     def mod_log(self) -> ModLog:
@@ -71,9 +69,9 @@ class VoiceGate(Cog):
         """
         # If user has received a ping in voice_verification, delete the message
         if message_id := await self.redis_cache.get(ctx.author.id, NO_MSG):
+            log.trace(f"Removing voice gate reminder message for user: {ctx.author.id}")
             with suppress(discord.NotFound):
-                self.voice_verification_channel = self.bot.get_channel(Channels.voice_gate)
-                await self.bot.http.delete_message(self.voice_verification_channel, message_id)
+                await self.bot.http.delete_message(Channels.voice_gate, message_id)
             await self.redis_cache.set(ctx.author.id, NO_MSG)
 
         try:
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -199,8 +199,7 @@ class VoiceGate(Cog):
             log.trace("User not in a voice channel. Ignore.")
             return
 
-        in_cache = await self.redis_cache.get(member.id, NO_MSG)
-        if in_cache:
+        if await self.redis_cache.contains(member.id):
             log.trace("User already in cache. Ignore.")
             return
 
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -32,14 +32,14 @@ MESSAGE_FIELD_MAP = {
 class VoiceGate(Cog):
     """Voice channels verification management."""
 
-    # RedisCache[t.Union[discord.User.id, discord.Member.id], t.Optional[discord.Message.id]]
+    # RedisCache[t.Union[discord.User.id, discord.Member.id], t.Union[discord.Message.id, bool]]
     redis_cache = RedisCache()
 
     def __init__(self, bot: Bot) -> None:
         self.bot = bot
         self._init_task = self.bot.loop.create_task(self._async_init())
 
-    async def _aysnc_init(self) -> None:
+    async def _async_init(self) -> None:
         await self.bot.wait_until_guild_available()
         self._voice_verification_channel = self.bot.get_channel(Channels.voice_gate)
 
@@ -62,10 +62,11 @@ class VoiceGate(Cog):
         """
 
         # If user has received a ping in voice_verification, delete the message
-        if message_id := self.redis_cache.get(ctx.author.id, None) is not None:
-            ping_message = await ctx.channel.fetch_message(message_id)
-            await ping_message.delete()
-            await self.redis_cache.update(ctx.author.id, None)
+        if message_id := await self.redis_cache.get(ctx.author.id, None):
+            with suppress(discord.NotFound):
+                ping_message = await ctx.channel.fetch_message(message_id)
+                await ping_message.delete()
+            await self.redis_cache.set(ctx.author.id, False)
 
         try:
             data = await self.bot.api_client.get(f"bot/users/{ctx.author.id}/metricity_data")
@@ -181,19 +182,20 @@ class VoiceGate(Cog):
             log.trace("User is a bot. Ignore.")
             return
 
-        in_cache = await self.redis_cache.get(member.id)
+        in_cache = await self.redis_cache.get(member.id, None)
 
         # member.voice will return None if the user is not in a voice channel
-        if not in_cache and member.voice is not None:
+        if in_cache is not None and member.voice is not None:
             log.trace("User not in cache and is in a voice channel")
             verified = any(Roles.voice_verified == role.id for role in member.roles)
             if verified:
                 log.trace("User is verified, add to the cache and ignore")
-                await self.redis_cache.set(member.id, None)
+                #  redis cache does not accept None, so False is used to signify no message
+                await self.redis_cache.set(member.id, False)
                 return
 
             log.trace("User is unverified. Send ping.")
-            message = self._voice_verification_channel.send(
+            message = await self._voice_verification_channel.send(
                 f"Hello, {member.mention}! Wondering why you can't talk in the voice channels? "
                 "Use the `!voiceverify` command in here to verify. "
                 "If you don't yet qualify, you'll be told why!"
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -29,7 +29,7 @@ MESSAGE_FIELD_MAP = {
 }
 
 VOICE_PING = (
-    "Hello, {}! Wondering why you can't talk in the voice channels? "
+    "Hello, {0}! Wondering why you can't talk in the voice channels? "
     "Use the `!voiceverify` command in here to verify. "
     "If you don't yet qualify, you'll be told why!"
 )
@@ -66,7 +66,6 @@ class VoiceGate(Cog):
         - You must have accepted our rules over a certain number of days ago
         - You must not be actively banned from using our voice channels
         """
-
         # If user has received a ping in voice_verification, delete the message
         if message_id := await self.redis_cache.get(ctx.author.id, None):
             with suppress(discord.NotFound):
@@ -186,7 +185,7 @@ class VoiceGate(Cog):
 
     @Cog.listener()
     async def on_voice_state_update(self, member: Member, *_) -> None:
-        """Pings a user if they've never joined the voice chat before and aren't verified"""
+        """Pings a user if they've never joined the voice chat before and aren't verified."""
         if member.bot:
             log.trace("User is a bot. Ignore.")
             return
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -48,6 +48,8 @@ class VoiceGate(Cog):
 
     def __init__(self, bot: Bot) -> None:
         self.bot = bot
+        # voice_verification_channel set to None so that we have a placeholder to get it later in the cog
+        self.voice_verification_channel = None
 
     @property
     def mod_log(self) -> ModLog:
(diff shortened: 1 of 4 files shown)


tagptroll1 (approved):
looks good to me :) (after fixing the linting/tests)

bot/exts/moderation/voice_gate.py
```diff
@@ -144,6 +166,10 @@ async def on_message(self, message: discord.Message) -> None:
 
         # When it's bot sent message, delete it after some time
         if message.author.bot:
+            # Comparing the message with the voice ping constant
+            if message.content.endswith(VOICE_PING[-45:]):
```
Akarys42, line 170: I don’t quite like this, if we ever change the message and the first 45 characters aren’t the same for any reason, the code will break. Also, a user could spam the channel using the same 45 characters. What I think should be done is not deleting messages in `on_message` if the user is a bot, and instead handle deletion were the message is created, you’d just have to delete the message sent line 147 in the same function.

Akarys42 (changes requested):
The PR looks good, there’s just one thing with how you handle deletion that could lead to trouble.

bot/exts/moderation/voice_gate.py
```diff
@@ -28,12 +29,26 @@
...
+    "If you don't yet qualify, you'll be told why!"
+)
+
 
 class VoiceGate(Cog):
     """Voice channels verification management."""
 
-    def __init__(self, bot: Bot):
+    # RedisCache[t.Union[discord.User.id, discord.Member.id], t.Union[discord.Message.id, bool]]
+    redis_cache = RedisCache()
```
kwzrd, line 43: I feel like there should be a comment here explaining what the values represent. There are some bits of information scattered around the cog, but better to have them here.

bot/exts/moderation/voice_gate.py
MrHemlock: The message is only checked if the message is from a bot, so a user couldn't trigger it

bot/exts/moderation/voice_gate.py
```diff
@@ -28,12 +29,26 @@
...
+
+    def __init__(self, bot: Bot) -> None:
         self.bot = bot
+        self._init_task = self.bot.loop.create_task(self._async_init())
+
+    async def _async_init(self) -> None:
+        await self.bot.wait_until_guild_available()
+        self._voice_verification_channel = self.bot.get_channel(Channels.voice_gate)
```
kwzrd, line 51: I wonder if this is completely safe, i.e. could the instance get stale on bot restart or similar? We do it elsewhere too, just wondering. I'm guessing the linter doesn't yell about the lack of docstring because it's a private func.

bot/exts/moderation/voice_gate.py
Akarys42: > The message is only checked if the message is from a bot, so a user couldn't trigger it

Right, I totally missed that. But I still think there’s a better solution. 😄

bot/exts/moderation/voice_gate.py
```diff
@@ -160,6 +186,35 @@ async def on_message(self, message: discord.Message) -> None:
...
+
+            log.trace("User is unverified. Send ping.")
+            message = await self._voice_verification_channel.send(VOICE_PING.format(member.mention))
+            await self.redis_cache.set(member.id, message.id)
+
+            # Message will try to be deleted after 1 minutes. If it fails, it'll do so silently
+            await message.delete(delay=GateConf.voice_ping_delete_delay)
+        else:
+            log.trace("User is either in the cache or not in a voice channel. Ignore.")
+            return
```
kwzrd, line 216: The return at the end of the function feels awkward. Could flip the conditionals and return early like we do with the `member.bot` check, I'd probably find that cleaner (it's the same pattern of `check` - `log` - `return`).

config-default.yml
```diff
@@ -522,7 +522,7 @@ voice_gate:
     minimum_messages: 50  # How many messages a user must have to be eligible for voice
     bot_message_delete_delay: 10  # Seconds before deleting bot's response in Voice Gate
     minimum_activity_blocks: 3  # Number of 10 minute blocks during which a user must have been active
-
+    voice_ping_delete_delay: 60 # Seconds before deleting the bot's ping to user in Voice Gate
```
kwzrd, line 525: Rest of the config file uses two spaces before the hash.

bot/exts/moderation/voice_gate.py
kwzrd, line 170: What's the justification for `[-45:]`? I find this awkward. The message constant could simply not contain the mention placeholder. When you send the message, do:
```python
f"Hello, {member}! {MESSAGE}"
```
Then this holds with no slicing:
```python
message.content.endswith(MESSAGE)
```

bot/exts/moderation/voice_gate.py
```diff
@@ -160,6 +186,35 @@ async def on_message(self, message: discord.Message) -> None:
         with suppress(discord.NotFound):
             await message.delete()
 
+    @Cog.listener()
+    async def on_voice_state_update(self, member: Member, *_) -> None:
+        """Pings a user if they've never joined the voice chat before and aren't verified."""
```
kwzrd, line 191: I don't care too much, but I don't see a reason to obfuscate the other params. I think for clarity's sake it's better to show what we receive, even if we don't use it. Later on, it may cause someone to build a needlessly complex solution to a problem that can be solved by simply using the params.

bot/exts/moderation/voice_gate.py
```diff
@@ -160,6 +186,35 @@ async def on_message(self, message: discord.Message) -> None:
...
+
+        in_cache = await self.redis_cache.get(member.id, None)
+
+        # member.voice will return None if the user is not in a voice channel
+        if in_cache is None and member.voice is not None:
+            log.trace("User not in cache and is in a voice channel")
+            verified = any(Roles.voice_verified == role.id for role in member.roles)
+            if verified:
+                log.trace("User is verified, add to the cache and ignore")
+                #  redis cache does not accept None, so False is used to signify no message
```
kwzrd, line 204: There's an extra space after the hash.

kwzrd (dismissed):
Writing `False` into the cache feels a little hacky to me. How would you feel about writing a `-1` or similar to signify that there's no message to be removed? It's not an ideal solution, ideally we'd want a set of pinged user IDs, but I think that's not currently possible with our wrapper, so "abusing" the keys sort of makes sense.

bot/exts/moderation/voice_gate.py
```diff
@@ -160,6 +186,35 @@ async def on_message(self, message: discord.Message) -> None:
...
+                await self.redis_cache.set(member.id, False)
+                return
+
+            log.trace("User is unverified. Send ping.")
+            message = await self._voice_verification_channel.send(VOICE_PING.format(member.mention))
+            await self.redis_cache.set(member.id, message.id)
```
kwzrd, line 210: This races with `_async_init`. There is no guarantee that the task has completed, so the attr may not yet exist. You can await the task, or simply get rid of `_async_init`, await the guild ready event here, and grab a fresh instance every time. I also think that attrs that are not set in the `__init__` should be explicitly annotated or assigned a placeholder `None` at init time, so that you don't have to worry about the potential `AttributeError`

bot/exts/moderation/voice_gate.py
MrHemlock: Should I correct it in the other places in the file as well?  The `voice_verify` command has that as well.

kwzrd: If you want. :smile: I wouldn't ask for it. This comment is opinionated anyway, if you prefer it with the `*_` I will live.

kwzrd: Oh, it's different for a command. I imagine the command has it so that you can do `!voice_verify a b c d`. For the listener, we know exactly what we will receive, so we can annotate the params.

bot/exts/moderation/voice_gate.py
MrHemlock: Hadn't thought about that.

MrHemlock: Corrected in latest commit

bot/exts/moderation/voice_gate.py
MrHemlock: Added in latest commit

bot/exts/moderation/voice_gate.py
MrHemlock: Ended up removing the `_async_init` in favor of moving it all to the `on_voice_state_update` event listener.

bot/exts/moderation/voice_gate.py
MrHemlock: Restructured with this in mind in latest commit.

config-default.yml
MrHemlock: Corrected in latest commit

bot/exts/moderation/voice_gate.py
MrHemlock: Changed in the latest commit.

bot/exts/moderation/voice_gate.py
MrHemlock: Changed in the latest commit.

bot/exts/moderation/voice_gate.py
MrHemlock: Corrected in latest commit.

bot/exts/moderation/voice_gate.py
MrHemlock: Corrected in the latest commit.

bot/exts/moderation/voice_gate.py
```diff
@@ -160,6 +183,46 @@ async def on_message(self, message: discord.Message) -> None:
...
+
+                log.trace("User is unverified. Send ping.")
+                await self.bot.wait_until_guild_available()
+                voice_verification_channel = self.bot.get_channel(Channels.voice_gate)
+
+                message = await voice_verification_channel.send(f"Hello, {member.mention}! {VOICE_PING}")
+                await self.redis_cache.set(member.id, message.id)
+
+                # Message will try to be deleted after 1 minutes. If it fails, it'll do so silently
+                await message.delete(delay=GateConf.voice_ping_delete_delay)
```
kwzrd, line 224: Looking good, thanks for addressing all the feedback.

We no longer need many of the else clauses here because of having adjusted the logic to return early. Due to the rewrite, we could have just e.g.:
```python
if member.voice is None:
    ...
    return

in_cache = ...

if in_cache:
    ...
    return

verified = ...
```

The early return helps us avoid the deep indentation, which is difficult to follow.

MrHemlock: Addressed in latest commit.

bot/exts/moderation/voice_gate.py
```diff
@@ -53,6 +67,13 @@ async def voice_verify(self, ctx: Context, *_) -> None:
         - You must not be actively banned from using our voice channels
         - You must have been active for over a certain number of 10-minute blocks
         """
+        # If user has received a ping in voice_verification, delete the message
+        if message_id := await self.redis_cache.get(ctx.author.id, NO_MSG):
+            with suppress(discord.NotFound):
+                ping_message = await ctx.channel.fetch_message(message_id)
+                await ping_message.delete()
```
kwzrd, line 74: This could be simplified by directly using the http method if you want:
```python
await self.bot.http.delete_message(channel_id, message_id)
```
Then we'd only be making one API call as we don't have to fetch the message first.

bot/exts/moderation/voice_gate.py
```diff
@@ -160,6 +185,45 @@ async def on_message(self, message: discord.Message) -> None:
...
+
+        # member.voice will return None if the user is not in a voice channel
+        if member.voice is None:
+            log.trace("User not in a voice channel. Ignore.")
+            return
+
+        in_cache = await self.redis_cache.get(member.id, None)
+        if in_cache:
+            log.trace("User already in cache. Ignore.")
+            return
```
kwzrd, line 208: This should check `if in_cache is None`. There can be the `NO_MSG` flag, which is falsey.

kwzrd (commented):
Solution works fine, although there's one small bug.

The listener method is very easy to read now after it was refactored to exit early, good job. The trace logging makes this very pleasant to test.

kwzrd (commented):
I think there's room for improvement at the end of the voice state listener. Although it doesn't fail in an observable way, it suffers from two issues:
* We always attempt to delete the message after 60 seconds with no regard for whether it was already deleted by the `on_message`
* Once the listener deletes it, it does not write this information into the cache, so the `on_message` will still attempt to delete it

Both of these issues are relatively easy to resolve. Instead of the "fire and forget" delete with a delay, we could sleep the coroutine for the 60 seconds, then:
* Check the cache ~ is there a `NO_MSG`? The message was already cleared by the `on_message`, and we shouldn't make the request
* Otherwise, delete the message, and put the `NO_MSG` in the cache ~ the `on_message` will then not attempt to also delete the message

bot/exts/moderation/voice_gate.py
```diff
@@ -53,6 +69,13 @@ async def voice_verify(self, ctx: Context, *_) -> None:
         - You must not be actively banned from using our voice channels
         - You must have been active for over a certain number of 10-minute blocks
         """
+        # If user has received a ping in voice_verification, delete the message
+        if message_id := await self.redis_cache.get(ctx.author.id, NO_MSG):
+            with suppress(discord.NotFound):
+                self.voice_verification_channel = self.bot.get_channel(Channels.voice_gate)
+                await self.bot.http.delete_message(self.voice_verification_channel, message_id)
```
kwzrd, line 76: Sorry, I may have confused you. The http method does *not* take the channel object, it takes the ID, which is why it makes this situation easier. You can do:
```py
delete_message(Channels.voice_gate, message_id)
```
Without fetching or getting the channel. The channel never needs to be retrieved in any way.

bot/exts/moderation/voice_gate.py
```diff
@@ -160,6 +187,47 @@ async def on_message(self, message: discord.Message) -> None:
...
+        await self.bot.wait_until_guild_available()
+        voice_verification_channel = self.bot.get_channel(Channels.voice_gate)
+
+        message = await voice_verification_channel.send(f"Hello, {member.mention}! {VOICE_PING}")
+        await self.redis_cache.set(member.id, message.id)
+
+        await asyncio.sleep(60)
+        if message := await self.redis_cache.get(member.id):
+            await message.delete()
+            await self.redis_cache.set(member.id, NO_MSG)
```
kwzrd, line 229: This should still use the configured value, not a hardcoded `60`. For safety it can still suppress `NotFound` on the delete, although it should no longer be possible unless we get very unlucky and the `on_message` runs in parallel (the cache access is not synchronous).

bot/exts/moderation/voice_gate.py
```diff
@@ -160,6 +187,47 @@ async def on_message(self, message: discord.Message) -> None:
...
+
+        # member.voice will return None if the user is not in a voice channel
+        if member.voice is None:
+            log.trace("User not in a voice channel. Ignore.")
+            return
+
+        in_cache = await self.redis_cache.get(member.id, NO_MSG)
+        if in_cache:
+            log.trace("User already in cache. Ignore.")
+            return
```
kwzrd, line 210: I think this still doesn't behave properly. Having a default of `NO_MSG` is bad because then you cannot tell whether the member was or was not in the cache. I originally proposed that it checks whether `in_cache` is None (with a None default), but now I realize that it could also use `redis_cache.contains`, that would read the best.

kwzrd: To expand on this: we want the if-condition to hold only if the member ID is not in the cache. If we `get` with a None default, `in_cache` being None means that the member wasn't there, and `NO_MSG` means that it *was* there but the message is deleted. But we need to check for the None explicitly, because both are falsey. If we use a `NO_MSG` default we cannot tell whether the ID was there or not, so it doesn't make sense for a default value.

Ultimately this should probably just be:
```python
if not await self.redis_cache.contains(member.id):
```

It did not occur to me initially, sorry.

kwzrd:
We've agreed with @MrHemlock that I'd help address some of the feedback so that we can merge it soon.

bot/exts/moderation/voice_gate.py
kwzrd: Resolved by 59b34c7.

bot/exts/moderation/voice_gate.py
kwzrd: Use config in d6820a2, correctly delete in 92132af.

bot/exts/moderation/voice_gate.py
kwzrd: See dbff099.

bast0006:
I'm spying a race issue here, when users might run into the pathological case when discord repeatedly spam joins/leaves the voice channel. This can result in over 6 joins/leave events per second, and while I'm not sure if the issue is fixed and do not have a way to reproduce it under my control, I think we should be careful and address that case.

At the moment, the user-ping lockout in the cache is split over a send message and a wait-ready event, which makes it especially prone to getting re-entered in that situation, especially since there have been incidents where the bot takes 1-2 seconds to respond to things, depending on load and events happening elsewhere. We probably need a better lockout. 

Would it be reasonable to have a sentry "sending message" value to set into redis while we wait to send the message?

thomaspet (approved):
I am not apposed to the new changes, looks good still

kwzrd:
@bast0006 You're definitely right that there's still plenty of opportunity for racing, both with concurrent voice events and the `on_message` too. I'd consider it pathological because I wasn't able to create such a situation myself, but one needs to consider that in the production environment the event loop is busier, which I imagine creates larger "windows" of opportunity for things to go wrong.

> Would it be reasonable to have a sentry "sending message" value to set into redis while we wait to send the message?

Using Redis already comes with the problem that you can only interface with it asynchronously, so I don't think it's a bulletproof solution. A lock probably makes more sense, because you never have to `await` anything to acquire it.

I think I have an idea for how to solve it using a smart locking util that @MarkKoz wrote for another PR. It should allow us to prevent both the message & the voice state listener from runninng concurrently for the same user, using a decorator, which avoids having to add extra logic to the actual code. I'm planning to look at it soon.

kwzrd (dismissed):
I'm going to block this again until we make it async-safe.

kwzrd (approved):
Okay, I believe that this PR should no longer be prone to race conditions. The voice state listener performs two primary actions:

* Send the newcomer notification, if appropriate
* Delete the notification after the configured delay, if it hasn't been deleted yet

This first part suffered from a dangerous race condition, as pointed out by @bast0006. This has been addressed in 4b60c21 by making the whole procedure an atomic transaction. The second part was then prone to racing with the `!voiceverify` command, and is similarly made atomic by b4220a3.

I think this is now fine for production.

thomaspet (approved):
A side form ghostpings, which i suppose we can't do much about.. Still looks good

MrHemlock:
Everything is working as intended.  Pushing her through.

PR #1273, opened 2020-11-06
Detect codeblock language specifiers with special characters

SebastiaanZ:
The highlighting package Discord uses, `highlight.js`, has a couple of language specifiers that include the characters `-`, `+`, and `.`. One example is a Python-related language specifier: `python-repl`. However, the regex pattern we used to detect codeblocks did not account for these characters and, as such, did not detect that codeblocks using language specifiers that included these characters had a language specifier.

This PR changes the character set we use to match language specifiers to make sure we match all valid language specifiers (see [this overview](https://github.com/highlightjs/highlight.js/blob/master/SUPPORTED_LANGUAGES.md) of supported language specifiers in `highlight.js`).

### DIFF
--- a/bot/exts/info/codeblock/_parsing.py
+++ b/bot/exts/info/codeblock/_parsing.py
@@ -36,7 +36,7 @@ _RE_CODE_BLOCK = re.compile(
         (?P<tick>[{''.join(_TICKS)}]) # Put all ticks into a character class within a group.
         \2{{2}}                       # Match previous group 2 more times to ensure the same char.
     )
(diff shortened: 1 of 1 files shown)


kwzrd (approved):
I don't feel qualified to review regex, but this looks straight-forward. I can confirm that it fixes the issue with `python-repl` & works fine in [regex101.com](https://regex101.com/).

Den4200 (approved):
Similarly to kwzrd, I've tested this and `python-repl` does not trigger a filter anymore. The regex looks fine to me, but I'm no expert with it either.

PR #1275, opened 2020-11-07
Parse URL and match domain name in filtering to avoid false positives

### LINKED ISSUE #1260: Adding special logic when matching domains (bar.com matches foobar.com)
When a domain like `cookies.com` is added to the blacklist, websites with domains containing `cookies` are triggered as well, for example, `deliciouscookies.com`. (can't link actual example because it's NSFW)

Maybe some special casing can be implemented to prevent `deliciouscookies.com` from triggering? I know that you can do `http://cookies.com` and `https://cookies.com`, but this particular website was linked without the protocol.

ks129:
Fixes #1260 

Before bot just used `in` check to find blacklisted domains, but this resulted false positives like is showed in linked issue. Now this find all URLs and parse them with `urllib` parser, then get domain name and check it against blacklist.

### DIFF
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -3,6 +3,7 @@ import logging
 import re
 from datetime import datetime, timedelta
 from typing import Any, Dict, List, Mapping, NamedTuple, Optional, Union
+from urllib import parse
 
 import dateutil
 import discord.errors
@@ -444,8 +445,9 @@ class Filtering(Cog):
         text = text.lower()
         domain_blacklist = self._get_filterlist_items("domain_name", allowed=False)
 
-        for url in domain_blacklist:
-            if url.lower() in text:
+        for url in URL_RE.findall(text):
+            parsed = parse.urlparse(url)
+            if parsed.netloc.lower() in domain_blacklist:
                 return True
 
         return False


MarkKoz (changes requested):
This isn't too accurate because not all URLs posted by users will include the HTTP scheme. Unfortunately, this will cause problems with urllib even if the URL regex is adjusted:

> Following the syntax specifications in RFC 1808, urlparse recognizes a netloc only if it is properly introduced by ‘//’. Otherwise the input is presumed to be a relative URL and thus to start with a path component.

Furthermore, the `netloc` may include a port and/or a subdomain e.g. `www.cwi.nl:80`. Some filters may need to match only on specific subdomains while others will need to match all subdomains. Adding a separate filter for each possible subdomain is infeasible, so the code needs to somehow account for this.

ks129:
@MarkKoz #1276 this issue is for new filter that handles subdomains too, not only exact matches.

MarkKoz:
Though I think that issue could be addressed together with this one, even if set aside, my other points still need to be addressed. Someone could easily spoof the filter by omitting the scheme or specifying a port.

ks129:
Leaving this to somebody smarter 😅

Akarys42:
Reopening this PR in order to keep the work already done. If somebody wants to take over, feel free to do so!

PR #1283, opened 2020-11-12
Remove selenium from the element list

Akarys42:
This could lead to some confusion with the users believing that this channel is reserved to help related to the selenium tool.

I’ve gone through the list, there don’t seem to be any other problematic element, although it would be nice if someone could give it a quick second look.

### DIFF
--- a/bot/resources/elements.json
+++ b/bot/resources/elements.json
@@ -32,7 +32,6 @@
     "gallium",
     "germanium",
     "arsenic",
-    "selenium",
     "bromine",
     "krypton",
     "rubidium",


Den4200:
I believe @SebastiaanZ and I agreed to just keep it until we actually saw issues with the name. It looks okay so far - nothing yet.

Akarys42:
Fair ‘nuff, shall we close this PR?

mbaruh:
So going over the channel there seem to be some people who get confused from occasionally and drop selenium questions there in the middle of a help session, but those users could possibly do the same in another occupied channel otherwise.
Personally I wouldn't mind either way. If we think it's a possible issue lets just get rid of the name.

SebastiaanZ:
I wanted to see what would happen. If you've observed selenium-related questions being dropped in the middle of ongoing conversations in that channel, then we should just remove it. It's just a name.

The only consideration I have is that if we do ever add atomic numbers to the channel names, there would be a more obvious gap, but how obvious is a missing number in an unsorted list of 40 channels split over three categories? I don't really think it's going to be an issue.

Akarys42:
I don't think we would notice a missing number, since they aren't sorted, yeah.

mbaruh:
You can try `in:help-selenium selenium`. It's not very often, but it seems to happen occasionally.

SebastiaanZ:
Let's just remove it then. It's just a name.

mbaruh (approved):

ks129 (approved):

PR #1284, opened 2020-11-13
Migrate CI Pipeline to GitHub Actions

SebastiaanZ:
I've migrated our Azure CI Pipeline to GitHub Actions. While the general workflow is the same, there are a few changes:

- `flake8` is no longer run by `pre-commit`, but rather by a separate action that adds annotations to the GH Action results page.

- As we no longer have need for xml-formatted coverage files, the xmlrunner for unittest has been removed as a dependency. Instead, we now publish our coverage results to [coveralls.io](https://coveralls.io). Each Actions run will echo a unique jobs link to the output and our README.md has a badge that points to coveralls as well.

- We use version 2 of docker's GitHub Action build-and-push flow, which is split over multiple steps instead of one.

- I have changed the badges to GitHub Actions and coveralls.io badges.

Note: Because we accept PRs from forks, we need to be a bit careful with our secrets. While we do use the `pull_request_target` event, we should not expose secrets in steps that run code from the repository.


.github/workflows/lint-test-build.yml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Install dependencies using pipenv
+        if: steps.python_cache.outputs.cache-hit != 'true'
+        run: |
+          pip install pipenv
+          pipenv install --dev --deploy --system
+
+      - name: Pre-commit Environment Caching
+        uses: actions/cache@v2
+        id: pre_commit_cache
```
MarkKoz, line 62: An ID is not necessary since nothing is referencing this step.

.github/workflows/lint-test-build.yml
```diff
@@ -0,0 +1,122 @@
...
+        with:
+          path: ${{ env.PRE_COMMIT_HOME }}
+          key: "precommit-0-${{ runner.os }}-${{ env.PRE_COMMIT_HOME }}-\
+          ${{ steps.python.outputs.python-version }}-\
+          ${{ hashFiles('./.pre-commit-config.yaml') }}"
+
+      # We will not run `flake8` here, as we will use a separate flake8 action
+      - name: Run pre-commit hooks
+        run: export PIP_USER=0; SKIP=flake8 pre-commit run --all-files
```
MarkKoz, line 71: I think the comment about `PIP_USER=0` was useful.

.github/workflows/lint-test-build.yml
```diff
@@ -0,0 +1,122 @@
...
+            python -Wignore -m coverage run -m unittest
+            coverage report -m
+
+      # This step will publish the coverage reports coveralls.io and
+      # print a "job" link in the output of the GitHub Action
+      - name: Publish coverage report to coveralls.io
+        env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          pip install coveralls
```
MarkKoz, line 94: This is going to get written to the cache even if it was restored successfully, right? Should this be in the pipfile anyway? The XML dependency was, after all.

.github/workflows/lint-test-build.yml
```diff
@@ -0,0 +1,122 @@
...
+          coveralls
+
+  build-and-push:
+    needs: lint-test
+    if: github.event_name != 'pull_request_target' && github.ref == 'refs/heads/master'
```
MarkKoz, line 99: Again, does this need `success()` i.e. ensuring the previous job succeeded?

MarkKoz (changes requested):

.github/workflows/lint-test-build.yml
SebastiaanZ: The `success()` is not needed here as `needs` requires the specified jobs to have completed successfully for this job to run:

> Identifies any jobs that must complete successfully before this job will run. ([source](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds))

I've tested the combination of `needs` and `if` with test runs by varying the result of the `lint-test` job and the `github.ref` we're looking for: It seems to work as you'd expect.

.github/workflows/lint-test-build.yml
SebastiaanZ: I was a bit torn on this, but I agree that it should probably in there. It is currently persisted, the caching process will run just before the job clean-up steps are run, but it's better to have it where all our other dependencies are.

The reason I did not do it was because this is only needed in CI; there's no need to install or require this dependency in local dev environments. However, keeping all dependencies in one spot is also valuable and it's a relatively small package anyway.

.github/workflows/lint-test-build.yml
SebastiaanZ: I'll add it.

.github/workflows/lint-test-build.yml
```diff
@@ -0,0 +1,122 @@
...
+        uses: actions/cache@v2
+        id: python_cache
+        with:
+          path: ${{ env.PYTHONUSERBASE }}
+          key: "python-0-${{ runner.os }}-${{ env.PYTHONUSERBASE }}-\
+          ${{ steps.python.outputs.python-version }}-\
+          ${{ hashFiles('./Pipfile', './Pipfile.lock') }}"
+
+      - name: Install dependencies using pipenv
+        if: steps.python_cache.outputs.cache-hit != 'true'
```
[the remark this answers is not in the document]
SebastiaanZ: Not sure, the expression I've used here was the recommended expression for steps that need to run conditionally on the cache being restored in the [README for the cache action](https://github.com/actions/cache#skipping-steps-based-on-cache-hit).

I think the documentation implies that if you have no explicit "status" check function, it will use `success()` by default. The `if` we have does not use a status function, so it should also depend on the `success()` function result by default.

This is indeed consistent with the behavior I had in my test runs: When I added an explicit failing step as the first step in the job, this step did not get run. See [this run](https://github.com/python-discord/bot/runs/1398356542?check_suite_focus=true) with the manual `exit 1` at the start of the job.

.github/workflows/lint-test-build.yml
```diff
@@ -0,0 +1,122 @@
...
+      PIPENV_NOSPIN: 1
+      PRE_COMMIT_HOME: ${{ github.workspace }}/.cache/pre-commit-cache
+      PYTHONUSERBASE: ${{ github.workspace }}/.cache/py-user-base
+
+    steps:
+      - name: Add custom PYTHONUSERBASE to PATH
+        run: echo '${{ env.PYTHONUSERBASE }}/bin/' >> $GITHUB_PATH
+
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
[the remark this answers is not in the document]
SebastiaanZ: I'll add that.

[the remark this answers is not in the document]
SebastiaanZ: Added in the next line.

.github/workflows/lint-test-build.yml
```diff
@@ -0,0 +1,125 @@
...
+      # action. As pre-commit does not support user installs, and we don't
+      # really need it, we set PIP_USER=0.
+      - name: Run pre-commit hooks
+        run: export PIP_USER=0; SKIP=flake8 pre-commit run --all-files
+
+      # This step requires `pull_request_target` due to the use of annotations
+      - name: Run flake8
+        uses: julianwachholz/flake8-action@v1
+        with:
+          checkName: lint
```
jb3, line 82:
```suggestion
          checkName: lint-test
```

.github/workflows/lint-test-build.yml
```diff
@@ -0,0 +1,125 @@
...
+      REDDIT_SECRET: ham
+      REDIS_PASSWORD: ''
+
+      PIP_NO_CACHE_DIR: false
+      PIP_USER: 1
+      PIPENV_HIDE_EMOJIS: 1
+      PIPENV_IGNORE_VIRTUALENVS: 1
+      PIPENV_NOSPIN: 1
+      PRE_COMMIT_HOME: ${{ github.workspace }}/.cache/pre-commit-cache
+      PYTHONUSERBASE: ${{ github.workspace }}/.cache/py-user-base
```
jb3, line 30: Some brief documentation on these variables would be nice

jb3 (changes requested):

MarkKoz (approved):

lemonsaurus (approved):
Looks great! Let's do it!

PR #1287, opened 2020-11-16
Updated available help channel embed

gustavwilliam:
Closes parts of python-discord/organisation#305.

This new embed is meant to be clearer and more user friendly, as well as providing more easy to understand tips for asking better questions.

### Screenshots
This is the previous version:
Screenshot 2020-11-16 at 01 40 17

This is the new version:
Screenshot 2020-11-16 at 01 39 28

### Reasoning
The green color of the embed — which is the same as one of the colors used in SeasonalBot — is meant to provide a clear feeling of the channel being available. The current grey color doesn’t give that. The ✅ emoji also helps contribute to that feeling and goes well with the category name of the available help channels.

After extensive discussions (and bike-shedding) in `#organisation`, we arrived at this message, which should provide maximum effect due to the conciseness and friendly tone of it. I would like to underline the importance of keeping this embed short, since that is what will result in more users actually reading it — as well as the striking green color.

I think this new embed will work much better and we'll hopefully even be seeing some improved question in the near future!


Akarys42 (approved):
Bike-shedding nicely done, this new embed is way better!

bot/exts/help_channels.py
```diff
@@ -28,17 +28,21 @@
...
-Try to write the best question you can by providing a detailed description and telling us what \
-you've tried already. For more information on asking a good question, \
-check out our guide on **[asking good questions]({ASKING_GUIDE_URL})**.
+**Send your question here to claim the channel**
+This channel will be dedicated to answering your question only. We’ll try to answer and help you solve the issue.
+
+**Keep in mind:**
+• It's always ok to just ask your question. You don't need permission.
+• Explain what you expect to happen and what actually happens.
+• Include a code sample and error message, if you got one.
```
ChrisLovering, line 37: Possible grammer edit needed?
`Include a code sample and/or error message, if you have either.`

gustavwilliam: The intention was that "if you got one" would mean "if you got any of the two things. Maybe using "any" instead of "one" would be better, though. What do you think?

gustavwilliam: This has been updated. I'm happy with how it looks now.

gustavwilliam:
The embed was updated to use a new icon and the author field instead of a title. This allows the icon to be centered.
Screenshot 2020-11-17 at 23 09 18

bot/exts/help_channels.py
ChrisLovering: I'm still not sure about the use of got here. Surely "have" is the better possessive, since they still have them.

I'll leave this to the core team to decide on though.

gustavwilliam: The word "got" here refers to the things they got when they were trying it.

mbaruh (changes requested):
A couple of things:

* "We'll try" - I'd like to avoid a rhetoric that implies that there is someone dedicated to helping, and say something like "Other members will try".
* I wonder if we should add a couple of words about the help channel being moved to the `Occupied Help Channels` category

gustavwilliam:
Updated to say "others will" instead of "we'll".

@mbaruh I think adding that last part is a bit verbose. If we notice a drastic _increase_ in people being confused by the where their channel went, let's add it. Right now though, I think we're better off not including it. The same goes for the rest of the embed as well; if we see something that needs to be mentioned, let's change it then.

mbaruh:
@gustavwilliam This currently gives me two checkmarks?

image

gustavwilliam:
Should be all good now.

mbaruh (approved):

PR #1290, opened 2020-11-18
Ensure flake8 actually lints pull requests

SebastiaanZ:
Unfortunately, the way I'd configured our workflow meant that flake8 was running in the context of the target repository, which means it was linting our master branch *without* the changes introduced by the PR. This is obviously a bit useless when it comes to guarding our codebase from linting errors. That's why I've changed the way we run `flake8`.

I've also decided to split the workflow file in two separate ones: one to lint/test and one to build/deploy.

Finally, I've updated the docker-compose file to use GHCR for pulling snekbox images. This should already work, although it's probably best if we wait until merging python-discord/snekbox#79.


.github/workflows/build.yml
```diff
@@ -0,0 +1,51 @@
...
+        id: sha_tag
+        run: |
+          tag=$(cut -c 1-7 <<< $GITHUB_SHA)
+          echo "::set-output name=tag::$tag"
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
```
lemonsaurus, line 29: Can you add some information in a comment as to why we need Buildx and what we're using it for?

lemonsaurus (approved):
This is gorgeous. Can you document buildx? After that it's good to go, I think.

.github/workflows/build.yml
SebastiaanZ: We're using it because it's the current recommended workflow by docker, but, sure, I can add a comment.

MarkKoz (approved):

.github/workflows/build.yml
SebastiaanZ: I've added a comment. Is it enough?

PR #1293, opened 2020-11-20
Include Unicode emojis to emojis filter

### LINKED ISSUE #1292: Filter functionality for emoji-spam messages
Now that someone's figured out you can cause havok by dumping massive amounts of emojis into a channel, we need a filter to automatically manage this ASAP. Detection should be fairly simple, a very high emoji count is required for the effect (>20/message to >250 for varying impact) but emoji variance is not required. 

Tangentially: This is also likely related to a discord update that went out that has slowed the emoji loading process recently.

I think we can safely filter out and autodelete high emoji messages without significant false positives. 

We may want to handle both unicode emojis and server emojis. 

It's possible for unicode emojis to get a higher "load" because they could as one character for message transmission purposes but server emojis count for more. However, server emojis may be animated, and will apply significant load by that fact as well.

ks129:
Closes #1292 

- Installed `emoji` package for parsing.
- Include Unicode emojis to emojis rule.


Akarys42 (approved):
Good work, as always!

MrHemlock (approved):
Looks good to me.  Simple tweak is all it takes.

PR #1294, opened 2020-11-20
Help Channel Refactor

MarkKoz:
Functions were split into separate modules because the single module was monolithic. What remains in the cog is too interconnected to refactor any further (relying on a bunch of shared state like queues).

There isn't much else to go into here since the changes mostly consist of moving things without further modification. There are a few exceptions like 840348b642f2d197e15d98a823d6e431b0a65d2a and 5ca816cccfe25eebba49d1c86b53d6b9166ce41a.


bot/exts/help_channels/_cog.py
```diff
@@ -14,14 +11,14 @@
 
 from bot import constants
 from bot.bot import Bot
+from bot.exts.help_channels import _channels
+from bot.exts.help_channels._names import MAX_CHANNELS_PER_CATEGORY, create_name_queue
```
lemonsaurus: Feels inconsistent. Maybe better to just import `_names` as well and have consistent namespacing?

lemonsaurus (approved):
I really like this. Excellent work.

bot/exts/help_channels/_cog.py
lemonsaurus: This may have been changed in a later commit..

MarkKoz: It was still similar. Changed in 47d65fc

lemonsaurus: nice, I like it. 👍🏼

Numerlor (approved):

bot/exts/help_channels/_cog.py
```diff
@@ -439,8 +436,9 @@ async def on_message(self, message: discord.Message) -> None:
         if not is_available or _channel.is_excluded_channel(channel):
             return  # Ignore messages outside the Available category or in excluded channels.
 
-        log.trace("Waiting for the cog to be ready before processing messages.")
-        await self.ready.wait()
+        if not self.init_task.done():
+            log.trace("Waiting for the cog to be ready before processing messages.")
+            await self.init_task
```
SebastiaanZ, line 431: I suppose you don't even need the `.done()` check first, although it makes sense from a logical point of view. Not sure about the performance characteristics, but I doubt it really matters here anyway.

SebastiaanZ (approved):
Left a few comments, but nothing that stops an approval. It looks good and it really helps with keeping things tidy and organized.

bot/exts/help_channels/_message.py
```diff
@@ -1,23 +1,183 @@
...
+            return False
+
+        if _match_bot_embed(msg, AVAILABLE_MSG):
+            log.trace(f"#{channel} ({channel.id}) has the available message embed.")
+            return True
+
+    return False
+
+
+async def notify(channel: discord.TextChannel, last_notification: t.Optional[datetime]) -> t.Optional[datetime]:
```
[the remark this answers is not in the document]
MarkKoz: Yes. I forgot to document it. e5b073c

bot/exts/help_channels/_channel.py
```diff
@@ -0,0 +1,57 @@
...
+
+import discord
+
+from bot import constants
+from bot.exts.help_channels import _caches, _message
+
+log = logging.getLogger(__name__)
+
+MAX_CHANNELS_PER_CATEGORY = 50
+EXCLUDED_CHANNELS = (constants.Channels.how_to_get_help, constants.Channels.cooldown)
```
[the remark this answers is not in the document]
MarkKoz: I removed the constant but kept it as a tuple to maintain support for excluding multiple channels in the future.

bot/exts/help_channels/_cog.py
MarkKoz: I think I did this to avoid sending the log message, but I now realise it causes a bug so I've removed the `done()` check. See the message of d9f87cc for more details.

bot/exts/help_channels/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+
+from bot.bot import Bot
+
+log = logging.getLogger(__name__)
+
+
+def setup(bot: Bot) -> None:
+    """Load the HelpChannels cog."""
+    # Defer import to reduce side effects from importing the sync package.
+    from bot.exts.help_channels import _cog
```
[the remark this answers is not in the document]
MarkKoz: Automatic extension discovery has to import this subpackage to check if it's a valid extension. This is mainly to be able to list all extensions with the command that shows the loaded/unloaded status. Otherwise, it could have taken a more dynamic approach which attempts to load extensions as requested and considers them invalid if discord.py throws an error.

If this extension is not being loaded, then any code that gets executed upon importing the package is just a waste. Side effects could be compiling regex, reading file content, importing expensive third-party packages, etc. I wasn't thinking of anything specific in this case; it was done more as a precaution. In any case, it isn't too effective since the `_channel` module has to be imported to get the constant needed by `validate_config`. Could be avoided if I cared enough, but the gains are not very significant so I'm not sure why I bothered.

PR #1296, opened 2020-11-22
Removed unnecessary shadow infractions

mbaruh:
We have several infractions that are never used, most moderators don't even know they exist, and frankly there is no reason they should be used. This PR removes them.

These infractions are:

* shadowkick 
* shadowmute
* shadowtempban


MarkKoz (approved):

mbaruh:
After some discussions in the mod channels, we decided to do some things differently. I'll open a different pull request.

PR #1297, opened 2020-11-22
Implements Voice Channel Muting

### LINKED ISSUE #1160: !shh Equivalent Command for Voice Chat
## Description
There is currently no way to quickly silence users in voice chat with a command. This currently makes it somewhat difficult to moderate when there are a lot of users in chat talking over each other and the mods. Users also don't always pay attention to the #voice-chat text channel if mods try to de-escalate a situation via the text channel. 

This suggestion would be to implement a !shh-equivalent command that would be run in the voice chat text channels that would apply mutes or somehow otherwise silence everyone who wasn't a moderator+ in the voice chat itself. This would help the moderator de-escalate situations and effectively moderate. 

## Considerations
The implementation of this command would have to consider how to apply the silence appropriately to users in the VC as well as how to handle users leaving/joining the VC while the !shh-equivalent command is in effect.

### LINKED ISSUE #1282: Add ability for mods to shut down a voice channel
Mods should have the ability to prevent access to the channel, and boot anyone inside.
They currently have the ability to disconnect users one by one, and admins can revoke role perms, but it should be a simple command.

HassanAbouelela:
## Description
Adds voice channel muting functionality, similar to text channel muting. 

- Allows muting any channel from any other channel or itself
- Allows unmuting any channel from any other channel or itself
- Relays mute confirmation/failure to both the invocation channel and the muted channel (in the case of VC, the muted channel is replaced with the text-channel with the same name).
- Updates shh and unshh commands to match the silence and unsilence interface.

Closes #1160, closes #1282.

### Tests
- [x] Write tests for silence command
- [x] Write tests for unsilence command
- [x] Write tests for error handler (shh, unshh)

### Other
- [x] Relay failure messages to invocation channel when muting VC
- [x] Replace hardcoded matching of VC to VC-chat channel names 
- [x] Make member kicking in voice channel optional.


HassanAbouelela:
Note on 985b681 which implements an optional kick parameter:

Ideally, the kick parameter would be the last parameter, as it is optional even if other arguments such as channel are passed, but because `channel` is starred, that wouldn't be possible.

bot/exts/moderation/silence.py
```diff
@@ -93,87 +93,222 @@ async def _async_init(self) -> None:
...
+            if name in channel.name.lower():
+                return self.bot.get_channel(channels[name])
+
+    async def send_message(self, message: str, source_channel: TextChannel,
+                           target_channel: Union[TextChannel, VoiceChannel],
+                           alert_target: bool = False, duration: HushDurationConverter = 0) -> None:
+        """Helper function to send message confirmation to `source_channel`, and notification to `target_channel`."""
+        await source_channel.send(
+            message.replace("current", target_channel.mention if source_channel != target_channel else "current")
+                   .replace("{duration}", str(duration))
```
MarkKoz, line 126: This will be more readable as a normal if statement rather than an inline one. It also slightly bothers me that in the else case you're redundantly replacing the string with itself.

bot/exts/moderation/silence.py
```diff
@@ -93,87 +93,222 @@ async def _async_init(self) -> None:
...
+        for name in channels.keys():
+            if name in channel.name.lower():
+                return self.bot.get_channel(channels[name])
+
+    async def send_message(self, message: str, source_channel: TextChannel,
+                           target_channel: Union[TextChannel, VoiceChannel],
+                           alert_target: bool = False, duration: HushDurationConverter = 0) -> None:
+        """Helper function to send message confirmation to `source_channel`, and notification to `target_channel`."""
+        await source_channel.send(
+            message.replace("current", target_channel.mention if source_channel != target_channel else "current")
```
MarkKoz, line 125: Consider replacing "current channel" rather than just "current". This addresses two issues:

1. When replacing with a mention, it's already obvious that the mention is a channel. Therefore writing "channel" right after is redundant. I'd also find it awkward to read (though when it's "current" it feels fine for some reason).
2. It's more specific. You wouldn't need to make changes such as the one to `MSG_UNSILENCE_MANUAL` to avoid accidental replacements.

MarkKoz (changes requested):
Didn't get a chance to review tests tonight. Will do that later.

bot/constants.py
```diff
@@ -398,6 +398,8 @@ class Channels(metaclass=YAMLGetter):
     change_log: int
     code_help_voice: int
     code_help_voice_2: int
+    admins_voice: int
+    staff_voice: int
```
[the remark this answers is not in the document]
HassanAbouelela: These are already in the default_config.
[Lines 203 & 204](https://github.com/python-discord/bot/blob/master/config-default.yml#L203)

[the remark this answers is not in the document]
MarkKoz: Sorry, I didn't realise. However, those are the IDs for the voice channels, not the text channels. In any case, with my suggestion in another comment, you'd need the IDs for both the text and voice channels.

bot/exts/moderation/silence.py
```diff
@@ -93,87 +93,222 @@ async def _async_init(self) -> None:
...
+            "admin": Channels.admins_voice,
+            "staff": Channels.staff_voice
+        }
+        for name in channels.keys():
+            if name in channel.name.lower():
+                return self.bot.get_channel(channels[name])
+
+    async def send_message(self, message: str, source_channel: TextChannel,
+                           target_channel: Union[TextChannel, VoiceChannel],
+                           alert_target: bool = False, duration: HushDurationConverter = 0) -> None:
```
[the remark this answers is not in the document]
HassanAbouelela: I hadn't seen that guide before, thank you. I'll fix the rest of the style issues.

bot/constants.py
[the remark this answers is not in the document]
HassanAbouelela: Thanks for the heads up, the rest of the channels grouped with that one in the config are the text channels, so there will need to be quite a few channels added to the config.

bot/exts/moderation/silence.py
```diff
@@ -93,87 +93,222 @@ async def _async_init(self) -> None:
...
+
+                await voice_chat.send(
+                    message.replace("{duration}", str(duration)).replace("current", target_channel.mention)
+                )
+
+            else:
+                await target_channel.send(message.replace("{duration}", str(duration)))
+
     @commands.command(aliases=("hush",))
     @lock_arg(LOCK_NAMESPACE, "ctx", attrgetter("channel"), raise_error=True)
```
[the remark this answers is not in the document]
HassanAbouelela: I don't follow what you mean here. The function is only responsible for sending alerts to the source and target channel (replacing the target channel with a voice chat channel in case of VC). Choosing between the invocation channel and a channel argument was left up to the caller.

There is one check that is sort of related to that, the one that compares source to target channel to prevent duplicate messages.

bot/exts/moderation/silence.py
```diff
@@ -93,87 +93,222 @@ async def _async_init(self) -> None:
...
-        await channel.set_permissions(self._verified_role, overwrite=overwrite)
+        if isinstance(channel, TextChannel):
+            overwrite.update(send_messages=False, add_reactions=False)
+            await channel.set_permissions(self._verified_msg_role, overwrite=overwrite)
+        else:
+            overwrite.update(speak=False)
+            if kick:
+                overwrite.update(connect=False)
+
+            await channel.set_permissions(self._verified_voice_role, overwrite=overwrite)
```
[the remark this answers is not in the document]
HassanAbouelela: How would you get the overwrite object in the if and else branches without calling the `overwrites_for` first? Since it's different for both branches, you'll need to have duplicate code inside the branches or have two checks (current implementation).

The code of the bottom branches (line 238-247) can be refactored into the top branches (226-233), but then the call on 235/236 has to be duplicated.

bot/exts/moderation/silence.py
```diff
@@ -93,87 +93,222 @@ async def _async_init(self) -> None:
...
+        If `member` is passed, the mute only occurs to that member.
+        Permission modification has to happen before this function.
+
+        If `kick_all` is True, members will not be added back to the voice channel.
+        """
+        # Handle member picking logic
+        if member is not None:
+            members = [member]
+        else:
+            members = channel.members
```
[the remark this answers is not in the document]
HassanAbouelela: That is true, however, this function was intended to be a more general kick/mute function that can work for other features (see related: #1160).

I'd be fine with removing that until I get started implementing the other features if you believe that would be better.

bot/exts/moderation/silence.py
```diff
@@ -188,14 +323,21 @@ async def _unsilence(self, channel: TextChannel) -> bool:
...
-        overwrite = channel.overwrites_for(self._verified_role)
+        if isinstance(channel, TextChannel):
+            overwrite = channel.overwrites_for(self._verified_msg_role)
+        else:
+            overwrite = channel.overwrites_for(self._verified_voice_role)
+
         if prev_overwrites is None:
             log.info(f"Missing previous overwrites for #{channel} ({channel.id}); defaulting to None.")
-            overwrite.update(send_messages=None, add_reactions=None)
+            overwrite.update(send_messages=None, add_reactions=None, speak=None)
```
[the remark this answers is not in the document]
HassanAbouelela: They are just ignored (I believe at the d.py level, as `PermissionOverwrite` objects can accept any permissions). Do you think it would be better to add checks here and set the overwrites based on the channel type?

bot/exts/moderation/silence.py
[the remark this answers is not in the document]
MarkKoz: The `lock_arg` is locking on `ctx.channel`. This is no longer adequate since you've added a channel arg that may be specified instead of using the context. Therefore, the lock decorator needs to be adjusted so it can lock on the correct channel.

bot/exts/moderation/silence.py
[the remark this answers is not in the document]
MarkKoz: I don't understand how that issue differs from what you're already doing. It compares it to `!shh`, which is just an alias for the silence feature in this cog. The two issues look like duplicates. Muting individual users could be done through the Discord UI. If needed, a command could be added to keep track of user infractions like we do for bans and kicks, which could also be done through the UI. However, that is out of scope.

bot/exts/moderation/silence.py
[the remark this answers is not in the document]
MarkKoz: No, it's fine if they're being ignored.

bot/exts/moderation/silence.py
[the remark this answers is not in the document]
MarkKoz: I missed that, sorry. The `overwrites_for` will just have to be duplicated in the if and else blocks at the top.

bot/exts/moderation/silence.py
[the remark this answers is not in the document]
MarkKoz: It's still incorrect. Every argument must be on a separate line You also inconsistently used 4 and 8 space indents. There is also some test code which has similar style issues. For long function calls, consider assigning an argument to a variable and then passing the variable to the function. This will avoid having to break up the function into multiple lines. For example,

```py
            self.scheduler.schedule_later(
                30, 
                afk_channel.id, 
                afk_channel.delete(reason="Deleting temp mute channel.")
            )
```
versus
```py
            coro = afk_channel.delete(reason="Deleting temp mute channel.")
            self.scheduler.schedule_later(30, afk_channel.id, coro)
```

MarkKoz:
I didn't mean to completely get rid of the converter, just to change it's implementation to rely on the existing converters. When using a union, there isn't going to be an error for an ambiguous matches.

An alternative implementation is to create a subset of the text channel converter which does not support matching by plain text names. Such converter in a union with the regular voice channel converter would effectively funnel all plain text names to the voice converter. Of course the downside is that plain text names cannot be used to specify text channels, but I see no reason why this would be relied upon when the Discord client makes it very easy to mention text channels.

bot/exts/moderation/silence.py
```diff
@@ -93,87 +93,229 @@ async def _async_init(self) -> None:
...
+        await source_channel.send(source_reply)
+
+        # Reply to target channel
+        if alert_target and source_channel not in [target_channel, voice_chat]:
+            if target_is_voice_channel:
+                if voice_chat is not None:
+                    await voice_chat.send(message.replace("current channel", target_channel.mention))
+
+            else:
+                await target_channel.send(message)
```
MarkKoz, line 145: This can be simplified

```suggestion
        # Reply to invocation channel
        source_reply = message
        if source_channel != target_channel:
            source_reply = source_reply.replace("current channel", target_channel.mention)
        await source_channel.send(source_reply)

        # Reply to target channel
        if alert_target:
            if isinstance(target_channel, VoiceChannel):
                voice_chat = await self._get_related_text_channel(target_channel)
                if voice_chat and source_channel != voice_chat:
                    await voice_chat.send(message.replace("current channel", target_channel.mention))
            elif source_channel != target_channel:
                await target_channel.send(message)
```

bot/exts/moderation/silence.py
```diff
@@ -188,14 +330,25 @@ async def _unsilence(self, channel: TextChannel) -> bool:
...
+            overwrite.update(send_messages=None, add_reactions=None, speak=None, connect=None)
         else:
             overwrite.update(**json.loads(prev_overwrites))
 
-        await channel.set_permissions(self._verified_role, overwrite=overwrite)
+        if is_text_channel:
+            await channel.set_permissions(self._verified_msg_role, overwrite=overwrite)
+        else:
+            await channel.set_permissions(self._verified_voice_role, overwrite=overwrite)
+            await self._force_voice_sync(channel)
```
MarkKoz, line 350: Saving the result of `isinstance` is unnecessary since it's a cheap operation. The code could be simplified as follows:

```py
        if isinstance(channel, TextChannel):
            role = self._verified_msg_role
        else:
            role = self._verified_voice_role
        
        overwrite = channel.overwrites_for(role)

        if prev_overwrites is None:
            log.info(f"Missing previous overwrites for #{channel} ({channel.id}); defaulting to None.")
            overwrite.update(send_messages=None, add_reactions=None, speak=None, connect=None)
        else:
            overwrite.update(**json.loads(prev_overwrites))

        await channel.set_permissions(role, overwrite=overwrite)

        if isinstance(channel, VoiceChannel):
            await self._force_voice_sync(channel)
```

bot/exts/moderation/silence.py
```diff
@@ -93,87 +93,229 @@ async def _async_init(self) -> None:
...
         else:
-            await channel.send(MSG_UNSILENCE_SUCCESS)
+            await self.send_message(MSG_UNSILENCE_SUCCESS, msg_channel, channel, True)
 
-    async def _set_silence_overwrites(self, channel: TextChannel) -> bool:
+    async def _set_silence_overwrites(self, channel: Union[TextChannel, VoiceChannel], kick: bool = False) -> bool:
         """Set silence permission overwrites for `channel` and return True if successful."""
-        overwrite = channel.overwrites_for(self._verified_role)
-        prev_overwrites = dict(send_messages=overwrite.send_messages, add_reactions=overwrite.add_reactions)
+        is_text_channel = isinstance(channel, TextChannel)
```
MarkKoz, line 228: I think you misunderstood what I was aiming for. The problem is that there are if statements that are avoidable. The problem is not with calling `isinstance` twice. What you've done still keeps the avoidable if statements. I've corrected my original code suggestion for this (https://github.com/python-discord/bot/pull/1297#discussion_r529262670) and I still think you should implement it like that.

MarkKoz (changes requested):

bot/exts/moderation/silence.py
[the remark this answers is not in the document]
HassanAbouelela: That's a much neater implementation, thank you.

bot/exts/moderation/silence.py
HassanAbouelela: I definitely did, I'll go ahead and implement that correctly.

HassanAbouelela:
I removed the original converter because I realized all that's lost would be the ability to mix case, and as you mentioned, the ambiguity warning.

It didn't seem like those were critical features, and the discord library just picked the first option when the argument was ambiguous. Regardless, I'll go ahead and reimplement it with the suggested changes as it doesn't complicate the code base much.

But on the note of mixing case, would it be better to not allow that? It can lead to ambiguity when voice channels only differ by case. Alternatively, I can avoid changing the case, or I can rerun the search code with the original case before raising the bad argument error. Which do you think would be better?

bot/exts/moderation/silence.py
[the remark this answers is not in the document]
HassanAbouelela: I had originally understood that issue as something similar to mute, acting on a per-user basis. After rereading, this functionality will not be needed for either of the issues, so I'll remove it.

MarkKoz:
Yeah, I agree that those features aren't critical. I just figured that since you already did most the of the work, you may as well leave it in. However, since you've removed it again, I could live without it. I'll leave that decision up to you and other reviewers.

discord.py's converters are case sensitive so you don't need to worry about mixed case. I'm suggesting that you rely on them to do all the comparison work (i.e. by calling their `convert()` methods within your `convert()`) rather than doing it yourself.

mbaruh:
How does this relate to #1305? since both are making changes to constants.py and are about the voice channels.

HassanAbouelela:
This PR was just adding an existing config parameter to the constants file. #1305 adds new fields and restructures the default config, so I split it out into its own PR, as to not clutter this one.

bot/exts/moderation/silence.py
```diff
@@ -93,87 +93,222 @@ async def _async_init(self) -> None:
...
+        if afk_channel is None:
+            overwrites = {
+                channel.guild.default_role: PermissionOverwrite(speak=False, connect=False, view_channel=False)
+            }
+            afk_channel = await channel.guild.create_voice_channel("mute-temp", overwrites=overwrites)
+            log.info(f"Failed to get afk-channel, created temporary channel #{afk_channel} ({afk_channel.id})")
+
+            # Schedule channel deletion in case function errors out
+            self.scheduler.schedule_later(30, afk_channel.id,
+                                          afk_channel.delete(reason="Deleting temp mute channel."))
```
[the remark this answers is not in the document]
HassanAbouelela: I don't know what the final decision will be yet, but I've moved that code into its own function until we decide what to do.

bot/exts/moderation/silence.py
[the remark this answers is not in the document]
HassanAbouelela: Updated the PR description to reflect this.

HassanAbouelela:
Would appreciate a re-review, bearing in mind some changes still need discussion.

Discussion needed on:
- Keeping or removing the custom channel converter, in favor of a text-voice union
- Should the AFK voice channel be assumed to always exist (or should we keep the temporary channel creation function)

ks129:
@HassanAbouelela Can you please solve conflicts?

tests/bot/exts/moderation/test_silence.py
```diff
@@ -268,7 +415,66 @@ async def test_sent_correct_message(self):
...
+                        voice_sync.assert_called_once_with(target)
+
+            ctx.channel.send.reset_mock()
+            if target is not None and isinstance(target, MockTextChannel):
+                target.send.reset_mock()
+
+    @mock.patch.object(silence.Silence, "_set_silence_overwrites", return_value=True)
+    @mock.patch.object(silence.Silence, "_kick_voice_members")
+    @mock.patch.object(silence.Silence, "_force_voice_sync")
+    async def test_sync_or_kick_called(self, sync, kick, _):
```
MarkKoz, line 458: There should also be a test that neither are called when the target is not a voice channel.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -268,7 +415,66 @@ async def test_sent_correct_message(self):
...
+
+            with self.subTest("Test calls kick"):
+                await self.cog.silence.callback(self.cog, ctx, 10, kick=True, channel=channel)
+                kick.assert_called_once_with(channel)
+                sync.assert_not_called()
+
+            kick.reset_mock()
+            sync.reset_mock()
+
+            with self.subTest("Test calls sync"):
```
MarkKoz, line 474: I'm on the fence about whether I like subtests being used in this manner. I tend to only use them when inputs change but the asssertions don't  (so I can loop over the inputs). Otherwise, I write separate functions. I suppose they can still provide DRY benefits like this, especially when there's a lot of mock setup involved. However, that could be alleviated by setting up mocks in a separate function, kind of like pytest's fixtures.


Anyway, this is not a request to change it. I just felt inclined to share my thoughts.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -343,34 +549,45 @@ async def test_cached_unsilence_time(self, datetime_mock):
...
         """A task was not scheduled for a permanent silence."""
-        ctx = MockContext(channel=self.channel)
+        ctx = MockContext(channel=self.text_channel)
         await self.cog.silence.callback(self.cog, ctx, None)
         self.cog.scheduler.schedule_later.assert_not_called()
 
+    async def test_correct_permission_updates(self):
+        """Tests if _set_silence_overwrites can correctly get and update permissions."""
+        self.assertTrue(await self.cog._set_silence_overwrites(self.text_channel))
+        self.assertFalse(self.text_overwrite.send_messages or self.text_overwrite.add_reactions)
```
MarkKoz, line 583: These assertions are redundant. They're already covered by `test_silenced_channel`.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -343,34 +549,45 @@ async def test_cached_unsilence_time(self, datetime_mock):
...
+
+        self.assertTrue(await self.cog._set_silence_overwrites(self.voice_channel))
+        self.assertFalse(self.voice_overwrite.speak)
```
MarkKoz, line 586: There are many tests above that are specific to text channels only currently. Instead of these simple assertions, you should have duplicates of the text tests or subtests within which test the same things for the voice channels. Use your best judgement for which tests need voice-specific tests and which don't (e.g. the notifier ones probably don't).

tests/bot/exts/moderation/test_silence.py
```diff
@@ -413,6 +630,42 @@ async def test_sent_correct_message(self):
...
+                    if ctx.channel == target or target is None:
+                        ctx.channel.overwrites_for.return_value = unsilenced_overwrite
+                    else:
+                        target.overwrites_for.return_value = unsilenced_overwrite
+
+                    await self.cog.unsilence.callback(self.cog, ctx, channel=target)
+
+                    # Check Messages
+                    if ctx.channel == target or target is None:
+                        ctx.channel.send.assert_called_once_with(message)
```
MarkKoz, line 658: The same request I made for the silence test applies here too.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -491,3 +748,86 @@ async def test_preserved_other_overwrites(self):
                 del new_overwrite_dict['add_reactions']
 
                 self.assertDictEqual(prev_overwrite_dict, new_overwrite_dict)
+
+    @mock.patch.object(silence.Silence, "_unsilence", return_value=False)
+    @mock.patch.object(silence.Silence, "send_message")
+    async def test_unsilence_helper_fail(self, send_message, _):
```
MarkKoz, line 754: I guess this is a similar situation to `test_sent_to_correct_channel`. You could probably merge this into `test_sent_correct_message` too. What you're doing with the custom class and the patching seems overcomplicated anyway.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -491,3 +748,86 @@ async def test_preserved_other_overwrites(self):
...
+                    await self.cog._unsilence_wrapper(channel, context)
+
+                    overwrites.assert_called_once_with(role)
+                    send_message.assert_called_once_with(message, ctx.channel, channel)
+
+                    send_message.reset_mock()
+
+    @mock.patch.object(silence.Silence, "_force_voice_sync")
+    @mock.patch.object(silence.Silence, "send_message")
+    async def test_correct_overwrites(self, send_message, _):
```
MarkKoz, line 791: As with the silence test, there seems to be redundancies here. I'm not sure why you rolled your own tests instead of re-using and modifying the existing tests for text channels.

bot/exts/moderation/silence.py
```diff
@@ -92,88 +102,221 @@ async def _async_init(self) -> None:
...
     @commands.command(aliases=("hush",))
-    @lock_arg(LOCK_NAMESPACE, "ctx", attrgetter("channel"), raise_error=True)
-    async def silence(self, ctx: Context, duration: HushDurationConverter = 10) -> None:
+    @lock(LOCK_NAMESPACE, _select_lock_channel, raise_error=True)
+    async def silence(
+        self,
+        ctx: Context,
+        duration: HushDurationConverter = 10,
+        kick: bool = False,
+        *, channel: TextOrVoiceChannel = None
```
MarkKoz, line 156: Doesn't it make more sense to put the kick parameter after the channel? I understand you did it so you can consume all remaining text in case a voice channel is specified by a name containing spaces. However, I suspect most of the time the voice channel will be specified by its ID. The convenience of not having to give the default value for the `kick` parameter (which is _only_ relevant when `channel` is also specified) likely outweighs the convenience of not having to put quotes around channel names with spaces.

MarkKoz (changes requested):

bot/exts/moderation/silence.py
```diff
@@ -92,88 +102,221 @@ async def _async_init(self) -> None:
...
+    @staticmethod
+    async def _get_afk_channel(guild: Guild) -> VoiceChannel:
+        """Get a guild's AFK channel, or create one if it does not exist."""
+        afk_channel = guild.afk_channel
+
+        if afk_channel is None:
+            overwrites = {
+                guild.default_role: PermissionOverwrite(speak=False, connect=False, view_channel=False)
+            }
+            afk_channel = await guild.create_voice_channel("mute-temp", overwrites=overwrites)
```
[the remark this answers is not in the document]
HassanAbouelela: It gets added to the very top of the channel list, and is only visible to admins, and possibly mods. It shouldn't matter too much either way, as that channel should only exist for a few seconds max.

bot/exts/moderation/silence.py
HassanAbouelela: I have had that exact same line of thought, but honestly, I don't feel too strongly about either case. I'll switch the order anyway because it's a lot more likely for the kick argument to not be passed.

bot/exts/moderation/silence.py
```diff
@@ -92,88 +102,221 @@ async def _async_init(self) -> None:
...
+        log.debug(f"Unsilencing channel #{channel} from {ctx.author}'s command.")
+        await self._unsilence_wrapper(channel, ctx)
 
     @lock_arg(LOCK_NAMESPACE, "channel", raise_error=True)
-    async def _unsilence_wrapper(self, channel: TextChannel) -> None:
+    async def _unsilence_wrapper(self, channel: TextOrVoiceChannel, ctx: Optional[Context] = None) -> None:
         """Unsilence `channel` and send a success/failure message."""
+        msg_channel = channel
+        if ctx is not None:
+            msg_channel = ctx.channel
```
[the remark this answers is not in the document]
HassanAbouelela: The context is None when the wrapper is being called from the scheduler  ([Line](https://github.com/python-discord/bot/blob/b6c966608d57a28a5fbbecc7b833955ecc47b998/bot/exts/moderation/silence.py#L392)). If we want to cut down on the arguments, we'd have to settle on not sending a message when the channel gets automatically unmuted/ when unmuting fails.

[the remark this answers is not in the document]
MarkKoz: Sorry, I missed that. At the least you should still update the docstring.

bot/exts/moderation/silence.py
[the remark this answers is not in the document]
MarkKoz: That sounds fine.

bot/exts/moderation/silence.py
```diff
@@ -92,88 +112,222 @@ async def _async_init(self) -> None:
...
+        """
+        # Obtain temporary channel
+        delete_channel = channel.guild.afk_channel is None
+        afk_channel = await self._get_afk_channel(channel.guild)
+
+        try:
+            # Move all members to temporary channel and back
+            for member in channel.members:
+                # Skip staff
+                if self._helper_role in member.roles:
```
mbaruh, line 303: This should affect anyone who doesn't have explicit connect and speak permissions in the channel, as in text. For simplicity, lets assume it's anyone who isn't mod+. This should use the has_any_roles check with the moderation roles, as in other parts of the codebase.

mbaruh (dismissed):

bot/exts/moderation/silence.py
```diff
@@ -92,88 +112,222 @@ async def _async_init(self) -> None:
...
+                    await member.move_to(channel, reason="Muting VC member.")
+                    log.debug(f"Moved {member.name} to original voice channel.")
+                except Exception as e:
+                    log.debug(f"Failed to move {member.name}. Reason: {e}")
+                    continue
+
+        finally:
+            # Delete VC channel if it was created.
+            if delete_channel:
+                await afk_channel.delete(reason="Deleting temp mute channel.")
```
Akarys42, line 319: Just a small something, but I always feel like such aliases should be banned from the surface of earth, even more if they appear in the audit log, 
```suggestion
                await afk_channel.delete(reason="Deleting temporary mute channel.")
```

bot/exts/moderation/silence.py
HassanAbouelela: Using the `has_any_role_check` without passing a fake context didn't seem feasible, so I opted to use the `any(role.id in Moderation_Roles)` structure seen elsewhere in the codebase.

Come to think of it, that should become it's own helper, but that's out of scope for this PR.

tests/bot/exts/moderation/test_silence.py
HassanAbouelela: I've gone ahead and broke those out into their own tests, in part because I had to add more tests and didn't want a long list of subtests, and in part because of your suggestion.

I've also tried to apply that to the rest of my tests.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -268,7 +415,66 @@ async def test_sent_correct_message(self):
...
+            with self.subTest(target_channel=target, message=message):
+                with mock.patch.object(self.cog, "bot") as bot_mock:
+                    bot_mock.get_channel.return_value = AsyncMock()
+                    await self.cog.silence.callback(self.cog, ctx, 10, False, channel=target)
+
+                if ctx.channel == target or target is None:
+                    ctx.channel.send.assert_called_once_with(message)
+
+                else:
+                    ctx.channel.send.assert_called_once_with(message.replace("current channel", target.mention))
```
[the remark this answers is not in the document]
MarkKoz: @HassanAbouelela this still needs to be addressed.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -405,13 +610,22 @@ async def test_sent_correct_message(self):
             (False, silence.MSG_UNSILENCE_MANUAL, PermissionOverwrite(send_messages=False)),
             (False, silence.MSG_UNSILENCE_MANUAL, PermissionOverwrite(add_reactions=False)),
         )
+
         for was_unsilenced, message, overwrite in test_cases:
             ctx = MockContext()
-            with self.subTest(was_unsilenced=was_unsilenced, message=message, overwrite=overwrite):
+
+            for target in [None, MockTextChannel()]:
```
MarkKoz, line 617: You can reduce nesting by using `zip` instead of a nested loop. Do you think this will look nicer?

tests/bot/exts/moderation/test_silence.py
```diff
@@ -405,13 +610,22 @@ async def test_sent_correct_message(self):
...
+                ctx.channel.overwrites_for.return_value = overwrite
+                if target:
+                    target.overwrites_for.return_value = overwrite
```
MarkKoz, line 620: Setting overrwrites for the target is unnecessary.

tests/bot/exts/moderation/test_silence.py
MarkKoz: @HassanAbouelela This still needs to be addressed. `test_sent_correct_message` takes care of all the message stuff, so this test should absolutely not make any assertions on `send_message`. It only needs to assert the role with which `overwrites_for` was called. It doesn't even need to mock the return values of `overwrites_for` because `test_sent_correct_message` already tests for that.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -491,3 +710,168 @@ async def test_preserved_other_overwrites(self):
...
+                    send_message.assert_called_once_with(message, context.channel, channel, alert_target=True)
+
+                channel.set_permissions.assert_called_once_with(role, overwrite=overwrites)
+                if channel != ctx.channel:
+                    ctx.channel.send.assert_not_called()
+
+            await reset()
+
+
+class SendMessageTests(unittest.IsolatedAsyncioTestCase):
```
MarkKoz, line 798: All of these look great 👍

tests/bot/exts/moderation/test_silence.py
MarkKoz: @HassanAbouelela This still needs to be addressed. I really think this test should be scrapped. To test overwrites, take inspiration from (or directly modify) `test_preserved_other_overwrites`, `test_restored_overwrites`, and `test_cache_miss_used_default_overwrites`.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -287,21 +474,39 @@ async def test_skipped_already_silenced(self):
...
+    async def test_silenced_voice_channel_full(self):
+        """Channel had `speak` and `connect` permissions revoked for verified role."""
+        self.assertTrue(await self.cog._set_silence_overwrites(self.voice_channel, kick=True))
+        self.assertFalse(self.voice_overwrite.speak or self.voice_overwrite.connect)
+        self.voice_channel.set_permissions.assert_awaited_once_with(
+            self.cog._verified_voice_role,
+            overwrite=self.voice_overwrite
         )
 
     async def test_preserved_other_overwrites(self):
```
MarkKoz, line 505: Test this for voice channels too.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -268,7 +386,76 @@ async def test_sent_correct_message(self):
...
+    @voice_sync_helper
+    async def test_kick_not_called(self, ctx, sync, kick):
+        """Tests that silence command does not call kick on a text channel."""
+        channel = MockTextChannel()
+        await self.cog.silence.callback(self.cog, ctx, 10, channel, kick=True)
+
+        sync.assert_not_called()
+        kick.assert_not_called()
 
     async def test_skipped_already_silenced(self):
```
MarkKoz, line 460: Test this for voice channels too.

MarkKoz (changes requested):

tests/bot/exts/moderation/test_silence.py
```diff
@@ -268,7 +386,76 @@ async def test_sent_correct_message(self):
...
+        """Test function sends messages to the correct channels."""
+        text_channel = MockTextChannel()
+        voice_channel = MockVoiceChannel()
+        ctx = MockContext()
+
+        test_cases = (
+            (None, silence.MSG_SILENCE_SUCCESS.format(duration=10)),
+            (text_channel, silence.MSG_SILENCE_SUCCESS.format(duration=10)),
+            (voice_channel, silence.MSG_SILENCE_SUCCESS.format(duration=10)),
+            (ctx.channel, silence.MSG_SILENCE_SUCCESS.format(duration=10)),
```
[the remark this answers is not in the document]
HassanAbouelela: Test has been removed in 1b9f8ee9.

tests/bot/exts/moderation/test_silence.py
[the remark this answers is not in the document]
HassanAbouelela: Thanks for the suggestion, addressed in 1b9f8ee9.

tests/bot/exts/moderation/test_silence.py
HassanAbouelela: Yeah, this test is quite redundant. I cleaned it up in aef979e6.

HassanAbouelela:
Thanks for the last review round. I'm slowly working through everything, but I'll rerequest review when the branch is ready.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -159,6 +169,100 @@ async def test_cog_check(self, role_check):
...
+    def create_erroneous_members() -> Tuple[List[MockMember], List[MockMember]]:
+        """
+        Helper method to generate a list of members that error out on move_to call.
+
+        Returns the list of erroneous members,
+        as well as a list of regular and erroneous members combined, in that order.
+        """
+        def failing_move_to(*_):
+            raise Exception()
+        erroneous_members = [MockMember(move_to=Mock(failing_move_to)) for _ in range(5)]
```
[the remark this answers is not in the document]
HassanAbouelela: I knew there must've been a better way haha. Thank you. Changed in b867fb21.

tests/bot/exts/moderation/test_silence.py
HassanAbouelela: When the target channel is set, the unsilence wrapper uses its permissions to determine behavior. Without setting the overwrites, the error message changes from `MSG_UNSILENCE_FAIL` to `MSG_UNSILENCE_MANUAL`.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -159,6 +169,100 @@ async def test_cog_check(self, role_check):
...
+
+    async def test_kick_move_to_error(self):
+        """Test to ensure move_to gets called on all members during kick, even if some fail."""
+        await self.cog._async_init()
+        failing_members, members = self.create_erroneous_members()
+
+        await self.cog._kick_voice_members(MockVoiceChannel(members=members))
+        for member in members:
+            member.move_to.assert_called_once()
+            member.reset_mock()
```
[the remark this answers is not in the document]
HassanAbouelela: Fixed in b0381d71.

tests/bot/exts/moderation/test_silence.py
```diff
@@ -159,6 +169,100 @@ async def test_cog_check(self, role_check):
...
+
+        Returns the list of erroneous members,
+        as well as a list of regular and erroneous members combined, in that order.
+        """
+        def failing_move_to(*_):
+            raise Exception()
+        erroneous_members = [MockMember(move_to=Mock(failing_move_to)) for _ in range(5)]
+
+        members = []
+        for i in range(5):
```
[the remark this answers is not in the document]
HassanAbouelela: Resolved in 2446c172 (along with all other instances of this).

tests/bot/exts/moderation/test_silence.py
```diff
@@ -268,7 +386,76 @@ async def test_sent_correct_message(self):
...
+                target.send.reset_mock()
+
+    @voice_sync_helper
+    async def test_sync_called(self, ctx, sync, kick):
+        """Tests if silence command calls sync on a voice channel."""
+        channel = MockVoiceChannel()
+        await self.cog.silence.callback(self.cog, ctx, 10, channel, kick=False)
+
+        sync.assert_called_once_with(self.cog, channel)
+        kick.assert_not_called()
```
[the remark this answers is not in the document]
HassanAbouelela: Resolved in da694fde.

tests/bot/exts/moderation/test_silence.py
HassanAbouelela: I believe I've duplicated all relevant tests now. db5b0751.

tests/bot/exts/moderation/test_silence.py
HassanAbouelela: Resolved in db5b0751.

tests/bot/exts/moderation/test_silence.py
HassanAbouelela: Added in bac68d1d.

tests/bot/exts/moderation/test_silence.py
HassanAbouelela: The challenge with that is we need to use two items from the target list, for each item of the test_cases. I'm struggling to see how I could solve that with zip. One solution would be to just double the test_cases list, but that would mean the test runner is completely dependent on there being an odd number of cases.

tests/bot/exts/moderation/test_silence.py
HassanAbouelela: This has been addressed through various other comments, and I believe all the existing text tests have a voice equivalent where relevant.

HassanAbouelela:
@MarkKoz it's been a while, but I believe I've worked through all the remaining review comments. There are a lot though, so forgive me if anything got missed.

tests/bot/exts/moderation/test_silence.py
MarkKoz: You're right, `zip` won't work. What's needed is a Cartesian product (`itertools.product`). Up to you if you want to make that change.

tests/bot/exts/moderation/test_silence.py
MarkKoz: Looks like you are correct. You're talking about this, right? https://github.com/python-discord/bot/pull/1297/files#diff-f64697ccedd0a53137e360dba5f9f39c7f7a7b2fa76e29153024d1eb7a1d51c0R228

HassanAbouelela: Yeah that's the one. The if statement right below it switches the message.

tests/bot/exts/moderation/test_silence.py
HassanAbouelela: That does help with the nesting, thanks. 27223522

MarkKoz (dismissed):

bot/exts/moderation/silence.py
```diff
@@ -92,88 +113,230 @@ async def _async_init(self) -> None:
...
-            overwrite = channel.overwrites_for(self._everyone_role)
-            if overwrite.send_messages is False or overwrite.add_reactions is False:
-                await channel.send(MSG_UNSILENCE_MANUAL)
+            if isinstance(channel, VoiceChannel):
+                overwrite = channel.overwrites_for(self._verified_voice_role)
+                manual = overwrite.speak is False
             else:
-                await channel.send(MSG_UNSILENCE_FAIL)
+                overwrite = channel.overwrites_for(self._everyone_role)
+                manual = overwrite.send_messages is False or overwrite.add_reactions is False
```
Numerlor, line 228: The `manual` name didn't really explain the var's purpose for me, not sure what would be the most fitting, maybe something like `has_manual_overwrites`

bot/exts/moderation/silence.py
```diff
@@ -92,88 +113,230 @@ async def _async_init(self) -> None:
...
+    async def _get_afk_channel(guild: Guild) -> VoiceChannel:
+        """Get a guild's AFK channel, or create one if it does not exist."""
+        afk_channel = guild.afk_channel
+
+        if afk_channel is None:
+            overwrites = {
+                guild.default_role: PermissionOverwrite(speak=False, connect=False, view_channel=False)
+            }
+            afk_channel = await guild.create_voice_channel("mute-temp", overwrites=overwrites)
+            log.info(f"Failed to get afk-channel, created temporary channel #{afk_channel} ({afk_channel.id})")
```
Numerlor, line 275: Would this log be more fitting around the uses of this method? The method itself doesn't enforce it being temporary

bot/exts/moderation/silence.py
```diff
@@ -92,88 +113,230 @@ async def _async_init(self) -> None:
...
+        log.debug(f"Removing all non staff members from #{channel.name} ({channel.id}).")
+
+        for member in channel.members:
+            # Skip staff
+            if any(role.id in constants.MODERATION_ROLES for role in member.roles):
+                continue
+
+            try:
+                await member.move_to(None, reason="Kicking member from voice channel.")
+                log.debug(f"Kicked {member.name} from voice channel.")
```
Numerlor, line 291: Logging all of the members that were kicked doesn't seem too important, a trace level log feels more appropriate

bot/exts/moderation/silence.py
```diff
@@ -92,88 +113,230 @@ async def _async_init(self) -> None:
...
+                    log.debug(f"Moved {member.name} to original voice channel.")
+                except Exception as e:
+                    log.debug(f"Failed to move {member.name}. Reason: {e}")
+                    continue
+
+        finally:
+            # Delete VC channel if it was created.
+            if delete_channel:
+                await afk_channel.delete(reason="Deleting temporary mute channel.")
+
```
Numerlor, line 329: Could the util methods above be moved near end of the cog? 
Most of the methods related to the silencing/unsilencing itself are above them except for `_unsilence` and `_schedule_unsilence`; I believe it'd be clearer if they were grouped

bot/exts/moderation/silence.py
```diff
@@ -92,88 +113,230 @@ async def _async_init(self) -> None:
...
+        self.notifier = SilenceNotifier(self.bot.get_channel(constants.Channels.mod_log))
         await self._reschedule()
 
+    async def send_message(
+        self,
+        message: str,
+        source_channel: TextChannel,
+        target_channel: TextOrVoiceChannel,
+        *, alert_target: bool = False
```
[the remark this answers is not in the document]
HassanAbouelela: I missed that, thanks. c896414b

bot/exts/moderation/silence.py
```diff
@@ -92,88 +113,230 @@ async def _async_init(self) -> None:
...
     @commands.command(aliases=("hush",))
-    @lock_arg(LOCK_NAMESPACE, "ctx", attrgetter("channel"), raise_error=True)
-    async def silence(self, ctx: Context, duration: HushDurationConverter = 10) -> None:
+    @lock(LOCK_NAMESPACE, _select_lock_channel, raise_error=True)
+    async def silence(
+        self,
+        ctx: Context,
+        duration: HushDurationConverter = 10,
+        channel: TextOrVoiceChannel = None,
+        *, kick: bool = False
```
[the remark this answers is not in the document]
HassanAbouelela: c896414b

bot/exts/moderation/silence.py
HassanAbouelela: This method was only intended to be used within the scope of this cog, so I hadn't considered any problems with that log. I see what you mean though, but it would add useless complexity if we had to convey to the caller that a channel was created. I opted instead to remove the temporary part so the function is not assuming intentions. d9ff7575

bot/exts/moderation/silence.py
HassanAbouelela: Changed in d9ff7575.

bot/exts/moderation/silence.py
HassanAbouelela: I moved the functions to be closer to their callers. 0253f8f6

bot/exts/moderation/silence.py
HassanAbouelela: I'm not sure what to call it either. Manual implies there's an automatic overwrite, but that's not really the case. A channel either follows the category's overwritten, or it is manually overwritten. I dropped the overwrite part for `has_channel_overwrite`, but I'm happy to change it again. 52bfbf31

bot/exts/moderation/silence.py
```diff
@@ -92,88 +113,230 @@ async def _async_init(self) -> None:
...
+            source_reply = source_reply.replace("current channel", target_channel.mention)
+        await source_channel.send(source_reply)
+
+        # Reply to target channel
+        if alert_target:
+            if isinstance(target_channel, VoiceChannel):
+                voice_chat = self.bot.get_channel(VOICE_CHANNELS.get(target_channel.id))
+                if voice_chat and source_channel != voice_chat:
+                    await voice_chat.send(message.replace("current channel", target_channel.mention))
+
```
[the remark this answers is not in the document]
HassanAbouelela: Changed in 73d120e.

HassanAbouelela:
I'm not sure how good an idea this is when we're this far along the development process, but I'm thinking of changing up the interface for the silence command. When testing recently, I found it very clunky, especially compared to the other moderation tools. Most commands (ban, mute, etc) have the format: `!<command> <target> <duration> <other args>`, compared to silence which has `<duration> <target>`.

This was originally done because the current usage of silence is either `!silence` or `!silence duration`, and I wanted to minimize the changes the mods experience. Because this is a mod-only command, I'll make this same suggestion internally, but I think the mods will find it most useful to have the `<target> <duration>` format.

Another idea I had was making the second argument accept both duration and target. This should be doable by having the argument type as `[TextChannel, VoiceChannel, int]`.

HassanAbouelela:
I put it up to a vote with the mods, and most people were leaning towards making the first argument accept either the duration or the channel, so I did that. From my testing, the converters are able to properly choose between a channel, and an int, so not much work was needed.

The subject of the `!shh` command did come up during discussions. One concern raised was the disparity between the interface of two very similar commands. I'll try adding an optional channel argument, but I'm not sure how that'll work out yet.

HassanAbouelela:
It took some figuring out, but now both changes have been made, and the PR is ready for review again.

kosayoda (changes requested):
Almost everything I've tried works to my expectation (almost merged it :x), but `|silence forever` only silences the channel for the fallback 10 minutes. I suspect this has something to do with dpy's parser, would you mind looking into it?

HassanAbouelela:
@kosayoda thanks for your review, good catch! I missed adding in the behavior for that argument when redesigning the first argument, but I've fixed that in f92338ef.

kosayoda (approved):
🚂

Xithrius (approved):
Tested it with author of this PR, it works very well!

PR #1302, opened 2020-11-24
Adding <prefix>stream and <prefix>revokestream commands

Harbys:
These commands would help in moderation by allowing mods to give temporary screen sharing permissions with one command
<prefix>stream <user mention> {duration default 1} {time unit default hour} - gives user video role for specified duration
<prefix>revokestream <user mention> - takes away video role


Harbys:
ok forgot to fix flake8 issues

PR #1303, opened 2020-11-24
Adding <prefix>stream and <prefix>revokestream commands

Harbys:
These commands would help in moderation by allowing mods to give temporary screen sharing permissions with one command
stream {duration default 1} {time unit default hour} - gives user video role for specified duration
revokestream - takes away video role


bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+        argument user - required user mention, any errors should be handled by upper level handler
+        duration - int must be higher than 0 - defaults to 1
+        time_format - str defining what time unit you want to use, must be any of FORMATS - defaults to h
+        Command give user permission to stream and takes it away after provided duration
+        """
+        # Check for required user argument
+        # if not provided send NO_USER_SPECIFIED message
+        if not user:
+            await ctx.send(NO_USER_SPECIFIED)
+            return
```
MarkKoz, line 72: If you want to enforce the user argument then don't define it as optional. The default error message for a missing argument will suffice in that case.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+        return duration * self._link_from_alias(time_format)[0]["mul"]
+
+    @commands.command(aliases=("streaming", "share"))
+    @commands.has_any_role(*STAFF_ROLES)
+    async def stream(
+            self,
+            ctx: commands.Context,
+            user: discord.Member = None,
+            duration: int = 1,
+            time_format: str = "h",
```
MarkKoz, line 57: Use the extant [`Duration`](https://github.com/python-discord/bot/blob/master/bot/converters.py#L301) converter instead of rolling your own. In fact, it'd be even better to use `Expiry` defined there too. This will support specifying exact timestamps too and make it consistent with the infraction commands.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+        """Get TIME_FORMATS key and entry by time format or any of its aliases."""
+        for format_key, val in TIME_FORMATS.items():
+            if format_key == time_format or time_format in val["aliases"]:
+                return TIME_FORMATS[format_key], format_key
+
+    def _parse_time_to_seconds(self, duration: int, time_format: str) -> int:
+        """Get time in seconds from duration and time format."""
+        return duration * self._link_from_alias(time_format)[0]["mul"]
+
+    @commands.command(aliases=("streaming", "share"))
```
MarkKoz, line 50: I think "share" is too generic to be used as an alias.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+    @tasks.loop(seconds=30)
+    async def remove_permissions(self) -> None:
+        """Background loop for removing streaming permission."""
+        all_entries = await self.user_cache.items()
+        for user_id, delete_time in all_entries:
+            if time.time() > delete_time:
+                member = self.guild_static.fetch_memebr(user_id)
+                if member:
+                    await member.remove_roles(discord.Object(Roles.video), reason="Temporary streaming access revoked")
+                    await self.user_cache.pop(user_id)
```
MarkKoz, line 105: Use a [`Scheduler`](https://github.com/python-discord/bot/blob/master/bot/utils/scheduling.py#L60-L82) to schedule removal of the role at a precise time rather than a task that polls.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+        """Wait for bot to be ready before starting remove_permissions loop and get guild by id."""
+        await self.bot.wait_until_ready()
+        self.guild_static = self.bot.get_guild(Guild.id)
+
+    @commands.command(aliases=("unstream", ))
+    @commands.has_any_role(*STAFF_ROLES)
+    async def revokestream(
+            self,
+            ctx: commands.Context,
+            user: discord.Member = None
```
MarkKoz, line 118: This shouldn't have a default value. It's should always be a required argument.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+    ) -> None:
+        """
+        Revokestream handles <prefix>revokestream command.
+
+        argument user - required user mention, any errors should be handled by upper level handler
+        command removes streaming permission from a user
+        """
```
MarkKoz, line 125: The description of the function should always be in the summary section i.e. the first line. Implementation details like error handling need not be documented for something user-facing like a command. The documentation of the `user` argument could be integrated into the summary as it is fairly self-explanatory.

```suggestion
        """Revoke streaming permissions from a user."""
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+            *_
+    ) -> None:
+        """
+        Stream handles <prefix>stream command.
+
+        argument user - required user mention, any errors should be handled by upper level handler
+        duration - int must be higher than 0 - defaults to 1
+        time_format - str defining what time unit you want to use, must be any of FORMATS - defaults to h
+        Command give user permission to stream and takes it away after provided duration
+        """
```
MarkKoz, line 67: In a similar vein to the other docstring comment I left, this one could be improved. I also included the boilerplate used in other docstrings that use the `Expiry` converter, since I'm assuming you're going to make that change too.

```suggestion
        """
        Temporarily grant streaming permissions to a user for a given duration.

        A unit of time should be appended to the duration.
        Units (∗case-sensitive):
        \u2003`y` - years
        \u2003`m` - months∗
        \u2003`w` - weeks
        \u2003`d` - days
        \u2003`h` - hours
        \u2003`M` - minutes∗
        \u2003`s` - seconds
        Alternatively, an ISO 8601 timestamp can be provided for the duration.
        """
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+    ) -> None:
+        """
+        Revokestream handles <prefix>revokestream command.
+
+        argument user - required user mention, any errors should be handled by upper level handler
+        command removes streaming permission from a user
+        """
+        not_allowed = not any(Roles.video == role.id for role in user.roles)
+        if not_allowed:
+            await user.remove_roles(discord.Object(Roles.video))
```
MarkKoz, line 128: There should be a confirmation message sent when this is successful.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+
+from bot.bot import Bot
+from bot.constants import Guild, Roles, STAFF_ROLES, TIME_FORMATS
+
+# Constant error messages
+NO_USER_SPECIFIED = "Please specify a user"
+TIME_FORMAT_NOT_VALID = "Please specify a valid time format ex. 10h or 1day"
+TIME_LESS_EQ_0 = "Duration can not be a 0 or lower"
+USER_ALREADY_ALLOWED_TO_STREAM = "This user can already stream"
+USER_ALREADY_NOT_ALLOWED_TO_STREAM = "This user already can't stream"
```
MarkKoz, line 15: To make it more clear that these are errors, prefix them with `bot.constants.Emoji.cross_mark`. This is a common practice in the bot. Similarly, success messages should be prefixed with `bot.constants.Emoji.check_mark`.

MarkKoz, line 15: These need to end in full stops.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+# FORMATS holds a combined list of all allowed time units
+# made from TIME_FORMATS constant
+FORMATS = []
+for key, entry in TIME_FORMATS.items():
+    FORMATS.extend(entry["aliases"])
+    FORMATS.append(key)
+
+
+class Stream(commands.Cog):
+    """Stream class handles giving screen sharing permission with commands."""
```
MarkKoz, line 27: I find it redundant to include the class's name in the docstring. I also don't think "class" should be included since 1. it's obvious it's a class, its declaration is one line above and 2. this docstring is user-facing via the help command. Here is a suggestion that's more in line with other cog docstrings:

```suggestion
    """Grant and revoke streaming permissions from users."""
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+        await user.add_roles(discord.Object(Roles.video), reason="Temporary streaming access granted")
+        await ctx.send(f"{user.mention} can now stream for {duration} {self._link_from_alias(time_format)[1]}/s")
+
+    @tasks.loop(seconds=30)
+    async def remove_permissions(self) -> None:
+        """Background loop for removing streaming permission."""
+        all_entries = await self.user_cache.items()
+        for user_id, delete_time in all_entries:
+            if time.time() > delete_time:
+                member = self.guild_static.fetch_memebr(user_id)
```
MarkKoz, line 102: This function is misspelled. Even though I am suggested to replace all this code with a `Scheduler`, it makes me worried for two reasons:

1. It indicates you haven't thoroughly tested this manually.
2. The tests are still passing despite this error.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,135 @@
...
+        # Check if user already has streaming permission
+        already_allowed = any(Roles.video == role.id for role in user.roles)
+        if already_allowed:
+            await ctx.send(USER_ALREADY_ALLOWED_TO_STREAM)
+            return
+
+        # Set user id - time in redis cache and add streaming permission role
+        await self.user_cache.set(user.id, time.time() + self._parse_time_to_seconds(duration, time_format))
+        await user.add_roles(discord.Object(Roles.video), reason="Temporary streaming access granted")
+        await ctx.send(f"{user.mention} can now stream for {duration} {self._link_from_alias(time_format)[1]}/s")
```
MarkKoz, line 94: As mentioned earlier, this should use the check mark emoji as a prefix to clearly communicate that the operation was successful.

MarkKoz: Please end sentences in full stops.

MarkKoz (changes requested):

Harbys:
I think I fixed all pointed out issues and implemented requested changes. Thx for pointing out the Scheduler as it actually made it much simpler

mbaruh:
Hi, thank you for contributing!
The thing with changes to moderation features, is that they need follow internal discussions to decide on whether we want it, and how we would use it - which in turn affects the implementation.

That is not to say that you can't PR such a feature, but rather that the right course of action here would be to open an issue first to allow us to discuss it properly, and to prevent a case where you would do a lot of work that would end up not being used.

I will put this PR on hold for now until we can discuss it properly among the mods.

cloudyti (approved):

ks129:
@Harbys Mods and admins team have made decision that we should have this feature, so I take this out from stalled status and you can continue your work. But please read #1343 issue, because implementation should be little bit different.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,86 @@
...
+        self.scheduler = Scheduler(self.__class__.__name__)
+
+    @staticmethod
+    async def _remove_streaming_permission(schedule_user: discord.Member) -> None:
+        """Remove streaming permission from Member."""
+        await schedule_user.remove_roles(discord.Object(Roles.video), reason="Temporary streaming access revoked")
+
+    @commands.command(aliases=("streaming",))
+    @commands.has_any_role(*STAFF_ROLES)
+    async def stream(
```
MrHemlock, line 31: We'll want to add a default duration for the command.  We're thinking 30 minutes would be a good one.  Along with that, we'll need that default added to the config.yaml and the constants file.

MrHemlock (changes requested):
Just a couple tweaks and I think we'll be golden.  I appreciate all the hard work you're doing!

config-default.yml
```diff
@@ -254,6 +254,9 @@ guild:
         jammers:        737249140966162473
         team_leaders:   737250302834638889
 
+        # Streaming
+        video: 764245844798079016
```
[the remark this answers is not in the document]
Harbys: done in commit [a50239e6fe46e2da36f925967db21c95f940597a](https://github.com/python-discord/bot/pull/1303/commits/a50239e6fe46e2da36f925967db21c95f940597a)

bot/exts/moderation/stream.py
Harbys: done in commit [074ce91205539ca06c1c048c62b7e649c4ae78b5](https://github.com/python-discord/bot/pull/1303/commits/074ce91205539ca06c1c048c62b7e649c4ae78b5) and [a0e2179da398a9164582e7debb7e139754434385](https://github.com/python-discord/bot/pull/1303/commits/a0e2179da398a9164582e7debb7e139754434385)

Harbys:
@MrHemlock I think it's all done as requested

MrHemlock (approved):

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,89 @@
...
+            self,
+            ctx: commands.Context,
+            user: discord.Member,
+            duration: Expiry =
+            datetime.datetime.utcnow() + datetime.timedelta(minutes=VideoPermission.default_permission_duration),
```
MarkKoz, line 38: Default values are only computed once. That happens when the function is defined. Therefore, `now()` will not actually be the time at which the command is invoked, but the time at which the function was defined. The only practical solution is to calculate this within the function rather than as a default value.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,89 @@
...
+        A unit of time should be appended to the duration.
+        Units (∗case-sensitive):
+        \u2003`y` - years
+        \u2003`m` - months∗
+        \u2003`w` - weeks
+        \u2003`d` - days
+        \u2003`h` - hours
+        \u2003`M` - minutes∗
+        \u2003`s` - seconds
+        Alternatively, an ISO 8601 timestamp can be provided for the duration.
```
MarkKoz, line 53: Add a newline for readability.
```suggestion

        Alternatively, an ISO 8601 timestamp can be provided for the duration.
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,89 @@
...
+
+from bot.bot import Bot
+from bot.constants import Emojis, Roles, STAFF_ROLES, VideoPermission
+from bot.converters import Expiry
+from bot.utils.scheduling import Scheduler
+from bot.utils.time import format_infraction_with_duration
+
+# Constant error messages
+TIME_FORMAT_NOT_VALID = f"{Emojis.cross_mark}Please specify a valid time format ex. 10h or 1day."
+TIME_LESS_EQ_0 = f"{Emojis.cross_mark}Duration can not be a 0 or lower."
```
MarkKoz, line 14: They're no longer used anywhere.
```suggestion
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,89 @@
...
+USER_ALREADY_ALLOWED_TO_STREAM = f"{Emojis.cross_mark}This user can already stream."
+USER_ALREADY_NOT_ALLOWED_TO_STREAM = f"{Emojis.cross_mark}This user already can't stream."
```
MarkKoz, line 16: Add a space between the emoji and the text.
```suggestion
USER_ALREADY_ALLOWED_TO_STREAM = f"{Emojis.cross_mark} This user can already stream."
USER_ALREADY_NOT_ALLOWED_TO_STREAM = f"{Emojis.cross_mark} This user already can't stream."
```

MarkKoz, line 16: Why are these two define as constants? They're each only used in one place. The only benefit I'd see of defining messages in variables like this is for testing purpose, yet your tests do not rely on these either.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,89 @@
...
+    ) -> None:
+        """Take away streaming permission from a user."""
+        # Check if user has the streaming permission to begin with
+        allowed = any(Roles.video == role.id for role in user.roles)
+        if allowed:
+            # Cancel scheduled task to take away streaming permission to avoid errors
+            if user.id in self.scheduler:
+                self.scheduler.cancel(user.id)
+            await user.remove_roles(discord.Object(Roles.video))
+            await ctx.send(f"{Emojis.check_mark}Streaming permission taken from {user.display_name}")
```
MarkKoz, line 82:
```suggestion
            await ctx.send(f"{Emojis.check_mark} Streaming permission taken from {user.display_name}.")
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,89 @@
...
+        already_allowed = any(Roles.video == role.id for role in user.roles)
+        if already_allowed:
+            await ctx.send(USER_ALREADY_ALLOWED_TO_STREAM)
+            return
+
+        # Schedule task to remove streaming permission from Member
+        self.scheduler.schedule_at(duration, user.id, self._remove_streaming_permission(user))
+        await user.add_roles(discord.Object(Roles.video), reason="Temporary streaming access granted")
+        await ctx.send(f"{Emojis.check_mark}{user.mention} can now stream until "
+                       f"{format_infraction_with_duration(str(duration))}.")
```
MarkKoz, line 65: This doesn't follow our style guide. In any case, it'd look better if the duration was formatted on a separate line. It also needs a space between the text and the emoji.

```suggestion
        duration = format_infraction_with_duration(str(duration))
        await ctx.send(f"{Emojis.check_mark} {user.mention} can now stream until {duration}.")
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,89 @@
...
+USER_ALREADY_ALLOWED_TO_STREAM = f"{Emojis.cross_mark}This user can already stream."
+USER_ALREADY_NOT_ALLOWED_TO_STREAM = f"{Emojis.cross_mark}This user already can't stream."
+
+
+class Stream(commands.Cog):
+    """Grant and revoke streaming permissions from users."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.scheduler = Scheduler(self.__class__.__name__)
```
MarkKoz, line 24: You need to define a `cog_unload` which will cancel all the tasks.

tests/bot/exts/moderation/test_stream.py
```diff
@@ -0,0 +1,25 @@
...
+from tests.helpers import MockBot, MockMember, MockRole
+
+
+class StreamCommandTest(unittest.IsolatedAsyncioTestCase):
+
+    def setUp(self) -> None:
+        self.bot = MockBot()
+        self.cog = Stream(self.bot)
+
+    def test_checking_if_user_has_streaming_permission(self):
```
MarkKoz, line 15: This test doesn't call into any code in the cog, so I don't see the point in having it.

MarkKoz (changes requested):
You need to cache the users in redis and re-schedule tasks to revoke permissions when the bot restarts or the cog reloads (cause you'll cancel tasks when the cog unloads). If you need some examples of how to use redis, look at the silence command or the help channel system. I can't think of any simpler examples. Let someone know if you still need help understanding redis.

Xithrius:
@Harbys are you planning on continuing this PR?

I'd like to see these changes get implemented.

Harbys:
@Xithrius Yes, I want to finish this pr, but I have very little spare time recently so I'm sorry it will be probably another week until I get it done.

Xithrius:
> @Xithrius Yes, I want to finish this pr, but I have very little spare time recently so, I'm sorry it will be probably another week until I get it done.

Take your time, I'm just making sure this will eventually get done.

Have a good one! :heart:

bot/exts/moderation/stream.py
Harbys: Had to remove it cause Flake8 is supper unhappy about this
D202 No blank lines allowed after function docstring

Harbys:
still need to do unit test, but I think that the cog is good for review

bot/exts/moderation/stream.py
Harbys: Now it default to None and calculates time if duration is None

bot/exts/moderation/stream.py
Harbys: done

bot/exts/moderation/stream.py
Harbys: gotten rid of these

bot/exts/moderation/stream.py
Harbys: Done.

MrHemlock (approved):
Everything looks good.  Went through and tested it on my end.  Was able to remove the role, let it naturally be removed by the time set, as well as restart the bot in the middle of it having someone in the timeout list and then having it sucessfully remove the role.  

I'm more than happy to see this go forward.

Pipfile.lock
```diff
@@ -231,10 +232,10 @@
         },
         "fakeredis": {
             "hashes": [
-                "sha256:8070b7fce16f828beaef2c757a4354af91698685d5232404f1aeeb233529c7a5",
-                "sha256:f8c8ea764d7b6fd801e7f5486e3edd32ca991d506186f1923a01fc072e33c271"
+                "sha256:01cb47d2286825a171fb49c0e445b1fa9307087e07cbb3d027ea10dbff108b6a",
```
MarkKoz, line 235: There weren't any changes to `Pipfile` so relocking is unnecessary. Please undo these changes to the lock file.

tests/bot/exts/moderation/test_stream.py
MarkKoz: @Harbys This still needs to be addressed. You either need to fix this test so it actually tests the cog, or decide that you don't want to work on tests (which is fine; we don't require tests) and remove this test file entirely.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+        await self.task_cache.set(user_id, timestamp)
+
+    async def _reload_tasks_from_redis(self) -> None:
+        await self.bot.wait_until_guild_available()
+        items = await self.task_cache.items()
+        for key, value in items:
+            member = await self.bot.get_guild(Guild.id).fetch_member(key)
+            self.scheduler.schedule_at(datetime.datetime.utcfromtimestamp(value),
+                                       key,
+                                       self._remove_streaming_permission(member))
```
MarkKoz, line 42: This doesn't follow our style guide.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+        self.reload_task = self.bot.loop.create_task(self._reload_tasks_from_redis())
+
+    async def _remove_streaming_permission(self, schedule_user: discord.Member) -> None:
+        """Remove streaming permission from Member."""
+        await self._delete_from_redis(schedule_user.id)
+        await schedule_user.remove_roles(discord.Object(Roles.video), reason="Streaming access revoked")
+
+    async def _add_to_redis_cache(self, user_id: int, timestamp: float) -> None:
+        """Adds 'task' to redis cache."""
+        await self.task_cache.set(user_id, timestamp)
```
MarkKoz, line 33: This function is redundant. All it does is forward arguments to another function without changing anything.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+        await self.bot.wait_until_guild_available()
+        items = await self.task_cache.items()
+        for key, value in items:
+            member = await self.bot.get_guild(Guild.id).fetch_member(key)
+            self.scheduler.schedule_at(datetime.datetime.utcfromtimestamp(value),
+                                       key,
+                                       self._remove_streaming_permission(member))
+
+    async def _delete_from_redis(self, key: str) -> None:
+        await self.task_cache.delete(key)
```
MarkKoz, line 45: This function is also redundant.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+    # Stores tasks to remove streaming permission
+    # User id : timestamp relation
+    task_cache = RedisCache()
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.scheduler = Scheduler(self.__class__.__name__)
+        self.reload_task = self.bot.loop.create_task(self._reload_tasks_from_redis())
+
+    async def _remove_streaming_permission(self, schedule_user: discord.Member) -> None:
```
MarkKoz, line 26: Just name the argument `member`.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @commands.command(aliases=("pstream",))
+    @commands.has_any_role(*STAFF_ROLES)
+    async def permanentstream(
+            self,
+            ctx: commands.Context,
+            user: discord.Member,
+            *_
+    ) -> None:
+        """Permanently give user a streaming permission."""
```
MarkKoz, line 97: I find "streaming permission" to be an awkward term.
```suggestion
        """Permanently grant a user the permission to stream."""
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+        await ctx.send(f"{Emojis.check_mark} {user.mention} can now stream forever")
+
+    @commands.command(aliases=("unstream", ))
+    @commands.has_any_role(*STAFF_ROLES)
+    async def revokestream(
+            self,
+            ctx: commands.Context,
+            user: discord.Member
+    ) -> None:
+        """Take away streaming permission from a user."""
```
MarkKoz, line 119:
```suggestion
        """Revoke the permissiont to stream from a user."""
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+            *_
+    ) -> None:
+        """Permanently give user a streaming permission."""
+        # Check if user already has streaming permission
+        already_allowed = any(Roles.video == role.id for role in user.roles)
+        if already_allowed:
+            if user.id in self.scheduler:
+                self.scheduler.cancel(user.id)
+                await self._delete_from_redis(user.id)
+                await ctx.send(f"{Emojis.check_mark} Moved temporary permission to permanent")
```
MarkKoz, line 104:
```suggestion
                await ctx.send(f"{Emojis.check_mark} Changed temporary permission to permanent.")
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+                return
+            await ctx.send(f"{Emojis.cross_mark} This user can already stream.")
+            return
+
+        await user.add_roles(discord.Object(Roles.video), reason="Permanent streaming access granted")
+        await ctx.send(f"{Emojis.check_mark} {user.mention} can now stream forever")
```
MarkKoz, line 110:
```suggestion
        await ctx.send(f"{Emojis.check_mark} Permanently granted {user.mention} the permission to stream.")
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+    ) -> None:
+        """Take away streaming permission from a user."""
+        # Check if user has the streaming permission to begin with
+        allowed = any(Roles.video == role.id for role in user.roles)
+        if allowed:
+            # Cancel scheduled task to take away streaming permission to avoid errors
+            if user.id in self.scheduler:
+                self.scheduler.cancel(user.id)
+            await self._remove_streaming_permission(user)
+            await ctx.send(f"{Emojis.check_mark} Streaming permission taken from {user.display_name}.")
```
MarkKoz, line 127: Revoked is a nicer word. Also, why does this use the display name rather than the mention? Whichever it is, it should be consistent.
```suggestion
            await ctx.send(f"{Emojis.check_mark} Revoked the permission to stream from {user.mention}.")
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+
+    async def _add_to_redis_cache(self, user_id: int, timestamp: float) -> None:
+        """Adds 'task' to redis cache."""
+        await self.task_cache.set(user_id, timestamp)
+
+    async def _reload_tasks_from_redis(self) -> None:
+        await self.bot.wait_until_guild_available()
+        items = await self.task_cache.items()
+        for key, value in items:
+            member = await self.bot.get_guild(Guild.id).fetch_member(key)
```
MarkKoz, line 39: You should try to use `get_member` before resorting to using an API call.

Also, you should catch the exception this may raise. Separate the handling of the member not being found from other network issues. If the member isn't found, that can just be logged at debug level and ignored. If there is a different exception, then that should be logged with `log.exception`. Note that member not found has a code of 10007, which is not the same at `NotFound` / an HTTP 404 status.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.scheduler = Scheduler(self.__class__.__name__)
+        self.reload_task = self.bot.loop.create_task(self._reload_tasks_from_redis())
+
+    async def _remove_streaming_permission(self, schedule_user: discord.Member) -> None:
+        """Remove streaming permission from Member."""
+        await self._delete_from_redis(schedule_user.id)
+        await schedule_user.remove_roles(discord.Object(Roles.video), reason="Streaming access revoked")
```
MarkKoz, line 29: What does Discord do if the permission is revoked while the user is streaming? Does it kick them out or let them keep streaming? What kind of behaviour would we want? @MrHemlock

MarkKoz (changes requested):

bot/exts/moderation/stream.py
MrHemlock: @MarkKoz It does not kick them out.  We would have to boot them from the voice channel.  Currently, I'd prefer if we don't implement anything for it at the moment, as Mods+ do have permissions to boot someone from VC manually.  

Typically if it came down to that, either they'd be receiving a voice ban or some other full infraction, which would bump them off call anyway.  Even a regular server mute does so.

MrHemlock: In my opinion, the use case for revoke would be if someone finishes up their help session or what have you before it would expire, rather than having to remove it due to bad behavior.

Xithrius:
@Harbys will you be completing this PR? If not, it can be handed off to someone else if you don't have enough time.

Harbys:
> 
> 
> @Harbys will you be completing this PR? If not, it can be handed off to someone else if you don't have enough time.

It would probably be better if you hand it off to someone else

Xithrius:
Thank you for the quick response.

MarkKoz:
Continued in #1493

PR #1308, opened 2020-11-27
Add `build-tools` tag

Mushinako:
Infoblock tag helping people install Microsoft Visual C++ Build Tools on Windows.

Special thanks to those who helped proof-reading this on Discord.


mbaruh (approved):

Akarys42:
Hello, thanks for your contribution! The tag looks pretty good, although I'd rather prefer have it named something like `microsoft-build-tools` (which should still make `build-tools` match) to communicate that it is a Windows/Microsoft specific thing.

Mushinako:
> Hello, thanks for your contribution! The tag looks pretty good, although I'd rather prefer have it named something like `microsoft-build-tools` (which should still make `build-tools` match) to communicate that it is a Windows/Microsoft specific thing.

Thanks for the suggestion! Just made the change 54540bf9dcf88f7d3e2e0f389a2127456d6c877f

Akarys42 (approved):
Thank you for your contribution!

PR #1310, opened 2020-11-28
General moderation improvements

### LINKED ISSUE #1278: Short aliases for infraction editing
At the moment, !infr edit last or !infr edit l is the fastest way to amend a previous infraction. Similar to #1170, but more limited in scope. Something like `!i e l` would be nice, if doable, if not `!infr e l` or even `!iel`.

mbaruh:
This PR implements some small but useful changes which were discussed internally:

* Removes the unnecessary and undesired `shadowkick` and `shadowmute` infractions.
* Gives mutes a one hour default duration.
* Adds the `i` alias to the infraction command group, and the `e` alias to the `edit` infraction subcommand, which closes #1278

### DIFF
--- a/bot/exts/moderation/infraction/infractions.py
+++ b/bot/exts/moderation/infraction/infractions.py
@@ -10,7 +10,7 @@ from discord.ext.commands import Context, command
 from bot import constants
 from bot.bot import Bot
 from bot.constants import Event
-from bot.converters import Expiry, FetchedMember
+from bot.converters import Duration, Expiry, FetchedMember
 from bot.decorators import respect_role_hierarchy
 from bot.exts.moderation.infraction import _utils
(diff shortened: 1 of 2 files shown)


Den4200 (changes requested):
I don't think we want to keep `!shadowtempban` either - it can probably be removed as well.
https://github.com/python-discord/bot/blob/7b3d8ed6e655b787c2538541ca35301f2fadb508/bot/exts/moderation/infraction/infractions.py#L201-L225

SebastiaanZ:
I'd like to keep shadowtempban. It's the one that was actually being used from time to time in case we want to tempban someone who's just left the server.

Since it was actually being used, I'd rather not remove it
 There's no harm in keeping it.

Den4200 (approved):
Looks good to me.

ks129 (approved):
Looks good for me!

PR #1313, opened 2020-11-30
Adds Member Checks Before Changing Voice

### LINKED ISSUE #1309: Voice ban bug for non-members
Trying to voice ban someone who left the server leads to:
image

The voice ban still registers, but the exception should be caught.

### LINKED ISSUE #1370: Gracefully handle user not being present in voice ban
When a user is voice banned without being a member of the server the bot still tries to move them to a different voice channel. We should gracefully handle the user not being in voice or not being in the server (the infraction still applies fine) by doing a membership check before attempting to move the user.

Sentry Issue: [BOT-KT](https://sentry.io/organizations/python-discord/issues/2159677697/?referrer=github_integration)

```
AttributeError: 'User' object has no attribute 'move_to'
  File "bot/exts/moderation/infraction/infractions.py", line 95, in voiceban
    await self.apply_voice_ban(ctx, user, reason)
  File "bot/decorators.py", line 135, in wrapper
    await func(*args, **kwargs)
  File "bot/exts/moderation/infraction/infractions.py", line 354, in apply_voice_ban
    await user.move_to(None, reason="Disconnected from voice to apply voiceban.")

Error executing command invoked by joe#6000: !voiceban [REDACTED] [REDACTED]
```

HassanAbouelela:
Closes #1309, closes #1370.

Checks is a user is a guild member, before performing guild operations (such as kicking from voice or removing roles) when using voice mute or general mute.


ks129:
I think this is related with https://github.com/python-discord/bot/pull/1016

HassanAbouelela:
Ah, you're right, I didn't see that other issue.

HassanAbouelela:
Reopening, as this isn't actually a duplicate of #1016, and has been reported again in #1370. This handles a different error at a different stage from the one fixed in the other PR.

bot/exts/moderation/infraction/infractions.py
```diff
@@ -351,9 +355,13 @@ async def apply_voice_ban(self, ctx: Context, user: UserSnowflake, reason: t.Opt
         if reason:
             reason = textwrap.shorten(reason, width=512, placeholder="...")
 
-        await user.move_to(None, reason="Disconnected from voice to apply voiceban.")
+        action = None
+
+        # Skip members that left the server
+        if isinstance(user, Member):
+            await user.move_to(None, reason="Disconnected from voice to apply voiceban.")
+            action = user.remove_roles(self._voice_verified_role, reason=reason)
```
MarkKoz, line 363: This was actually not implemented correctly originally. Both of these functions should be the action, not just `remove_roles`. The benefit of being an action is that `apply_infraction` will do error handling if the action fails.

bot/exts/moderation/infraction/infractions.py
```diff
@@ -351,9 +355,13 @@ async def apply_voice_ban(self, ctx: Context, user: UserSnowflake, reason: t.Opt
         if reason:
             reason = textwrap.shorten(reason, width=512, placeholder="...")
 
-        await user.move_to(None, reason="Disconnected from voice to apply voiceban.")
+        action = None
```
MarkKoz, line 358: Something strange I noticed is that the infraction won't be scheduled to expire if it doesn't have an `action`.

https://github.com/python-discord/bot/blob/41e861c52c5e8c3f5fb751345661d8f062976d87/bot/exts/moderation/infraction/_scheduler.py#L176-L182

I'm not sure why it was done this way, but it means that a temp voice ban will only expire when the bot restarts and infractions are rescheduled. A mute doesn't suffer from this since the type check is inside the action.

MarkKoz (changes requested):

bot/exts/moderation/infraction/infractions.py
HassanAbouelela: From what I gather, the reason it was done like this is so you can have non-expiring infractions (notes and warnings for example). It is probably unintuitive behavior, and it is probably better for the scheduling to be an optional argument.

Either way, I'll add a note in the docstring for now, but I can implement my suggestion if you believe it'll be more useful.

MarkKoz: I don't think that's the reason. It already uses `if expiry`, so setting `expiry` to `None` is a way to not schedule expiration. The confusing part is why it's _also_ dependent on an action being given.

MarkKoz (approved):

PR #1318, opened 2020-12-04
Create command for showing Discord snowflake creation time

### LINKED ISSUE #1279: Command for snowflake info
When verifying users reports and when searching for information about an issue, it's helpful to be able to take a discord ID and get relevant information from it. All discord ID's encode their date+time of creation within them, so this can be used to date a dm channel's creation, a DM's timing, and in some cases can be used to jump directly back into a channel to find the actual conversation a deleted message was present in by manually constructing a channel/message link (as discord's default behavior when given a link to a message that doesn't exist is to jump as close as it can).

A command to do this manually, rather than having to rely on manual conversion or the !user command, would be helpful. 

Discord.py provides a utility function for getting the creation time from a snowflake conveniently--`discord.utils.snowflake_time`. When I personally used this, I returned the exact time of creation in UTC, and the delta between now and then, looking like this:

```
Snowflake created at 2020-11-08 16:17:58.949000, 9h 24m 12s ago.
```

A proposed name is `!snf` or `!snowflake`.

This could probably be made nicer, and dropped into an embed perhaps? I'm not sure if embedification is desired or necessary.

ks129:
Title say everything... Here is one screenshot:

Screen Shot 2020-12-04 at 14 43 09

Closes #1279

### DIFF
--- a/bot/exts/utils/utils.py
+++ b/bot/exts/utils/utils.py
@@ -168,7 +168,7 @@ class Utils(Cog):
         embed.description = best_match
         await ctx.send(embed=embed)
 
-    @command(aliases=("snf", "snfl"))
+    @command(aliases=("snf", "snfl", "sf"))
     @in_whitelist(channels=(Channels.bot_commands,), roles=STAFF_ROLES)
     async def snowflake(self, ctx: Context, snowflake: int) -> None:
         """Get Discord snowflake creation time."""
--- a/bot/exts/utils/utils.py
+++ b/bot/exts/utils/utils.py
@@ -9,6 +9,7 @@ from typing import Dict, Optional, Tuple, Union
 
 from discord import Colour, Embed, utils
 from discord.ext.commands import BadArgument, Cog, Context, clean_content, command, has_any_role
+from discord.utils import snowflake_time
 
 from bot.bot import Bot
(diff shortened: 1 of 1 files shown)


Den4200 (approved):
Works great, maybe a `sf` alias as well?

ks129:
Added `sf` alias to command.

bot/exts/utils/utils.py
```diff
@@ -166,6 +168,21 @@ async def zen(self, ctx: Context, *, search_value: Union[int, str, None] = None)
         embed.description = best_match
         await ctx.send(embed=embed)
 
+    @command(aliases=("snf", "snfl", "sf"))
+    @in_whitelist(channels=(Channels.bot_commands,), roles=STAFF_ROLES)
+    async def snowflake(self, ctx: Context, snowflake: int) -> None:
```
mbaruh, line 173: Seems like the Snowflake converter could fit here

mbaruh (changes requested):

bot/exts/utils/utils.py
ks129: What? Custom converter?

ks129: Solved in 47a2607

mbaruh (approved):
Thanks, looks good

Issue #1224, opened 2020-10-09
Improving Superstar command


mbaruh:
Out of the moderation tools we have available, the superstar command is the odd one out.

* It doesn't ping the user being superstarred, instead posting a special embed (albeit a pretty funny one)
* It doesn't send the reason in the DM.

That leads to some confusion for users on what happened and why it happened.
We don't have to remove the special embed, but:
* It should still give a standard "applied superstar to <user>" so that the user is pinged if it's applied in a public channel
* The embed should be sent in a DM as well
* The optional reason should be integrated into the embed to give any additional context provided by the moderator.

spitfire-hash:
Can you assign this to me?

Akarys42:
Done! As always, if you have any question, feel free to ask in #dev-contrib!

spitfire-hash:
Okay, thank you! I'll work on it in 2 days.

mbaruh:
This might clash with https://github.com/python-discord/bot/pull/1204/files so you might want to check that out

spitfire-hash:
@mbaruh what should I do know? wait for merging #1204 and then work on that?

mbaruh:
@spitfire-hash #1204 makes some extensive changes to the command, I think it's best if you waited for it to be merged, yes.

mbaruh:
#1204 is now merged, so this issue is now up for grabs. If anyone wants to be assigned feel free to say so.

PR #1319, opened 2020-12-06
Relay the moderator's reason for `!superstarify` to the user

swfarnsworth:
This PR closes issue #1224.

Previously, when a moderator used the `!superstarify` command, the reason they gave only gets recorded in the user's infraction history and is not relayed to the user in the direct message sent by the bot. This PR changes the message the user is sent to fix this.

However, we're discussing if the special embed that gets posted in the channel where the command is issued is actually appropriate.

image

### DIFF
--- a/bot/exts/moderation/infraction/superstarify.py
+++ b/bot/exts/moderation/infraction/superstarify.py
@@ -157,32 +157,29 @@ class Superstarify(InfractionScheduler, Cog):
             New nickname: `{forced_nick}`
         """).strip()
 
-        formatted_reason = f'**Additional details:** {reason}\n\n' if reason else ''
-
-        embed_reason = (
+        user_message = (
             f"Your previous nickname, **{old_nick}**, "
-            f"didn't comply with our nickname policy. "
+            f"was so bad that we have decided to change it. "
             f"Your new nickname will be **{forced_nick}**.\n\n"
-            f"{formatted_reason}"
+            "{reason}"
             f"You will be unable to change your nickname until **{expiry_str}**. "
             "If you're confused by this, please read our "
             f"[official nickname policy]({NICKNAME_POLICY_URL})."
-        )
+        ).format
 
         successful = await self.apply_infraction(
             ctx, infraction, member, action(),
-            user_reason=embed_reason,
+            user_reason=user_message(reason=f'**Additional details:** {reason}\n\n' if reason else ''),
             additional_info=nickname_info
         )
 
-        # Send an embed with the infraction information to the invoking context if
-        # superstar was successful.
+        # Send an embed with to the invoking context if superstar was successful.
         if successful:
             log.trace(f"Sending superstar #{id_} embed.")
             embed = Embed(
-                title="Congratulations!",
+                title="Superstarified!",
                 colour=constants.Colours.soft_orange,
-                description=embed_reason
+                description=user_message(reason='')
             )
             await ctx.send(embed=embed)
 
--- a/bot/exts/moderation/infraction/superstarify.py
+++ b/bot/exts/moderation/infraction/superstarify.py
@@ -104,7 +104,7 @@ class Superstarify(InfractionScheduler, Cog):
 
             await self.reapply_infraction(infraction, action)
 
-    @command(name="superstarify", aliases=("force_nick", "star"))
+    @command(name="superstarify", aliases=("force_nick", "star", "starify"))
     async def superstarify(
         self,
         ctx: Context,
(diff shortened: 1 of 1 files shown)


swfarnsworth:
Here's an alternative message (I have not committed or pushed this yet):

image

This is the DM that they get:
image

This is what the mod would see in the user's infraction history:
image

swfarnsworth:
Per @mbaruh's suggestion, I made another change whereby the "Additional details: " portion of the message does not appear in the embed displayed in the channel (since it's just restating what the mod wrote in the command), whereas the message still appears in the message sent to the user. The messages are otherwise the same.

swfarnsworth:
image

Per an internal discussion, we'll retain the original wording of the embed, but the title will still be "Superstarify!" instead of "Congratulations!"

bot/exts/moderation/infraction/superstarify.py
```diff
@@ -128,16 +128,16 @@ async def superstarify(
 
         Alternatively, an ISO 8601 timestamp can be provided for the duration.
 
-        An optional reason can be provided. If no reason is given, the original name will be shown
-        in a generated reason.
+        An optional reason can be provided, which would be added to a message stating their old nickname
+        and linking to the nickname policy.
```
HassanAbouelela, line 132: Was there an official consensus on echoing back the user's name, or is that not a concern?

bot/exts/moderation/infraction/superstarify.py
```diff
@@ -152,15 +152,26 @@ async def action() -> None:
...
         """).strip()
 
+        formatted_reason = f'**Additional details:** {reason}\n\n' if reason else ''
+
+        embed_reason = (
+            f"Your previous nickname, **{old_nick}**, "
+            f"didn't comply with our nickname policy. "
+            f"Your new nickname will be **{forced_nick}**.\n\n"
+            f"{formatted_reason}"
+            f"You will be unable to change your nickname until **{expiry_str}**. "
```
HassanAbouelela, line 167: The expiry time is duplicated in two different spots in the direct message embed ([L115](https://github.com/python-discord/bot/blob/c3a927569da782c24299c8ae75df28ae6cd3f2ba/bot/exts/moderation/infraction/_scheduler.py#L115)). Can you remove this from the main embed, and only add it to the message sent to the invocation context?

bot/exts/moderation/infraction/superstarify.py
```diff
@@ -171,18 +182,11 @@ async def action() -> None:
             embed = Embed(
                 title="Congratulations!",
```
HassanAbouelela, line 183: Don't forget to update this, based on the poll results.

HassanAbouelela (changes requested):

bot/exts/moderation/infraction/superstarify.py
mbaruh: This is referring to what will appear in the infraction history.

bot/exts/moderation/infraction/superstarify.py
mbaruh: I think it's important to specify that they will be unable to change the nickname until the infraction expires, and in that case duplicating the date doesn't seem too bad.

swfarnsworth:
The public embed is always the same:
image

The infraction message depends on whether or not the mod gave a reason:
image

image

The message in the infraction history is unchanged from the previous set of screenshots.

bot/exts/moderation/infraction/superstarify.py
HassanAbouelela: Sure, but it does appear in the infraction history, DMs, and server invocation channel. I am curious if we want to be sending the old names as part of the (public-facing) embeds or not. It was a point touched upon during discussions, but not fully explored.

bot/exts/moderation/infraction/superstarify.py
HassanAbouelela: We can still mention that you can't change your name while this is in effect, without adding the date, but at that point, we're splitting hairs. If you're fine with the duplication, then I won't complain.

bot/exts/moderation/infraction/superstarify.py
```diff
@@ -152,37 +152,38 @@ async def action() -> None:
...
 
+        user_message = (
+            f"Your previous nickname, **{old_nick}**, "
+            f"was so bad that we have decided to change it. "
+            f"Your new nickname will be **{forced_nick}**.\n\n"
+            "{reason}"
+            f"You will be unable to change your nickname until **{expiry_str}**. "
+            "If you're confused by this, please read our "
+            f"[official nickname policy]({NICKNAME_POLICY_URL})."
+        ).format
```
Den4200, line 168: This feels a bit unconventional and odd, but it's quite smart.

Den4200 (approved):
Looks and works great!

bot/exts/moderation/infraction/superstarify.py
mbaruh: infraction history and DM - yes, we want to know what the old nick was for future reference, and I don't see a reason for it to not be DMed.
For the public embed, from what I gathered people didn't seem to think it was an major issue, and most starts are applied form private channels anyway. Also in the vote it says to keep the content, and I can only assume it includes the old nick.

mbaruh (approved):

HassanAbouelela (approved):

PR #1322, opened 2020-12-10
New `!verify` command to manually apply the Developer role

swfarnsworth:
This PR submits a new `!verify` command for moderators to apply the Developer role to any user. This is so that we can give the role to users who inform us via modmail that they're having difficulty verifying.

Previously, `!verify` was an alias for `!accept`.

Closes python-discord/mods#5


jb3 (changes requested):
Code all looks good, very clean!

I'm not sure this feature needs a cog of it's own — it fits fairly well into the existing `Verification` cog.

I'd be intereted to hear from others though, so I'll request changes for now.

swfarnsworth:
The method for the actual command could be copied over to the Verification cog almost unchanged, so that would be an easy fix. There wasn't any particular reason this had to be its own cog, so I'm happy to make that change unless someone feels otherwise.

Den4200:
Putting this in the `Verification` cog sounds good to me.

jb3 (approved):
Looking good! :shipit:

bot/exts/moderation/verification.py
```diff
@@ -848,6 +848,22 @@ async def bot_check(ctx: Context) -> bool:
         else:
             return True
 
+    @command(name='verify')
+    @has_any_role(*constants.MODERATION_ROLES)
+    async def apply_developer_role(self, ctx: Context, user: discord.Member) -> None:
+        """Command for moderators to apply the Developer role to any user."""
+        log.trace(f'verify command called by {ctx.author} for {user.id}.')
+        developer_role = self.bot.get_guild(constants.Guild.id).get_role(constants.Roles.verified)
```
mbaruh, line 856: This should probably use `constants.Roles.verified`, and then it can be combined with the next line.

bot/exts/moderation/verification.py
```diff
@@ -848,6 +848,22 @@ async def bot_check(ctx: Context) -> bool:
...
+
+        if developer_role in user.roles:
+            log.trace(f'{user.id} is already a developer, aborting.')
+            await ctx.send(f'{constants.Emojis.cross_mark} {user} is already a developer.')
```
mbaruh, line 860: Nitpicky and not important, but I'd use `already verified` here. Feel free to ignore this one.

mbaruh (changes requested):

mbaruh (approved):

PR #1326, opened 2020-12-14
Fix backtick escapes in the codeblock tag

Numerlor:
On some devices the previous escaping didn't work properly, escaping
all backticks will make sure none of them get registered as Markdown

### DIFF
--- a/bot/resources/tags/codeblock.md
+++ b/bot/resources/tags/codeblock.md
@@ -1,7 +1,7 @@
 Here's how to format Python code on Discord:
 
-\```py
+\`\`\`py
 print('Hello world!')
-\```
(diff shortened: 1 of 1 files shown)


MarkKoz:
Does this fix need to be applied to the examples used by the code block cog too?

Numerlor:
The example string in the cog already escapes all of the backticks https://github.com/python-discord/bot/blob/bcab67375c778fb30c86d8edd19bed854b0f8b45/bot/exts/info/codeblock/_instructions.py#L11-L15

MarkKoz (approved):

ks129 (approved):

PR #1328, opened 2020-12-18
Manual verification with `!verify` sends the user the confirmation DM.

swfarnsworth:
Previously, users wouldn't get the DM from the bot that one usually gets when they verify if a moderator verifies them manually. This PR solves this.

Per @Den4200's request, the response from the bot also mentions the user rather than stating their account name.


swfarnsworth:
I didn't realize the native gate rollout was imminent when making this PR. I'll submit a new version that checks if the user is pending, and if not, assigns and subsequently removes the developer role. I'll probably also rename the function accordingly.

bot/exts/moderation/verification.py
```diff
@@ -55,7 +55,7 @@
 """
 
 ALTERNATE_VERIFIED_MESSAGE = f"""
-Thanks for accepting our rules!
+You're now verified!
```
Xithrius, line 58: It may just be me, but I feel this would be "easier" to read for more users by saying 
```suggestion
You are now verified!
```

Xithrius (commented):
Besides this one line in the `ALTERNATE_VERIFIED_MESSAGE` docstring that may need to be changed, everything looks good to me.

bot/exts/moderation/verification.py
swfarnsworth: I made that change under the pretense that we'd still have bot verification. Does anyone actually get the alternate verified message anymore?

Xithrius (approved):
LGTM!

bot/exts/moderation/verification.py
```diff
@@ -834,19 +834,21 @@ async def bot_check(ctx: Context) -> bool:
...
-        """Command for moderators to apply the Developer role to any user."""
+    async def perform_manual_verification(self, ctx: Context, user: discord.Member) -> None:
+        """Command for moderators to verify any user."""
         log.trace(f'verify command called by {ctx.author} for {user.id}.')
-        developer_role = self.bot.get_guild(constants.Guild.id).get_role(constants.Roles.verified)
 
-        if developer_role in user.roles:
-            log.trace(f'{user.id} is already a developer, aborting.')
-            await ctx.send(f'{constants.Emojis.cross_mark} {user} is already a developer.')
+        if user.pending:
```
ks129, line 841: When user have pending True, I think this means that user isn't already verified?

swfarnsworth: I think you're right--this should probably be `if not user.pending`.

ks129 (approved):

PR #1329, opened 2020-12-19
Better handle native gated users

jb3:
Until Discord.py emits the `pending` property in `member_update` events we must use a workaround to watch for users who pass the native gate.

This PR creates an internal set of members who are pending and then eventually removes them once the pending property gets set to false. It is only a temporary solution before the pending event is handled by Discord.py.


mbaruh (approved):

Den4200 (approved):
Let's do this thing.

PR #1331, opened 2020-12-19
Use member update event for verification DMs

### LINKED ISSUE #1330: Use on_member_update for verification DMs
We want to DM users some information once they pass the native Discord verification gate. To do this we need to watch for the `pending` attribute of the user change from `True` to `False`.

As of now we have implemented this by keeping track of a list of users who need to receive a welcome DM in a set called `pending_members` in the `Verification` cog but we should switch this to use `on_member_update(before, after)` as soon as discord.py makes this available.

Once a user passes the native gate we want to send them the `ALTERNATE_VERIFIED_MESSAGE` defined in this cog.

As of now Discord.py do not have an upstream issue for this, but we should keep track of it to watch for when one is opened.

jb3:
- Install git in Docker container
- Bump discord.py to a unreleased ref with pending flag
- Update verification.py to use on_member_update, closes #1330

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,11 @@ ENV PIP_NO_CACHE_DIR=false \
     PIPENV_IGNORE_VIRTUALENVS=1 \
     PIPENV_NOSPIN=1
 
+RUN apt-get -y update \
+    && apt-get install -y \
+        git \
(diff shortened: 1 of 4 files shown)


mbaruh (approved):

Den4200 (approved):
Let's do it again!

PR #1334, opened 2020-12-19
Fix flake8 pre-commit hook running through PyCharm

MarkKoz:
Running the hook through PyCharm used to work a long time ago due to a pipenv bug which eventually got fixed. Now it normally fails because `PIPENV_IGNORE_VIRTUALENVS` is unset. Manually editing the hook and adding `os.putenv('PIPENV_IGNORE_VIRTUALENVS', '1')` or similar fixes it.

Recently, I've found out about `language: system` for the hook, which is a better alternative to setting the environment variable. As per https://github.com/pre-commit/pre-commit/issues/1412#issuecomment-619263744 it means "I don't want pre-commit to manage anything, I promise the tools will be installed globally on my system." I'm not sure exactly why it fixes things but I guess it's because pre-commit will no longer try to invoke pipenv while its own Python environment is active.

Try out the fix and comment on its success or failure. Don't forget to remove any manual modifications you made to your hook.


Xithrius (approved):
When commiting with a fault that flake8 wouldn't accept (3 newlines), the precommit hook alerts me that I cannot continue. 

This change seems to be successful for me.

image

ks129 (approved):
Tested locally, works correctly.

PR #1338, opened 2020-12-25
Made a json tag for the commonly seen error JSONDecodeError: Expecting value: line 1 column 1 (char 0)

sebkuip:
I have seen a lot of people come for help to the python discord with the error JSONDecodeError: Expecting value: line 1 column 1 (char 0). Which really is a simple fix. To speed up the helping progress I thought having it as a tag would help.

Feel free to suggest any rewording/renaming and I will be happy to comply.


bot/resources/tags/json.md
```diff
@@ -0,0 +1,15 @@
+JSON is a very handy way of storing values. But when starting out with JSON, you might run into the following error.
```
MarkKoz, line 1: > But when starting out with JSON

This statement could be interpreted as implying the error is a beginner issue, but that isn't really relevant. The emphasis should be on the fact that this often happens when a new empty file is created/opened, not on anything relating to the user's level of experience.

bot/resources/tags/json.md
```diff
@@ -0,0 +1,15 @@
...
+
+`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
+
+What this error means is that the decoder (the reading of the JSON file) is expecting a character on line 1 column 1. In other words, your JSON is missing a critical character at the start. So it's saying that your JSON is invalid.
```
MarkKoz, line 5: > What this error means is that the decoder (the reading of the JSON file) is expecting a character on line 1 column 1.

This statement is just restating the error so it's redundant.

bot/resources/tags/json.md
```diff
@@ -0,0 +1,15 @@
...
+Looking at how JSON is set up, you can see that they require the following at minimum if you want to start out.
```
MarkKoz, line 6: > your JSON is missing a critical character at the start

> they require the following at minimum if you want to start out.

This isn't correct. The JSON standard basically expects the root element to be a "value". That term has a specific definition and that's what the error is referring to. The grammar defines a "value" as any of the following:
```
object
array
string
number
"true"
"false"
"null"
```

I don't think this tag necessarily has to go into such detail. It could just point out the common cases of using an array or object as the root element. However, it should be phrased in a way that doesn't imply these are the only options.

MarkKoz (changes requested):

bot/resources/tags/json.md
sebkuip: Thank you for this list. I have pushed a new change hopefully resolving these issues you pointed out.

bot/resources/tags/json.md
```diff
@@ -12,4 +24,4 @@ Looking at how JSON is set up, you can see that they require the following at mi
 }
 ```
 
-If you add at least this to your JSON file, it should work. Make sure to put all your data between the `{}`, just like you would when making a dictionary.
+Make sure to put all your data between the `{}`, just like you would when making a dictionary.
```
MarkKoz, line 27: It should be clearer that an object is equivalent to a dictionary and an array is equivalent to a list.

bot/resources/tags/json.md
```diff
@@ -0,0 +1,27 @@
+When creating a new JSON file you might run into the following error.
+
+`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
+
+In short this error means your JSON is invalid in it's current state.
```
MarkKoz, line 5: Grammar
```suggestion
In short, this error means your JSON is invalid in its current state.
```

bot/resources/tags/json.md
```diff
@@ -0,0 +1,27 @@
...
+
+```
+object
+array
+string
+number
+"true"
+"false"
+"null"
+```
```
MarkKoz, line 16: I should've been more unequivocal. Listing out all possibilities isn't really beneficial and will likely serve to confuse people. 99% of the time people will either want an object or an array. Therefore, this tag should only mention those two cases e.g. "You probably want to use an array or object, blah blah blah..."

bot/resources/tags/json.md
```diff
@@ -0,0 +1,27 @@
+When creating a new JSON file you might run into the following error.
+
+`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
+
+In short this error means your JSON is invalid in it's current state.
+A JSON may never be completely empty and must always at least have one of the following items.
```
MarkKoz, line 6: I'd lead into this more following the previous statement e.g. "This is probably because your new file is empty."

bot/resources/tags/json.md
```diff
@@ -0,0 +1,27 @@
...
+object
+array
+string
+number
+"true"
+"false"
+"null"
+```
+
+To resolve this issue, you create one of the above values in your JSON. It is very common to use `{}` to make an object. Adding the following to your JSON should resolve this issue.
```
MarkKoz, line 18: It should be clearer that an object is equivalent to a dictionary and an array is equivalent to a list.

MarkKoz (changes requested):
The tag should be named something like `empty_json` rather than just `json`.

Xithrius:
@sebkuip are you planning on finishing this?

sebkuip:
> @sebkuip are you planning on finishing this?

Ah I did forget a bit about it after a few school projects. Should have more time in the upcoming week so I’ll see if I can get back at it with a fresh mind

Xithrius:
Alright cool, thanks for the speedy reply.

sebkuip:
Sorry for the history added, since I left the project for a bit I was about 190 commits on master behind and had to rebase.

sebkuip:
Issues trying to resolve conflicts with master, going to close this. Make a new branch and new PR to hopefully fix this.

PR #1339, opened 2020-12-25
`!user` command says if a user is "Verified", rather than "Pending"

swfarnsworth:
Previously, the `!user` command would tell mods if a user is "Pending", which is True before they are verified and False thereafter. I think most of us are interested in whether or not they are "Verified", which is False when they are pending and True otherwise. This PR aligns the message in the embed with the expected semantics.


mbaruh (approved):

Xithrius (approved):
LGTM

PR #1342, opened 2020-12-29
Fix webhook message publishing

jb3:
Discord.py now uses the webhook HTTP state when operating on the return of `discord.Webhook.send`, this means that methods like `publish` on `discord.Message` objects no longer work.

This PR corrects places where the bot automatically publishes announcements by sending with the webhook and then fetching the message from the channel using the bot context and `discord.TextChannel.fetch_message`.

Once we have that object we are back to using the bot HTTP state again so we can use the publish methods.


MarkKoz (approved):
It works thank you

Den4200 (approved):

jb3:
Wait, don't merge this

Den4200:
#1345 should resolve the issue stated here.

PR #1346, opened 2021-01-05
Fix Race Conditions in Help Channels

### LINKED ISSUE #1341: Race condition resulting in a double close in help commands
It's possible for !close to fire at the exact same time as the autodormant check, which runs at a variable amount of time since the last messages. While rare, it results in a desynchronization of the bots internal management of help channel locations, and has in the wild resulted in a double-declaim of a help channel. 

Essentially, the !close command can attempt to deschedule an idle check that is already running at that exact moment, resulting in both the idle check and the close command placing the channel in the idle category AND in the upcoming queue for new available channels. Neither the close command nor the autodormant check to make sure the channel they are operating on is within the in-use channels, and thus they add the channel to be "made available" twice. 

There are several avenues for a fix. First, we could check to ensure a channel is not within the idle channels before scheduling it, which will fix all cases where the idle timer would finish firing before the channel is rescheduled (but would permit the bug to remain in very-high-volume times). Another is to check the available channel queue before moving a channel to the idle category, which will prevent the double-queue, however, it doesn't fix the double-scheduled idle task and could also cause issues during high volume times.

It may be a better idea to use the on_message_lock during !close and the idle timer, to prevent them from running at the same time and stepping on each other's toes, and to prevent potential future race conditions involving the on_message handler and the idle timer and close command.

MarkKoz:
The following race conditions were fixed:

* User claiming a second channel while their first channel is in the middle of being claimed.
* Multiple people claiming the same channel (this was accounted for before, but now each channel has its own lock).
* User's channel being unclaimed while they're in the middle of claiming a new channel. This would cause their cooldown role to get incorrectly removed.
* Multiple unclaims on the same channel either from multiple `!close` invocations or `!close` being called while the channel is automatically being unclaimed. Fixes #1341.

To facilitate these fixes, some refactoring had to be done. The lock decorator also had to be updated to support waiting for an `asyncio.Lock` instead of always aborting if the lock is unavailable. There was also some other refactoring sprinkled throughout the help channel modules.

### DIFF
--- a/bot/exts/help_channels/_channel.py
+++ b/bot/exts/help_channels/_channel.py
@@ -4,8 +4,10 @@ from datetime import datetime, timedelta
 
 import discord
 
+import bot
 from bot import constants
 from bot.exts.help_channels import _caches, _message
+from bot.utils.channel import try_get_channel
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


Numerlor (approved):
I feel like the `is_auto` param name doesn't clearly explain its purpose, maybe something like `invoked_automatically` would be better; seems to be fine otherwise.

HassanAbouelela (approved):
Everything looks in order.

jb3 (approved):

PR #1350, opened 2021-01-10
Annihilate all traces of Developer and Unverified roles

mbaruh:
Surgically remove anything related to the Developer and Unverified roles using a sledgehammer.


jb3 (approved):
Spectacular. Let's terminate with extreme prejudice.

Den4200 (approved):
Thanks!

PR #1351, opened 2021-01-15
Add New Voice Channels To Config

HassanAbouelela:
Adds a new general voice and text channel to the default config, and renames general and help channels to match the new configuration in the config file and constants file. [Announcment with details](https://discord.com/channels/267624335836053506/748238795236704388/799648464061464577).

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -397,10 +397,10 @@ class Channels(metaclass=YAMLGetter):
     big_brother_logs: int
     bot_commands: int
     change_log: int
+    code_help_chat_0: int
     code_help_chat_1: int
-    code_help_chat_2: int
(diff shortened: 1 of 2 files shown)


MarkKoz:
Can you elaborate on why these are being added? Are they required by #1297? If so, it isn't necessary to open a separate PR.

HassanAbouelela:
I originally asked to update it for that PR, but was then asked to update in general ([conversation](https://discord.com/channels/267624335836053506/551789653284356126/799653265980784682)). If you think it should be added on that PR, feel free to close this one.

MarkKoz:
I believe what was told to you is false; nothing else relies on these constants. I think it makes more sense to make these changes in #1297.

PR #1354, opened 2021-01-16
Remove unwatch reason from the !nominate output

Akarys42:
As the watch reason can contain private information, we shouldn't share it with the whole staff. This commit removes the previous unwatch reason from the `!nominate` output.


ks129 (approved):

Xithrius (approved):
LGTM!

PR #1358, opened 2021-01-16
Make sure that moderators can interact with trashcans and restrict paginator usage

Akarys42:
* Add an override for moderation roles in `wait_for_deletion` 5e93396e1655715187d176d1cfdd32aa54616a1a
* Restrict the paginator usage to `ctx.author` by default and add an override for mods c25929cfef967a87fda96c2d0b26dcaf74dc5f77


bot/pagination.py
```diff
@@ -242,6 +246,9 @@ def event_check(reaction_: discord.Reaction, user_: discord.Member) -> bool:
                         scale_to_size=scale_to_size)
         current_page = 0
 
+        if not restrict_to_user:
+            restrict_to_user = ctx.author
+
         if not lines:
```
Inheritanc-e, line 252: Why not combine these 2 if statements into one using `and`?

Inheritanc-e (commented):
Just one small change.

bot/pagination.py
Inheritanc-e: Oh, my bad. 
Didn't see the entire thing.

bot/pagination.py
```diff
@@ -218,10 +222,10 @@ async def paginate(
         def event_check(reaction_: discord.Reaction, user_: discord.Member) -> bool:
             """Make sure that this reaction is what we want to operate on."""
             no_restrictions = (
-                # Pagination is not restricted
-                not restrict_to_user
```
MarkKoz, line 222: This removes support for unrestricted pagination. Now, all pagination must either be restricted to the given user or staff. Was this not a feature we relied upon?

Akarys42: As far as I can see, defaulting to `ctx.author` won't break anything, the features using the paginator should still work correctly.

bot/pagination.py
```diff
@@ -218,10 +222,10 @@ async def paginate(
...
                 # The reaction was by a whitelisted user
-                or user_.id == restrict_to_user.id
+                user_.id == restrict_to_user.id
+                # The reaction was by a moderator
+                or any(role.id in MODERATION_ROLES for role in user_.roles)
```
MarkKoz, line 228: I don't believe the paginator is inherently restricted to guilds. Therefore, this event may get a `User` which doesn't have `roles`.

MarkKoz (changes requested):

bot/pagination.py
Akarys42: That's a good point. Fixed in 46ff6949 !

MarkKoz (approved):

HassanAbouelela (approved):

HassanAbouelela (approved):

PR #1367, opened 2021-01-19
Sync: chunk user requests

MarkKoz:
The site can't handle huge syncs. Even a bulk patch of 10k users will crash the service. Chunk the requests into groups of 1000 users and await them sequentially. Testing showed that concurrent requests are not scalable and would also crash the service.

10k user patch without chunking. This caused the site service to crash and respond with 524.

bild
bild

10k user patch with chunking and concurrent requests. The operation succeeded but there was some CPU throttling. It implies that this won't scale well (the situation that triggered the discovery of this issue was a sync of over 80k users).

bild

This compares the two aforementioned tests (start times indicates by the vertical lines):

bild

I have no graphs for chunking without concurrent requests (i.e. what this PR does) but we know that this approach works as it was done in production with internal evals too.

### DIFF
--- a/bot/exts/backend/sync/_syncers.py
+++ b/bot/exts/backend/sync/_syncers.py
@@ -5,11 +5,14 @@ from collections import namedtuple
 
 from discord import Guild
 from discord.ext.commands import Context
+from more_itertools import chunked
 
 import bot
 from bot.api import ResponseCodeError
 
 log = logging.getLogger(__name__)
+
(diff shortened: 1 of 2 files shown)


jb3 (approved):
Looks good, a similar solution has already been tested in the production cluster and proven to work so this should fix our timeout problems.

Den4200 (approved):
Was lurking in #dev-ops and tested in the test server!

PR #1166, opened 2020-09-21
Enhance the !server command.

kosayoda:
The `!server` command is now more compact, and has more information.

#### In a regular channel
image

#### In a moderation channel
image

For contrast, the old embed:
image

I have no idea what I'm doing with the test.

PR #1372, opened 2021-01-20
Duckpond check

### LINKED ISSUE #1355: Make sure that messages from private channels can't be ducked
Since our `#duck-pond` channel can be viewed by all our Helpers and we *apparently* can't trust our Moderators (me included :pensive:), we need to add a check to make sure that messages from channels not viewable by Helpers can't be relayed to the duck pond. 

This can be done by adding a new check function in [`bot/bot/exts/fun/duck_pond.py`](https://github.com/python-discord/bot/blob/master/bot/exts/fun/duck_pond.py) and using it in `on_raw_reaction_add()`.

anand2312:
Aims to resolve #1355 
I also changed two type hints in functions to remain uniform with the rest of the file.


bot/exts/fun/duck_pond.py
```diff
@@ -165,6 +176,10 @@ async def on_raw_reaction_add(self, payload: RawReactionActionEvent) -> None:
         message = await channel.fetch_message(payload.message_id)
         member = discord.utils.get(message.guild.members, id=payload.user_id)
 
+        # Was the message sent in a channel Helpers can see?
+        if not self.is_helper_viewable(channel):
+            return
```
MarkKoz, line 181: Move this before the message and member are retrieved. If this condition turns out to be false, those aforementioned operations would end up being wasteful.

MarkKoz (changes requested):

MarkKoz (approved):

kosayoda (changes requested):
Good work. Since we merged #1166, we have a `is_staff_channel` helper function at `bot/bot/utils/channel.py`. We should use that instead now. Thanks!

mbaruh:
@kosayoda hmm looking at this again I may have misunderstood this issue.
Your function checks whether the channel invisible to non-staff, while here it checks whether it's visible to helpers (including public channels)

Akarys42:
> @kosayoda hmm looking at this again I may have misunderstood this issue.
> Your function checks whether the channel is invisible to non-staff, while here it checks whether it's visible to helpers (including public channels)

Yep, we can’t use the current implementation here. Maybe we could support passing a custom role list to `is_staff_channel`? I think this PR is fine though.

kosayoda:
Ah, that's my bad then.

mbaruh:
Yeah there's no reason to over-complicate the existing function

bot/exts/fun/duck_pond.py
```diff
@@ -44,6 +44,17 @@ def is_staff(member: Union[User, Member]) -> bool:
...
+    @staticmethod
+    def is_helper_viewable(channel: TextChannel) -> bool:
+        """Check if helpers can view a specific channel."""
+        guild = channel.guild
+        helper_role = guild.get_role(constants.Roles.helpers)
+        # check channel overwrites for both the Helper role and @everyone and
+        # return True for channels that they have explicit permissions to view.
+        helper_overwrites = channel.overwrites_for(helper_role)
+        default_overwrites = channel.overwrites_for(guild.default_role)
+        return default_overwrites.view_channel or helper_overwrites.view_channel
```
kosayoda, line 56:
```suggestion
        return default_overwrites.view_channel is None or helper_overwrites.view_channel is True
```
The [default value](https://discordpy.readthedocs.io/en/latest/api.html?highlight=permissionoverwrite#discord.PermissionOverwrite) of a permission *overwrite* for a channel is `None`.

Regular channels:
```
helper_overwrites = None
default_overwrites = None
```
Staff channels:
```
helper_overwrites = True
default_overwrites = False
```
Mod+ channels:
```
helper_overwrites = None
default_overwrites = False
```
Please also change the comment accordingly.

kosayoda (changes requested):

bot/exts/fun/duck_pond.py
anand2312: Oh, I knew this but I didn't know how the permissions were set up in the channels - and I wrote it this way based on what [Scaleious told me in dev-contrib] (https://discord.com/channels/267624335836053506/635950537262759947/801492910952546315)

kosayoda (approved):

PR #1374, opened 2021-01-21
Added slowmode statistic for python-general

mbaruh:
Added a statistic for advanced monitoring of the cooldown in python-general. Also made the slowmode reset command use the slowmode setting command under the hood to avoid repetition.

### DIFF
--- a/bot/exts/moderation/slowmode.py
+++ b/bot/exts/moderation/slowmode.py
@@ -7,7 +7,7 @@ from discord import TextChannel
 from discord.ext.commands import Cog, Context, group, has_any_role
 
 from bot.bot import Bot
-from bot.constants import Emojis, MODERATION_ROLES
+from bot.constants import Channels, Emojis, MODERATION_ROLES
 from bot.converters import DurationDelta
 from bot.utils import time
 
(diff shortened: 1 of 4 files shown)


bot/exts/moderation/slowmode.py
```diff
@@ -58,6 +58,10 @@ async def set_slowmode(self, ctx: Context, channel: Optional[TextChannel], delay
             log.info(f'{ctx.author} set the slowmode delay for #{channel} to {humanized_delay}.')
 
             await channel.edit(slowmode_delay=slowmode_delay)
+            if channel.id == Channels.python_general:
```
Akarys42, line 61: Why are we tracking only one channel? I think having multiple samples would be worth it, like we often have #ot0 in slowmode.

ks129: Maybe we should track all channels?

mbaruh: That was the plan, but channels can change names (especially ot), while having it by id would be pretty annoying, and we don't really need other channels for what this is intended for, so I agreed with @jb3 to limit it to just pygen.

Numerlor: I don't know how it looks on the moderation front for slowmode, but maybe discordpy would also be nice here. Dynamic tracking doesn't sound too good for channels that get a sm set rarely

Akarys42: We could have a mapping of usually slowmode-ed channel IDs to human readable name, and use that before submitting the stats.

mbaruh: I guess I could add dpy and ot0 in case it's needed in the future. It's not super important to me though.

Akarys42: Sounds good

Akarys42 (approved):
LGTM

bot/exts/moderation/slowmode.py
Shivansh-007, line 61: Instead of restricting it to just python general and manually adding it in future for other channels, why not create a list like `!otn`, and the moderators can add channels to the list and remove it.

ks129: This would make this unnecessary complex. But I still think discord.py channel should be added at least.

ks129 (approved):

PR #1376, opened 2021-01-22
Update badges on the README file

Akarys42:
Some small changes with the README badges:
* Use the new Python Discord badge
* Use the vanity URL
* Use a static license badge as the current one is broken
* Remove the site badge


Numerlor (approved):
The pydis github badge is a bit too small but I suppose that should be fixed at the branding end.

Just making sure it's intentional, should the site badge be removed?

lemonsaurus (approved):
Good change!

PR #1378, opened 2021-01-22
Create voice.md

PH-KDX:
This tag would provide info for users which are not voice-verified, so that they can easily be directed toward the appropriate channel.


Akarys42:
Thanks for the PR! I am not sure about the tag name though. I think it is better to call it `voice-verification` and rely on fuzzy matching to support `voice`.

Akarys42 (approved):
lgtm

Senjan21 (approved):
I am approving this on behalf of Mr. Hemlock.

PR #1383, opened 2021-01-25
Removed helper interpreter.

Inheritanc-e:
Removed `self.interpreter` from internal cog, and deleted the interpreter helper as it was not being used anywhere and it was basically dead code.


ks129 (approved):

Akarys42 (approved):
Why not, LGTM

PR #1392, opened 2021-01-31
Don't fuzzy search for tags when tag is on cooldown

### LINKED ISSUE #1373: Command suggester wrongly invoked for commands under cooldown
When a command is under cooldown and is invoked again, the error handler will wrongly try to suggest an alternative as if the command does not exist, when it should not do anything at all.

image

ChrisLovering:
Closes #1373.

By returning True within this command, the fuzzy search in [error_handler.py](https://github.com/python-discord/bot/blob/ced0af2b2936b1fab4f5601bd7b1c00bbbd9130a/bot/exts/backend/error_handler.py#L169) is not hit.
As the tag_get function deals with cooldowns internally, we can reserve False for when a tag is not found.

See https://discord.com/channels/267624335836053506/635950537262759947/805458232943837194 for a discussion on this.


Senjan21 (approved):
Looks like a good change, thanks.

Xithrius (approved):
Very good.

PR #1393, opened 2021-01-31
Made a tag to help people fix an empty file JSON error.

sebkuip:
From personal experience I have seen people struggle when just starting out with JSON due to having an invalid JSON form error. This was caused by an empty JSON file. This tag will help explain what is wrong in a normalised fashion.

This is the continuation of https://github.com/python-discord/bot/pull/1338


bot/resources/tags/empty-json.md
```diff
@@ -0,0 +1,27 @@
+When creating a new JSON file you might run into the following error.
+
+`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
+
+In short this error means your JSON is invalid in it's current state.
```
swfarnsworth: While true, this statement may be unnecessarily technical for some subset of the intended audience, who already know that their JSON isn't working. The line that follows it will probably be enough.

bot/resources/tags/empty-json.md
```diff
@@ -0,0 +1,27 @@
...
+A JSON may never be completely empty and must always at least have one of the following items.
+
+```
+object
+array
+string
+number
+"true"
```
swfarnsworth: Aren't `true`, `false`, and `null` valid literals in a JSON? If so, why are they in quotes here? It might be better to have an example.

bot/resources/tags/empty-json.md
```diff
@@ -0,0 +1,27 @@
+When creating a new JSON file you might run into the following error.
+
+`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
+
+In short this error means your JSON is invalid in it's current state.
+A JSON may never be completely empty and must always at least have one of the following items.
+
+```
+object
```
swfarnsworth: An object is the dict-like structure that you can have in a JSON, yes? I think it would be easier for most of the intended readers to describe the valid structures in terms of their Python analogues, since the goal isn't to teach anyone anything about JSONs beyond how to debug this error.

bot/resources/tags/empty-json.md
```diff
@@ -0,0 +1,27 @@
...
+"null"
+```
+
+To resolve this issue, you create one of the above values in your JSON. It is very common to use `{}` to make an object. Adding the following to your JSON should resolve this issue.
+
+```json
+{
+
+
+}
```
swfarnsworth: Does this mean that the JSON in this code example, if saved exactly as-is in a .json file, would be valid because it has the extra line? That's my takeaway but that seems to conflict with what else is conveyed in this message.

swfarnsworth (changes requested):

bot/resources/tags/empty-json.md
sebkuip: You might be looking at an old commit. In the latest one they are removed. It's purely list and object now.

bot/resources/tags/empty_json.md
```diff
@@ -0,0 +1,24 @@
+When creating a new JSON file you might run into the following error.
+
+`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
+
+In short, this means that your JSON is invalid in its current state. This could very well happen because the file is just new and empty.
+A JSON may never be completely empty. It is recommended to have at least one of the following in your json:
```
MarkKoz, line 6:
```suggestion
JSON data must not be completely empty. It is recommended to have at least one of the following in your json:
```

bot/resources/tags/empty_json.md
```diff
@@ -0,0 +1,24 @@
...
+
+```
+object
+array
+```
+
+To resolve this issue, you create one of the above values in your JSON. It is very common to use `{}` to make an object, which is similar to a dictionary in python.
```
MarkKoz, line 13: The first sentence would likely get removed anyway based on my other suggestions, but just in case...
```suggestion
To resolve this issue, create one of the above values in your JSON. It is very common to use `{}` to make an object, which is similar to a dictionary in python.
```

bot/resources/tags/empty_json.md
```diff
@@ -0,0 +1,24 @@
...
+
+`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
+
+In short, this means that your JSON is invalid in its current state. This could very well happen because the file is just new and empty.
+A JSON may never be completely empty. It is recommended to have at least one of the following in your json:
+
+```
+object
+array
+```
```
MarkKoz, line 11: No need to use a list when there are only two items. Furthermore, I don't like using the term "recommend" here. I suggest rephrasing it as "You probably want to use an object or an array."

MarkKoz (changes requested):

bot/resources/tags/empty-json.md
sebkuip: That extra white line between the brackets is removed.
As stated earlier when it's COMPLETELY empty (like when it's just created) it's invalid. But this makes it have at least SOME content to make it valid.
Mayb there is a better way to word it? or maybe an extra note to add?

bot/resources/tags/empty-json.md
sebkuip: Outdated. Has more info on the description below it now.

bot/resources/tags/empty-json.md
sebkuip: In the current text below it, says it's common to use {} to make an object, which is similar to a dictionary in python. Should I move this line up?

swfarnsworth:
I would suggest wording more along these lines:

```
When creating a new JSON file you might run into the following error:

`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

This means that your `.json` *file* is completely empty, but it still needs to represent some kind of data, even if it's an empty container.

To solve the error, simply change the file to read `{}`. This will save an empty "object", which is very similar to what a dict is in Python. To learn more about JSON, check [this](we need a URL that can go here) out.
```

As discussed in the development channel within Discord, these messages should be brief.

bot/resources/tags/empty-json.md
HassanAbouelela: Even if it is the correct term, I am also of the mindset that we use python keywords. The tags in my opinion are by no means an extensive resource, but more of a quick overview to explain a concept, and set someone on the right path of looking into it.

The source you included uses `object`. If someone is interested in learning more about JSON, they can follow that and learn more. It does not make sense to have it that far removed from python though when the original question is probably in python.

sebkuip: > Even if it is the correct term, I am also of the mindset that we use python keywords. The tags in my opinion are by no means an extensive resource, but more of a quick overview to explain a concept, and set someone on the right path of looking into it.
> 
> The source you included uses `object`. If someone is interested in learning more about JSON, they can follow that and learn more. It does not make sense to have it that far removed from python though when the original question is probably in python.

Ah right now I understand what you are saying. In the discord chat I didn't completely understand what you were saying. Now with context I understand more. I will fix this.

bot/resources/tags/empty-json.md
```diff
@@ -0,0 +1,11 @@
...
+```
+JSONDecodeError: Expecting value: line 1 column 1 (char 0)
+```
+This error could have appeared because you just created the JSON file and there is nothing in it at the moment.
+
+Whilst having the data empty is no problem, the file itself may never be completely empty.
+
+You most likely wanted to structure your JSON as a dictionary. To do this, change your JSON to read `{}`.
+
+Different data types are also supported. If you wish to read more on these, please reffer to the following article: https://www.tutorialspoint.com/json/json_data_types.htm
```
kosayoda, line 11:
```suggestion
Different data types are also supported. If you wish to read more on these, please refer to the following article: https://www.tutorialspoint.com/json/json_data_types.htm
```

bot/resources/tags/empty-json.md
```diff
@@ -0,0 +1,11 @@
...
+```
+JSONDecodeError: Expecting value: line 1 column 1 (char 0)
+```
+This error could have appeared because you just created the JSON file and there is nothing in it at the moment.
+
+Whilst having the data empty is no problem, the file itself may never be completely empty.
+
+You most likely wanted to structure your JSON as a dictionary. To do this, change your JSON to read `{}`.
+
+Different data types are also supported. If you wish to read more on these, please refer to the following article: https://www.tutorialspoint.com/json/json_data_types.htm
```
Numerlor, line 11:
```suggestion
Different data types are also supported. If you wish to read more on these, please refer to [this article](https://www.tutorialspoint.com/json/json_data_types.htm).
```
Making this a text link hides the long link, saving some space

Numerlor (changes requested):

bot/resources/tags/empty-json.md
```diff
@@ -0,0 +1,11 @@
+When using JSON, you might run into the following error:
+```
+JSONDecodeError: Expecting value: line 1 column 1 (char 0)
+```
+This error could have appeared because you just created the JSON file and there is nothing in it at the moment.
+
+Whilst having the data empty is no problem, the file itself may never be completely empty.
```
MarkKoz, line 7: Maybe it's just me, but I found the original phrasing a bit awkward.
```suggestion
Whilst having empty data is no problem, the file itself may never be completely empty.
```

bot/resources/tags/empty-json.md
```diff
@@ -0,0 +1,11 @@
...
+
+You most likely wanted to structure your JSON as a dictionary. To do this, change your JSON to read `{}`.
```
MarkKoz, line 9: It can be more explicit about what it's trying to convey
```suggestion
You most likely wanted to structure your JSON as a dictionary. To do this, edit your empty JSON file so that it instead contains `{}`.
```

MarkKoz (changes requested):
Seems like there has been a lot of iteration on the tag content. I fundamentally like the final result but there are just a few minor improvements to phrasing.

Xithrius:
@sebkuip will you be updating this PR relatively soon within the future?

sebkuip:
> @sebkuip will you be updating this PR relatively soon within the future?

Sorry was under the impression that I already merged the requested changes. I have merged these now.

MarkKoz (approved):

Xithrius:
@swfarnsworth Please review when you're available.

swfarnsworth (approved):
Thanks!

PR #1395, opened 2021-02-02
Add tag on float imprecision

wookie184:
Adds a tag on float imprecision
image

### DIFF
--- a/bot/resources/tags/floats.md
+++ b/bot/resources/tags/floats.md
@@ -8,12 +8,13 @@ You may have noticed that when doing arithmetic with floats in Python you someti
 Internally your computer stores floats as as binary fractions. Many decimal values cannot be stored as exact binary fractions, which means an approximation has to be used.
 
 **How you can avoid this**
-If you require an exact decimal representation, you can use the [decimal](https://docs.python.org/3/library/decimal.html) or [fractions](https://docs.python.org/3/library/fractions.html) module. Here is an example using the decimal module:
+ You can use [math.isclose](https://docs.python.org/3/library/math.html#math.isclose) to check if two floats are close, or to get an exact decimal representation, you can use the [decimal](https://docs.python.org/3/library/decimal.html) or [fractions](https://docs.python.org/3/library/fractions.html) module. Here are some examples:
 ```python
->>> from decimal import Decimal
->>> Decimal('0.1') + Decimal('0.2')
+>>> math.isclose(0.1 + 0.2, 0.3)
+True
+>>> decimal.Decimal('0.1') + decimal.Decimal('0.2')
 Decimal('0.3')
 ```
-Note that we enter in the number we want as a string so we don't pass on the imprecision from the float.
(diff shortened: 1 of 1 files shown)


Shivansh-007:
Can we also give an example where there are represented exactly:
```py
# This is because 0.125 is precisely 1/8, or 1/(2^3)
>>> format(0.125, '.25f')
'0.1250000000000000000000000'
```
Then the `math.isclose()` for Floats Equality Testing:
```py
>>> x = 0.1 + 0.1 + 0.1
>>> y = 0.3
>>> x == y
False
>>> from math import isclose
>>> x = 0.0000001
>>> y = 0.0000002
>>> 
>>> a = 123456789.01
>>> b = 123456789.02
>>> 
>>> print('x = y:', isclose(x, y, abs_tol=0.0001, rel_tol=0.01))
x = y: True
>>> print('a = b:', isclose(a, b, abs_tol=0.0001, rel_tol=0.01))
a = b: True
>>> 
```
And last not to use `float` datatypes while giving it the Decimal Function
```py
# As you can see, since we passed an approximate binary float to the Decimal constructor it did it's best to represent that binary float exactly!!
# So, instead, use strings or tuples in the Decimal constructor.
>>> from decimal import Decimal
>>> format(0.1, '.25f')
'0.1000000000000000055511151'
>>> Decimal(0.1)
Decimal('0.1000000000000000055511151231257827021181583404541015625')
```

wookie184:
The tag is already fairly long, and our aim is to try and keep them fairly short so i've only put in what I think is most important, and linked to other things. Regarding your three points:

1. it's explained at the start of the first link, https://docs.python.org/3/tutorial/floatingpoint.html, much more clearly then I could if I tried to fit it in.
2. I think the tag would be too big with this explained properly, although I think it could help to add a line like `You can also check if two floats are close with math.isclose` and link to the docs, especially as it doesn't seem to be mentioned in the links.
3. It's already mentioned, `Note that we enter in the number we want as a string so we don't pass on the imprecision from the float.`

Shivansh-007:
> The tag is already fairly long, and our aim is to try and keep them fairly short so i've only put in what I think is most important, and linked to other things. Regarding your three points:
> 
> 1. it's explained at the start of the first link, https://docs.python.org/3/tutorial/floatingpoint.html, much more clearly then I could if I tried to fit it in.
> 2. I think the tag would be too big with this explained properly, although I think it could help to add a line like `You can also check if two floats are close with math.isclose` and link to the docs, especially as it doesn't seem to be mentioned in the links.
> 3. It's already mentioned, `Note that we enter in the number we want as a string so we don't pass on the imprecision from the float.`

Alright, You can do that for the 2nd point, maybe add a single line example?

wookie184:
Updated, now looks like this
image

Shivansh-007 (approved):
Thanks! Looks Good To Me!

Xithrius (approved):

Akarys42 (approved):
Lovely

PR #1325, opened 2020-12-13
Add dynamic message in #how-to-get-help

mudkipdev:
Implements #1320.

I think all the code needed is there but I may have done it wrong since it is my first time contributing.

PR #1396, opened 2021-02-05
Dynamic available help channels message

### LINKED ISSUE #1320: Add a dynamic message at the bottom of `#how-to-get-help`
Actions https://github.com/python-discord/organisation/issues/315. 

We should have a dynamic message at the bottom of #how-to-get-help saying something along the lines of 
“The currently available channels are #help-element1 and #help-element2, feel free to ask in any of those two channels!”, 
to avoid getting anyone getting confused by the currently open channels like we had in the past.

Xithrius:
Closes #1320
Rewrite of #1325

Users sometimes get confused on which channels are available for them to claim.

With this PR, a message in the #how-to-get-help channel will be edited to reflect which channels are currently available.

If the bot cannot edit the most recent message in the #how-to-get-help channel, then it knows it's trying to edit the webhook message that contains the guide for claiming a channel. It will send the dynamic message if this occurs.

If we do decide to remove or add more channels in the `help: available` category, then the bot will be able to pick up on them.

image

### DIFF
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -495,18 +495,16 @@ class HelpChannels(commands.Cog):
                 c.id for c in available_channels_category.channels if 'help-' in c.name
             )
 
-        if self.how_to_get_help is None:
-            self.how_to_get_help = await channel_utils.try_get_channel(Channels.how_to_get_help)
-
-        if self.dynamic_message is None:
-            last_message = await self.how_to_get_help.history(limit=1)
-            self.dynamic_message = next(last_message)
-
         available_channels = AVAILABLE_HELP_CHANNELS.format(
             available=', '.join(f"<#{c}>" for c in self.available_help_channels)
         )
 
-        try:
-            await self.dynamic_message.edit(content=available_channels)
-        except discord.Forbidden:
-            self.dynamic_message = await self.how_to_get_help.send(available_channels)
+        if self.how_to_get_help is None:
+            self.how_to_get_help = await channel_utils.try_get_channel(Channels.how_to_get_help)
+
+        if self.dynamic_message is None:
+            try:
+                self.dynamic_message = await self.how_to_get_help.fetch_message(self.how_to_get_help.last_message_id)
+                await self.dynamic_message.edit(content=available_channels)
+            except discord.NotFound:
+                self.dynamic_message = await self.how_to_get_help.send(available_channels)
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -21,7 +21,7 @@ NAMESPACE = "help"
 HELP_CHANNEL_TOPIC = """
 This is a Python help channel. You can claim your own help channel in the Python Help: Available category.
 """
-AVAILABLE_HELP_CHANNELS = """**Currently available help channel(s):** {available}"""
+AVAILABLE_HELP_CHANNELS = "**Currently available help channel(s):** {available}"
 
 
 class HelpChannels(commands.Cog):
@@ -500,3 +500,4 @@ class HelpChannels(commands.Cog):
             await self.dynamic_message.edit(content=available_channels)
         except discord.NotFound:
             self.dynamic_message = await self.how_to_get_help.send(available_channels)
+            log.trace("A dynamic message was sent for later modification because one couldn't be found.")
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -123,12 +123,11 @@ class HelpChannels(commands.Cog):
 
         await _caches.unanswered.set(message.channel.id, True)
 
-        # Not awaited because it may indefinitely hold the lock while waiting for a channel.
-        scheduling.create_task(self.move_to_available(), name=f"help_claim_{message.id}")
-
         # Removing the help channel from the dynamic message, and editing/sending that message.
         self.available_help_channels.remove(message.channel.id)
-        await self.update_available_help_channels()
+
+        # Not awaited because it may indefinitely hold the lock while waiting for a channel.
+        scheduling.create_task(self.move_to_available(), name=f"help_claim_{message.id}")
 
     def create_channel_queue(self) -> asyncio.Queue:
         """
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -21,9 +21,7 @@ NAMESPACE = "help"
 HELP_CHANNEL_TOPIC = """
 This is a Python help channel. You can claim your own help channel in the Python Help: Available category.
 """
-AVAILABLE_HELP_CHANNELS = """
-**Currently available help channel(s):** {available}
-"""
+AVAILABLE_HELP_CHANNELS = """**Currently available help channel(s):** {available}"""
 
 
 class HelpChannels(commands.Cog):
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -123,11 +123,12 @@ class HelpChannels(commands.Cog):
 
         await _caches.unanswered.set(message.channel.id, True)
 
-        self.available_help_channels.remove(message.channel.id)
-        await self.update_available_help_channels()
-
         # Not awaited because it may indefinitely hold the lock while waiting for a channel.
         scheduling.create_task(self.move_to_available(), name=f"help_claim_{message.id}")
+
+        # Removing the help channel from the dynamic message, and editing/sending that message.
+        self.available_help_channels.remove(message.channel.id)
+        await self.update_available_help_channels()
 
     def create_channel_queue(self) -> asyncio.Queue:
         """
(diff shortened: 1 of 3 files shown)


bot/exts/help_channels/_cog.py
```diff
@@ -461,3 +487,25 @@ async def wait_for_dormant_channel(self) -> discord.TextChannel:
...
+            self.available_help_channels = set(
+                c.id for c in available_channels_category.channels if 'help-' in c.name
+            )
+
+        available_channels = AVAILABLE_HELP_CHANNELS.format(
+            available=', '.join(f"<#{c}>" for c in self.available_help_channels) or None
+        )
+
+        if self.how_to_get_help is None:
+            self.how_to_get_help = await channel_utils.try_get_channel(Channels.how_to_get_help)
```
HassanAbouelela, line 504: Could you add a check for `how_to_get_help` being an instance of TextChannel? GuildChannels could be voice or category channels, which don't all impleemnt fetch_message, send, and last_message.

Perhaps we should log a warning and abort if the channel is not of the correct type.

bot/exts/help_channels/_cog.py
```diff
@@ -332,6 +354,10 @@ async def move_to_available(self) -> None:
             category_id=constants.Categories.help_available,
         )
 
+        # Adding the help channel to the dynamic message, and editing/sending that message.
+        self.available_help_channels.add(channel.id)
+        await self.update_available_help_channels()
```
HassanAbouelela, line 359: You can save one edit if you remove this update. The next thing that runs after this task is another update.

HassanAbouelela (changes requested):
Everything is looking solid so far. I have a few functionality concerns, but everything works as expected in general.

bot/exts/help_channels/_cog.py
Xithrius: I don't think this is necessary since there's no plans of having a voice channel anywhere near the available help channels.

bot/exts/help_channels/_cog.py
```diff
@@ -461,3 +486,25 @@ async def wait_for_dormant_channel(self) -> discord.TextChannel:
         self.queue_tasks.remove(task)
 
         return channel
+
+    async def update_available_help_channels(self) -> None:
+        """Updates the dynamic message within #how-to-get-help for available help channels."""
+        if not self.available_help_channels:
+            available_channels_category = await channel_utils.try_get_channel(Categories.help_available)
```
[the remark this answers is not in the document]
Xithrius: Forgot that the `self.available_category` attribute exists.

HassanAbouelela (approved):
Clean, and works as expected. LGTM

bot/exts/help_channels/_cog.py
```diff
@@ -20,6 +21,9 @@
 HELP_CHANNEL_TOPIC = """
 This is a Python help channel. You can claim your own help channel in the Python Help: Available category.
 """
+AVAILABLE_HELP_CHANNELS = """
+Currently available help channel(s): {available}
```
Shivansh-007, line 25: How does it look if we do this:
```suggestion
**Currently available help channel(s)**: {available}
```

Shivansh-007 (approved):
Looks Good!

bot/exts/help_channels/_cog.py
Xithrius: image

I prefer the first one.

Shivansh-007: Hmm, i like it with bold, let's see what others prefer.

Vthechamp22: I think it looks fine normally (without bold).

Akarys42: I prefer the bold version to catch the eyes

Xithrius: After further consideration I see the appeal of the bold. I will change this now.

Xithrius:
Here's what the dynamic message now looks like.
image

bot/exts/help_channels/_cog.py
```diff
@@ -114,6 +123,9 @@ async def claim_channel(self, message: discord.Message) -> None:
 
         await _caches.unanswered.set(message.channel.id, True)
 
+        # Removing the help channel from the dynamic message, and editing/sending that message.
+        self.available_help_channels.remove(message.channel.id)
```
Numerlor, line 127: Why are we storing ids instead of the channels themselves?

bot/exts/help_channels/_cog.py
```diff
@@ -20,6 +21,9 @@
 HELP_CHANNEL_TOPIC = """
 This is a Python help channel. You can claim your own help channel in the Python Help: Available category.
 """
+AVAILABLE_HELP_CHANNELS = """
+**Currently available help channel(s):** {available}
+"""
```
Numerlor, line 26: This fits into one line comfortably

bot/exts/help_channels/_cog.py
```diff
@@ -275,6 +287,15 @@ async def init_cog(self) -> None:
...
         self.close_command.enabled = True
 
+        # Getting channels that need to be included in the dynamic message.
+        task = asyncio.create_task(self.update_available_help_channels())
+        self.queue_tasks.append(task)
+
+        await task
+
+        log.trace("Dynamic available help message updated.")
+        self.queue_tasks.remove(task)
```
Numerlor, line 297: What's the reason for creating a task and immediately awaiting it? If it's because of `queue_tasks` and their cancellation then it should be unnecessary, as the init task itself is cancelled when the cog is unloaded

bot/exts/help_channels/_cog.py
```diff
@@ -461,3 +486,24 @@ async def wait_for_dormant_channel(self) -> discord.TextChannel:
         self.queue_tasks.remove(task)
 
         return channel
+
+    async def update_available_help_channels(self) -> None:
+        """Updates the dynamic message within #how-to-get-help for available help channels."""
+        if not self.available_help_channels:
+            self.available_help_channels = set(
+                c.id for c in self.available_category.channels if 'help-' in c.name
```
Numerlor, line 494: Use the `_channel.is_excluded_channel` util

Numerlor (changes requested):

bot/exts/help_channels/_cog.py
mbaruh: +1 to this. It will also make formatting the message more straightforward with channel.metion

bot/exts/help_channels/_cog.py
Xithrius: I thought back then that it would be necessary for this task to be cancelled if the cog was unloaded. 

I now see that there's no significant reason to do so.

bot/exts/help_channels/_cog.py
```diff
@@ -20,6 +21,7 @@
 HELP_CHANNEL_TOPIC = """
 This is a Python help channel. You can claim your own help channel in the Python Help: Available category.
 """
+AVAILABLE_HELP_CHANNELS = """**Currently available help channel(s):** {available}"""
```
Shivansh-007, line 24: Can we spread this across lines?
```suggestion
AVAILABLE_HELP_CHANNELS = """
**Currently available help channel(s):** {available}
"""
```

bot/exts/help_channels/_cog.py
```diff
@@ -461,3 +479,24 @@ async def wait_for_dormant_channel(self) -> discord.TextChannel:
...
+
+        if self.how_to_get_help is None:
+            self.how_to_get_help = await channel_utils.try_get_channel(Channels.how_to_get_help)
+
+        try:
+            if self.dynamic_message is None:
+                self.dynamic_message = await self.how_to_get_help.fetch_message(self.how_to_get_help.last_message_id)
+            await self.dynamic_message.edit(content=available_channels)
+        except discord.NotFound:
+            self.dynamic_message = await self.how_to_get_help.send(available_channels)
```
Shivansh-007, line 502: Can we have a `log.trace()` statement here?

Shivansh-007 (changes requested):

bot/exts/help_channels/_cog.py
Xithrius: https://github.com/python-discord/bot/pull/1396#discussion_r570897703

Shivansh-007: It seems more readable, `"""` are multi line strings, so either change it to `"` single quotes or keep it as it was before
(my opinion)

Xithrius: After looking at other cogs, it does seem that one liners should have single double quotes. An example can be seen [here](<https://github.com/python-discord/bot/blob/e5b073c5ee9c7c887a193ec05d813d259eca58ee/bot/exts/help_channels/_message.py#L30>).

Shivansh-007 (approved):
Thanks! Looks Good To Me!

HassanAbouelela (approved):
Anyone using the server template will need to update their config to move the how-to-get-help channel out of the help available category, but it shouldn't be a big deal. This is not important for the test server, as it uses the same channel for cooldown and how-to-get-help.

gustavwilliam (approved):
Thanks! Works like a charm — unless we go into some edge cases.

When lowering the maximim amount of help channels (restarting bot in-between), the higher number of help channels will be listed in the available help channels message. Let me show you an example, where we go from 3 to 2 channels. After restarting the bot with the new config, the following will happen:

- The bot keeps updating the message with 3 help channels. One of them will be there statically (and incorrectly), while the other two are correctly and continuously updated
- When deleting the original "these channels are available" message, the bot will resend the same three-channel message

One workaround for this is deleting the message and *then* restarting the bot.

The other edge case comes into play when we increase the number of help channels that should be available, like going from 2 to 3. This simply sends a second message with two channels and keeps updating that. Just delete the first message manually and things will work just fine.

However, both of these scenarios are uncommon enough that I'll still approve this PR. If you want to fix the mentioned bugs, that would be fantastic, but I really don't mind leaving it at this point. It works flawlessly, unless we get into some uncommon and strange scenarios.

Xithrius:
Thank you for your approval! 

I think that any of these edge cases will have to include a bot redeploy, as rules would probably have to be changed in order for them to take effect. If we encounter these edge cases, I'll be sure to put it into high/critical priority.

Xithrius:
Bot can fetch a message that was previously deleted, which causes a `discord.NotFound` error since it cannot find the message to edit.

On the actual server, no message has been deleted. The bot this time finds the webhook message, then tries to edit it.

I will make a change to check if the last message was sent by the bot. If not, it will send a dynamic message (or something similar to this)

image

jb3:
Where are we at on this? Should we open a new PR?

Xithrius:
I will be opening a new PR for this, I'm still working out the logic.

PR #1397, opened 2021-02-05
Add jumplink to deleted messages

### LINKED ISSUE #1280: Add link to deleted message for message log
Being able to jump immediately back to when it was deleted would be helpful (especially in the case of messages for which the content is no longer available) since this is the default when discord gets a jumplink to a message that does not exist (except for on android, but it would be useful enough on other platforms I think that's worth ignoring.

Senjan21:
Closes #1280 

Implements jumplinks to deleted messages in logs.


Akarys42 (approved):
Lgtm

Xithrius (approved):
Thank you @Senjan21, very cool.

PR #1398, opened 2021-02-05
Make `KeyError` tag

anand2312:


bot/resources/tags/keyerror.md
```diff
@@ -0,0 +1,17 @@
+Often while using dictionaries in Python, you may run into `KeyErrors`. This error is raised when you try to access a key that isn't present in your dictionary. \
+While you can use a `try` and `except` block to catch the `KeyError`, Python also gives you some other neat ways to handle them.
+## __The `dict.get` method__
```
Akarys42, line 3: Discord doesn’t support markdown titles sadly 
```suggestion

**The `dict.get` method**

```

bot/resources/tags/keyerror.md
```diff
@@ -0,0 +1,17 @@
+Often while using dictionaries in Python, you may run into `KeyErrors`. This error is raised when you try to access a key that isn't present in your dictionary. \
+While you can use a `try` and `except` block to catch the `KeyError`, Python also gives you some other neat ways to handle them.
```
Akarys42, line 2: I believe there is one too many spaces, since there is one on this line and one on the previous line before the backslash.

bot/resources/tags/keyerror.md
```diff
@@ -0,0 +1,17 @@
...
+## __The `dict.get` method__
+The [dict.get](https://docs.python.org/3/library/stdtypes.html#dict.get) method will return the value for the key if it exists, or None (or a default value that you specify) if the key doesn't exist. Hence it will _never raise_ a KeyError.
```
Akarys42, line 4: Let’s put an inline codeblock around the function names.
```suggestion
The [`dict.get`](https://docs.python.org/3/library/stdtypes.html#dict.get) method will return the value for the key if it exists, or None (or a default value that you specify) if the key doesn't exist. Hence it will _never raise_ a KeyError.
```

bot/resources/tags/keyerror.md
```diff
@@ -0,0 +1,17 @@
...
+>>> print(my_dict.get("foo"))
+1
+>>> print(my_dict.get("foobar"))
+None
+>>> print(my_dict.get("foobar", 3))    # here 3 is the default value to be returned, in case the key doesn't exist
+3
+>>> print(my_dict)
+{'foo': 1, 'bar': 2}    # note that the new key was NOT added to the dictionary
+```
+\
```
Akarys42, line 16: This should be an empty line to visually split the blocks and keep it readable.

bot/resources/tags/keyerror.md
```diff
@@ -0,0 +1,17 @@
...
+Some other methods that can be used for handling KeyErrors gracefully are the [dict.setdefault](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) method, or by using [collections.defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict).
```
Akarys42, line 17: Some missing inline codeblocks. 
```suggestion
Some other methods that can be used for handling `KeyErrors` gracefully are the [`dict.setdefault`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) method, or by using [`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict).
```

Akarys42 (changes requested):
Hello, thank you for your Pull Request! The tag content is quite good, although I wonder if it would be worth renaming the tag to `dict-get` or something like that. 

I left a few comments related to formatting.

Shivansh-007:
Maybe include a example of `KeyErrors` being raised.
```py
>>> ages = {'Jim': 30, 'Pam': 28}
>>> ages['Michael']
  File "<stdin>", line 1, in <module>
KeyError: 'Michael'
```

bot/resources/tags/keyerror.md
anand2312: Ah, figured the title would be used in the `Embed`s title field

bot/resources/tags/dict-get.md
```diff
@@ -0,0 +1,15 @@
...
+While you can use a `try` and `except` block to catch the `KeyError`, Python also gives you some other neat ways to handle them.\
+
+__**The `dict.get` method**__
+
+The [`dict.get`](https://docs.python.org/3/library/stdtypes.html#dict.get) method will return the value for the key if it exists, or None (or a default value that you specify) if the key doesn't exist. Hence it will _never raise_ a KeyError.
+```py
+>>> my_dict = {"foo": 1, "bar": 2}
+>>> print(my_dict.get("foobar"))
+None
+>>> print(my_dict.get("foobar", 3))    # here 3 is the default value to be returned, in case the key doesn't exist
```
Akarys42, line 11:
```suggestion
>>> print(my_dict.get("foobar", 3))    # here 3 is the default value to be returned, because the key doesn't exist
```

bot/resources/tags/dict-get.md
```diff
@@ -0,0 +1,15 @@
+Often while using dictionaries in Python, you may run into `KeyErrors`. This error is raised when you try to access a key that isn't present in your dictionary.\
+While you can use a `try` and `except` block to catch the `KeyError`, Python also gives you some other neat ways to handle them.\
```
Akarys42, line 2: This shouldn't contain a backslash since we want to have a line break
```suggestion
While you can use a `try` and `except` block to catch the `KeyError`, Python also gives you some other neat ways to handle them.
```

Akarys42 (changes requested):
a couple of things

bot/resources/tags/dict-get.md
```diff
@@ -0,0 +1,15 @@
+Often while using dictionaries in Python, you may run into `KeyErrors`. This error is raised when you try to access a key that isn't present in your dictionary.\
+While you can use a `try` and `except` block to catch the `KeyError`, Python also gives you some other neat ways to handle them.
```
gustavwilliam, line 2:
```suggestion
While using dictionaries in Python, you may run into `KeyError`s. This is raised when you try to access a dictionary item that isn't present in the dictionary. Python gives you some neat ways to handle them.
```
This could be condensed and more clear.

bot/resources/tags/dict-get.md
```diff
@@ -0,0 +1,15 @@
...
+
+__**The `dict.get` method**__
```
gustavwilliam, line 4:
```suggestion
**Providing default values**
```
We don't really underline sections in the tags. They're generally just bolded, so this would seem more consistent.

The change of wording isn't necessary, but it's something to consider.

bot/resources/tags/dict-get.md
```diff
@@ -0,0 +1,15 @@
...
+
+The [`dict.get`](https://docs.python.org/3/library/stdtypes.html#dict.get) method will return the value for the key if it exists, or None (or a default value that you specify) if the key doesn't exist. Hence it will _never raise_ a KeyError.
```
gustavwilliam, line 6:
```suggestion
The [`dict.get`](https://docs.python.org/3/library/stdtypes.html#dict.get) method will return the value corresponding to the key, if it exists, and None (or a default value that you provide) if the it doesn't. Hence it will _never raise_ a KeyError.
```

bot/resources/tags/dict-get.md
```diff
@@ -0,0 +1,15 @@
...
+```py
+>>> my_dict = {"foo": 1, "bar": 2}
+>>> print(my_dict.get("foobar"))
+None
+>>> print(my_dict.get("foobar", 3))    # here 3 is the default value to be returned, because the key doesn't exist
+3
```
gustavwilliam, line 12:
```suggestion
>>> print(my_dict.get("foobar", 3))  # Here, 3 is specified as the default value
3  # The default value is returned, since the key doesn't exist
```
Let's make sure that the comment follows PEP 8. I'd also like to see this split up in a way like the above, but I'll leave the details to you.

bot/resources/tags/dict-get.md
```diff
@@ -0,0 +1,15 @@
...
+```
+
```
gustavwilliam, line 14:
```suggestion
```
This creates a double space when rendered on Discord. Let's remove this space.

bot/resources/tags/dict-get.md
```diff
@@ -0,0 +1,15 @@
...
+Some other methods that can be used for handling KeyErrors gracefully are the [`dict.setdefault`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) method, or by using [`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict) (check out the `!defaultdict` tag).
```
gustavwilliam, line 15:
```suggestion
Some other methods for handling `KeyError`s gracefully are the [`dict.setdefault`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) method and [`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict) (check out the `!defaultdict` tag).
```
Let's make this less verbose.

gustavwilliam (changes requested):
Thanks for the PR! This seems like a highly useful tag to have.

The tag looks good in general, but I have a few thoughts how we could perfect this, before going live.

bot/resources/tags/dict-get.md
anand2312: I was told to mention the `try - except` option in a single line here, do you think I should omit it?

bot/resources/tags/dict-get.md
anand2312: I didn't know how to split the comment whilst also keeping to the REPL look (with the `>>>`), wouldn't a lone comment look a bit odd?

bot/resources/tags/dict-get.md
gustavwilliam: If you want to do so, include it. I personally think it's a bit redundant.

bot/resources/tags/dict-get.md
gustavwilliam: Hmmm. This looks fine to me. It might also be worth considering putting the explanation outside of the code block, if you don't like putting the comment on the new line.

Akarys42 (approved):
Lgtm, thank you!

Shivansh-007 (approved):
Thanks! Looks Good To Me!

bot/resources/tags/dict-get.md
Vthechamp22: `...the key, if it exists, and None (or a default value that you provide) if it doesn't...` Small grammar correction :smile_cat:

Xithrius:
Maybe we should split this into two different code blocks so this can be more easily read?

image

Xithrius (approved):
Resolved all unresolved discussions, since almost every suggestion was eventually put into this new embed, or worked out.

Cool stuff!

PR #1400, opened 2021-02-05
Make `defaultdict` tag

anand2312:
Messed up and made a separate branch for this, my bad.
I've edited the `dict-get` tag to also refer to this.

### DIFF
--- a/bot/resources/tags/defaultdict.md
+++ b/bot/resources/tags/defaultdict.md
@@ -1,6 +1,6 @@
 **[`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict)**
 
-The Python `defaultdict` type behaves almost exactly like a regular Python dictionary, but if you try to access or modify a missing key, the `defaultdict` will automatically create the key and generate a default value for it. 
+The Python `defaultdict` type behaves almost exactly like a regular Python dictionary, but if you try to access or modify a missing key, the `defaultdict` will automatically create the key and generate a default value for it.
 While instantiating a `defaultdict`, we pass in a function that tells it how to create a default value for missing keys.
 
 ```py
--- a/bot/resources/tags/defaultdict.md
+++ b/bot/resources/tags/defaultdict.md
@@ -1,20 +1,21 @@
 **[`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict)**
 
-The Python `defaultdict` type behaves almost exactly like a regular Python dictionary, but if you try to access or modify a missing key, the `defaultdict` will automatically create the key and generate a default value for it.
+The Python `defaultdict` type behaves almost exactly like a regular Python dictionary, but if you try to access or modify a missing key, the `defaultdict` will automatically insert the key and generate a default value for it.
 While instantiating a `defaultdict`, we pass in a function that tells it how to create a default value for missing keys.
 
 ```py
 >>> from collections import defaultdict
->>> my_dict = defaultdict(int, {"foo": 1, "bar": 2})
->>> print(my_dict)
-defaultdict(<class 'int'>, {'foo': 1, 'bar': 2})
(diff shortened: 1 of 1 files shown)


swfarnsworth:
For reference, this is what one would see if they evoke the `!docs collections.defaultdict` command that is available currently.

image

I agree that we can make one that's better for our purposes.

bot/resources/tags/defaultdict.md
```diff
@@ -0,0 +1,20 @@
+**[`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict)**
+
+The Python `defaultdict` type behaves almost exactly like a regular Python dictionary, but if you try to access or modify a missing key, the `defaultdict` will automatically create the key and generate a default value for it. 
+While instantiating a `defaultdict`, we pass in a function that tells it how to create a default value for missing keys.
+
+```py
+>>> from collections import defaultdict
+>>> my_dict = defaultdict(int, {"foo": 1, "bar": 2})
```
swfarnsworth, line 8: I don't think it's very popular to instantiate a defaultdict from another dict, so it may be more useful to just have `my_dict = defaultdict(int)`. Otherwise people might erroneously assume that passing an empty dict is necessary if they don't need any key-values to start, or something.

bot/resources/tags/defaultdict.md
```diff
@@ -0,0 +1,20 @@
...
+>>> print(my_dict)
```
swfarnsworth, line 9: In a repl setting, `my_dict` without the call to `print` would display the value.

bot/resources/tags/defaultdict.md
```diff
@@ -0,0 +1,20 @@
...
+defaultdict(<class 'int'>, {'foo': 1, 'bar': 2})
+```
+
+In this example, we've used the `int` function - this means that if we try to access a non-existent key, it provides the default value of 0.
```
swfarnsworth, line 13: Calling `int` a function in this context is dubious because it is an instance of `type` that returns `0` when called. You might consider:
> In this example, we've used the `int` class, which returns `0` when called like a function, so any missing key will get a default value of `0`. You can also get an empty list by default with `list` or an empty string with `str`.

bot/resources/tags/defaultdict.md
```diff
@@ -0,0 +1,20 @@
...
+
+```py
+>>> print(my_dict["foobar"])
+0
+>>> print(my_dict)
+defaultdict(<class 'int'>, {'foo': 1, 'bar': 2, 'foobar': 0})
```
swfarnsworth, line 19: If you're going to use `int` as the default factory in the example, I would do some calculation that uses `+=`, since that's how `defaultdict(int)`s often get used. Note also my earlier suggestion about using `print` in repls.

bot/resources/tags/defaultdict.md
```diff
@@ -0,0 +1,20 @@
+**[`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict)**
+
+The Python `defaultdict` type behaves almost exactly like a regular Python dictionary, but if you try to access or modify a missing key, the `defaultdict` will automatically create the key and generate a default value for it.
```
swfarnsworth: The key is not being created; the key is being entered into the dict and a default value is created.

I think everyone will know what you mean by "access or modify a missing key", though one could argue that it is the *value* that is being accessed or modified via a certain key. Though it's probably not worth adding extra words to uphold that distinction.

swfarnsworth (changes requested):
Thank you for submitting this PR; I hope that my suggestions don't obscure the appreciation that I have for the time you've taken to make it. Let me know what you think and we can proceed from there.

swfarnsworth (approved):
Thanks!

kosayoda (approved):
Good job.

PR #1402, opened 2021-02-06
Implement showing filterlist entry comment in alerts

### LINKED ISSUE #1394: Mention the reason on watchlist triggers
We have quite a few watchlist triggers, such as regexes and domains which will ping the moderators when someone post a message matching our denylist. 

One of the issue we are currently facing is that it can be hard to know why we watch that particular word or domain. We already allow adding reasons when adding filters to our denylist, but we don’t display them when pinging the mods. We should totally do that!

ks129:
Closes #1394 

This PR make filtering functions returning reason about triggered entry what will be shown in main embed (except for guild invites what show this in guild info embed):
GH-Guild-Blacklist
GH-URL-blacklist

### DIFF
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -426,6 +426,7 @@ class Filtering(Cog):
 
         `word_watchlist`'s patterns are placed between word boundaries while `token_watchlist` is
         matched as-is. Spoilers are expanded, if any, and URLs are ignored.
+        Second return value is a reason written to database about blacklist entry (can be None).
         """
         if SPOILER_RE.search(text):
             text = self._expand_spoilers(text)
@@ -443,7 +444,11 @@ class Filtering(Cog):
         return False, None
 
     async def _has_urls(self, text: str) -> Tuple[bool, Optional[str]]:
-        """Returns True if the text contains one of the blacklisted URLs from the config file."""
+        """
+        Returns True if the text contains one of the blacklisted URLs from the config file.
+
+        Second return value is a reason of URL blacklisting (can be None).
+        """
         if not URL_RE.search(text):
             return False, None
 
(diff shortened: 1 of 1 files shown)


kosayoda (changes requested):
Couple comments:
1. There are a few missed conditional branches for filters that do not return a reason:
https://github.com/python-discord/bot/blob/81af0099ffd552aa5cb2a61de30cf7bd16a013eb/bot/exts/filters/filtering.py#L435
https://github.com/python-discord/bot/blob/81af0099ffd552aa5cb2a61de30cf7bd16a013eb/bot/exts/filters/filtering.py#L448
https://github.com/python-discord/bot/blob/81af0099ffd552aa5cb2a61de30cf7bd16a013eb/bot/exts/filters/filtering.py#L497

2. The majority of filters do not document the second return value, it'll be nice to have some form of documentation.

Good job overall :+1:

mbaruh:
Why do invites need a special case?

ks129:
@mbaruh If one message contains many invites, then every invite is posted as a separate embed. Reasons is a different for every guild blacklist entry (I suppose). But sure, I'll change wording.

mbaruh (commented):
Can you move the reason for invites to the footer as well?

ks129:
@mbaruh Do you mean something like: `Guild ID: ... | Reason: ...`?

mbaruh:
@ks129 yeah. Maybe the other way around

ks129:
@mbaruh Moved guild reason to footer

mbaruh:
Always returning None by certain functions, only to match the signature of other unrelated functions seems a bit iffy to me. I'm wondering if it's better to not unpack the return value right away and check if the return value is a tuple.

ks129:
@mbaruh Now this doesn't have compatibility None returns anymore.

mbaruh (approved):
Seems to be working as intended.

Xithrius (approved):
It works very well.

kosayoda (approved):

PR #1404, opened 2021-02-06
Ignore attachment-only messages for duplicates antispam rule

### LINKED ISSUE #1356: Add an `attachments` rule
# Abstract 

We should have an antispam rule filtering small burst of images. 

# Rationale

Currently, when a user posts 4 images in less than 10 seconds without any comment, the `duplicates` rule will trigger. While we still want to be informed when many images are posted, having the `duplicates` rule trigger doesn't make much sense. Besides, if different message content is given for each image, it will only trigger `burst` if more than 9 messages are sent in 10 seconds. 

# Specification

- [ ] Make sure that the `duplicates` filter won't be triggered by messages with images. We can safely skip empty messages with attachments.
- [ ] Create an `images` filter based on `duplicates` that will trigger when more than 3 images are posted in less than 10 seconds. It should ignore the message content.

ks129:
Closes #1356 because we already have attachments rule, but this still should ignore attachment-only messages in duplicates.

### DIFF
--- a/bot/rules/duplicates.py
+++ b/bot/rules/duplicates.py
@@ -13,7 +13,7 @@ async def apply(
         if (
             msg.author == last_message.author
             and msg.content == last_message.content
-            and (msg.content and not msg.attachments)
+            and msg.content
         )
     )
 
--- a/bot/rules/duplicates.py
+++ b/bot/rules/duplicates.py
@@ -13,6 +13,7 @@ async def apply(
         if (
             msg.author == last_message.author
             and msg.content == last_message.content
+            and (msg.content and not msg.attachments)
         )
     )
 
--- a/config-default.yml
+++ b/config-default.yml
@@ -367,7 +367,7 @@ anti_spam:
     rules:
         attachments:
             interval: 10
-            max: 3
+            max: 6
 
         burst:
             interval: 10
--- a/config-default.yml
+++ b/config-default.yml
@@ -367,7 +367,7 @@ anti_spam:
     rules:
         attachments:
             interval: 10
-            max: 9
+            max: 3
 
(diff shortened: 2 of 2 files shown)


kosayoda (approved):
:+1: Tested locally.

Shivansh-007 (approved):
Thanks! Looks Good To Me!

bot/rules/duplicates.py
```diff
@@ -13,6 +13,7 @@ async def apply(
         if (
             msg.author == last_message.author
             and msg.content == last_message.content
+            and (msg.content and not msg.attachments)
```
mbaruh, line 16: The attachments filter checks whether too many attachments are sent. This isn't mutually exclusive with duplicate messages, as someone might still send the same text but with a message attached. So it seems to me like this should just check that there is any text at all, and not skip messages with attachments.

config-default.yml
```diff
@@ -367,7 +367,7 @@ anti_spam:
     rules:
         attachments:
             interval: 10
-            max: 9
+            max: 3
```
mbaruh, line 370: This seems too strict to me. Most of filter trips are for people who send a lot of screenshots in help channels, or a lot of design images in the branding channel. Let's start with 6 and see how many false positives to false negatives we have.

mbaruh (changes requested):

config-default.yml
Akarys42: Fwiw the idea was to keep the current behavior but use the right filter when someone does spam images. I'm fine with changing the threshold though.

mbaruh (approved):

mbaruh (approved):

PR #1405, opened 2021-02-06
New wording for some tag messages

swfarnsworth:
I rewrote three tags for reasons described in the respective commit messages. The content I'm submitting is informed by my belief that tag messages should be specific so that we're not trying to address too many potential questions in the relatively small amount of space that we have.

### DIFF
--- a/bot/resources/tags/dictcomps.md
+++ b/bot/resources/tags/dictcomps.md
@@ -11,5 +11,4 @@ One can use a dict comp to change an existing dictionary using its `items` metho
 >>> {key.upper(): value * 2 for key, value in first_dict.items()}
 {'I': 2, 'LOVE': 8, 'PYTHON': 12}
 ```
-
 For more information and examples, check out [PEP 274](https://www.python.org/dev/peps/pep-0274/)
--- a/bot/resources/tags/dictcomps.md
+++ b/bot/resources/tags/dictcomps.md
@@ -1,20 +1,15 @@
-**Dictionary Comprehensions**
-
-Like lists, there is a convenient way of creating dictionaries:
+Dictionary comprehensions (*dict comps*) provide a convenient way to make dictionaries, just like list comps:
 ```py
->>> ftoc = {f: round((5/9)*(f-32)) for f in range(-40,101,20)}
->>> print(ftoc)
-{-40: -40, -20: -29, 0: -18, 20: -7, 40: 4, 60: 16, 80: 27, 100: 38}
+>>> {word.lower(): len(word) for word in ('I', 'love', 'Python')}
+{'i': 1, 'love': 4, 'python': 6}
 ```
-In the example above, I created a dictionary of temperatures in Fahrenheit, that are mapped to (*roughly*) their Celsius counterpart within a small range. These comprehensions are useful for succinctly creating dictionaries from some other sequence.
+The syntax is very similar to list comps except that you surround it with curly braces and have two expressions: one for the key and one for the value.
 
-They are also very useful for inverting the key value pairs of a dictionary that already exists, such that the value in the old dictionary is now the key, and the corresponding key is now its value:
+One can use a dict comp to change an existing dictionary using its `items` method
 ```py
->>> ctof = {v:k for k, v in ftoc.items()}
->>> print(ctof)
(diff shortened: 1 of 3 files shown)


bot/resources/tags/dictcomps.md
```diff
@@ -1,20 +1,15 @@
...
-They are also very useful for inverting the key value pairs of a dictionary that already exists, such that the value in the old dictionary is now the key, and the corresponding key is now its value:
+One can use a dict comp to change an existing dictionary using its `items` method
 ```py
->>> ctof = {v:k for k, v in ftoc.items()}
->>> print(ctof)
-{-40: -40, -29: -20, -18: 0, -7: 20, 4: 40, 16: 60, 27: 80, 38: 100}
+>>> first_dict = {'i': 1, 'love': 4, 'python': 6}
+>>> {key.upper(): value * 2 for key, value in first_dict.items()}
+{'I': 2, 'LOVE': 8, 'PYTHON': 12}
 ```
```
gustavwilliam, line 14:
```suggestion
```
This results in two spaces after the code block, like in the following screenshot. Let's remove it.
Screenshot 2021-02-06 at 22 16 55
With one space, it instead looks like this:
Screenshot 2021-02-06 at 22 17 19

bot/resources/tags/f-strings.md
```diff
@@ -1,17 +1,9 @@
...
-print("{0} are some of the largest snakes in the world".format(snake))
-
-# Or keyword arguments
-print("{family} are some of the largest snakes in the world".format(family=snake))
+>>> snake = "pythons"
+>>> number = 21
+>>> f"There are {number * 2} {snake} on the plane."
+"There are 42 pythons on the plane."
 ```
+Note that even when you include an expression that isn't a string, like `number * 2`, Python will handle converting it to a string.
```
gustavwilliam, line 9:
```suggestion
Note that even when you include an expression that isn't a string, like `number * 2`, Python will convert it to a string for you.
```
Both work, but this version seems more direct.

bot/resources/tags/listcomps.md
```diff
@@ -1,14 +1,19 @@
...
-even_numbers = [n for n in range(20) if n % 2 == 0]
+>>> [n ** 2 for n in range(5)]
+[0, 1, 4, 9, 16]
+```
+List comprehensions also get an `if` statement:
+```python
+>>> [n ** 2 for n in range(5) if n % 2 == 0]
+[0, 4, 16]
 ```
-This also works for generators, dicts and sets by using `()` or `{}` instead of `[]`.
```
gustavwilliam, line 12: Is there a reason to remove this piece of information? We may want to say that it works in a similar way, instead of that it work out of the box, for the other types. From using the tag in the community, I've found this to be a nice addition to the tag, by showing them that there's more to the comps than just list comps. It's a pretty nice fact.

gustavwilliam (changes requested):
Overall, you've significantly reduced the amount of clutter. I like this change. There are just a few minor things to consider before I'm fully satisfied.

bot/resources/tags/listcomps.md
swfarnsworth: Other types of comprehensions have differences that people need to know about in order to use them and given that it would take too many words to describe those differences, I think we shouldn't refer to them at all. I think this is especially true for generator expressions, as I think most people learn list and dict comprehensions (and maybe set comprehensions as well) before they learn about lazy iteration.

`!dictcomp` is another tag.

gustavwilliam: I think you're right that they have quite a few differences that you need to know. However, this is something that I actually found very useful when reading the tag a while back. List comps are pretty common, so mentioning that there are similar things and underlining the fact that there are differences and that they'll need to do some further resource is crucial. At the same time, I see this as a great opportunity to sneak in some "you didn't know you wanted to know this, but you really want to know this" kind of content.

gustavwilliam (approved):
I would love to see a mention of generators, dicts and sets as well, but I'll leave it to you to decide. Either way, I'm positive to this change. Thanks!

Xithrius (approved):
Good idea on simplifying examples, should be much easier for people to understand them now.

PR #1409, opened 2021-02-09
Makes Off Topic Name Search Insensitive

### LINKED ISSUE #1408: make the "!otn s" command non case sensitive
make the "!otn s" command non case sensitive for easier querying. We would do this by casting the query into lowercase using `str.lower()` on the search, therefor making the command non case sensitive, because channel names are also stored in full lowercase.

HassanAbouelela:
Closes #1408.

This PR allows the search command for otn to accept punctuation and differently cased letters. In order to accomplish this, results from the database, and the user query run through the reverse translation used to make channel safe names, before being passed through the regular searching.

Results displayed to the user are the same names stored in the database.


kosayoda (approved):
:+1: Tested locally. Not quite sure why we translate the query since we'll almost always search using regular characters, but there's no harm so it's fine.

Senjan21 (approved):
Looks good, nice addition of the comments, thanks!

PR #1411, opened 2021-02-10
Internal API routing

jb3:
Migrates away from the API fetching through Cloudflare to use cluster-local DNS and hit the Django service directly.

- Migrate API utilities to use internal DNS routing
- Migrate ping command to ping internal API

### DIFF
--- a/config-default.yml
+++ b/config-default.yml
@@ -335,9 +335,10 @@ keys:
 urls:
     # PyDis site vars
     site:        &DOMAIN       "pythondiscord.com"
-    site_api:    &API    !JOIN ["api.", *DOMAIN]
+    site_api:    &API          "pydis-api.default.svc.cluster.local"
     site_paste:  &PASTE  !JOIN ["paste.", *DOMAIN]
     site_schema: &SCHEMA       "https://"
+    site_api_schema:           "http://"
     site_staff:  &STAFF  !JOIN ["staff.", *DOMAIN]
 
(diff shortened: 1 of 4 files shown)


config-default.yml
```diff
@@ -335,9 +335,10 @@ keys:
 urls:
     # PyDis site vars
     site:        &DOMAIN       "pythondiscord.com"
-    site_api:    &API    !JOIN ["api.", *DOMAIN]
+    site_api:    &API          "pydis-api.default.svc.cluster.local"
     site_paste:  &PASTE  !JOIN ["paste.", *DOMAIN]
     site_schema: &SCHEMA       "https://"
+    site_api_schema:           "http://"
```
MarkKoz, line 341: Please keep it alphabetically sorted.

MarkKoz (changes requested):
Can you update contrib guidelines to mention that the `site_api` has to be set too? I think it just needs to state that the value is identical to `site` for development purposes.

Checking if the production stuff is adequately set up for this is out of my scope.

Den4200 (approved):
Other than Mark's requested changes, the code is solid.

config-default.yml
jb3: Resolved in 9111f9f

jb3:
Contribution guides at https://pythondiscord.com/pages/contributing/bot/ have been updated.

MarkKoz (approved):

PR #1414, opened 2021-02-11
Reworked dynamic available help message logic.

### LINKED ISSUE #1320: Add a dynamic message at the bottom of `#how-to-get-help`

Xithrius:
Closes #1320 
First iteration: #1325

In the previous iteration of this PR, #1396, the branch wasn't tested enough by me, causing many errors. One is shown below.

image

This was caused by the bot trying to edit the webhook in #how-to-get-help, which I didn't think was going to happen. On the test server, fetching the previous message with `message.last_message` could get a deleted message. 

This caused me to think that the message fetched wouldn't get one sent by a webhook.

In final iteration (at least for now) of this feature, the bot first checks if the dynamic message is set at all. If it is, it'll go ahead and attempt to edit the message. If the message isn't found, that means it doesn't exist. It'll then send a message, set the dynamic message to it's ID, then save it within the Redis cache.


kosayoda (changes requested):
Helo, apparently `.history` still gets deleted messages. I get a `NotFound` if I run it, remove the first generated message, then update the channels.

Xithrius:
That's odd, I'll look into it.

kwzrd:
The approach taken is perfectly fine, I'd just like to mention the option of persisting the message ID with Redis, since it seems that we could remove some complexity that way. In fact, as long as you know the message ID (in Redis) and channel ID (in config), editing a message is possible with a direct call:

```python
self.bot.http.edit_message(channel_id, message_id, content=new_content)
```

This avoids having to acquire and store the channel and message objects, allowing you to work with just the IDs, and you never have to "guess" which message is the one you want to edit. The downside is that I believe it's undocumented.

The logic could (very roughly) be:

```python
fresh_content = "Available channels: ..."
message_id: t.Optional[int] = self.message_cache.get("available_channels")  # Probably needs better names

if message_id is not None:  # We know the message ID ~ message should exist
    try:
        self.bot.http.edit_message(channel_id, message_id, content=fresh_content)
    except NotFound:  # Someone must have deleted the message, we'll have to make a new one
        pass
    else:
        return  # We're done!

# If we reach here, either we don't have a message ID (we never sent a message),
# or the cached message ID is expired (someone deleted the message)
m = channel.send(fresh_content)  # Can optionally also use a HTTP method to avoid needing a channel obj
self.message_cache.put("available_channels", m.id)  # Cache the latest message ID for future use
```

This way there is no "warm-up" logic on bot restart and you avoid having to acquire the objects. But I don't know whether it's ok to use undocumented methods for this project. Either way, just thought I'd mention that this is possible.

Xithrius:
@kwzrd Thank you for this information! 

I'll see where I can get with Redis caching, and hopefully get this implemented.

Xithrius:
I'm coming back to this, just figuring out Redis things. Should be done relatively soon.

kosayoda (changes requested):
Haven't tested yet, but I noticed that we're defining caches in [bot/exts/help_channels/_caches.py](https://github.com/python-discord/bot/blob/master/bot/exts/help_channels/_caches.py), so it would be a good idea to have that there too.
Going by convention then, the namespace could be `HelpChannels.dynamic_message`? I'd also suggest simplifying the redis key, maybe `message_id` or just `id`

bot/exts/help_channels/_cog.py
```diff
@@ -465,3 +488,30 @@ async def wait_for_dormant_channel(self) -> discord.TextChannel:
...
+        available_channels = AVAILABLE_HELP_CHANNELS.format(
+            available=', '.join(c.mention for c in self.available_help_channels) or None
+        )
+
+        if self.dynamic_message is not None:
+            try:
+                await self.bot.http.edit_message(
+                    constants.Channels.how_to_get_help, self.dynamic_message, content=available_channels
+                )
+                log.trace("Help channels have changed, dynamic message has been edited.")
```
Shivansh-007, line 508: "Help channels have been updated" Sounds more accurate imo.

Shivansh-007 (changes requested):
Code looks good and it works pretty well too. Except one change.

bot/exts/help_channels/_cog.py
```diff
@@ -72,6 +74,12 @@ def __init__(self, bot: Bot):
 
         self.last_notification: t.Optional[datetime] = None
 
+        # Caching the message object for the dynamic message.
+        self.dynamic_message_cache = async_rediscache.RedisCache(namespace="Dynamic Message")
```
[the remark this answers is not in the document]
Xithrius: Yep. See [this comment](<https://github.com/python-discord/bot/pull/1414#pullrequestreview-606123644>).

bot/exts/help_channels/_cog.py
Xithrius: Could you explain what you mean by "more accurate"? I don't really see the main purpose of changing between "changed" to "updated".

Shivansh-007: I got the sentence wrong, nvm.

bot/exts/help_channels/_cog.py
```diff
@@ -465,3 +484,32 @@ async def wait_for_dormant_channel(self) -> discord.TextChannel:
...
+                await self.bot.http.edit_message(
+                    constants.Channels.how_to_get_help, self.dynamic_message, content=available_channels
+                )
+                log.trace("Help channels have changed, dynamic message has been edited.")
+            except discord.NotFound:
+                pass
+            else:
+                return
+
+        log.trace("No How-to-get-help dynamic message could be found in the Redis cache. Setting a new one.")
```
kwzrd, line 510: This log is misleading; we hit this line either in the described case or when the message ID was there but the edit failed.

bot/exts/help_channels/_cog.py
```diff
@@ -465,3 +484,32 @@ async def wait_for_dormant_channel(self) -> discord.TextChannel:
...
+        available_channels = AVAILABLE_HELP_CHANNELS.format(
+            available=', '.join(c.mention for c in self.available_help_channels) or None
+        )
+
+        if self.dynamic_message is not None:
+            try:
+                await self.bot.http.edit_message(
+                    constants.Channels.how_to_get_help, self.dynamic_message, content=available_channels
+                )
+                log.trace("Help channels have changed, dynamic message has been edited.")
```
kwzrd, line 504: I think it's better for this log to come before the edit call. If it's after, nothing will be logged if the call fails.

bot/exts/help_channels/_cog.py
```diff
@@ -118,6 +122,9 @@ async def claim_channel(self, message: discord.Message) -> None:
 
         await _caches.unanswered.set(message.channel.id, True)
 
+        # Removing the help channel from the dynamic message, and editing/sending that message.
+        self.available_help_channels.remove(message.channel)
+
```
kwzrd, line 127: I think this forgets to call the update function?

bot/exts/help_channels/_cog.py
```diff
@@ -72,6 +73,9 @@ def __init__(self, bot: Bot):
 
         self.last_notification: t.Optional[datetime] = None
 
+        self.dynamic_message: t.Optional[int] = None
+        self.available_help_channels: t.Set[discord.TextChannel] = set()
```
kwzrd, line 77: I think you don't need to keep the `dynamic_message` attribute on the class ~ you can just retrieve it from the cache every time you need it (inside the update function). Although this will have a little less overhead, so maybe it makes sense.

Regarding the `available_help_channels` attribute. In terms of code cleanliness I think it would be a lot nicer to always look at the `d.py` representation (as we do when the set isn't populated yet) rather than keeping our own state, but I understand that it may lag behind as `d.py` won't update the channels in the category when we send the request, but when some event comes back from Discord? So this is probably better, it just comes with the burden of having to make sure that we don't accidentally forget to update the set somewhere. Furthermore, it will desync if an admin moves a channel manually.

kwzrd (commented):
Looks great, I think the caching reduced the solution's complexity a lot, which is always good.

bot/exts/help_channels/_cog.py
```diff
@@ -279,6 +286,14 @@ async def init_cog(self) -> None:
         # This may confuse users. So would potentially long delays for the cog to become ready.
         self.close_command.enabled = True
 
+        # Acquiring the dynamic message ID, if it exists within the cache.
+        log.trace("Attempting to fetch How-to-get-help dynamic message ID.")
+        self.dynamic_message = await _caches.dynamic_message.get("message_id")
+
+        # Getting channels that need to be included in the dynamic message.
+        await self.update_available_help_channels()
+        log.trace("Dynamic available help message updated.")
```
kwzrd, line 295: Perhaps this should be moved down below the `init_available` call. Makes more sense to wait for the channels to be synced before we edit the message.

bot/exts/help_channels/_cog.py
kosayoda: I actually have a different opinion on this one, I think the log should come after, since the whole purpose of logging is to record events, ie. if we see the log it means *x* happened, if it doesn't then we never reached that point.

kwzrd: I'd log before the edit, then both in the `except` and the `else` clause to document the outcome. :smile:

bot/exts/help_channels/_cog.py
Xithrius: Actually, the call is not needed here. Works perfectly fine without it.

bot/exts/help_channels/_cog.py
Xithrius: Will do, thanks!

bot/exts/help_channels/_cog.py
Xithrius: I'll put it above the edit log.

MarkKoz (changes requested):

bot/exts/help_channels/_cog.py
```diff
@@ -465,3 +484,32 @@ async def wait_for_dormant_channel(self) -> discord.TextChannel:
...
+        )
+
+        if self.dynamic_message is not None:
+            try:
+                log.trace("Help channels have changed, dynamic message has been edited.")
+                await self.bot.http.edit_message(
+                    constants.Channels.how_to_get_help, self.dynamic_message, content=available_channels
+                )
+            except discord.NotFound:
+                pass
```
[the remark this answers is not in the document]
Xithrius: Thank you for this information. I'll see what I can do.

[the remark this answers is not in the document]
Xithrius: @MarkKoz I'm going to go ahead and leave the channel "checking" as it is.

As for missing messages, we receive a `discord.errors.NotFound: 404 Not Found (error code: 10008): Unknown Message` (tested with random IDs). I haven't seen an instance of where the return code is a 400.

bot/exts/help_channels/_cog.py
```diff
@@ -472,3 +491,32 @@ async def wait_for_dormant_channel(self) -> discord.TextChannel:
...
+            else:
+                return
+
+        log.trace("Dynamic message could not be edited or found. Creating a new one.")
+        new_dynamic_message = await self.bot.http.send_message(
+            constants.Channels.how_to_get_help, available_channels
```
MarkKoz, line 519: This constant no longer exists.

bot/exts/help_channels/_cog.py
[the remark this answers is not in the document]
MarkKoz: Okay, thanks for looking into it.

MarkKoz (changes requested):

bot/exts/help_channels/_cog.py
Xithrius: Added in c5806d0.

dawnofmidnight (approved):
:shipit:

ToxicKidz (approved):
:shipit:

bot/exts/help_channels/_caches.py
```diff
@@ -20,3 +20,7 @@
 # This cache maps a help channel to original question message in same channel.
 # RedisCache[discord.TextChannel.id, discord.Message.id]
 question_messages = RedisCache(namespace="HelpChannels.question_messages")
+
+# This cache keeps track of the dynamic message ID for
+# the continuously updated message in the #How-to-get-help channel.
+dynamic_message = RedisCache(namespace="HelpChannels.dynamic_message")
```
MarkKoz, line 26: I think redis is good enough, but a more robust approach would be to use message history to find the message when the cog is initialised. It would only need to get the most recent message and make sure the author is the bot (and optionally check the start of the message is "Currently available help channel").

MarkKoz (approved):

bot/exts/help_channels/_caches.py
Xithrius: We can stick with the Redis cache for now since I think we'll probably want to shove more information in the dynamic message in the future.

HassanAbouelela (approved):
Everything looks good, fingers crossed everything goes as expected this time.

PR #1428, opened 2021-02-22
pep 8 and inline-codeblock tag updates

wookie184:
Rewrote the inline tag as it was unnecessarily long, and added the pep 8 song to the pep 8 tag.
Inline tag now looks like this:
image

Pep 8 tag now looks like this:
image


Senjan21 (approved):
Those changes look good to me, thanks!

Xithrius (approved):
Good stuff, I like the examples!

Tested locally, it works well.

PR #1429, opened 2021-02-23
DM on open help channel

### LINKED ISSUE #1174: DM when opening help channel
## Use case/rationale
Sometimes, users don't remember which help channel is theirs. Other times, they simply wanna look back at the old help channel threads and review what was said.

## Proposed solution
DM users with an embed when opening a new help channel. The following info would be included:

1. Help channel name
2. Preview of first message sent
3. "Jump to message" link

(1) Would allow them to quickly look up the channel name or remind themselves what channel was claimed, without necessarily going there.

(2) Let's them quickly remember what issue was discussed in the help channel, after the help session is over

(3) Allows the user to easily navigate to the beginning of the thread, both during a help session and when it's done.

gustavwilliam:
Closes #1174.

This PR introduces a feature that will DM users when they claim help channels, with links to the original message and more. It looks like this:
Screenshot 2021-02-23 at 04 14 17

This also introduces a new message truncation feature, that we can use whenever the content of a message needs to be truncated to a certain length. The bot simply ignores to send a DM if the claimant's DMs aren't open.

## Before merging
The thumbnail icon is currently waiting to get merged into master on the branding repo. Please have a look at python-discord/branding#121 and merge it as soon as possible. That PR needs to be merged in order for the icon to be accessed by the bot.

### DIFF
--- a/bot/exts/help_channels/_message.py
+++ b/bot/exts/help_channels/_message.py
@@ -111,7 +111,7 @@ async def dm_on_open(message: discord.Message) -> None:
         name="Your message", value=truncate_message(message, limit=100), inline=False
     )
     embed.add_field(
-        name="Want to go there?",
+        name="Conversation",
         value=f"[Jump to message!]({message.jump_url})",
         inline=False,
     )
--- a/bot/exts/help_channels/_message.py
+++ b/bot/exts/help_channels/_message.py
@@ -8,6 +8,7 @@ import bot
 from bot import constants
 from bot.exts.help_channels import _caches
 from bot.utils.channel import is_in_category
+from bot.utils.messages import truncate_message
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


bot/exts/help_channels/_message.py
```diff
@@ -92,6 +93,38 @@ async def is_empty(channel: discord.TextChannel) -> bool:
...
+    embed = discord.Embed(
+        title="Help channel opened",
+        description=f"You claimed {message.channel.mention}.",
+        colour=bot.constants.Colours.bright_green,
+        timestamp=message.created_at,
+    )
+
+    embed.set_thumbnail(url=constants.Icons.green_questionmark)
+    embed.add_field(
+        name="Your message", value=truncate_message(message, limit=100), inline=False
```
codephile1221, line 111:
```suggestion
        name="Your message:", value=truncate_message(message, limit=100), inline=False
```

codephile1221 (commented):
Solid code, I don't have much to say.

bot/utils/messages.py
```diff
@@ -154,3 +154,12 @@ async def send_denial(ctx: Context, reason: str) -> None:
...
     return f"{user.mention} (`{user.id}`)"
+
+
+def truncate_message(message: discord.Message, limit: int) -> str:
+    """Returns a truncated version of the message content, up to the specified limit."""
+    text = message.content
+    if len(text) > limit:
+        return text[:limit-3] + "..."
+    else:
+        return text
```
MarkKoz, line 165: This seems redundant considering `textwrap.shorten` exists.

bot/exts/help_channels/_message.py
```diff
@@ -92,6 +93,38 @@ async def is_empty(channel: discord.TextChannel) -> bool:
...
+    )
+    embed.add_field(
+        name="Want to go there?",
```
jb3, line 114: Let's formalise this a tad.
```suggestion
        name="Jump to conversation",
```

bot/exts/help_channels/_message.py
jb3: A field makes it render pretty fine without one.

jb3 (changes requested):
Nice work! One minor language thing and then just Mark's comment need resolving and I think we're good.

bot/exts/help_channels/_message.py
gustavwilliam: So, the reason I chose this, is that both the field name and description will be "Jump to X", making it slightly repetitive. I tried something similar to what Sir Lancebot is using for `bookmark`.

Do you have any other suggestions of ways to formulate this, which are a bit more formal?

jb3: Maybe just "Conversation" as the header and link being "Jump to first message"?

bot/utils/messages.py
gustavwilliam: Great idea, thanks for telling me about it!

bot/exts/help_channels/_message.py
gustavwilliam: That could definitely work. I'll update it.

jb3 (approved):
:shipit:

Xithrius (approved):
Tested locally. It works well!

Nice grammar you got there.

PR #1434, opened 2021-02-25
Send command help for BadUnionArgument errors

Numerlor:
Currently the bot only sends the error embed when a BadUnionArgument or ArgumentParsingError occurs, unlike all the other handled exceptions which send the help for the command. 
The PR adds sending the help under the BadUnionArgument error. 
For ArgumentParsingError I've decided to only close the coroutine, as the error fails on the user's formatting and occurs regardless of the user passing in the expected args or not


Senjan21 (approved):
Looks and sounds good! Thanks.

Shivansh-007 (approved):
Thanks! Looks Good To Me!

PR #1435, opened 2021-02-25
Instructions to dispute an infraction vary by infraction type.

### LINKED ISSUE #1363: Upon receiving a non-ban infraction, the DM from the bot should tell them about ModMail
Currently, users get a message like this one when they get an infraction:

image

For infractions other than bans (regardless of duration), the embed should tell them to use the ModMail bot if they want to discuss or ask questions about the infraction. We might also except Superstarify infractions.

I'm not sure if this should await the redesign of the infraction system or if we can implement it without sabotaging Scragly's efforts.

swfarnsworth:
Previously, the user was instructed to email the appeals email for infraction types that don't remove one from the server. They are now instructed to DM ModMail, except for infractions that remove them from the server (which all get mapped to the string `Ban` by the time the relevant code is involved). 

There was also a header that read "Please review our rules over at <actual URL>", and this now reads "Please review our rules". The header itself was and still is a link to the URL in question. I thought that stating the URL literally was visually unappealing.

Closes #1363 

image

On second consideration, I think the message for ban-like infractions should begin "If you want to appeal" so we don't suggest that appealing the infraction is the default course of action. If anyone feels similarly, they or I can probably make that change through the GitHub UI.


bot/exts/moderation/infraction/_utils.py
```diff
@@ -178,17 +179,18 @@ async def notify_infraction(
     embed.title = INFRACTION_TITLE
     embed.url = RULES_URL
 
-    if infr_type in APPEALABLE_INFRACTIONS:
-        embed.set_footer(text=INFRACTION_APPEAL_FOOTER)
+    embed.set_footer(
+        text=INFRACTION_APPEAL_EMAIL_FOOTER if infr_type == 'Ban' else INFRACTION_APPEAL_MODMAIL_FOOTER
```
Akarys42, line 183: I'd rather a constant set of email appealable infractions, but it is fine like this

bot/exts/moderation/infraction/_utils.py
```diff
@@ -31,8 +30,10 @@
 
 APPEAL_EMAIL = "appeals@pythondiscord.com"
 
-INFRACTION_TITLE = f"Please review our rules over at {RULES_URL}"
-INFRACTION_APPEAL_FOOTER = f"To appeal this infraction, send an e-mail to {APPEAL_EMAIL}"
+INFRACTION_TITLE = "Please review our rules"
+INFRACTION_APPEAL_EMAIL_FOOTER = f"To appeal this infraction, send an e-mail to {APPEAL_EMAIL}"
+INFRACTION_APPEAL_MODMAIL_FOOTER = ('If you would like to discuss or appeal this infraction, '
+                                    'send a message to the ModMail bot')
```
Akarys42, line 36: We use this style of multiline strings throughout the code base.
```suggestion
INFRACTION_APPEAL_MODMAIL_FOOTER = (
    'If you would like to discuss or appeal this infraction, '
    'send a message to the ModMail bot'
)
```

Akarys42, line 36: Could we make it a mention of the modmail bot? Changing `ModMail` to `<@683001325440860340>` should do it.

Akarys42 (changes requested):
It looks pretty good, I left a few little comments. Also, it seems like it broke the modlog tests.

bot/exts/moderation/infraction/_utils.py
swfarnsworth: For all infractions that remove the individual from the server, the `infr_type` argument that gets passed to this function is `'Ban'` (upper-case B), as opposed to `'ban'` or `'kick'`, which are the names for those infractions used elsewhere in the code base. I therefore decided on a string literal in the function so that we don't give the false impression that `'Ban'` is used elsewhere.

bot/exts/moderation/infraction/_utils.py
swfarnsworth: I tried this and it didn't render as expected, presumably because the DM is out-of-guild content.

swfarnsworth: That's fine. Perhaps we could add that rule to the pre-commit checks?

swfarnsworth:
@Akarys42 I'll commit the change you requested regarding the formatting of multi-line strings along with any other requested changes.

bot/exts/moderation/infraction/_utils.py
Akarys42: Alright, it is fine like this then

bot/exts/moderation/infraction/_utils.py
Akarys42: Hmm yeah it can cause issues if it isn't in the cache. Could we maybe have both, like having a mention in parenthesis along with the bot name?

Shivansh-007 (approved):
Looks good, other than the changes mentioned by Akarys I don’t think anything is needed.

And I don’t think the indentation changes in the function are required.

kosayoda (approved):
Nice job!

bot/exts/moderation/infraction/_utils.py
kosayoda: I don't think that would improve the footer, as it's already long as it is. @Akarys42 thoughts?

Akarys42: I’m fine with this current text :+1:

Akarys42 (approved):
Looks good, thanks a lot!

bot/exts/moderation/infraction/_utils.py
swfarnsworth: I don't think it's worth risking the footer being uninformative (I think it would just say `invalid-user`) if the bot, for some reason, weren't in their cache.

dawnofmidnight: Personally, it seems far too wide. At the very least, add a `\n` character.

Issue #1385, opened 2021-01-26
Redis Startup Error


HassanAbouelela:
## Description
Make fake redis an environment variable (like it is in sir-lancebot), and possibly have the default fake redis value be true.

## Reasoning
A lot of developers try to start the project, just to be met with a long traceback ending with:
> socket.gaierror: [Errno 11004] getaddrinfo failed

It is unclear that the cause here is a lack of a redis setup, and the main indicator that it is, is to look further up in the traceback to find the call to start redis, then to put two and two together.

A lot of debug time can be saved if we have the default value be true, as there aren't major differences for a developer as far as I know. Additionally, making it an env var (call it `USE_FAKEREDIS` to maintain consistency) makes it easier to debug and deploy the app.

Make sure to add the env var to the [contributing guide](https://pythondiscord.com/pages/contributing/bot/#environment-variables), or get someone to do it for you. You should be able to copy the name and description from the [sir-lancebot guide](https://pythondiscord.com/pages/contributing/sir-lancebot/sir-lancebot-env-var-reference/#debug).

## Possible Alternatives
My main concern here is that it is hard for new comers to the project to figure out the error. One possible alternative is to wrap the redis connection code in a try/except, and to log a more meaningful message to help the user.

This should probably be accompanied by a change in the documentation on the contributing guide, or some other resource to help fix the issue.

mbaruh:
I think it's fine, but maybe we should make it fall back to fakeredis if there's no redis? and log a warning if redis was desired.

Regardless of the approach, making the error more meaningful is desirable.

ChrisLovering:
Fixed in #1436

PR #1436, opened 2021-02-25
Helpful Startup Errors

### LINKED ISSUE #1349: Waiting for the website to be up before starting in development mode
# Abstract

When the development mode is enabled, the bot should wait for the website to be up before continuing the startup process.

# Rationale

With the current setup using Docker Compose, the bot will start making requests to the website before the website even starts accepting requests, which will cause the bot to fail to start. 

# Specification

We can have a loop trying to ping the server URL every second for a short duration before moving on. Note that the bot should fail to start after some time.

HassanAbouelela:
# Description
One of the biggest hurdles that people face when setting up the bot, is the lack of helpful information when things go wrong. From redis, to the API, plenty can go wrong, and it requires experience to read the traceback and figure out why.

Closes #1349.

# Inspiration
This PR aims to take a step in fixing that, by offering more descriptive errors on startup. This is inspired by #1385, but only covers a portion of the proposed changes.

# Implementation
This PR does two things:
1. It logs a message for redis and API errors, so we don't keep getting confused people asking why they got a 100 line traceback when they hit run.
2. It adds retries to site connection, so it solves the problem of the bot trying to connect to the site before it warms up in docker. (Right now that is at 3 attempts with 5 seconds between each).

### DIFF
--- a/bot/__main__.py
+++ b/bot/__main__.py
@@ -15,7 +15,7 @@ try:
     bot.instance.run(constants.Bot.token)
 except StartupError as e:
     message = "Unknown Startup Error Occurred."
-    if isinstance(e.exception, aiohttp.ClientConnectorError):
+    if type(e.exception) in [aiohttp.ClientConnectorError, aiohttp.ServerDisconnectedError]:
         message = "Could not connect to site API. Is it running?"
     elif isinstance(e.exception, OSError):
         message = "Could not connect to Redis. Is it running?"
--- a/bot/__main__.py
+++ b/bot/__main__.py
@@ -15,7 +15,7 @@ try:
     bot.instance.run(constants.Bot.token)
 except StartupError as e:
     message = "Unknown Startup Error Occurred."
-    if type(e.exception) in [aiohttp.ClientConnectorError, aiohttp.ServerDisconnectedError]:
+    if isinstance(e.exception, (aiohttp.ClientConnectorError, aiohttp.ServerDisconnectedError)):
         message = "Could not connect to site API. Is it running?"
     elif isinstance(e.exception, OSError):
         message = "Could not connect to Redis. Is it running?"
@@ -24,3 +24,5 @@ except StartupError as e:
     log = logging.getLogger("bot")
     log.fatal("", exc_info=e.exception)
     log.fatal(message)
+
+    exit(69)
--- a/bot/__main__.py
+++ b/bot/__main__.py
@@ -1,10 +1,26 @@
+import logging
+
+import aiohttp
+
 import bot
 from bot import constants
-from bot.bot import Bot
+from bot.bot import Bot, StartupError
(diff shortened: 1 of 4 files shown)


bot/bot.py
```diff
@@ -19,6 +19,14 @@
 LOCALHOST = "127.0.0.1"
 
 
+class StartupError(Exception):
+    """Exception class for startup errors."""
+
+    def __init__(self, base: Exception):
+        super()
```
Akarys42, line 26: I believe you meant
```suggestion
        super().__init__()
```

bot/bot.py
```diff
@@ -81,6 +89,22 @@ async def cache_filter_list_data(self) -> None:
...
+        while True:
+            try:
+                log.info(f"Attempting site connection: {attempts + 1}/{constants.URLs.connect_max_retries}")
+                await self.api_client.get("healthcheck")
+                break
+
+            except (aiohttp.ClientConnectorError, aiohttp.ServerDisconnectedError) as e:
+                attempts += 1
+                if attempts == constants.URLs.connect_max_retries:
+                    raise e
```
Akarys42, line 105: Nitpicky, but it can just be
```suggestion
                    raise
```

bot/__main__.py
```diff
@@ -1,10 +1,26 @@
...
+    message = "Unknown Startup Error Occurred."
+    if type(e.exception) in [aiohttp.ClientConnectorError, aiohttp.ServerDisconnectedError]:
+        message = "Could not connect to site API. Is it running?"
+    elif isinstance(e.exception, OSError):
+        message = "Could not connect to Redis. Is it running?"
+
+    # The exception is logged with an empty message so the actual message is visible at the bottom
+    log = logging.getLogger("bot")
+    log.fatal("", exc_info=e.exception)
+    log.fatal(message)
```
Akarys42, line 26: This should exit with a non-zero code

bot/__main__.py
```diff
@@ -1,10 +1,26 @@
...
-bot.instance = Bot.create()
-bot.instance.load_extensions()
-bot.instance.run(constants.Bot.token)
+try:
+    bot.instance = Bot.create()
+    bot.instance.load_extensions()
+    bot.instance.run(constants.Bot.token)
+except StartupError as e:
+    message = "Unknown Startup Error Occurred."
+    if type(e.exception) in [aiohttp.ClientConnectorError, aiohttp.ServerDisconnectedError]:
```
Akarys42, line 18: This should be an `isinstance` check ~~and it should be a tuple, not a list :P~~

Akarys42 (changes requested):
This seems to be working fine, although I have some comments to make

Akarys42 (approved):
LGTM, thanks!

Shivansh-007 (approved):
Thanks! Looks Good To Me!

PR #1442, opened 2021-03-01
Bump license year to 2021

gustavwilliam:
It seems like we were enjoying a good old 2018 license. Time to update!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2018 Python Discord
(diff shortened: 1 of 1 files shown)


jb3:
Since we have implicit copyright for 95 years it isn't necessary to update this. If we make major alterations to the project we can add a year like ":copyright: 2018, 2021" but at this stage I don't think that is necessary.

For certain, we should not rebuke our 2018 license, as that is the time of publication.

PR #1446, opened 2021-03-04
Enhancing information within the incidents channel.

### LINKED ISSUE #1226: Enhancing information within the incidents channel.
Given that message links jump to deleted messages sometimes, it's hard to figure out who was the perputrator.

Within the incidents channel, it may be a good idea for the bot to provide a user ID for the message link owner so mods+ can look into their message history. 

If the message is still there, the bot could tell what it contains. If exceeding a certien limit, the bot will give a shortened version of the content.

If the message link returns nothing, then maybe (if it's possible) the bot can send a message to redirect to a deleted message in the message-change-log channel.

Shivansh-007:
## Relevant Issues
Closes #1226 

## Description
Whenever a staff member would post a message in `#incidents` channel, it would be searched through a regex and all the message links in it would be extracted.

Then the bot would send this the messages the link point towards in embeds.
```md
The Embed would contain the following information -->
        Author: @Jason Terror ♦ (736234578745884682)
        Channel: Special/#bot-commands (814190307980607493)
        Content: This is a very important message!
```
These embeds would be sent in groups of 10 using a webhook.

Once the incident is actioned/rejected (using reactions), the bot would delete these webhook messages (**NOT** send them to `#incidents-archive`)

### Has a link and doesn't has a link message
Screenshot from 2021-03-04 13-57-32

### On putting ✅ reaction on message with link
Screenshot from 2021-03-04 13-57-44

### In Incidents-Archive Channel (the message link embeds are not sent)
Screenshot from 2021-03-04 13-57-50

## Additional Details
Adds a webhook to channel #incidents with id `816650601844572212` (Thanks joe)

## Did you:

- [x] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x] **Lint your code** (`pipenv run lint`)?
- [x] Set the PR to **allow edits from contributors**?


bot/exts/moderation/incidents.py
```diff
@@ -130,6 +137,32 @@ def has_signals(message: discord.Message) -> bool:
...
+    The Embed would contain the following information -->
+        Author: @Jason Terror ♦ (736234578745884682)
+        Channel: Special/#bot-commands (814190307980607493)
+        Content: This is a very important message!
+    """
+    channel_id = int(message_link.split("/")[3])
+    msg_id = int(message_link.split("/")[4])
+
+    channel = incident.guild.get_channel(channel_id)
+    message = await channel.fetch_message(msg_id)
```
ks129, line 153: Don't parse manually, use `MessageConverter` from discord.py instead.

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +137,32 @@ def has_signals(message: discord.Message) -> bool:
...
+
+    text = message.content
+    channel = message.channel
```
ks129, line 156: Message can also be `None` if not found.

ks129 (changes requested):
Quick first review

bot/exts/moderation/incidents.py
```diff
@@ -113,9 +121,9 @@ def is_incident(message: discord.Message) -> bool:
     """True if `message` qualifies as an incident, False otherwise."""
     conditions = (
         message.channel.id == Channels.incidents,  # Message sent in #incidents
-        not message.author.bot,                    # Not by a bot
-        not message.content.startswith("#"),       # Doesn't start with a hash
-        not message.pinned,                        # And isn't header
+        not message.author.bot,  # Not by a bot
+        not message.content.startswith("#"),  # Doesn't start with a hash
+        not message.pinned,  # And isn't header
```
mbaruh, line 126: I'm not sure why you removed the spaces here, they were added for readability

bot/exts/moderation/incidents.py
```diff
@@ -404,9 +460,41 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+                else:
+                    await self.message_link_embeds_cache.set(message.id, ','.join(map(str, webhook_msg_ids)))
+                    log.trace("Message Link Embeds Sent successfully!")
+
+            log.trace(f"Skipping discord message link detection on {message.id}: message doesn't qualify.")
             await add_signals(message)
 
 
-def setup(bot: Bot) -> None:
+def setup(bot: bot.Bot) -> None:
```
mbaruh, line 498: Similarly here, your change isn't in line with the rest of the codebase

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,39 @@ def has_signals(message: discord.Message) -> bool:
...
+        Channel: Special/#bot-commands (814190307980607493)
+        Content: This is a very important message!
+    """
+    embed = discord.Embed()
+
+    try:
+        message_convert_object = MessageConverter()
+        message = await message_convert_object.convert(ctx, message_link)
+
+    except Exception as e:
```
mbaruh, line 156: This shouldn't be a bare exception

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,39 @@ def has_signals(message: discord.Message) -> bool:
...
+        embed.title = f"{e}"
+        embed.colour = Colours.soft_red
```
mbaruh, line 158: Is the error going to be posted to the channel? I'm not sure that's desired. I'd say it should log the error and return without creating an embed.

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,39 @@ def has_signals(message: discord.Message) -> bool:
...
+
+    else:
+        text = message.content
+        channel = message.channel
+
+        embed.description = (
+            f"**Author:** {format_user(message.author)}\n"
+            f"**Channel:** {channel.category}/#{channel.name} (`{channel.id}`)\n"
+            f"**Content:** {text[:2045] + '...' if len(text) > 2048 else text}\n"
```
mbaruh, line 167: This can become quite bulky. Let's limit it to ~500-1000 characters.

bot/exts/moderation/incidents.py
```diff
@@ -178,7 +219,11 @@ class Incidents(Cog):
     Please refer to function docstrings for implementation details.
     """
 
-    def __init__(self, bot: Bot) -> None:
+    # This dictionary maps a incident message to the message link embeds(s) sent by it
+    # RedisCache[discord.Message.id, List[discord.Message.id]]
+    message_link_embeds_cache = RedisCache()
+
+    def __init__(self, bot: bot.Bot) -> None:
```
mbaruh, line 226: As earlier, there is no reason for this line change

bot/exts/moderation/incidents.py
```diff
@@ -404,9 +460,41 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+
+                    for x, embed in enumerate(webhook_embed_list):
+                        webhook_msg = await webhook.send(
+                            embeds=[x for x in embed if x is not None],
+                            username=sub_clyde(message.author.name),
+                            avatar_url=message.author.avatar_url,
+                            wait=True
+                        )
+                        webhook_msg_ids.append(webhook_msg.id)
+                        log.trace(f"Message Link Embed {x + 1}/{len(webhook_embed_list)} Sent Succesfully")
```
mbaruh, line 485: The try clause should be minimized to the part you think will throw an exception, e.g without the part before the loop.

I also suggest you export the loop to a utility function.

mbaruh (changes requested):
First review. Looks pretty good overall.

bot/exts/moderation/incidents.py
Shivansh-007: > Similarly here, your change isn't in line with the rest of the codebase

I needed import of `Bot` from `discord.ext.commands` earlier, but later I didn't need it and I forgot to remove it from here.

bot/exts/moderation/incidents.py
Shivansh-007: > Is the error going to be posted to the channel? I'm not sure that's desired. I'd say it should log the error and return without creating an embed.

It would be posted along the other embeds with the webhook. I did this so it is easier to know what happened about that message link. I could just log it with a `log.info` and not create a embed for it

bot/exts/moderation/incidents.py
Shivansh-007: In many places of the code we use a bare exception, and for the similar case I using this, If any error occurs it should be mentioned in the embed, whether it is Channel Not Found, Channel Unreadable or Message Not Found. This is then set as the header of the error embed for that particular message link.

mbaruh:
You should keep commit titles short. If you want to add more detail you should do it in the commit body.

bot/exts/moderation/incidents.py
mbaruh: Maybe `DiscordException`?

Shivansh-007: Yeah, `DiscordException` would be better.

bot/exts/moderation/incidents.py
mbaruh: Let's just leave it for now. We can revisit it if we decide it's not helpful.

bot/exts/moderation/incidents.py
```diff
@@ -394,18 +484,76 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+                    embeds.append(await make_message_link_embed(ctx, message_link[0]))
+
+                webhook = await self.bot.fetch_webhook(Webhooks.incidents)
+                webhook_embed_list = list(grouper(embeds, 10))
+
+                await self.send_webhooks(webhook_embed_list, message, webhook)
+
+            log.trace(
+                f"Skipping discord message link detection on {message.id}: message doesn't qualify."
+            )
```
mbaruh, line 512: I'm wondering if this part should be exported to a helper function, and called again on message edits.

mbaruh: Meaning, the embeds of the message should be deleted and resent

mbaruh: Or just edit the embeds message? unsure

bot/exts/moderation/incidents.py
```diff
@@ -178,6 +231,10 @@ class Incidents(Cog):
     Please refer to function docstrings for implementation details.
     """
 
+    # This dictionary maps a incident message to the message link embeds(s) sent by it
+    # RedisCache[discord.Message.id, List[discord.Message.id]]
+    message_link_embeds_cache = RedisCache()
```
kwzrd, line 236: The type hint is incorrect, the value is just a string, delimited with commas. It'd probably be better explained using natural language; the hints don't really work.

bot/exts/moderation/incidents.py
```diff
@@ -394,18 +484,76 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+
+                webhook = await self.bot.fetch_webhook(Webhooks.incidents)
+                webhook_embed_list = list(grouper(embeds, 10))
+
+                await self.send_webhooks(webhook_embed_list, message, webhook)
+
+            log.trace(
+                f"Skipping discord message link detection on {message.id}: message doesn't qualify."
+            )
             await add_signals(message)
```
kwzrd, line 513: This logic should be moved to a function like `add_signals` just below is.

kwzrd (commented):
This PR adds a large amount of code style changes that I cannot really wrap my head around. I see you decided to run Black on the code, do you think the lines that got broken were too long? New logging is inconsistent in style with both itself and the existing code; some existing logging was changed but not all.

None of the documentation, class or function, was updated to mention the new behaviour.

I'm still not convinced that we need to support more than 10 embeds per report. Such an amount of embeds would flood the channel, and it adds a non-trivial amount of complexity to the solution. But this has to be decided by staff.

kwzrd:
Additionally, I'm not sure whether sending exceptions to `#incidents` makes sense. Wouldn't it be better to log the exception and let it be handled by Sentry?

bot/exts/moderation/incidents.py
Shivansh-007: Well we have two cases here:

1. If we support more than 10 links, then say we get 10 links in the message, and the author adds another, it would be 11 links, message edit is not possible in this case. In this case I would go with message delete and send new webhook message.

2. If we don't support more than 10 links, then we can just edit the embeds message.

Shivansh-007:
> do you think the lines that got broken were too long? 

Not exactly sure, if they were too long or if `black` decided to break them.

> New logging is inconsistent in style with both itself and the existing code; some existing logging was changed but not all.

Could you point them out?

> None of the documentation, class or function, was updated to mention the new behaviour.

I completely forgot about them, will have them updated.

> I'm still not convinced that we need to support more than 10 embeds per report. Such an amount of embeds would flood the channel, and it adds a non-trivial amount of complexity to the solution. But this has to be decided by staff.

I completely agree with you here, I have asked the moderators in dev-contrib (Akarys has passed on the message).

bot/exts/moderation/incidents.py
```diff
@@ -394,18 +484,76 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
         if is_incident(message):
+            message_links = DISCORD_MESSAGE_LINK_RE.findall(str(message.content))
+
+            if message_links:
+                embeds = []
+                for message_link in message_links:
+                    ctx = await self.bot.get_context(message)
+                    embeds.append(await make_message_link_embed(ctx, message_link[0]))
+
+                webhook = await self.bot.fetch_webhook(Webhooks.incidents)
```
Akarys42, line 505: The webhook should be a cog level attribute to not fetch it every time.

bot/exts/moderation/incidents.py
```diff
@@ -323,6 +398,19 @@ async def process_event(self, reaction: str, incident: discord.Message, member:
...
+        log.trace("Deleting discord links webhook message.")
+        webhook_msg_ids = await self.message_link_embeds_cache.get(incident.id)
+
+        if webhook_msg_ids:
+            webhook_msg_ids = webhook_msg_ids.split(",")
+            webhook = await self.bot.fetch_webhook(Webhooks.incidents)
+
+            for x, msg in enumerate(webhook_msg_ids):
+                await webhook.delete_message(msg)
+                log.trace(f"Deleted discord links webhook message{x}/{len(webhook_msg_ids)}")
```
Akarys42, line 410: I believe there is a space missing here
```suggestion
                log.trace(f"Deleted discord links webhook message {x}/{len(webhook_msg_ids)}")
```
Besides I'd rather log the message ID than the number.

Akarys42 (changes requested):
Thanks for the PR! I have a few comments to do:
- Please revert lines broken down by Black. You have free to use formatters on your contributions but they must comply with our code style. We don't use 79 chars limits.
- We decided to not allow more than 10 embeds per report. This can reduce code complexity.
- Entries from the cache should be deleted after the incident is cleared. We shouldn't keep the cache growing exponentially.
- Message edits aren't handled properly, they should. Also I haven't tested it but is it handling message deletion too?
- The docstring hasn't been updated to match the new features
- I would also like to see tests for this new feature
Note that I haven't done a functional review yet.

bot/exts/moderation/incidents.py
mbaruh: Since it seems like we're not supporting more than 10, let's make it work for edits.

bot/exts/moderation/incidents.py
```diff
@@ -139,7 +190,9 @@ async def add_signals(incident: discord.Message) -> None:
     existing_reacts = own_reactions(incident)
 
     for signal_emoji in Signal:
-        if signal_emoji.value in existing_reacts:  # This would not raise, but it is a superfluous API call
+        if (
+            signal_emoji.value in existing_reacts
+        ):  # This would not raise, but it is a superfluous API call
```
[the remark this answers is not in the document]
Shivansh-007: Yup, I would revert the `black` formatting changes.

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +148,39 @@ def has_signals(message: discord.Message) -> bool:
...
+
+    else:
+        text = message.content
+        channel = message.channel
+
+        embed.description = (
+            f"**Author:** {format_user(message.author)}\n"
+            f"**Channel:** {channel.category}/#{channel.name} (`{channel.id}`)\n"
+            f"**Content:** {textwrap.shorten(text, 300, placeholder='...')}\n"
+            "\n"
```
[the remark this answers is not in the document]
Shivansh-007: Yes they for styling, if the newlines aren't added after each, then it would be kept to one line, which would look bad.

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +148,39 @@ def has_signals(message: discord.Message) -> bool:
...
+        Content: This is a very important message!
+    """
+    embed = discord.Embed()
+
+    try:
+        message_convert_object = MessageConverter()
+        message = await message_convert_object.convert(ctx, message_link)
+
+    except discord.DiscordException as e:
+        embed.title = f"{e}"
```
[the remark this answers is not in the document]
Shivansh-007: The tests were complaining that the title is not a string, so I added it here.

bot/exts/moderation/incidents.py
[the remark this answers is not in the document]
ChrisLovering: I'm specifically talking about the newline on this line, since there's already a new link at the end of the previous one.

bot/exts/moderation/incidents.py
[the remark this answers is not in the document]
ChrisLovering: Ahh right, maybe just a `str()` call would suffice then? Just to show intent.

Shivansh-007:
The following changes are mentioned in commit 78333e7 :

> - Entries from the cache should be deleted after the incident is cleared. We shouldn't keep the cache growing exponentially.
> - Message edits aren't handled properly, they should. Also I haven't tested it but is it handling message deletion too?
> - The docstring hasn't been updated to match the new features

bot/exts/moderation/incidents.py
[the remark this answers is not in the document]
Shivansh-007: Mentioned in commit 7aadd79

bot/exts/moderation/incidents.py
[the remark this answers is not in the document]
Shivansh-007: Mentioned in commit 2ece953.

bot/exts/moderation/incidents.py
Shivansh-007: Resolved in commit 92f37ec and 0f8ecf3.

bot/exts/moderation/incidents.py
Akarys42: You marked the conversation as resolved, have you reverted this?

Shivansh-007: > You marked the conversation as resolved, have you reverted this?

It was resolved and the change had been done, but sometimes PyCharm pops up a message of `Reformat code`, and on clicking that, this change happens.

I must have missed this piece of code while going through the code before committing.

Shivansh-007: Resolved in commit 93f9243.

bot/exts/moderation/incidents.py
Shivansh-007: Resolved in commit 93f9243.

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +139,39 @@ def has_signals(message: discord.Message) -> bool:
...
+    """
+    embed = discord.Embed()
+
+    try:
+        message_convert_object = MessageConverter()
+        message = await message_convert_object.convert(ctx, message_link)
+
+    except discord.DiscordException as e:
+        embed.title = str(e)
+        embed.colour = Colours.soft_red
```
Akarys42, line 159: Errors shouldn't be sent in `#incidents`. Instead, they should be logged with `log.exception` and the function should return.

bot/exts/moderation/incidents.py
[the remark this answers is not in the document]
Akarys42: This hasn't been addressed

bot/exts/moderation/incidents.py
Akarys42: This should still be reverted

bot/exts/moderation/incidents.py
```diff
@@ -173,18 +243,30 @@ class Incidents(Cog):
...
 
+        self.bot.loop.create_task(self.get_webhook())
+
         self.event_lock = asyncio.Lock()
         self.crawl_task = self.bot.loop.create_task(self.crawl_incidents())
 
+    async def get_webhook(self) -> None:
+        """Fetch and store message link embeds webhook, present in #incidents channel."""
+        await self.bot.wait_until_guild_available()
+        self.incidents_webhook = await self.bot.fetch_webhook(Webhooks.incidents)
```
Akarys42, line 268: If the webhook isn't available and returns `None`  this should be properly logged instead of letting some NPEs appear later on.

Akarys42 (changes requested):
I'd also still like to see tests for the new feature.

bot/exts/moderation/incidents.py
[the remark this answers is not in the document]
Shivansh-007: > This hasn't been addressed

Yes, I know that, I still need to have a look what happens when you remove the last `\n`, hence I didn't mark it as resolved.

bot/exts/moderation/incidents.py
Shivansh-007: Scaleios suggested doing something like this: https://discord.com/channels/267624335836053506/635950537262759947/833165964014256159.

Is there any better solution or should I go ahead and try doing this?

bot/exts/moderation/incidents.py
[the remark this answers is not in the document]
Shivansh-007: Hmm, I can find any difference, and for a proof, I even ran a `internal eval` code to see the difference.

Screenshot from 2021-04-19 05-51-42

bot/exts/moderation/incidents.py
Shivansh-007: Ok, I understood it wrong, will do the required changes.

mbaruh:
Something that bothers me is that it isn't clear which embed relates to which message link, which is exacerbated if any of the messages error out and get skipped. Add the message link in the title (but then the embed is even bulkier)? in the footer? just the message ID?

Shivansh-007:
> Something that bothers me is that it isn't clear which embed relates to which message link, which is exacerbated if any of the messages error out and get skipped. Add the message link in the title (but then the embed is even bulkier)? in the footer? just the message ID?

Yeah, I would add the message ID in the footer, I had tried adding the message link as the Title and Footer but then it breaks (half link in line, second half in the next line) due to embed lengths and the embed is bulkier.

mbaruh:
I also think that the channel ID should be a channel mention, and the user should also be mentioned.

mbaruh:
And a minor nitpick, but can you try a gold color for the embed? :D

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +139,39 @@ def has_signals(message: discord.Message) -> bool:
...
+    else:
+        text = message.content
+        channel = message.channel
+
+        embed = discord.Embed(
+            colour=discord.Colour.gold(),
+            description=(
+                f"**Author:** {format_user(message.author)}\n"
+                f"**Channel:** <#{channel.id}> ({channel.category}/#{channel.name})\n"
+                f"**Content:** {textwrap.shorten(text, 300, placeholder='...')}\n"
```
mbaruh, line 167: If someone spams without spaces, this will just show "...".

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +492,104 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        the embeds in the `webhook_embed_list` parameter.
+
+        After sending each webhook it maps the `message.id`
+        to the `webhook_msg_ids` IDs in the async redis-cache.
+        """
+        try:
+            webhook_msg = await webhook.send(
+                embeds=[x for x in webhook_embed_list if x is not None],
+                username=sub_clyde(message.author.name),
+                avatar_url=message.author.avatar_url,
```
mbaruh, line 567: This is an interesting idea, but I'm not sure if it fits here. Maybe just the Python logo and the name "Incident Context"? or maybe some info symbol for the avatar.

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +492,104 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+            log.trace("Message Link Embed Sent successfully!")
+            return webhook_msg.id
+
+    async def delete_msg_link_embed(self, message: discord.Message) -> None:
+        """Delete discord message link message found in cache for `message`."""
+        log.trace("Deleting discord links webhook message.")
+        webhook_msg_id = await self.message_link_embeds_cache.get(message.id)
+
+        if webhook_msg_id:
+            await self.incidents_webhook.delete_message(webhook_msg_id)
```
mbaruh, line 588: This raises `discord.errors.NotFound` whenever a webhook message is deleted, not sure why. Also this doesn't account for a case where the webhook message was already deleted for some reason.

mbaruh (changes requested):
Seems to work well, mostly minor comments:

bot/exts/moderation/incidents.py
mbaruh: You know what scratch that, it can help identify who sent what.

bot/exts/moderation/incidents.py
mbaruh: This doesn't raise an error anymore, but it still fires the "skipping" log message whenever I action an incident, even if the embed is still there.

bot/exts/moderation/incidents.py
Shivansh-007: Fixed in commit d6872422.

mbaruh:
It's not a bug with `textwrap.shorten`. `shorten` will truncate the "word" which causes the string to pass the set length. If the only word makes it pass the length, that word will be truncated, and an empty string will be left. So that works as expected.

If there's such a long "word", I feel like showing 300 characters of it is unnecessary. In that specific case ~50 should be enough. I also think your usage of `and` is a bit obscure here.

bot/exts/moderation/incidents.py
Shivansh-007: I couldn't replicate this on my test bot, could you post a screenshot of it?

Shivansh-007:
> It's not a bug with `textwrap.shorten`. `shorten` will truncate the "word" which causes the string to pass the set length. If the only word makes it pass the length, that word will be truncated, and an empty string will be left. So that works as expected.
> 
> If there's such a long "word", I feel like showing 300 characters of it is unnecessary. In that specific case ~50 should be enough. I also think your usage of `and` is a bit obscure here.
> 
> At the same time, there should probably be a line limit (let's say three).

So, if the sentence is three words (say) in the 300 shortened characters, then we reduce it to 50 characters. Otherwise, keep it to 300 characters (normal length).

And on this shortened text, we check and make sure it is not more than 3 lines. (Counting the `\n`).

mbaruh:
> So, if the sentence is three words (say) in the 300 shortened characters, then we reduce it to 50 characters. Otherwise, keep it to 300 characters (normal length).

I think just making sure it's not a single word is enough. So if `shorten` gives you an empty string, take 50 chars.

> And on this shortened text, we check and make sure it is not more than 3 lines. (Counting the `\n`).
yes

tests/bot/exts/moderation/test_incidents.py
```diff
@@ -768,3 +788,64 @@ async def test_on_message_non_incident(self):
...
+            id=777,
+            content=f"I would like to report the following messages, "
+                    f"as they break our rules: \n{', '.join(msg_links)}"
+        )
+
+        embeds = await extract_message_links(incident_msg, self.cog_instance.bot)
+        description = (
+            f"**Author:** {format_user(msg.author)}\n"
+            f"**Channel:** {msg.channel.mention} ({msg.channel.category}/#{msg.channel.name})\n"
+            f"**Content:** {('Hello, World!' * 3000)[:300] + '...'}\n"
```
mbaruh, line 828: I still think you should slice before counting lines, but regardless you should probably add cases for the 50 chars and the 3 lines scenarios.

Shivansh-007: Resolved in commit 488be0c.

tests/bot/exts/moderation/test_incidents.py
```diff
@@ -768,3 +787,73 @@ async def test_on_message_non_incident(self):
...
+        tests = {
+            "thisisasingleword"*10: ('thisisasingleword'*10)[:50]+"...",
+            "\n".join("Lets make a new line test".split()): "Lets\nmake\na"+"...",
+            'Hello, World!' * 300: ('Hello, World!' * 300)[:300] + '...'
+        }
+
+        for test, value in tests.items():
+            self.assertEqual(
+                str(incidents.shorten_text(test)), value
+            )
```
mbaruh, line 806: You should use subtests here. You'll find an example in the test module's README.

bot/exts/moderation/incidents.py
mbaruh: Was this addressed? doesn't seem so.

mbaruh (changes requested):

bot/exts/moderation/incidents.py
Shivansh-007: Yeah I know that hence didn’t resolve it, had a [doubt](https://discord.com/channels/267624335836053506/635950537262759947/840543183674933270).

Shivansh-007: Resolved in 9cfd4082.

tests/bot/exts/moderation/test_incidents.py
Shivansh-007: Resolved in 8e39adae.

bot/exts/moderation/incidents.py
```diff
@@ -146,6 +208,35 @@ async def add_signals(incident: discord.Message) -> None:
...
+    Checks if there's any message links in the text content.
+
+    Then passes the the message_link into `make_message_link_embed` to format a
+    embed for it containing information about the link.
+
+    As discord only allows a max of 10 embeds in a single webhook, just send the
+    first 10 embeds and don't care about the rest.
+
+    If no links are found for the message, it logs a trace statement.
+    """
```
kwzrd, line 222: Why is this not an instance method on the cog if it uses the cog's state?

Consider making it a method and accessing `self.bot` rather than passing it as a param.

The return type annotation can be extended to be `List[Embed]`.

bot/exts/moderation/incidents.py
```diff
@@ -146,6 +208,35 @@ async def add_signals(incident: discord.Message) -> None:
...
+
+    if message_links:
+        embeds = []
+        for message_link in message_links:
+            ctx = await bot.get_context(message)
+            embed = await make_message_link_embed(ctx, message_link[0])
+            if embed:
+                embeds.append(embed)
+
+        return embeds[:10]
```
kwzrd, line 233: Instead of processing all and then returning the first 10, it may make more sense to return after processing 10.

bot/exts/moderation/incidents.py
```diff
@@ -146,6 +208,35 @@ async def add_signals(incident: discord.Message) -> None:
...
+
+    Then passes the the message_link into `make_message_link_embed` to format a
+    embed for it containing information about the link.
+
+    As discord only allows a max of 10 embeds in a single webhook, just send the
+    first 10 embeds and don't care about the rest.
+
+    If no links are found for the message, it logs a trace statement.
+    """
+    message_links = DISCORD_MESSAGE_LINK_RE.findall(str(message.content))
```
kwzrd, line 223: Does this need to convert to string?

bot/exts/moderation/incidents.py
```diff
@@ -173,18 +265,33 @@ class Incidents(Cog):
         * If `Signal.ACTIONED` or `Signal.NOT_ACTIONED` were chosen, attempt to
           relay the incident message to #incidents-archive
         * If relay successful, delete original message
+        * Search the cache for the webhook message for this message, if found delete it.
```
kwzrd, line 268: No other list item ends with a period.

Consider rewording to something like: "Delete quotation message if cached." Webhook message will make no sense to developers reading the docs without knowledge of the internals.

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +515,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        """
+        if is_incident(msg_after):
+            webhook_embed_list = await extract_message_links(msg_after, self.bot)
+            webhook_msg_id = await self.message_link_embeds_cache.get(msg_before.id)
+
+            if webhook_msg_id:
+                await self.incidents_webhook.edit_message(
+                    message_id=webhook_msg_id,
+                    embeds=[x for x in webhook_embed_list if x is not None],
+                )
```
kwzrd, line 556: Please use a better name than `x`. Why not `embed`?

According to the return type annotation of `extract_message_links`, the rvalue can be None. This code seems to assume that the rvalue is always iterable.

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +515,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
 
     @Cog.listener()
     async def on_message(self, message: discord.Message) -> None:
-        """Pass `message` to `add_signals` if and only if it satisfies `is_incident`."""
+        """
+        Pass `message` to `add_signals` and `extract_message_links` if it satisfies `is_incident`.
+
+        If the message (`message`) is an incident report, then run it through `extract_message_links`
+        to get all the message link embeds (embeds which contain information about that particular
+        link).These message link embeds are then sent into the channel.
```
kwzrd, line 523: Why not just "If `message` is an incident report, ..."

Mind the missing space after the dot on the L523.

bot/exts/moderation/incidents.py
```diff
@@ -173,18 +265,33 @@ class Incidents(Cog):
...
         self.event_lock = asyncio.Lock()
         self.crawl_task = self.bot.loop.create_task(self.crawl_incidents())
 
+    async def get_webhook(self) -> None:
+        """Fetch and store message link embeds webhook, present in #incidents channel."""
+        await self.bot.wait_until_guild_available()
+        self.incidents_webhook = await self.bot.fetch_webhook(Webhooks.incidents)
+
+        if not self.incidents_webhook:
+            log.error(f"Failed to fetch incidents webhook with id `{Webhooks.incidents}`.")
```
kwzrd, line 293: This approach is problematic for a number of reasons.
* The attribute will not exist until the coro completes. This means that if a reaction triggers an event, it will fail on an attribute error. At the very least, code accessing this attribute must await the task.
* According to the docs, [`fetch_webhook`](https://discordpy.readthedocs.io/en/stable/api.html#discord.Client.fetch_webhook) will raise rather than return None if not found. This code assumes it will return None.
* If None is returned (in a hypothetical case, as per above), an error is logged, but code accessing the attribute will begin failing because it assumes the fetch to be successful.

tests/bot/exts/moderation/test_incidents.py
```diff
@@ -768,3 +787,73 @@ async def test_on_message_non_incident(self):
...
+class TestMessageLinkEmbeds(TestIncidents):
+    """Tests for `extract_message_links` coroutine."""
+
+    async def test_shorten_text(self):
+        """Test all cases of text shortening by mocking messages."""
+        tests = {
+            "thisisasingleword"*10: ('thisisasingleword'*10)[:50]+"...",
+            "\n".join("Lets make a new line test".split()): "Lets\nmake\na"+"...",
+            'Hello, World!' * 300: ('Hello, World!' * 300)[:300] + '...'
+        }
```
kwzrd, line 801: This is really difficult to read. Could we figure out a better way? Verbosity is sometimes worth it in tests, you want the cases to be obvious to whoever will have to read them if something fails.

The code seems to mix single and double quotes and doesn't seem to have spaces around some operators, is that intended?

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +515,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+
+        Deletes cache value (`message_link_embeds_cache`) of `msg_before` if it exists and removes the
+        webhook message for that particular link from the channel.
+
+        If the message edit (`msg_after`) is a incident then run it through `extract_message_links`
+        to get all the message link embeds (embeds which contain information about that particular
+        link), this message link embeds are then sent into the channel.
+
+        The edited message is also passed into `add_signals` if it is a incident message.
+        """
```
kwzrd, line 547: Have you tested whether this listener works for messages that aren't in the bot's cache? The reason why my implementation uses a raw listener is that the regular reaction listener will not fire when it cannot give you the message instance, i.e. if the bot restarts and a reaction is placed on an older message. The bot should obviously still react to such events. I believe this is explained in the docstring.

Having a quick look at the docs, it seems that the [edit](https://discordpy.readthedocs.io/en/stable/api.html#discord.on_message_edit) and [delete](https://discordpy.readthedocs.io/en/stable/api.html#discord.on_message_delete) listeners probably behave the same way. In practice, this would mean that if the bot restarts, it will not clear the quotation message for incidents sent before the current session, nor will it react to edits.

Please consult the docs, there are raw counterparts for both listeners.

tests/bot/exts/moderation/test_incidents.py
Shivansh-007: Do you have ideas for making it more readable.

No, that is not intended behavior, I will change them.

bot/exts/moderation/incidents.py
Shivansh-007: > This approach is problematic for a number of reasons.
> 
>     * The attribute will not exist until the coro completes. This means that if a reaction triggers an event, it will fail on an attribute error. At the very least, code accessing this attribute must await the task.
> 
>     * According to the docs, [`fetch_webhook`](https://discordpy.readthedocs.io/en/stable/api.html#discord.Client.fetch_webhook) will raise rather than return None if not found. This code assumes it will return None.
> 
>     * If None is returned (in a hypothetical case, as per above), an error is logged, but code accessing the attribute will begin failing because it assumes the fetch to be successful.

According to the second point, there is no need for doing this check, as it will raise when it fails to fetch the web-hook.

kwzrd: I do not follow sorry. :smile:

bot/exts/moderation/incidents.py
kwzrd: Which test is that?

Shivansh-007: `ERROR: test_on_message_incident (tests.bot.exts.moderation.test_incidents.TestOnMessage)`

bot/exts/moderation/incidents.py
Shivansh-007: Resolved in 248ecbf.

bot/exts/moderation/incidents.py
Shivansh-007: Resolved in 88cefa6.

bot/exts/moderation/incidents.py
Shivansh-007: Resolved in 88cefa6.

bot/exts/moderation/incidents.py
Shivansh-007: Resolved in efd100b and e2bf492.

bot/exts/moderation/incidents.py
Shivansh-007: Resolved in ffc39fb.

bot/exts/moderation/incidents.py
Shivansh-007: Resolved in 0b35f2a and 0c5561d.

bot/exts/moderation/incidents.py
kwzrd: It's failing because it doesn't mock the `content` attribute. So accessing it will resolve in a `MagicMock` instance. By converting it to a string in the tested code, you're running the regex on the following string:

```py
"<MagicMock name='mock.content' id='...'>"
```

Either way, I guess what I'm trying to ask is whether there's any reason to be converting the attribute to a string. [The docs](https://discordpy.readthedocs.io/en/stable/api.html#discord.Message.content) don't seem to think that it can be anything else. But even if it could, would it make sense to convert an arbitrary object into a string and be done with it?

By the way, the failing test is not testing this function directly. This code gets called by the tested function. The test does not make an attempt to assert what happens here, it only cares about whether the object is delegated to the correct functions under the correct conditions. Because the tested code was changed, the test's docstring is not entirely honest, since it does more than it's claiming to. I'm not saying that's necessarily wrong, just pointing it out.

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +486,147 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+
+        else:
+            await self.message_link_embeds_cache.set(int(message.id), int(webhook_msg.id))
+            log.trace("Message Link Embed Sent successfully!")
+            return webhook_msg.id
+
+    async def delete_msg_link_embed(self, message_id: int) -> None:
+        """Delete discord message link message found in cache for `message`."""
+        log.trace("Deleting discord links webhook message.")
+        webhook_msg_id = await self.message_link_embeds_cache.get(int(message_id))
```
kwzrd, line 619: Again, this looks off. For `message_id`, either the annotation is wrong, or it doesn't need to be converted to an int.

kwzrd: This is minor, but the docstring refers to a param name from a previous revision (presumably) that doesn't exist anymore.

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +486,147 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+                ctx = await self.bot.get_context(message)
+                embed = await make_message_link_embed(ctx, message_link[0])
+                if embed:
+                    embeds.append(embed)
+
+            return embeds
+
+        log.trace(
+            f"Skipping discord message link detection on {message.id}: message doesn't qualify."
+        )
```
kwzrd, line 581: This seems inaccurate. Detection wasn't skipped, it just didn't detect anything. :smile:

bot/exts/moderation/incidents.py
Shivansh-007: This was added, as it was having the same issue as [this](https://github.com/python-discord/bot/pull/1446#discussion_r631685982). Will do the required changes.

bot/exts/moderation/incidents.py
Shivansh-007: Ah yes 😅

bot/exts/moderation/incidents.py
Shivansh-007: Uhh, I am not completely sure what you mean here.

Shivansh-007:
@kwzrd Could you give it a review again? I also have a few comments on your previous reviews, Thanks!

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,60 @@ def has_signals(message: discord.Message) -> bool:
...
+        Author: @Jason Terror ♦ (736234578745884682)
+        Channel: Special/#bot-commands (814190307980607493)
+        Content: This is a very important message!
+    """
+    try:
+        message: discord.Message = await MessageConverter().convert(ctx, message_link)
+
+    except discord.DiscordException as e:
+        log.exception(f"Failed to make message link embed for '{message_link}', raised exception: {e}")
+
```
bast0006, line 175: I think we should specifically handle a few cases here that have been missed:
- if a message has been deleted, the converter will raise NotFound. We should have a special message for this case
- When it's deleted, we can search through #message-change-log in our message cache looking for the appropriate message id: {id} string, to link there and extract content from there instead.
- since incidents is within helper view, we should at the very least restrict this converter from accessing messages inside, say, #admins with some kind of a channel check.

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +486,147 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        Uses the `webhook` passed in as a parameter to send
+        the embeds in the `webhook_embed_list` parameter.
+
+        After sending each webhook it maps the `message.id`
+        to the `webhook_msg_ids` IDs in the async redis-cache.
+        """
+        try:
+            webhook_msg = await webhook.send(
+                embeds=[embed for embed in webhook_embed_list if embed is not None],
```
bast0006, line 600: No need to check for None here, as `webhook_embed_list` should never contain None.

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +486,147 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+    @Cog.listener()
+    async def on_raw_message_delete(self, payload: discord.RawMessageDeleteEvent) -> None:
+        """
+        Delete message link embeds for `payload.message_id`.
+
+        Search through the cache for message, if found delete it from cache and channel.
+        """
+        await self.delete_msg_link_embed(payload.message_id)
+
+    async def extract_message_links(self, message: discord.Message) -> t.Optional[t.List[discord.Embed]]:
```
bast0006, line 555: Consider changing this to always return a list, just empty if no embeds are found. Should simplify processing?

bot/exts/moderation/incidents.py
```diff
@@ -402,10 +486,147 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+                wait=True,
+            )
+
+        except discord.DiscordException:
+            log.exception(
+                f"Failed to send message link embed {message.id} to #incidents."
+            )
+
+        else:
+            await self.message_link_embeds_cache.set(int(message.id), int(webhook_msg.id))
```
bast0006, line 612: I don't think these need int() casts.. do they?

bast0006 (changes requested):

bot/exts/moderation/incidents.py
Shivansh-007: This was added as it was having the same issue as [this](https://github.com/python-discord/bot/pull/1446#discussion_r631685982).

bast0006: Try not adjusting code if there are tests failing and you can't figure out how to reproduce the test failure with manual code. `message.id` should never be non-int, so if you're having to cast there you're having other problems (such as with the test). Otherwise the tests become part of your code, and are not really suitable as tests any longer. 

This issue is because the Mock being used isn't having .id set appropriately. Try using `breakpoint()` or another debugger when this errors to figure out exactly what is being passed here during your tests.

bot/exts/moderation/incidents.py
Shivansh-007: I don't see how it would simplify the process, we would either need to check if `extract_message_links` returns an empty list or `None`?

jb3:
ℹ️ **Note to merger:** This PR should be squashed to avoid a wildly inflated git history.

Xithrius:
I realize that this PR was never approved to go forth with, planning was stated [here](https://github.com/python-discord/bot/issues/1226#issuecomment-790176085).

Please wait for approval before going ahead with PRs next time.

bot/exts/moderation/incidents.py
Shivansh-007: Mentioned in ed30b1e and 94bcc93.

mbaruh:
Can you elaborate here what was discussed?

jb3:
> Can you elaborate here what was discussed?

@mbaruh We talked about whether the editing things were necessary given that we already have edit logs which not only show edited message but all iterations of edits.

The changes mean that the message will be relayed as-is at time of reporting (which in my mind at least makes more sense), it will still be cleaned up once the incident is actioned.

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,94 @@ def has_signals(message: discord.Message) -> bool:
...
+    lines = text.count("\n")
+    # Limit to a maximum of three lines
+    if lines > 3:
+        text = "\n".join(line for line in text.split('\n')[:3])
+    # If it is a single word, then truncate it to 50 characters
+    if text.count(" ") < 1:
+        text = text[:50]
+    # Truncate text to a maximum of 300 characters
+    if len(text) > 300:
+        text = text[:300]
```
mbaruh, line 153: This should be the first operation, and then count characters in the reduced string.

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,94 @@ def has_signals(message: discord.Message) -> bool:
     return ALL_SIGNALS.issubset(own_reactions(message))
 
 
+def shorten_text(text: str) -> str:
+    """Truncate the text if there are over 3 lines or 300 characters, or if it is a single word."""
+    original_length = len(text)
+    lines = text.count("\n")
+    # Limit to a maximum of three lines
+    if lines > 3:
+        text = "\n".join(line for line in text.split('\n')[:3])
```
mbaruh, line 147: Nitpicking, but can be slightly improved with maxsplit=3

bot/exts/moderation/incidents.py
```diff
@@ -180,18 +277,33 @@ class Incidents(Cog):
...
 
+    # This dictionary maps an incident report message to the message link embed's ID
+    # RedisCache[discord.Message.id, discord.Message.id]
+    message_link_embeds_cache = RedisCache()
+
     def __init__(self, bot: Bot) -> None:
         """Prepare `event_lock` and schedule `crawl_task` on start-up."""
         self.bot = bot
 
+        self.bot.loop.create_task(self.get_webhook())
```
mbaruh, line 294: Can you adjust it to the approach in #1825 ?

bot/exts/moderation/incidents.py
mbaruh: This doesn't solve the issues kwzrd outlined. What will happen is:
* An exception will be raised for the missing webhook.
* An exception will be raised for every single operation in the cog thereafter which tries to use this attribute.

I'm fine with an exception being logged in Sentry for the missing webhook, in which case the check above is not necessary, but the rest of the cog then should not assume the attribute exists, and only use the feature implemented in this PR if it has the value set (with a default of None set in the init).

bot/exts/moderation/incidents.py
```diff
@@ -291,6 +403,9 @@ async def process_event(self, reaction: str, incident: discord.Message, member:
         This ensures that if there is a racing event awaiting the lock, it will fail to find the
         message, and will abort. There is a `timeout` to ensure that this doesn't hold the lock
         forever should something go wrong.
+
+        Deletes cache value (`message_link_embeds_cache`) of `msg_before` if it exists and removes the
```
mbaruh, line 407: Got a bit confused by the wording at first
```suggestion
        Deletes cache value (`message_link_embeds_cache`) of `msg_before` if it exists. It then removes the
```
Also I don't see any variable called `msg_before`. It will also be helpful if you mention what that variable represents in your own words, since people will be first reading the docstring and then the code.

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+    @Cog.listener()
+    async def on_raw_message_delete(self, payload: discord.RawMessageDeleteEvent) -> None:
+        """
+        Delete message link embeds for `payload.message_id`.
+
+        Search through the cache for message, if found delete it from cache and channel.
+        """
+        await self.delete_msg_link_embed(payload.message_id)
+
+    async def extract_message_links(self, message: discord.Message) -> t.Optional[t.List[discord.Embed]]:
```
mbaruh, line 564: Since we're now on 3.9 (this PR has been open for awhile) you can use `list` instead.

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        """
+        Checks if there's any message links in the text content.
+
+        Then passes the the message_link into `make_message_link_embed` to format a
```
mbaruh, line 568:
```suggestion
        Then passes the message_link into `make_message_link_embed` to format an
```

tests/bot/exts/moderation/test_incidents.py
mbaruh: I'd rather these were string literals without the extra processing, even if it takes space. Just clear cases of input and output.

mbaruh (changes requested):

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,94 @@ def has_signals(message: discord.Message) -> bool:
...
 
+def shorten_text(text: str) -> str:
+    """Truncate the text if there are over 3 lines or 300 characters, or if it is a single word."""
+    original_length = len(text)
+    lines = text.count("\n")
+    # Limit to a maximum of three lines
+    if lines > 3:
+        text = "\n".join(line for line in text.split('\n')[:3])
+    # If it is a single word, then truncate it to 50 characters
+    if text.count(" ") < 1:
```
mbaruh, line 149: Alternatively find the first occurence of a space so you don't have to keep counting if you do find one.

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,94 @@ def has_signals(message: discord.Message) -> bool:
...
+        text = text[:50]
+    # Truncate text to a maximum of 300 characters
+    if len(text) > 300:
+        text = text[:300]
+    # Add placeholder if the text was shortened
+    if len(text) < original_length:
```
mbaruh, line 155: It doesn't seem like the string is stripped before this line as requested in https://github.com/python-discord/bot/pull/1446#issuecomment-830607365

mbaruh (changes requested):

bot/exts/moderation/incidents.py
```diff
@@ -21,6 +23,12 @@
 # Seconds for `crawl_task` to sleep after adding reactions to a message
 CRAWL_SLEEP = 2
 
+DISCORD_MESSAGE_LINK_RE = re.compile(
+    r"(https?:\/\/(?:(ptb|canary|www)\.)?discord(?:app)?\.com\/channels\/"
+    r"[0-9]{15,21}"
+    r"\/[0-9]{15,21}\/[0-9]{15,21})"
```
Bluenix2, line 29:
```suggestion
    r"[0-9]{15,20}"
    r"\/[0-9]{15,20}\/[0-9]{15,20})"
```
```python
>>> int('1'*64, 2)
18446744073709551615
>>> len(str(18446744073709551615))
20
```
The biggest size a Discord snowflake can be is 20.

bot/exts/moderation/incidents.py
Bluenix2, line 147:
```suggestion
    lines = text.split("\n")
    # Limit to a maximum of three lines
    if len(lines) > 3:
        text = "\n".join(line for line in lines[:3])
```

bot/exts/moderation/incidents.py
Bluenix2: I agree, this could come before the one-word count.

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,94 @@ def has_signals(message: discord.Message) -> bool:
...
+    The Embed would contain the following information -->
+        Author: @Jason Terror ♦ (736234578745884682)
+        Channel: Special/#bot-commands (814190307980607493)
+        Content: This is a very important message!
+    """
+    embed = None
+
+    try:
+        message: discord.Message = await MessageConverter().convert(ctx, message_link)
+
```
Bluenix2, line 174:
```suggestion
```

bot/exts/moderation/incidents.py
```diff
@@ -130,6 +138,94 @@ def has_signals(message: discord.Message) -> bool:
...
+    except MessageNotFound:
+        mod_logs_channel = ctx.bot.get_channel(Channels.mod_log)
```
Bluenix2, line 176: What if we have an incorrect config setup and the channel cannot be found?

bot/exts/moderation/incidents.py
```diff
@@ -180,18 +277,33 @@ class Incidents(Cog):
...
         """Prepare `event_lock` and schedule `crawl_task` on start-up."""
         self.bot = bot
 
+        self.bot.loop.create_task(self.get_webhook())
+
         self.event_lock = asyncio.Lock()
         self.crawl_task = self.bot.loop.create_task(self.crawl_incidents())
 
+    async def get_webhook(self) -> None:
+        """Fetch and store message link embeds webhook, present in #incidents channel."""
```
Bluenix2, line 300: Did you mean `embed's`? Is there a missing `the` here? This docstring is a bit confusing.

bot/exts/moderation/incidents.py
```diff
@@ -180,18 +277,33 @@ class Incidents(Cog):
...
     def __init__(self, bot: Bot) -> None:
         """Prepare `event_lock` and schedule `crawl_task` on start-up."""
         self.bot = bot
 
+        self.bot.loop.create_task(self.get_webhook())
+
         self.event_lock = asyncio.Lock()
         self.crawl_task = self.bot.loop.create_task(self.crawl_incidents())
 
+    async def get_webhook(self) -> None:
```
Bluenix2, line 299:
```suggestion
    async def fetch_webhook(self) -> None:
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        These message link embeds are then sent into the channel.
+
+        Also passes the message into `add_signals` if the message is an incident.
+        """
         if is_incident(message):
             await add_signals(message)
 
+            webhook_embed_list = await self.extract_message_links(message)
+            if webhook_embed_list:
+                await self.send_message_link_embeds(webhook_embed_list, message, self.incidents_webhook)
```
Bluenix2, line 553:
```suggestion
        if not is_incident(message):
            return

        await add_signals(message)

        if embed_list := await self.extract_message_links(message):
            await self.send_message_link_embeds(embed_list, message, self.incidents_webhook)
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        """
+        Delete message link embeds for `payload.message_id`.
+
+        Search through the cache for message, if found delete it from cache and channel.
+        """
+        await self.delete_msg_link_embed(payload.message_id)
+
+    async def extract_message_links(self, message: discord.Message) -> t.Optional[t.List[discord.Embed]]:
+        """
+        Checks if there's any message links in the text content.
```
Bluenix2, line 566:
```suggestion
        Check if there's any message links in the text content.
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+
+        Then passes the the message_link into `make_message_link_embed` to format a
+        embed for it containing information about the link.
+
+        As discord only allows a max of 10 embeds in a single webhook, just send the
```
Bluenix2, line 571:
```suggestion
        As Discord only allows a max of 10 embeds in a single webhook, just send the
```

bot/exts/moderation/incidents.py
Bluenix2, line 568:
```suggestion
        Then pass the the message_link into `make_message_link_embed` to format a
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        first 10 embeds and don't care about the rest.
+
+        If no links are found for the message, it logs a trace statement.
```
Bluenix2, line 574:
```suggestion
        If no links are found for the message, just log a trace statement.
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+                ctx = await self.bot.get_context(message)
+                embed = await make_message_link_embed(ctx, message_link[0])
+                if embed:
+                    embeds.append(embed)
+
+            return embeds
+
+        log.trace(
+            f"No message links detected on incident message with id {message.id}."
+        )
```
Bluenix2, line 590:
```suggestion
        message_links = DISCORD_MESSAGE_LINK_RE.findall(str(message.content))

        if not message_links:
            log.trace(
                f"No message links detected on incident message with id {message.id}."
            )
            return

        embeds = []
        for message_link in message_links[:10]:
            ctx = await self.bot.get_context(message)
            embed = await make_message_link_embed(ctx, message_link[0])
            if embed:
                embeds.append(embed)

        return embeds        
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+
+    async def send_message_link_embeds(
+            self,
+            webhook_embed_list: t.List,
+            message: discord.Message,
+            webhook: discord.Webhook,
+    ) -> t.Optional[int]:
+        """
+        Send Message Link Embeds to #incidents channel.
```
Bluenix2, line 599:
```suggestion
        Send message link embeds to #incidents channel.
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+
+        Uses the `webhook` passed in as a parameter to send
```
Bluenix2, line 601:
```suggestion
        Using the `webhook` passed in as a parameter to send
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        the embeds in the `webhook_embed_list` parameter.
+
+        After sending each webhook it maps the `message.id`
+        to the `webhook_msg_ids` IDs in the async redis-cache.
+        """
+        try:
+            webhook_msg = await webhook.send(
+                embeds=[embed for embed in webhook_embed_list if embed],
```
Bluenix2, line 609:
```suggestion
                embeds=[embed for embed in webhook_embed_list if embed],
```
I don't follow why this is necessary? When is the item in the list falsely?

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+                username=sub_clyde(message.author.name),
+                avatar_url=message.author.avatar_url,
+                wait=True,
+            )
+
+        except discord.DiscordException:
+            log.exception(
+                f"Failed to send message link embed {message.id} to #incidents."
+            )
+
```
Bluenix2, line 619:
```suggestion
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        else:
+            await self.message_link_embeds_cache.set(message.id, webhook_msg.id)
+            log.trace("Message Link Embed Sent successfully!")
```
Bluenix2, line 622:
```suggestion
            log.trace("Message link embeds sent successfully to #incidents!")
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+            message: discord.Message,
+            webhook: discord.Webhook,
+    ) -> t.Optional[int]:
+        """
+        Send Message Link Embeds to #incidents channel.
+
+        Uses the `webhook` passed in as a parameter to send
+        the embeds in the `webhook_embed_list` parameter.
+
+        After sending each webhook it maps the `message.id`
```
Bluenix2, line 604:
```suggestion
        After sending each embed it maps the `message.id`
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+                f"Failed to send message link embed {message.id} to #incidents."
+            )
+
+        else:
+            await self.message_link_embeds_cache.set(message.id, webhook_msg.id)
+            log.trace("Message Link Embed Sent successfully!")
+            return webhook_msg.id
+
+    async def delete_msg_link_embed(self, message_id: int) -> None:
+        """Delete discord message link message found in cache for `message`."""
```
Bluenix2, line 626:
```suggestion
        """Delete the Discord message link message found in cache for `message_id`."""
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        log.trace("Deleting discord links webhook message.")
```
Bluenix2, line 627:
```suggestion
        log.trace("Deleting Discord message link's webhook message.")
```

bot/exts/moderation/incidents.py
```diff
@@ -418,10 +536,106 @@ async def on_raw_reaction_add(self, payload: discord.RawReactionActionEvent) ->
...
+        webhook_msg_id = await self.message_link_embeds_cache.get(int(message_id))
+
+        if webhook_msg_id:
+            try:
+                await self.incidents_webhook.delete_message(webhook_msg_id)
+            except discord.errors.NotFound:
+                log.trace(f"Incidents message link embed (`{webhook_msg_id}`) has already been deleted, skipping.")
+
+        await self.message_link_embeds_cache.delete(message_id)
+        log.trace("Successfully deleted discord links webhook message.")
```
Bluenix2, line 637: There's a lot of trace calls here, is it not better if we somehow summarize it in the end of what happened?

Xithrius:
@Shivansh-007 will you be able to resolve these requests?

Shivansh-007:
Hey @Xithrius,

Unfortunately I haven't been getting enough time to mention the reviews due to my exams going on, if this is PR is on a high priority (or any of my other PRs), someone else can take them up as I won't be able to mention them until 15th October at-least. Otherwise, I would get them mentioned as soon a I am back from the break. And thanks bluenix and zig for the reviews!

You would probably forgot, I had pinged you in #dev-contrib with the same before I went off into a break.

Thanks

Shivansh-007:
Update: Good news! I am almost done with my exams, would be going through the reviews and mentioning the required changes within this week, I would be focusing on this PR over the other 2 first 🙂

mbaruh:
Why only for a single word? 😅 
The idea is that you might end up with a slice ending at a space. It's best to strip right before checking whether the length of the string changed.

bot/exts/moderation/incidents.py
```diff
@@ -181,18 +284,35 @@ class Incidents(Cog):
...
         self.crawl_task = scheduling.create_task(self.crawl_incidents(), event_loop=self.bot.loop)
 
+    async def fetch_webhook(self) -> None:
+        """Fetch the incidents webhook object, so we can post message link embeds to it."""
+        await self.bot.wait_until_guild_available()
+
+        try:
+            self.incidents_webhook = await self.bot.fetch_webhook(Webhooks.incidents)
+        except discord.HTTPException:
+            log.error(f"Failed to fetch incidents webhook with id `{Webhooks.incidents}`.")
```
Bluenix2, line 314: Can we differentiate between not finding the webhook and missing permissions?

bot/exts/moderation/incidents.py
```diff
@@ -131,6 +139,100 @@ def has_signals(message: discord.Message) -> bool:
...
+                f" not sending message link embed for {message_link}"
+            )
+            return
+
+        embed = discord.Embed(
+            colour=discord.Colour.gold(),
+            description=(
+                f"**Author:** {format_user(message.author)}\n"
+                f"**Channel:** {channel.mention} ({channel.category}/#{channel.name})\n"
+            ),
```
Bluenix2, line 224: Wouldn't it be easier to use the author field here?

bot/exts/moderation/incidents.py
```diff
@@ -131,6 +139,100 @@ def has_signals(message: discord.Message) -> bool:
...
+        channel = message.channel
+        if not channel.permissions_for(channel.guild.get_role(Roles.helpers)).view_channel:
+            log.info(
+                f"Helpers don't have read permissions in #{channel.name},"
+                f" not sending message link embed for {message_link}"
+            )
+            return
+
+        embed = discord.Embed(
+            colour=discord.Colour.gold(),
```
Bluenix2, line 220: I think that [discord.Colour.dark_theme()](https://discordpy.readthedocs.io/en/master/api.html#discord.Colour.dark_theme) might fit better here?

The yellowish-gold is pretty obnoxious.

mbaruh: I actually like the gold color 😔 I always found the transparent color pretty lackluster.

bot/exts/moderation/incidents.py
mbaruh: I don't think you can have mentions there, can you?

bot/exts/moderation/incidents.py
mbaruh: You mean catching `discord.Forbidden` as well? I don't think this is something we bother doing anywhere with webhooks. If the bot doesn't have perms for it we probably have bigger problems 😅

bot/exts/moderation/incidents.py
Bluenix2: Ah right, maybe not. That could be an issue.

bot/exts/moderation/incidents.py
```diff
@@ -131,6 +139,100 @@ def has_signals(message: discord.Message) -> bool:
...
+            ),
+            timestamp=message.created_at
+        )
+        embed.add_field(
+            name="Content",
+            value=shorten_text(message.content)
+        )
+        embed.set_footer(text=f"Message ID: {message.id}")
+
+    return embed
```
Bluenix2, line 233: Can we also display attachments if there are some? You don't need to download them. Just use [`discord.Embed.set_image`](https://discordpy.readthedocs.io/en/latest/api.html#discord.Embed.set_image) with the attachment link.

bot/exts/moderation/incidents.py
Bluenix2: I don't really care all that much, so I'll just mark this as resolved. It's fine

bot/exts/moderation/incidents.py
Bluenix2: I tried it just now, you cannot. This style of embed looks good enough.

bot/exts/moderation/incidents.py
```diff
@@ -131,6 +139,100 @@ def has_signals(message: discord.Message) -> bool:
...
+                f"Helpers don't have read permissions in #{channel.name},"
+                f" not sending message link embed for {message_link}"
+            )
+            return
+
+        embed = discord.Embed(
+            colour=discord.Colour.gold(),
+            description=(
+                f"**Author:** {format_user(message.author)}\n"
+                f"**Channel:** {channel.mention} ({channel.category}/#{channel.name})\n"
```
Bluenix2, line 223: This doesn't account for threads

Bluenix2 (approved):
Thanks for this Shivansh-007, nice work!

bot/exts/moderation/incidents.py
```diff
@@ -131,6 +139,100 @@ def has_signals(message: discord.Message) -> bool:
...
+                    )
+                )
+        if not embed:
+            embed = discord.Embed(
+                colour=discord.Colour.red(),
+                title="Bad Message Link",
+                description=f"Message {message_link} not found."
+            )
+    except discord.DiscordException as e:
+        log.exception(f"Failed to make message link embed for '{message_link}', raised exception: {e}")
```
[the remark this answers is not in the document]
jchristgit: IMO if we want to explicitly pass the exception, then this is the cleaner way to do it.
Otherwise, we end up duplicating information in the traceback already printed by `log.exception` and the separate exception printed here.
```suggestion
    except discord.DiscordException as e:
        log.exception(f"Failed to make message link embed for '{message_link}'", exc_info=e)
```

PR #1450, opened 2021-03-07
Only allow some characters within a request URL to PyPi.

Xithrius:
If someone spams a bunch of characters, we don't want to send a request to PyPi with a bunch of gibberish. This also protects against injections (if possible, if any).

Some examples of cases with illegal characters within them:
image

### DIFF
--- a/bot/exts/info/pypi.py
+++ b/bot/exts/info/pypi.py
@@ -1,6 +1,7 @@
 import itertools
 import logging
 import random
+import re
 
 from discord import Embed
 from discord.ext.commands import Cog, Context, command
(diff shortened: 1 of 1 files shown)


Xithrius:
There are some changes from a different branch on here. Fixing.

bot/exts/info/pypi.py
```diff
@@ -32,6 +36,11 @@ async def get_package_info(self, ctx: Context, package: str) -> None:
         )
         embed.set_thumbnail(url=PYPI_ICON)
 
+        if (character := re.search(ILLEGAL_CHARACTERS, package)) is not None:
+            embed.description = f"Illegal character passed into command: '{escape_markdown(character.group(0))}'"
```
Shivansh-007, line 40: Missing a `.` full stop at the end of the sentence,

```suggestion
            embed.description = f"Illegal character passed into command: '{escape_markdown(character.group(0))}'."
```

Shivansh-007 (approved):
Thanks! Looks Good To Me! Just one thing, not sure if it is needed or not tho.

Tested locally.

bot/exts/info/pypi.py
Xithrius: I tested this a while ago. I didn't think it looked right. Here's a comparison.

With period:
image

And without:
image

Shivansh-007: Yeah, even I had the same feeling.

ChrisLovering (approved):
All looks good to me!

scragly (approved):

PR #1452, opened 2021-03-08
Remove invoked command and message after failure.

### LINKED ISSUE #1443: Delete messages when `!pypi` module is invalid
When users use the `!pypi` command with an invalid module, you currently get a response saying `"Package could not be found."`. 

What I'm proposing it that after a delay (I believe we have `bot.constants.RedirectOutput.delete_delay` for this) the bot should delete both its response and the invocation message, like [`!d` currently does](https://github.com/python-discord/bot/blob/master/bot/exts/info/doc.py#L363-L365).

This would help reduce channel flooding when someone is trying to remember the name of a module, and since it's not a trivial change I'm happy to try and implement this myself if it's something we decide we want.

Xithrius:
Closes #1443.

Removes the invoked command and any error embed the bot sent (from the `pypi` cog) after a constant amount of time that's set in the config.

### DIFF
--- a/bot/exts/info/pypi.py
+++ b/bot/exts/info/pypi.py
@@ -8,7 +8,7 @@ from discord.ext.commands import Cog, Context, command
 from discord.utils import escape_markdown
 
 from bot.bot import Bot
-from bot.constants import Colours, NEGATIVE_REPLIES
+from bot.constants import Colours, NEGATIVE_REPLIES, RedirectOutput
 
 URL = "https://pypi.org/pypi/{package}/json"
 FIELDS = ("author", "requires_python", "summary", "license")
@@ -17,6 +17,7 @@ PYPI_ICON = "https://cdn.discordapp.com/emojis/766274397257334814.png"
 PYPI_COLOURS = itertools.cycle((Colours.yellow, Colours.blue, Colours.white))
 
 ILLEGAL_CHARACTERS = re.compile(r"[^a-zA-Z0-9-.]+")
+INVALID_INPUT_DELETE_DELAY = RedirectOutput.delete_delay
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


Xithrius:
Branch name should be `pypi-delete-on-error`, whoops.

bot/exts/info/pypi.py
```diff
@@ -17,6 +17,7 @@
 PYPI_COLOURS = itertools.cycle((Colours.yellow, Colours.blue, Colours.white))
 
 ILLEGAL_CHARACTERS = re.compile(r"[^a-zA-Z0-9-.]+")
+NOT_FOUND_DELETE_DELAY = RedirectOutput.delete_delay
```
Shivansh-007, line 20: Not sure if it is needed as a wrapper but I think it is good to have it since it describes the purpose of the constant better than `RedirectOutput.delete_delay`.

bot/exts/info/pypi.py
```diff
@@ -36,42 +37,49 @@ async def get_package_info(self, ctx: Context, package: str) -> None:
...
+                            )
 
-            else:
-                embed.description = "There was an error when fetching your PyPi package."
-                log.trace(f"Error when fetching PyPi package: {response.status}.")
+                    error = False
 
-        await ctx.send(embed=embed)
+                else:
+                    embed.description = "There was an error when fetching your PyPi package."
```
Shivansh-007, line 75: I feel setting `error` To `False` as default is better and then setting it to `True` here.

But yeah that’s just my personal preference.

Shivansh-007 (approved):
Thanks! Looks Good To Me! Just two comments.

bot/exts/info/pypi.py
Numerlor, line 20: I think with multiple renamed uses like this it'd be better to move it value to a more appropriate key, although I'm not sure where something general like it would fit

Xithrius: I'm not sure either, but I think this should be discussed within the future. 

I'm going to rename `NOT_FOUND_DELETE_DELAY` to `ON_ERROR_DELETE_DELAY` to describe the situations more accurately.

fisher60 (approved):
LGTM

Akarys42 (approved):
Setting error to true then false feels a bit weird, but I guess that works. Thanks sir Xith.

PR #1455, opened 2021-03-08
Allow the sf command to take more than one snowflake at the time

Akarys42:
Here is how it looks with many snowflakes:

image

To avoid users making the bot hit ratelimits, it is still limited to one snowflake per invocation for non-staffers

image

### DIFF
--- a/bot/exts/utils/utils.py
+++ b/bot/exts/utils/utils.py
@@ -2,7 +2,7 @@ import difflib
 import logging
 import re
 import unicodedata
-from typing import Tuple, Union, List
+from typing import Tuple, Union
 
 from discord import Colour, Embed, utils
 from discord.ext.commands import BadArgument, Cog, Context, clean_content, command, has_any_role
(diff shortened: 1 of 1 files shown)


bot/exts/utils/utils.py
```diff
@@ -156,18 +157,22 @@ async def zen(self, ctx: Context, *, search_value: Union[int, str, None] = None)
...
-            description=f"**Created at {created_at}** ({time_since(created_at, max_units=3)}).",
-            colour=Colour.blue()
-        )
-        embed.set_author(
-            name=f"Snowflake: {snowflake}",
-            icon_url="https://github.com/twitter/twemoji/blob/master/assets/72x72/2744.png?raw=true"
-        )
-        await ctx.send(embed=embed)
+        if len(snowflakes) > 1 and await has_no_roles_check(ctx, *STAFF_ROLES):
+            raise BadArgument("Cannot process more than one snowflake in one invocation.")
```
ks129, line 163: Please mention here that this is only case if user is not staff member.

jb3: That's probably not necessary. We can let the staff know that or alternatively add it to the help section of the command.

Akarys42: I don’t think it is necessary, we don’t mention that only moderators can use `!user` on other users than themselves. I don’t mind adding it to the docstring but I feel like that will just add noise since the new command signature will reflect this change, what do you think?

MarkKoz (approved):

Xithrius (approved):

PR #1460, opened 2021-03-11
Compose: read all env vars from '.env'

kwzrd:
When running with Compose, the `GITHUB_API_KEY` env var isn't being read from `.env` into the container, causing harsh rate limits when making unauthenticated requests. With this PR, we'll be reading all env vars from `.env` automatically.


MarkKoz:
It could just read everything from the file with `env_file: .env`. Why don't we do that?

kwzrd:
Okay, let's do that.

Xithrius (approved):

MarkKoz (approved):

PR #1465, opened 2021-03-13
Use .gitattributes to normalise line endings on check-in

MarkKoz:
Remove the mixed line endings pre-commit hook because it is obsolete.

Relying on git to handle line endings means contributors have more flexibility with which line endings they want to use on check-out. The settings in `.gitattributes` only impose which line endings will be used upon check-in (`LF`), which should not impact local development; git will still respect the `core.eol` and `core.autocrlf` settings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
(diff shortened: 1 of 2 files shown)


HassanAbouelela (approved):
I always appreciate changes that lower the barrier to entry. I've tested this setup, and it seems hard to mess things up, intentionally or accidentally. Thank you.

Den4200 (approved):

PR #1467, opened 2021-03-15
Code block: remove null bytes before parsing AST

MarkKoz:
`ast.parse` raises a ValueError complaining that source code strings cannot contain null bytes. It seems like they may accidentally get pasted into Discord by users sometimes. I couldn't manage to paste a null byte myself.

Fixes [BOT-XR](https://sentry.io/organizations/python-discord/issues/2273274549/)


Akarys42 (approved):
Good stuff, cheers!

kosayoda (approved):
solid

PR #1468, opened 2021-03-15
Account for defcon threshold being None when displaying server info

MarkKoz:
Fixes [BOT-XK](https://sentry.io/organizations/python-discord/issues/2264325856).

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -64,7 +64,8 @@ class Information(Cog):
 
         defcon_info = ""
         if cog := self.bot.get_cog("Defcon"):
-            defcon_info = f"Defcon threshold: {humanize_delta(cog.threshold)}\n"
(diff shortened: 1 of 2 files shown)


mbaruh:
This works as a fix for the server command, but there's an underlying issue which is that `self.threshold` is set to None by default, but then becomes an empty `relativedelta()` upon reset. My plan was to replace the None's with a `relativedelta()`, as it is apparently false-y anyway, and then it's possible it's not needed to change the server command.

MarkKoz:
Yeah you could avoid it being None, but it still seemed to be an intentional choice to display the dash character rather than "0 days", so that is what I stuck to here.

mbaruh (approved):
Yeah that seems reasonable

Akarys42 (approved):
Lgtm

PR #1470, opened 2021-03-16
Extend close time logic to differentiate between the claimant and other users

### LINKED ISSUE #1451: Improve help channel closing by differentiating between the OP and other users.
### Goal
Implement a system that closes help channels more quickly if the OP (the person who originally opened the channel) has not responded in a while. This is to prevent channels from going stale for a very long time.

### Implementation suggestion

The initial implementation suggestion is that 10 minutes after the last user message, and 30 minutes after the last OP message the channel will be closed.

This issue is meant as a place to discuss different implementations and ideas relating to the Goal.

ChrisLovering:
Closes #1451

This extends the closing logic for help channels to closed based on the claimant's last message time and last message in general.

This results in the help channel closing time being reduced if only others are chatting, and the claimant has already left.

Thanks to laundmo, Darr, Griff, LX and Hemlock for the group coding on this PR :D


laundmo (dismissed):
I was there in VC watching this be implemented and tested, all seems good to me.

Shivansh-007 (changes requested):
Few nitpicks:

bot/exts/help_channels/_message.py
```diff
@@ -28,7 +28,7 @@
 
 AVAILABLE_TITLE = "Available help channel"
 
-AVAILABLE_FOOTER = f"Closes after {constants.HelpChannels.idle_minutes} minutes of inactivity or when you send !close."
+AVAILABLE_FOOTER = "Closes after being idle for some time, or when you send !close."
```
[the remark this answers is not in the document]
laundmo: we (chrisj, me, darr) all couldn't really come up with a one line explanation of the new system. if you would have one, that would be greatly appreciated.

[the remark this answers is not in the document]
ChrisLovering: I've slightly changed the wording of this line as discussed in #dev-contrib.
I think trying to explain how this new system works in a sentence would over simplify it too much, and cause confusion/more questions.

People who really want to know how it works will ask anyway, and we can then explain it properly.

bot/exts/help_channels/_cog.py
```diff
@@ -427,6 +430,21 @@ async def on_message(self, message: discord.Message) -> None:
...
+                await _caches.last_message_times.set(
+                    message.channel.id,
+                    message.created_at.timestamp()
+                )
+        elif channel_utils.is_in_category(message.channel, constants.Categories.help_in_use):
+            # Overwrite the claimant message time, if its from the claimant.
+            if message.author == await _caches.claimants.get(message.channel.id):
+                await _caches.claimant_last_message_times(message.channel.id, message.created_at.timestamp())
+
+            await _caches.last_message_times.set(message.channel.id, message.created_at.timestamp())
```
MarkKoz, line 447: This could be moved into `check_for_answer`  since it shares similar logic. In fact, I think you could get rid of `last_message_times` and instead repurpose the unanswered cache by storing times instead of booleans (where None would mean unanswered).

MarkKoz (changes requested):

MarkKoz:
Has it been decided to "take the hit" when the bot is offline and can't cache messages that get sent?

laundmo:
@MarkKoz might need to add a condition to the old method fallback, that checks for bot uptime smaller than idle_minutes_claimant so that there is at least that much time after a restart after which channels wont be unduly closed.

that is, if i understood correctly what you were referring to.

MarkKoz:
> 
> 
> @MarkKoz might need to add a condition to the old method fallback, that checks for bot uptime smaller than idle_minutes_claimant so that there is at least that much time after a restart after which channels wont be unduly closed.
> 
> that is, if i understood correctly what you were referring to.

What I was referring to is the cache not having the latest times due to the bot going offline and more messages being sent in channels. Upon going online again, the bot may prematurely close channels due to the outdated cache leading to the perception that the channels have been inactive.

I don't understand what you're proposing.

laundmo:
@MarkKoz as it stands currently, it will fall back to the old behavior if the cache isn't available (empty/ returns None)

I am proposing extending the fallback conditions to always use the old method for a while after startup, as it does not rely on cache. This way, instead of channels being closed prematurely they might take a bit longer to close.

bot/exts/help_channels/_channel.py
```diff
@@ -25,23 +25,43 @@ def get_category_channels(category: discord.CategoryChannel) -> t.Iterable[disco
...
+
+    # We want to get the time at which a channel should be closed.
+    last_message_time = datetime.fromtimestamp(last_message_time)
+    claimant_last_message_time = datetime.fromtimestamp(claimant_last_message_time)
+
+    last_message_time += timedelta(minutes=idle_minutes)
+    claimant_last_message_time += timedelta(minutes=constants.HelpChannels.idle_minutes_claimant)
+
+    # The further away closing time is what we should use.
+    closing_time = max(claimant_last_message_time, last_message_time)
```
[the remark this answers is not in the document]
ChrisLovering: If `_message.is_empty(channel)` returns `True`, this indicates that there isn't a message in the channel. So both cache's would be empty. This would mean we enter the if statement and return before this code is hit.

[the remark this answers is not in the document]
MarkKoz: That is incorrect because the cache doesn't get cleared when the channel goes dormant. `is_empty` doesn't mean the channel has no messages at all; it basically means the channel's latest message is the bot's embed.

MarkKoz:
> 
> 
> @MarkKoz as it stands currently, it will fall back to the old behavior if the cache isn't available (empty/ returns None)
> 
> I am proposing extending the fallback conditions to always use the old method for a while after startup, as it does not rely on cache. This way, instead of channels being closed prematurely they might take a bit longer to close.

Yes, that could work. It'd have to reschedule it for the maximum time.

bot/exts/help_channels/_channel.py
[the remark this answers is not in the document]
ChrisLovering: I added code to clear the caches when the channel is unclaimed in a recent commit, so would that mean it's now correct?
https://github.com/python-discord/bot/pull/1470/files#diff-c31098978b8d838eadb52b2bf23494072cc66686e738d7f99c4f59e6495bfe6fR380-R381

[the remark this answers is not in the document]
MarkKoz: Maybe, but is there a way to fix it without relying on the cache being cleared?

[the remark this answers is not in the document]
ChrisLovering: How do you feel about returning early, within the `if await _message.is_empty(channel):` block?

bot/exts/help_channels/_channel.py
```diff
@@ -26,19 +26,20 @@ def get_category_channels(category: discord.CategoryChannel) -> t.Iterable[disco
 
 
 async def get_closing_time(channel: discord.TextChannel) -> datetime:
-    """Return the timestamp at which the given help `channel` should be closed."""
+    """Return the timestamp at which the given help channel, `channel`, should be closed."""
     log.trace(f"Getting the closing time for #{channel} ({channel.id}).")
 
-    if await _message.is_empty(channel):
+    if is_empty := await _message.is_empty(channel):
```
laundmo, line 32: Does this need a walrus operator? Why not just make a variable in front?

ChrisLovering: I doesn't *need* one, but is this not what the walrus operator is for?

ChrisLovering:
> > @MarkKoz as it stands currently, it will fall back to the old behavior if the cache isn't available (empty/ returns None)
> > I am proposing extending the fallback conditions to always use the old method for a while after startup, as it does not rely on cache. This way, instead of channels being closed prematurely they might take a bit longer to close.
> 
> Yes, that could work. It'd have to reschedule it for the maximum time.

What about clearing the caches on init, so the bot uses the fallback on startup, until the caches have had time to populate?

bot/exts/help_channels/_channel.py
laundmo: it may be that i just generally don't like the walrus operator. i have never seen it make the code more readable.

laundmo:
i have committed some changes that implement the cache clearing on bot startup to force fallback behaviour, and to write stats specifically for different reasons the channel could be closed. 

https://github.com/laundmo/bot/tree/help-channel-closing-delay-changes

Please merge them if they are okay, i did not want to make a new PR for this.

MarkKoz:
> 
> 
> > > @MarkKoz as it stands currently, it will fall back to the old behavior if the cache isn't available (empty/ returns None)
> > > I am proposing extending the fallback conditions to always use the old method for a while after startup, as it does not rely on cache. This way, instead of channels being closed prematurely they might take a bit longer to close.
> > 
> > 
> > Yes, that could work. It'd have to reschedule it for the maximum time.
> 
> What about clearing the caches on init, so the bot uses the fallback on startup, until the caches have had time to populate?
> Another open is populating the cache's on init too.

I'm not a fan of relying on cache behaviour like this. It feels too fragile and implicit.

laundmo:
> I'm not a fan of relying on cache behaviour like this. It feels too fragile and implicit.

Personally i think that clearing the caches is fine, as that solves exactly the issues we would have, which is outdated caches. The code has to handle non-existent caches anyway, and treating them as non-existent after a restart seems exactly the behaviour we want.

i understand that it doesn't feel clean since we have to manually clear them. If we were to handle that data as a python only variable, there wouldn't be a discussion about this at all.

Otherwise, we would have to complicate this more by somehow updating or otherwise handling outdated caches, which i personally like less.

MarkKoz:
> 
> 
> > I'm not a fan of relying on cache behaviour like this. It feels too fragile and implicit.
> 
> Personally i think that clearing the caches is fine, as that solves exactly the issues we would have, which is outdated caches. The code has to handle non-existent caches anyway, and treating them as non-existent after a restart seems exactly the behaviour we want.
> 
> i understand that it doesn't feel clean since we have to manually clear them. If we were to handle that data as a python only variable, there wouldn't be a discussion about this at all.
> 
> Otherwise, we would have to complicate this more by somehow updating or otherwise handling outdated caches, which i personally like less.

It's not complicated. Check if the cog is starting for the first time with `self.init_task.done()`. If it's true, then schedule channels to close 30 minutes (`idle_minutes_claimant`, the largest of all possible values) after the latest sent message. No need to worry about outdated caches because if a new message gets sent, the caches will get updated. If no messages get sent, then the outdated caches values will be old enough anyway to cause the channel to close. What are everyone's thoughts on that?

In any case, the current behaviour is still incorrect because it is using `idle_minutes_others` instead of `idle_minutes_claimant` when the cache is empty. It's not going to extensively search the message history for the claimant's last message. Therefore, it must assume the worst case and use `idle_minutes_claimant` to avoid prematurely closing a channel.

bot/exts/help_channels/_message.py
```diff
@@ -42,25 +42,28 @@
...
-                # The mapping for this channel doesn't exist, we can't do anything.
-                return
-
-            # Check the message did not come from the claimant
-            if claimant_id != message.author.id:
-                # Mark the channel as answered
-                await _caches.unanswered.set(channel.id, False)
+        log.trace(f"Checking if #{channel} ({channel.id}) has had a reply.")
+        # Must use a timezone-aware datetime to ensure a correct POSIX timestamp.
+        timestamp = datetime.now(timezone.utc).timestamp()
```
MarkKoz, line 53: Need to be careful mixing aware and naïve datetimes. They all need to be one or the other. I don't remember if it was for help channels that I chose to use aware datetimes somewhere. In any case, I don't believe it's strictly necessary as long as all the code is "on the same page" and using all naïve or all aware objects. Since it's dealing with times coming from discord.py, which gives naïve objects, the path of least resistance is likely to use naïve objects everywhere.

If you swich to naïve datetimes then you should just use the time from the `message` object over `datetime.now`.

bot/exts/help_channels/_message.py
```diff
@@ -42,25 +42,28 @@
...
+
+        # Overwrite the claimant message time, if its from the claimant.
+        if message.author == await _caches.claimants.get(channel.id):
+            await _caches.claimant_last_message_times.set(channel.id, timestamp)
+            return
+
+        claimant_id = await _caches.claimants.get(channel.id)
```
MarkKoz, line 60: The claimant ID is being fetched twice. That can be avoided by saving the first value.

bot/exts/help_channels/_cog.py
```diff
@@ -410,15 +418,20 @@ async def move_to_in_use(self, channel: discord.TextChannel) -> None:
...
         _stats.report_counts()
 
     @commands.Cog.listener()
     async def on_message(self, message: discord.Message) -> None:
-        """Move an available channel to the In Use category and replace it with a dormant one."""
+        """
+        Move an available channel to the In Use category and replace it with a dormant one.
+
+        Update the `last_message_times` cache based on the current timestamp. If the message
+        author is the claimant of this channel, also update the `claimant_last_message_times` cache.
```
MarkKoz, line 433: Do you think this should be removed now that the code for this has been moved elsewhere? `on_message` still technically causes this to happen, but it's no longer directly responsible for it.

bot/exts/help_channels/_cog.py
```diff
@@ -410,15 +418,20 @@ async def move_to_in_use(self, channel: discord.TextChannel) -> None:
             category_id=constants.Categories.help_in_use,
         )
 
-        timeout = constants.HelpChannels.idle_minutes * 60
+        timeout = constants.HelpChannels.idle_minutes_others * 60
```
MarkKoz, line 421: Shouldn't it use `idle_minutes_claimant` since the message responsible for this was sent by the claimant?

bot/exts/help_channels/_cog.py
```diff
@@ -375,11 +377,17 @@ async def unclaim_channel(self, channel: discord.TextChannel, *, is_auto: bool =
...
+    async def _delete_message_time_caches(self, channel: discord.TextChannel) -> None:
+        """Delete message time caches."""
+        await _caches.claimant_last_message_times.delete(channel.id)
+        await _caches.non_claimant_last_message_times.delete(channel.id)
 
-    async def _unclaim_channel(self, channel: discord.TextChannel, claimant_id: int, is_auto: bool) -> None:
+    async def _unclaim_channel(self, channel: discord.TextChannel, claimant_id: int, closed_on: str) -> None:
         """Actual implementation of `unclaim_channel`. See that for full documentation."""
         await _caches.claimants.delete(channel.id)
+        await self._delete_message_time_caches(channel)
```
MarkKoz, line 390: Is the deletion still necessary? I hold the opinion that the fewer operations, the better. So if it can get away with leaving junk values in cache then it should just leave them. Or does that sound like a bad idea?

bot/exts/help_channels/_cog.py
```diff
@@ -271,6 +275,8 @@ async def init_cog(self) -> None:
 
         log.trace("Moving or rescheduling in-use channels.")
         for channel in _channel.get_category_channels(self.in_use_category):
+            # clear the cache here so moving doesn't rely on old cached messages.
+            await self._delete_message_time_caches(channel)
```
MarkKoz, line 279: Again, I don't think this is necessary.

bot/exts/help_channels/_channel.py
```diff
@@ -25,23 +25,58 @@ def get_category_channels(category: discord.CategoryChannel) -> t.Iterable[disco
...
+        idle_minutes = constants.HelpChannels.idle_minutes_claimant
+
+    non_claimant_last_message_time = await _caches.non_claimant_last_message_times.get(channel.id)
+    claimant_last_message_time = await _caches.claimant_last_message_times.get(channel.id)
+
+    if is_empty or not all(
+        init_done,
+        non_claimant_last_message_time,
+        claimant_last_message_time,
+    ):
```
MarkKoz, line 46: Is this logic right? It should do it if either the cog is still initialising or the caches are empty.

MarkKoz (changes requested):

bot/exts/help_channels/_stats.py
```diff
@@ -22,21 +22,21 @@ def report_counts() -> None:
...
 
-async def report_complete_session(channel_id: int, is_auto: bool) -> None:
+async def report_complete_session(channel_id: int, closed_on: str) -> None:
     """
     Report stats for a completed help session channel `channel_id`.
 
     Set `is_auto` to True if the channel was automatically closed or False if manually closed.
     """
-    caller = "auto" if is_auto else "command"
+    caller = f"auto.{closed_on}" if closed_on else "command"
```
laundmo, line 31: this makes sure that auto is not prepended if the channel is closed by commend. 

alternatively, get rid of this prepending entirely and explicitly pass "auto.cleanup" etc.

```suggestion
    caller = f"auto.{closed_on}" if closed_on != "command" else "command"
```

bot/exts/help_channels/_cog.py
```diff
@@ -187,7 +191,7 @@ async def close_command(self, ctx: commands.Context) -> None:
         # Don't use a discord.py check because the check needs to fail silently.
         if await self.close_check(ctx):
             log.info(f"Close command invoked by {ctx.author} in #{ctx.channel}.")
-            await self.unclaim_channel(ctx.channel, is_auto=False)
+            await self.unclaim_channel(ctx.channel, closed_on="command")
```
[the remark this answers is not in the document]
laundmo: see https://github.com/python-discord/bot/pull/1470/files/2ac3e1f9f24033c77145fa143350814e15542fd2#r601151967 for my suggested fix

bot/exts/help_channels/_message.py
laundmo: it seems this would also check the author object against the id which even if discord.py handles that (not sure) is not as explicit as checking the id.
```suggestion
        claimant_id = await _caches.claimants.get(channel.id)

        if message.author.id == claimant_id:
            await _caches.claimant_last_message_times.set(channel.id, timestamp)
            return
```

bot/exts/help_channels/_cog.py
laundmo: yep, definitely should be claimant.
```suggestion
        timeout = constants.HelpChannels.idle_minutes_claimant * 60
```

bot/exts/help_channels/_cog.py
```diff
@@ -393,12 +401,12 @@ async def _unclaim_channel(self, channel: discord.TextChannel, claimant_id: int,
...
 
         await _message.unpin(channel)
-        await _stats.report_complete_session(channel.id, is_auto)
+        await _stats.report_complete_session(channel.id, closed_on)
         await self.move_to_dormant(channel)
 
         # Cancel the task that makes the channel dormant only if called by the close command.
         # In other cases, the task is either already done or not-existent.
-        if not is_auto:
+        if not closed_on:
```
laundmo, line 409: this needs to explicitly check for the closed_on reason being `command`
initially i had considered passing False or None for command closings, but since command now passes the string command explicitly, this check will never be true.

```suggestion
        if closed_on == "command":
```

bot/exts/help_channels/_channel.py
ChrisLovering: If any of the variables within the `all()` call are falsely, we want to enter the if block and use the fallback behaviour. IE if the cog is initialising or if either cache is empty

bot/exts/help_channels/_cog.py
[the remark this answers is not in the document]
ChrisLovering: Resolved with https://github.com/python-discord/bot/pull/1470/commits/65c009736507974ebce20a1f36b7e24a4ecc349b

bot/exts/help_channels/_message.py
ChrisLovering: Resolved with https://github.com/python-discord/bot/commit/9f576bcf28aa228b6c7f7d57072bba3135047200

bot/exts/help_channels/_channel.py
```diff
@@ -85,8 +85,8 @@ async def get_in_use_time(channel_id: int) -> t.Optional[timedelta]:
 
     claimed_timestamp = await _caches.claim_times.get(channel_id)
     if claimed_timestamp:
-        claimed = datetime.utcfromtimestamp(claimed_timestamp)
-        return datetime.utcnow() - claimed
+        claimed = datetime.fromtimestamp(claimed_timestamp)
```
MarkKoz, line 88: discord.py still returns times in UTC; they're just naïve. It uses `utcfromtimestamp`. Sorry if my previous comment was misleading.

MarkKoz (changes requested):

bot/exts/help_channels/_channel.py
MarkKoz: Right... `¬(A & B & C) == (¬A | ¬B | ¬C)`

bot/exts/help_channels/_stats.py
```diff
@@ -22,21 +22,22 @@ def report_counts() -> None:
...
         bot.instance.stats.timing("help.in_use_time", in_use_time)
 
-    unanswered = await _caches.unanswered.get(channel_id)
-    if unanswered:
+    non_claimant_last_message_time = await _caches.non_claimant_last_message_times.get(channel_id)
+    if non_claimant_last_message_time is None:
         bot.instance.stats.incr("help.sessions.unanswered")
-    elif unanswered is not None:
+    else:
         bot.instance.stats.incr("help.sessions.answered")
```
MarkKoz, line 43: This is now broken due to
```py
        # non_claimant needs to be set too, to satisfy the condition in `_channel.get_closing_time` the first time.
        # Otherwise it will fall back to the old method if no other messages are sent.
        await _caches.non_claimant_last_message_times.set(message.channel.id, message.created_at)
```
The value should be None when the channel is claimed, but then you'd need to fix the fallback logic.

MarkKoz (changes requested):

bot/exts/help_channels/_cog.py
```diff
@@ -112,11 +114,11 @@ async def claim_channel(self, message: discord.Message) -> None:
 
         self.bot.stats.incr("help.claimed")
 
-        # Must use a timezone-aware datetime to ensure a correct POSIX timestamp.
-        timestamp = datetime.now(timezone.utc).timestamp()
-        await _caches.claim_times.set(message.channel.id, timestamp)
-
-        await _caches.unanswered.set(message.channel.id, True)
+        await _caches.claim_times.set(message.channel.id, message.created_at)
```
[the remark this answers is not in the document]
ChrisLovering: wdym, we still use this cache right?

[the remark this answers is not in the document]
MarkKoz: Oh sorry, I didn't realise the diff was collapsed.

bot/exts/help_channels/_stats.py
ChrisLovering: Good point, I'll look into this 1

MarkKoz:
I'm trying this out with the following config:

```yml
    idle_minutes_claimant: 2
    idle_minutes_others: 1
    deleted_idle_minutes: 1
```

I simply claim a channel, and it correctly schedules it to close in 2 minutes. However, when the closing task runs, it calculates a wrong time. The log timestamps on the left are in UTC-8. Converted to UTC the times would be around `2021-03-26 20:58:06`. Therefore, the time it calculates is off by 7 hours.

```
2021-03-26 13:58:06 | bot.exts.help_channels._channel | TRACE | #help-burrito (800698210392997918) should be closed at 2021-03-27 03:58:05.915000 due to claimant timeout.
...
2021-03-26 13:58:06 | bot.exts.help_channels._cog | INFO | #help-burrito (800698210392997918) is still active; scheduling it to be moved after 25199 seconds.
...
2021-03-26 13:58:06 | bot.utils.scheduling.HelpChannels | TRACE | Waiting 25199 seconds before awaiting coroutine for #800698210392997918.
```

bot/exts/help_channels/_stats.py
```diff
@@ -22,21 +22,22 @@ def report_counts() -> None:
             log.warning(f"Couldn't find category {name!r} to track channel count stats.")
 
 
-async def report_complete_session(channel_id: int, is_auto: bool) -> None:
+async def report_complete_session(channel_id: int, closed_on: str) -> None:
     """
     Report stats for a completed help session channel `channel_id`.
 
     Set `is_auto` to True if the channel was automatically closed or False if manually closed.
```
jb3, line 29: This docstring needs updating with `closed_on` and the potential values.

bot/exts/help_channels/_cog.py
```diff
@@ -112,11 +115,13 @@ async def claim_channel(self, message: discord.Message) -> None:
 
         self.bot.stats.incr("help.claimed")
 
-        # Must use a timezone-aware datetime to ensure a correct POSIX timestamp.
-        timestamp = datetime.now(timezone.utc).timestamp()
-        await _caches.claim_times.set(message.channel.id, timestamp)
+        # datetime.timestamp() would assume it's local, despite d.py giving a (naïve) UTC time.
```
jb3, line 118: We should probably just use `naive` here :smile:, which is the more usual spelling nowadays.
```suggestion
        # datetime.timestamp() would assume it's local, despite d.py giving a (naive) UTC time.
```

bot/exts/help_channels/_message.py
```diff
@@ -42,25 +43,27 @@
...
+        if not claimant_id:
+            # The mapping for this channel doesn't exist, we can't do anything.
+            return
 
-            # Check the message did not come from the claimant
-            if claimant_id != message.author.id:
-                # Mark the channel as answered
-                await _caches.unanswered.set(channel.id, False)
+        # datetime.timestamp() would assume it's local, despite d.py giving a (naïve) UTC time.
```
jb3, line 59:
```suggestion
        # datetime.timestamp() would assume it's local, despite d.py giving a (naive) UTC time.
```

jb3 (changes requested):
Code looks good and functionality works, just a few documentation things.

bot/exts/help_channels/_cog.py
MarkKoz: Leave my diacritics out of this. Jokes aside, this spelling is much more common than you might think.

laundmo: personally i would prefer the normal i as i have no idea how to properly type the one with 2 dots.

jb3 (approved):
*Actual* approval now, :shipit:.

laundmo (approved):

jb3 (approved):

PR #1471, opened 2021-03-16
Remove reactions from everyone when paginating and waiting for trashcan reaction.

### LINKED ISSUE #1454: Remove reactions from users that aren't the OP for pagination and trash can.
### Goal
When using pagination or the trash can symbol (the one used to delete embeds), it is set to only be able to use by the OP. It may help clean it up to delete reactions from people that aren't the OP.

### Implementation Suggestion
Python already checks whether the person adding the reactions for pagination or the trash can is used by the OP, so it would be simple to use `remove_reaction` to get rid of the useless reactions.

ToxicKidz:
Closes #1454

This will remove reactions when someone adds it, even if they aren't whitelisted to do so.


bot/pagination.py
```diff
@@ -229,20 +229,36 @@ def event_check(reaction_: discord.Reaction, user_: discord.Member) -> bool:
...
                     # There were no restrictions
                     no_restrictions
                 ))
             )
 
+            if whitelisted:
+                return True
+
+            elif right_reaction and user_.id != ctx.bot.user.id:
+                ctx.bot.loop.create_task(
```
kosayoda, line 255: This method call easily fits on one line.

bot/pagination.py
```diff
@@ -229,20 +229,36 @@ def event_check(reaction_: discord.Reaction, user_: discord.Member) -> bool:
...
+                    reaction_.message.remove_reaction(
+                        reaction_.emoji, user_
+                    )
+                )
+                log.debug(f"Got reaction: {reaction_.emoji} from non-whitelisted user")
+
```
kosayoda, line 261: In the event the reaction is not `whitelisted`, the check should return `False`. Right now it's implicitly returning `None`.

bot/pagination.py
```diff
@@ -229,20 +229,36 @@ def event_check(reaction_: discord.Reaction, user_: discord.Member) -> bool:
...
+            if whitelisted:
+                return True
+
+            elif right_reaction and user_.id != ctx.bot.user.id:
+                ctx.bot.loop.create_task(
+                    reaction_.message.remove_reaction(
+                        reaction_.emoji, user_
+                    )
+                )
+                log.debug(f"Got reaction: {reaction_.emoji} from non-whitelisted user")
```
kosayoda, line 260:
```suggestion
                log.debug(f"Got reaction: {reaction_.emoji} from non-whitelisted user, reaction deleted")
```
Some additional clarification would be nice.

kosayoda (changes requested):
Thanks for the PR! For future pull requests, you can run `pipenv run precommit` in the project's root directory to prevent yourself from being able to commit if the lint fails.

kosayoda:
When you commit, there should be something like this:
image

You shouldn't have to run `precommit` every single time, just once when you setup the project.

ToxicKidz:
> When you commit, there should be something like this:
> image
> 
> You shouldn't have to run `precommit` every single time, just once when you setup the project.

Oh so I only have to `pipenv run pre-commit` once?

ToxicKidz:
> I'm thinking we should squash this when we merge due to the messy history at the beginning.

Like a new PR?

ToxicKidz:
Never mind I read it incorrectly

bot/pagination.py
```diff
@@ -229,20 +229,35 @@ def event_check(reaction_: discord.Reaction, user_: discord.Member) -> bool:
...
                     no_restrictions
                 ))
             )
 
+            if whitelisted:
+                return True
+
+            elif right_reaction and user_.id != ctx.bot.user.id:
+
+                ctx.bot.loop.create_task(reaction_.message.remove_reaction(reaction_.emoji, user_))
```
MarkKoz, line 256: Use `asyncio.create_task` instead. In fact, it would be even better to use [`scheduling.create_task`](https://github.com/python-discord/bot/blob/main/bot/utils/scheduling.py#L160) since it logs exceptions. I believe that even with `asyncio.create_task`, any exception raised would eventually show up anyway as uncaught (typically when the bot shuts down), so it's worth handling it. In this case, we'd want to suppress the exception raised when the message no longer exists.

MarkKoz (changes requested):

bot/pagination.py
ToxicKidz: So pretty much I should change to `scheduling.create_task` and suppress `discord.HttpException` (or just `NotFound`)?

MarkKoz:
`create_task` doesn't raise the exception. It has to be suppressed within the coroutine being scheduled.

ToxicKidz:
> `create_task` doesn't raise the exception. It has to be suppressed within the coroutine being scheduled.

So don't suppress it at all?

MarkKoz:
> 
> 
> > `create_task` doesn't raise the exception. It has to be suppressed within the coroutine being scheduled.
> 
> So don't suppress it at all?

No. I explicitly wrote it has to be suppressed. You're just suppressing it at the wrong time/place.

ToxicKidz:
`create_task` isn't where the coroutine is being scheduled?

MarkKoz:
A correct example:

```py
async def coroutine():
    with suppress(SomeException):
        return await do_thing()

create_task(coroutine)
```

ToxicKidz:
Oh, so I need to create a function that suppresses `HTTPException`?

MarkKoz:
Yes... unless you can think of a nicer way.

ToxicKidz:
> Yes... unless you can think of a nicer way.

Would a new utility function work since I'm using it multiple times?

ToxicKidz:
But I don't see why suppress the exception if `scheduling.create_task` logs the exception and ultimately suppresses it.

MarkKoz:
> 
> 
> > Yes... unless you can think of a nicer way.
> 
> Would a new utility function work since I'm using it multiple times?

Yeah, sure. See how much you can turn into a utility function. Maybe some logic in the check could be moved to the utility function too.

> 
> 
> But I don't see why suppress the exception if `scheduling.create_task` logs the exception and ultimately suppresses it.

It suppresses the raising of the exception, but it does not suppress the logging of it. It indiscreminantly logs all exceptions. We don't need to see a log for when the message is missing because it's not actually an error that we need to fix.

bot/utils/scheduling.py
```diff
@@ -157,18 +157,20 @@ def _task_done_callback(self, task_id: t.Hashable, done_task: asyncio.Task) -> N
...
 
-def _log_task_exception(task: asyncio.Task) -> None:
+def _log_task_exception(task: asyncio.Task, suppress_exceptions: t.Tuple[Exception] = ()) -> None:
     """Retrieve and log the exception raised in `task` if one exists."""
     with contextlib.suppress(asyncio.CancelledError):
         exception = task.exception()
         # Log the exception if one exists.
         if exception:
+            if isinstance(exception, suppress_exceptions):
+                return
```
MarkKoz, line 174:
```suggestion
        if exception and not isinstance(exception, suppress_exceptions):
```

bot/utils/scheduling.py
```diff
@@ -157,18 +157,20 @@ def _task_done_callback(self, task_id: t.Hashable, done_task: asyncio.Task) -> N
...
+def create_task(suppress_exceptions: t.Tuple[Exception], *args, **kwargs) -> asyncio.Task:
     """Wrapper for `asyncio.create_task` which logs exceptions raised in the task."""
     task = asyncio.create_task(*args, **kwargs)
-    task.add_done_callback(_log_task_exception)
+    task.add_done_callback(partial(_log_task_exception, suppress_exceptions=suppress_exceptions))
     return task
 
 
-def _log_task_exception(task: asyncio.Task) -> None:
+def _log_task_exception(task: asyncio.Task, suppress_exceptions: t.Tuple[Exception] = ()) -> None:
```
MarkKoz, line 167: `suppressed_exceptions` would be a better name since the current name sounds more like a Boolean.

bot/utils/scheduling.py
```diff
@@ -157,18 +157,20 @@ def _task_done_callback(self, task_id: t.Hashable, done_task: asyncio.Task) -> N
                 self._log.error(f"Error in task #{task_id} {id(done_task)}!", exc_info=exception)
 
 
-def create_task(*args, **kwargs) -> asyncio.Task:
+def create_task(suppress_exceptions: t.Tuple[Exception], *args, **kwargs) -> asyncio.Task:
```
MarkKoz, line 160: Making this the first argument and non-optional breaks all other uses of this function. The coroutine should be the first argument and the function should not require callers to always specify exceptions. Consider using `*args` for the exceptions since it'd be more convenient for callers.

MarkKoz (changes requested):

bot/utils/scheduling.py
ToxicKidz: > Making this the first argument and non-optional breaks all other uses of this function. The coroutine should be the first argument and the function should not require callers to always specify exceptions. Consider using `*args` for the exceptions since it'd be more convenient for callers.

But it's already using `*args`. How about making it a kwarg?

MarkKoz: It's not actually using it. It was just done this way to make it future-proof against changes to `asyncio.create_task`'s signature, but this is unlikely to happen anyway.

bot/pagination.py
```diff
@@ -229,20 +230,37 @@ def event_check(reaction_: discord.Reaction, user_: discord.Member) -> bool:
...
 
+            if whitelisted:
+                return True
+
+            elif right_reaction and user_.id != ctx.bot.user.id:
+
+                scheduling.create_task(
+                    (discord.HTTPException,),  # Suppress the HTTPException if adding the reaction fails
+                    reaction.message.remove_reaction(reaction.emoji, user)
+                )
```
MarkKoz, line 260: You forgot to change these calls to use the new signature with `*args`.

bot/utils/scheduling.py
```diff
@@ -157,18 +157,18 @@ def _task_done_callback(self, task_id: t.Hashable, done_task: asyncio.Task) -> N
                 self._log.error(f"Error in task #{task_id} {id(done_task)}!", exc_info=exception)
 
 
-def create_task(*args, **kwargs) -> asyncio.Task:
+def create_task(coro: t.Awaitable, *suppressed_exceptions: Exception, **kwargs) -> asyncio.Task:
     """Wrapper for `asyncio.create_task` which logs exceptions raised in the task."""
-    task = asyncio.create_task(*args, **kwargs)
-    task.add_done_callback(_log_task_exception)
+    task = asyncio.create_task(coro, **kwargs)
+    task.add_done_callback(partial(_log_task_exception, *suppressed_exceptions))
```
MarkKoz, line 163: Does this actually work? I'm thinking it might put an exception in place of the `task` argument.

MarkKoz (changes requested):

bot/utils/scheduling.py
ToxicKidz: In what case would that happen?

ToxicKidz: Oh I see it now

MarkKoz: If it's a problem then I _think_ the fix would be to pass it as a tuple through a kwarg rather than unpacking it.

ToxicKidz: I'll just make it take a `tuple` since it's only going to be used once.

bot/pagination.py
ToxicKidz: Whoops, I should have tested it more.

MarkKoz (changes requested):
I get an error upon reacting with a different user

```
  File "/home/mark/repos/python/bot-pydis/.venv/lib/python3.8/site-packages/discord/ext/commands/core.py", line 85, in wrapped
  File "/home/mark/repos/python/bot-pydis/bot/decorators.py", line 78, in inner
    await func(self, ctx, *args, **kwargs)
  File "/home/mark/repos/python/bot-pydis/bot/exts/info/help.py", line 68, in command_callback
    await self.send_bot_help(mapping)
  File "/home/mark/repos/python/bot-pydis/bot/exts/info/help.py", line 344, in send_bot_help
    await LinePaginator.paginate(pages, self.context, embed=embed, max_lines=1, max_size=2000)
  File "/home/mark/repos/python/bot-pydis/bot/pagination.py", line 326, in paginate
    reaction, user = await ctx.bot.wait_for("reaction_add", timeout=timeout, check=event_check)
  File "/opt/pyenv/versions/3.8.6/lib/python3.8/asyncio/tasks.py", line 491, in wait_for
    return fut.result()
  File "/home/mark/repos/python/bot-pydis/.venv/lib/python3.8/site-packages/discord/client.py", line 370, in dispatch
    result = condition(*args)
  File "/home/mark/repos/python/bot-pydis/bot/pagination.py", line 258, in event_check
    reaction.message.remove_reaction(reaction.emoji, user),
NameError: free variable 'reaction' referenced before assignment in enclosing scope
```

ToxicKidz:
Right, I forgot to append the `_` after `reaction, user`, but I never saw that before.

MarkKoz:
Still getting the same error. Can you please test this yourself before pushing the fix?

ToxicKidz:
> Still getting the same error. Can you please test this yourself before pushing the fix?

I've tested multiple times and haven't gotten that error, I did get a different one though.

MarkKoz:
Definitely still getting the same error... I think `reaction.message.remove_reaction(reaction_.emoji, user_)` needs to be `reaction_.message.remove_reaction(reaction_.emoji, user_)`

ToxicKidz:
> Definitely still getting the same error... I think `reaction.message.remove_reaction(reaction_.emoji, user_)` needs to be `reaction_.message.remove_reaction(reaction_.emoji, user_)`

Yeah that makes sense. I'm just wondering why I didn't get any errors.

ToxicKidz:
Ah it's because I reacted to it first so reaction got defined.

MarkKoz:
Okay, that's fixed, but now there is a different bug. I think the bot is detecting its own reaction as a deletion. It happens fast, so it's hard to tell what's going on. Enter `!codeblock` and observe that the message immediately gets deleted.

ToxicKidz:
Fixed.

MarkKoz (approved):
It was working fine, but I took the opportunity to refactor it into a separate function.

I am a bit concerned about reactions being spammed, but I tried it and the bot seemed to be able to keep up.

bot/utils/scheduling.py
```diff
@@ -157,18 +157,18 @@ def _task_done_callback(self, task_id: t.Hashable, done_task: asyncio.Task) -> N
                 self._log.error(f"Error in task #{task_id} {id(done_task)}!", exc_info=exception)
 
 
-def create_task(*args, **kwargs) -> asyncio.Task:
+def create_task(coro: t.Awaitable, *suppressed_exceptions: t.Type[Exception], **kwargs) -> asyncio.Task:
```
Numerlor, line 160: Passing in the exceptions as positional args doesn't seem clear to me from the caller's side. 
A tuple kwarg would explain the uses of the exception types properly without a comment like the one in the messages module where this functionality is currently used

Numerlor (changes requested):

vcokltfre (approved):

kosayoda (approved):

PR #1477, opened 2021-03-20
Added nomination voting channel to config

wookie184:
Also changed talentpool review cog to post there instead of mod-announcements


codephile1221 (approved):
Looks good to me!

mbaruh (approved):

PR #1479, opened 2021-03-21
Limit nominate command channels

wookie184:
This limits the `!nominate` command to the nominations channel, adds a force nominate command so mods can override that, and also adds the `moderators` category to the config as a moderation category, so `nomination-discussion` and any new channels added in the future can have moderation commands run in them.


config-default.yml
```diff
@@ -193,12 +194,10 @@ guild:
         helpers:            &HELPERS        385474242440986624
         incidents:                          714214212200562749
         incidents_archive:                  720668923636351037
-        mods:               &MODS           305126844661760000
+        mods:                               305126844661760000
         mod_alerts:                         473092532147060736
-        mod_appeals:        &MOD_APPEALS    808790025688711198
```
Akarys42, line 198: Since those channels have been from here, they also need to be removed from `bot/constants.py`.

Akarys42 (changes requested):
Just one small thing!

wookie184:
I don't think the ones I removed were in the constants.py (unless I'm missing something)

MarkKoz:
Why make a separate command instead of allowing the original one to work everywhere for moderators?

wookie184:
> 
> 
> Why make a separate command instead of allowing the original one to work everywhere for moderators?

To discourage moderators from running the nominate command in channels other than the `nominations` channel, for example if they clicked on the wrong channel or didn't realise they should only use the `nominations` channel, while still allowing them to bypass that if necessary. Granted theses cases would be rare, but I don't think there's much harm in adding a separate command.

Akarys42 (approved):

Akarys42 (approved):
you can do it review bot

Xithrius (approved):
Looks good to my eyes

Akarys42 (approved):
I don't like you review bot

PR #1480, opened 2021-03-22
Update eval to use SIGTERM for killed processes

Akarys42:
Due to a recent change in snekbox (https://github.com/python-discord/snekbox/pull/96) processes killed by cgroup limits will use SIGTERM instead of SIGKILL. This commit account for this change by updating the numerical value to 15 (SIGTERM portable value).

### DIFF
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -43,7 +43,7 @@ EVAL_CHANNELS = (Channels.bot_commands, Channels.esoteric)
 EVAL_CATEGORIES = (Categories.help_available, Categories.help_in_use, Categories.voice)
 EVAL_ROLES = (Roles.helpers, Roles.moderators, Roles.admins, Roles.owners, Roles.python_community, Roles.partners)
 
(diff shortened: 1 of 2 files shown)


MarkKoz:
This is incorrect. SIGTERM is only used to stop the process when stdout is too large. Nsjail still uses SIGKILL for time, memory, and seccomp violations.

Akarys42:
> This is incorrect. SIGTERM is only used to stop the process when stdout is too large. Nsjail still uses SIGKILL for time, memory, and seccomp violations.

Alright, I misunderstood the change then my bad. What about using SIGKILL and SIGTERM to trigger this message then?

MarkKoz:
If anything it should show an entirely separate message. SIGTERM will never have anything to do with the current message so it'd be inaccurate to use it.

Akarys42:
> If anything it should show an entirely separate message. SIGTERM will never have anything to do with the current message so it'd be inaccurate to use it.

Alright, let’s do that in another PR then.

PR #1486, opened 2021-03-28
!dmrelay command

Den4200:
## Description
This PRs wipes out the old `DMRelay` cog and its functionalities. We realize that relaying all DMs sent to the bot into a specific channel may come in conflict with Discord's ToS, so that feature has now been removed, along with the `!reply` command. I've replaced it with the `!dmrelay` command, allowing moderators to view DM history between the bot and a specific user when necessary.

Here is an [example message log](https://paste.pythondiscord.com/ogeromerur.txt?noredirect).

For now, the message history is uploaded to our [paste site](https://paste.pydis.com). In the future, when [Logcord](https://github.com/python-discord/logcord) is up and ready, we will transfer over to use that.


bot/exts/moderation/dm_relay.py
```diff
@@ -1,132 +1,63 @@
...
 
 log = logging.getLogger(__name__)
 
 
 class DMRelay(Cog):
-    """Relay direct messages to and from the bot."""
-
-    # RedisCache[str, t.Union[discord.User.id, discord.Member.id]]
-    dm_cache = RedisCache()
+    """Relay direct messages from the bot."""
```
jb3, line 15:
```suggestion
    """Inspect messages sent to the bot."""
```

bot/utils/services.py
```diff
@@ -47,7 +47,13 @@ async def send_to_paste_service(contents: str, *, extension: str = "") -> Option
             continue
         elif "key" in response_json:
             log.info(f"Successfully uploaded contents to paste service behind key {response_json['key']}.")
-            return URLs.paste_service.format(key=response_json['key']) + extension
+
+            paste_link = URLs.paste_service.format(key=response_json['key']) + extension
+
+            if extension == '.py':
```
jb3, line 53: Shouldn't extension not have a `.` here?
```suggestion
            if extension == 'py':
```

bot/utils/services.py
```diff
@@ -47,7 +47,13 @@ async def send_to_paste_service(contents: str, *, extension: str = "") -> Option
...
+                return paste_link
+            return paste_link + "?noredirect"
```
jb3, line 55:
```suggestion

            return paste_link + "?noredirect"
```

jb3 (changes requested):
A few minor things noted in the code, as well as the below bugs.

## Does not handle client user

Screenshot 2021-03-28 at 12 40 57

We should gracefully reject requests to relay the bot's DM history with itself.

## Incorrect "no DM history"

After booting the bot and running this without sending any messages, the bot incorrectly identified me as having no history, after I sent a message to the (existing) channel it *did* identify that there was a history and uploaded it.

Maybe we should instead just use an async iterator to ensure we get a channel and if we yield 0 messages then we report back that there was no history, I presume this is a cache thing.

Screenshot 2021-03-28 at 12 41 30

bot/utils/services.py
Den4200: A `.` is tacked onto the extension, if one is given beforehand.

https://github.com/python-discord/bot/blob/4be90b3c454138e3548c7394fcb2a1182b05b7d7/bot/utils/services.py#L22

jb3: Aha, gotcha.

jb3 (approved):
This looks good, however it now uses 2 API calls instead of one, which I think we can optimise down to one.

Switching around the order of "no channel" with "fetch messages" means you can do something along the lines of:
```py
found_messages = False

async for msg in user.history(...):
  found_messages = True

if not found_messages:
  await ctx.send("... no direct message hitsory ...")
```

I'm fine with this implementation though, since it's an infrequently used command anyway.

jb3 (approved):
I like it. :shipit:

bot/utils/services.py
```diff
@@ -47,7 +47,14 @@ async def send_to_paste_service(contents: str, *, extension: str = "") -> Option
             continue
         elif "key" in response_json:
             log.info(f"Successfully uploaded contents to paste service behind key {response_json['key']}.")
-            return URLs.paste_service.format(key=response_json['key']) + extension
+
+            paste_link = URLs.paste_service.format(key=response_json['key']) + extension
+
+            if extension == '.py':
+                return paste_link
```
Numerlor, line 54: Does py need to be special cased here? the noredirect shouldn't change anything for it and the util isn't used anywhere where the compactness would matter

ChrisLovering (approved):
My only gripe is we're building the output string using `+=` rather than having a list and using `.join()` at the end. Not enough to stop the merge though! :D

bot/exts/moderation/dm_relay.py
```diff
@@ -1,132 +1,68 @@
...
-                log.exception("Failed to send an attachment to the webhook")
-
-    async def cog_check(self, ctx: commands.Context) -> bool:
+        metadata = textwrap.dedent(f"""\
+            User: {user} ({user.id})
+            Channel ID: {user.dm_channel.id}\n
+        """)
+
+        paste_link = await send_to_paste_service(metadata + output, extension="txt")
+        await ctx.send(paste_link)
```
Numerlor, line 59: The case where the paste service fails and returns None needs to be handled here

ks129 (approved):

bot/exts/moderation/dm_relay.py
```diff
@@ -1,132 +1,68 @@
...
-                    avatar_url=message.author.avatar_url
-                )
-            except discord.HTTPException:
-                log.exception("Failed to send an attachment to the webhook")
-
-    async def cog_check(self, ctx: commands.Context) -> bool:
+        metadata = textwrap.dedent(f"""\
+            User: {user} ({user.id})
+            Channel ID: {user.dm_channel.id}\n
+        """)
```
Numerlor, line 56:
```suggestion
        metadata = (
            f"User: {user} ({user.id})\n"
            f"Channel ID: {user.dm_channel.id}\n\n"
        )
```
Using implicit line continuation and joining looks a bit nicer to me; the textwrap import can also be removed if you decide to use this

bot/utils/services.py
Den4200: I guess not, but I still think it's nice to account for it.

Den4200:
I'll address these comments in a new PR.

PR #1488, opened 2021-03-28
Filter codeblock escapes and allow no mentions for !raw command

### LINKED ISSUE #1485: Filter bypass in !raw command, can make the bot @mention
The `!raw` command takes a link to a message in the discord, and then shows the message contents inside a code block.

This line tries to prevent the message from breaking the code block:
https://github.com/python-discord/bot/blob/76b3b5898dd4163d8bac6ff330209f27f437e3d6/bot/exts/info/information.py#L452

But consider the message: ` ````` ` (5 backticks). The first 3 will be replaced by the filter, but that's it. Finally giving this result: ` `` ``` `, and actually breaking out of the code block, since we got 3 backticks in a row.

POC with ` ````` <@...> ``` ` to have the bot mention someone:

image

` ````` @everyone ``` ` might be possible, but I didn't try 😋

I suggest changing the filter to ```.replace('`', '\\`')```, so backticks will just show up "escaped" to users (they will actually see the backslash).
Another fix is to remove the `!raw` command, idk why it exists.

ToxicKidz:
## Changes
1. Instead of adding a space between ``` (this was easily bypassed), replace each backtick with "`\u200b".
2. If someone (somehow) bypasses the codeblocks, they cannot mention everyone because the message will use [`AllowedMentions.none`](https://discordpy.readthedocs.io/en/latest/api.html#discord.AllowedMentions.none)

## Relevant Issues
Closes #1485

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 from typing import Any, DefaultDict, Dict, Mapping, Optional, Tuple, Union
 
 import fuzzywuzzy
-from discord import Colour, Embed, Guild, Message, Role
+from discord import AllowedMentions, Colour, Embed, Guild, Message, Role
 from discord.ext.commands import BucketType, Cog, Context, Paginator, command, group, has_any_role
 
 from bot import constants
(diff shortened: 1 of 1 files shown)


bot/exts/info/information.py
```diff
@@ -447,9 +447,9 @@ async def raw(self, ctx: Context, *, message: Message, json: bool = False) -> No
 
         def add_content(title: str, content: str) -> None:
             paginator.add_line(f'== {title} ==\n')
-            # replace backticks as it breaks out of code blocks. Spaces seemed to be the most reasonable solution.
-            # we hope it's not close to 2000
-            paginator.add_line(content.replace('```', '`` `'))
+            # Replace backticks as it breaks out of code blocks.
+            # An invisble character seemed to be the most reasonable solution. We hope it's not close to 2000.
```
jb3, line 451:
```suggestion
            # An invisible character seemed to be the most reasonable solution. We hope it's not close to 2000.
```

jb3 (changes requested):
Looking good! Minor typo but otherwise fine.

jb3 (approved):
:shipit:

mbaruh (approved):

PR #1492, opened 2021-03-30
Use deleted reason if help channel is closed due to being empty

ChrisLovering:
Previously `get_closing_time()` would always return `ClosingReason.LATEST_MESSSAGE` regardless for if it was using the fallback behaviour due to empty caches, or if the channel was empty.

This PR fixes that incorrect behaviour, and correctly returns `ClosingReason.DELETED` if the help channel was closed due to being empty.


Shivansh-007 (approved):
:shipit:

MarkKoz (approved):

Akarys42 (approved):
:shipit: :shipit: :shipit:

PR #1493, opened 2021-03-31
Stream and revokestream commands

### LINKED ISSUE #1343: Command to let mods add video role for a user to share their screen for voice chat help
Streaming and video permissions are currently disabled for non-staff or contributors.  However, at times it's beneficial for users to share their screen to show the errors that they're getting.  This currently requires an admin to apply the Video role to the person.  Admins aren't always around, so this makes things difficult.  

My proposal is that we should have a command to let mods+ be able to give the video role.  This has a potential pitfall in that people may forget to remove it.  My solution to that is that the role should be revoked as soon as the screen share starts.  The user's stream will not be closed even if their permissions are revoked.  Additionally, we can set a timeout that removes the role if they don't start a stream in say... 5 to 10 minutes (the amount of time is just off the cuff.)

Command details summary:

- Grants the Video role to the user (would be the User ID, mention, etc)
~~- Waits to see when the user starts a stream and removes the role~~
~~- If the user does not start the stream after x minutes, remove the role~~
- Removes the role after a set amount of time
- Also has a command for revoking the role immediately

Should be relatively simple.

ChrisLovering:
Closes #1343

This is a fork from #1303 as the user could not continue with the project.

These commands allow mods to give temporary screen sharing permissions to users, with a default value pulled from the config.yml (currently set to 30minutes).

This cog also allows users to be given permanent video permissions, or have their permissions revoked.

This uses the redis cache to ensure that revoking temporary permissions is not forgotten about when the bot restarts.

Screenshots:
image
image


ChrisLovering:
We should squash some of these commits before pushing. I'll do so tomorrow if no one else has done so by then.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+        Alternatively, an ISO 8601 timestamp can be provided for the duration.
+        """
+        log.trace(f"Attempting to give temporary streaming permission to {user} ({user.id}).")
+        # if duration is none then calculate default duration
+        if duration is None:
+            now = datetime.datetime.utcnow()
+            duration = now + datetime.timedelta(minutes=VideoPermission.default_permission_duration)
+
+        # Check if user already has streaming permission
+        already_allowed = any(Roles.video == role.id for role in user.roles)
```
ToxicKidz, line 92:
```suggestion
        already_allowed = user._roles.has(Roles.video)
```
This should be changed in other places too.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+            log.debug(f"Scheduling {member} ({member.id}) to have streaming permission revoked at {revoke_time}")
+            self.scheduler.schedule_at(
+                revoke_time,
+                key,
+                self._remove_streaming_permission(member)
+            )
+
+    @commands.command(aliases=("streaming",))
+    @commands.has_any_role(*STAFF_ROLES)
+    async def stream(self, ctx: commands.Context, user: discord.Member, duration: Expiry = None) -> None:
```
ToxicKidz, line 69:
```suggestion
    async def stream(self, ctx: commands.Context, member: discord.Member, duration: Expiry = None) -> None:
```
This should be changed in other places too.

ToxicKidz (changes requested):
Looks good, a couple changes.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+        \u2003`M` - minutes∗
+        \u2003`s` - seconds
+
+        Alternatively, an ISO 8601 timestamp can be provided for the duration.
+        """
+        log.trace(f"Attempting to give temporary streaming permission to {user} ({user.id}).")
+        # if duration is none then calculate default duration
+        if duration is None:
+            now = datetime.datetime.utcnow()
+            duration = now + datetime.timedelta(minutes=VideoPermission.default_permission_duration)
```
Shivansh-007, line 89: `now` is a redundant variable, it is just getting used once, so we can merge it with `duration`.
```suggestion
            duration = datetime.datetime.utcnow() + datetime.timedelta(minutes=VideoPermission.default_permission_duration)
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+        items = await self.task_cache.items()
+        for key, value in items:
+            member = self.bot.get_guild(Guild.id).get_member(key)
+
+            if not member:
+                try:
+                    member = await self.bot.get_guild(Guild.id).fetch_member(key)
+                except discord.errors.NotFound:
+                    log.debug(
+                        f"Member {key} left the guild before we could scheudle "
```
Shivansh-007, line 51:
```suggestion
                        f"Member {key} left the guild before we could schedule "
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class Stream(commands.Cog):
+    """Grant and revoke streaming permissions from users."""
+
+    # Stores tasks to remove streaming permission
+    # User id : timestamp relation
+    task_cache = RedisCache()
```
Shivansh-007, line 22: Uhh, the variable name is possibly giving the wrong information, it doesn't store "task" cache, but rather the user ids. for which it has to remove the video roles.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+        await user.add_roles(discord.Object(Roles.video), reason="Temporary streaming access granted")
+        revoke_time = format_infraction_with_duration(str(duration))
+        await ctx.send(f"{Emojis.check_mark} {user.mention} can now stream until {revoke_time}.")
+        log.debug(f"Successfully given {user} ({user.id}) permission to stream until {revoke_time}.")
+
+    @commands.command(aliases=("pstream",))
+    @commands.has_any_role(*STAFF_ROLES)
+    async def permanentstream(self, ctx: commands.Context, user: discord.Member) -> None:
+        """Permanently grants the given user the permission to stream."""
+        log.trace(f"Attempting to give permenant streaming permission to {user} ({user.id}).")
```
Shivansh-007, line 110:
```suggestion
        log.trace(f"Attempting to give permanent streaming permission to {user} ({user.id}).")
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+        self.bot = bot
+        self.scheduler = Scheduler(self.__class__.__name__)
+        self.reload_task = self.bot.loop.create_task(self._reload_tasks_from_redis())
+
+    def cog_unload(self) -> None:
+        """Cancel all scheduled tasks."""
+        self.reload_task.cancel()
+        self.reload_task.add_done_callback(lambda _: self.scheduler.cancel_all())
+
+    async def _remove_streaming_permission(self, member: discord.Member) -> None:
```
Shivansh-007, line 34:
```suggestion
    async def _revoke_streaming_permission(self, member: discord.Member) -> None:
```
Revoke sounds better here.

Shivansh-007 (changes requested):

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+                    member = await self.bot.get_guild(Guild.id).fetch_member(key)
+                except discord.errors.NotFound:
+                    log.debug(
+                        f"Member {key} left the guild before we could scheudle "
+                        "the revoking of their streaming permissions."
+                    )
+                    await self.task_cache.delete(key)
+                    continue
+                except discord.HTTPException as e:
+                    log.exception(f"Exception while trying to retrieve member {key} from discord\n{e}")
```
MarkKoz, line 57: This is redundant cause the entire traceback will already be included anyway.
```suggestion
                    log.exception(f"Exception while trying to retrieve member {key} from Discord.")
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+        """Reload outstanding tasks from redis on startup, delete the task if the member has since left the server."""
+        await self.bot.wait_until_guild_available()
+        items = await self.task_cache.items()
+        for key, value in items:
+            member = self.bot.get_guild(Guild.id).get_member(key)
+
+            if not member:
+                try:
+                    member = await self.bot.get_guild(Guild.id).fetch_member(key)
+                except discord.errors.NotFound:
```
MarkKoz, line 49: This may actually return a 400 with a code of `10007` (unknown member) rather than a 404 (or even both may be possible). Please double-check this.

bot/exts/moderation/stream.py
ChrisLovering: It stores a list of userIDs that need to be scheduled to be removed, I think I agree with the original author that `task_cache` is the best way to describe this.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+        # Check if user already has streaming permission
+        already_allowed = any(Roles.video == role.id for role in user.roles)
+        if already_allowed:
+            await ctx.send(f"{Emojis.cross_mark} This user can already stream.")
+            log.debug(f"{user} ({user.id}) already has permission to stream.")
+            return
+
+        # Schedule task to remove streaming permission from Member and add it to task cache
+        self.scheduler.schedule_at(duration, user.id, self._remove_streaming_permission(user))
+        await self.task_cache.set(user.id, duration.timestamp())
```
MarkKoz, line 100: This suffers from the same issue the recent help channels PR did. `duration` is in UTC, but it is naïve. Therefore, `timestamp()` will interpret it in local time. However, when it's read from the cache and converted to a datetime, it's interpreted as UTC again with `utcfromtimestamp()`. I suggest using tz-aware objects through `arrow` to fix this.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+        await user.add_roles(discord.Object(Roles.video), reason="Temporary streaming access granted")
+        revoke_time = format_infraction_with_duration(str(duration))
+        await ctx.send(f"{Emojis.check_mark} {user.mention} can now stream until {revoke_time}.")
+        log.debug(f"Successfully given {user} ({user.id}) permission to stream until {revoke_time}.")
```
MarkKoz: There are several places in this file that have these clusters of lines. Adding some blank lines between different "sections" would help with readability.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,149 @@
...
+                await ctx.send(f"{Emojis.check_mark} Changed temporary permission to permanent.")
+                log.debug(f"Successfully upgraded temporary streaming permission for {user} ({user.id}) to permanent.")
+                return
+            await ctx.send(f"{Emojis.cross_mark} This user can already stream.")
+            log.debug(f"{user} ({user.id}) already had permanent streaming permission.")
+            return
+
+        await user.add_roles(discord.Object(Roles.video), reason="Permanent streaming access granted")
+        await ctx.send(f"{Emojis.check_mark} Permanently granted {user.mention} the permission to stream.")
+        log.debug(f"Successfully given {user} ({user.id}) permanent streaming permission.")
```
MarkKoz, line 126:
```suggestion
        log.debug(f"Successfully gave {user} ({user.id}) permanent streaming permission.")
```

MarkKoz (changes requested):

bot/exts/moderation/stream.py
Shivansh-007: Well I can't come up with something concise and better, so lets go with this.

bot/exts/moderation/stream.py
```diff
@@ -85,8 +85,9 @@ async def stream(self, ctx: commands.Context, member: discord.Member, duration:
         log.trace(f"Attempting to give temporary streaming permission to {member} ({member.id}).")
         # If duration is none then calculate default duration
         if duration is None:
-            now = datetime.datetime.utcnow()
-            duration = now + datetime.timedelta(minutes=VideoPermission.default_permission_duration)
+            duration = datetime.datetime.utcnow() + datetime.timedelta(
+                minutes=VideoPermission.default_permission_duration
+            )
```
MarkKoz: This is worse because it's 2 lines versus 3 and putting the argument on a separate line is less readable. Just a comment; not going make you go back and forth on this.

bot/exts/moderation/stream.py
```diff
@@ -85,7 +87,7 @@ async def stream(self, ctx: commands.Context, member: discord.Member, duration:
         log.trace(f"Attempting to give temporary streaming permission to {member} ({member.id}).")
         # If duration is none then calculate default duration
         if duration is None:
-            duration = datetime.datetime.utcnow() + datetime.timedelta(
+            duration = arrow.utcnow().naive + timedelta(
```
MarkKoz, line 90: Why are you making it naive and then using `Arrow.fromdatetime` on it again to make it aware?

bot/exts/moderation/stream.py
ChrisLovering: I've tried to find a case where a 400 is raised, and d.py don't seem to handle 400 specifically, so they would fall into a `HTTPException`. unknown members will raise a 404, which return up as `NotFound` with the code you mentioned.

bot/exts/moderation/stream.py
MarkKoz: Oh, I missed your Discord message. Anyway, you can avoid converting it back to aware by keeping it aware and only passing the naive version when the schedule function is called; adding the delta will still result in an `Arrow` object, right?

MarkKoz: Alternatively, `schedule_at` may be changed to somehow support both naïve and aware objects (probably by using now() in the object's timezone if it has one and utcnow otherwise).

bot/exts/moderation/stream.py
ChrisLovering: Due to the private nature of this API, it carries risks to use.

We were indecisive in whether to use it or not, so I'm going to leave it as-is or now.

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,161 @@
...
+            log.debug(f"Scheduling {member} ({member.id}) to have streaming permission revoked at {revoke_time}")
+            self.scheduler.schedule_at(
+                revoke_time,
+                key,
+                self._revoke_streaming_permission(member)
+            )
+
+    @commands.command(aliases=("streaming",))
+    @commands.has_any_role(*STAFF_ROLES)
+    async def stream(self, ctx: commands.Context, member: discord.Member, duration: Expiry = None) -> None:
```
MarkKoz, line 71: The `Expiry` converter returns a naïve object. Forgot about that!

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,161 @@
...
+        # Schedule task to remove streaming permission from Member and add it to task cache
+        self.scheduler.schedule_at(duration, member.id, self._revoke_streaming_permission(member))
+        await self.task_cache.set(member.id, Arrow.fromdatetime(duration).timestamp())
+
+        await member.add_roles(discord.Object(Roles.video), reason="Temporary streaming access granted")
+
+        # Convert here for nicer logging and output
+        revoke_time = format_infraction_with_duration(str(duration))
+        await ctx.send(f"{Emojis.check_mark} {member.mention} can now stream until {revoke_time}.")
+        log.debug(f"Successfully given {member} ({member.id}) permission to stream until {revoke_time}.")
```
MarkKoz, line 108: Missed this one, sorry.

```suggestion
        log.debug(f"Successfully gave {member} ({member.id}) permission to stream until {revoke_time}.")
```

bot/utils/scheduling.py
```diff
@@ -59,14 +59,18 @@ def schedule(self, task_id: t.Hashable, coroutine: t.Coroutine) -> None:
 
     def schedule_at(self, time: datetime, task_id: t.Hashable, coroutine: t.Coroutine) -> None:
         """
-        Schedule `coroutine` to be executed at the given naïve UTC `time`.
+        Schedule `coroutine` to be executed at the given `time`.
+
+        If `time` is timezone aware, then use that timezone to calculate now() when subtracting.
+        If `time` is naïve, then we use UTC.
```
MarkKoz, line 65: First-person (and second-person) pronouns should be avoided, especially in docstrings. Use the imperative mood.

```suggestion
        If `time` is naïve, then use UTC.
```

MarkKoz (changes requested):

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,166 @@
...
+        # Check if the member already has streaming permission
+        already_allowed = any(Roles.video == role.id for role in member.roles)
+        if already_allowed:
+            await ctx.send(f"{Emojis.cross_mark} This member can already stream.")
+            log.debug(f"{member} ({member.id}) already has permission to stream.")
+            return
+
+        # Schedule task to remove streaming permission from Member and add it to task cache
+        self.scheduler.schedule_at(duration, member.id, self._revoke_streaming_permission(member))
+        await self.task_cache.set(member.id, Arrow.fromdatetime(duration).timestamp())
```
MarkKoz, line 106: This conversion is redundant.

MarkKoz (approved):

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,166 @@
...
+                    member = await self.bot.get_guild(Guild.id).fetch_member(key)
+                except discord.errors.NotFound:
+                    log.debug(
+                        f"Member {key} left the guild before we could schedule "
+                        "the revoking of their streaming permissions."
+                    )
+                    await self.task_cache.delete(key)
+                    continue
+                except discord.HTTPException:
+                    log.exception(f"Exception while trying to retrieve member {key} from Discord.")
```
Shivansh-007, line 59: We can understand it is a `Exception` by the log level, but not which exception, so you can update it to:
```suggestion
                    log.exception(f"HTTPException while trying to retrieve member {key} from Discord.")
```

bot/exts/moderation/stream.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # Schedule task to remove streaming permission from Member and add it to task cache
+        self.scheduler.schedule_at(duration, member.id, self._revoke_streaming_permission(member))
+        await self.task_cache.set(member.id, duration.timestamp())
+
+        await member.add_roles(discord.Object(Roles.video), reason="Temporary streaming access granted")
+
+        # Convert here for nicer logging and output
+        revoke_time = format_infraction_with_duration(str(duration))
+        await ctx.send(f"{Emojis.check_mark} {member.mention} can now stream until {revoke_time}.")
```
Shivansh-007, line 112: Everyone has different time zones, and it can be difficult to figure out the time for a member, so we can use embeds here and set the timestamp as `revoke_time`, with some footer text before like `The user can now stream until`, and the timestamp would come after that, while the embed description can be `{Emojis.check_mark} {member.mention} has been granted permissions to stream`.

Shivansh-007 (changes requested):

bot/exts/moderation/stream.py
ChrisLovering: This line is just to set context, the full traceback will give the reader all they need to know.

bot/exts/moderation/stream.py
ChrisLovering: I'll give it a look when I'm back.

Akarys42 (changes requested):
It seems like if the permissions should be revoked while the bot is offline, the task will be removed from the cache without firing. 

You should make sure that it gets removed either through the reschedule function or inside the scheduler directly.

Akarys42 (approved):
Fake Redis was the culprit, let's go!

PR #1503, opened 2021-04-06
Add a !tp get_review command to get the nomination text without posting it

Akarys42:
This can be used when information should be added to the post, or someone wants to review the user. 

Approved by our nomination overlord @wookie184.

### DIFF
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -114,7 +114,7 @@ class Reviewer:
         vote_request = (
             "*Refer to their nomination and infraction histories for further details*.\n"
             f"*Please react {seen_emoji} if you've seen this post."
-            " Then react 👍 for approval, or 👎 for disapproval*."
+            " Then react :+1: for approval, or :-1: for disapproval*."
         )
 
         review = "\n\n".join(part for part in (opening, current_nominations, review_body, vote_request))
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -77,7 +77,7 @@ class Reviewer:
         log.trace(f"Posting the review of {user_id}")
         message = (await self._bulk_send(channel, review))[-1]
         if seen_emoji:
-            for reaction in (seen_emoji, "👍", "👎"):
+            for reaction in (seen_emoji, "\N{THUMBS UP SIGN}", "\N{THUMBS DOWN SIGN}"):
                 await message.add_reaction(reaction)
 
         if update_database:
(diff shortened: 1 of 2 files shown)


wookie184 (changes requested):
Just a small thing. 

Also, Discord doesn't seem to display the emojis properly in the file preview, at least for me, I get something like this:

image

The actual file when downloaded seems fine though, seems it could be a discord bug? I guess we could make it display with the `:thumbsup:` format, but I don't really mind leaving it as-is.

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -104,10 +118,7 @@ async def post_review(self, user_id: int, update_database: bool) -> None:
         )
 
         review = "\n\n".join(part for part in (opening, current_nominations, review_body, vote_request))
```
ToxicKidz, line 120:
```suggestion
        review = "\n\n".join((opening, current_nominations, review_body, vote_request))
```

ToxicKidz (changes requested):
Mostly looks good, one small suggestion.

bot/exts/recruitment/talentpool/_review.py
Akarys42: That's isn't my code :pensive: but thanks haha

wookie184 (approved):
Might make sense for line 273 in `_review.py`, to be changed to use `:eyes:` instead of `👀` since you've changed the other ones, but this looks good to me, thanks!

Akarys42:
> Might make sense for line 273 in `_review.py`, to be changed to use `:eyes:` instead of `👀` since you've changed the other ones, but this looks good to me, thanks!

To be honest, if we were to do that they should be Unicode points because Git messes up encoding sometimes. But we can't use the emoji names everwhere, it will fail in some places. I don't mind switching everything though.

ToxicKidz (approved):
Looks good to me!

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -66,26 +66,40 @@ def schedule_review(self, user_id: int) -> None:
...
+        if not review:
+            return
+
+        guild = self.bot.get_guild(Guild.id)
+        channel = guild.get_channel(Channels.nomination_voting)
+
         log.trace(f"Posting the review of {user_id}")
+        message = (await self._bulk_send(channel, review))[-1]
+        if seen_emoji:
+            for reaction in (seen_emoji, "\U0001f44d", "\U0001f44e"):
```
Numerlor, line 80: I think using `"\N{THUMBS UP SIGN}"` and `"\N{THUMBS DOWN SIGN}"` is a bit clearer for the emojis

Numerlor (changes requested):
Some of the sentences sent to the channel are missing periods but that includes some of the string from he previous revisions - feel free to fix them if you think it's worth it.

L205 in the `_cog` module is still using the emoji version of a checkmark

ChrisLovering (commented):
🔥

ChrisLovering (approved):
Looks good 👍

PR #1507, opened 2021-04-07
Add custom command checks tag

vcokltfre:
This PR adds a tag for creating custom decorators in discord.py for commands (different to the existing decorators tag in that it specifically shows how to make discord.py checks, a commonly required subject, rather than just decorators)

How the embed looks:
image


vcokltfre:
Updated embed:
image

bot/resources/tags/customchecks.md
```diff
@@ -0,0 +1,21 @@
+**Custom Command Checks in discord.py**
+
+You may find yourself in need of a check decorator to do something that doesn't exist in discord.py by default, but fear not, you can make your own! Using discord.py you can use `discord.ext.commands.check` to create you own checks like this:
```
MarkKoz, line 3: First sentence is unnecessary. If someone wants to know how to create a check then they already have a reason for why the want to do it. In any case, the notion of having to create something because it doesn't already exist is pretty obvious and doesn't need to be explained.

bot/resources/tags/customchecks.md
```diff
@@ -0,0 +1,21 @@
...
+from discord.ext.commands import check, Context
+
+def in_any_channel(*channels):
+  async def predicate(ctx: Context):
+    return ctx.channel.id in channels
+  return check(predicate)
+```
+There's a fair bit to break down here, so let's start with what we're trying to achieve with this check. As you can probably guess from the name it's locking a command to a **list of channels**. The inner function named `predicate` is used to perform the actual check on the command context. Here you can do anything that requires a `Context` object. This inner function should return `True` if the check is **successful** or `False` if the check **fails**.
+
+Here's how we might use our new check:
```
MarkKoz, line 14: Definitely a personal preference, but I really dislike the use of "we" like this. Furthermore, it's inconsistent with the rest of the text, which uses the second-person.

MarkKoz (changes requested):

vcokltfre:
Updated embed:
image

vcokltfre:
> First sentence is unnecessary. If someone wants to know how to create a check then they already have a reason for why they want to do it. In any case, the notion of having to create something because it doesn't already exist is pretty obvious and doesn't need to be explained.

I've elected to keep the sentence, but I've made it more matter of fact. The reasoning behind this is that I don't want it to sound too robotic or boring for those reading since this is a tag generally for beginners, not the docs page. I believe that without it the sentence sounds too 'un human' and sounds weird, and given it's a single line I see no harm in having it that way.

MarkKoz (approved):

Xithrius (approved):
Pretty darn useful tag.

PR #1509, opened 2021-04-07
Update YouTube terms in the `ytdl` tag

### LINKED ISSUE #1505: Update YouTube terms in `ytdl` tag
The [YouTube Terms of Service](https://www.youtube.com/static?template=terms) have been updated a bit since March 14, 2020 when they were last edited in the `ytdl` tag. (They were last changed March 17, 2021 according to the site.)

Specifically, it looks like we need to fix the wording in restrictions 1 and 3.

https://github.com/python-discord/bot/blob/eeacceae01b95e39eaeecab2fd14f6edfb19b94b/bot/resources/tags/ytdl.md

I'm happy to tackle this one really quick once it's assigned to me. :)

bsoyka:
Fixes #1505.

### DIFF
--- a/bot/resources/tags/ytdl.md
+++ b/bot/resources/tags/ytdl.md
@@ -1,12 +1,12 @@
 Per [PyDis' Rule 5](https://pythondiscord.com/pages/rules), we are unable to assist with questions related to youtube-dl, commonly used by Discord bots to stream audio, as its use violates YouTube's Terms of Service.
 
-For reference, this usage is covered by the following clauses in [YouTube's TOS](https://www.youtube.com/static?template=terms), as of 2019-07-22:
+For reference, this usage is covered by the following clauses in [YouTube's TOS](https://www.youtube.com/static?template=terms), as of 2021-03-17:
 ```
-The following restrictions apply to your use of the Service. You are not allowed to:  
+The following restrictions apply to your use of the Service. You are not allowed to:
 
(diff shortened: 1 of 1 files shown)


ChrisLovering (approved):

wookie184:
The Terms of Service seem to be different in different countries, the UK version seems to be the one that was used previously, and the relevant info for that one hasn't changed as far as I can tell, see https://www.youtube.com/static?gl=GB&template=terms. It's just that the US version seems to have slightly different wording. I guess it doesn't really matter which one we use though

bsoyka:
Ah, that would make sense. Does the link in the tag automatically redirect to the UK version if that's where you're at? If so, we should probably just change the link instead, but if not, the wording should be the US version so it matches up.

MarkKoz:
I thought it detects one's location and shows the appropriate version.

bsoyka:
That's what I'm wondering - if that's true then we should probably see if there's a way to have a link that points to one specific version regardless of location so that the tag shows the right wording.

wookie184:
Yeah what I meant to say was that it shows a different document depending on the location it detects, although it seems you can override it by providing `gl=` in the URL, hence the URL I posted before, https://www.youtube.com/static?gl=GB&template=terms to show the UK version. `gl=US` should show the US version in the same way

bsoyka:
Just discussed with Joe briefly [in `#dev-contrib`](https://discord.com/channels/267624335836053506/635950537262759947/829449751866638336) and I'm gonna switch the wording back to the UK version and fix the link, then it should be all good to go.

Xithrius (approved):

PR #1510, opened 2021-04-07
Made bot to send DM when a reminder is sent

traliotube:


mbaruh:
You need to properly open an issue and have it approved, and you have lint and syntax errors.

PR #1511, opened 2021-04-08
Close voice channels with defcon shutdown

onerandomusername:
## Description

Simple addition to remove the connect permission from the everyone role and restore it when done.

## Reasoning

Keep people from joining voice channels during a raid, making the entire server as read only as it can get.

## Did you:

- [x] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x] If dependencies have been added or updated, run `pipenv lock`?
- [x] **Lint your code** (`pipenv run lint`)?
- [x] Set the PR to **allow edits from contributors**?

### DIFF
--- a/bot/exts/moderation/defcon.py
+++ b/bot/exts/moderation/defcon.py
@@ -181,7 +181,7 @@ class Defcon(Cog):
         role = ctx.guild.default_role
         permissions = role.permissions
 
-        permissions.update(send_messages=False, add_reactions=False)
+        permissions.update(send_messages=False, add_reactions=False, connect=False)
         await role.edit(reason="DEFCON shutdown", permissions=permissions)
         await ctx.send(f"{Action.SERVER_SHUTDOWN.value.emoji} Server shut down.")
 
@@ -192,7 +192,7 @@ class Defcon(Cog):
         role = ctx.guild.default_role
         permissions = role.permissions
 
-        permissions.update(send_messages=True, add_reactions=True)
+        permissions.update(send_messages=True, add_reactions=True, connect=True)
         await role.edit(reason="DEFCON unshutdown", permissions=permissions)
         await ctx.send(f"{Action.SERVER_OPEN.value.emoji} Server reopened.")
 


dawnofmidnight (approved):
LGTM! :shipit:

ChrisLovering (commented):
If we're going for shutting down voice channels, should we set `speak` to False too?

We could also take it a step further and kick people from VCs too.

onerandomusername:
> 
> 
> If we're going for shutting down voice channels, should we also set `speak` to False too?

No, because if you can't connect to a voice channel, you can't talk in them.

ChrisLovering:
> > If we're going for shutting down voice channels, should we also set `speak` to False too?
> 
> No, because if you can't connect to a voice channel, you can't talk in them.

I'm referring to users who are already in voice channels.

onerandomusername:
> 
> 
> > > If we're going for shutting down voice channels, should we also set `speak` to False too?
> > 
> > 
> > No, because if you can't connect to a voice channel, you can't talk in them.
> 
> I'm referring to users who are already in voice channels.

If they're already in voice channel, permissions aren't applied until one leaves and rejoins. To fix this, the bot would need to recurse through everyone in the voice channels and disconnect them from the channels.

dawnofmidnight:
> > > > If we're going for shutting down voice channels, should we also set `speak` to False too?
> > > 
> > > 
> > > No, because if you can't connect to a voice channel, you can't talk in them.
> > 
> > 
> > I'm referring to users who are already in voice channels.
> 
> If they're already in voice channel, permissions aren't applied until one leaves and rejoins. To fix this, the bot would need to recurse through everyone in the voice channels and disconnect them from the channels.
> 
> I'm not sure we have the presence intent to get everyone in the voice channels and the ratelimits to remove them. Previously, staff has deleted the voice channels, which would be another possible solution to remove everyone at once.

Why not just hide voice channels?

ChrisLovering:
> > > > If we're going for shutting down voice channels, should we also set `speak` to False too?
> > > 
> > > 
> > > No, because if you can't connect to a voice channel, you can't talk in them.
> > 
> > 
> > I'm referring to users who are already in voice channels.
> 
> If they're already in voice channel, permissions aren't applied until one leaves and rejoins. To fix this, the bot would need to recurse through everyone in the voice channels and disconnect them from the channels.
> 
> I'm not sure we have the presence intent to get everyone in the voice channels and the ratelimits to remove them. Previously, staff has deleted the voice channels, which would be another possible solution to remove everyone at once.

We have the intent, but agreed, it's not worth the effort of iterating through connected members and kicking them.

ChrisLovering (approved):

vcokltfre (approved):

Akarys42:
Hey @onerandomusername, thanks for the PR! It seems like you don't say *why* those changes were made. Chris pointed me to the discussion in `#dev-contrib` from last time, if you could mention it on your next PRs that'd be perfect!

Akarys42 (approved):
🚀

PR #1512, opened 2021-04-09
Log exception info for unhandled exceptions from event listeners

MarkKoz:
It's pretty difficult to debug without this information.

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -308,7 +308,7 @@ class Bot(commands.Bot):
             scope.set_extra("args", args)
             scope.set_extra("kwargs", kwargs)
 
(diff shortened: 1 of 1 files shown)


MarkKoz:
Actually, since d.py calls this from an `except` block, it should already have the exception info implicitly. However, for some reason, it doesn't exist for [this issue](https://sentry.io/organizations/python-discord/issues/2326913752/?project=2724195).

Akarys42:
> Is that from an internal eval?

@MarkKoz this specific issue is indeed coming from an `int e`

MarkKoz:
In that case, I'll close this.

PR #1514, opened 2021-04-10
AntiSpam: prevent attempts to punish a user multiple times 

### LINKED ISSUE #902: Only punish users once to prevent a race condition in AntiSpam
If a user manages to take part of multiple filter violations, the cog will attempt to `punish()` them for each of those violations. While `punish()` does check for the muted role before applying the infraction, the asynchronous nature of the code can mean a second `punish()` is called and passes the role check before the first `punish()` has gotten a chance to apply the mute. This would result in an attempt to apply a second mute infraction, which will ultimately fail and display a message stating the user already has such infraction.

I believe the fix is to keep track of which members were already punished so that they can be skipped when processing members for the rule violation. One has to consider this globally, not just for the current deletion context. Maybe using locks for each member would just be easier.

MarkKoz:
Fixes #902

A user may manage to send multiple message that violate filters before the mute is applied. Because of a race condition, subsequent punish attempts did not detect the mute role exists and therefore proceeded to apply another mute.

To avoid the race condition, abort any subsequent punish attempts while one is already ongoing for a given user. It could be possible to wait instead of abort, but the first attempt failing very likely means subsequent attempts would fail too.


jb3 (approved):
Looks good!

Numerlor (approved):

Xithrius (approved):

PR #1517, opened 2021-04-12
Create identity.md

SavagePastaMan:
Tag to demonstrate the difference between `is` and `==`.


vcokltfre:
For those curious here's what the tag looks like:
image

bot/resources/tags/identity.md
```diff
@@ -0,0 +1,25 @@
...
+To check if two things are equal, use the equality operator (`==`).
+```py
+x = 5
+if x == 5:
+    print("x equals 5")
+if x == 3:
+    print("x equals 3")
+# Prints 'x equals 5'
+```
+
```
Akarys42, line 14: This empty line should be removed

Akarys42 (commented):
I'm not a fan of the use of the word "thing". Maybe "object" would be a good replacement?

mbaruh:
I wonder if this will result in questions about `None` and booleans. Maybe there should be a short note about that.

Xithrius:
Author will not be able to complete tag. This PR is now up for grabs. Ask a core developer if you'd like to take over.

vcokltfre:
Updated embed:
image

bot/resources/tags/identity.md
```diff
@@ -0,0 +1,24 @@
...
+To check if two things are equal, use the equality operator (`==`).
+```py
+x = 5
+if x == 5:
+    print("x equals 5")
+if x == 3:
+    print("x equals 3")
+# Prints 'x equals 5'
+```
+To check if two things are actually the same thing in memory, use the identity comparison operator (`is`).
```
Akarys42, line 14: I still don’t like the use of `things` here. Perhaps `object` is a better word?

bot/resources/tags/identity.md
```diff
@@ -0,0 +1,24 @@
...
+```py
+x = [1, 2, 3]
+y = [1, 2, 3]
+if x is [1, 2, 3]:
+    print("x is y")
+z = x
```
Akarys42, line 20: This snippet could get more explicit variable names too, it is a bit hard to follow right now. Also using the integer `420` would be less noisy than `[1, 2, 3]` and keep the same behavior.

Akarys42 (changes requested):
Couple of things

bot/resources/tags/identity.md
mbaruh: I think there's merit in using specifically lists, as it is something that confuses a lot of people.

For integers and strings this works only if the value is cached and shouldn't be relied on.

vcokltfre: I've kept it using the list given what zig said

Akarys42 (approved):
Lgtm, thanks!

Akarys42 (commented):
.

Akarys42 (commented):
Please review bot

mbaruh (approved):

PR #1520, opened 2021-04-12
Create str-join.md

SavagePastaMan:
Create a tag to showcase `str.join`

Here is a screenshot of the tag 
image


Akarys42:
Hey @SavagePastaMan, can you please add a bit more content to the PR body? right now it isn't saying much about the PR either how it looks inside the server. Also, don't forget to lint!

SavagePastaMan:
i'm just using the github editor :pensive:. I also don't have a local instance of the bot so I can't get screenshots 😬.

ChrisLovering (approved):

bot/resources/tags/str-join.md
```diff
@@ -0,0 +1,28 @@
+**Joining Iterables**
+
+Suppose you want to nicely display a list (or some other iterable). The naive solution would be something like this.
```
mbaruh, line 3:
```suggestion
If you want to display a list (or some other iterable), you can write:
```

bot/resources/tags/str-join.md
```diff
@@ -0,0 +1,28 @@
...
+```py
+colors = ['red', 'green', 'blue', 'yellow']
+output = ""
+separator = ", "
+for color in colors:
+    output += color + separator
+print(output)
+# Prints 'red, green, blue, yellow, '
+```
+However, this solution is flawed. The separator is still added to the last element, and it is slow.
```
mbaruh, line 13:
```suggestion
However, the separator is still added to the last element, and it is relatively slow.
```

bot/resources/tags/str-join.md
```diff
@@ -0,0 +1,28 @@
...
+
+The better solution is to use `str.join`.
```
mbaruh, line 15:
```suggestion
A better solution is to use `str.join`.
```

bot/resources/tags/str-join.md
```diff
@@ -0,0 +1,28 @@
...
+```py
+colors = ['red', 'green', 'blue', 'yellow']
+separator = ", "
+print(separator.join(colors))
+# Prints 'red, green, blue, yellow'
+```
+This solution is much simpler, faster, and solves the problem of the extra separator. An important thing to note is that you can only `str.join` strings. For a list of ints,
```
mbaruh, line 22:
```suggestion
An important thing to note is that you can only `str.join` strings. For a list of ints,
```

mbaruh (changes requested):
I'd keep it as short as possible, and also maybe a bit less opinionated.

Xithrius:
Author will not be able to complete this PR, so I'll be taking over.

mbaruh (approved):

Akarys42 (approved):

PR #1521, opened 2021-04-12
Use == instead of str.startswith when getting the source for a Command

### LINKED ISSUE #1519: Logic in Source Command's Help Command Check
## Description

Its impossible to get the source for the HelpChannel cog.

## Steps to Reproduce

`!src HelpChannel`
Due to the logic of the command, all that needs to be said is `!src help` followed by anything.

## Expected Behaviour

Source link to the HelpChannel cog is shown.

## Actual Behaviour

Help command source link is shown.

## Known Impacted Platforms

- [x] Web
- [x] Desktop
- [x] Android App
- [x] iOS App

## Possible Solutions

These lines are the probable cause.
https://github.com/python-discord/bot/blob/51af1369e0d9c2ad185f0c0920b599b7187ed077/bot/exts/info/source.py#L19-L20
After looking at sir-lancebot, it might even be possible to remove these two lines altogether.
## Would you like to implement a fix?

***Note: For high-priority or critical bugs, fixes may be implemented by staff.***

- [ ] I'd like to implement the bug fix
- [x] Anyone can implement the bug fix

ToxicKidz:
Currently, when you do `!source HelpChannels` (or any command/cog/tag that starts with help), it shows the source for the help command. This is because we check if the lowered string *starts with* help, and not checking if it is equal to help. This PR uses == instead of str.startswith instead.
## Relevant Issues
Closes #1519


onerandomusername (approved):
:shipit:

MarkKoz (approved):

ChrisLovering (approved):

PR #1524, opened 2021-04-13
Use a paginated embed to output multiple snowflakes

ChrisLovering:
## Description
Previously each snowflake passed to the command would have their own embed, which may cause the bot to send many embeds if a staff unknowingly passed it a bunch of snowflakes. This change makes sure that we don't run into rate limits on the bot by sending all of the snowflakes in one embed.

## Screenshots
No Snowflake
image

1 snowflake
image

3 snowflakes
image

Many snowflakes
image

### DIFF
--- a/bot/exts/utils/utils.py
+++ b/bot/exts/utils/utils.py
@@ -162,6 +162,9 @@ class Utils(Cog):
         if len(snowflakes) > 1 and await has_no_roles_check(ctx, *STAFF_ROLES):
             raise BadArgument("Cannot process more than one snowflake in one invocation.")
 
+        if not snowflakes:
+            raise BadArgument("At least one snowflake must be provided.")
+
         embed = Embed(
             colour=Colour.blue()
         )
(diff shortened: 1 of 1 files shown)


bot/exts/utils/utils.py
```diff
@@ -162,17 +162,29 @@ async def snowflake(self, ctx: Context, *snowflakes: Snowflake) -> None:
         if len(snowflakes) > 1 and await has_no_roles_check(ctx, *STAFF_ROLES):
             raise BadArgument("Cannot process more than one snowflake in one invocation.")
 
+        if not snowflakes:
+            raise BadArgument("At least one snowflake must be provided.")
+
+        embed = Embed(
+            colour=Colour.blue()
+        )
```
MarkKoz, line 170: I don't think the readability is poor as a single line.

MarkKoz (approved):

bot/exts/utils/utils.py
ChrisLovering: Agreed, this was a copy-paste oversight from when we also set the description here.

ks129 (approved):

PR #1527, opened 2021-04-15
Re-work clean command to be more efficient, have more features.

Senjan21:
# Justification
Our current implementation of `clean` is fairly inefficient with how it handles message deletion.
**Few examples of that are:**

- Using two different ways of deleting messages, depending on command used. [Line 164](https://github.com/python-discord/bot/blob/main/bot/exts/utils/clean.py#L164) and [Line 166](https://github.com/python-discord/bot/blob/main/bot/exts/utils/clean.py#L166).
- Using minimum 3 API calls in order to delete <=100 messages in single channel ([Channel history (1)](https://github.com/python-discord/bot/blob/main/bot/exts/utils/clean.py#L130-L152), [Purge (2 Calls, 1 for reading channel history, 1 for message deletion)](https://github.com/python-discord/bot/blob/main/bot/exts/utils/clean.py#L166)).
- Cleaning can only be stopped before message reading starts ([Line 126](https://github.com/python-discord/bot/blob/main/bot/exts/utils/clean.py#L126)) which often times might be too late.
- In order to delete messages across multiple channels, all channels must be listed.

# Changes made
- Re-worked underlying `clean` system to not use `TextChannel.purge` but other methods such as `TextChannel.delete_messages`.
- Checking for cleaning cancelation at multiple points throughout whole command execution.
- Created new command for cleaning messages within range of messages.
- `clean` now uses Message Cache in few places in order to make the command more friendly to use for deletion of messages from multiple channels.
- Changed name of `message` subcommand to `until` for better self documentation.

# Example Usage

**This 2 examples use `clean user` however same syntax will work with `clean regex`**
Delete last 10 messages sent by `X` user, regardless of which channels they were sent in:
(This will only work if said messages were sent after bot started, while they are within the message cache)
`!clean user X 10`

Scan last 10 messages sent in each specified channel(s) and delete the ones that were sent by `X` user.
(This will work regardless of when messages were sent)
`!clean user X 10 channel1 channel2 channel3...`

Delete all messages between two specified messages, including said messages:
(Both specified messages must be from the same channel)
`!clean from-to first_message last_message`


bot/exts/utils/clean.py
```diff
@@ -36,15 +41,80 @@ def mod_log(self) -> ModLog:
...
+                return
+            try:
+                await message.delete()
+            except NotFound:
+                # Message doesn't exist or was already deleted
+                continue
+
+    def _get_messages_from_cache(self, amount: int, predicate: CheckHint) -> Tuple[DefaultDict, List[int]]:
+        """Helper function for getting messages from the cache."""
+        message_mappings = defaultdict(lambda: [])
```
ToxicKidz, line 57:
```suggestion
        message_mappings = defaultdict(list)
```

bot/exts/utils/clean.py
```diff
@@ -36,15 +41,80 @@ def mod_log(self) -> ModLog:
...
+        return message_mappings, message_ids
+
+    async def _get_messages_from_channels(
+        self,
+        amount: int,
+        channels: Iterable[TextChannel],
+        predicate: CheckHint,
+        until_message: Optional[Message] = None
+    ) -> DefaultDict:
+        message_mappings = defaultdict(lambda: [])
```
ToxicKidz, line 82:
```suggestion
        message_mappings = defaultdict(list)
```

bot/exts/utils/clean.py
```diff
@@ -36,15 +41,80 @@ def mod_log(self) -> ModLog:
...
+        message_ids = []
+
+        for channel in channels:
+
+            async for message in channel.history(amount=amount):
```
ToxicKidz, line 87:
```suggestion
            async for message in channel.history(limit=amount):
```
`amount` isn't a valid kwarg, but `limit` is.

bot/exts/utils/clean.py
```diff
@@ -245,11 +370,14 @@ async def clean_regex(
         channels: commands.Greedy[TextChannel] = None
     ) -> None:
         """Delete all messages that match a certain regex, stop cleaning after traversing `amount` messages."""
-        await self._clean_messages(amount, ctx, regex=regex, channels=channels)
+        use_cache = True
+        if channels:
+            use_cache = False
```
ToxicKidz, line 375: Just a nitpick
```suggestion
        use_cache = not channels
```

ToxicKidz (changes requested):

bot/exts/utils/clean.py
```diff
@@ -15,6 +17,9 @@
 
 log = logging.getLogger(__name__)
 
+# Type alias for checks
+CheckHint = Callable[[Message], bool]
```
mbaruh, line 21: `Predicate` might be a better name here.

bot/exts/utils/clean.py
```diff
@@ -36,15 +41,80 @@ def mod_log(self) -> ModLog:
...
+                    return message_mappings, message_ids
+
+        # Amount exceeds amount of messages matching the check
+        return message_mappings, message_ids
+
+    async def _get_messages_from_channels(
+        self,
+        amount: int,
+        channels: Iterable[TextChannel],
+        predicate: CheckHint,
```
mbaruh, line 79: I would change the name of the argument here and in other functions to `to_delete`, as then you get a sweet `if to_delete(message):`

bot/exts/utils/clean.py
```diff
@@ -79,6 +149,10 @@ def predicate_regex(message: Message) -> bool:
             else:
                 return bool(re.search(regex.lower(), content.lower()))
 
+        def predicate_range(message: Message) -> bool:
+            """Check if message is older than message provided in after_message but younger than until_message."""
+            return message.created_at > after_message.created_at and message.created_at < until_message.created_at
```
mbaruh, line 154: As the `until` clean is inclusive, meaning it deletes the message which is specified in the command, I think this should be inclusive as well, meaning have `<=` and `>=`.

bot/exts/utils/clean.py
```diff
@@ -89,6 +163,38 @@ def predicate_regex(message: Message) -> bool:
...
+                )
+                await ctx.send(embed=embed)
+                return
+
+            # Check if the messages are not in same channel
+            if after_message.channel != until_message.channel:
+                embed = Embed(
+                    color=Colour(Colours.soft_red),
+                    title=random.choice(NEGATIVE_REPLIES),
+                    description="You cannot do range clean across different channel."
```
mbaruh, line 183:
```suggestion
                    description="You cannot do range clean across several channels."
```

bot/exts/utils/clean.py
```diff
@@ -89,6 +163,38 @@ def predicate_regex(message: Message) -> bool:
             await ctx.send(embed=embed)
             return
 
+        if after_message:
+
+            # Ensure that until_message is specified.
+            if not until_message:
+                embed = Embed(
```
mbaruh, line 170: It seems to me that this embed and others would be better as thrown exceptions, such as `MissingRequiredArgument`, `BadArgument`, or possibly custom errors subclassed from `UserInputError` in `discord.ext.commands.error`.

bot/exts/utils/clean.py
```diff
@@ -258,6 +386,18 @@ async def clean_message(self, ctx: Context, message: Message) -> None:
             until_message=message
         )
 
+    @clean_group.command(name="from-to", aliases=["after-until"])
```
mbaruh, line 389: I suggest `between` as an alias.

bot/exts/utils/clean.py
```diff
@@ -121,54 +229,68 @@ def predicate_regex(message: Message) -> bool:
...
+
+        if not self.cleaning:
+            # Means that the cleaning was canceled
+            return
+
+        # Now let's delete the actual messages with purge.
+        self.mod_log.ignore(Event.message_delete, *message_ids)
+
+        # Creates ID like int object that would represent an object that is exactly 14 days old
+        minimum_time = int((time.time() - 14 * 24 * 60 * 60) * 1000.0 - 1420070400000) << 22
```
mbaruh, line 252: I would change the name of the variable to better document the meaning of the comparison a few lines below. Something like `two_week_old_snowflake`.

mbaruh: Also, is there a reason to not check the age of the message instead?

mbaruh (changes requested):
Will test functionality soon

bot/exts/utils/clean.py
Senjan21: Sounds good, both changed, https://github.com/python-discord/bot/pull/1527/commits/11be1666fce272d13e72467563fde53eae0f8419

bot/exts/utils/clean.py
Senjan21: Am not sure why did I miss that, thanks for the sharp eye! https://github.com/python-discord/bot/pull/1527/commits/5bf42c7d3c4927dae2a130a95d83295db746338d

bot/exts/utils/clean.py
Senjan21: Looks better, thanks, https://github.com/python-discord/bot/pull/1527/commits/add078121aba4630fef86a44b261211de89f4f95

bot/exts/utils/clean.py
Senjan21: I agree, https://github.com/python-discord/bot/pull/1527/commits/477810f86387c85ae6da0b80ffccb40dac60e26e

bot/exts/utils/clean.py
Senjan21: Sounds good, thanks, https://github.com/python-discord/bot/pull/1527/commits/708063c70bcea2cdeecb1eb66c703817e0195042

bot/exts/utils/clean.py
Senjan21: Sounds better, alias and this was changed in https://github.com/python-discord/bot/pull/1527/commits/c2705492ec13984aa75f4c31532ca18b1756121d, thanks

bot/exts/utils/clean.py
Senjan21: Main reason as to why this logic even exists is to ensure that, if someone decides to use the function, not from the command interface, they'd get a response, am not sure how useful it is.
I think for that matter it could be removed.
As far as regular user is concerned, that should already be the case as per [Line 387](https://github.com/python-discord/bot/blob/cleanrework/bot/exts/utils/clean.py#L387) the command will error out

bot/exts/utils/clean.py
Senjan21: > Also, is there a reason to not check the age of the message instead?
This part of the code was inspired by how `purge` works internally.
Comparison on message age could possibly be less accurate which in turn would resort in problems with message deletion if said messages are very close to the 14d mark.

mbaruh: Might be a little less readable this way, although if there's a known limitation I suppose it's fine. In that case though this kind of decision should be added to the comment above the line.

bot/exts/utils/clean.py
mbaruh: I think this part should be exported to a helper function and replaced with errors. The error handler should translate them into embeds and send them.

mbaruh: But yes, if we don't expect the function to be used in another context then I'm not sure there's a point

mbaruh:
Can you give invocation examples for the different commands that use the new functionalities?

bot/exts/utils/clean.py
```diff
@@ -36,15 +42,91 @@ def mod_log(self) -> ModLog:
...
+                        # Means we have found the message until which we were supposed to be deleting.
+                        break
+
+                if to_delete(message):
+                    message_mappings[message.channel].append(message)
+                    message_ids.append(message.id)
+
+        return message_mappings, message_ids
+
+    def is_older_than_14d(self, message: Message) -> bool:
```
mbaruh, line 108: This method can be static

bot/exts/utils/clean.py
```diff
@@ -106,69 +224,84 @@ def predicate_regex(message: Message) -> bool:
...
-                    messages.append(message)
+                if len(to_delete) == 100:
+                    # we can only delete up to 100 messages in a bulk
+                    await channel.delete_messages(to_delete)
+                    to_delete.clear()
 
-                # If the message passes predicate, let's save it.
-                if predicate is None or predicate(message):
-                    message_ids.append(message.id)
+            if len(to_delete) > 0:
```
mbaruh, line 290: For whatever reason this isn't deleting messages when they come from the cache. I made sure that the code actually reaches that point, and it seems like `to_delete` is populated as expected.

bot/exts/utils/clean.py
```diff
@@ -106,69 +224,84 @@ def predicate_regex(message: Message) -> bool:
...
+        for messages in message_mappings.values():
+            log_messages.extend(messages)
+
+        if log_messages:
+            # Reverse the list to restore chronological order
+            log_messages = reversed(log_messages)
+            log_url = await self.mod_log.upload_log(log_messages, ctx.author.id)
         else:
             # Can't build an embed, nothing to clean!
             embed = Embed(
```
mbaruh, line 307: This is a bit out of scope, but I think this section until the end of the function should be exported to a helper function such as `log_message_clean` or whatever name might fit.

mbaruh (changes requested):
I still think that the error embeds should be removed. Either completely, or placed in a separate function and replaced with error invocations, as it's the error handler's role to catch errors, convert them into embeds, and send them.

bot/exts/utils/clean.py
```diff
@@ -36,15 +42,91 @@ def mod_log(self) -> ModLog:
...
+        channels: Iterable[TextChannel],
+        to_delete: Predicate,
+        until_message: Optional[Message] = None
+    ) -> DefaultDict:
+        message_mappings = defaultdict(list)
+        message_ids = []
+
+        for channel in channels:
+
+            async for message in channel.history(limit=amount):
```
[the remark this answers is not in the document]
mbaruh: To be clear, `until` works as expected, `between` doesn't.

janine9vn:
Hi! This is a request for this PR. Totally 1000% is this request is out of scope for this PR though.
Could we have an option to essentially allow `!clean user <id> 30M` which would clean a users messages from all channels. It would only target messages that have been made during the specified duration. So in the example here, it would clean all the messages the user has posted in the past 30 minutes from all channels.

I can provide reasoning for this here or in the mod channels if you prefer.

Xithrius:
@Senjan21 How's progress on this PR? Will you be able to finish it?

bot/exts/utils/clean.py
```diff
@@ -36,15 +42,91 @@ def mod_log(self) -> ModLog:
...
+        # Amount exceeds amount of messages matching the check
+        return message_mappings, message_ids
+
+    async def _get_messages_from_channels(
+        self,
+        amount: int,
+        channels: Iterable[TextChannel],
+        to_delete: Predicate,
+        until_message: Optional[Message] = None
+    ) -> DefaultDict:
```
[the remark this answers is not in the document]
ChrisLovering: Resolved in b58e47df

bot/exts/utils/clean.py
ChrisLovering: Resolved in 02b3c8af

bot/exts/utils/clean.py
ChrisLovering: Added in https://github.com/python-discord/bot/pull/1527/commits/ed352272a67224182178bbd5583746053ec912a6

bot/exts/utils/clean.py
```diff
@@ -79,6 +161,10 @@ def predicate_regex(message: Message) -> bool:
             else:
                 return bool(re.search(regex.lower(), content.lower()))
 
+        def predicate_range(message: Message) -> bool:
+            """Check if message is older than message provided in after_message but younger than until_message."""
+            return message.created_at >= after_message.created_at and message.created_at <= until_message.created_at
```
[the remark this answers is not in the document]
ChrisLovering: Done in https://github.com/python-discord/bot/pull/1527/commits/770528c70ff38b739c963c88b89ec6401d687d16

bot/exts/utils/clean.py
```diff
@@ -36,15 +42,91 @@ def mod_log(self) -> ModLog:
...
+        message_mappings = defaultdict(list)
+        message_ids = []
+
+        for channel in channels:
+
+            async for message in channel.history(limit=amount):
+
+                if not self.cleaning:
+                    # Cleaning was canceled
+                    return (message_mappings, message_ids)
```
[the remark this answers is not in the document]
ChrisLovering: Done in https://github.com/python-discord/bot/pull/1527/commits/0cc135de21c8fe8a85b2c42b95b04779b3af7baa

bot/exts/utils/clean.py
```diff
@@ -106,69 +224,84 @@ def predicate_regex(message: Message) -> bool:
...
 
-        messages = []
-        message_ids = []
         self.cleaning = True
 
-        # Find the IDs of the messages to delete. IDs are needed in order to ignore mod log events.
-        for channel in channels:
-            async for message in channel.history(limit=amount):
+        if use_cache:
+            message_mappings, message_ids = self._get_messages_from_cache(amount=amount, predicate=predicate)
```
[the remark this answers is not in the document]
ChrisLovering: Fixed in https://github.com/python-discord/bot/pull/1527/commits/b2e76ddc6f4d3ccd327f48d9333eb977ddfb72d2

bot/exts/utils/clean.py
```diff
@@ -106,69 +224,84 @@ def predicate_regex(message: Message) -> bool:
...
+        else:
+            message_mappings, message_ids = await self._get_messages_from_channels(
+                amount=amount,
+                channels=channels,
+                predicate=predicate,
```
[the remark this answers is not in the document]
ChrisLovering: fixed in https://github.com/python-discord/bot/pull/1527/commits/b2e76ddc6f4d3ccd327f48d9333eb977ddfb72d2

bot/exts/utils/clean.py
ChrisLovering: Could you explain how you produced this bug? I naively just spammed some messages, and then used clean user, and it worked as expected

mbaruh:
I'll be taking over this PR.

mbaruh:
Opening a new PR as this one has become pretty messy and the description is outdated.

PR #1528, opened 2021-04-15
Allow eval almost everywhere

mbaruh:
## Description
Adds a check to blacklist a command only in a specific context, with an option for a role override. The check is applied to the eval command to blacklist it only from python-general.

## Questionable choices
* The check uses the in_whitelist_check check internally. It's possible that it's better to manually handle the logic of the predicate.
* This still uses the InWhiteListCheckFailure exception, even though it explicitly states in its docstring that it's raised for the in_whitelist_check. I wasn't sure if it's better to do this, or to duplicate the class's code with a different name. There might be a third solution I'm not seeing.

### DIFF
--- a/bot/decorators.py
+++ b/bot/decorators.py
@@ -11,7 +11,7 @@ from discord.ext.commands import Cog, Context
 
 from bot.constants import Channels, DEBUG_MODE, RedirectOutput
 from bot.utils import function
-from bot.utils.checks import InWhitelistCheckFailure, in_whitelist_check
+from bot.utils.checks import ContextCheckFailure, in_whitelist_check
 from bot.utils.function import command_wraps
 
 log = logging.getLogger(__name__)
@@ -43,6 +43,10 @@ def in_whitelist(
         return in_whitelist_check(ctx, channels, categories, roles, redirect, fail_silently)
 
     return commands.check(predicate)
+
+
+class NotInBlacklistCheckFailure(ContextCheckFailure):
+    """Raised when the 'not_in_blacklist' check fails."""
 
 
 def not_in_blacklist(
@@ -77,7 +81,7 @@ def not_in_blacklist(
         success = not_blacklisted or overridden
 
         if not success and not fail_silently:
-            raise InWhitelistCheckFailure(redirect)
+            raise NotInBlacklistCheckFailure(redirect)
 
         return success
 
--- a/bot/decorators.py
+++ b/bot/decorators.py
@@ -11,7 +11,7 @@ from discord.ext.commands import Cog, Context
 
 from bot.constants import Channels, DEBUG_MODE, RedirectOutput
 from bot.utils import function
-from bot.utils.checks import in_whitelist_check
+from bot.utils.checks import InWhitelistCheckFailure, in_whitelist_check
 from bot.utils.function import command_wraps
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


bot/decorators.py
```diff
@@ -45,6 +45,45 @@ def predicate(ctx: Context) -> bool:
...
+    If the command was invoked in a context that was blacklisted, the member is either
+    redirected to the `redirect` channel that was passed (default: #bot-commands) or simply
+    told that they're not allowed to use this particular command (if `None` was passed).
+
+    The blacklist can be overridden through the roles specified in `override_roles`.
+    """
+    def predicate(ctx: Context) -> bool:
+        """Check if command was issued in a blacklisted context."""
+        not_blacklisted = not in_whitelist_check(ctx, channels, categories, roles, fail_silently=True)
+        overridden = in_whitelist_check(ctx, roles=override_roles, fail_silently=True)
```
MarkKoz, line 75: What do you think about making this a separate decorator like `@override_blacklist`? It would work by catching the `InWhitelistCheckFailure` raised and suppresing it. I feel like shoving so much into a single decorator makes the API confusing.

bot/decorators.py
```diff
@@ -45,6 +45,45 @@ def predicate(ctx: Context) -> bool:
...
+
+        success = not_blacklisted or overridden
+
+        if not success and not fail_silently:
+            raise InWhitelistCheckFailure(redirect)
```
MarkKoz, line 80: Yeah, this should be a new, distinct exception type.

MarkKoz (changes requested):
> The check uses the in_whitelist_check check internally

I don't have a problem with that. It means fewer redundancies. The trace logs in the check will be a bit misleading, but they're just trace logs, so I don't mind.

bot/decorators.py
mbaruh: So just copy the code from the lib with a different class name? I guess that's what's been bugging me.

MarkKoz: You can make it a base class and make the whitelist/blacklist types subclasses with empty class bodies.

mbaruh: Hmm the whitelist type is a part of the dpy library, I can't really edit that.

MarkKoz: It's not. It's right here https://github.com/python-discord/bot/blob/main/bot/utils/checks.py#L23

mbaruh: Oh, woops

bot/decorators.py
mbaruh: I think it sound ok, I just need to figure out how to do it lol

MarkKoz (approved):

bot/decorators.py
HassanAbouelela: It seems this hasn't been addressed. I have no opinion on the matter, but everything else looks good to me.

mbaruh: @HassanAbouelela Yeah, we talked about it in the server and decided it's more trouble than it's worth, since checks are handled like normal decorators. This would mean this would need to be either rewritten as non-dpy checks, or monkey-patch an attribute to be inspected in another part of the code.

HassanAbouelela: For sir-lancebot, we do monkey patching like [this](https://github.com/python-discord/sir-lancebot/blob/1e9a88f5b59cf7d1c4e24f48cc09aa98e5f6499b/bot/utils/decorators.py#L206), and it isn't too bad. I do agree this isn't necessary here, as blacklist is not a global check, and can be directly modified by the caller.

HassanAbouelela (approved):

Issue #1502, opened 2021-04-05
Disable TRACE log levels by default, with ability to selectively enable them for certain loggers


Numerlor:
The trace log level should be used for verbose logging, however because of the current setup where every logger always outputs trace logs, it is hard to properly utilize the log level without adding noise to the logs while that module isn't being worked on.

The root logger should be set to the debug level, with trace logs only being enabled for loggers with names that were specified by the user. These names can be listed in an env var, with the ability to also set the root logger to the trace level.

MarkKoz:
I agree that trace logs can get in the way of development sometimes. This is aggravated by logs coming from event handlers, like the ones in the code block detection cog.

Several people, including myself, were in favour of using an environment variable/config field to set the log level on a per-module basis.

PR #1529, opened 2021-04-15
Make trace logging optional and allow selective enabling

Numerlor:
closes: #1502

Implements the `BOT_TRACE_LOGGERS` env var that can be used to set which loggers should log trace logs instead of defaulting to logging them all in a development environment.

The env var follows the format of a comma delimited list prefixed with either a `*` or a `!`. 

- When neither of the prefixes are present, the listed loggers are set to the trace level. 
- If the `!` prefix is used, all of the loggers except the listed ones are set to the trace level.
- Otherwise if `*` is used, the root logger is set to the trace level.

I removed the option to use the `COLOREDLOGS_LOG_LEVEL` to change the coloredlogs level as I don't think it's something that was really used as the levels are controlled by our loggers instead of the different handlers needing to filter them, and it made the code a bit more complicated, but I'm open to changing it back if there are any opinions against it.

I left the `None` value (unset env var) to not enable any trace logs as it made sense with the `None`'s general purpose, although I'm not sure if it is a good default, thoughts?

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -199,6 +199,7 @@ class Bot(metaclass=YAMLGetter):
     prefix: str
     sentry_dsn: Optional[str]
     token: str
+    trace_loggers: Optional[str]
 
 
 class Redis(metaclass=YAMLGetter):
--- a/bot/log.py
+++ b/bot/log.py
@@ -20,7 +20,6 @@ def setup() -> None:
     logging.addLevelName(TRACE_LEVEL, "TRACE")
     Logger.trace = _monkeypatch_trace
 
-    log_level = TRACE_LEVEL if constants.DEBUG_MODE else logging.INFO
     format_string = "%(asctime)s | %(name)s | %(levelname)s | %(message)s"
     log_format = logging.Formatter(format_string)
 
@@ -30,7 +29,6 @@ def setup() -> None:
     file_handler.setFormatter(log_format)
 
     root_log = logging.getLogger()
-    root_log.setLevel(log_level)
     root_log.addHandler(file_handler)
 
     if "COLOREDLOGS_LEVEL_STYLES" not in os.environ:
@@ -44,11 +42,9 @@ def setup() -> None:
     if "COLOREDLOGS_LOG_FORMAT" not in os.environ:
         coloredlogs.DEFAULT_LOG_FORMAT = format_string
 
-    if "COLOREDLOGS_LOG_LEVEL" not in os.environ:
-        coloredlogs.DEFAULT_LOG_LEVEL = log_level
+    coloredlogs.install(level=logging.TRACE, logger=root_log, stream=sys.stdout)
 
-    coloredlogs.install(logger=root_log, stream=sys.stdout)
-
+    root_log.setLevel(logging.DEBUG if constants.DEBUG_MODE else logging.INFO)
     logging.getLogger("discord").setLevel(logging.WARNING)
     logging.getLogger("websockets").setLevel(logging.WARNING)
     logging.getLogger("chardet").setLevel(logging.WARNING)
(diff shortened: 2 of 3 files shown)


kwzrd (commented):
Looks & works great. I agree that having trace logs enabled globally by default is annoying, so this will be a great change. Let me share some thoughts.

Have you considered defining the config directly in the YAML? It would allow us to use the list structure, rather than a comma-delimited string. The local config is already git-ignored, so I don't think we would lose out on the convenience there. I think this would also allow us to make it a list always (empty by default), rather than consider None.

It may even allow us to drop support for the * and ROOT aliases since having an empty string in the config would already make the code grab the root logger. Though that's probably not obvious to the user unless you already know `logging` behaves like that.

I think there may be some merit in allowing the root aliases to be part of the list, rather than the sole entry. It seems unexpected that `a,a.b` would work to enable `a.c`, bot not `*,a,a.b` to enable `c`.

Numerlor:
> Have you considered defining the config directly in the YAML? It would allow us to use the list structure, rather than a comma-delimited string. The local config is already git-ignored, so I don't think we would lose out on the convenience there. I think this would also allow us to make it a list always (empty by default), rather than consider None.

I considered it at first, but saw the env var as something that's easier to change for the user with the config being mostly static

> It may even allow us to drop support for the * and ROOT aliases since having an empty string in the config would already make the code grab the root logger...

This already happens if the env var is present but not set, but I decided to just ignore the empty string to not cause confusion and left the root logger to the explicit names

> I think there may be some merit in allowing the root aliases to be part of the list, rather than the sole entry. It seems unexpected that `a,a.b` would work to enable `a.c`, bot not `*,a,a.b` to enable `c`.

This sounds good, one of my goals was to keep this simple as it's only dev facing but the tiny bit of complexity is worth it here as  currently the whole value has to overwritten to change the root logger

MarkKoz:
> I think there may be some merit in allowing the root aliases to be part of the list, rather than the sole entry. It seems unexpected that a,a.b would work to enable a.c, bot not *,a,a.b to enable c.

I actually think it would be a good opportunity to implement a blacklist. That is, if  `*` is present, then all other specified loggers have trace disabled. If `*`, then all specified loggers have trace enabled. A blacklist works well with my flow since there are only a couple modules that have annoying trace logs; the rest don't bother me most of the time. A blacklist would mean I don't have to change the value of the env var every time I test a different feature.

Numerlor:
> I actually think it would be a good opportunity to implement a blacklist. That is, if `*` is present, then all other specified loggers have trace disabled. If `*`, then all specified loggers have trace enabled. A blacklist works well with my flow since there are only a couple modules that have annoying trace logs; the rest don't bother me most of the time. A blacklist would mean I don't have to change the value of the env var every time I test a different feature.

That also sounds useful but now I don't really want it to go back to having to remove the whole env var to enable it globally, maybe we could ditch `ROOT` and then use some tokens as prefixes? For example `*` to set the root logger and `!` to turn it into a blacklist, having them separate could also save some confusion

MarkKoz:
I made a typo. I meant that if `*` is _not_ present, then specified loggers will be set to trace. To enable it globally all you'd need is a lone `*`.

Examples:

* `*,a.b.c`: enable trace for all except `a.b.c`
* `*`: enable trace for all
* `a.b.c`: enable trace only for `a.b.c`

MarkKoz:
But yeah, I suppose making it a separate token is fine.

Numerlor:
> I made a typo. I meant that if `*` is _not_ present, then specified loggers will be set to trace. To enable it globally all you'd need is a lone `*`.

Yep, but to enable it, the old logger names would be lost from the env var which could be annoying when doing it temporarily.

So we'd have something like
`a` - enables for module `a`
`*,a` or `*a` - enables for everything
`!,a` or `!,a` - disables for a

MarkKoz:
That looks good.

MarkKoz (approved):

kwzrd (approved):

vcokltfre (approved):
LGTM :shipit:

HassanAbouelela (approved):

PR #1531, opened 2021-04-17
Redirect output to bot-commands channel for the !eval command

### LINKED ISSUE #1498: Show the result of !eval in #bot-commands when it is used in a channel where it is not allowed.
## Description
When the command `!help` is used in a channel such as `#python-general`, Python prints out the result in `#bot-commands`. However, when a command like `!eval` is used, Python simply tells the user that they can't do that here. It would be nice if Python showed the results of `!eval` in `#bot-commands` the same way it does for `!help`.

## Reasoning
It would enhance the user experience because they wouldn't have to type it all over again.

## Proposed Implementation
The same way Python currently prints the output of `!help` in `#bot-commands` when it's used in a channel where that's not allowed, it should do the same with `!eval`.

## Alternate Ideas
The only alternate implementation that I can think of is the current one, where someone is forced to do it again in `#bot-commands.`

## Would you like to implement this yourself?
- [ ] I'd like to implement this feature myself
- [x] Anyone can implement this feature

ToxicKidz:
## Relevant Issues
Closes #1498

## Description
This PR makes it so that if you do the `!e` command, and you use it in a blacklisted channel (i.e `#python-general`), if you don't have a bypassing role, then the output will be redirected.

## Screenshots
image

### DIFF
--- a/bot/decorators.py
+++ b/bot/decorators.py
@@ -111,12 +111,15 @@ def redirect_output(
     destination_channel: int,
     bypass_roles: t.Optional[t.Container[int]] = None,
     channels: t.Optional[t.Container[int]] = None,
-    categories: t.Optional[t.Container[int]] = None
+    categories: t.Optional[t.Container[int]] = None,
+    ping_user: bool = True
 ) -> t.Callable:
     """
     Changes the channel in the context of the command to redirect the output to a certain channel.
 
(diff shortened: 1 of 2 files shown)


doublevcodes (approved):
Looks good!

Xithrius (approved):
Tested, it works well.

ChrisLovering (approved):
Looks good! nice one :)

PR #1532, opened 2021-04-18
!superstarify default time

### LINKED ISSUE #1515: Default duration for !superstar command if no duration is specified
## Context
The `!superstar` command currently requires a duration to be specified when invoked. 

Often we want to change someone's nickname from something very noisy or that takes up a lot of room until the user can respond and change it themselves. The duration in these cases is not usually that important as we want it to be a quick change.

## Proposal
It would be useful if there was a short default duration for this command. Something like 5 or 15 minutes. It would function similarly to the `!mute` command where if no duration is given, it opts for a 1 hour duration.

doublevcodes:
### Relevant issues
Closes #1515 

### Description
Makes the `duration` parameter optional giving it a default value of 1 hour as suggested in the issue

### DIFF
--- a/bot/exts/moderation/infraction/superstarify.py
+++ b/bot/exts/moderation/infraction/superstarify.py
@@ -135,7 +135,7 @@ class Superstarify(InfractionScheduler, Cog):
         if await _utils.get_active_infraction(ctx, member, "superstar"):
             return
 
-        # Set the duration to 1 hour if none was provided
+        # Set to default duration if none was provided.
         duration = duration or await Duration().convert(ctx, SUPERSTARIFY_DEFAULT_DURATION)
 
         # Post the infraction to the API
--- a/bot/exts/moderation/infraction/superstarify.py
+++ b/bot/exts/moderation/infraction/superstarify.py
@@ -136,8 +136,9 @@ class Superstarify(InfractionScheduler, Cog):
             return
 
         # Set the duration to 1 hour if none was provided
-        duration = datetime.datetime.now() + datetime.timedelta(hours=1)
-
+        if not duration:
+            duration = datetime.datetime.now() + datetime.timedelta(hours=1)
+            
         # Post the infraction to the API
         old_nick = member.display_name
         infraction_reason = f'Old nickname: {old_nick}. {reason}'
--- a/bot/exts/moderation/infraction/superstarify.py
+++ b/bot/exts/moderation/infraction/superstarify.py
@@ -138,7 +138,7 @@ class Superstarify(InfractionScheduler, Cog):
         # Set the duration to 1 hour if none was provided
         if not duration:
             duration = datetime.datetime.now() + datetime.timedelta(hours=1)
-            
+
         # Post the infraction to the API
         old_nick = member.display_name
         infraction_reason = f'Old nickname: {old_nick}. {reason}'
--- a/bot/exts/moderation/infraction/superstarify.py
+++ b/bot/exts/moderation/infraction/superstarify.py
@@ -109,7 +109,7 @@ class Superstarify(InfractionScheduler, Cog):
         self,
         ctx: Context,
         member: Member,
-        duration: Expiry,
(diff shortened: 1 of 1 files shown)


bot/exts/moderation/infraction/superstarify.py
```diff
@@ -134,6 +135,10 @@ async def superstarify(
         if await _utils.get_active_infraction(ctx, member, "superstar"):
             return
 
+        # Set the duration to 1 hour if none was provided
+        if not duration:
+            duration = datetime.datetime.now() + datetime.timedelta(hours=1)
```
RohanJnr, line 140:
```suggestion
        # Set the duration to 1 hour if none was provided
        duration = duration or datetime.datetime.utcnow() + datetime.timedelta(hours=1)
```

RohanJnr (changes requested):

bot/exts/moderation/infraction/superstarify.py
RohanJnr: also, it would be better to store the default duration as a constant

RohanJnr: like
```python
DEFAULT_DURATION = 1
```
```python
duration = duration or datetime.datetime.utcnow() + datetime.timedelta(hours=DEFAULT_DURATION)
```
in the file, as a constant

mbaruh:
I'm wondering if it would be better to do the same thing done to set the default duration of the mute command, meaning use the `Duration` converter to create the default time. This would ensure that if there is ever a change to the implementation of the converter, the behaviors won't diverge.

RohanJnr:
> I'm wondering if it would be better to do the same thing done to set the default duration of the mute command, meaning use the `Duration` converter to create the default time. This would ensure that if there is ever a change to the implementation of the converter, the behaviors won't diverge.

I thought of doing that, but the Duration().converter(...) needs to parse and convert the string `1h` to a timedelta object first and then add the date and timedelta object.
Whereas here, we can directly add the date and timedelta object

mbaruh:
> I thought of doing that, but the Duration().converter(...) needs to parse and convert the string `1h` to a timedelta object first and then add the date and timedelta object.
> Whereas here, we can directly add the date and timedelta object

True, but I don't really mind the extra operation. It's cheap and hidden anyway.

RohanJnr:
> > I thought of doing that, but the Duration().converter(...) needs to parse and convert the string `1h` to a timedelta object first and then add the date and timedelta object.
> > Whereas here, we can directly add the date and timedelta object
> 
> True, but I don't really mind the extra operation. It's cheap and hidden anyway.

alright then, we can use the Duration() class directly

bot/exts/moderation/infraction/superstarify.py
```diff
@@ -134,6 +135,9 @@ async def superstarify(
         if await _utils.get_active_infraction(ctx, member, "superstar"):
             return
 
+        # Set the duration to 1 hour if none was provided
+        duration = duration or datetime.datetime.utcnow() + datetime.timedelta(hours=1)
```
RohanJnr, line 139:
```suggestion
        duration = duration or await Duration().convert(ctx, "1h")
```
You can have the `1h` as a constant at the top of the file.

RohanJnr: make sure to import `Duration`

bot/exts/moderation/infraction/superstarify.py
```diff
@@ -134,6 +135,9 @@ async def superstarify(
         if await _utils.get_active_infraction(ctx, member, "superstar"):
             return
 
+        # Set the duration to 1 hour if none was provided
```
mbaruh, line 138: This comment is outdated
```suggestion
        # Set to default duration if none was provided
```

mbaruh (changes requested):

mbaruh (approved):
Thanks for contributing!

PR #1538, opened 2021-04-19
Suspend stream when revoked

### LINKED ISSUE #1506: Stop an in-progress stream if a user's role is revoked
If a user's streaming permission is removed via the revoke stream command, we want to stop any stream that may be in-progress.

If a user's video role drops off due to the normal schedule, then we don't need to do this.

`discord.ActivityType.streaming` shows whether a user is streaming.

This implementation should work even if the member don't have the video role anymore.

Concept approved by @MrHemlock

RohanJnr:
closes #1506 

Suspend a member's stream when the `!rstream` command (used to revoke stream permissions).

This is done by moving the member to the AFK voice channel and then back to the original voice channel.

### DIFF
--- a/bot/exts/moderation/stream.py
+++ b/bot/exts/moderation/stream.py
@@ -1,5 +1,6 @@
 import logging
 from datetime import timedelta, timezone
+from operator import itemgetter
 
 import arrow
 import discord
@@ -8,8 +9,9 @@ from async_rediscache import RedisCache
 from discord.ext import commands
 
 from bot.bot import Bot
-from bot.constants import Colours, Emojis, Guild, Roles, STAFF_ROLES, VideoPermission
+from bot.constants import Colours, Emojis, Guild, MODERATION_ROLES, Roles, STAFF_ROLES, VideoPermission
 from bot.converters import Expiry
+from bot.pagination import LinePaginator
 from bot.utils.scheduling import Scheduler
 from bot.utils.time import format_infraction_with_duration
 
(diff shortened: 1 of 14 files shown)


ChrisLovering (approved):
Looks good to me, tested too and works exactly like I had imagined :D

Shivansh-007 (approved):
Tested it! Works perfectly!

bot/exts/moderation/stream.py
```diff
@@ -70,6 +70,27 @@ async def _reload_tasks_from_redis(self) -> None:
...
+
+        if not voice_state:
+            return
+
+        # If the user is streaming.
+        if voice_state.self_stream:
+            # End user's stream by moving them to AFK voice channel and back.
+            original_vc = voice_state.channel
+            await member.move_to(self.bot.get_channel(Channels.afk_voice))
+            await member.move_to(original_vc)
```
kwzrd, line 85: You can get None from `get_channel` if the cache isn't ready yet. Usually this problem is circumvented by awaiting the `wait_until_guild_available` event on the bot instance. Although I guess it's not much of a concern here since it's only called from a command.

From the docs, it looks like if you do get None here it will kick the user from voice, and L85 will fail since you cannot reconnect them.

ChrisLovering: Yea, we don't want to kick them from voice with this, so adding a wait until guild available makes sense.

RohanJnr: I think `member.voice` also uses the cache, hence I will wait for the cache ready even before I call `member.voice`

config-default.yml
```diff
@@ -206,6 +206,7 @@ guild:
         staff_announcements:    &STAFF_ANNOUNCEMENTS    464033278631084042
 
         # Voice Channels
+        afk_voice:                          756327105389920306
```
Akarys42, line 209: [`Guild.afk_channel`](https://discordpy.readthedocs.io/en/latest/api.html#discord.Guild.afk_channel) provides the AFK channel of the guild already.

Akarys42 (changes requested):
Works like a charm, thanks a lot! There is just a tiny change that I'd like to see.

Akarys42 (approved):
LGTM! Happy to merge as soon as linting is fixed.

Akarys42 (commented):
Refresh

Issue #1473, opened 2021-03-18
Move static content site commands to tags


Numerlor:
The site commands tools, resources, help and home construct embeds from mostly static strings, with only the site url being variable. The url is already referenced in other tags so this won't bring any additional maintenance in the case it does change for some reason. 
Moving them to tags will also allow the more direct fuzzy matching and per channel cooldowns to take effect.
With the other commands removed the only left will be the rules command; I think we should move it to the Information or Utils cog and get rid of the Site cog

MarkKoz:
I agree. That would be a good refactor.

Numerlor:
Which commands do we want moved?
Looking at it there's the site home/about and site help commands, which don't looks like they'd fit a tag, at least if the names are kept. But I don't think they get much use, what do the usage stats look like for them?

MarkKoz:
Last 6 months:
```
site_resources  4 K
site_rules  4 K
site_tools  84
site_home   72
site_faq    43
site_help   10
```

I don't see why they all can't be tags (except rules). I guess the names should be like `sitefaq`, `sitehome`, etc.

Numerlor:
My main concern here is `site-help`, as a tag I don't think the name really fits it and it may be used in a fashion similar to how the ask tag was used.

ChrisLovering:
I think each of them could be ported across to tags, using their root aliases. I rarely ever see people doing `!site resources`, instead they opt for `!resources`, similar story for the others.

As for the site-help tag specifically, I brought it up with other core devs, and we agreed that it should be dropped entirely.

PR #1539, opened 2021-04-19
Move static content site commands to tags

Numerlor:
Creates the `resources`, `faq` and `tools` tags from the corresponding site commands and the `site` tag from the `home`/`about` command. The `rules` command was moved to the `Information` cog, and `site help` was removed along with the cog.

closes: #1473


Numerlor:
I've only moved the content to the bodies of the tags so far, do we also want to have a "title" in the tags?

ToxicKidz:
> do we also want to have a "title" in the tags?

How would this work? Currently we don't have embed titles for tags, unless you mean we should make the start of the files `**title**`, which I think would be a good idea.

jb3:
We could use a similar approach to the branding repo and put the tag metadata in the YAML frontmatter of the markdown documents.

For example:
```md
---
title: How to get rich, fast.
---
1. Buy stonks
2. Watch stonks
3. ????
4. Profit
```

Numerlor:
> We could use a similar approach to the branding repo and put the tag metadata in the YAML frontmatter of the markdown documents.
> 
> For example:
> 
> ```lisp
> ---
> title: How to get rich, fast.
> ---
> 1. Buy stonks
> 2. Watch stonks
> 3. ????
> 4. Profit
> ```

I was only planning to do it with a bold text at the start like in ToxicKidz' comment, but implementing this shouldn't be too hard and will give more flexibility around tags in general so I'll do that.
Should the URLs also be moved to the title? I think it is a bit clearer inline but it may be worth it to have them in both places

jb3:
> I was only planning to do it with a bold text at the start like in ToxicKidz' comment, but implementing this shouldn't be too hard and will give more flexibility around tags in general so I'll do that.

Thinking about it, that is likely out of scope for this specific PR. Maybe we group that in with #1545 so we can have richer tag data.

> Should the URLs also be moved to the title?

Up to you, I know that the URL embeds don't look too great, and I think that the inline hyperlinks looks fine.

Akarys42 (approved):
LGTM

Akarys42:
I don't think we should wait for tag grouping before moving along with this. It wouldn't change anything about the current functionality, we can group them later on when we merge the other PR.

Numerlor:
> I don't think we should wait for tag grouping before moving along with this. It wouldn't change anything about the current functionality, we can group them later on when we merge the other PR.

The PR is stalled because the group PR should also introduce the ability to pass in embed arguments through the tag file to be used as titles, after the issue is approved. I suppose it could also be done in a PR separate from the grouping functionality but I don't think it'd be worth complicating it just to merge this PR

bot/exts/info/information.py
```diff
@@ -475,6 +476,39 @@ async def json(self, ctx: Context, message: Message) -> None:
         """Shows information about the raw API response in a copy-pasteable Python format."""
         await ctx.invoke(self.raw, message=message, json=True)
 
+    @command(aliases=("rule",))
+    async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
+        """Provides a link to all rules or, if specified, displays specific rule(s)."""
+        rules_embed = Embed(title="Rules", color=Colour.blurple(), url=f"{URLs.site_schema}{URLs.site}/pages/rules")
```
onerandomusername, line 482: `og_blurple`

onerandomusername (approved):
Looks good, one detail.

Testing the current behavior with the soon-to-be-removed site command, and noticing, with buttons, being able to link tags and get a nice view like this but with tags would be awesome.

image

mbaruh:
@Numerlor since the tags PR is merged I removed the stalled label.

Numerlor:
> @Numerlor since the tags PR is merged I removed the stalled label.

should be good to go now

bot/exts/info/information.py
```diff
@@ -523,6 +525,40 @@ async def json(self, ctx: Context, message: Message) -> None:
         """Shows information about the raw API response in a copy-pasteable Python format."""
         await self.send_raw_content(ctx, message, json=True)
 
+    @command(aliases=("rule",))
+    async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
+        """Provides a link to all rules or, if specified, displays specific rule(s)."""
+        rules_embed = Embed(title="Rules", color=Colour.og_blurple(), url=f"{URLs.site_schema}{URLs.site}/pages/rules")
```
ChrisLovering, line 531: Setting the URL here won't work in dev for people using Docker, since Discord's API does validation on embed URLs and deems `http://web:8000/pages/rules` an invalid url.

I wonder if it's *just fine* to hardcode this particular embed url to the prod site

ChrisLovering (commented):
Tested and all works good, bar my one comment.

bot/exts/info/information.py
Numerlor: Don't see an issue with it if it may make using it locally a bit easier, some other places in the code hardcode the urls anyway, pushed in 1f2b603

ChrisLovering (approved):

PR #1546, opened 2021-04-20
Use Embed.timestamp for the !remind command

### LINKED ISSUE #1536: Use embed timestamp in reminder
When a user sets a reminder, a confirmation embed is sent, to inform the user that the reminder was successfully saved.

In this reminder, we currently include the time at which it is due to be sent. This time is in UTC.

I propose we refactor this part of the code to instead use the timestamp field in the Embed.
https://github.com/python-discord/bot/blob/f2970481e1cd8950dbbb2d3b871df68e90f5a932/bot/exts/utils/reminders.py#L85-L105
 This would mean that the time would automatically get converted to the user's current time zone by Discord.

ToxicKidz:
## Description
This PR makes it so that when you the use `!remind` command, the success embed sends the date in the embed's timestamp, and not in the footer. This is so that the time can update how discord formats it, and so that it will convert to your local timezone. This closes #1536.

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -100,7 +100,7 @@ class Reminders(Cog):
 
         if delivery_dt:
             # Reminder deletion will have a `None` `delivery_dt`
-            footer_str += ', Done at'
+            footer_str += ', Due'
             embed.timestamp = delivery_dt
 
         embed.set_footer(text=footer_str)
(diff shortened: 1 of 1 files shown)


kwzrd:
Here is the current output:

image

I think that's pretty good, if grammatically a little awkward. I think the original wording ~

image

~ reads somewhat better to me, but maybe it's not as direct in what it's trying to convey. What do you think?

ToxicKidz:
I wasn't sure if `Due •10/20/21`, but now thinking about it, it looks fine.

kwzrd:
Yeah, perhaps what you have will be the easiest to understand. I don't feel strongly either way.

ChrisLovering (approved):

Akarys42 (approved):
Lgtm, thanks! :shipit:

PR #1551, opened 2021-04-27
Fixed the code snippets limits

dolphingarlic:
- Made the line limit 15 (including lines between two snippets)
- Halved the char limit to be 1000

### DIFF
--- a/bot/exts/info/code_snippets.py
+++ b/bot/exts/info/code_snippets.py
@@ -8,6 +8,7 @@ from discord import Message
 from discord.ext.commands import Cog
 
 from bot.bot import Bot
+from bot.constants import Channels
 from bot.utils.messages import wait_for_deletion
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


vcokltfre:
I'd like to suggest before this gets reviewed that instead of halving the limit which may cut off stuff people want to see, if it gets past 1k characters the output is redirect to #bot-commands instead

jb3 (approved):

jb3:
Whoops, approved, but I agree with @vcokltfre's comment.

vcokltfre (approved):
Looks great!

ChrisLovering (changes requested):
You can get None from `bot. get_channel` if the cache isn't ready yet. Usually this problem is circumvented by awaiting the wait_until_guild_available event on the bot instance.

ToxicKidz:
I don't think you'd receive the message event if the cache wasn't ready yet.

ChrisLovering:
> I don't think you'd receive the message event if the cache wasn't ready yet.

Are you sure about that? I thought message events are still emitted while the cache is filling?

ToxicKidz:
I guess it would be fine to just wait until the guild is ready just in case.

vcokltfre:
> Are you sure about that? I thought message events are still emitted while the cache is filling?

Yep, message events can still be received before the ready event is received

ChrisLovering (approved):

PR #1552, opened 2021-04-27
Use a specific error message when a warned user isn't in the guild

Akarys42:
This commit changes slightly how the warn, kick and mute commands to take a fetched member as their argument and to return a little error message if the user isn't in the guild rather than showing the whole help page.

**Previously:**
image
(note that help message of the mute command is way longer)

**Now:**
image


mbaruh (approved):
:shipit:

ChrisLovering (approved):
🚢

PR #1555, opened 2021-04-29
Code Snippets: Fix Error Handling for HTTP Request Errors

### LINKED ISSUE #1553: Code snippet cog doesn't handle 404 errors
Sentry Issue: [BOT-Z4](https://sentry.io/organizations/python-discord/issues/2368344750/?referrer=github_integration)

The following error occurs when trying to fetch from a non-exisiting repository:

```
Failed to fetch code snippet from https://api.github.com/repos/fake/link/branches. HTTP Status: 404. Message: 404, message='Not Found', url=URL('https://api.github.com/repos/fake/link/branches').
```

It should be handled by the cog and passed silently.

### LINKED ISSUE #1554: Code Snippets: Check if API Response is None
Sentry Issue: [BOT-Z7](https://sentry.io/organizations/python-discord/issues/2369621285/?referrer=github_integration)

When `_fetch_response` catches an exception, it returns `None`. The code that uses the response does not check if it's `None`, leading to the error below. I think it's only used in `_snippet_to_codeblock`, so a `None` check in there should suffice.

```
AttributeError: 'NoneType' object has no attribute 'splitlines'
  File "discord/client.py", line 343, in _run_event
    await coro(*args, **kwargs)
  File "bot/exts/info/code_snippets.py", line 231, in on_message
    snippet = await handler(**match.groupdict())
  File "bot/exts/info/code_snippets.py", line 92, in _fetch_github_snippet
    return self._snippet_to_codeblock(file_contents, file_path, start_line, end_line)
  File "bot/exts/info/code_snippets.py", line 180, in _snippet_to_codeblock
    split_file_contents = file_contents.splitlines()

```

MarkKoz:
Fixes #1554 
Fixes #1553

Ignore 404s and avoid returning None as the response when there's an exception.

### DIFF
--- a/bot/exts/info/code_snippets.py
+++ b/bot/exts/info/code_snippets.py
@@ -1,6 +1,7 @@
 import logging
 import re
 import textwrap
+from typing import Any
 from urllib.parse import quote_plus
 
 from aiohttp import ClientResponseError
@@ -43,7 +44,7 @@ class CodeSnippets(Cog):
     Matches each message against a regex and prints the contents of all matched snippets.
     """
 
-    async def _fetch_response(self, url: str, response_format: str, **kwargs) -> str:
+    async def _fetch_response(self, url: str, response_format: str, **kwargs) -> Any:
         """Makes http requests using aiohttp."""
         async with self.bot.http_session.get(url, raise_for_status=True, **kwargs) as response:
             if response_format == 'text':
@@ -61,7 +62,7 @@ class CodeSnippets(Cog):
                 ref = possible_ref['name']
                 file_path = path[len(ref) + 1:]
                 break
-        return (ref, file_path)
+        return ref, file_path
 
     async def _fetch_github_snippet(
         self,
@@ -145,8 +146,8 @@ class CodeSnippets(Cog):
         repo: str,
         ref: str,
         file_path: str,
-        start_line: int,
-        end_line: int
+        start_line: str,
+        end_line: str
     ) -> str:
         """Fetches a snippet from a BitBucket repo."""
         file_contents = await self._fetch_response(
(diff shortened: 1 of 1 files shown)


bot/exts/info/code_snippets.py
```diff
@@ -43,16 +44,13 @@ class CodeSnippets(Cog):
     Matches each message against a regex and prints the contents of all matched snippets.
     """
 
-    async def _fetch_response(self, url: str, response_format: str, **kwargs) -> str:
+    async def _fetch_response(self, url: str, response_format: str, **kwargs) -> Any:
```
ChrisLovering, line 47: Would a return type of `Union[str, dict]` make more sense here?

bot/exts/info/code_snippets.py
```diff
@@ -228,8 +226,16 @@ async def on_message(self, message: Message) -> None:
 
             for pattern, handler in self.pattern_handlers:
                 for match in pattern.finditer(message.content):
-                    snippet = await handler(**match.groupdict())
-                    all_snippets.append((match.start(), snippet))
+                    try:
+                        snippet = await handler(**match.groupdict())
+                        all_snippets.append((match.start(), snippet))
+                    except ClientResponseError as error:
+                        error_message = error.message  # noqa: B306
```
ChrisLovering, line 233: Out of interest, what was the reason for using `error.message` here with a noqa, instead of `str(error)`?

MarkKoz: Because it wouldn't consist solely of the message

https://github.com/aio-libs/aiohttp/blob/v3.7.4/aiohttp/client_exceptions.py#L87-L92

It's on a separate line cause the noqa wasn't working when it was in the f-string for some reason.

bot/exts/info/code_snippets.py
MarkKoz: No. Technically JSON is not limited to objects (or arrays) as the root element. Plus it's an internal helper method, so I don't feel being extremely accurate is necessary, especially considering the caller literally has to pass "json" or "text" to the function.

ChrisLovering (approved):

Xithrius (approved):
Tested, this works very well.

PR #1556, opened 2021-04-29
feat: Use embed timestamp in modpings off

### LINKED ISSUE #1547: Use embed timestamp in mod pings off
When a mods turns off mod pings, a confirmation is sent to inform the user that their pings have successfully been turned off.

In this confirmation, we currently include the time at which it is due to be sent, this time is in UTC.

I propose we refactor this part of the code to instead use an Embed, with a the timestamp field.
https://github.com/python-discord/bot/blob/ce819ade482e82ecbc474bce5fb8ac9dd8b37b40/bot/exts/moderation/modpings.py#L107
This would mean that the time would automatically get converted to the user's current time zone by Discord.

ToxicKidz:
## Relevant Issues
Closes #1547 

## Description
This PR makes it so that when the bot sends the mod pings off embed, it uses and embed timestamp to show when the role will be given back, so that it can be converted to local readable time, instead of using an ISO format.

### DIFF
--- a/bot/exts/moderation/modpings.py
+++ b/bot/exts/moderation/modpings.py
@@ -105,7 +105,7 @@ class ModPings(Cog):
         self._role_scheduler.schedule_at(duration, mod.id, self.reapply_role(mod))
 
         embed = Embed(timestamp=duration, colour=Colours.bright_green)
-        embed.set_footer(text="Moderators role has been removed", icon_url=Icons.green_checkmark)
+        embed.set_footer(text="Moderators role has been removed until", icon_url=Icons.green_checkmark)
         await ctx.send(embed=embed)
 
     @modpings_group.command(name='on')
--- a/bot/exts/moderation/modpings.py
+++ b/bot/exts/moderation/modpings.py
@@ -3,11 +3,11 @@ import logging
 
 from async_rediscache import RedisCache
 from dateutil.parser import isoparse
-from discord import Member
+from discord import Embed, Member
 from discord.ext.commands import Cog, Context, group, has_any_role
 
 from bot.bot import Bot
-from bot.constants import Emojis, Guild, MODERATION_ROLES, Roles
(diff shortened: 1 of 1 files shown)


bot/exts/moderation/modpings.py
```diff
@@ -104,7 +104,9 @@ async def off_command(self, ctx: Context, duration: Expiry) -> None:
             self._role_scheduler.cancel(mod.id)
         self._role_scheduler.schedule_at(duration, mod.id, self.reapply_role(mod))
 
-        await ctx.send(f"{Emojis.check_mark} Moderators role has been removed until {until_date}.")
+        embed = Embed(timestamp=duration, colour=Colours.bright_green)
+        embed.set_footer(text="Moderators role has been removed", icon_url=Icons.green_checkmark)
```
ChrisLovering, line 108:
```suggestion
        embed.set_footer(text="Moderators role has been removed until", icon_url=Icons.green_checkmark)
```
I'm not sure what other's think, but I preferred the wording you have before, with `until` here. It adds context to the date.

mbaruh: We could also move the current footer to the embed's text, and just have "Until" in the footer.

ToxicKidz: > I'm not sure what other's think, but I preferred the wording you have before, with `until` here. It adds context to the date.

I removed it by accident, my bad.

mbaruh: Just adding back until seems ok, on second thought

ToxicKidz: Resolved in https://github.com/python-discord/bot/pull/1556/commits/5c1cf2b5fd94359d42d147ce8687dd27b05e193d.

ChrisLovering (approved):

Akarys42 (approved):

PR #1569, opened 2021-05-06
Prevent accidental addition of users to talentpool cache when using `!talentpool get_review`

SebastiaanZ:
## Bug Description
When asking for a review for a user that isn't currently nominated using the `!talentpool get_review <userid>` command, the user is added to the talentpool `watched_users` cache, causing their messages to be relayed to the talentpool watch channel.

## Steps to reproduce
Use `!talentpool get_review <user_id>`, where `<user_id>` is the ID of a user with no active nomination. The command will correctly reply that the user does not have an active nomination, but their ID will nonetheless be added as a key to the `defaultdict` used to keep track of watched users in the talentpool .

## Solution
Replace all regular getitem usages with `.get(<user_id>)`, as the `Reviewer` should never insert IDs using the regular `defaultdict` path.

### Additional note
I've replaced all occurrences of regular getitem access into the defaultdict, even those that are normally not reachable with the id of a user that's currently not nominated, to prevent a future refactor from accidentally introducing this bug again.

### DIFF
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -57,7 +57,7 @@ class Reviewer:
         """Schedules a single user for review."""
         log.trace(f"Scheduling review of user with ID {user_id}")
 
-        user_data = self._pool.watched_users[user_id]
+        user_data = self._pool.watched_users.get(user_id)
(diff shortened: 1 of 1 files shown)


Akarys42 (approved):
That's one deadly bug right here. Thanks seb!

HassanAbouelela (approved):

PR #1570, opened 2021-05-07
Add a persistent scheduler

mbaruh:
It is gradually becoming a more common pattern to use a combination of the Scheduler class with an async redis cache to persistent scheduled tasks across restarts. Both a Scheduler and a redis cache are initialized in a cog, and there's a function to reschedule tasks on load. It is currently used in the `ModPings` and `Stream` cogs, and I foresee it being used in a couple more places in the near future as well.

I created a class that abstracts that combo away, to create the interface of a scheduler which can be assumed to persist across restarts (within the limitations of persistence of a redis cache).

Additionally, doing so allows us to standardize the usage of unix timestamps.

## Choices and limitations
* This scheduler is limited to taking a single async function, which is used to schedule all tasks, represented by keys in the cache. I don't see an easy way to automatically reschedule those tasks when the persistent scheduler is created.
* At first I tried to make it subclass the Scheduler, but couldn't make it work cleanly, as I needed to change method signatures, and the immediate `schedule` method didn't make much sense in this context. So there's now a Scheduler attribute instead,
* Currently, the async function can only take the key as an argument, and the value in the cache is the timestamp in which the task should go off. This is a bit inflexible. It might be better to make the value a JSON string with a `'timestamp'` field, and another field for keyword arguments the async function can take (that way it might be possible to use it for the help channel system as well).

### DIFF
--- a/bot/exts/moderation/modpings.py
+++ b/bot/exts/moderation/modpings.py
@@ -1,15 +1,12 @@
 import datetime
 import logging
 
-from async_rediscache import RedisCache
-from dateutil.parser import isoparse
-from discord import Member
 from discord.ext.commands import Cog, Context, group, has_any_role
 
 from bot.bot import Bot
 from bot.constants import Emojis, Guild, MODERATION_ROLES, Roles
 from bot.converters import Expiry
-from bot.utils.scheduling import Scheduler
+from bot.utils.persistent_scheduling import PersistentScheduler
 
 log = logging.getLogger(__name__)
 
@@ -17,47 +14,40 @@ log = logging.getLogger(__name__)
 class ModPings(Cog):
     """Commands for a moderator to turn moderator pings on and off."""
 
-    # RedisCache[discord.Member.id, 'Naïve ISO 8601 string']
-    # The cache's keys are mods who have pings off.
-    # The cache's values are the times when the role should be re-applied to them, stored in ISO format.
-    pings_off_mods = RedisCache()
-
     def __init__(self, bot: Bot):
         self.bot = bot
-        self._role_scheduler = Scheduler(self.__class__.__name__)
+        self._role_scheduler = PersistentScheduler(self.__class__.__name__, self.reapply_role, bot)
 
         self.guild = None
         self.moderators_role = None
 
-        self.reschedule_task = self.bot.loop.create_task(self.reschedule_roles(), name="mod-pings-reschedule")
+        self.reschedule_task = self.bot.loop.create_task(self.normalize_roles(), name="mod-pings-reschedule")
 
-    async def reschedule_roles(self) -> None:
+    async def normalize_roles(self) -> None:
         """Reschedule moderators role re-apply times."""
         await self.bot.wait_until_guild_available()
+        await self._role_scheduler.wait_until_ready()
+
         self.guild = self.bot.get_guild(Guild.id)
         self.moderators_role = self.guild.get_role(Roles.moderators)
 
         mod_team = self.guild.get_role(Roles.mod_team)
         pings_on = self.moderators_role.members
-        pings_off = await self.pings_off_mods.to_dict()
 
         log.trace("Applying the moderators role to the mod team where necessary.")
         for mod in mod_team.members:
             if mod in pings_on:  # Make sure that on-duty mods aren't in the cache.
-                if mod in pings_off:
-                    await self.pings_off_mods.delete(mod.id)
-                continue
+                if mod.id in self._role_scheduler:
+                    await self._role_scheduler.delete(mod.id)
 
             # Keep the role off only for those in the cache.
-            if mod.id not in pings_off:
-                await self.reapply_role(mod)
-            else:
-                expiry = isoparse(pings_off[mod.id]).replace(tzinfo=None)
-                self._role_scheduler.schedule_at(expiry, mod.id, self.reapply_role(mod))
+            elif mod.id not in self._role_scheduler:
+                await self.reapply_role(mod.id)
 
-    async def reapply_role(self, mod: Member) -> None:
+    async def reapply_role(self, mod_id: int) -> None:
         """Reapply the moderator's role to the given moderator."""
-        log.trace(f"Re-applying role to mod with ID {mod.id}.")
+        log.trace(f"Re-applying role to mod with ID {mod_id}.")
+        mod = self.guild.get_member(mod_id)
         await mod.add_roles(self.moderators_role, reason="Pings off period expired.")
 
     @group(name='modpings', aliases=('modping',), invoke_without_command=True)
@@ -97,12 +87,10 @@ class ModPings(Cog):
         until_date = duration.replace(microsecond=0).isoformat()  # Looks noisy with microseconds.
         await mod.remove_roles(self.moderators_role, reason=f"Turned pings off until {until_date}.")
 
-        await self.pings_off_mods.set(mod.id, duration.isoformat())
-
         # Allow rescheduling the task without cancelling it separately via the `on` command.
         if mod.id in self._role_scheduler:
             self._role_scheduler.cancel(mod.id)
-        self._role_scheduler.schedule_at(duration, mod.id, self.reapply_role(mod))
+        await self._role_scheduler.schedule_at(duration, mod.id)
 
         await ctx.send(f"{Emojis.check_mark} Moderators role has been removed until {until_date}.")
 
@@ -117,10 +105,7 @@ class ModPings(Cog):
 
         await mod.add_roles(self.moderators_role, reason="Pings off period canceled.")
 
-        await self.pings_off_mods.delete(mod.id)
-
-        # We assume the task exists. Lack of it may indicate a bug.
-        self._role_scheduler.cancel(mod.id)
+        await self._role_scheduler.delete(mod.id)
 
         await ctx.send(f"{Emojis.check_mark} Moderators role has been re-applied.")
 
--- a/bot/exts/moderation/stream.py
+++ b/bot/exts/moderation/stream.py
@@ -1,18 +1,17 @@
 import logging
-from datetime import timedelta, timezone
+from datetime import timedelta
 from operator import itemgetter
 
 import arrow
 import discord
 from arrow import Arrow
-from async_rediscache import RedisCache
 from discord.ext import commands
 
 from bot.bot import Bot
 from bot.constants import Colours, Emojis, Guild, MODERATION_ROLES, Roles, STAFF_ROLES, VideoPermission
 from bot.converters import Expiry
 from bot.pagination import LinePaginator
-from bot.utils.scheduling import Scheduler
+from bot.utils.persistent_scheduling import PersistentScheduler
 from bot.utils.time import format_infraction_with_duration
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


MarkKoz (changes requested):

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,161 @@
...
+
+        If a task with `task_id` already exists, close `coroutine` instead of scheduling it. This
+        prevents unawaited coroutine warnings. Don't pass a coroutine that'll be re-used elsewhere.
+        """
+        self._scheduler.schedule_later(delay, task_id, self._to_schedule(task_id))
+
+        time = arrow.utcnow() + timedelta(seconds=delay)
+        await self.cache.set(task_id, time.timestamp())
+
+    def cancel_all(self) -> None:
```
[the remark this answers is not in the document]
mbaruh: I probably forgot to add it after turning the scheduler into an attribute.

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,161 @@
...
+    `delete` it.
+    The `in` operator is supported for checking if a task with a given ID is currently scheduled.
+
+    Any exception raised in a scheduled task is logged when the task is done.
+    """
+
+    def __init__(self, name: str, task_factory: SchedulerTaskFactory, bot: Bot):
+        name = f"{__name__}.{name}"
+
+        self._scheduler = Scheduler(name)
```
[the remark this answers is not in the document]
mbaruh: Ah, yes. The reason I did this is that I wanted both the scheduler and the cache to have the same name. Hopefully it's not too much of an issue. A way to around it is to add an optional namespace argument to the scheduler init, although I'm not sure if that's desirable.

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,161 @@
...
+        await self.cache.set(task_id, time.timestamp())
+
+    def cancel_all(self) -> None:
+        """Unschedule all known tasks."""
+        self.reschedule_task.cancel()
+        self.reschedule_task.add_done_callback(lambda _: self._scheduler.cancel_all())
+
+    async def _start_scheduler(self) -> None:
+        """Starts the persistent scheduler by awaiting the guild before rescheduling the tasks."""
+        await self.bot.wait_until_guild_available()
```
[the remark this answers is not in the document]
mbaruh: In what format? I can avoid keeping the bot instance as an attribute, and call the rescheduling separately from outside, but it doesn't seem to be better in terms of coupling. Otherwise, I need the rescheduling task object for the cancels and deletes to work correctly, as you mentioned above.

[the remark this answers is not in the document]
MarkKoz: > I can avoid keeping the bot instance as an attribute, and call the rescheduling separately from outside, but it doesn't seem to be better in terms of coupling.

Yes, that is exactly what I mean. The rescheduling function is part of the public interface anyway. It is better for coupling since it won't rely on `Bot` or even an event loop (for `create_task`).

bot/utils/persistent_scheduling.py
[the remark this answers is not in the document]
MarkKoz: > Hopefully it's not too much of an issue
The logger name will be really long since `__name__` is full qualified and two fully qualified names are being concatenated.

> A way to around it is to add an optional namespace argument to the scheduler init, although I'm not sure if that's desirable.

I'd be okay with it if it was really necessary, but I don't think it's important to keep the names the same. In fact, it may be a bad idea to use `__name__` for the cache, since it will cause cache misses if there's a refactor that moves modules around or renames them.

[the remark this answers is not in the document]
mbaruh: It would be nice to be able to refer to both using the same name for debugging purposes, although it's not a must.

bot/utils/persistent_scheduling.py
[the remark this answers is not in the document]
mbaruh: Just to be sure, you mean you want the rescheduling function to receive a bot instance and create the task there?
Or do you mean you want whatever instantiated the class to create the task? because then I still need an indicator inside the scheduler that the rescheduling finished.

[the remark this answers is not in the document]
MarkKoz: Originally, I meant that the user of the scheduler would call `reschedule_all_tasks` and keep track of the task itself, if needed. The responsibility of ensuring rescheduling is finished before attempting to delete something would be delegated to the user of the class.

However, on second thought, having to manage the task externally sounds inconvenient. I'd feel fine with passing an event loop to the scheduler so it can schedule task. As for waiting for the guild to be available, that could be delegated to the task factory given; not all tasks will necessarily rely on the guild anyway.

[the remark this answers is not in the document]
mbaruh: An alternative is to bring back the event object, and then the task can be created from outside.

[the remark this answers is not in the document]
MarkKoz: Well, what do you think about just passing an event loop?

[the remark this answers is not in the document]
mbaruh: Yeah, that's fine, just throwing an idea

mbaruh:
What do you think about the last point in the original post? It will allow for future flexibility, but it's not currently needed and complicates the class a bit. Then again, if we do need it, then adding it later will be a headache that will invalidate the caches.

bot/utils/persistent_scheduling.py
```diff
@@ -99,7 +99,8 @@ async def delete_all(self) -> None:
         self._reschedule_task.cancel()
         await self.cache.clear()
 
-        self._reschedule_task.add_done_callback(self._scheduler.cancel_all())
+        self._reschedule_task.cancel()
+        self._scheduler.cancel_all()
```
MarkKoz: Why is the cancel repeated twice? The way to do it with `add_done_callback` would've been to create a lambda that ignores the argument given.

However, while the exact mechanics of task cancellation still elude me, I'm not sure if the `add_done_callback` approach is even necessary; the exception is thrown on the very next cycle of the event loop.

bot/utils/persistent_scheduling.py
```diff
@@ -68,8 +68,10 @@ async def to_dict(self) -> t.Dict[CacheKey, Arrow]:
 
     async def reschedule_task(self, task_id: CacheKey) -> None:
         """Reschedule the task according to `task_factory` and `task_id`."""
-        time = (await self.get(task_id)).datetime
-        self._scheduler.schedule_at(time, task_id, coroutine=self._to_schedule(task_id))
+        time = await self.get(task_id)
+        if not time:
+            return
```
MarkKoz, line 73: Do you not think it's worth raising an error?

bot/exts/moderation/modpings.py
```diff
@@ -27,11 +27,11 @@ def __init__(self, bot: Bot):
     async def normalize_roles(self) -> None:
         """Reschedule moderators role re-apply times."""
         await self.bot.wait_until_guild_available()
-        await self._role_scheduler.wait_until_ready()
-
         self.guild = self.bot.get_guild(Guild.id)
         self.moderators_role = self.guild.get_role(Roles.moderators)
 
+        await self._role_scheduler.wait_until_ready()
+
```
MarkKoz, line 34: I don't see how moving it down after getting the role and guild makes a difference.

bot/exts/moderation/modpings.py
```diff
@@ -1,63 +1,55 @@
...
-                self._role_scheduler.schedule_at(expiry, mod.id, self.reapply_role(mod))
+            elif mod.id not in self._role_scheduler:
+                await self.reapply_role(mod.id)
 
-    async def reapply_role(self, mod: Member) -> None:
+    async def reapply_role(self, mod_id: int) -> None:
         """Reapply the moderator's role to the given moderator."""
-        log.trace(f"Re-applying role to mod with ID {mod.id}.")
+        log.trace(f"Re-applying role to mod with ID {mod_id}.")
+        await self.reschedule_task
```
MarkKoz, line 51: Doesn't this create a deadlock?

The scheduler calls `reapply_role` and it waits for `reschedule_task`. `reschedule_task` calls `normalize_roles`, which waits for the scheduler to be ready. However, the scheduler is not ready until all `reapply_role` are called. That's a cycle, is it not?

MarkKoz (changes requested):

MarkKoz:
> 
> 
> What do you think about the last point in the original post? It will allow for future flexibility, but it's not currently needed and complicates the class a bit. Then again, if we do need it, then adding it later will be a headache that will invalidate the caches.

I'm not sure, but I have a feeling this would be better saved for another PR, especially since this one is pretty close to being done already. It doesn't sound like a terrible idea - I did store JSON once in redis for silence channel overwrites, but that was a lot simpler (no arbitrary keys). I'd want opinions from other people.

bot/utils/persistent_scheduling.py
mbaruh: Yeah I don't know why I did that, removed in b8d58e0

bot/utils/persistent_scheduling.py
mbaruh: edec063

bot/exts/moderation/modpings.py
mbaruh: Yeah, I tried to be smart and tripped on my own fit. Untangled in 76e62b3

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,155 @@
...
+
+        if not timestamp:
+            return None
+        return Arrow.utcfromtimestamp(timestamp)
+
+    async def to_dict(self) -> t.Dict[CacheKey, Arrow]:
+        """
+        Get a dict representation of the currently scheduled tasks.
+
+        The dictionary contains the task_id's as keys, and the times in which the tasks should go off as values.
```
MarkKoz, line 64:
```suggestion
        The dictionary contains the task_id's as keys, and the times at which the tasks should run as values.
```

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,155 @@
...
+        self._reschedule_task.add_done_callback(lambda _: self._scheduler.cancel(task_id))
+
+    def cancel_all(self) -> None:
+        """Unschedule all known tasks."""
+        self._reschedule_task.cancel()
+        self._reschedule_task.add_done_callback(lambda _: self._scheduler.cancel_all())
+
+    async def _to_schedule(self, task_id: CacheKey) -> None:
+        """
+        Wraps the task to run, so that we can delete the key from the cache.
```
MarkKoz, line 148:
```suggestion
        Wraps the task factory, so that it can first delete the task from the cache.
```

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,155 @@
...
+        await self._reschedule_task  # Wait for rescheduling to finish to avoid race conditions.
+        self._scheduler.cancel(task_id)
+
+    async def delete_all(self) -> None:
+        """
+        Unschedule all known tasks and wipe the cache.
+
+        `cancel_all` remains untouched so that it can be used in instances such as cog unloads while preserving cache.
+        """
+        self._reschedule_task.cancel()
```
MarkKoz, line 101: Since it's in a coroutine, you can cancel the task and then await it to make sure it's cancelled before moving on to `cancel_all`.

MarkKoz: Alternatively, you could just call `self.cancel_call` which takes care of both of these steps. It seems fine to be able to move the cache clear before/after that call. Could also use `self.cancel` inside of `delete()`.

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,155 @@
...
+        if time.tzinfo is None:
+            time.replace(tzinfo=timezone.utc)
+        await self.cache.set(task_id, time.timestamp())
+
+    async def schedule_later(self, delay: t.Union[int, float], task_id: CacheKey) -> None:
+        """
+        Schedule `coroutine` to be executed after the given `delay` number of seconds and add it to the cache.
+
+        If a task with `task_id` already exists, close `coroutine` instead of scheduling it. This
+        prevents unawaited coroutine warnings. Don't pass a coroutine that'll be re-used elsewhere.
```
MarkKoz, line 129: I think the closing coroutine part is not relevant for this class since it relies on a factory rather than directly being given a coroutine.

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        Get a dict representation of the currently scheduled tasks.
+
+        The dictionary contains the task_id's as keys, and the times in which the tasks should go off as values.
+        """
+        dict_ = await self.cache.to_dict()
+        return {task_id: Arrow.utcfromtimestamp(timestamp) for task_id, timestamp in dict_.items()}
+
+    async def reschedule_task(self, task_id: CacheKey) -> None:
+        """Reschedule the task according to `task_factory` and `task_id`."""
```
MarkKoz, line 70: Please document what happens if the task is already scheduled.

MarkKoz: On second thought, what is even the use case of this? The task wouldn't exist in the cache unless it's already scheduled. Furthermore, `reschedule_all_tasks` takes care of rescheduling it on restart.

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,155 @@
...
+    the instance is suggested. The same name must be provided after a restart for the tasks to be preserved.
+
+    A coroutine must be provided in order to reschedule each of the keys contained in the cache at init.
+    The coroutine will take a single argument -- the ID that was used to create the Scheduler task in the first place.
+
+    Coroutines can be scheduled with `schedule_at` or `schedule_later`.
+    A unique ID is required to be given in order to keep track of the resulting Tasks.
+    Any scheduled task can be cancelled prematurely using `cancel` by providing the same ID used to schedule it, but
+    the task will be rescheduled from the cache on the next initialization. To remove the task completely, you can
+    `delete` it.
```
MarkKoz, line 30:
```suggestion
    the task will be rescheduled from the cache on the next initialization. To remove the task completely, use
    `delete` instead.
```

MarkKoz (changes requested):
Hello, I finally have time again to get around to reviews. The API sensible but there are still some things to address. I haven't run the code yet to test it out.

Xithrius:
@mbaruh status on this PR?

mbaruh:
@Xithrius It still lives, just taking me a bit to get to it.

Xithrius:
> @Xithrius It still lives, just taking me a bit to get to it.

Thank you for the update! :heart:

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,155 @@
...
+        for task_id, timestamp in tasks:
+            time = Arrow.utcfromtimestamp(timestamp)
+            self._scheduler.schedule_at(time.datetime, task_id, coroutine=self._to_schedule(task_id))
+
+    async def delete(self, task_id: CacheKey) -> None:
+        """
+        Unschedule the task identified by `task_id` and delete it from the cache.
+
+        `cancel` is left as-is in order to be able to pause execution until the next reschedule.
+        Log a warning if the task doesn't exist.
```
Bluenix2, line 88: This method doesn't do that?

bot/utils/persistent_scheduling.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def cancel(self, task_id: CacheKey) -> None:
+        """Unschedule the task identified by `task_id`. Log a warning if the task doesn't exist."""
+        # Wait for rescheduling to finish to avoid race conditions.
+        self._reschedule_task.add_done_callback(lambda _: self._scheduler.cancel(task_id))
+
+    def cancel_all(self) -> None:
+        """Unschedule all known tasks."""
+        self._reschedule_task.cancel()
+        self._reschedule_task.add_done_callback(lambda _: self._scheduler.cancel_all())
```
Bluenix2, line 144: Shouldn't these be in opposite order?

bot/exts/moderation/stream.py
```diff
@@ -21,54 +20,34 @@
...
-
-            revoke_time = Arrow.utcfromtimestamp(value)
-            log.debug(f"Scheduling {member} ({member.id}) to have streaming permission revoked at {revoke_time}")
-            self.scheduler.schedule_at(
-                revoke_time,
-                key,
-                self._revoke_streaming_permission(member)
-            )
+
+        guild = self.bot.get_guild(Guild.id)
```
Bluenix2, line 35: Isn't there a helper for this?

Xithrius:
@mbaruh may you resolve the conflicting files? Thanks!

PR #1572, opened 2021-05-07
Escape markdown in the edited message contents before getting their diff.

Qwerty-133:
This prevents embeds from looking like this:
Screenshot 2021-05-08 at 12 35 22 AM
when a user edits a message with the content

> a
> ```
> b
> ```

to

> c
> ```
> b
> ```

and instead displays a nicer embed:
Screenshot 2021-05-08 at 12 38 10 AM


jb3 (approved):
Good spot, nice.

PR #1574, opened 2021-05-09
Fix Ping Bugs On Localhost

### LINKED ISSUE #1407: Ping Command Bugs
There are a couple bugs in the ping command, that only manifest in local development.

## Bug 1: `gaierror: [Errno 11001] getaddrinfo failed`
This bug is pretty easy to explain. On many development bots, the site URL will have some sort of port (for example `pythondiscord.local:8000` or `web:8000`).  aioping.ping does not seem to handle that, and will throw a getadderinfo error. ([Full TB if needed](https://paste.pythondiscord.com/pahuherice.sql)). This issue does not affect the Python bot, as the URL does not contain a port.

The fix for this should be simple enough, you just need to extract netloc from the URL, and pass it to the ping function. One possible way to do that could be through using the builtin [urllib.parse.urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse).

## Bug 2: Negative Processing Time
Last I checked, my bot had not yet developed time-traveling abilities, which makes this output a bit strange:

Time Magic

The line that calculates that time is:
https://github.com/python-discord/bot/blob/888c427466b370b9ae51e47496e979c2b6faed0c/bot/exts/utils/ping.py#L36

I'm not too sure why this happens, but it happens consistently. My guess would be a difference between the internal timing for discord, and the internal timing of my system. This is also not a problem on the production server.

I propose that all negative times be replaced with a warning, or a 0, or something to signify a nonsensical result. If someone has a better explanation or a better fix, feel free to comment.

HassanAbouelela:
## Description
Closes #1407.

## Changes
- Improves parsing of site URL, to make it work locally and in prod. This has been tested in prod, locally, and on docker.
- Changes the output of the bot latency for when ping is < 0. This is usually caused by a desynced closk. Interestingly, the docker container is always a few seconds out of sync, even after syncing my system clock, and rebuilding the container. This seems like a docker problem though, so I'm not quite sure how to fix it.
- Added another fix where the ping utility is not available on Linux based OSes, if not running as root. Simplest solution was to just say it didn't work.

## Screenshots
The new error message when ping is < 0.

New error message

Error message for missing permissions.

Missing perms

### DIFF
--- a/bot/exts/utils/ping.py
+++ b/bot/exts/utils/ping.py
@@ -42,8 +42,12 @@ class Latency(commands.Cog):
 
         try:
             url = urllib.parse.urlparse(URLs.site_schema + URLs.site).hostname
-            delay = await aioping.ping(url, family=socket.AddressFamily.AF_INET) * 1000
-            site_ping = f"{delay:.{ROUND_LATENCY}f} ms"
+            try:
+                delay = await aioping.ping(url, family=socket.AddressFamily.AF_INET) * 1000
+                site_ping = f"{delay:.{ROUND_LATENCY}f} ms"
+            except OSError:
+                # Some machines do not have permission to run ping
+                site_ping = "Permission denied, could not ping."
 
         except TimeoutError:
             site_ping = f"{Emojis.cross_mark} Connection timed out."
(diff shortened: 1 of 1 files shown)


vcokltfre (approved):
Tested, works as expected

Note that site ping doesn't work on some systems due to requiring root to do ICMP messages, but the error is handled so that's fine!

ToxicKidz (approved):
:shipit:

Xithrius (approved):

PR #1577, opened 2021-05-10
Code snippet: support the two dots syntax

Akarys42:
Lines can be highlighted in GitHub using the `L00..L42` syntax, currently not supported by the regex. This commits adds it.


ToxicKidz (approved):
Works well :shipit:

Xithrius (approved):
Tested, works well.

PR #1587, opened 2021-05-14
Prioritize DM over channel message for voice verification ping.

### LINKED ISSUE #1448: Add DM to current #voice-verification ping
When a user joins the voice chat for the first time, they receive a ping in the `#voice-verification` channel.  This was done in an attempt to solve new users' confusion as to why they couldn't talk when joining the chat.  However, it does not seem to be having the desired effect, as there are still quite a few users still confused. 

The proposal now is to send a DM to the user with a link and explanation so that they have a lasting link and reminder on how to get verified.  Message requirements would be:

- A quick explanation of the system.
- A clickable link to the `#voice-verification` system.

Should be relatively quick and easy to implement.

RohanJnr:
Closes #1448


bot/exts/moderation/voice_gate.py
```diff
@@ -144,8 +155,13 @@ async def voice_verify(self, ctx: Context, *_) -> None:
...
                 log.warning(f"Got response code {e.status} while trying to get {ctx.author.id} Metricity data.")
-
-            await ctx.author.send(embed=embed)
+            try:
+                await ctx.author.send(embed=embed)
+            except discord.Forbidden:
+                log.info("Could not send user DM. Sending in voice-verify channel and scheduling delete.")
+                message = await ctx.send(embed=embed)
+                await asyncio.sleep(GateConf.voice_ping_delete_delay)
+                await message.delete()
```
ChrisLovering, line 164: Ah, it seems we don't need to do this since it already gets deleted here https://github.com/python-discord/bot/blob/67934bd582ae5a6fac459c02ba3fa6c17d085d1a/bot/exts/moderation/voice_gate.py#L207-L214

ChrisLovering (changes requested):

bot/exts/moderation/voice_gate.py
```diff
@@ -75,7 +81,7 @@ async def _delete_ping(self, member_id: int) -> None:
             log.trace(f"Voice gate reminder message for user {member_id} was already removed")
 
     @redis_cache.atomic_transaction
-    async def _ping_newcomer(self, member: discord.Member) -> bool:
+    async def _ping_newcomer(self, member: discord.Member) -> tuple:
```
MrHemlock, line 84: Docstring needs to be updated to reflect the new return type, detailing what the second element represents.

MrHemlock (changes requested):
Works fine with testing.  Only things that need tweaked are the doc string that I pointed out (possibly looking through the others just to make sure I didn't miss anything), and the thing that Chris pointed out.

ChrisLovering (approved):
Tested locally, all looks good to me 👍

kosayoda (approved):
Tests good to me!

PR #1590, opened 2021-05-16
Constants: use in-server emojis for incident

Akarys42:
The default config is currently referencing the emoji server versions, making it unable to work with the nomination archive automation


ChrisLovering (approved):

ToxicKidz (approved):
:shipit:

PR #1593, opened 2021-05-17
Fix pre-commit, since flake8 isn't a task

ChrisLovering:
The pre-commit file current runs flake8 via `poetry run task flake8` but flake 8 isn't a task, so it fails.


ToxicKidz (approved):
:shipit:

HassanAbouelela (approved):

Kronifer (approved):
Thanks, LGTM!

Issue #1585, opened 2021-05-12
Tag: async-await


vcokltfre:
This issue proposes a new tag that explains some basics about asyncio and concurrency. Essentially the content would be along the lines of:
- What's an event loop?
- How does it work?
- How do I use it?

If approved I'd like to write this myself

Objectivitix:
AMAZING! I WOULD LOVE THAT BECAUSE WE HAVE A LOT OF ASYNC/AWAIT QUESTIONS

ChrisLovering:
Go for it

vcokltfre:
Awesome, thanks!

PR #1594, opened 2021-05-17
feat: add async-await tag

vcokltfre:
Async await tag as approved in #1585 

image


bot/resources/tags/async-await.md
```diff
@@ -0,0 +1,28 @@
...
+To call an async function we can either `await` it, or run it in an event loop which we get from `asyncio`.
+
+To create a coroutine that can be used with asyncio we need to define a function using the async keyword:
+```py
+async def main():
+    await something_awaitable()
+```
+Which means we can call `await something_awaitable()` directly from within the function. If this were a non-async function this would have raised an exception like: `SyntaxError: 'await' outside async function`
+
+To run the top level async function from outside of the even loop we can get an event loop from `asyncio`, and then use that loop to run the function:
```
ToxicKidz, line 16:
```suggestion
To run the top level async function from outside of the event loop we can get an event loop from `asyncio`, and then use that loop to run the function:
```

bot/resources/tags/async-await.md
```diff
@@ -0,0 +1,28 @@
...
+```py
+from asyncio import get_event_loop
+
+async def main():
+    await something_awaitable()
+
+loop = get_event_loop()
+loop.run_until_complete(main())
+```
+Note that in the `run_until_complete()` where we appear to be calling `main()`, this does not execute the code in `main`, rather it returns a `Coroutine` object which is then handled and run by the event loop via `run_until_complete()`.
```
ToxicKidz, line 26:
```suggestion
Note that in the `run_until_complete()` where we appear to be calling `main()`, this does not execute the code in `main`, rather it returns a `coroutine` object which is then handled and run by the event loop via `run_until_complete()`.
```

bot/resources/tags/async-await.md
```diff
@@ -0,0 +1,28 @@
...
+
+To learn more about asyncio and it's use, see the [asyncio documentation](https://docs.python.org/3/library/asyncio.html).
```
ToxicKidz, line 28:
```suggestion
To learn more about asyncio and its use, see the [asyncio documentation](https://docs.python.org/3/library/asyncio.html).
```

ToxicKidz (changes requested):
Hey, this is a good tag, just a couple of suggestions. Is it worth mentioning the use of `asyncio.run` as well?

Xithrius (approved):
Yes.

ToxicKidz (approved):
:shipit:

PR #1602, opened 2021-05-20
Add the timeit command

### LINKED ISSUE #1025: Command to time code execution

ToxicKidz:
## Relevant Issues
Closes #1025 
Continues #1234 

## Description
This PR adds the `!timeit` command, which passes the `["-m", "timeit"]` arguments to the python process in snekbox.

##  Screenshots
image


Numerlor:
I don't think the current output redirecting approach is a good way to do it, as it heavily influences the timing of code that has its runtime on the lower end.
Somewhat hacky to persist the state across runs, but something like this works as the setup code and doesn't affect code that doesn't interact with stdout.
```py
import atexit
import sys
from collections import deque

if not hasattr(sys, "_setup_finished"):
    class Writer(deque):
        def __init__(self):
            super().__init__(maxlen=1)

        def write(self, string):
            if string.strip():
                self.append(string)

        def read(self):
            return self.pop()

        def flush(self):
            pass

    sys.stdout = Writer()

    def print_last_line():
        print(sys.stdout.read(), file=sys.__stdout__)

    atexit.register(print_last_line)
    sys._setup_finished = None
```
Or you could copy over a simplified version of timeit's main to normal code and wrap the part that executes code in the redirect instead.

The command should also provide a way to accept a setup (for example the first codeblock if multiple are detected) as it is fairly important for timing things.

Xithrius:
@ToxicKidz status?

Xithrius:
@ToxicKidz will you be able to finish this, or should I put it up for grabs?

ToxicKidz:
> @ToxicKidz will you be able to finish this, or should I put it up for grabs?

Sorry for the delay, I was stuck on the implementation for the PR and then I forgot about it. I'll try to finish this PR as soon as I can.

wookie184:
RE handling extra stdout from the user, why do we need to support suppressing it? I can't think of any reason timing code should need to send anything to stdout, so we'd probably be fine just having the command error if there's anything more than the `timeit` output imo.

ToxicKidz:
That makes sense, I guess I should just put a warning about printing anything to the stdout.

Xithrius:
@ToxicKidz What's your status on this PR?

ToxicKidz:
> @ToxicKidz What's your status on this PR?

Sorry, I got busy and forgot about this PR, again. I'll try to finish this PR by the end of the week.

ToxicKidz:
@wookie184 I was experimenting with Numerlor's implementation and found it didn't cause too much trouble to add, so I went through with it. It also stops the job from getting terminated if the `stdout` is too large, which is a plus.

bot/exts/utils/snekbox.py
```diff
@@ -336,6 +347,67 @@ async def eval_command(self, ctx: Context, *, code: str = None) -> None:
...
+        """
+        Parses the time from the end of the output given by timeit.
+
+        If an error happened, then it won't contain the time and instead proceed with regular formatting.
+        """
+        split_output = output.rstrip("\n").rsplit("\n", 1)
+        if len(split_output) == 2 and TIMEIT_OUTPUT_REGEX.fullmatch(split_output[1]):
+            return split_output[1], None
+
+        return await self.format_output(output)
```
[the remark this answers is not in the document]
ToxicKidz: This is still needed to determine whether an error happened, so we know whether we should format the error or just send the timings.

bot/exts/utils/snekbox.py
```diff
@@ -73,34 +112,37 @@ async def upload_output(self, output: str) -> Optional[str]:
...
-    def prepare_input(code: str) -> str:
+    def prepare_input(code: str) -> list[str]:
         """
         Extract code from the Markdown, format it, and insert it into the code template.
 
         If there is any code block, ignore text outside the code block.
         Use the first code block, but prefer a fenced code block.
         If there are several fenced code blocks, concatenate only the fenced code blocks.
+
+        Retrun a list of code blocks if any, otherwise return a list with a single string of code.
```
Bluenix2, line 123:
```suggestion
        Return a list of code blocks if any, otherwise return a list with a single string of code.
```

Bluenix2 (commented):
There's just one typo I found

wookie184 (changes requested):
Rerunning of code doesn't currently seem to be working, when I click the reaction to rerun code (for eval or timeit command) I get this error:
```py
bot_1        | 2022-01-15 13:06:01 | bot.exts.backend.error_handler | ERROR | Error executing command invoked by wookie#4017: !e 2
bot_1        | Traceback (most recent call last):
bot_1        |   File "/usr/local/lib/python3.9/site-packages/discord/ext/commands/core.py", line 167, in wrapped
bot_1        |     ret = await coro(*args, **kwargs)
bot_1        |   File "/bot/bot/decorators.py", line 136, in inner
bot_1        |     await func(self, ctx, *args, **kwargs)
bot_1        |   File "/bot/bot/exts/utils/snekbox.py", line 409, in eval_command
bot_1        |     await self.run_eval(ctx, code, format_func=self.format_output)
bot_1        |   File "/bot/bot/exts/utils/snekbox.py", line 367, in run_eval
bot_1        |     response = await self.send_eval(ctx, code, args=args, format_func=format_func)
bot_1        |   File "/bot/bot/exts/utils/snekbox.py", line 242, in send_eval
bot_1        |     results = await self.post_eval(code, args=args)
bot_1        |   File "/bot/bot/exts/utils/snekbox.py", line 102, in post_eval
bot_1        |     async with self.bot.http_session.post(url, json=data, raise_for_status=True) as resp:
bot_1        |   File "/usr/local/lib/python3.9/site-packages/aiohttp/client.py", line 1117, in __aenter__
bot_1        |     self._resp = await self._coro
bot_1        |   File "/usr/local/lib/python3.9/site-packages/aiohttp/client.py", line 625, in _request
bot_1        |     resp.raise_for_status()
bot_1        |   File "/usr/local/lib/python3.9/site-packages/aiohttp/client_reqrep.py", line 1000, in raise_for_status
bot_1        |     raise ClientResponseError(
bot_1        | aiohttp.client_exceptions.ClientResponseError: 400, message='Bad Request', url=URL('http://snekbox:8060/eval')
```
Not too sure what the cause of this is.

Other than that this seems pretty good

wookie184 (dismissed):
Good stuff 👌

bot/exts/utils/snekbox.py
```diff
@@ -36,6 +36,39 @@
...
+    sys.stdout = Writer()
+
+    def print_last_line():
+        if sys.stdout:
+            print(sys.stdout.read(), file=sys.__stdout__)
+
+    atexit.register(print_last_line)
+    sys._setup_finished = None
+{setup}
+"""
```
Numerlor, line 68: This should be documented more

bot/exts/utils/snekbox.py
```diff
@@ -279,25 +355,18 @@ async def get_code(self, message: Message) -> Optional[str]:
...
-    @guild_only()
-    @redirect_output(
-        destination_channel=Channels.bot_commands,
-        bypass_roles=EVAL_ROLES,
-        categories=NO_EVAL_CATEGORIES,
-        channels=NO_EVAL_CHANNELS,
-        ping_user=False
-    )
-    async def eval_command(self, ctx: Context, *, code: str = None) -> None:
+    async def run_eval(
```
Numerlor, line 358: I think a more generic command like run_job (and send_job for send_eval) would be better with it being expanded  and not strictly running a conventional eval

bot/exts/utils/snekbox.py
```diff
@@ -326,17 +391,81 @@ async def eval_command(self, ctx: Context, *, code: str = None) -> None:
...
+        """
+        Parses the time from the end of the output given by timeit.
+
+        If an error happened, then it won't contain the time and instead proceed with regular formatting.
+        """
+        split_output = output.rstrip("\n").rsplit("\n", 1)
+        if len(split_output) == 2 and TIMEIT_OUTPUT_REGEX.fullmatch(split_output[1]):
+            return split_output[1], None
+
+        return await self.format_output(output)
```
Numerlor, line 414: doesn't match typehint

Numerlor, line 414: To continue on the resolved conversation, why is this necessary at all? 

If the command runs properly, the output should be a single line we can directly send anyway. While passing it through format_output is a bit unnecessary, the changeable format func does complicate things considerably compared to just not doing it

From my understanding that is what's happening anyway as the split_output will only contain a single element if it's the timeit stdout

Numerlor (changes requested):

bot/exts/utils/snekbox.py
ToxicKidz: Yeah that makes sense. Do you think it would also make more sense to say `f"Your {job_name} job has failed"`, along with the other messages, rather than just `f"Your eval job has failed"?

bot/exts/utils/snekbox.py
```diff
@@ -248,18 +319,25 @@ async def continue_eval(self, ctx: Context, response: Message) -> Optional[str]:
                     timeout=10
                 )
 
-                code = await self.get_code(new_message)
+                code = await self.get_code(new_message, ctx.command)
```
Numerlor, line 322: While going through this I noticed we don't check for None on the code result which can cause an error down the line

bot/exts/utils/snekbox.py
```diff
@@ -36,6 +36,41 @@
...
+    def print_last_line():
+        if sys.stdout:
+            print(sys.stdout.read(), file=sys.__stdout__)
+
+    atexit.register(print_last_line)
+    sys._setup_finished = None
+{setup}
+"""
+
+TIMEIT_OUTPUT_REGEX = re.compile(r"\d+ loops, best of \d+: \d(?:\.\d\d?)? [mnu]?sec per loop")
```
Numerlor, line 72: No longer used

bot/exts/utils/snekbox.py
```diff
@@ -48,6 +83,8 @@
 REEVAL_EMOJI = '\U0001f501'  # :repeat:
 REEVAL_TIMEOUT = 30
 
+FormatFunc = Callable[[str], Awaitable[tuple[str, Optional[str]]]]
```
Numerlor, line 86: No longer used

bot/exts/utils/snekbox.py
```diff
@@ -222,14 +290,17 @@ async def send_eval(self, ctx: Context, code: str) -> Message:
...
-    async def continue_eval(self, ctx: Context, response: Message) -> Optional[str]:
+    async def continue_eval(
+        self, ctx: Context, response: Message, command: Command
+    ) -> tuple[Optional[str], Optional[list[str]]]:
         """
         Check if the eval session should continue.
 
-        Return the new code to evaluate or None if the eval session should be terminated.
+        If the code is to be evaluated, return the new code and the args if the command is the timeit command.
+        Otherwise return None if the eval session should be terminated.
```
Numerlor, line 303:
```suggestion
        If the code is to be re-evaluated, return the new code, and the args if the command is the timeit command.
        Otherwise return None as the first element if the eval session should be terminated.
```

It took me a bit to understand these two sentences, I believe the above suggestion makes it a bit better but maybe it could use a full reword. The code also never returns None alone so the second sentence should mention it'll return None for the code in the tuple if it's terminated

Numerlor (changes requested):

bot/exts/utils/snekbox.py
```diff
@@ -326,17 +388,67 @@ async def eval_command(self, ctx: Context, *, code: str = None) -> None:
...
+    @command(name="timeit", aliases=("ti",))
+    @guild_only()
+    @redirect_output(
+        destination_channel=Channels.bot_commands,
+        bypass_roles=EVAL_ROLES,
+        categories=NO_EVAL_CATEGORIES,
+        channels=NO_EVAL_CHANNELS,
+        ping_user=False
+    )
+    async def timeit_command(self, ctx: Context, *, code: str) -> str:
```
Numerlor, line 433:
```suggestion
    async def timeit_command(self, ctx: Context, *, code: str) -> None:
```

Numerlor (changes requested):
It seems like if any more job types are added, they'll all use the prepare input function, what do you think of making it a converter that'd be applied to the code parameter?

Some of the methods are still using the eval naming (in both their names, and some strings inside), I think `send_eval` and `post_eval` should be changed to job like the others. There's also `filter_eval` from the `Filtering` cog, but in that case I think it's clearer to use eval as it doesn't have the context of being in the `Snekbox` cog, or maybe something like `filter_snekbox_output` would work.

Looks fine otherwise

bot/exts/utils/snekbox.py
```diff
@@ -22,17 +22,87 @@
...
+        def read(self):
+            return self.pop()
+
+        def flush(self):
+            pass
+
+    sys.stdout = Writer()
+
+    def print_last_line():
+        if sys.stdout:
```
Bluenix2, line 50: I don't understand this, when would `sys.stdout` be falsely?

Bluenix2 (changes requested):
I only have one code-specific comment, but can you add a few timeit tests as well? I'd say the following two will do:

- Test setup codeblock detection
- Submitting timeit jobs carries the correct args

I did also get an issue running this locally:
```python
bot_1        |   File "/usr/local/lib/python3.9/site-packages/discord/ext/commands/core.py", line 167, in wrapped
bot_1        |     ret = await coro(*args, **kwargs)
bot_1        |   File "/bot/bot/decorators.py", line 136, in inner
bot_1        |     await func(self, ctx, *args, **kwargs)
bot_1        |   File "/bot/bot/exts/utils/snekbox.py", line 438, in timeit_command
bot_1        |     await self.run_job("timeit", ctx, code=code, args=args)
bot_1        |   File "/bot/bot/exts/utils/snekbox.py", line 382, in run_job
bot_1        |     response = await self.send_job(ctx, code, args=args, job_name=job_name)
bot_1        |   File "/bot/bot/exts/utils/snekbox.py", line 271, in send_job
bot_1        |     filter_triggered = await filter_cog.filter_snekbox_job(msg, ctx.message)
bot_1        | AttributeError: 'Filtering' object has no attribute 'filter_snekbox_job'
```
It seems like you renamed it to `filter_snekbox_output` but use `filter_snekbox_jon`.

bot/exts/utils/snekbox.py
ToxicKidz: If the queue is empty.

bot/exts/utils/snekbox.py
lemonsaurus: I agree wholeheartedly. This code sorely needs block comments and docstrings just like code that isn't inside of a triple quoted string would. Just because the linter doesn't mind, doesn't mean we should lower our readability standards.

ToxicKidz: Agreed. Resolved in https://github.com/python-discord/bot/pull/1602/commits/afdb2b4d5ebeda20d90e794711ca7e56244cfd6b.

Bluenix2 (approved):
I haven't really tried to escape or abuse the code before it, this is purely a functional approval. Works nicely and will be super useful! Thanks.

wookie184 (approved):
All seems good to me 🎉

Issue #1548, opened 2021-04-21
"!d if" and "!d for" commands are grabbing from django docs.


camprevail:
The doc commands `!d if` and `!d for` (and possibly others) are grabbing from django docs. The expected behavior is the python syntax for if/elif/else. 
`!d label.if` kind of pulls the right doc, but it's the grammatical usage and not very helpful for trying to understand the usage. 
Maybe there could be some dedicated tags for them?
I've written a suggestion explanation for a dedicated `!if-else` tag here. This is just how I might explain it. Maybe it should be simplified, but I'll leave it to the pros. 
```
if condition:
    # condition can be a variable or expression.
    # Ex. if condition > 0:
    print('condition is True'.)
elif condition2: 
    # elif is optional and can be used to test the same condition  
    # or another condition.
    print('condition is False, but condition2 is True.')
else:
    # The default action if the previous conditions are False
    print('Both conditions were False.')
```

RohanJnr:
If you are interested in writing a tag, go for it and can also open a PR for it.

The problem of the doc command fetching the Django template syntax for if-else will need some digging and investigation.

kosayoda:
The django/python docs conflict has been solved in #1606. I've updated your issue to reflect the tag suggestion.

PR #1606, opened 2021-05-21
Improve doc symbol name conflict resolution

Numerlor:
Instead of only using the package name as a fallback to resolve the conflicts, it's now used as the first path with the group renames only used for conflicts within packages, this should make it clearer what the symbol may be about when it's shown in the similar names footer.
This also makes the names a bit less dependent on the order the inventories were fetched in.

In addition to using the package name, conflicts within a package where both of the symbols have their group in `FORCE_PREFIX_GROUPS` are now handled with respect to their order in the constant, where the symbol with the higher priority group uses the un-modified key.

Takes care of the initial issue in #1548 where non python docs were fetched for certain symbols.

### DIFF
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -30,6 +30,7 @@ FORCE_PREFIX_GROUPS = (
     "term",
     "label",
     "token",
+    "doc",
     "pdbcommand",
     "2to3fixer",
 )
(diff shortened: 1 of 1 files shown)


kosayoda (approved):
Solves the problem.

Xithrius (approved):
It works well.

PR #1608, opened 2021-05-21
Enforce one help channel per user

### LINKED ISSUE #1576: Enforce only 1 Help Channel at a time
# Proposal
Currently, if a user opens a help channel they have the "Help Cooldown" role until they close the help channel or until 15 minutes has passed, whichever occurs first. This allows a person to open multiple help channels if they wait 15 minutes and don't close their first one. This has led to some sub-par help sessions since a person is trying to get help in multiple channels and it's a bad experience all around.

During the staff meeting, the consensus was that we wanted to enforce only 1 help channel per user open at a time.

# Implementation Details
The Help Cooldown role should only be removed when a user closes the channel. This should hopefully only involve removing the 15 minute remove-role option.

swfarnsworth:
This PR removes the cool down functionality from the help channel cooldown role, and instead applies the role when their help channel opens and removes it exactly when that channel closes, thereby ensuring that a given user is never the claimant of more than one channel.

If you test this functionality, please list which cases you tried.

Closes #1576.


wookie184:
Have played around with it a bit and it seems to work fine. My only worry is that removing the timed cooldown creates a potential for trolls to quickly open and close their channel to mess with the system, I think that was one of the reasons for having a cooldown. On the other hand, being able to simplify the system is nice, and the amount of harm somebody could cause is likely pretty low and we would notice/deal with it pretty quickly.

To clarify, that would still involve limiting the help channels to one per person, but on top of that there would be a short cooldown (e.g. 5 minutes) to prevent abuse. Thoughts?

swfarnsworth:
@wookie184 I don't think that concern was raised during the meeting where we green-lit this change.

I'm hesitant to have the "channel is closed and five minutes" cooldown as it re-introduces some of the complexity that this seeks to eliminate. I have also at times closed help sessions that have been overrun by trolls and instructed the claimant to open a new session, and I wouldn't want to lose the option to do that without penalizing the claimant.

wookie184 (approved):
My bad, I forgot that the role was already removed. In that case this isn't a change of behaviour in that regard so it's fine. This looks good to me, have tested and the role is added and removed correctly, with command `!close` and natural close.

You could add log messages for when the role is added or removed, as was in `add_cooldown_role` and `remove_cooldown_role` previously

MarkKoz (changes requested):
I'm not too concerned about `get_role` failing, but `add_roles` should have some exception handling. Failing to change the role isn't a critical failure; it shouldn't leave the system in an invalid/broken state. Some of the functions you removed were taking care of the error handling. I don't see why you can't just leave them in and keep using them.

bot/exts/help_channels/_cog.py
```diff
@@ -106,9 +125,11 @@ async def claim_channel(self, message: discord.Message) -> None:
         """
         log.info(f"Channel #{message.channel} was claimed by `{message.author.id}`.")
         await self.move_to_in_use(message.channel)
-        await _cooldown.revoke_send_permissions(message.author, self.scheduler)
+        cooldown_role = self.bot.get_guild(constants.Guild.id).get_role(constants.Roles.help_cooldown)
```
MarkKoz, line 128: This line can be put into the function to reduce code redundancy.

MarkKoz (changes requested):

MarkKoz (approved):

PR #1609, opened 2021-05-22
Fix infraction rescheduler breaking with more than 100 in flight reactions

### LINKED ISSUE #1581: Temporary infractions don't always get lifted
The issue is probably our new limit to infractions returned from the site API, combined with the heavy mass ban we recently did. We have probably crossed 100 active infractions since those were applied, and as a result, they don't show up to be rescheduled when the scheduler restarts.

Having >100 infractions active at a time is a reasonable figure for us to have right now with our size, even without the mass bans.

It does appear to have an easy temporary fix though, `reschedule_infractions()` should add a filter to the site query to ignore permanent infractions.

A permanent fix would be to remove the pagination cap for this specific request in the site api.

cc @bast0006 @mbaruh

bast0006:
Make sure to only fetch infractions that actually need rescheduling by filtering by type and permanent status. We don't reschedule permanents as they will never be automatically expired, so they're a waste and clog to filter out manually. There is a PR for `site` to add the requisite filters (`types` and `permanent`) for this part.
We also only reschedule the soonest-expiring infractions, waiting until we've processed all of them before fetching the next batch by ordering them by expiration time.

This requires python-discord/site#510 to be merged first, so that the api changes it needs are available. 

Fixes #1581


bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -48,11 +48,32 @@ async def reschedule_infractions(self, supported_infractions: t.Container[str])
 
         infractions = await self.bot.api_client.get(
             'bot/infractions',
-            params={'active': 'true'}
+            params={
+                'active': 'true',
+                'ordering': 'expires_at',
+                'permanent': 'false',
+                'types': ','.join(supported_infractions),
+            },
```
ToxicKidz, line 56: I know this was using single quotes already, but [our style guide](https://pythondiscord.com/pages/guides/pydis-guides/contributing/style-guide/#quote-marks) says that we should prefer single quotes over double quotes.
```suggestion
            params={
                "active": "true",
                "ordering": "expires_at",
                "permanent": "false",
                "types": ",".join(supported_infractions),
            },
```

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -48,11 +48,32 @@ async def reschedule_infractions(self, supported_infractions: t.Container[str])
...
         )
-        for infraction in infractions:
-            if infraction["expires_at"] is not None and infraction["type"] in supported_infractions:
-                self.schedule_expiration(infraction)
+
+        to_schedule = [i for i in infractions if not i['id'] in self.scheduler]
```
ToxicKidz, line 59: Prefer `x not in y` over `not x in y`
```suggestion
        to_schedule = [i for i in infractions if i['id'] not in self.scheduler]
```

ToxicKidz (commented):
A couple suggestions related to styling

ChrisLovering (dismissed):
Tested and all looks good to me! 

Backwards compatible too, so can be merged before the bot PR

ChrisLovering (approved):
Tested locally with the new site changes, all looks good to me!

ToxicKidz (approved):
:shipit:

Xithrius (approved):

PR #1611, opened 2021-05-24
Up duck pond threshold to 6

gustavwilliam:
Makes duck pond entries less common, by requiring more ducks for a message to be ducked.


gustavwilliam:
Moving this to a different branch and changing the limit to 7.

PR #1612, opened 2021-05-24
Up duck pond threshold to 7

gustavwilliam:
Makes duck pond entries less common, by requiring more ducks in order for a message to be ducked.


mbaruh (approved):

vcokltfre (approved):
For the record I think this will kill the duck pond, however, I also think it's good to try things out since it can be so easily reverted so... let's go for it!

PR #1614, opened 2021-05-24
Update pixels token regex to false match less

ChrisLovering:
This new regex means there needs to be a minimum number of characters between each `.`, along with requiring there to be 3 "sections" to the API token.

This should reduce the number of false matches


vcokltfre (approved):
<:ok_handbutflipped:779364331350523909>

mbaruh (approved):

PR #1615, opened 2021-05-27
Recruitment: reverse nomination order

Akarys42:
This makes review appear in chronological order, making it easier to say things like "^^ what they said".


ChrisLovering (approved):

HassanAbouelela (approved):
Now we need a PR title hall of fame.

PR #1626, opened 2021-06-07
Pytest-xdist

### LINKED ISSUE #1622: Unittest Parallelization & Pytest
## Description
Over time, this project's test suite has grown to be very large, and so too has its operation time. On my machine, the test suite currently takes about 80s to complete. You can see a full breakdown [here](https://hassanamr.dev/pydis/tests.html).

This issue proposes parallelization to help shorten the run time of tests.

## Approaches
Naturally, there are many ways to go about doing this, each with varying degrees of difficulty.

**Approach 1:** [`unittest-parallel`](https://pypi.org/project/unittest-parallel/)
unittest-parallel is a pypi plugin for unittest. It theoretically should be the least effort required to get parallelization working, yet, I could not get it to work after a lot of testing. Random tests were failing, I could not get it to work at all.

Other complaints include:
- Can't kill the test runner. No combination of `ctrl+combo` would stop the tests after they started.
- It was not faster than plain unittests from the looks of it. I obviously couldn't get the entire test suite to run, but the part that did took much longer than plain unittests.
- There were other various formatting complaints, but they'll take a lot of space, and they're completely moot if the points above are not addressed.

**Approach 2:** [`pytest-xdist`](https://pypi.org/project/pytest-xdist/)
Alright, hold your horses. Don't start typing out how terrible this idea is yet. Yes, this idea relies on pytest, but we'll get to that later. The possibility of adding pytest is a large part of this issue. Now, back to xdist:

It's a pytest plugin that parallelizes your code. I don't really have anything else to say about that. It worked really well out of the box, and I didn't run into any problems. Here are some numbers:

|                              | Unittests | Pytest | xdist |
| ------------------- | ---------- | ------ | ----- |
| **No Coverage** | 60s           | 71s     | 45s  |
| **Coverage**       | 79s           | 130s   | 75s  |
| **CI Coverage**   | 91s           | ~       | 52s   |

## Pytest
This is the next large part of this issue. Using pytest has a few benefits, which we can get into in a second. There are two approaches however: using pytest as a runner, and using it as a framework.

**Using as a runner:**
We could use pytest exclusively to run the tests. That would give us access to pytest plugins, without requiring we use pytest. It's the least work since pytest has native support for unittest suites, and the only changes needed are installing the dependencies.

**Using as a framework:**
Alternatively, we could use pytest in full. That will give us access to all the things mentioned above, as well as other pytest features, such as fixtures. This could be a much larger job depending on how it's approached. There is no requirement to immediately refractor all old tests into pytest format, we could have a mix of both. We could stop accepting any more non-pytest test suites, and slowly update old suites as they are worked on. Not the most ideal solution because it would require maintaining two test suites, but it's an idea.

HassanAbouelela:
Closes #1622 

This PR makes a few changes to the way tests are handled:
- Switches to pytest purely  as a test runner.
- Adds pytest-xdist
- Switches to pytest-cov to generate `.coverage` files, but leaves coverage.py to process them.

The last point about pytest-cov is due to the simple fact that coverage.py doesn't support xdist as far as I could tell, and pytest-cov did not support manipulating the data after the fact. The intention here is to minimize any changes one might experience in the contributing flow.


HassanAbouelela:
Converted this PR to a draft while I figure out why we lost 10% coverage

HassanAbouelela:
I figure it's because I've excluded `tests` from source, hence an overall decrease in coverage. I don't think it makes much sense to have that, since you aren't testing your tests, but I'm willing to hear other's opinions.

HassanAbouelela:
I've made another change. I split the test task into two: `fast-test` and `test`. `test` is unchanged, but `fast-test` has been added to run the tests without coverage, since including coverage can take up 2x the time.

As a sidenote, I now realize the original numbers in the issue are incorrect, as they included coverage for unittests, but not pytest. I've updated the numbers there to reflect that.

kwzrd:
Afaik the reason to include the tests themselves in the coverage report is that it becomes easier to find test code that wasn't used, or tests that didn't run, when they probably should have. You may also find yourself patching in a block of code that mocks something and is defined in the tests package; seeing coverage for it may be useful.

I don't know how compelling these arguments are to you, but I remember them from reading Ves' docs on the testing utils they had prepared.

tests/README.md
```diff
@@ -11,10 +11,15 @@ We are using the following modules and packages for our unit tests:
...
+- [pytest-cov](https://pytest-cov.readthedocs.io/en/latest/index.html)
+
+We also use the following package as a test runner:
+- [pytest](https://docs.pytest.org/en/6.2.x/)
 
 To ensure the results you obtain on your personal machine are comparable to those generated in the CI, please make sure to run your tests with the virtual environment defined by our [Poetry Project](/pyproject.toml). To run your tests with `poetry`, we've provided two "scripts" shortcuts:
 
-- `poetry run task test` will run `unittest` with `coverage.py`
+- `poetry run task fast-test` will run `pytest`.
+- `poetry run task test` will run `pytest` with `pytest-cov`.
```
kwzrd, line 22: The line above says "two scripts" but it is four now.

kwzrd: Or maybe just three. :smile:

HassanAbouelela:
I don't understand the point about files in the tests directory not running. How does a test not run? How does that make it past review? Helpers make more sense, but I don't know if it's a compelling enough reason to add tests as a src.

Regardless, adding them is the path of least resistance, so... fb053e4.

tests/README.md
HassanAbouelela: Counting is hard. e45843e

HassanAbouelela:
And now the coverage has jumped up 5 percent...

I'll just let it be for tonight.

MarkKoz:
Why are you passing options to coverage through the command rather than letting it use the `.coveragerc` file? Does it not support reading from that file?

HassanAbouelela:
I removed it when it was no longer needed, and never added it back.

That's been reverted in e2064b4.

pyproject.toml
```diff
@@ -58,6 +61,7 @@ lint = "pre-commit run --all-files"
 precommit = "pre-commit install"
 build = "docker build -t ghcr.io/python-discord/bot:latest -f Dockerfile ."
 push = "docker push ghcr.io/python-discord/bot:latest"
-test = "coverage run -m unittest"
+fast-test = "pytest -n auto"
```
MarkKoz, line 64: How about this?

1. Nicer to start it with "test"
2. "nocov" is more meaningful than "fast" but perhaps no everyone will catch on to it meaning "no coverage"

```suggestion
test-nocov = "pytest -n auto"
```

.github/workflows/lint-test.yml
```diff
@@ -97,12 +97,9 @@ jobs:
         --format='::error file=%(path)s,line=%(row)d,col=%(col)d::\
         [flake8] %(code)s: %(text)s'"
 
-      # We run `coverage` using the `python` command so we can suppress
-      # irrelevant warnings in our CI output.
       - name: Run tests and generate coverage report
         run: |
-            python -Wignore -m coverage run -m unittest
-            coverage report -m
+            pytest -n auto --cov --disable-warnings -q
```
MarkKoz, line 102: This doesn't have to be on a separate line any more but whatever.

MarkKoz (commented):
Can anyone confirm xdist works on all three major operating systems?

pyproject.toml
HassanAbouelela: I didn't particularly like "fast", but I didn't have a better name. I would usually just call this test, and have something indicating coverage in the other one, but I didn't want to break the existing convention.

nocov works, let's use that.

.github/workflows/lint-test.yml
HassanAbouelela: Might as well fix it.

HassanAbouelela:
Tested and can confirm the branch works on Windows 10 and Linux (ubuntu 20.04. CI also runs ubuntu).

tests/README.md
```diff
@@ -11,10 +11,15 @@ We are using the following modules and packages for our unit tests:
...
+- [pytest-cov](https://pytest-cov.readthedocs.io/en/latest/index.html)
 
-To ensure the results you obtain on your personal machine are comparable to those generated in the CI, please make sure to run your tests with the virtual environment defined by our [Poetry Project](/pyproject.toml). To run your tests with `poetry`, we've provided two "scripts" shortcuts:
+We also use the following package as a test runner:
+- [pytest](https://docs.pytest.org/en/6.2.x/)
 
-- `poetry run task test` will run `unittest` with `coverage.py`
+To ensure the results you obtain on your personal machine are comparable to those generated in the CI, please make sure to run your tests with the virtual environment defined by our [Poetry Project](/pyproject.toml). To run your tests with `poetry`, we've provided the following "script" shortcuts:
+
+- `poetry run task fast-test` will run `pytest`.
```
MarkKoz, line 21: You forgot to update the documentation.
```suggestion
- `poetry run task test-nocov` will run `pytest`.
```

MarkKoz (approved):

tests/README.md
HassanAbouelela: Thanks

pyproject.toml
```diff
@@ -58,6 +61,7 @@ lint = "pre-commit run --all-files"
 precommit = "pre-commit install"
 build = "docker build -t ghcr.io/python-discord/bot:latest -f Dockerfile ."
 push = "docker push ghcr.io/python-discord/bot:latest"
-test = "coverage run -m unittest"
+test-nocov = "pytest -n auto"
+test = "pytest -n auto --cov-report= --cov"
```
Numerlor, line 65: What's the reason for suppressing the coverage printing?

HassanAbouelela: I wanted to maintain as much of the original interface as possible. The coverage also takes up a lot of space.

Numerlor: Ah, I must've misremembered. Thought it was printed as a part of the test command before, but looks like it always was a separate step.

Numerlor (approved):

Xithrius (approved):
It works well.

jb3 (approved):
Looks good! Signing off for DevOps.

graingert:
> How does a test not run? How does that make it past review?

https://nedbatchelder.com/blog/201908/dont_omit_tests_from_coverage.html

I had a horror story in which pytest-twisted was not installed so most tests were passing without actually scheduling any coroutines or running any assertions

pyproject.toml
```diff
@@ -47,6 +47,9 @@ pep8-naming = "~=0.9"
 pre-commit = "~=2.1"
 taskipy = "~=1.7.0"
 python-dotenv = "~=0.17.1"
+pytest = "~=6.2.4"
+pytest-cov = "~=2.12.1"
+pytest-xdist = { version = "~=2.2.1", extras = ["psutil"] }
```
graingert, line 52: this pytest-xdist is now outdated I upgraded it here https://github.com/python-discord/bot/pull/1659

PR #1629, opened 2021-06-07
helpdm for participants

### LINKED ISSUE #1567: Help Channel DM's For The Person Who Is Helping.
So we already have the help channel DM's that sends the user a DM whenever they claim a help channel. But why not also send a DM to the person who is helping in the channel as well. We could also make this toggleable just incase someone doesn't want to receive the DM's.

JakeM0001:
closes #1567

This PR is for the help channel participant DMs, me and Griff went on and tested this in and we ran into no problems at all, we even went as far as testing it with multiple people.

This is the embed that the user receives in DMs.
image

Also added feedback if the user already has help DMs turned on already(this is the same for when the user turns them off)
image

### DIFF
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -577,7 +577,7 @@ class HelpChannels(commands.Cog):
             embed = discord.Embed(
                 title="Currently Helping",
                 description=f"You're currently helping in {message.channel.mention}",
-                color=discord.Colour.green(),
+                color=constants.Colours.soft_green,
                 timestamp=message.created_at
             )
             embed.add_field(name="Conversation", value=f"[Jump to message]({message.jump_url})")
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -577,7 +577,7 @@ class HelpChannels(commands.Cog):
 
             embed = discord.Embed(
                 title="Currently Helping",
-                description=f"You're currently helping in <#{message.channel.id}>",
+                description=f"You're currently helping in {message.channel.mention}",
                 color=discord.Colour.green(),
                 timestamp=message.created_at
             )
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -470,8 +470,6 @@ class HelpChannels(commands.Cog):
 
         elif channel_utils.is_in_category(message.channel, constants.Categories.help_in_use):
             await self.notify_session_participants(message)
-
-        else:
             await _message.update_message_caches(message)
 
     @commands.Cog.listener()
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -12,6 +12,7 @@ from discord.ext import commands
 
 from bot import constants
 from bot.bot import Bot
+from bot.converters import allowed_strings
 from bot.exts.help_channels import _caches, _channel, _message, _name, _stats
 from bot.utils import channel as channel_utils, lock, scheduling
 
@@ -577,7 +578,7 @@ class HelpChannels(commands.Cog):
 
             embed = discord.Embed(
                 title="Currently Helping",
-                description=f"You're currently helping in {message.channel.mention}",
+                description=f"You're currently helping in <#{message.channel.id}>",
                 color=discord.Colour.green(),
                 timestamp=message.created_at
             )
@@ -589,42 +590,43 @@ class HelpChannels(commands.Cog):
                 self._serialise_session_participants(session_participants)
             )
 
-    @commands.group(name="helpdm")
-    async def help_dm_group(self, ctx: commands.Context) -> None:
+    @commands.command(name="helpdm")
+    async def helpdm_command(
+            self,
+            ctx: commands.Context,
+            state: allowed_strings("on", "off") = None  # noqa: F821
+    ) -> None:
         """
-        User will receive a embed when they are "helping" in a help channel.
+        Allows user to toggle "Helping" dms.
 
-        If they have helpdms off they will won't receive an embed.
-        If they have helpdms on they will receive an embed.
+        If this is set to off the user will not receive a dm for channel that they are participating in.
+        If this is set to on the user will receive a dm for the channel they are participating in.
         """
-        if ctx.invoked_subcommand is None:
-            await ctx.send_help(ctx.command)
+        state_bool = state.lower() == "on"
 
-    @help_dm_group.command(name="on")
-    async def on_command(self, ctx: commands.Context) -> None:
-        """Turns help dms on so the user will receive the participating dm."""
-        if await _caches.help_dm.get(ctx.author.id):
-            await ctx.send(f"{constants.Emojis.cross_mark}{ctx.author.mention} Help DMs are already ON!")
+        requested_state_bool = state.lower() == "on"
+        if requested_state_bool == await _caches.help_dm.get(ctx.author.id, False):
+            if await _caches.help_dm.get(ctx.author.id):
+                await ctx.send(f"{constants.Emojis.cross_mark}{ctx.author.mention} Help DMs are already ON!")
 
-            log.trace(f"{ctx.author.id} Attempted to turn Help DMs on but they are already ON")
+                log.trace(f"{ctx.author.id} Attempted to turn Help DMs on but they are already ON")
+                return
+
+            if not await _caches.help_dm.get(ctx.author.id):
+                await ctx.send(f"{constants.Emojis.cross_mark}{ctx.author.mention} Help DMs are already OFF!")
+
+                log.trace(f"{ctx.author.id} Attempted to turn Help DMs on but they are already OFF")
+                return
+
+        if state_bool:
+            await _caches.help_dm.set(ctx.author.id, True)
+
+            await ctx.send(f"{constants.Emojis.ok_hand} {ctx.author.mention} Help DMs ON!")
+
+            log.trace(f"{ctx.author.id} Help DMs ON")
             return
 
-        await _caches.help_dm.set(ctx.author.id, True)
-
-        await ctx.send(f"{constants.Emojis.ok_hand} {ctx.author.mention} Help DMs ON!")
-
-        log.trace(f"{ctx.author.id} Help DMs OFF")
-
-    @help_dm_group.command(name="off")
-    async def off_command(self, ctx: commands.Context) -> None:
-        """Turns help dms off so the user wont receive the participating dm."""
-        if not await _caches.help_dm.get(ctx.author.id):
-            await ctx.send(f"{constants.Emojis.cross_mark} {ctx.author.mention} Help DMs are already OFF!")
-
-            log.trace(f"{ctx.author.id} Attempted to turn Help DMs on but they are already OFF")
-            return
-
-        await _caches.help_dm.set(ctx.author.id, False)
+        await _caches.help_dm.delete(ctx.author.id)
 
         await ctx.send(f"{constants.Emojis.ok_hand} {ctx.author.mention} Help DMs OFF!")
 
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -589,9 +589,9 @@ class HelpChannels(commands.Cog):
 
     @commands.command(name="helpdm")
     async def helpdm_command(
-            self,
-            ctx: commands.Context,
-            state_bool: bool
+        self,
+        ctx: commands.Context,
+        state_bool: bool
     ) -> None:
         """
         Allows user to toggle "Helping" dms.
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -417,10 +417,10 @@ class HelpChannels(commands.Cog):
         return await _unclaim_channel(channel, claimant_id, closed_on)
 
     async def _unclaim_channel(
-            self,
-            channel: discord.TextChannel,
-            claimant_id: int,
-            closed_on: _channel.ClosingReason
+        self,
+        channel: discord.TextChannel,
+        claimant_id: int,
+        closed_on: _channel.ClosingReason
     ) -> None:
         """Actual implementation of `unclaim_channel`. See that for full documentation."""
         await _caches.claimants.delete(channel.id)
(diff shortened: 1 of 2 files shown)


bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +540,94 @@ async def update_available_help_channels(self) -> None:
...
+        """
+        User will receive a embed when they are "helping" in a help channel.
+
+        If they have helpdms off they will won't receive an embed.
+        If they have helpdms on they will receive an embed.
+        """
+        if ctx.invoked_subcommand is None:
+            await ctx.send_help(ctx.command)
+
+    @help_dm_group.command(name="on")
```
ChrisLovering, line 605: There's quite a bit of repeated code in these two subcommands, I think they could both be brought into a single command which takes an argument of on/off.

See bot.converters.allowed_strings on how to limit input like that.

you could do something like 
```py
      requested_state_bool = requested_state.lower() == "on"
      if requested_state_bool == await _caches.help_dm.get(ctx.author.id, False):
```
where requested_state is the string yes or no from the function args. This if will tell you if they are trying to set their state to whats already stored, so you can show them the error, if not, then you can set their state to that

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +540,94 @@ async def update_available_help_channels(self) -> None:
...
+    @help_dm_group.command(name="off")
+    async def off_command(self, ctx: commands.Context) -> None:
+        """Turns help dms off so the user wont receive the participating dm."""
+        if not await _caches.help_dm.get(ctx.author.id):
+            await ctx.send(f"{constants.Emojis.cross_mark} {ctx.author.mention} Help DMs are already OFF!")
+
+            log.trace(f"{ctx.author.id} Attempted to turn Help DMs on but they are already OFF")
+            return
+
+        await _caches.help_dm.set(ctx.author.id, False)
```
ChrisLovering, line 629: Instead of overwriting the value with False, maybe we can just remove them from the dict entirely?

That way, this cache only includes those that have opted in

bot/exts/help_channels/_cog.py
```diff
@@ -417,13 +417,14 @@ async def unclaim_channel(self, channel: discord.TextChannel, *, closed_on: _cha
...
 
     async def _unclaim_channel(
-        self,
-        channel: discord.TextChannel,
-        claimant_id: int,
-        closed_on: _channel.ClosingReason
+            self,
+            channel: discord.TextChannel,
+            claimant_id: int,
+            closed_on: _channel.ClosingReason
```
ChrisLovering, line 423: Any reason for the extra indent here?

ChrisLovering (changes requested):
Just a few comments,. overall looks pretty good :D

bot/exts/help_channels/_cog.py
JakeM0001: So quick question, why does `await _caches.help_dm.get(ctx.author.id, False)` return true. I am testing it out and it works, but I don't understand why that would return true.

ChrisLovering: The second parameter defines what to return if the key isn't found in redis. So if `ctx.author.id` is in there, it will return the value, if it's not, it will return False

JakeM0001: Before I push this, should I change anything? (I am going to add logging and the docstring)
```py
@commands.command(name="helpdm")
    async def helpdm_command(
            self,
            ctx: commands.Context,
            state: allowed_strings("on", "off") = None
    ) -> None:
        state_bool = state.lower() == "on"

        requested_state_bool = state.lower() == "on"
        if requested_state_bool == await _caches.help_dm.get(ctx.author.id, False):
            if await _caches.help_dm.get(ctx.author.id):
                await ctx.send(f"{constants.Emojis.cross_mark}{ctx.author.mention} Help DMs are already ON!")
                return

            if not await _caches.help_dm.get(ctx.author.id):
                await ctx.send(f"{constants.Emojis.cross_mark}{ctx.author.mention} Help DMs are already OFF!")
                return

        if state_bool:
            await _caches.help_dm.set(ctx.author.id, True)

            await ctx.send(f"{constants.Emojis.ok_hand} {ctx.author.mention} Help DMs ON!")
            return

        await _caches.help_dm.delete(ctx.author.id)

        await ctx.send(f"{constants.Emojis.ok_hand} {ctx.author.mention} Help DMs OFF!")
```

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +538,94 @@ async def update_available_help_channels(self) -> None:
...
+        )
+
+        if message.author.id not in session_participants:
+            session_participants.add(message.author.id)
+
+            user = self.bot.get_user(message.author.id)
+
+            embed = discord.Embed(
+                title="Currently Helping",
+                description=f"You're currently helping in <#{message.channel.id}>",
```
ToxicKidz, line 580:
```suggestion
                description=f"You're currently helping in {message.channel.mention}",
```

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +538,94 @@ async def update_available_help_channels(self) -> None:
...
+                color=discord.Colour.green(),
+                timestamp=message.created_at
+            )
+            embed.add_field(name="Conversation", value=f"[Jump to message]({message.message.jump_url})")
```
ToxicKidz, line 584: What is `message.message`?

ToxicKidz (changes requested):

bot/exts/help_channels/_cog.py
JakeM0001: Should be `message.jump_url` I will change that.

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +541,85 @@ async def update_available_help_channels(self) -> None:
...
+            return  # Ignore messages that are closing the channel
+
+        session_participants = self._deserialise_session_participants(
+            await _caches.session_participants.get(message.channel.id) or ""
+        )
+
+        if not message.author.id in session_participants:
+            session_participants.add(message.author.id)
+
+            user = self.bot.get_user(message.author.id)
```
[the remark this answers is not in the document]
JakeM0001: There is no difference.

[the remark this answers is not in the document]
ToxicKidz: Right, so `get_user` shouldn't be used at all here.

bot/exts/help_channels/_cog.py
ToxicKidz: This was unresolved in another commit.

JakeM0001:
All requested changes have been done.

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +539,93 @@ async def update_available_help_channels(self) -> None:
...
+        session_participants = self._deserialise_session_participants(
+            await _caches.session_participants.get(message.channel.id) or ""
+        )
+
+        if message.author.id not in session_participants:
+            session_participants.add(message.author.id)
+
+            embed = discord.Embed(
+                title="Currently Helping",
+                description=f"You're currently helping in <#{message.channel.id}>",
```
GDWR: You could just use `message.channel.mention` for this?

bot/exts/help_channels/_cog.py
ChrisLovering: That's pretty close to what I'd do, just 3 things
1. I don't think making the default value be None is needed, we want the user to be required to give this arg.
2. The extra if statements aren't needed inside the `if requested_state_bool == await _caches.help_dm.get(ctx.author.id, False):` line. As this if statement is only entered if they are trying to set their status to what's already stored. So this could be simplified to just 
```py
if requested_state_bool == await _caches.help_dm.get(ctx.author.id, False):
    await ctx.send(f"{constants.Emojis.cross_mark} {ctx.author.mention} Help DMs are already {state}!")
    return
```
3. The last block could also be simplified to something like
```py
if state_bool:
    await _caches.help_dm.set(ctx.author.id, True)
else:
    await _caches.help_dm.delete(ctx.author.id)
await ctx.send(f"{constants.Emojis.ok_hand} {ctx.author.mention} Help DMs {state}!")
```

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +539,93 @@ async def update_available_help_channels(self) -> None:
...
+                color=discord.Colour.green(),
```
GDWR, line 580: https://github.com/python-discord/bot/blob/main/bot/constants.py#L262

Should this be using the colour from constants? so the colour is exact across the bot.

bot/exts/help_channels/_cog.py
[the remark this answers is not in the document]
JakeM0001: Yep, I fixed that

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +539,93 @@ async def update_available_help_channels(self) -> None:
...
+            ctx: commands.Context,
+            state: allowed_strings("on", "off") = None  # noqa: F821
+    ) -> None:
+        """
+        Allows user to toggle "Helping" dms.
+
+        If this is set to off the user will not receive a dm for channel that they are participating in.
+        If this is set to on the user will receive a dm for the channel they are participating in.
+        """
+        state_bool = state.lower() == "on"
```
GDWR: If the state is not passed in, it will be None.
This will now make this line of code raise an error.

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +539,93 @@ async def update_available_help_channels(self) -> None:
...
+
+        requested_state_bool = state.lower() == "on"
```
GDWR, line 604: Repeated line of code?

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +539,93 @@ async def update_available_help_channels(self) -> None:
...
+        if requested_state_bool == await _caches.help_dm.get(ctx.author.id, False):
+            if await _caches.help_dm.get(ctx.author.id):
```
GDWR: Might be worth saving to a variable the `await _caches.help_dm.get(ctx.author.id)` so that it isn't requested twice if the first if condition is False.

GDWR (changes requested):

bot/exts/help_channels/_cog.py
JakeM0001: Fixed that in the latest commit

bot/exts/help_channels/_cog.py
JakeM0001: Fixed that as well in the newest commit.

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +539,76 @@ async def update_available_help_channels(self) -> None:
...
+
+        If they meet the requirements they are notified.
+        """
+        if await _caches.claimants.get(message.channel.id) == message.author.id:
+            return  # Ignore messages sent by claimants
+
+        if not await _caches.help_dm.get(message.author.id):
+            return  # Ignore message if user is opted out of help dms
+
+        if message.content == f"{self.bot.command_prefix}close":
```
ChrisLovering, line 567: The close command has some aliases, and you can enter text after the close part too.

This line should be something like `if self.bot.get_context(message).command == self.close_command:` instead

ChrisLovering (changes requested):
Just one last change and the rest looks good. I've tested locally too and all seems well.

After the requested change is done, I'll merge in main and do a final test

ChrisLovering (approved):
Looks good and works for me locally 👍

GDWR (approved):
LGTM

JakeM0001:
@ToxicKidz thinking about it, I am not too sure about the 0, 1 to toggle the command.

GDWR:
> 
> 
> @ToxicKidz thinking about it, I am not too sure about the 0, 1 to toggle the command.

If you was to add `0`, `1` would you add `true`, `false` etc. etc.
The only other means of changing I would say is having the base command `!helpdm` toggle your current state

ToxicKidz:
I was thinking of adding the [bool converter](https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#bool) which is pretty useful, it's still case sensitive and supports `on` and `off`.

JakeM0001:
What about just having `!helpdm` and that will toggle it?

GDWR:
> 
> 
> I was thinking of adding the [bool converter](https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#bool) which is pretty useful, it's still case insensitive and supports `on` and `off`.

That is a really good suggestion, 👍

bot/exts/help_channels/_cog.py
```diff
@@ -535,3 +539,76 @@ async def update_available_help_channels(self) -> None:
...
+
+    @staticmethod
+    def _serialise_session_participants(participants: set[int]) -> str:
+        """Convert a set to a comma separated string."""
+        return ','.join(str(p) for p in participants)
+
+    @staticmethod
+    def _deserialise_session_participants(s: str) -> set[int]:
+        """Convert a comma separated string into a set."""
+        return set(int(user_id) for user_id in s.split(",") if user_id != "")
```
MarkKoz, line 551: Aren't sets in redis more appropriate for this sort of thing? It would remove the need to serialise and deserialise. It wouldn't even need to retrieve all items since there are commands to check set membership.

JakeM0001: Async Rediscache doesn't support sets.

MarkKoz: aioredis does support it and async-redis cache can trivially be extended to support it (yes I think it's worth doing that).

MarkKoz: Note I'm not suggesting implementing the set operations with all the type conversion stuff. Just a basic SADD call or whatever that will store a string as given.

JakeM0001:
Those weren't there before.... But as soon as I push they show up.

bot/exts/help_channels/_cog.py
JakeM0001: I'm not too familiar with Redis can you please give more information about how I would implement this? Also I think it would be best to do this as a separate feature.

MarkKoz: That's alright. Let's just leave your implementation for now. I can try to merge set support into async-rediscache in the future.

JakeM0001: Alright, sounds good!

ToxicKidz (approved):

PR #1634, opened 2021-06-11
Modpings schedule

### LINKED ISSUE #1588: Modpings schedule feature
# Abstract

Moderators should be able to set a schedule for when their moderator ping role gets removed, to avoid waking up to a ton of unread pings.

# Specification

A new command, `!modpings schedule set [start] [stop]` (with the two arguments being UTC time), should store those two dates in a Redis cache and make sure to remove the role when `start` hits and re-add it when `stop` hits using a scheduler.

`!modpings schedule reset` should clear the cache entry and disable it.

# Reasoning

Currently, a few moderators are using `!modpings off` to not wake up to 20-50 unread mod pings and be able to focus on pings that still require their attention. This command would help by not having to remember (and sometimes forget!) to run the command every day.

Shivansh-007:
Closes #1588 

### Description

Screenshot from 2021-06-11 10-25-10

Screenshot from 2021-06-11 10-33-43

I have disabled the check for the minimum work time in the second test above.

### Reasoning

The PR adds another scheduler to manage the modpings schedule feature since the two features (on/off and schedule), both use the same format for the ID i.e. the moderator's discord ID. Therefore if we use the same scheduler, a moderator cannot schedule his modpings if has a task in modpings.

The modpings on/off feature would override this feature, so that if one is in `pings_off_mods` cache i.e. have there pings off for sometime, then it won't add the role on their schedule time.


bot/exts/moderation/modpings.py
```diff
@@ -126,6 +183,38 @@ async def on_command(self, ctx: Context) -> None:
...
+        self._modpings_scheduler.schedule_at(
+            start,
+            ctx.author.id,
+            self.add_role_schedule(ctx.author, work_time, start)
+        )
+
+        await ctx.send(
+            f"{Emojis.ok_hand} {ctx.author.mention} Scheduled mod pings from "
+            f"{start: %I:%M%p} to {end: %I:%M%p} UTC Timing!"
+        )
```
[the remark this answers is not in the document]
Shivansh-007: > Instead of the user having to convert from UTC to their timezone, you can use an embed timestamp.
> 
> [Here](https://github.com/python-discord/bot/blob/main/bot/exts/moderation/modpings.py#L89-L109) is an example.

Yes, but I don't think it is possible to put two timestamps in a single embed.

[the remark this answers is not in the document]
Xithrius: This is unfortunately true.

bot/exts/moderation/modpings.py
```diff
@@ -126,12 +183,47 @@ async def on_command(self, ctx: Context) -> None:
...
+
+        self._modpings_scheduler.schedule_at(
+            start,
+            ctx.author.id,
+            self.add_role_schedule(ctx.author, work_time, start)
+        )
+
+        await ctx.send(
+            f"{Emojis.ok_hand} {ctx.author.mention} Scheduled mod pings from "
+            f"{start: %I:%M%p} to {end: %I:%M%p} UTC Timing!"
```
Numerlor, line 215: Why does this use the 12 hour clock? Where time is formatted on the bot side, the iso format is used which uses the 24 hour clock

bot/exts/moderation/modpings.py
```diff
@@ -55,6 +65,53 @@ async def reschedule_roles(self) -> None:
...
+            start_timestamp, work_time = schedule.split("|")
+            start = datetime.datetime.fromtimestamp(start_timestamp)
+
+            mod = self.bot.fetch_user(mod_id)
+            self._modpings_scheduler.schedule_at(
+                start,
+                mod_id,
+                self.add_role_schedule(mod, work_time, start)
+            )
+
```
Numerlor, line 84: This method is full of errors, attach an exception logging done callback to the task to see what needs to be fixed.

Numerlor (changes requested):
A quick review for now as there are some things preventing the scheduling from fully working.

The functionality should account for users leaving the moderation team and remove the scheduling when that happens.
There should also be a command to remove the schedule for yourself.

Would need confirmation as I'm not sure if I got it right myself but I don't think the 16 hour limit was meant in this way, but the other way around (as a maximum). With a 16 hour minimum the moderator will always have less than 8 hours for personal non-ping time.

bot/exts/moderation/modpings.py
Numerlor: #1637 was just merged to replace the uses of `self.bot.loop.create_task` that swallow exceptions with `scheduling.create_task(coro, event_loop=bot.loop)` to log exceptions

Akarys42:
Hey @Shivansh-007! What is the state of this PR? Is it ready for re-review?

Shivansh-007:
Yep it is waiting for a re-review :D

Bluenix2:
How does this work/compare to #1570?

Shivansh-007:
It isn't using the persistent scheduler, depending on whichever PR would get merged first, the other can do the required changes, it would be a quick change anyhow.

Bluenix2 (approved):

Bluenix2:
This PR has been open for *quite a while*.. if I am gonna be honest I thought it was merged shortly after I approved it a month ago (not to be disrespectful to anyone's available time of course).

I believe this PR needs a core-dev approval (and/or another contributor approval).. let's give this some love :shipit:

bot/exts/moderation/modpings.py
```diff
@@ -60,6 +74,53 @@ async def reschedule_roles(self) -> None:
...
+            mod = await self.bot.fetch_user(mod_id)
+            self._modpings_scheduler.schedule_at(
+                start,
+                mod_id,
+                self.add_role_schedule(mod, work_time, start)
+            )
+
+    async def remove_role_schedule(self, mod: Member, work_time: int, schedule_start: datetime.datetime) -> None:
+        """Removes the moderator's role to the given moderator."""
+        log.trace(f"Removing moderator role to mod with ID {mod.id}")
```
jchristgit, line 96:
```suggestion
        log.trace(f"Removing moderator role from mod with ID {mod.id}")
```

jchristgit (approved):
This is really cool. I do think this could be even better if we were to store the modping schedules in the PostgreSQL database of the site, because then we can query it and inspect there, but that would just be a "nice to have".

ChrisLovering (approved):
Proxy +1 for Volcyy

PR #1636, opened 2021-06-11
Add black-formatter to reminders overrides

ichard26:
Adds the black-formatter channel to the remind command overrides to
allow usage of the command in the channel. This isn't the cleanest
patch, ideally the whole OSS category would be whitelisted but the
filter currently doesn't support categories.

Co-authored-by:  Hassan Abouelela <hassan@hassanamr.com>

--- 

Signed off here: https://discord.com/channels/267624335836053506/846434317021741086/852958371497377892

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -433,6 +433,8 @@ class Channels(metaclass=YAMLGetter):
     off_topic_1: int
     off_topic_2: int
 
+    black_formatter: int
+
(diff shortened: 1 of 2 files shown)


HassanAbouelela (approved):
Thanks!

Xithrius (approved):
That's the good stuff.

PR #1637, opened 2021-06-12
Add an optional loop kwarg to scheduling.create_task

Numerlor:
The `create_task` util is used to create tasks which log their exceptions to make debugging easier, but with its current implementation it cannot be used to schedule tasks from the init of cogs as there's no running loop when they execute.

The loop kwarg allows the caller to pass in `bot.loop` in these situations to get a task with the same exception logging.

### DIFF
--- a/bot/utils/scheduling.py
+++ b/bot/utils/scheduling.py
@@ -161,9 +161,21 @@ class Scheduler:
                 self._log.error(f"Error in task #{task_id} {id(done_task)}!", exc_info=exception)
 
 
-def create_task(coro: t.Awaitable, *suppressed_exceptions: t.Type[Exception], **kwargs) -> asyncio.Task:
-    """Wrapper for `asyncio.create_task` which logs exceptions raised in the task."""
-    task = asyncio.create_task(coro, **kwargs)
+def create_task(
+        coro: t.Awaitable,
+        *suppressed_exceptions: t.Type[Exception],
+        event_loop: asyncio.AbstractEventLoop = None,
+        **kwargs
+) -> asyncio.Task:
+    """
+    Wrapper for creating asyncio `Task`s which logs exceptions raised in the task.
+
+    If the loop kwarg is provided, the task is created from that event loop, otherwise the running loop is used.
+    """
+    if event_loop is not None:
+        task = event_loop.create_task(coro, **kwargs)
+    else:
+        task = asyncio.create_task(coro, **kwargs)
     task.add_done_callback(partial(_log_task_exception, suppressed_exceptions=suppressed_exceptions))
     return task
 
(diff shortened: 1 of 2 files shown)


bot/utils/scheduling.py
```diff
@@ -161,9 +161,21 @@ def _task_done_callback(self, task_id: t.Hashable, done_task: asyncio.Task) -> N
                 self._log.error(f"Error in task #{task_id} {id(done_task)}!", exc_info=exception)
 
 
-def create_task(coro: t.Awaitable, *suppressed_exceptions: t.Type[Exception], **kwargs) -> asyncio.Task:
-    """Wrapper for `asyncio.create_task` which logs exceptions raised in the task."""
-    task = asyncio.create_task(coro, **kwargs)
+def create_task(
+        coro: t.Awaitable,
+        *suppressed_exceptions: t.Type[Exception],
```
Numerlor, line 166: I've asked this in <https://github.com/python-discord/bot/pull/1471#discussion_r606778142> but the review didn't get addressed; thoughts on replacing this with an optional kwarg instead of a varag? When calling the function, the purpose of the passed in exceptions isn't immediately clear.

Numerlor (commented):
I held off on replacing the `self.bot.create_task` pattern with this function for now as it'd be a larger change so I'd like some confirmation beforehand, but making sure only the utility function is used where applicable will prevent future errors popping up on already existing uses.

bot/utils/scheduling.py
```diff
@@ -161,9 +161,21 @@ def _task_done_callback(self, task_id: t.Hashable, done_task: asyncio.Task) -> N
...
+        event_loop: asyncio.AbstractEventLoop = None,
```
MarkKoz, line 167: Thoughts on naming this `loop` to be consistent with the (deprecated) parameter in asyncio?

bot/utils/scheduling.py
MarkKoz: I agree with you about the clarity. On the other hand, it's kind of annoying to write the kwarg and wrap everything in a tuple/list, especially when there's only one Exception. I'll say this: as far as I am concerned, if you want to change it, you can. If not, that's fine too.

bot/utils/scheduling.py
ToxicKidz, line 167:
```suggestion
        event_loop: t.Optional[asyncio.AbstractEventLoop] = None,
```

Numerlor: I went with `loop` at first, but found `event_loop` to be a bit more explicit in what it is. The create_task interface also never used the loop kwarg so I didn't see a need to match the names for consistency

bot/utils/scheduling.py
Numerlor: I went ahead and moved it to a kwarg; I'm not too concerned about the small ease of use difference as it's not used that commonly

ToxicKidz (approved):

MarkKoz (approved):

PR #1639, opened 2021-06-13
Allowing python community and partners to run !poll

### LINKED ISSUE #1632: Allow `!poll` functionality for Partners
Since we now have Partners more integrated into our server and our spaces, I'd like to provide them access to some of the more useful bot functionality such as `!poll`. The main driver of this is to allow the `black` maintainers the ability to run quick polls in their dev channel.

This should be a fairly simple change since it's adding another role to the decorator.

Xithrius:
Closes #1632.

More people can now run the poll command!

Moved `Roles.project_leads` and `Roles.domain_leads` to a constant with the newly added roles since the decorator would be a bit too long otherwise.

### DIFF
--- a/bot/exts/utils/utils.py
+++ b/bot/exts/utils/utils.py
@@ -40,6 +40,7 @@ If the implementation is hard to explain, it's a bad idea.
 If the implementation is easy to explain, it may be a good idea.
 Namespaces are one honking great idea -- let's do more of those!
 """
+LEADS_AND_COMMUNITY = (Roles.project_leads, Roles.domain_leads, Roles.partners, Roles.python_community)
 
 
 class Utils(Cog):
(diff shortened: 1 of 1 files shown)


bot/exts/utils/utils.py
```diff
@@ -40,6 +40,7 @@
 If the implementation is easy to explain, it may be a good idea.
 Namespaces are one honking great idea -- let's do more of those!
 """
+LEADS_AND_COMMUNITY = (Roles.project_leads, Roles.domain_leads, Roles.partners, Roles.python_community)
```
Shivansh-007, line 43: Leads could be made a separate constant like `MODERATION_ROLES` in `constants.py`, same for community.

Xithrius: If we do eventually use these roles more often, then we probably should.

At this current moment, `LEADS_AND_COMMUNITY` contains roles that have never been used anywhere else.

jb3 (approved):

Shivansh-007 (approved):

ToxicKidz (approved):

PR #1643, opened 2021-06-15
feat: add for-else tag

vcokltfre:
This PR adds a tag showing how you can use `for/else`, as approved by kutie [here](https://canary.discord.com/channels/267624335836053506/464905259261755392/854421970509889536)

Here's how it looks in Discord:
image

### DIFF
--- a/bot/resources/tags/for-else.md
+++ b/bot/resources/tags/for-else.md
@@ -1,0 +1,16 @@
+**for-else**
+
+In Python it's possible to attach an `else` clause to a for loop. The code under the `else` block will be run if the `for` block is not broken out of, by either a `break`, `return`, or `raise` statement.
+
+Here's an example of its usage:
+```py
+numbers = [1, 3, 5, 7, 8, 9, 11]
+
+for number in numbers:
+    if number % 2 == 0:
+        print("Found an even number:", number)
+        break
+else:
+    print("All numbers are odd. How odd.")
+```
+Try running this example but with an even number in the list, see how the output changes as you do so.


JakeM0001 (approved):
Looks good to me.

bot/resources/tags/for-else.md
```diff
@@ -0,0 +1,16 @@
+**for-else**
+
+In Python it's possible to attach an `else` clause to a for loop. The code under the `else` block will be run if the `for` block is not broken out of, by either a `break`, `return`, or `raise` statement.
```
jb3, line 3: Listing `return` or `raise` here seem odd since they are non-specific to loops, they shift the flow of execution in a different manner.

I'd rather we document it similar to [this guide](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops) and write it something like this:

```suggestion
In Python it's possible to attach an `else` clause to a for loop. The code under the `else` block will be run when the iterable is exhausted (there are no more items to iterate over). Code within the else block will **not** run if the loop is broken out using `break`.
```

bot/resources/tags/for-else.md
```diff
@@ -0,0 +1,16 @@
...
+
+Here's an example of its usage:
+```py
+numbers = [1, 3, 5, 7, 8, 9, 11]
```
jb3, line 7: Let's demo the actual functionality here
```suggestion
numbers = [1, 3, 5, 7, 9, 11]
```

bot/resources/tags/for-else.md
```diff
@@ -0,0 +1,16 @@
...
+
+for number in numbers:
+    if number % 2 == 0:
+        print("Found an even number:", number)
+        break
```
jb3, line 12: To make the output a bit more visually interesting, let's add some prints for each iteration.
```suggestion
        break
    print(f"{number} is odd.")
```

bot/resources/tags/for-else.md
```diff
@@ -0,0 +1,16 @@
...
+
+In Python it's possible to attach an `else` clause to a for loop. The code under the `else` block will be run if the `for` block is not broken out of, by either a `break`, `return`, or `raise` statement.
+
+Here's an example of its usage:
+```py
+numbers = [1, 3, 5, 7, 8, 9, 11]
+
+for number in numbers:
+    if number % 2 == 0:
+        print("Found an even number:", number)
```
jb3, line 11: F-strings.
```suggestion
        print(f"Found an even number: {number}")
```

jb3 (changes requested):

Xithrius (approved):

PR #1644, opened 2021-06-16
Add dunder methods tag

### LINKED ISSUE #1598: Tag: dunder methods
## Description
This would be a tag explaining why dunder methods are special, why we would sometimes override them, and give some examples using `__init__` (constructor), `__str__` (string representation), and `__repr__` (dev-friendly representation).

Objectivitix:
## Relevant Issues
Closes #1598.

## Description
Adds the dunder methods tag described in that issue.

Sorry for the second to fourth commits, I forgot about good commit messages lol, so here they are
1. Add text to the tag explaining dunder methods
2. Add an example codeblock of `__init__`, `__str__`, and `__repr__`
3. Make the codeblock structured like a script instead of an interactive REPL session


bot/resources/tags/dunder-methods.md
```diff
@@ -0,0 +1,28 @@
...
+Other common dunder methods to override are `__str__` and `__repr__`. `__str__` is the user-friendly string representation of an object, and is implicitly called on arguments passed into the `str` function. `__repr__` is the developer-friendly string representation of an object - usually the syntax to recreate it - and is called by the `repr` function.
+
+```py
+class Foo:
+    def __init__(self, value):  # constructor
+        self.value = value
+    def __str__(self):
+        return f"This is a Foo object, with a value of {self.value}!"  # string representation
+    def __repr__(self):
+        return f"Foo({self.value})"  # way to recreate this object
```
bast0006, line 16: Consider doing `f"Foo({repr(self.value)})"`, if that doesn't make things too complicated

bot/resources/tags/dunder-methods.md
```diff
@@ -0,0 +1,28 @@
+### Dunder methods
```
bast0006, line 1: While these are MD files these are actually just output into discord embeds. While most markdown syntax (including links!) works in embeds, ### for titles doesn't. Other tags, like `decorators.md` and `async-await.md` use `**Title**` so if we intend to have a title (which I think works great here) we should use the same syntax/formatting as otherwise it ends up as literally `### Title` and looks a little out of place

bot/resources/tags/dunder-methods.md
```diff
@@ -0,0 +1,28 @@
...
+bar = Foo(5)
+
+print(bar)  # print also implicitly calls __str__
+# Output: This is a Foo object, with a value of 5!
+
+print(repr(bar))  # dev-friendly representation
+# Output: Foo(5)
+```
+
+Another example: did you know that when you use the `<left operand> + <right operand>` syntax, you're implicitly calling `<left operand>.__add__(<right operand>)`? The same applies to other operators!
```
bast0006, line 28: You might want to add a reference to python's [`operator` module documentation](https://docs.python.org/3/library/operator.html) on this line, too, maybe just as a link from `operators`--see how you can fit it in, if you can.   While it's not comprehensive for dunders it does include references and dunders for the standard operators and can be quite helpful

bot/resources/tags/dunder-methods.md
```diff
@@ -0,0 +1,28 @@
+### Dunder methods
+
+Double-underscore methods, or "dunder" methods, are special methods defined in a class that are invoked implicitly. Like the name suggests, they are prefixed and suffixed with dunders. You've probably already seen some, such as the `__init__` dunder method, also known as the "constructor" of a class, which is implicitly invoked when you instantiate an instance of a class.
+
+When you create a new class, all the default dunder methods are inherited from its superclass (which is `object` if no superclass is specified). However, we can override them by redefining the methods within this new class. For example, the default `__init__` method from `object` doesn't take any arguments, so we almost always override that to fit our needs.
+
+Other common dunder methods to override are `__str__` and `__repr__`. `__str__` is the user-friendly string representation of an object, and is implicitly called on arguments passed into the `str` function. `__repr__` is the developer-friendly string representation of an object - usually the syntax to recreate it - and is called by the `repr` function.
```
bast0006, line 7: Add on to the end of this (or in another, better place if you find one) that `__str__` falls back to `__repr__`?

bot/resources/tags/dunder-methods.md
```diff
@@ -0,0 +1,28 @@
+### Dunder methods
+
+Double-underscore methods, or "dunder" methods, are special methods defined in a class that are invoked implicitly. Like the name suggests, they are prefixed and suffixed with dunders. You've probably already seen some, such as the `__init__` dunder method, also known as the "constructor" of a class, which is implicitly invoked when you instantiate an instance of a class.
+
+When you create a new class, all the default dunder methods are inherited from its superclass (which is `object` if no superclass is specified). However, we can override them by redefining the methods within this new class. For example, the default `__init__` method from `object` doesn't take any arguments, so we almost always override that to fit our needs.
```
bast0006, line 5: Perhaps you could word the inheritance portion a little differently? All methods are inherited, dunders are just special because they come with defaults. The way this is phrased suggests that the inheritance is special. 

For the curious: You can use `dir(object)` to see what's available by default and thus almost universally available.

bot/resources/tags/dunder-methods.md
```diff
@@ -0,0 +1,28 @@
...
+        self.value = value
+    def __str__(self):
+        return f"This is a Foo object, with a value of {self.value}!"  # string representation
+    def __repr__(self):
+        return f"Foo({self.value})"  # way to recreate this object
+
+
+bar = Foo(5)
+
+print(bar)  # print also implicitly calls __str__
```
bast0006, line 21: Nit: Try moving these comments to right before the function call, then put the output line on the same line as the call. You can look at `functions-are-objects` or `names` for an example.

bast0006 (changes requested):
Thanks for taking the time to write this out, it's great. A chunk of these are just suggestions that could make it better rather than required changes, if you can't fit them feel free to just say so/let another reviewer look at it.

bot/resources/tags/dunder-methods.md
ToxicKidz:
```suggestion
        return f"Foo({self.value!r})"  # way to recreate this object
```

bast0006: TIL. Yes, that's better. Thank you!

Objectivitix: got it, resolved :D

bot/resources/tags/dunder-methods.md
Objectivitix: yep, i forgot about that, i will use the double asterisks

bot/resources/tags/dunder-methods.md
Objectivitix: yep, i will add that in the commit i'm editing right now!

bot/resources/tags/dunder-methods.md
Objectivitix: resolved

bot/resources/tags/dunder-methods.md
Objectivitix: yep, rephrased

bot/resources/tags/dunder-methods.md
Objectivitix: resolved

Xithrius:
@Objectivitix When creating a commit message, please make sure that if a person was to look back into git history they'd be able to get a hint of what occurred in a change/addition/deletion.

For more information on how to create a great commit message, see [this](<https://pythondiscord.com/pages/guides/pydis-guides/contributing/contributing-guidelines/supplemental-information/#writing-good-commit-messages>) tutorial.

bot/resources/tags/dunder-methods.md
```diff
@@ -1,10 +1,12 @@
...
 
-When you create a new class, all the default dunder methods are inherited from its superclass (which is `object` if no superclass is specified). However, we can override them by redefining the methods within this new class. For example, the default `__init__` method from `object` doesn't take any arguments, so we almost always override that to fit our needs.
+When you create a new class, all methods - including dunder methods - are inherited from its superclass, which is by default `object`. `object` has a set of default dunder methods, but we can override them by redefining the methods within this new class. For example, the default `__init__` method from `object` doesn't take any arguments, so we almost always override that to fit our needs.
 
-Other common dunder methods to override are `__str__` and `__repr__`. `__str__` is the user-friendly string representation of an object, and is implicitly called on arguments passed into the `str` function. `__repr__` is the developer-friendly string representation of an object - usually the syntax to recreate it - and is called by the `repr` function.
+> NOTE: You can use `dir(object)` to see the available default dunder methods.
```
bast0006, line 7: I'm not sure on this. The `NOTE` seems a bit extra. Giving it a moments thought I think we shouldn't include this--one of the goals with tags is as a famous engineering adage goes: 

> Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away

The goal here is to explain dunders correctly, succinctly, and with the goal of educating the reader. I think this note could probably be dropped to make this easier to read and digits without losing much of the overall success.

bot/resources/tags/dunder-methods.md
```diff
@@ -1,10 +1,12 @@
-### Dunder methods
+**Dunder methods**
 
 Double-underscore methods, or "dunder" methods, are special methods defined in a class that are invoked implicitly. Like the name suggests, they are prefixed and suffixed with dunders. You've probably already seen some, such as the `__init__` dunder method, also known as the "constructor" of a class, which is implicitly invoked when you instantiate an instance of a class.
 
-When you create a new class, all the default dunder methods are inherited from its superclass (which is `object` if no superclass is specified). However, we can override them by redefining the methods within this new class. For example, the default `__init__` method from `object` doesn't take any arguments, so we almost always override that to fit our needs.
+When you create a new class, all methods - including dunder methods - are inherited from its superclass, which is by default `object`. `object` has a set of default dunder methods, but we can override them by redefining the methods within this new class. For example, the default `__init__` method from `object` doesn't take any arguments, so we almost always override that to fit our needs.
```
bast0006, line 5: Sorry, I'm going to be a little nitpicky here. I think this can be worded a little better. At the moment, the first sentence is quiet long, with a lot of asides in it. We could probably get away with something much shorter. I don't think we need to really explain inheritance here. The first sentence attempts to do that a little, but I think we can get away with just stating that there are defaults that come from `object`.

bast0006 (changes requested):
Looks good, just two last things

bot/resources/tags/dunder-methods.md
Objectivitix: got it, i'll delete it

bot/resources/tags/dunder-methods.md
Objectivitix: yep, resolved

Objectivitix:
@Xithrius thanks for telling me this. From now on, I'm putting all the changes in the extended description.

Xithrius:
> @Xithrius thanks for telling me this. From now on, I'm putting all the changes in the extended description.

I think this should only occur for commits that absolutely need more of an explanation. The message should be short enough to where it's not truncated by GitHub, and give a very quick insight to the change.

bast0006 (approved):

Objectivitix:
@Xithrius got it

Objectivitix:
All checks passed, exapoggers

Xithrius (approved):

Objectivitix:
E X A P O G G E R S

Objectivitix:
Thanks to everyone who reviewed!!

PR #1648, opened 2021-06-18
Don't voice verify ping users who join a stage channel

ChrisLovering:
We recently had our first large stage channel event. This led to a lot of people joining a voice channel for the first time. This normally means that they get pinged to tell them how to verify for voice.

We do not want this to happen for stage channels, as the user's don't get perms to speak there anyway.

This PR ignores users who haven't been pinged yet, when trying to ping them the voice verification warning. When they join a normal voice channel for the first time, they will be pinged.

Since Discord.py only added support for stage channels in 1.7, this change required a version bump. As such, the below site change was also required.

https://github.com/python-discord/site/pull/534 must be merged first!!

### DIFF
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -254,6 +254,10 @@ class VoiceGate(Cog):
             log.trace("User not in a voice channel. Ignore.")
             return
 
+        if isinstance(after.channel, discord.StageChannel):
+            log.trace("User joined a stage chanel. Ignore.")
+            return
+
         # To avoid race conditions, checking if the user should receive a notification
         # and sending it if appropriate is delegated to an atomic helper
         notification_sent, message_channel = await self._ping_newcomer(member)
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -255,7 +255,7 @@ class VoiceGate(Cog):
             return
 
         if isinstance(after.channel, discord.StageChannel):
-            log.trace("User joined a stage chanel. Ignore.")
+            log.trace("User joined a stage channel. Ignore.")
             return
 
         # To avoid race conditions, checking if the user should receive a notification
--- a/bot/pagination.py
+++ b/bot/pagination.py
@@ -51,22 +51,25 @@ class LinePaginator(Paginator):
         suffix: str = '```',
         max_size: int = 2000,
         scale_to_size: int = 2000,
-        max_lines: t.Optional[int] = None
+        max_lines: t.Optional[int] = None,
+        linesep: str = "\n"
     ) -> None:
         """
         This function overrides the Paginator.__init__ from inside discord.ext.commands.
 
         It overrides in order to allow us to configure the maximum number of lines per page.
         """
-        self.prefix = prefix
-        self.suffix = suffix
-
         # Embeds that exceed 2048 characters will result in an HTTPException
         # (Discord API limit), so we've set a limit of 2000
         if max_size > 2000:
             raise ValueError(f"max_size must be <= 2,000 characters. ({max_size} > 2000)")
 
-        self.max_size = max_size - len(suffix)
+        super().__init__(
+            prefix,
+            suffix,
+            max_size - len(suffix),
+            linesep
+        )
 
         if scale_to_size < max_size:
             raise ValueError(f"scale_to_size must be >= max_size. ({scale_to_size} < {max_size})")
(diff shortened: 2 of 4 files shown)


bot/exts/moderation/voice_gate.py
```diff
@@ -254,6 +254,10 @@ async def on_voice_state_update(self, member: Member, before: VoiceState, after:
             log.trace("User not in a voice channel. Ignore.")
             return
 
+        if isinstance(after.channel, discord.StageChannel):
+            log.trace("User joined a stage chanel. Ignore.")
```
Numerlor, line 258:
```suggestion
            log.trace("User joined a stage channel. Ignore.")
```

Numerlor (approved):

bot/exts/moderation/voice_gate.py
```diff
@@ -254,6 +254,10 @@ async def on_voice_state_update(self, member: Member, before: VoiceState, after:
             log.trace("User not in a voice channel. Ignore.")
             return
 
+        if isinstance(after.channel, discord.StageChannel):
+            log.trace("User joined a stage channel. Ignore.")
+            return
```
Afilnor, line 259: would it be interesting to refactor like this?
```
not_applicable = any(
    member.bot,  # is a bot
    member.voice is None,  # not in a voice channel
    isinstance(after.channel, discord.StageChannel)  # user is muted in a StageChannel
)

if not_applicable:
    return
```
Or do you need the log.trace for each case?

pyproject.toml
```diff
@@ -18,7 +18,7 @@ beautifulsoup4 = "~=4.9"
 colorama = { version = "~=0.4.3", markers = "sys_platform == 'win32'" }
 coloredlogs = "~=14.0"
 deepdiff = "~=4.0"
-"discord.py" = "~=1.6.0"
+"discord.py" = "~=1.7.3"
```
tooboredtocode, line 21: Why does this PR increase the discord.py version?

bot/pagination.py
```diff
@@ -51,22 +51,25 @@ def __init__(
...
         if max_size > 2000:
             raise ValueError(f"max_size must be <= 2,000 characters. ({max_size} > 2000)")
 
-        self.max_size = max_size - len(suffix)
+        super().__init__(
+            prefix,
+            suffix,
+            max_size - len(suffix),
+            linesep
+        )
```
tooboredtocode, line 72: I'm not sure why calling the init function of the parent class seems better than doing it manually. Additionally this _should_ have been a separate PR.

tooboredtocode (changes requested):

pyproject.toml
HassanAbouelela: Stage channels were introduced in 1.7 ([reference](https://discordpy.readthedocs.io/en/latest/api.html#discord.ChannelType.stage_voice)).

tooboredtocode: Fair, enough my bad, I seem to have overlooked that part in the message.

bot/pagination.py
ChrisLovering: This is a breaking change in d.py 1.7 so needs to be part of this PR.

Calling the parent constructor was just for simplicity.

bot/exts/moderation/voice_gate.py
ChrisLovering: I found the separate trace logs for each case very useful during dev testing. I'm hesitant to make this change for that reason.

Afilnor: ok makes sense indeed

bot/pagination.py
tooboredtocode: Sure, but wouldn't it be cleaner to call the functions manually? Since that's the way it was done before

ToxicKidz: This same change was made when sir-lancebot was upgraded to 1.17

tooboredtocode: Sure, but this isn't a PR that bumps the version

tooboredtocode: Anyhow, this _is_ kinda nitpicky 😅

Numerlor: We definitely don't need a separate PR for every small change, the feature needed a version bump of a dependency so that was done.
Calling the super class' init feels better to me as in case the underlying init changes we won't be storing the attributes in a possibly incompatible way

Xithrius (approved):

ToxicKidz (approved):

Xithrius:
Don't merge this yet.

PR #1651, opened 2021-06-19
Only fetch message counts if in mod channel

### LINKED ISSUE #1635: Add typing feedback to !user command
# Issue
When using the `!user` command I have found that if the user isn't cached the command can sometimes take upwards of 10 seconds to give a response. So for a fix I was thinking we can add `async with ctx.typing` just so it gives the user some sort of feedback so they know the command is in fact working.

wookie184:
Changes to only fetch message counts for the user command if in a mod channel, as they are not displayed otherwise, and so an unnecessary api call.

This was done as the `!user` command can currently be quite slow even when ran in non mod channels. This would probably mean we can close #1635

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -241,8 +241,6 @@ class Information(Cog):
             if is_set and (emoji := getattr(constants.Emojis, f"badge_{badge}", None)):
                 badges.append(emoji)
 
-        activity = await self.user_messages(user)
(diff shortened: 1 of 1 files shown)


vcokltfre (approved):

ToxicKidz (approved):

Xithrius (approved):
It works well.

PR #1654, opened 2021-06-22
Reorder everyone ping filter so it fires after watch_regex

bast0006:
This means messages with both `@everyone` and a watched term will now ping the mod team instead of hiding beneath the everyone ping's log message which does not ping


Xithrius (approved):
Poifect.

ChrisLovering (approved):

PR #1658, opened 2021-06-27
Add alias for voice_verify command

wookie184:
Added voice-verify alias as the channel and tag are hyphenated so it may be confusing as it is currently.


HassanAbouelela (approved):
Thanks for the double quotes 👀

Akarys42 (approved):
:shipit:

PR #1660, opened 2021-06-29
Ensure mods cannot be watched

ChrisLovering:
This meant admins could also be watched, meaning their messages in any channel would be relayed to the BB channel.


Akarys42 (approved):
:shipit:

ToxicKidz (approved):

Issue #1545, opened 2021-04-20
Create tag groups


mbaruh:
It's been a concern lately of tags cluttering the bot's command namespace. To mitigate the issue, we can split certain tags into groups, and then having a lot of related tags under that group won't be an issue.

For example, `!intents` can become `!dpy intents`, and then we can add however many tags we want under `dpy`.

The tag group's md files can be stored in a common folder under `resources`.

Numerlor:
> The tag group's md files can be stored in a common folder under `resources`.

The current implementation uses folders for the (unused) role restricted tags, would we just ditch that as it's not currently in use by anything?

mbaruh:
We could. I suppose we can also create a special `_roles` or a similarly named folder and to put restricted tags there, if we ever have any.

wookie184:
Thoughts on making it so some groups are tied to certain channels, e.g. so `!intents` run in the `#discord-py` channel would work as it would search in the dpy group. An issue with this is that it may be confusing for certain tags to work in some channels and not others though.

mbaruh:
Yeah that sounds a bit unintuitive. Long time users learn to remember certain tags, and it would be harder if they were called differently in different channels

Numerlor:
> We could. I suppose we can also create a special `_roles` or a similarly named folder and to put restricted tags there, if we ever have any.

With the metadata described in https://github.com/python-discord/bot/pull/1539#issuecomment-827060796 we can do role restrictions through an attribute there, as having multiple folders would be a bit difficult to navigate if a tag was both role and group restricted.

I can start working on this if that sounds good.

Are there any group you currently have in mind except the dpy one?

mbaruh:
A point that was made is that it might be simpler to just use hyphens (e.g `!dpy-intents`), and then no code changes are needed.

The reason this issue was raised is that we've been wary of adding new tags too liberally (for example, recently there was a suggestion for more ToS related tags), fearing too much clutter, and this approach seemed like a solution.

That said, simply using hyphenated names is completely valid as well.

Opinions are welcome.

Numerlor:
There could be issues with fuzzy matching if we just use the prefixes so I think fully separating it with non fuzzy matched groups would be a bit easier in the code. 
Using the group name to list all tags in that group would also be nice but apart from that I don't think there are UX differences between the two approaches.

RohanJnr:
It would also be nice if we could call the tag name without the group name,
[...]

PR #1663, opened 2021-06-30
Add tag groups and in file metadata

Numerlor:
Introduces tag groups described in #1545.

When matching a tag with a group, the group is matched with no minimum amount of chars (unlike a tag that has to have 3 chars of its name) to allow a short invocation name to be used and possibly prevent some conflicts with tag names where multiple tags would be suggested instead of a s tag being sent.

Additionally tags now support metadata through a frontmatter block like in [meta files in branding](https://raw.githubusercontent.com/python-discord/branding/main/events/earth_day/meta.md)
Currently, the `restricted_to` key accepts a list of roles to restric the tag to, and the `embed` key is passed to `discord.Embed.from_dict`, this allows customization of the embed for individual tags

The module was redesigned a bit to separate the logic of all the fuzzy matching and name parsing from the cog itself.


bot/exts/info/tags.py
```diff
@@ -172,125 +243,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+
+            if tag.accessible_by(ctx.author):
+                result_lines.append(f"**\N{RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK}** {identifier.name}")
+                group_accessible = True
+
+        embed = Embed(title="Current tags")
+        await LinePaginator.paginate(result_lines, ctx, embed=embed, max_lines=15, empty=False, footer_text=FOOTER_TEXT)
+
+    async def list_tags_in_group(self, ctx: Context, group: str) -> None:
+        """Send a paginator with all tags under `group`, that are accessible by `ctx.author`."""
```
Numerlor, line 340: As it is, listing the top level group alone isn't possible because it lacks a group to look it up with. Would some kind of a string (`None`, `nogroup` or something like that) be worth using here to list them alone instead of only being accessible through the whole list

bot/exts/info/tags.py
```diff
@@ -24,145 +29,214 @@
 FOOTER_TEXT = f"To show a tag, type {constants.Bot.prefix}tags <tagname>."
 
 
+class COOLDOWN(enum.Enum):
+    """Sentinel value to signal that a tag is on cooldown."""
+
+    obj = object()
```
Bluenix2, line 35: What's the difference to this and `COOLDOWN = object()`?

I assume this has the benefit of `__repr__` right?

bot/exts/info/tags.py
```diff
@@ -24,145 +29,214 @@
...
+                matching_tags.append((identifier, tag))
 
         return matching_tags
 
-    async def _send_matching_tags(self, ctx: Context, keywords: str, matching_tags: list) -> None:
+    async def _send_matching_tags(
+            self,
+            ctx: Context,
+            keywords: str,
+            matching_tags: list[tuple[TagIdentifier, Tag]],
```
Bluenix2, line 210: We define this type a lot, isn't there a way to create an alias. As in:
```py
tag_pair = tuple[TagIdentifier, Tag]
```
Then this code can become:
```suggestion
            matching_tags: list[tag_pair],
```
May not be a huge benefit, but I thought I'd mention it.

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+    ) -> Optional[Union[Embed, Literal[COOLDOWN.obj]]]:
         """
-        If a tag is not found, display similar tag names as suggestions.
-
-        If a tag is not specified, display a paginated embed of all tags.
+        Generate an embed of the requested tag or of suggestions if the tag doesn't exist/isn't accessible by the user.
 
-        Tags are on cooldowns on a per-tag, per-channel basis. If a tag is on cooldown, display
-        nothing and return True.
+        If the requested tag is on cooldown or no suggestions were found, return None.
```
Bluenix2, line 267: This claims that if the tag is on cool-down that None will be returned.

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
-                and tag_name in self.tag_cooldowns
-                and (now - self.tag_cooldowns[tag_name]["time"]) < constants.Cooldowns.tags
-                and self.tag_cooldowns[tag_name]["channel"] == ctx.channel.id
+        filtered_tags = [
+            (ident, tag) for ident, tag in
+            self.get_fuzzy_matches(tag_identifier)[:10]
+            if tag.accessible_by(ctx.author)
+        ]
+
+        if (tag := self._tags.get(tag_identifier)) is None:
```
Bluenix2, line 275: I'd honestly place this outside of the if-statement to make it clear that `tag` is used after the if-statement, even though Python doesn't have strict scoping like JS for example.

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
-            time_elapsed = time.time() - self.tag_cooldowns[tag_name]["time"]
-            time_left = constants.Cooldowns.tags - time_elapsed
-            log.info(
-                f"{ctx.author} tried to get the '{tag_name}' tag, but the tag is on cooldown. "
-                f"Cooldown ends in {time_left:.1f} seconds."
+        else:
+            if not filtered_tags:
+                return None
+            suggested_tags_text = "\n".join(
+                f"**\N{RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK}** {identifier}"
```
Bluenix2, line 297: I assume that this is the Unicode name for `»`?

Either replace this with that character, or change the usage seen a bit further up with this. For consistency.

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
-                    self.tag_cooldowns[tag_name] = {
-                        "time": time.time(),
-                        "channel": ctx.channel.id
-                    }
 
-                self.bot.stats.incr(f"tags.usages.{tag['title'].replace('-', '_')}")
+    async def list_all_tags(self, ctx: Context) -> None:
+        """Send a paginator with all loaded tags accessible by `ctx.author`, groups first, and alphabetically sorted."""
+        def tag_sort_key(tag_item: tuple[TagIdentifier, tag]) -> str:
+            ident = tag_item[0]
```
Bluenix2, line 309: We may want to use tuple assignment here? Feel free to leave that as it is though.

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+            if ident.group is None:
+                # Max codepoint character to force tags without a group to the end
+                group = chr(0x10ffff)
+            else:
+                group = ident.group
+            return group+ident.name
+
+        result_lines = []
+        current_group = object()
```
Bluenix2, line 318: Is this also a sentinel value? Can you not use a empty string (I know that None is used when a tag doesn't have a group).

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+                if not group_accessible:
+                    # Remove group separator line if no tags in the previous group were accessible by the user.
+                    result_lines.pop()
+                # A new group began, add a separator with the group name.
+                if identifier.group is not None:
+                    group_accessible = False
+                    result_lines.append(f"\n\N{BULLET} **{identifier.group}**")
+                else:
+                    result_lines.append("\n\N{BULLET}")
+                current_group = identifier.group
```
Bluenix2, line 333: You could move this like next to the comment about "A new group began".

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+            if tag.accessible_by(ctx.author):
+                result_lines.append(f"**\N{RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK}** {identifier.name}")
+                group_accessible = True
+
+        embed = Embed(title="Current tags")
+        await LinePaginator.paginate(result_lines, ctx, embed=embed, max_lines=15, empty=False, footer_text=FOOTER_TEXT)
+
+    async def list_tags_in_group(self, ctx: Context, group: str) -> None:
+        """Send a paginator with all tags under `group`, that are accessible by `ctx.author`."""
+        embed = Embed(title=f"**Tags under *{group}***")
```
Bluenix2, line 344: Can an embed title be bold?

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+        tag_lines = sorted(
+            f"**\N{RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK}** {identifier}"
+            for identifier, tag in self._tags.items()
+            if identifier.group == group and tag.accessible_by(ctx.author)
+        )
+        await LinePaginator.paginate(tag_lines, ctx, embed, footer_text=FOOTER_TEXT, empty=False, max_lines=15)
```
Bluenix2, line 350: On line 340 you use a kwarg for `embed`, shouldn't this also be a kwarg then or is that unnecessary?

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+            tag_name: TagNameConverter = None,
+    ) -> bool:
+        """
+        Get a specified tag, or a list of all tags if no tag is specified.
 
-                await wait_for_deletion(
-                    await ctx.send(embed=Embed.from_dict(tag['embed'])),
-                    [ctx.author.id],
-                )
+        Returns True if something was sent, or if the tag is on cooldown.
```
Bluenix2, line 361: Could use `response`, or `message` as opposed to `something`.

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+        await LinePaginator.paginate(tag_lines, ctx, embed, footer_text=FOOTER_TEXT, empty=False, max_lines=15)
+
+    @tags_group.command(name="get", aliases=("show", "g"))
+    async def get_command(
+            self, ctx: Context,
+            tag_name_or_group: TagNameConverter = None,
+            tag_name: TagNameConverter = None,
+    ) -> bool:
+        """
+        Get a specified tag, or a list of all tags if no tag is specified.
```
Bluenix2, line 359: Should probably mention that this can display a tag group.

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+        embed = await self.get_tag_embed(ctx, TagIdentifier(tag_group, tag_name))
+        if embed is not None:
+            if embed is not COOLDOWN.obj:
+                await wait_for_deletion(
+                    await ctx.send(embed=embed),
+                    (ctx.author.id,)
+                )
+            return True
+        else:
+            return False
```
Bluenix2, line 394: Could probably write this thing as:
```py
if embed is None:
    return False

if embed is not COOLDOWN.obj:
    await wait_for_deletion(...)

return True
```

bot/exts/info/tags.py
```diff
@@ -172,125 +246,163 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
 
 
 def setup(bot: Bot) -> None:
     """Load the Tags cog."""
     bot.add_cog(Tags(bot))
+
+
+def extract_tag_identifier(string: str) -> TagIdentifier:
+    """Create a `TagIdentifier` instance from beginning of `string`."""
```
Bluenix2, line 403:
```suggestion
    """Create a `TagIdentifier` instance from the beginning of `string`."""
```

bot/exts/info/tags.py
ToxicKidz: Although type aliases should be in PascalCase

bot/exts/info/tags.py
Numerlor: I used an Enum because I couldn't find a way to type the sentinel alone

bot/exts/info/tags.py
Numerlor: Do you mean doing `group, name = tag_item[0]` to avoid the else case?

bot/exts/info/tags.py
Numerlor: It looks a bit nicer, but I'm not sure if it's as clear on when True is returned

bot/exts/info/tags.py
Numerlor: I suppose an empty string could be used, but I don't think there's any benefit of using it instead of a plain object like the current use

bot/exts/info/tags.py
Numerlor: Yes, that'd be nicer, although I don't think `TagPair` is a fitting name as I'd expect a pair of `Tag` objs from that, not sure on a good name myself

bot/exts/info/tags.py
Numerlor: Reworded things a bit in e40cd9840c188210a7cfde6bffbf08ce563c2a11

bot/exts/info/tags.py
Numerlor: I think this was a leftover from when I was prototyping and had it in an elif, moved to a normal assignment in e40cd9840c188210a7cfde6bffbf08ce563c2a11

bot/exts/info/tags.py
Numerlor: Moved the assignment to the comment and changed the uses to use the name directly instead of identifier.group in 5cbb7d20713c0a7027c2e0909ddcdb6e02dbddc2

bot/exts/info/tags.py
Numerlor: Looks like it can, but other embed titles were changed to not be bold and I think it was a bit harder to read so I removed the bolding in 96a8390290f379a90fa0f966b74993dd1ddf9f44

bot/exts/info/tags.py
Numerlor: Changed the other use to not use a kwarg in e3b9924920a32e19adfa1274cfdbc4cf9ddb874a

bot/exts/info/tags.py
Numerlor: Reworded things a bit in e40cd98

bot/exts/info/tags.py
Bluenix2: Could fix that with some comments, like `If we reached here it was successful` or `Wait for the cooldown message to be deleted before returning True`.

bot/exts/info/tags.py
Bluenix2: I was originally thinking about doing `ident, _ = tag_item` since `tag_item` should be treated more like a pair than a tuple.

But that sounds great doing `group, name = tag_item[0]`, I didn't think about TagIdentifier being a named tuple.

bot/exts/info/tags.py
```diff
@@ -24,145 +29,217 @@
...
+    obj = object()
+
+
+class TagIdentifier(NamedTuple):
+    """Stores the group and name used as an identifier for a tag."""
+
+    group: Optional[str]
+    name: str
+
+    def get_fuzzy_score(self, fuzz_tag_identifier: TagIdentifier) -> float:
```
Bluenix2, line 44: What's wrong with simply calling that `identifier` or `match`?

bot/exts/info/tags.py
```diff
@@ -24,145 +29,217 @@
...
+
+class Tag:
+    """Provide an interface to a tag from resources with `file_content`."""
+
+    def __init__(self, file_content: str):
+        post = frontmatter.loads(file_content)
+        self.content = post.content
+        self.metadata = post.metadata
+        self._restricted_to: set[int] = set(self.metadata.get("restricted_to", ()))
+        self._cooldowns: dict[discord.TextChannel, float] = {}
```
Bluenix2, line 80: Can't redis be used for this?

bot/exts/info/tags.py
```diff
@@ -24,145 +29,217 @@
...
+            else:
+                return self._get_suggestions(tag_identifier)
+        else:
+            if len(tag_identifier.group) < 3:
+                suggestions = []
+            else:
+                # Try fuzzy matching with only a name first
+                suggestions = self._get_suggestions(TagIdentifier(None, tag_identifier.group))
+
+            if not len(tag_identifier.name) < 3:
```
Bluenix2, line 170:
```suggestion
            if len(tag_identifier.name) > 3:
```

bot/exts/info/tags.py
```diff
@@ -172,125 +249,168 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
-        matching_tags = self._get_tags_via_content(any, keywords or 'any', ctx.author)
+        matching_tags = self._get_tags_via_content(any, keywords or "any", ctx.author)
         await self._send_matching_tags(ctx, keywords, matching_tags)
 
-    async def display_tag(self, ctx: Context, tag_name: str = None) -> bool:
+    async def get_tag_embed(
+            self,
+            ctx: Context,
+            tag_identifier: TagIdentifier,
+    ) -> Optional[Union[Embed, Literal[COOLDOWN.obj]]]:
```
Bluenix2, line 266: Pretty much the same thing, but we avoid some unnecessary nesting
```suggestion
    ) -> Union[Embed, Literal[COOLDOWN.obj], None]:
```

bot/exts/info/tags.py
```diff
@@ -172,125 +249,168 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
-                and self.tag_cooldowns[tag_name]["channel"] == ctx.channel.id
+        filtered_tags = [
+            (ident, tag) for ident, tag in
+            self.get_fuzzy_matches(tag_identifier)[:10]
+            if tag.accessible_by(ctx.author)
+        ]
+
+        tag = self._tags.get(tag_identifier)
+        if tag is None:
+            if len(filtered_tags) == 1:
```
Bluenix2, line 280: This can become an `and`.

bot/exts/info/tags.py
```diff
@@ -172,125 +249,168 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+                tag_identifier = filtered_tags[0][0]
+                tag = filtered_tags[0][1]
+
+        if tag is not None:
```
Bluenix2, line 284: Should we perhaps add a comment why this can't be an `else:`?

bot/exts/info/tags.py
```diff
@@ -172,125 +249,168 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
                 return True
 
-        else:
-            tags = self._cache.values()
-            if not tags:
-                await ctx.send(embed=Embed(
-                    description="**There are no tags in the database!**",
-                    colour=Colour.red()
-                ))
+        if tag_name is None:
```
Bluenix2, line 380: Make this an `elif` to be clear that either of these happen.

bot/exts/info/tags.py
```diff
@@ -172,125 +249,168 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+        else:
+            return False
 
 
 def setup(bot: Bot) -> None:
     """Load the Tags cog."""
     bot.add_cog(Tags(bot))
+
+
+def extract_tag_identifier(string: str) -> TagIdentifier:
```
Bluenix2, line 410: How does moving this to a classmethod sound?

bot/exts/info/tags.py
Numerlor: What purpose would it serve? We gather the data from the tag file and after that it's unchanged at runtime

Bluenix2: Specifically I am talking about the cooldowns but that may not be big enough to experience a benefit?

Numerlor: I don't think they're important enough to persist if that's what you mean

Bluenix2: No, Redis will not store them persistently.

Redis is an in-memory cache (key-value store). Pretty much exactly like this dict.

I would like to hear others thought about it, as I am no expert on Redis, but I suggest you look into it if you're unaware what Redis is.

Numerlor: I'm aware of what it is, but I fail to see how it could be used here

bot/exts/info/tags.py
Numerlor: Sounds good, made it a constructor of the class in 359a17c00bb036f9a90fe7681592dee29f84c806

bot/exts/info/tags.py
Numerlor: Nothing wrong, but I wanted to make it clear that the method is fuzzy matching the passed in identifier with self as the target

bot/exts/info/tags.py
Numerlor: I find the Optional a bit nicer as it's immediately obvious that it either returns None or some other value
I can change it if others agree though

bot/exts/info/tags.py
Numerlor: I think it should be clear with `tag` being changed on the line right before the condition

bot/exts/info/tags.py
```diff
@@ -24,145 +29,227 @@
...
+
+    def get_fuzzy_score(self, fuzz_tag_identifier: TagIdentifier) -> float:
+        """Get fuzzy score, using `fuzz_tag_identifier` as the identifier to fuzzy match with."""
+        if self.group is None:
+            if fuzz_tag_identifier.group is None:
+                # We're only fuzzy matching the name
+                group_score = 1
+            else:
+                # Ignore tags without groups if the identifier contains a group
+                return .0
```
Bluenix2, line 52: Something I am a bit on the fence about, do we prefer `0.0` here?

bot/exts/info/tags.py
```diff
@@ -24,145 +29,227 @@
...
+    def accessible_by(self, member: discord.Member) -> bool:
+        """Check whether `member` can access the tag."""
+        return bool(
+            not self._restricted_to
+            or self._restricted_to & {role.id for role in member.roles}
+        )
+
+    def on_cooldown_in(self, channel: discord.TextChannel) -> bool:
+        """Check whether the tag is on cooldown in `channel`."""
+        return channel in self._cooldowns and self._cooldowns[channel] > time.time()
```
Bluenix2, line 108: How does this work over midnight, considering it is only time.

bot/exts/info/tags.py
Bluenix2: Well there's a benefit in annotating the code

bot/exts/info/tags.py
```diff
@@ -172,123 +259,156 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+
+    @tags_group.command(name="get", aliases=("show", "g"))
+    async def get_command(
+            self, ctx: Context,
+            tag_name_or_group: TagNameConverter = None,
+            tag_name: TagNameConverter = None,
+    ) -> bool:
+        """
+        When arguments are passed in:
+        If a single argument is given and it matches a group name, list accessible all tags from that group.
```
Bluenix2, line 372: I think this sentence needs revising.

bot/exts/info/tags.py
wookie184: `time.time` is seconds since the epoch, https://docs.python.org/3/library/time.html#time.time

bot/exts/info/tags.py
wookie184: If I was writing it myself i'd probably write `0.0`, but both ways are valid and I don't think our style guide indicates any preferred way, so probably doesn't need changing.

bot/exts/info/tags.py
Numerlor: I'm open to any suggestions.
Is "If a single argument matching a group name is given, list accessible all tags from that group." better?

bot/exts/info/tags.py
```diff
@@ -24,145 +29,227 @@
...
+    _targets = iter(REGEX_NON_ALPHABET.split(target.lower()))
+    _target = next(_targets)
+    try:
+        while True:
+            while index < len(_target) and _search[current] == _target[index]:
+                current += 1
+                index += 1
+            index, _target = 0, next(_targets)
+    except (StopIteration, IndexError):
+        pass
```
Bluenix2, line 128:
```suggestion
    for _target in _targets:
        while index < len(_target) and _search[current] == _target[index]:
            current += 1
            index += 1
        index = 0
```

bot/exts/info/tags.py
```diff
@@ -24,145 +29,227 @@
...
+        return channel in self._cooldowns and self._cooldowns[channel] > time.time()
+
+    def set_cooldown_for(self, channel: discord.TextChannel) -> None:
+        """Set the tag to be on cooldown in `channel` for `constants.Cooldowns.tags` seconds."""
+        self._cooldowns[channel] = time.time() + constants.Cooldowns.tags
+
+
+def _fuzzy_search(search: str, target: str) -> float:
+    """A simple scoring algorithm based on how many letters are found / total, with order in mind."""
+    current, index = 0, 0
```
Bluenix2, line 117: Perhaps we can move this down and remove the index assignment, we only need it in the loop.

bot/exts/info/tags.py
```diff
@@ -172,123 +259,156 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
-
-                self.bot.stats.incr(f"tags.usages.{tag['title'].replace('-', '_')}")
 
-                await wait_for_deletion(
-                    await ctx.send(embed=Embed.from_dict(tag['embed'])),
-                    [ctx.author.id],
-                )
+    async def list_all_tags(self, ctx: Context) -> None:
+        """Send a paginator with all loaded tags accessible by `ctx.author`, groups first, and alphabetically sorted."""
+        def tag_sort_key(tag_item: tuple[TagIdentifier, tag]) -> str:
```
Bluenix2, line 321:
```suggestion
        def tag_sort_key(tag_item: tuple[TagIdentifier, Tag]) -> str:
```

bot/exts/info/tags.py
Bluenix2: Specifically I was first talking about "list accessible all tags", but I like what you suggested better than what is currently there. So "If a single argument matching a group name is given, list all accessible tags from that group." it is

Numerlor: Oh I didn't even notice that, will fix

bot/exts/info/tags.py
```diff
@@ -24,145 +29,225 @@
...
+        """Create an embed for the tag."""
+        embed = Embed.from_dict(self.metadata.get("embed", {}))
+        embed.description = self.content
+        return embed
+
+    def accessible_by(self, member: discord.Member) -> bool:
+        """Check whether `member` can access the tag."""
+        return bool(
+            not self._restricted_to
+            or self._restricted_to & {role.id for role in member.roles}
```
Bluenix2, line 103: This is actually a bit sad when you think about it haha

Discord.py uses an internal array to store role IDs attached to a member, `.roles` is a property that gets all of the roles from cache by their ID. We then use a set comprehension to convert it to a set of.. role IDs.

bot/exts/info/tags.py
```diff
@@ -24,145 +29,225 @@
...
+    """A simple scoring algorithm based on how many letters are found / total, with order in mind."""
+    _search = REGEX_NON_ALPHABET.sub("", search.lower())
+    _targets = iter(REGEX_NON_ALPHABET.split(target.lower()))
+
+    current, index = 0, 0
+    for _target in _targets:
+        while index < len(_target) and _search[current] == _target[index]:
+            current += 1
+            index += 1
+        index = 0
```
Bluenix2, line 125:
```suggestion
    current = 0
    for _target in _targets:
        index = 0
        while index < len(_target) and _search[current] == _target[index]:
            current += 1
            index += 1
```

Bluenix2 (commented):
Looks good otherwise

bot/exts/info/tags.py
```diff
@@ -172,123 +261,156 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+
+        if tag is not None:
+            if tag.on_cooldown_in(ctx.channel):
+                log.debug(f"Tag {str(tag_identifier)!r} is on cooldown.")
+                return COOLDOWN.obj
+            tag.set_cooldown_for(ctx.channel)
+
+            self.bot.stats.incr(
+                f"tags.usages"
+                f"{'.' + tag_identifier.group.replace('-', '_') if tag_identifier.group else ''}"
```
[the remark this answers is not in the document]
Bluenix2: Oh I thought I deleted this comment, but yes it would behave differently. https://github.com/python-discord/bot/pull/1663#discussion_r697746775 would be equivalent.

But I am not that picky on the subject, only the usage of `f"tags.usages"` being an f-string though I guess that leads to consistency.

bot/exts/info/tags.py
```diff
@@ -24,145 +29,229 @@
...
+                group_score = 1
+            else:
+                # Ignore tags without groups if the identifier contains a group
+                return .0
+        else:
+            if fuzz_tag_identifier.group is None:
+                # Ignore tags with groups if the identifier does not have a group
+                return .0
+            else:
+                group_score = _fuzzy_search(fuzz_tag_identifier.group, self.group)
```
Bluenix2, line 58: I was able to simplify this code a bit considering there's 3 cases:
- One has a group and the other doesn't (XOR)
- Identical groups (potentially both None)
- Different strings and should be fuzzy searched

```suggestion
        if self.group is None != fuzzy_tag_identifier.group is None:
            # Ignore tags without groups if the identifier has a group and vice versa
            return .0

        if self.group == fuzzy_tag_identifier.group:
            # Completely identical, or both None
            group_score = 1
        else:
            group_score = _fuzzy_search(fuzz_tag_identifier.group, self.group)
```

bot/exts/info/tags.py
```diff
@@ -24,145 +29,229 @@
...
+    def accessible_by(self, member: discord.Member) -> bool:
+        """Check whether `member` can access the tag."""
+        return bool(
+            not self._restricted_to
+            or self._restricted_to & {role.id for role in member.roles}
+        )
+
+    def on_cooldown_in(self, channel: discord.TextChannel) -> bool:
+        """Check whether the tag is on cooldown in `channel`."""
+        return channel in self._cooldowns and self._cooldowns[channel] > time.time()
```
Bluenix2, line 108: This makes two lookups, what about the following instead?
```suggestion
        if cooldown := self._cooldowns.get(channel):
            return cooldown > time.time()

        return False
```

bot/exts/info/tags.py
```diff
@@ -24,145 +29,229 @@
...
-        cache = {}
+        self.tags: dict[TagIdentifier, Tag] = {}
+        self.initialize_tags()
 
+    def initialize_tags(self) -> None:
+        """Load all tags from resources into `self.tags`."""
         base_path = Path("bot", "resources", "tags")
+
         for file in base_path.glob("**/*"):
             if file.is_file():
```
Bluenix2, line 147: Technically we can skip a level of indentation if we invert this

bot/exts/info/tags.py
```diff
@@ -24,145 +29,229 @@
...
+            if len(tag_identifier.name) < 3:
+                return []
+            else:
+                return self._get_suggestions(tag_identifier)
+        else:
+            if len(tag_identifier.group) < 3:
+                suggestions = []
+            else:
+                # Try fuzzy matching with only a name first
+                suggestions = self._get_suggestions(TagIdentifier(None, tag_identifier.group))
```
Bluenix2, line 180:
```suggestion
        if len(tag_identifier.name) < 3 or len(tag_identifier.group) < 3:
            return []
    
        if tag_identifier.group is None:
            return self._get_suggestions(tag_identifier)
        else:
            # Try fuzzy matching with only a name first
            suggestions = self._get_suggestions(TagIdentifier(None, tag_identifier.group))
```

bot/exts/info/tags.py
```diff
@@ -172,123 +261,155 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+            self, ctx: Context,
+            tag_name_or_group: TagNameConverter = None,
+            tag_name: TagNameConverter = None,
+    ) -> bool:
+        """
+        If a single argument matching a group name is given, list all accessible tags from that group
+        Otherwise display the tag if one was found for the given arguments, or try to display suggestions for that name
 
-                self.bot.stats.incr(f"tags.usages.{tag['title'].replace('-', '_')}")
+        With no arguments, list all accessible tags
```
Bluenix2, line 376:
```suggestion
        With no arguments, list all accessible tags.
```

bot/exts/info/tags.py
```diff
@@ -172,123 +261,155 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+
+    @tags_group.command(name="get", aliases=("show", "g"))
+    async def get_command(
+            self, ctx: Context,
+            tag_name_or_group: TagNameConverter = None,
+            tag_name: TagNameConverter = None,
+    ) -> bool:
+        """
+        If a single argument matching a group name is given, list all accessible tags from that group
+        Otherwise display the tag if one was found for the given arguments, or try to display suggestions for that name
```
Bluenix2, line 374:
```suggestion
        Otherwise display the tag if one was found for the given arguments, or try to display suggestions for that name.
```

bot/exts/info/tags.py
Numerlor: The  `self.group is None != fuzzy_tag_identifier.group is None:` comparison won't work. we'd need to do something like (not sure if there's any nicer way to check that either of  them is None and they aren't equal)
```suggestion
        if (self.group is None or fuzz_tag_identifier.group is None) and self.group != fuzz_tag_identifier.group:
            # Ignore tags without groups if the identifier has a group and vice versa.
            return .0
        if self.group == fuzz_tag_identifier.group:
            # Completely identical, or both None.
            group_score = 1
        else:
            group_score = _fuzzy_search(fuzz_tag_identifier.group, self.group)
```
I think that's a bit easier to read overall but the first condition could be confusing at a glance

Bluenix2: No `self.group is None != fuzzy_tag_identifier.group is None:` will work, because they become bools. Bools only have 2 states, True or False. So we have 4 cases:
- both `self.group` and `fuzzy_tag_identifier.group` are None (`True != True` *False*)
- `self.group` is None and `fuzzy_tag_identifer.group` is set (`True != False` *True*)
- `self.group` is set and `fuzzy_tag_identifier.group` is None (`False != True` *True*)
- both `self.group` and `fuzzy_tag_identifier.group` are set (`False != False` *False*)

We want to return when it is True, which is exaclty what is happening here.

Numerlor: With how it's chained, it impossible for the condition to be True `self.group is None != fuzzy_tag_identifier.group is None` creates a `!=` comparison between `None` and `fuzzy_tag_identifier.group` which is then compared to be equal to `None`. The long equivalent (and how python handles it) would be 
`self_group is None and None != fuzzy_tag_identifier_group and fuzzy_tag_identifier_group is None`

```py
In [149]: for self_group, fuzzy_tag_identifier_group in itertools.product((None, "ba"), repeat=2):
     ...:     print(self_group, fuzzy_tag_identifier_group)
     ...:     print(self_group is None != fuzzy_tag_identifier_group is None)
     ...:     print((self_group is None or fuzzy_tag_identifier_group is None) and self_group != fuzzy_tag_identifier_g
     ...: roup)
     ...:     print()
     ...:
None None
False
False

None ba
False
True

ba None
False
True

ba ba
False
False
```

Bluenix2: Oh I see, so just add the parenthesis: `(self.group is None) != (fuzzy_tag_identifier.group is None)`

bot/exts/info/tags.py
Numerlor: these do not act the same, the current implementation allows groups shorter than 3 chars to be used to invoke tags as long as the name is longer than 3 and matches

bot/exts/info/tags.py
Numerlor: Ah, yes that should work, sorry I couldn't make sense of the comparison before

bot/exts/info/tags.py
Numerlor: I think doing the containment is a bit more readable and less verbose; I'm not too worried about the two checks as they're negligible in runtime to pretty much anything else that's happening in the bot.

We could also go with -inf as the default for a get and compare with that but again imo doing a normal containment check is the most explicit to show what's going on there

bot/exts/info/tags.py
Numerlor: The nesting is not too bad and there's nothing after it so I think it's fine as is

bot/exts/info/tags.py
Bluenix2: I think that actually sounds like a great idea
```suggestion
        return self._cooldowns.get(channel, float('inf')) > time.time()
```

bot/exts/info/tags.py
Bluenix2: Hmm, I see now. What about this instead?

We're removing a level of nesting in the if-statements introducing several dimensions to keep track, by flattening them we get a clear case of when to do what
```suggestion
        if len(tag_identifier.name) < 3:
            return []
        elif tag_identifier.group is None:
            return self._get_suggestions(tag_identifier)

        if len(tag_identifier.group) < 3:
            # No point in fuzzy matching if the group is less than 3 characters
            suggestions = []
        else:
            # Try fuzzy matching with only a name first
            suggestions = self._get_suggestions(TagIdentifier(None, tag_identifier.group))
```
Like this we have a pretty clear view..

- If the name is less then 3 characters we return immediately
- If the group is None we can just fuzzy match immediately and return
*I cut them in half here because the first two exit early, the second two continue down the function*
- If the group is less then 3 characters there's no point in fuzzy matching it and just continue with an empty list
- If name is longer then 3 characters, and we have a group over 3 character we fuzzy match the group and continue

Numerlor: This won't let us match only with a name if a tag is matched with a long group but a short name, I'm on the edge of that being worth it for the increased code complexity.
On one hand it's relatively unlikely to happen as `!<name> <2chars>` would have to be invoked (maybe an emoji would be an actual case where that can happen) but then the code is also not *that* much less complex without the first if taking care of that case and then the name length check at the end

Bluenix2: > This won't let us match with only a name if a tag is matched with a long group but a short name

I see now, so we just remove that if-statement and leave it to the code below this. That way we will at least match for groups yeah.
```suggestion
        if tag_identifier.group is None:
            return self._get_suggestions(tag_identifier)

        if len(tag_identifier.group) < 3:
            # No point in fuzzy matching if the group is less than 3 characters
            suggestions = []
        else:
            # Try fuzzy matching with only a name first
            suggestions = self._get_suggestions(TagIdentifier(None, tag_identifier.group))
```

Numerlor: That would skip the name length check altogether, allowing fuzzying of invocations like `!de` which may just be a typo

Bluenix2: No it wouldn't, check the code below this comment.

Numerlor: The code below won't run when the group is None

Bluenix2: Ah right, so we can special-case it and just add a second `and len(tag.name) >= 3:`

Numerlor: With that `None` groups would go beyond the check for short names

Bluenix2: I am starting to see why the if-statements look exactly like they do- because I tried to write a suggestion for the code to be exactly the same. That said I came up with something like this:
```py
        if tag_identifier.group is None and len (tag_identifier.name) >= 3:
            return self._get_suggestions(tag_identifier)

        if tag_identifier.group is None or len(tag_identifier.group) < 3:
            # No point in fuzzy matching if the group is less than 3 characters or None
            suggestions = []
        else:
            # Try fuzzy matching with only a name first
            suggestions = self._get_suggestions(TagIdentifier(None, tag_identifier.group))
```
But at this point, this suggestion can be refactored to consider the if-statements below it.

Bluenix2:
```
        suggestions = []

        if tag_identifier.group is not None and len(tag_identifier.group) < 3:
            # Try fuzzy matching with only a name first
            suggestions += self._get_suggestions(TagIdentifier(None, tag_identifier.group))

            if len(tag_identifier.name) >= 3:
                suggestions += self._get_suggestions(tag_identifier)

            return suggestions
```
I can't use the GitHub suggestions for this, because that would duplicate code. But what do you think about this?

Numerlor: That should work according to my tests after fixing the inequality in the first if, pushed it in ba2a74ac18f9e40fad917d51ffa2238a340416c0

Numerlor:
I decided to remove the converter for tag names in f5db0b8 as after the removal of adding tags through the bot, it just restricted the input which can safely be ignored by the cog and with the new handling it could refuse tags when parsing a message even if the unparseable part of the message wasn't a part of the tag identifier.

An error caused by that was fixed in f3634f9dbcb6c7cc6952f7d9e40879518c4e6eb1 but I believe that should be everything that could have problems with input that wouldn't pass through the converter before

Bluenix2:
Sending a reminder about [these messages from the Discord server](https://discord.com/channels/267624335836053506/635950537262759947/885576813278011452) so that it is also on record in the PR.

Numerlor:
> Sending a reminder about [these messages from the Discord server](https://discord.com/channels/267624335836053506/635950537262759947/885576813278011452) so that it is also on record in the PR.

Cleaned it up a bit in https://github.com/python-discord/bot/pull/1663/files/e07febdbde4815466b161b535f4a5eaf3593f755..3edce6120b48cf83f8a87f34da1007221967232c after I realized the `TagIdentifier.from_string` handles part of the logic and can be used instead. I've also moved all the paginating into the command itself, which makes it a bit busier but I think the interface of the methods that used it makes more sense in that way when it also returns like in your suggestion.

Unfortunately there's one issue with the new handling where I'm not sure how to modify the command help of the group so it's formatted like the get command through the `usage` kwarg

bot/exts/info/tags.py
```diff
@@ -172,123 +248,146 @@ async def search_tag_content(self, ctx: Context, *, keywords: str) -> None:
...
+
+    def accessible_tags_in_group(self, group: str, user: discord.Member) -> list[str]:
+        """Return a formatted list of tags in `group`, that are accessible by `user`."""
+        return sorted(
+            f"**\N{RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK}** {identifier}"
+            for identifier, tag in self.tags.items()
+            if identifier.group == group and tag.accessible_by(user)
+        )
+
+    @tags_group.command(name="get", aliases=("show", "g"), usage="[tag_group] [tag_name]")
```
Numerlor, line 350: Do you think `[[tag_group] <tag_name>]"` would make more sense for the usage? to highlight that they're either both optional, or the name is required and the group is optional

Bluenix2: I don't think people read these carefully enough for it to matter and if they did then that would simply be confusing.

I am open for the discussion but what we have right now seems good enough.

Bluenix2:
> Unfortunately there's one issue with the new handling where I'm not sure how to modify the command help of the group so it's formatted like the get command through the `usage` kwarg
> 

Reading this back, I don't think I follow exactly what issue you're referencing. The brief is the first line in the docstring, so if you can change the docstring to fit enough information in one line then that will work.

Otherwise override the brief (assuming that line is the brief from discord.py).

Numerlor:
> Reading this back, I don't think I follow exactly what issue you're referencing. The brief is the first line in the docstring, so if you can change the docstring to fit enough information in one line then that will work.
> 
> Otherwise override the brief (assuming that line is the brief from discord.py).

I was talking about the signature specifically which is a property on the class object, but after reading through the source it became apparent that specifying `usage` does work on groups too and I probably just didn't relaunch the bot or something like that so I through it didn't.
resolved it through a fixup on b56c340

Bluenix2 (approved):
After countless code-reviews, I have played around with this - very good job here!

Frontmatter, groups and cooldowns all working well! Thanks Numerlor. There are some conflicts to resolve before this can be merged though

tests/bot/exts/backend/test_error_handler.py
```diff
@@ -359,38 +357,12 @@ async def test_try_get_tag_command_error(self):
...
-        """Should call `ctx.invoke` with proper args/kwargs."""
-        self.ctx.reset_mock()
-        self.ctx.invoked_with = "foo"
-        self.assertIsNone(await self.cog.try_get_tag(self.ctx))
-        self.ctx.invoke.assert_awaited_once_with(self.tag.get_command, tag_name="foo")
-
     async def test_dont_call_suggestion_tag_sent(self):
         """Should never call command suggestion if tag is already sent."""
-        self.ctx.invoked_with = "foo"
+        self.ctx.message = MagicMock(content="foo")
```
onerandomusername, line 365: tests/helpers.py has a `MockMessage`, we should use that.

Bluenix2:
Oh I thought this was merged already.. kind of surprised it isn't.

onerandomusername:
> Oh I thought this was merged already.. kind of surprised it isn't.

Same here, it looked like it was going to be merged with the conflicts above sorted.

jchristgit (approved):
:shipit:

PR #1666, opened 2021-07-02
Make use of the new Discord features

Akarys42:
- [X] Bump embed limit to 4096 characters
- [x] Use localized timestamps


Akarys42:
This is now ready for review! That said since iOS doesn't have support for timestamps yet, I'm leaving this PR in a draft state for the time being.

Xithrius (commented):
Should we consider having the integer `4096` as a constant?

Akarys42:
> Should we consider having the integer `4096` as a constant?

The thing is some modules needs `4096`, some `4093`, `4090` or even `4000`. I think it would have been worth it from the start, now I doubt discord will ever change it again. I don't believe it is worth changing and retesting everything to use `4096`, minus some magic numbers in some cases.

Xithrius:
Agreed. These numbers will remain where they are.

Akarys42:
It seems like timestamps have been rolled on iOS 🥳

It can take a bit to be available everywhere, but by the time we merge this PR we should be fine.

Akarys42:
It seems like timestamps have been rolled on iOS 🥳

It can take a bit to be available everywhere, but by the time we merge this PR we should be fine.

tests/bot/exts/info/test_information.py
```diff
@@ -347,7 +346,7 @@ async def test_create_user_embed_expanded_information_in_moderation_channels(
 
         self.assertEqual(
             textwrap.dedent(f"""
-                Created: {"1 year ago"}
+                Created: {"<t:1:R>"}
```
Xithrius, line 349: When `<t:1:R>` is typed out in Discord, it seems that it converts to 52 years ago. Is that supposed to be correct here?

Akarys42: Since the decoded timestamp is 1, yes. We used to patch the function that gets the formatted timestamp to return `1 year ago` but since it is gone, we can just assert that it constructed a valid timestamp.

Xithrius: I see. Thank you for this information.

Xithrius (approved):

vcokltfre (approved):
All looks good to me, looking forward to the new features being in production!

PR #1669, opened 2021-07-05
Adds Documentation For Running A Single Test

HassanAbouelela:
Adds a portion to the testing README explaining how and when to run an individual test file when working with tests.

Additionally adds a table of contents as the document has become quite long.

I used nocov instead of the regular version in the documentation as the options seemed to interfere with the selection (using cov and cov-report leads to pytest not using the passed path, and instead collecting all the tests).

These changes do not cover how to further specify tests or test suites from a file, as it had already gotten quite long, and the time improvements gained from narrowing down to a specific test or suite are not as monumental as narrowing down to a single file. Interested users can figure it out though.


Xithrius (approved):
All hyperlinks work, I see no errors in the English language.

Very well.

ToxicKidz (approved):
:shipit:

PR #1670, opened 2021-07-06
Disable filter in codejam team channels

wookie184:
This disables `filter_invites` in the code jam team channels 

Code jam team channels are based on the category name `Code Jam` used by the jams cog, so we shouldn't put any public channels under a category with that name. This code will likely be replaced by the new filters system in the near future so this should be fine anyway.

### DIFF
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -283,7 +283,7 @@ class Filtering(Cog):
                             continue
 
                     if filter_name == "filter_invites":
-                        # Disable domains filter in codejam team channels
+                        # Disable invites filter in codejam team channels
                         category = msg.channel.category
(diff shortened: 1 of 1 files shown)


bot/exts/filters/filtering.py
```diff
@@ -281,6 +282,12 @@ async def _filter_message(self, msg: Message, delta: Optional[int] = None) -> No
                         if delta is not None and delta < 100:
                             continue
 
+                    if filter_name == "filter_invites":
+                        # Disable domains filter in codejam team channels
```
ChrisLovering, line 286:
```suggestion
                        # Disable invites filter in codejam team channels
```

ChrisLovering (changes requested):

ChrisLovering (approved):

mbaruh (approved):

PR #1678, opened 2021-07-08
Add `Leads` to Server info Embed.

### LINKED ISSUE #1667: Adding a "leads" role to the server information embed.
Currently when the `!server` command is used, we get an embed like the one below. Since a stage channel has been added recently, the roles/channels listing is now uneven.

I suggest that we combine the project leads and domain leads role into a singular role called "leads", then place it somewhere within the left listing of roles.

image

D0rs4n:
Closes #1667 

After I proposed a solution on _Discord_, I made a **possible** implementation, I tried to make it readable by creating a helper function that might be useful later on if you're planning to add other groups. Although I'm not fully sure about it.


JacobMonck (approved):
👍

jchristgit (approved):

jchristgit:
One small comment I also wanted to mention was that the `join_role_stats` function in its current form is not very useful for other code, returning a dictionary with the name that was passed in is a bit odd. That said, it's perfectly reasonable and useful for the code here, so I don't think we should generalize it at this point in time

bot/utils/helpers.py
```diff
@@ -30,3 +31,11 @@ def has_lines(string: str, count: int) -> bool:
 def pad_base64(data: str) -> str:
     """Return base64 `data` with padding characters to ensure its length is a multiple of 4."""
     return data + "=" * (-len(data) % 4)
+
+
+def join_role_stats(role_ids: List[int], name: str, guild: Guild) -> Dict[str, int]:
+    """Return a dict object with the number of `members` of each role given, and the `name` for this joined group."""
+    members = []
+    for role_id in role_ids:
+        members += guild.get_role(role_id).members
```
Xithrius, line 40: `guild.get_role(role_id)` can fail, leading to the following error:
```python
bot_1       | Traceback (most recent call last):
bot_1       |   File "/usr/local/lib/python3.9/site-packages/discord/ext/commands/core.py", line 85, in wrapped
bot_1       |     ret = await coro(*args, **kwargs)
bot_1       |   File "/bot/bot/exts/info/information.py", line 194, in server_info
bot_1       |     member_counts = self.get_member_counts(ctx.guild)
bot_1       |   File "/bot/bot/exts/info/information.py", line 57, in get_member_counts
bot_1       |     **join_role_stats([constants.Roles.project_leads, constants.Roles.domain_leads], "Leads", guild))
bot_1       |   File "/bot/bot/utils/helpers.py", line 40, in join_role_stats
bot_1       |     members += guild.get_role(role_id).members
bot_1       | AttributeError: 'NoneType' object has no attribute 'members'
```

Xithrius, line 40: `guild.get_role(role_id)` can return `None`, leading to the following error:
```python
bot_1       | Traceback (most recent call last):
bot_1       |   File "/usr/local/lib/python3.9/site-packages/discord/ext/commands/core.py", line 85, in wrapped
bot_1       |     ret = await coro(*args, **kwargs)
bot_1       |   File "/bot/bot/exts/info/information.py", line 194, in server_info
bot_1       |     member_counts = self.get_member_counts(ctx.guild)
bot_1       |   File "/bot/bot/exts/info/information.py", line 57, in get_member_counts
bot_1       |     **join_role_stats([constants.Roles.project_leads, constants.Roles.domain_leads], "Leads", guild))
bot_1       |   File "/bot/bot/utils/helpers.py", line 40, in join_role_stats
bot_1       |     members += guild.get_role(role_id).members
bot_1       | AttributeError: 'NoneType' object has no attribute 'members'
```
This is also said to the user:
image

Xithrius (changes requested):
Changes look pretty much good, but it seems that if a role doesn't exist, the command will fail.

bot/utils/helpers.py
D0rs4n: Okay, I pushed some changes and this should not be a problem anymore.

bot/utils/helpers.py
```diff
@@ -37,5 +37,8 @@ def join_role_stats(role_ids: List[int], name: str, guild: Guild) -> Dict[str, i
     """Return a dict object with the number of `members` of each role given, and the `name` for this joined group."""
     members = []
     for role_id in role_ids:
-        members += guild.get_role(role_id).members
+        if (role := guild.get_role(role_id)) is None:
+            raise BadArgument("Unable to fetch role data, the specified role does not exist.")
```
Xithrius, line 41: Error should be more specific. If this were to go off, neither the user or anyone looking at the logs would exactly be able to identify which role is missing.
```suggestion
            raise ValueError(f"Could not fetch data for role {role} for server embed.")
```

Xithrius (changes requested):
We should be good after this change.

bot/utils/helpers.py
```diff
@@ -30,3 +31,14 @@ def has_lines(string: str, count: int) -> bool:
 def pad_base64(data: str) -> str:
     """Return base64 `data` with padding characters to ensure its length is a multiple of 4."""
     return data + "=" * (-len(data) % 4)
+
+
+def join_role_stats(role_ids: List[int], name: str, guild: Guild) -> Dict[str, int]:
```
Numerlor, line 36: Like @jchristgit mentioned this is not really useful for other code so I don't think it fits into this module as a general util. I think having it in the information module or cog directly would make a bit more sense.

The bot is also running on python 3.9 so the generics f rom typing aren't required and builtins can be used directly

bot/exts/info/information.py
```diff
@@ -50,7 +52,10 @@ def get_member_counts(guild: Guild) -> Dict[str, int]:
                 constants.Roles.owners, constants.Roles.contributors,
             )
         )
-        return {role.name.title(): len(role.members) for role in roles}
+        role_stats = {role.name.title(): len(role.members) for role in roles}
+        role_stats.update(
+            **join_role_stats([constants.Roles.project_leads, constants.Roles.domain_leads], "Leads", guild))
```
Numerlor, line 57:
```suggestion
        role_stats.update(
            **join_role_stats([constants.Roles.project_leads, constants.Roles.domain_leads], "Leads", guild)
        )
```

Numerlor (changes requested):

bot/utils/helpers.py
Numerlor: Looking at it now, the code in `get_member_counts` that runs before the call to this function doesn't have any error handling and will raise an attribute error for non existing roles. 
We should make that consistent with not handling the same error (for different roles) here, raising a common error for both cases, or providing a default so the command can be used in servers that aren't completely configured

bot/exts/info/information.py
```diff
@@ -41,6 +42,17 @@ def get_channel_type_counts(guild: Guild) -> DefaultDict[str, int]:
 
         return channel_counter
 
+    @staticmethod
+    def join_role_stats(role_ids: List[int], name: str, guild: Guild) -> Dict[str, int]:
```
Numerlor, line 46:
```suggestion
    def join_role_stats(role_ids: list[int], name: str, guild: Guild) -> dict[str, int]:
```
Like I mentioned we can use the builtin generics now, you can also remove the List import with this

bot/exts/info/information.py
```diff
@@ -19,6 +19,7 @@
 from bot.utils.checks import cooldown_with_role_bypass, has_no_roles_check, in_whitelist_check
 from bot.utils.time import TimestampFormats, discord_timestamp, humanize_delta
 
+
```
Numerlor, line 22: Why was the line added here? It's inconsistent with other modules and creates an unnecessary diff

Numerlor (changes requested):

Numerlor:
I've also added a comment to the error conversation above that'd probably be addressed https://github.com/python-discord/bot/pull/1678#discussion_r679544471 not sure if it was collapsed before

bot/exts/info/information.py
D0rs4n: Most of the code still uses the typing module, I thought I should change it. But I'll change it then.

bot/utils/helpers.py
D0rs4n: I added a custom error to handle both cases!

bot/exts/info/information.py
```diff
@@ -42,15 +43,32 @@ def get_channel_type_counts(guild: Guild) -> DefaultDict[str, int]:
...
-                constants.Roles.owners, constants.Roles.contributors,
-            )
+        role_ids = [constants.Roles.helpers, constants.Roles.mod_team, constants.Roles.admins,
+                    constants.Roles.owners, constants.Roles.contributors]
+        roles = []
+        for role_id in role_ids:
+            if (role := guild.get_role(role_id)) is None:
+                raise NonExistentRoleError(role_id)
+            else:
+                roles.append(role)
```
Bluenix2, line 66: There's some unnecessary code-duplication here.

D0rs4n: I was reminded that if there is a terrible misconfiguration and those essential roles, such as helpers, or owners do not exist this bit of code would fail, so I added an additional check. In case of the other bit of code (`join_role_stats`) the roles that need to be joined might also not exist so I had to add another check)
I'm not sure how to get around this, the roles that needs to be joined must be handled differently than the ones that don't.

bot/exts/info/information.py
```diff
@@ -42,15 +43,33 @@ def get_channel_type_counts(guild: Guild) -> DefaultDict[str, int]:
...
+        roles = []
+        for role_id in role_ids:
+            if (role := guild.get_role(role_id)) is not None:
+                roles.append(role)
+            else:
+                raise NonExistentRoleError(role_id)
+
+        role_stats = {role.name.title(): len(role.members) for role in roles}
+        role_stats.update(
+            **Information.join_role_stats([constants.Roles.project_leads, constants.Roles.domain_leads], "Leads", guild)
```
Bluenix2, line 70: I actually don't remember if this unpacking is necessary.

Additionally, you can reach this staticmethod through `self`.

bot/exts/info/information.py
```diff
@@ -42,15 +43,33 @@ def get_channel_type_counts(guild: Guild) -> DefaultDict[str, int]:
         return channel_counter
 
     @staticmethod
-    def get_member_counts(guild: Guild) -> Dict[str, int]:
+    def join_role_stats(role_ids: list[int], name: str, guild: Guild) -> dict[str, int]:
```
Bluenix2, line 46: We should be able to make this useful other than to Leads.
```suggestion
    def join_role_stats(role_ids: list[int], guild: Guild, name: Optional[str] = None) -> dict[str, int]:
```

bot/exts/info/information.py
```diff
@@ -42,15 +43,33 @@ def get_channel_type_counts(guild: Guild) -> DefaultDict[str, int]:
...
+        """Return a dictionary with the number of `members` of each role given, and the `name` for this joined group."""
+        members = []
+        for role_id in role_ids:
+            if (role := guild.get_role(role_id)) is not None:
+                members.append(role.members)
+            else:
+                raise NonExistentRoleError(role_id)
+        return {name: len(set(members))}
```
Bluenix2, line 54:
```suggestion
        # name should only be None if there's only one role ID,
        # we can use the last role variable from the loop
        return {name or role.name.title(): len(set(members))}
```

bot/exts/info/information.py
```diff
@@ -42,15 +43,33 @@ def get_channel_type_counts(guild: Guild) -> DefaultDict[str, int]:
         return channel_counter
 
     @staticmethod
-    def get_member_counts(guild: Guild) -> Dict[str, int]:
+    def join_role_stats(role_ids: list[int], name: str, guild: Guild) -> dict[str, int]:
+        """Return a dictionary with the number of `members` of each role given, and the `name` for this joined group."""
+        members = []
```
Bluenix2, line 48: It's probably more efficient to keep a count rather than allocate a list like this.

bot/exts/info/information.py
```diff
@@ -42,15 +43,33 @@ def get_channel_type_counts(guild: Guild) -> DefaultDict[str, int]:
...
+        role_ids = [constants.Roles.helpers, constants.Roles.mod_team, constants.Roles.admins,
+                    constants.Roles.owners, constants.Roles.contributors]
+        roles = []
+        for role_id in role_ids:
+            if (role := guild.get_role(role_id)) is not None:
+                roles.append(role)
+            else:
+                raise NonExistentRoleError(role_id)
+
+        role_stats = {role.name.title(): len(role.members) for role in roles}
```
Bluenix2, line 68: Remove the loop above as well
```suggestion
        role_stats = {}
        for role_id in role_ids:
            role_stats.update(self.join_role_stats([role_id], guild))
```

bot/exts/info/information.py
Bluenix2: Note: I think we can also come up with a better name

bot/exts/info/information.py
D0rs4n: `get_member_counts` is a staticmethod as well.

bot/errors.py
```diff
@@ -41,3 +41,17 @@ class BrandingMisconfiguration(RuntimeError):
...
 
     pass
+
+
+class NonExistentRoleError(ValueError):
+    """
+    Raised by the Information Cog when encountering a role that does not exist.
+
+    Attributes:
+        `role_id` -- the ID of the role that does not exist
```
Bluenix2, line 51: To be consistent with the errors above, capitalize "Role".

bot/exts/info/information.py
```diff
@@ -42,15 +43,29 @@ def get_channel_type_counts(guild: Guild) -> DefaultDict[str, int]:
         return channel_counter
 
     @staticmethod
-    def get_member_counts(guild: Guild) -> Dict[str, int]:
+    def join_role_stats(role_ids: list[int], guild: Guild, name: Optional[str] = None) -> dict[str, int]:
+        """Return a dictionary with the number of `members` of each role given, and the `name` for this joined group."""
+        members = int()
```
Bluenix2, line 48: I think we can just do `0` here haha

bot/exts/info/information.py
```diff
@@ -42,15 +43,29 @@ def get_channel_type_counts(guild: Guild) -> DefaultDict[str, int]:
...
-                constants.Roles.owners, constants.Roles.contributors,
-            )
+        role_ids = [constants.Roles.helpers, constants.Roles.mod_team, constants.Roles.admins,
+                    constants.Roles.owners, constants.Roles.contributors]
+
+        role_stats = {}
+        for role_id in role_ids:
+            role_stats.update(Information.join_role_stats([role_id], guild))
+        role_stats.update(
+            Information.join_role_stats([constants.Roles.project_leads, constants.Roles.domain_leads], guild, "Leads")
```
Bluenix2, line 66: Elsewhere we use `self` from what I found (instead of `Information`).

D0rs4n: But it's a static method, and the reason it is a static method is because it's a helper function, but it is sort of unique to this cog, so I put it directly into this cog.

Bluenix2: It's still a method, see the following examples
https://github.com/python-discord/bot/blob/4aaf2bbe3f0a866451cf9fc91e771dfd7c54d029/bot/converters.py#L91-L92
https://github.com/python-discord/bot/blob/4aaf2bbe3f0a866451cf9fc91e771dfd7c54d029/bot/converters.py#L107
...
https://github.com/python-discord/bot/blob/4aaf2bbe3f0a866451cf9fc91e771dfd7c54d029/bot/exts/moderation/silence.py#L367-L368
https://github.com/python-discord/bot/blob/4aaf2bbe3f0a866451cf9fc91e771dfd7c54d029/bot/exts/moderation/silence.py#L182

D0rs4n: In the examples they might have been called with `self` but the context was different. In case of `silence` for example it was called directly there (`async def silence(...)`) which in fact is not a staticmethod but a normal method. But in my case, I'm calling a staticmethod inside a staticmethod.(Which in fact is strange, but I could not put it into the helpers unfortunately, so I put it directly into the cog as a staticmethod.).
But if you prefer those staticmethod decorators to be removed I'll do so. (Although originally the `get_member_counts` was a staticmethod as well)

Bluenix2: Oh sorry, I see now. You do not have `self` as an argument

Bluenix2 (approved):

ChrisLovering (approved):
looks good to me, tested and works well!

PR #1679, opened 2021-07-11
narrow wording of ytdl to permit legal usages

graingert:


Akarys42:
Thank you for the PR. As previously discussed, we won't merge this. It will just be less clear and more confusing for the users and moderators. We are not lawyers, the mods cannot know what is legal and what isn't, plus we don't think we can trust a user saying they will only download their own video, or something similar.

PR #1680, opened 2021-07-15
Add command to resend infraction DM

### LINKED ISSUE #1664: Command to resend infraction embed
When an infraction is dealt, the user is DMed an embed with infraction information.

We'd like to have a command that takes an infraction ID, and sends that embed again. A possible syntax can be `!infraction resend <id>`

MarkKoz:
Resolves #1664 

The command added is `!infraction resend <infraction_id>`. It uses the same ID converter that the other commands in the `!infraction` group use. It has the following aliases: `!infraction send`, `!infraction dm`, and `!infraction rs` (short for "resend").

A confirmation message is shown when the DM is sent. Similarly, an error is shown if it is not sent (usually DMs disabled or the user couldn't be found). Furthermore, the command rejects requests to resend hidden/shadow infractions.

Discord_2021-07-15_19-20-36

The embed attempts to clarify that it is not a new infraction:

bild

Unlike the automatic DM, this command does not send the full reason for superstarify infractions. As demonsrated in the image above, it instead sends the reason stored in the database, which is `f"Old nickname: {old_nick}. {reason}'"`, where `reason` is the reason given by the actor. This isn't difficult to support, but I didn't feel like it was worth handling this edge case; there would have to be a hardcoded check for the infraction type.

There is also a minor bug fix for the superstarify reason included in this PR. See the commit message for details.

<details>
<summary>Gallery</summary>

Normal DM.
bild

Resent DM for an expired infraction. Notice that the duration never includes the time remaining if it's already expired. 
bild

Resent DM for a permanent infraction.
bild

Normal DM with a long expiry. Notice that the remaining time is omitted because it was identical to the duration.
bild

Resent DM for an inactive infraction which expires in the future.
bild

</details>


MarkKoz:
Note to self: append to the reason something that clearly identifies the new DM as an edit rather than a new infraction.

mbaruh:
We should make sure to not accept hidden infractions, as those weren't meant to be sent to the user in the first place.

MarkKoz:
Please comment on whether the advisory appended to the reason is okay.

MarkKoz:
@mbaruh How should it handle expired infractions?

jchristgit (approved):
This is rock solid code as always, Mark. Nice catch on the superstarify bug.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -162,20 +162,12 @@ async def apply_infraction(
...
-            dm_log_text = "\nDM: **Failed**"
-
-            # Sometimes user is a discord.Object; make it a proper user.
-            try:
-                if not isinstance(user, (discord.Member, discord.User)):
-                    user = await self.bot.fetch_user(user.id)
-            except discord.HTTPException as e:
-                log.error(f"Failed to DM {user.id}: could not fetch user (status {e.status})")
+            if await _utils.notify_infraction(infraction, user, user_reason):
+                dm_result = ":incoming_envelope: "
```
Bluenix2, line 166: Should this emoji perhaps be put in `constants.Emoji`?

bot/exts/moderation/infraction/_utils.py
```diff
@@ -188,6 +190,40 @@ async def notify_infraction(
...
+
+    `user` and `reason` can be used to override what is in `infraction`. Otherwise, this data will
+    be retrieved from `infraction`.
+
+    Also return False if the user needs to be fetched but fails to be fetched.
+    """
+    if user is None:
+        user = discord.Object(infraction["user"])
+
+    # Sometimes user is a discord.Object; make it a proper user.
```
Bluenix2, line 209: Is there a case, other than showcased above when `user` is not passed, when this is called with a `discord.Object` instance?

Because if there's no other case, the if-statement and this fetch can be combined.

MarkKoz: Yes, there is. Besides, if the fetch fails, the except block needs to be able to access the `id` attribute.

bot/exts/moderation/infraction/_scheduler.py
MarkKoz: I don't know. It's a standard Unicode emoji, which normally doesn't have a constant. The only one that does is `ok_hand`, but the constant is only used once out of the nine references to that emoji.

MarkKoz: I just noticed the emojis right above `ok_hand` in the config. Didn't recognise them at first since they were Unicode escapes. Still, it's not very consistent, and I imagine it being cumbersome to use a constant instead of e.g. just typing `":x:"`.

mbaruh:
@MarkKoz sorry for taking so long to review it.

The appended advisory looks good. The feature itself looks solid.

As for expired or inactive infractions, I think replacing the expiry value with `Expired` should be enough. This has the disadvantage of not telling the user when their infraction expired, but this isn't really a case we're planning to use the command in.

MarkKoz:
> As for expired or inactive infractions, I think replacing the expiry value with `Expired` should be enough. This has the disadvantage of not telling the user when their infraction expired, but this isn't really a case we're planning to use the command in.

Can we defer this to after #1721 is merged? I think it would make this slightly easier.

---

I realised that the advisory may get truncated if the reason is too long. I might have to add a separate parameter for an advisory than appending it to the reason. I didn't want to have to do that. I'll try to see if there are other solutions.

mbaruh:
Isn't it just adding a condition at https://github.com/python-discord/bot/blob/d914046dc5b661d144537715dbf80ae6b361b0e8/bot/exts/moderation/infraction/_utils.py#L219 checking two fields in the infraction object?

Maybe I'm missing something

MarkKoz:
> 
> 
> Isn't it just adding a condition at
> 
> https://github.com/python-discord/bot/blob/d914046dc5b661d144537715dbf80ae6b361b0e8/bot/exts/moderation/infraction/_utils.py#L219
> checking two fields in the infraction object?
> 
> Maybe I'm missing something

I'd like to do something nicer where the expiration date, duration, and whether it's expired are all included.

bot/exts/moderation/infraction/_utils.py
```diff
@@ -188,6 +190,40 @@ async def notify_infraction(
...
+    except discord.HTTPException as e:
+        log.error(f"Failed to DM {user.id}: could not fetch user (status {e.status})")
+        return False
+
+    type_ = infraction["type"].replace("_", " ").title()
+    icon = INFRACTION_ICONS[infraction["type"]][0]
+    expiry = time.format_infraction_with_duration(infraction["expires_at"])
+
+    if reason is None:
+        reason = infraction["reason"]
```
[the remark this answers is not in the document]
Bluenix2: Pretty sure that will never happen because of `reason` coming from discord.py and there an empty reason would become the default (None).

When is an empty string and None not functionally equivalent?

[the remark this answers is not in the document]
MarkKoz: It is possible by invoking it as follows: `!command ""`

bot/exts/moderation/infraction/management.py
```diff
@@ -38,13 +39,29 @@ def infractions_cog(self) -> Infractions:
         """Get currently loaded Infractions cog instance."""
         return self.bot.get_cog("Infractions")
 
-    # region: Edit infraction commands
-
```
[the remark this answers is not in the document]
MarkKoz: In fact, the search group is under the same command group, so this group shouldn't have been in the edit region to begin with.

MarkKoz:
> 
> 
> Per [this comment](https://github.com/python-discord/bot/pull/1680#issuecomment-894713190), this PR will be deferred until the time utilities refactor is merged.

It wasn't a suggestion to defer the whole PR but rather to defer that specific request. However, if the moderation team is okay to wait for this command, then I have no problem just deferring the whole PR.

mbaruh:
Yeah it will be nice to have, but it's not urgent.

onerandomusername (commented):
Can't mark it since its just out of pr bounds, line 174 of infraction/_utils.py should be 4096 now

additionally, line 161, is a default of token removed good for the icon_url? I feel like it should be `None`

Xithrius:
The time utility refactor has been merged.

> It wasn't a suggestion to defer the whole PR but rather to defer that specific request.

@MarkKoz Sorry about my misunderstanding. Since it was such a long time ago, I don't remember if I was planning to follow up or not, but that's no longer relevant.

Would you like to continue this PR where it left off?

MarkKoz:
@Xithrius Yes, I will finish it up when I find some time. Hopefully that will be before next week.

MarkKoz:
This is "done", but I left it as a draft since I need to test it after the big merge.

When merging, I decided that the `notify_infraction` utility function doesn't need to be separated from the embed sending code. This is because a good bit of code could be removed regarding getting the Member object. However, this means the test for that function no longer works. I tried to fix it, but ultimately realised that it's going to be a headache. The current time has to be patched (which can be more difficult than it sounds without something like `freezegun`) since the output expiration has the duration relative to the current time. I honestly can't be bothered to deal with this, so I am skipping the test in the suite.

Bluenix2:
Is this still a Draft by choice? Is it ready for review?

MarkKoz:
> Is this still a Draft by choice? Is it ready for review?

Still draft until I address what I wrote in my previous comment.

Bluenix2 (approved):
The code looks good, and the feature works. Thanks Mark for putting this together!

PR #1682, opened 2021-07-16
Allow manual mutes to override auto mutes

### LINKED ISSUE #1665: Make mutes override auto-mutes
Various behaviors might cause the bot to mute a user.

When a moderator deals a mute, it should disable a pre-existing bot mute.

MarkKoz:
Resolves #1665

If a moderator mutes a user, allow them to override any active mute that was automatically applied by the bot (e.g. from antispam).

When deactivating the infraction, it sends the mod log as normal. However, it does not send a confirmation message to the mod nor does it send a DM to the user:

bild

The fact that an automatic mute is being overriden is transparent to the mod unless an error occurs during deactivation:

bild

If it succeeds then there's just a single confirm message for success, just like one would expect for a normal mute

bild


vcokltfre (approved):
Looks good to me, tested a day or two back and seems to be working

bot/exts/moderation/infraction/infractions.py
```diff
@@ -444,43 +463,51 @@ async def pardon_ban(self, user_id: int, guild: discord.Guild, reason: t.Optiona
...
-    async def pardon_voice_ban(self, user_id: int, guild: discord.Guild, reason: t.Optional[str]) -> t.Dict[str, str]:
-        """Add Voice Verified role back to user, DM them a notification, and return a log dict."""
+    async def pardon_voice_ban(
+        self,
+        user_id: int,
+        guild: discord.Guild,
+        *,
+        notify: bool = True
+    ) -> t.Dict[str, str]:
+        """Optionally DM the user a pardon notification and return a log dict."""
```
Numerlor, line 473: Did you mean to overwrite the whole docstring here?

Numerlor (changes requested):
The superstarify `_pardon_action` will need to be changed to accept the notify argument, currently it'll error out when an user is attempted to be unstarified.

bot/exts/moderation/infraction/infractions.py
MarkKoz: Yes, because it was incorrect: the function doesn't actually add the role back, and this was confirmed to be intentional behaviour.

MarkKoz:
> The superstarify _pardon_action will need to be changed to accept the notify argument, currently it'll error out when an user is attempted to be unstarified.

Good catch. Pushed a fix.

Numerlor (approved):

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -468,10 +476,15 @@ async def deactivate_infraction(
         return log_text
 
     @abstractmethod
-    async def _pardon_action(self, infraction: _utils.Infraction) -> t.Optional[t.Dict[str, str]]:
+    async def _pardon_action(
+        self,
+        infraction: _utils.Infraction,
+        notify: bool
```
Bluenix2, line 482: Could it be worth it to make this a kwarg like everywhere else you made this change?

bot/exts/moderation/infraction/infractions.py
```diff
@@ -280,8 +280,19 @@ async def unvoiceban(self, ctx: Context, user: FetchedMember) -> None:
...
+        if active := await _utils.get_active_infraction(ctx, user, "mute", send_msg=False):
+            if active["actor"] != self.bot.user.id:
+                await _utils.send_active_infraction_message(ctx, active)
+                return
+
+            # Let the current mute attempt override an automatically triggered mute.
+            log_text = await self.deactivate_infraction(active, notify=False)
+            if "Failure" in log_text:
+                await ctx.send(
+                    f":x: can't override infraction **mute** for {user.mention}: "
```
Bluenix2, line 292: This should be capitalized.

tests/bot/exts/moderation/infraction/test_infractions.py
```diff
@@ -206,7 +206,7 @@ async def test_voice_unban_user_found(self, format_user_mock, notify_pardon_mock
         notify_pardon_mock.return_value = True
         format_user_mock.return_value = "my-user"
 
-        result = await self.cog.pardon_voice_ban(self.user.id, self.guild, "foobar")
+        result = await self.cog.pardon_voice_ban(self.user.id, self.guild)
         self.assertEqual(result, {
             "Member": "my-user",
             "DM": "Sent"
```
Bluenix2, line 212: Won't this fail because you don't specify `notify`?

tests/bot/exts/moderation/infraction/test_infractions.py
```diff
@@ -221,7 +221,7 @@ async def test_voice_unban_dm_fail(self, format_user_mock, notify_pardon_mock):
         notify_pardon_mock.return_value = False
         format_user_mock.return_value = "my-user"
 
-        result = await self.cog.pardon_voice_ban(self.user.id, self.guild, "foobar")
+        result = await self.cog.pardon_voice_ban(self.user.id, self.guild)
         self.assertEqual(result, {
             "Member": "my-user",
             "DM": "**Failed**"
```
Bluenix2, line 227: You don't specify `notify` here either.

bot/exts/moderation/infraction/superstarify.py
```diff
@@ -192,8 +192,8 @@ async def unsuperstarify(self, ctx: Context, member: Member) -> None:
         """Remove the superstarify infraction and allow the user to change their nickname."""
         await self.pardon_infraction(ctx, "superstar", member)
 
-    async def _pardon_action(self, infraction: _utils.Infraction) -> t.Optional[t.Dict[str, str]]:
-        """Pardon a superstar infraction and return a log dict."""
+    async def _pardon_action(self, infraction: _utils.Infraction, notify: bool) -> t.Optional[t.Dict[str, str]]:
```
Bluenix2, line 195: See other comment about making this a kwarg.

bot/exts/moderation/infraction/superstarify.py
```diff
@@ -208,18 +208,19 @@ async def _pardon_action(self, infraction: _utils.Infraction) -> t.Optional[t.Di
             )
             return {}
 
+        log_text = {"Member": format_user(user)}
+
         # DM the user about the expiration.
```
Bluenix2, line 213: You should move this inside the if statement like in the other file.

bot/exts/moderation/infraction/infractions.py
```diff
@@ -280,8 +280,19 @@ async def unvoiceban(self, ctx: Context, user: FetchedMember) -> None:
...
     async def apply_mute(self, ctx: Context, user: Member, reason: t.Optional[str], **kwargs) -> None:
         """Apply a mute infraction with kwargs passed to `post_infraction`."""
-        if await _utils.get_active_infraction(ctx, user, "mute"):
-            return
+        if active := await _utils.get_active_infraction(ctx, user, "mute", send_msg=False):
+            if active["actor"] != self.bot.user.id:
+                await _utils.send_active_infraction_message(ctx, active)
+                return
+
+            # Let the current mute attempt override an automatically triggered mute.
```
Bluenix2, line 288: This wording is a bit ambiguous, is "attempt" in this context a noun or verb?

As in, "Let the current attempt to mute the user override an automatically triggered mute." or is it "Let the current mute attempt to override an automatically triggered mute."?

I prefer some variation of the former, I think it could be improved.

tests/bot/exts/moderation/infraction/test_infractions.py
MarkKoz: No. This function has a default value for that parameter. Furthermore, tests run in CI and you can see CI is passing.

bot/exts/moderation/infraction/infractions.py
MarkKoz: Lowercase is consistent with the other messages sent out by infraction stuff.

bot/exts/moderation/infraction/_scheduler.py
MarkKoz: It could be done, but not really worth it IMO because in practice this won't get called directly with a True/False value — instead a variable is forwarded to it. This is the same reason it doesn't have a default value.

bot/exts/moderation/infraction/superstarify.py
MarkKoz: Unlike the other function, this one has an early exit with `if not user: ... return`. Your suggestion would call for a refactor of the function, resulting in a more significant diff. I didn't want to change more than necessary.

bot/exts/moderation/infraction/infractions.py
MarkKoz: I just forgot to write "to" after "attempt".

MarkKoz: Well, I didn't forget. I was using it as a noun, but I agree it's ambiguous.

bot/exts/moderation/infraction/infractions.py
Bluenix2: Are you sure? Here's from that same file..
https://github.com/python-discord/bot/blob/693573412dcf726426271c66f832e626fa77a5d4/bot/exts/moderation/infraction/infractions.py#L355
https://github.com/python-discord/bot/blob/693573412dcf726426271c66f832e626fa77a5d4/bot/exts/moderation/infraction/infractions.py#L155
https://github.com/python-discord/bot/blob/693573412dcf726426271c66f832e626fa77a5d4/bot/exts/moderation/infraction/infractions.py#L73

bot/exts/moderation/infraction/superstarify.py
Bluenix2: To clarify, I mean the comment. You should move the comment down a line to be consistent.

bot/exts/moderation/infraction/infractions.py
MarkKoz: Turns out there's no consistency. Whether this message is lower or upper-cased, it will be inconsistent with some other message. The confirmation messages use lower case, so that is what I was originally thinking of. I don't think making the messages consistent should be addressed by this PR.

bot/exts/moderation/infraction/superstarify.py
MarkKoz: Oh. That's such a minor thing, I don't think it's worthwhile. Maybe if I had another reason to be modifying that area of code now, I'd throw in this change.

Bluenix2: Refering to our other conversation, this could be the same reasoning as to why capitalization is so inconsistent.

MarkKoz: That's just because no one noticed when reviewing the PRs which introduced those messages. Furthermore, there's an important difference: those messages are user-facing, but comments such as this one are not. Outside of readability/ambiguity reasons, I think worrying about the placement of comments is being too pedantic; someone doesn't need to go out of their way to address such issue.

Bluenix2: I won't continue arguing about this, but code is read more times than it is written. It most likely will never be changed when this is merged, so now is the time to change it.

MarkKoz: You're right, but in this case, I think that's okay; moving this comment would be so inconsequential. If we're concerned with consistency, the function should undergo a more significant refactor as I originally mentioned. In fact, I will later make a separate PR to make the message capitalisation consistent, and I can also refactor this function in that PR.

Bluenix2 (approved):

mbaruh:
When you added the notify argument to mute pardons, why you did so for voice bans as well?

MarkKoz:
> 
> 
> When you added the notify argument to mute pardons, why you did so for voice bans as well?

Because they also send a pardon DM. It'd be inconsistent to allow it for one infraction type but not another when both send a DM.

mbaruh:
> 
> 
> > When you added the notify argument to mute pardons, why you did so for voice bans as well?
> 
> Because they also send a pardon DM. It'd be inconsistent to allow it for one infraction type but not another when both send a DM.

What about the pardon_ban function then?

mbaruh (approved):
Looks good, thank you

PR #1685, opened 2021-07-20
Enabled charinfo for discord_py channel.

### LINKED ISSUE #1684: Enable charinfo command in discord_py channel.
In order to add a reaction in discordpy, we need the emoji's unicode. For most emotes/emojis we can do \:emoji: , but for some of them (such as numbers and letters) this option is not possible. The best replacement for it is the charinfo command, which currently isn't allowed in discord.py, but I feel should be for the above reason.

NIRDERIi:
Closes #1684 
Add support to charinfo command in discord.py channel


vcokltfre (approved):
:shipit:

Akarys42 (approved):
10/10, thanks a lot!

PR #1687, opened 2021-07-21
Reminder: remove footer

Akarys42:
Now that we have Discord timestamps, the timestamp in the footer isn't useful anymore since it can be hovered to have a localised timestamp.

image


ChrisLovering (approved):

wookie184 (commented):
The `delivery_dt` argument isn't used anymore, so I guess it could be removed

wookie184 (approved):
:shipit:

PR #1693, opened 2021-07-23
Update docstring tag pep and grammar

ChrisLovering:


bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,19 @@
...
-  return_string = f"Hello, {name} you are {age} years old!"
-  return return_string
+    :param name: The name to greet.
+    :type name: str
+    :param age: The age to display.
+    :type age: int
+
+    :return: String representation of the greeting.
+    """
+    return_string = f"Hello, {name} you are {age} years old!"
```
Bluenix2, line 14: Unnecessary variable?

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,19 @@
...
+    return return_string
 ```
-You can get the docstring by using `.__doc__` attribute. For the last example you can get it through: `print(greet.__doc__)`.
+You can get the docstring by using the `.__doc__` attribute. For the last example you can get it through: `print(greet.__doc__)`.
```
Bluenix2, line 17:
```suggestion
You can get the docstring by using the `.__doc__` attribute. For the last example you can print it like this: `print(greet.__doc__)`.
```

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,19 @@
-A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that often used in file, classes, functions, etc. A docstring should have a clear explanation of exactly what the function does. You can also include descriptions of the function's parameter(s) and its return type, as shown below.
+A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that's often used in files, classes, functions, etc. A docstring should contain a clear explanation of exactly what the function does. You can also include descriptions of the function's parameter(s) and its return type, as shown below.
 ```py
 def greet(name, age) -> str:
-  """
-  Return a string that greets the given person, including their name and age.
+    """
```
Bluenix2, line 4: Doesn't PEP 257 say docstrings should be written like this:
```py
"""A magic method, we don't know what it does.

Only kept here for legacy reasons, if we remove it the code breaks.
"""
```

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,19 @@
-A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that often used in file, classes, functions, etc. A docstring should have a clear explanation of exactly what the function does. You can also include descriptions of the function's parameter(s) and its return type, as shown below.
+A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that's often used in files, classes, functions, etc. A docstring should contain a clear explanation of exactly what the function does. You can also include descriptions of the function's parameter(s) and its return type, as shown below.
```
Bluenix2: Why have the `etc.`? That's all the uses..
```suggestion
A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that's often at the top of files, classes, and functions. A docstring should contain a clear explanation of exactly what the function does. You can also include descriptions of the function's parameter(s) and its return type, as shown below.
```

bot/resources/tags/docstring.md
ChrisLovering: I don't think it mandates that, it uses that in some examples, but then also uses the blank first line as an example lower down, saying they're equivalent. At least that's how I read the pep

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,16 @@
-A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that often used in file, classes, functions, etc. A docstring should have a clear explanation of exactly what the function does. You can also include descriptions of the function's parameter(s) and its return type, as shown below.
+A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that's placed at the top of files, classes and functions. A docstring should contain a clear explanation of what it's describing. You can also include descriptions of the subjects's parameter(s) and its return type, as shown below.
```
Bluenix2: > A docstring should contain a clear explanation of what it's describing.

Huh?

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,16 @@
...
-  :param name: The name to greet.
-  :type name: str
-  :param age: The age to display.
-  :type age: int
-  :return: String of the greeting.
-  """
-  return_string = f"Hello, {name} you are {age} years old!"
-  return return_string
+    :param name: The name to greet.
+    :param age: The age to display.
```
Bluenix2, line 8:
```suggestion
    :param age: The person's age.
```

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,16 @@
...
+
+    :return: String representation of the greeting.
+    """
+    return f"Hello {name}, you are {age} years old!"
 ```
-You can get the docstring by using `.__doc__` attribute. For the last example you can get it through: `print(greet.__doc__)`.
+You can get the docstring by using the `.__doc__` attribute. For the last example, you can print it by doing this: `print(greet.__doc__)`.
 
-For more details about what docstring is and it's usage check out this guide by [Real Python](https://realpython.com/documenting-python-code/#docstrings-background), or the [PEP-257 docs](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring).
+For more details about what a docstring is and its usage, check out this guide by [Real Python](https://realpython.com/documenting-python-code/#docstrings-background), or the [PEP-257 docs](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring).
```
Bluenix2, line 16: I kind of want to include some form of note that explains that a PEP is the official specification/proposal for those that don't know.

Maybe just?
```suggestion
For more details about what a docstring is and its usage, check out this guide by [Real Python](https://realpython.com/documenting-python-code/#docstrings-background), or the [official docstring specification](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring).
```

bot/resources/tags/docstring.md
ChrisLovering: I was actually about to commit `The age of the person to include in the greeting.` What do you think?

Bluenix2: Something more like `The age of the person to greet`? Since we want to know the age of the person that we're about to greet, the person isn't necessarily *included*.

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,16 @@
-A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that often used in file, classes, functions, etc. A docstring should have a clear explanation of exactly what the function does. You can also include descriptions of the function's parameter(s) and its return type, as shown below.
+A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that's placed at the top of files, classes and functions. A docstring should contain a clear explanation of what it's describing. You can also include descriptions of the subject's parameter(s) and what it returns, as shown below:
```
Bluenix2, line 1:
```suggestion
A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string (always using triple quotes) that's placed at the top of files, classes and functions. A docstring should contain a clear explanation of what it's describing. You can also include descriptions of the subject's parameter(s) and what it returns, as shown below:
```

bot/resources/tags/docstring.md
ChrisLovering: We didn't want to enumerate files, classes and functions again, so we went with this instead, not sure how to improve this.

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,16 @@
...
-  :param age: The age to display.
-  :type age: int
-  :return: String of the greeting.
-  """
-  return_string = f"Hello, {name} you are {age} years old!"
-  return return_string
+    :param name: The name of the person to greet.
+    :param age: The age of the person to greet.
+
+    :return: String representation of the greeting.
```
Bluenix2, line 10: We should be consistent here, either change all of them or change this to..
```suggestion
    :return: The string representation of the greeting.
```

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,16 @@
-A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string with triple quotes that often used in file, classes, functions, etc. A docstring should have a clear explanation of exactly what the function does. You can also include descriptions of the function's parameter(s) and its return type, as shown below.
+A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string, always using triple quotes, that's placed at the top of files, classes and functions. A docstring should contain a clear explanation of what it's describing. You can also include descriptions of the subject's parameter(s) and what it returns, as shown below:
```
Bluenix2, line 1: We could probably uses dashes for the triple quote comment?
```suggestion
A [`docstring`](https://docs.python.org/3/glossary.html#term-docstring) is a string - always using triple quotes - that's placed at the top of files, classes and functions. A docstring should contain a clear explanation of what it's describing. You can also include descriptions of the subject's parameter(s) and what it returns, as shown below:
```

bot/resources/tags/docstring.md
Bluenix2: What about..
> A docstring should contain a clear explanation of what it does and its purpose.

ChrisLovering: it would need to be referring to the subject here, rather than the docstring itself. So what about `A docstring should contain a clear explanation of what it's describing does and its purpose.`

Bluenix2: `what it's describing does` still rubs me the wrong way.. what if we change that to `whatever`?

Or I guess completely different more like `A docstring should be a clear explanation of the purpose of what it describes.`

Numerlor: I think something along the lines of `A docstring should describe the behaviour and the interface of what it's attached to` would fit.

bot/resources/tags/docstring.md
```diff
@@ -1,18 +1,16 @@
...
-  return return_string
+    :param name: The name of the person to greet.
+    :param age: The age of the person to greet.
+
+    :return: The greeting.
+    """
+    return f"Hello {name}, you are {age} years old!"
 ```
-You can get the docstring by using `.__doc__` attribute. For the last example you can get it through: `print(greet.__doc__)`.
+You can get the docstring by using the `.__doc__` attribute. For the last example, you can print it by doing this: `print(greet.__doc__)`.
```
Numerlor, line 14: Would `inspect.getdoc` make more sense here? I'm not sure if the attribute is intended to be accessed directly, or something like getdoc which also cleans it up is a better alternative

Bluenix2: It's kind of like `dir()` vs. `.__dict__`.

bot/resources/tags/docstring.md
Bluenix2: I like `what it's attached to`, but a docstring should not explain the interface. That is up to the function signature to do.

Numerlor: With the interface I meant describing what some of the more complex parameters mean, not just how the signature looks- - maybe the original wording is clearer there

bot/resources/tags/docstring.md
ChrisLovering: yea, good idea. Added in https://github.com/python-discord/bot/pull/1693/commits/39500b40ac2892cb469366103022bd854f7734b2

ChrisLovering: > It's kind of like `dir()` vs. `.__dict__`.

inspect.getdoc Also clears indents, which is nice

TizzySaurus (approved):

Bluenix2 (approved):
I don't have time to check it completely again, but I saw the test in Discord. It seems good!

Issue #1624, opened 2021-06-06
Do not delete !docs/!pypi invocation if it contains a ping


Akarys42:
# Current behavior
If a `!docs` invocation cannot find the appropriate symbol, the invocation and the bot answer is deleted after a few second, even if the message was a ping reply or contained a ping, which lead to a ghost ping.

# Suggested behavior
The bot answer is or is not deleted (maybe we can just have a :x: reaction instead?) but the message stays to avoid getting a ghost ping (only if it contained a ping).

TizzySaurus:
The `!pypi` command has the same issue so will work on that too and link to this issue. Will also edit title to reflect this.

PR #1694, opened 2021-07-23
Prevent ghost-pings in docs get command

TizzySaurus:
Solves #1624.

Will no longer delete the invoking message when the given symbol is invalid if the message contains user/role mentions. If it has mentions, allows deletion of error message through reactions.

### DIFF
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -14,7 +14,7 @@ import discord
 from discord.ext import commands
 
 from bot.bot import Bot
-from bot.constants import MODERATION_ROLES, RedirectOutput
+from bot.constants import Emojis, MODERATION_ROLES, RedirectOutput
 from bot.converters import Inventory, PackageName, ValidURL, allowed_strings
 from bot.pagination import LinePaginator
 from bot.utils.lock import SharedEvent, lock
@@ -341,6 +341,9 @@ class DocCog(commands.Cog):
             if doc_embed is None:
                 error_message = await send_denial(ctx, "No documentation found for the requested symbol.")
                 await wait_for_deletion(error_message, (ctx.author.id,), timeout=NOT_FOUND_DELETE_DELAY)
+
+                with suppress(discord.NotFound):
+                    await message.clear_reaction(Emojis.trashcan)
 
                 if not (ctx.message.mentions or ctx.message.role_mentions):
                     with suppress(discord.NotFound):
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -12,7 +12,6 @@ from typing import Dict, NamedTuple, Optional, Tuple, Union
 
 import aiohttp
 import discord
-
 from discord.ext import commands
 
 from bot.bot import Bot
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -343,7 +343,7 @@ class DocCog(commands.Cog):
                 await wait_for_deletion(error_message, (ctx.author.id,), timeout=NOT_FOUND_DELETE_DELAY)
 
                 with suppress(discord.NotFound):
-                    await message.clear_reaction(Emojis.trashcan)
+                    await error_message.clear_reaction(Emojis.trashcan)
 
                 if not (ctx.message.mentions or ctx.message.role_mentions):
                     with suppress(discord.NotFound):
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -13,7 +13,6 @@ from typing import Dict, NamedTuple, Optional, Tuple, Union
 import aiohttp
 import discord
 from discord import Colour, Embed, Message, NotFound, Reaction, User
-
 from discord.ext.commands import BadArgument, Cog, Context, group, has_any_role
 
 from bot.bot import Bot
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -344,6 +344,7 @@ class DocCog(commands.Cog):
                 with suppress(discord.NotFound):
                     await error_message.clear_reaction(Emojis.trashcan)
 
+                # Make sure that we won't cause a ghost-ping by deleting the message
                 if not (ctx.message.mentions or ctx.message.role_mentions):
                     with suppress(discord.NotFound):
                         await ctx.message.delete()
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -6,6 +6,7 @@ import sys
 import textwrap
 from collections import defaultdict
 from contextlib import suppress
+from functools import partial
 from types import SimpleNamespace
 from typing import Dict, NamedTuple, Optional, Tuple, Union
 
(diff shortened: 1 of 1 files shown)


bot/exts/info/doc/_cog.py
```diff
@@ -445,3 +468,7 @@ def cog_unload(self) -> None:
         self.inventory_scheduler.cancel_all()
         self.init_refresh_task.cancel()
         asyncio.create_task(self.item_fetcher.clear(), name="DocCog.item_fetcher unload clear")
+        
+        
+def predicate_emoji_reaction(ctx: Context, error_message: Message, reaction: Reaction, user: User):
```
Bluenix2, line 473: I'd honestly rather see this at the top before the cog rather than afterwards.

bot/exts/info/doc/_cog.py
```diff
@@ -6,12 +6,15 @@
...
 from collections import defaultdict
 from contextlib import suppress
+from functools import partial
 from types import SimpleNamespace
 from typing import Dict, NamedTuple, Optional, Tuple, Union
 
 import aiohttp
 import discord
-from discord.ext import commands
+from discord import Colour, Embed, Message, NotFound, Reaction, User
```
Bluenix2, line 15: This messes with the `git blame` and could be considered unnecessary.

TizzySaurus: Did it to shorten the `predicate_emoji_reaction` function signature.

bot/exts/info/doc/_cog.py
TizzySaurus: Was just copying what was done in the [eval code](https://github.com/python-discord/bot/blob/main/bot/exts/utils/snekbox.py#L340-L342). Will move it up if you still think I should.

Numerlor:
the pypi command and possibly others in the future also delete invalid invocations, we should extract this to an util function under `utils.messages` instead of only implementing the functionality for this cog.
Also, like @​Bluenix2 mentioned, please do not make unnecessary styling changes like the imports

bot/exts/info/doc/_cog.py
```diff
@@ -34,6 +35,7 @@
     "pdbcommand",
     "2to3fixer",
 )
+DELETE_ERROR_MESSAGE_REACTION = '\u274c'  # :x:
```
Bluenix2, line 38: You can actually refer to Unicode characters by their name, that might be a better option.

bot/exts/info/doc/_cog.py
Bluenix2: In that case keep it like that for consistency.

bot/exts/info/doc/_cog.py
Bluenix2: See the first option..
```py
>>> "\N{GREEK CAPITAL LETTER DELTA}"  # Using the character name
'\u0394'
>>> "\u0394"                          # Using a 16-bit hex value
'\u0394'
>>> "\U00000394"                      # Using a 32-bit hex value
'\u0394'
```
From Python 3 docs - [Unicode HOWTO](https://docs.python.org/3/howto/unicode.html).

bot/exts/info/doc/_cog.py
```diff
@@ -340,11 +340,13 @@ async def get_command(self, ctx: commands.Context, *, symbol_name: Optional[str]
 
             if doc_embed is None:
                 error_message = await send_denial(ctx, "No documentation found for the requested symbol.")
+
```
Bluenix2, line 343: For the sake of the diffs, can you remove this empty linee you added?

TizzySaurus: > For the sake of the diffs, can you remove this empty line you added?

Yep.

bot/exts/info/doc/_cog.py
```diff
@@ -340,11 +340,13 @@ async def get_command(self, ctx: commands.Context, *, symbol_name: Optional[str]
...
                 await wait_for_deletion(error_message, (ctx.author.id,), timeout=NOT_FOUND_DELETE_DELAY)
-                with suppress(discord.NotFound):
-                    await ctx.message.delete()
-                with suppress(discord.NotFound):
-                    await error_message.delete()
+
+                if not (ctx.message.mentions or ctx.message.role_mentions):
+                    with suppress(discord.NotFound):
+                        await ctx.message.delete()
+
```
Bluenix2, line 349: You still need to remove `error_message`.
```suggestion
                with suppress(discord.NotFound):
                    await error_message.delete()
```

TizzySaurus: > You still need to remove `error_message`.

That will be deleted in [L#343?](https://github.com/python-discord/bot/blob/c5ab6af7770745d69e0d1d04fc07d7b8c601f98a/bot/exts/info/doc/_cog.py#L343)

bot/exts/info/doc/_cog.py
```diff
@@ -341,10 +341,14 @@ async def get_command(self, ctx: commands.Context, *, symbol_name: Optional[str]
             if doc_embed is None:
                 error_message = await send_denial(ctx, "No documentation found for the requested symbol.")
                 await wait_for_deletion(error_message, (ctx.author.id,), timeout=NOT_FOUND_DELETE_DELAY)
+
```
Bluenix2, line 344: Might as well remove this line, I don't think it helps much in terms of readabily. Good to not introduce unneccessary changes.
```suggestion
```

TizzySaurus: Done :+1:

bot/exts/info/doc/_cog.py
```diff
@@ -342,9 +342,12 @@ async def get_command(self, ctx: commands.Context, *, symbol_name: Optional[str]
                 error_message = await send_denial(ctx, "No documentation found for the requested symbol.")
                 await wait_for_deletion(error_message, (ctx.author.id,), timeout=NOT_FOUND_DELETE_DELAY)
                 with suppress(discord.NotFound):
-                    await ctx.message.delete()
-                with suppress(discord.NotFound):
-                    await error_message.delete()
+                    await error_message.clear_reaction(Emojis.trashcan)
+
+                if not (ctx.message.mentions or ctx.message.role_mentions):
```
Bluenix2, line 347: Should we perhaps add a comment for this? I feel like it may not be immediately obvious.

TizzySaurus: Are you referring to the if statement?

TizzySaurus: If so, something like `if ...  # if message doesn't ping anyone`?

Bluenix2: You could add a comment above the if statement (but keeping a new line). For example:
```suggestion
                # Make sure that we won't cause a ghost-ping by deleting the message
                if not (ctx.message.mentions or ctx.message.role_mentions):
```

TizzySaurus: Yeah, that works better. Will do now :+1:

Bluenix2 (approved):
I haven't tested it, but I see nothing wrong. Looks good!

wookie184 (approved):
Looks good to me

NIRDERIi (approved):
The code seems to be solid and should work just fine.

ChrisLovering (approved):

Issue #1695, opened 2021-07-23
Change `utils.messages.wait_for_deletion` to delete reactions after timeout


TizzySaurus:
Discussed this in `#dev-log` with @Bluenix2. [Full conversation](https://discord.com/channels/267624335836053506/635950537262759947/868210825796677693).

Summary:

This is a nice quality of life feature as currently when the timeout expires (e.g. after 5 minutes for the `!eval` command), the reaction stays on the message but the bot will no longer react if the emoji is pressed. This makes it seem like the bot isn't working since there's no response. By removing the reactions after the timeout, it makes it clear that it's no longer possible to delete the message.

It also has a small bonus benefit of making life easier for [another issue](https://github.com/python-discord/bot/issues/1624) I'm working on (saves duplicate code of having to remove reactions manually after calling `wait_for_deletion()`).

PR #1696, opened 2021-07-23
Prevent ghost-pings in pypi command

TizzySaurus:
Solves #1624
Solves #1695

Will no longer delete the invoking message of `!pypi` command when the given package is invalid if the message contains user/role mentions. Always allows deletion of error message through reactions.

`utils.messages.wait_for_deletion()` now also [clears the added reaction(s) once timeout expires.](https://github.com/python-discord/bot/commit/df3d011bd6490eadc2b611c8d21062b14ebd18c4)

### DIFF
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -14,7 +14,7 @@ import discord
 from discord.ext import commands
 
 from bot.bot import Bot
-from bot.constants import Emojis, MODERATION_ROLES, RedirectOutput
+from bot.constants import MODERATION_ROLES, RedirectOutput
 from bot.converters import Inventory, PackageName, ValidURL, allowed_strings
 from bot.pagination import LinePaginator
 from bot.utils.lock import SharedEvent, lock
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -346,6 +346,7 @@ class DocCog(commands.Cog):
                 if not (ctx.message.mentions or ctx.message.role_mentions):
                     with suppress(discord.NotFound):
                         await ctx.message.delete()
+                        await error_message.delete()
 
             else:
                 msg = await ctx.send(embed=doc_embed)
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -341,8 +341,6 @@ class DocCog(commands.Cog):
             if doc_embed is None:
                 error_message = await send_denial(ctx, "No documentation found for the requested symbol.")
                 await wait_for_deletion(error_message, (ctx.author.id,), timeout=NOT_FOUND_DELETE_DELAY)
-                with suppress(discord.NotFound):
-                    await error_message.clear_reaction(Emojis.trashcan)
 
                 # Make sure that we won't cause a ghost-ping by deleting the message
                 if not (ctx.message.mentions or ctx.message.role_mentions):
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -10,12 +10,11 @@ from types import SimpleNamespace
 from typing import Dict, NamedTuple, Optional, Tuple, Union
 
 import aiohttp
-import asyncio
 import discord
 from discord.ext import commands
 
 from bot.bot import Bot
-from bot.constants import MODERATION_ROLES, RedirectOutput
+from bot.constants import Emojis, MODERATION_ROLES, RedirectOutput
 from bot.converters import Inventory, PackageName, ValidURL, allowed_strings
 from bot.pagination import LinePaginator
 from bot.utils.lock import SharedEvent, lock
@@ -35,7 +34,6 @@ FORCE_PREFIX_GROUPS = (
     "pdbcommand",
     "2to3fixer",
 )
-DELETE_ERROR_MESSAGE_REACTION = '\u274c'  # :x:
 NOT_FOUND_DELETE_DELAY = RedirectOutput.delete_delay
 # Delay to wait before trying to reach a rescheduled inventory again, in minutes
 FETCH_RESCHEDULE_DELAY = SimpleNamespace(first=2, repeated=5)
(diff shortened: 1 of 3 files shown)


bot/exts/info/pypi.py
```diff
@@ -67,8 +67,14 @@ async def get_package_info(self, ctx: Context, package: str) -> None:
                     log.trace(f"Error when fetching PyPi package: {response.status}.")
 
         if error:
-            await ctx.send(embed=embed, delete_after=INVALID_INPUT_DELETE_DELAY)
-            await ctx.message.delete(delay=INVALID_INPUT_DELETE_DELAY)
+            error_message = await ctx.send(embed=embed)
+            await wait_for_deletion(error_message, (ctx.author.id,), timeout=INVALID_INPUT_DELETE_DELAY)
+
+            # If won't ghost-ping when deleting message
```
Bluenix2, line 73: Please look over this again haha

TizzySaurus: Was referring to the fact I didn't need to import `bot.constants.Emojis` anymore. Import was removed

NIRDERIi (approved):
The code changed should work just fine...

bot/exts/info/pypi.py
```diff
@@ -1,14 +1,16 @@
+import contextlib
```
Bluenix2, line 1: Could be good to be consistent, everywhere else it's `from contextlib import suppress`.

bot/exts/info/pypi.py
```diff
@@ -67,8 +69,14 @@ async def get_package_info(self, ctx: Context, package: str) -> None:
                     log.trace(f"Error when fetching PyPi package: {response.status}.")
 
         if error:
-            await ctx.send(embed=embed, delete_after=INVALID_INPUT_DELETE_DELAY)
-            await ctx.message.delete(delay=INVALID_INPUT_DELETE_DELAY)
+            error_message = await ctx.send(embed=embed)
+            await wait_for_deletion(error_message, (ctx.author.id,), timeout=INVALID_INPUT_DELETE_DELAY)
+
+            # If won't ghost-ping when deleting message
```
Bluenix2, line 75: This comment is still very weirdly worded.

bot/exts/info/pypi.py
```diff
@@ -67,8 +69,14 @@ async def get_package_info(self, ctx: Context, package: str) -> None:
                     log.trace(f"Error when fetching PyPi package: {response.status}.")
 
         if error:
-            await ctx.send(embed=embed, delete_after=INVALID_INPUT_DELETE_DELAY)
-            await ctx.message.delete(delay=INVALID_INPUT_DELETE_DELAY)
+            error_message = await ctx.send(embed=embed)
+            await wait_for_deletion(error_message, (ctx.author.id,), timeout=INVALID_INPUT_DELETE_DELAY)
```
Bluenix2, line 73: I want to know how other feels about this:
```suggestion
            await wait_for_deletion(await ctx.send(embed=embed), (ctx.author.id,), timeout=INVALID_INPUT_DELETE_DELAY)
```
It could degrade readability and make less sense.

bot/utils/messages.py
```diff
@@ -95,8 +94,11 @@ async def wait_for_deletion(
         allow_mods=allow_mods,
     )
 
-    with contextlib.suppress(asyncio.TimeoutError):
+    try:
         await bot.instance.wait_for('reaction_add', check=check, timeout=timeout)
+    except asyncio.TimeoutError:
+        await message.clear_reactions()
```
Bluenix2, line 100: Perhaps we could adjust the docstring for this change?

bot/exts/info/pypi.py
TizzySaurus: Forgot to reword it when pasting over. Will do that

bot/exts/info/pypi.py
TizzySaurus: Wasn't sure if it would change the import orders (I don't really understand how the linting works for imports)

bot/exts/info/pypi.py
TizzySaurus: I considered that but opted not to in the end. If others think it's a good idea though I can commit this.

bot/utils/messages.py
TizzySaurus: Good idea. [Committed](https://github.com/python-discord/bot/pull/1696/commits/aa77ce65dff915611ce5bc56d77f41b5803a5e65)

bot/exts/info/pypi.py
TizzySaurus: [Committed](https://github.com/python-discord/bot/pull/1696/commits/d52bf9d921465320ba8f70fb36933aed8340480d) with no issues

bot/utils/messages.py
TizzySaurus: And [re-committed.](https://github.com/python-discord/bot/pull/1696/commits/1abd1c230aedea6c0e8e680253bd7690cc27fd4c)

Bluenix2 (approved):
Same as last time, can't test it but it looks good! Nice work

TizzySaurus:
> Same as last time, can't test it but it looks good! Nice work

Thanks! Appreciate your help today :+1:

vcokltfre (approved):
LGTM!

bot/exts/info/doc/_cog.py
```diff
@@ -342,29 +340,15 @@ async def get_command(self, ctx: commands.Context, *, symbol_name: Optional[str]
...
-                        await error_message.clear_reaction(DELETE_ERROR_MESSAGE_REACTION)
-
-                else:
-                    await wait_for_deletion(error_message, (ctx.author.id,), timeout=NOT_FOUND_DELETE_DELAY)
+                # Make sure that we won't cause a ghost-ping by deleting the message
+                if not (ctx.message.mentions or ctx.message.role_mentions):
                     with suppress(discord.NotFound):
                         await ctx.message.delete()
-                    with suppress(discord.NotFound):
-                        await error_message.delete()
```
kosayoda: Is this change intentional? Prior, both the invoking message and the error embed would be deleted, but now the error embed remains. I noticed that this is a duplicate commit of the other ghost ping PR, but the situation is still similar to the pypi command.

kosayoda (changes requested):

jchristgit (approved):
Straightforward and reasonable. Very nice. 👍

bot/exts/info/doc/_cog.py
TizzySaurus: Yeah, I changed this deliberately assuming we'd want consistent behaviour with the docs command. If you think this should be reverted to the old behaviour then I'll do so, but my thoughts was that this should be updated to be the same as docs.

TizzySaurus: For now I'll leave this conversation unsolved and wait for your response @kosayoda.

kosayoda: I am asking this question for both this and the docs PR. `wait_for_deletion` does not delete the embed after the timeout. Before the docs PR, the error embed would be deleted after the timeout (code being the removed lines), which is no longer the case now.

TizzySaurus: Right, I see. I personally prefer keeping the embed and just removing the reactions, but if we want to revert to old system we can do.

kosayoda: For these embeds specifically, the idea is to not have multiple error embeds lying around, since they have lose meaning right after being shown. This is in line with the short configured delay on the reaction check.

Bluenix2: I think we should add the embed delete inside the same `suppress()`, that way the embed only gets deleted if the bot is able to delete the invoking message.

See this diff:
```diff
                # Make sure that we won't cause a ghost-ping by deleting the message
                if not (ctx.message.mentions or ctx.message.role_mentions):
                    with suppress(discord.NotFound):
                        await ctx.message.delete()
+                       await error_message.delete()
```

If the invoking message pings someone, and they did not react with a trashcan, then both messages will stay up.
If the invoking messaes does not ping someone both messages will be deleted regardless of reacting with a trashcan in time.

TizzySaurus: That sounds good to me. I'll do this if @kosayoda thinks that's good too.

TizzySaurus: Kosayoda has approved on discord so will commit this :+1:

kosayoda (approved):
Works well!

PR #1700, opened 2021-07-27
Restrict internal eval command to bot owner when in debug mode

### LINKED ISSUE #1683: Restrict internal eval command to bot owner when in debug mode
It's a security risk to let anyone in a test server have essentially full access to the host machine. Thoughts on this? Does it need to be configurable or should it always be on when in debug mode?

MarkKoz:
Resolves #1683 

As mentioned in the issue, it's a security risk to let anyone in a test server have essentially full access to the host machine.

A new `debug` entry is introduced in the config file. This is false by default, but can be set to true in the user config. When set to true, an `is_owner` check is added to the internal eval command.

There isn't quite an easy way to disable this during runtime, or to disable this without disabling debug mode entirely. However, I don't feel this is necessary. It seems rare to want to allow someone else. To disable during runtime, one would have to remove the check from the `bot.get_cog("Internal").eval.checks` list.


Numerlor:
I don't like the debug mode being false as the config default as it should almost always be set for dev; what do you think about making it an env var with defaulting to enabled?

MarkKoz:
> 
> 
> I don't like the debug mode being false as the config default as it should almost always be set for dev; what do you think about making it an env var with defaulting to enabled?

That's a decent idea. The downside is it goes against the concept of the default config being the default values for production, but I don't see a simple way around this.

ChrisLovering:
> > I don't like the debug mode being false as the config default as it should almost always be set for dev; what do you think about making it an env var with defaulting to enabled?
> 
> That's a decent idea. The downside is it goes against the concept of the default config being the default values for production, but I don't see a simple way around this.

We could have it as an env var that gets set in the compose file, since that isn't use in prod

MarkKoz:
> 
> 
> > > I don't like the debug mode being false as the config default as it should almost always be set for dev; what do you think about making it an env var with defaulting to enabled?
> > 
> > 
> > That's a decent idea. The downside is it goes against the concept of the default config being the default values for production, but I don't see a simple way around this.
> 
> We could have it as an env var that gets set in the compose file, since that isn't use in prod

It's a partial fix because compose is not necessary to run the bot.

ChrisLovering:
> > > > I don't like the debug mode being false as the config default as it should almost always be set for dev; what do you think about making it an env var with defaulting to enabled?
> > > 
> > > 
> > > That's a decent idea. The downside is it goes against the concept of the default config being the default values for production, but I don't see a simple way around this.
> > 
> > 
> > We could have it as an env var that gets set in the compose file, since that isn't use in prod
> 
> It's a partial fix because compose is not necessary to run the bot.

That's true, if we want to default on, we could ask joe to add a debug = False env var to the cluster and then load it with a True default.

i can't really think of a good catch-all fix.

HassanAbouelela:
Do we really need to find a fix? We can just tell people to set the value like we do for most other values. The biggest downside I see is that it’ll leave a possible security vulnerability if the user doesn’t set it, and nothing will tell them so. That being said, I imagine most people are already running these bots in their own test servers, or with people they trust.

MarkKoz:
> The biggest downside I see is that it’ll leave a possible security vulnerability if the user doesn’t set it, and nothing will tell them so

Yeah, that's the concern, and I think it's a completely valid point. I'm thinking that production could just "take the hit" and set the env var to False explicitly. It has to set other env vars anyway.

Xithrius (approved):
Good stuff. Tested, works well.

mbaruh:
> 
> 
> > The biggest downside I see is that it’ll leave a possible security vulnerability if the user doesn’t set it, and nothing will tell them so
> 
> Yeah, that's the concern, and I think it's a completely valid point. I'm thinking that production could just "take the hit" and set the env var to False explicitly. It has to set other env vars anyway.

What is the verdict then? do we set an env var in prod, and default to true?

MarkKoz:
> 
> 
> > > The biggest downside I see is that it’ll leave a possible security vulnerability if the user doesn’t set it, and nothing will tell them so
> > 
> > 
> > Yeah, that's the concern, and I think it's a completely valid point. I'm thinking that production could just "take the hit" and set the env var to False explicitly. It has to set other env vars anyway.
> 
> What is the verdict then? do we set an env var in prod, and default to true? I don't mind that approach.

I am okay with it too. The minor problem is that a boolean cannot be specified as an environment variable. I guess it has to be true/false strings, and it'll have to be specified this way in YAML too, which isn't as nice. Unless someone has a better idea...

Numerlor:
> > > > The biggest downside I see is that it’ll leave a possible security vulnerability if the user doesn’t set it, and nothing will tell them so
> > > 
> > > 
> > > Yeah, that's the concern, and I think it's a completely valid point. I'm thinking that production could just "take the hit" and set the env var to False explicitly. It has to set other env vars anyway.
> > 
> > 
> > What is the verdict then? do we set an env var in prod, and default to true? I don't mind that approach.
> 
> I am okay with it too. The minor problem is that a boolean cannot be specified as an environment variable. I guess it has to be true/false strings, and it'll have to be specified this way in YAML too, which isn't as nice. Unless someone has a better idea...

I think the usual approach of the env var having a value assigned being the "true" value should be enough

wookie184 (approved):

Akarys42 (approved):

PR #1702, opened 2021-07-29
Add virtual environment (venv) tag

### LINKED ISSUE #1701: Tag: `!venv` explaining virtual environments
### Description

Add a `!venv` tag to explain the purpose, creation and activation of virtual environments.

It should outline the following:
- Purpose: keeps system clean, dependency management, etc.
- Creation: `python -m venv .venv`
- Activation for both PowerShell and Bash

### Notes

On Windows systems, the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) needs to be set before activating from PowerShell:
```powershell
Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
```

### Resources

- [Python docs](https://docs.python.org/3/library/venv.html)

Zedeldi:
Closes #1701

### DIFF
--- a/bot/resources/tags/venv.md
+++ b/bot/resources/tags/venv.md
@@ -12,7 +12,7 @@ or (Command Prompt): `.venv\Scripts\activate.bat`
 
 Packages can then be installed to the virtual environment using `pip`, as normal.
 
-For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html). Your IDE, for example [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) or [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html), likely has `venv` integration.
+For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html). If you run code through your editor, check its documentation on how to make it use your virtual environment. For example, see the [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) or [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html) docs.
 
 Tools such as [poetry](https://python-poetry.org/docs/basic-usage/) and [pipenv](https://pipenv.pypa.io/en/latest/) can manage the creation of virtual environments as well as project dependencies, making packaging and installing your project easier.
 
--- a/bot/resources/tags/venv.md
+++ b/bot/resources/tags/venv.md
@@ -10,7 +10,7 @@ Then, to activate the new virtual environment:
 or (Command Prompt): `.venv\Scripts\activate.bat`
 **MacOS / Linux** (Bash): `source .venv/bin/activate`
 
-Note: On Windows, you may need to change the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) first. This is only required once:
+Note: When using Windows PowerShell, you may need to change the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) first. This is only required once:
 `Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser`
 
 Packages can then be installed to the virtual environment using `pip`, as normal.
--- a/bot/resources/tags/venv.md
+++ b/bot/resources/tags/venv.md
@@ -2,11 +2,12 @@
 
 Virtual environments are isolated Python environments, which make it easier to keep your system clean and manage dependencies. By default, when activated, only libraries and scripts installed in the virtual environment are accessible.
 
-To create a new virtual environment, you can use the standard library `venv` module: `python3 -m venv .venv` (replace `python3` with `python` or `py` on windows)
+To create a new virtual environment, you can use the standard library `venv` module: `python3 -m venv .venv` (replace `python3` with `python` or `py` on Windows)
 
 Then, to activate the new virtual environment:
 
 PowerShell: `.venv\Scripts\Activate.ps1`
+Command Prompt: `.venv\Scripts\activate.bat`
 Bash: `source .venv/bin/activate`
 
 Note: On Windows, you may need to change the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) first:
@@ -15,6 +16,6 @@ Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
 ```
 Packages can then be installed to the virtual environment using `pip`, as normal.
 
-For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html). Your editor may also have venv integration, in which case you should check out your editors documentation. Here's the docs for [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) and [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html)
+For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html). Your IDE, such as [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) or [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html), may also have `venv` integration.
 
-There are also tools such as [poetry](https://python-poetry.org/docs/basic-usage/) and [pipenv](https://pipenv.pypa.io/en/latest/) you can use for managing creation of venvs as well as project dependency management, among other things
+Tools such as [poetry](https://python-poetry.org/docs/basic-usage/) and [pipenv](https://pipenv.pypa.io/en/latest/) can manage the creation of virtual environments as well as project dependencies, making packaging and installing your project easier.
--- a/bot/resources/tags/venv.md
+++ b/bot/resources/tags/venv.md
@@ -1,0 +1,18 @@
+**Virtual Environments**
+
+Virtual environments are isolated Python environments, which make it easier to keep your system clean and manage dependencies. By default, when activated, only libraries and scripts installed in the virtual environment are accessible.
+
+To create a new virtual environment, you can use the `venv` module: `python -m venv .venv`
+
+Then, to activate the new virtual environment:
+
+PowerShell: `.venv\Scripts\Activate.ps1`
+Bash: `source .venv/bin/activate`
+
+Note: On Windows, you may need to change the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) first:
+```powershell
+Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
+```
+Packages can then be installed to the virtual environment using `pip`, as normal.
+
+For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html).
--- a/bot/resources/tags/venv.md
+++ b/bot/resources/tags/venv.md
@@ -16,6 +16,6 @@ Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
 ```
 Packages can then be installed to the virtual environment using `pip`, as normal.
 
-For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html). Your IDE, such as [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) or [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html), may also have `venv` integration.
+For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html). Your IDE, for example [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) or [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html), may also have `venv` integration.
 
 Tools such as [poetry](https://python-poetry.org/docs/basic-usage/) and [pipenv](https://pipenv.pypa.io/en/latest/) can manage the creation of virtual environments as well as project dependencies, making packaging and installing your project easier.
--- a/bot/resources/tags/venv.md
+++ b/bot/resources/tags/venv.md
@@ -1,6 +1,6 @@
 **Virtual Environments**
 
-Virtual environments are isolated Python environments, which make it easier to keep your system clean and manage dependencies. By default, when activated, only libraries and scripts installed in the virtual environment are accessible, allowing dependency conflicts to be resolved, and your program requirements to be found easily.
+Virtual environments are isolated Python environments, which make it easier to keep your system clean and manage dependencies. By default, when activated, only libraries and scripts installed in the virtual environment are accessible, preventing cross-project dependency conflicts, and allowing easy isolation of requirements.
 
 To create a new virtual environment, you can use the standard library `venv` module: `python3 -m venv .venv` (replace `python3` with `python` or `py` on Windows)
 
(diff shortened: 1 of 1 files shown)


Xithrius:
I'm not sure if this is enough information. Maybe it would be a good idea to have this on the site, where we also go over poetry and/or pipenv?

wookie184:
I'd agree with Xithrius that I'm not sure this tag explains enough at the moment, and may be better off as a page on the site.

For example, there may be editor specific advice people should follow. If someone using Pycharm tried to follow the instructions in this tag they might end up more confused than before.

It could be worth having a tag that explains what a venv is and then just links to other pages for how to use them, maybe including one on our site if we write one.

NIRDERIi:
I agree with @wookie184 , but, why not make a tag that does show how to use venvs and create ones, with a comment on top suggesting for some editors that they have less confusing ways. But, if it's a lot to explain, and there are some examples to show which are important, I would suggest making some page on the pydis website(or the appropriate place) that covers almost everything needed to know about venvs. I wish I could be more of a help, but I have no idea what venv is, how do I use\create one, or what is it for, which is what makes me more excited about this. Hope I'll get to learn it from the amazing coders here!

Zedeldi:
Completely agree that it's worth mentioning other IDEs that integrate venvs already, else it could prove more confusing. I'm happy to move this to site content, though I think there should be a summary tag to briefly describe their purpose, linking to the site page, which explains how to set them up -- both for PyCharm, VS Code, etc. and manually.

> where we also go over poetry and/or pipenv

I imagine packaging should probably be a separate section, or at least subsection, though definitely worth including. It would require users to know about structuring their project first, and have a basic understanding of modules.

bot/resources/tags/venv.md
```diff
@@ -0,0 +1,18 @@
+**Virtual Environments**
+
+Virtual environments are isolated Python environments, which make it easier to keep your system clean and manage dependencies. By default, when activated, only libraries and scripts installed in the virtual environment are accessible.
+
+To create a new virtual environment, you can use the `venv` module: `python -m venv .venv`
```
wookie184, line 5:
```suggestion
To create a new virtual environment, you can use the standard library `venv` module: `python3 -m venv .venv` (replace `python3` with `python` or `py` on windows)
```
May need `python3` on unix, so probably worth changing to something like this. Also mention that's it's in the standard library

bot/resources/tags/venv.md
```diff
@@ -0,0 +1,18 @@
...
+
+Then, to activate the new virtual environment:
+
+PowerShell: `.venv\Scripts\Activate.ps1`
+Bash: `source .venv/bin/activate`
```
wookie184, line 10: I think it would also be worth adding the command for cmd,  `.venv\Scripts\activate.bat`

bot/resources/tags/venv.md
```diff
@@ -0,0 +1,18 @@
...
+
+Note: On Windows, you may need to change the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) first:
```
wookie184, line 12: This sort of feels a little out of scope for the tag. Would be nice if we could just link to this section in the venv docs next to the command above, although i'm not sure you can specifically link to that part. You could try linking to another page, like https://dev.to/aka_anoop/enabling-virtualenv-in-windows-powershell-ka3, may be nicer to have this info on our site at some point though,

bot/resources/tags/venv.md
```diff
@@ -0,0 +1,18 @@
...
+```powershell
+Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
+```
+Packages can then be installed to the virtual environment using `pip`, as normal.
+
+For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html).
```
wookie184, line 18:
```suggestion
For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html). Your editor may also have venv integration, in which case you should check out your editors documentation. Here's the docs for [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) and [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html)

There are also tools such as [poetry](https://python-poetry.org/docs/basic-usage/) and [pipenv](https://pipenv.pypa.io/en/latest/) you can use for managing creation of venvs as well as project dependency management, among other things
```

wookie184 (changes requested):
I think this tag probably could be useful. We could maybe update it more if/when we have a page on our site, although I don't think it's worth waiting for that at the moment.

Feel free to question my comments/rewrite them with your own wording.

bot/resources/tags/venv.md
Zedeldi: The issue with those pages is that they suggest changing the execution policy to `Unrestricted`, whereas `RemoteSigned` would be preferred -- as the Python docs suggest.

Until we get a page for the site, keeping something about execution policies here will probably be useful.

wookie184: Another possible option could be to link to a gist, something like https://gist.github.com/wookie184/4bec5f2785ad72736a6a46bb4cc7defc#file-executionpolicy-md, although if you'd like to leave the note in the tag for now I think that's fine too.

wookie184:
This is how the tag currently looks:
Discord_Nwc5mCCJny
would be interested to hear thoughts from others on what they think so far.

Zedeldi:
I've reworded a few things, just to make it a little more concise. Feel free to make any more suggestions :)

Information to activate from cmd has been added too.

bot/resources/tags/venv.md
```diff
@@ -0,0 +1,21 @@
+**Virtual Environments**
+
+Virtual environments are isolated Python environments, which make it easier to keep your system clean and manage dependencies. By default, when activated, only libraries and scripts installed in the virtual environment are accessible.
```
bast0006, line 3: I think we can fit a sentence in here explaining why clean dependencies are important. We have an opportunity to make this more beginner friendly without adding too much, and I think we should take it. Eg:
 - knowing what your program actually requires
 - dependency conflicts
 - partial upgrades rather than having to update every script on your system at once

bot/resources/tags/venv.md
```diff
@@ -0,0 +1,21 @@
...
+
+To create a new virtual environment, you can use the standard library `venv` module: `python3 -m venv .venv` (replace `python3` with `python` or `py` on Windows)
+
+Then, to activate the new virtual environment:
+
+PowerShell: `.venv\Scripts\Activate.ps1`
+Command Prompt: `.venv\Scripts\activate.bat`
+Bash: `source .venv/bin/activate`
```
bast0006, line 11: We should note that bash applies to linux/osx. It took me a moment ( despite extensive familiarity with all three here and the os's and python ) to realize that yeah, the line marked "bash" was the one that I was looking for. If we put (OSX/Linux) in there, or perhaps something better if you can figure it out, that might help make things clearer and more obvious.

bot/resources/tags/venv.md
bast0006: I don't think length is a problem, but I do think the prominence is a problem. Someone reading the tag might think that this is immediately important/applies all the time, given the font-size of codeblocks and the fact it comes before the very end of the tag, and this isn't strictly true, you only need to run the command once. It takes a lot of mindshare and I think that can be improved. 

Try moving it to the very, very end of the tag, as well as sticking it within \`regular code tags\` rather than a \`\`\`code block\`\`\`, and see if that's better?

bot/resources/tags/venv.md
```diff
@@ -0,0 +1,21 @@
...
+
+Note: On Windows, you may need to change the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) first:
+```powershell
+Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
+```
+Packages can then be installed to the virtual environment using `pip`, as normal.
+
+For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html). Your IDE, for example [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) or [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html), may also have `venv` integration.
```
bast0006, line 19: Would changing `may` to `is likely to` or `likely has` make this better?

bast0006 (changes requested):
I like this tag. It definitely fits the use case where a user doesn't know what a virtual environment is, or why they should need one. 

However, at the moment the wording does seem a little more technical than I think it needs to be. It assumes a bunch of baseline knowledge (why should the dependencies be clean, that I may need to set an execution policy, that I know what bash is, etc), and I think we can clean that up. Otherwise though, I like it. Thanks for your PR!

bot/resources/tags/venv.md
Zedeldi: Okey doke, totally agree. I'll try to throw some of those in there.

bot/resources/tags/venv.md
Zedeldi: My concern with specifying an OS would be that bash, for example, could be for Linux, OSX or WSL. Likewise, PowerShell can be installed on platforms other than Windows. The only specific difference between the three is the shell itself.

I guess this is a little pedantic, and can probably be ignored. Thoughts?

bast0006: WSL is linux, technically. While you can get bash underneath windows proper with MingW, the reason to mention the OS is to catch the people who are using it (starting anew on mac, or just getting their feet wet on linux), who think of "terminal" rather than "bash". 

It is a significant portion of the people likely to view the tag, rather than those who've installed powershell on something other than windows, who are unlikely to get confused on which applies since they've chosen to install the tech by name manually. 

If we did have a linux distro of some popularity utilizing powershell as the default we'd want to note that here. There isn't one, so thankfully we won't be tempted to.

Zedeldi: > If we did have a linux distro of some popularity utilizing powershell as the default

Sent shudders down my spine haha

Yeah, with the target audience, it's probably best to choose clarity over accuracy. Would something like below be okay?
> Shell (OS): `path/to/activate`

bast0006: Well, I'm not sure what the best option is, but again, we're painting targets. Like how bow targets are concentric rings and the comment button here is green and on the bottom right, we want the people searching the tag for what activate command they need to have their eyes immediately go to the right spot. 

I do think duplicating Windows (powershell) might be too much.. but maybe not. I'll defer to other's judgement on that.

Zedeldi: Writing "Windows" twice I think would be redundant. I've used bold for the OS to draw attention to the determining factor. This is the best way I can think to distinguish between PowerShell and Command Prompt:

**Windows** (PowerShell): `.venv\Scripts\Activate.ps1`
or (Command Prompt): `.venv\Scripts\activate.bat`
**MacOS / Linux** (Bash): `source .venv/bin/activate`

bot/resources/tags/venv.md
Zedeldi: My reasoning for its placement is that it follows directly after the activation examples, which is the stage where changing the execution policy is required. I'll make some changes to suggest that it's only required once.

I agree that it is rather prominent, however. Changing to single backticks seems to work, but it should probably be on a new line to avoid wrapping.

bot/resources/tags/venv.md
dawnofmidnight: I think that sounds pretty good and clear.

bot/resources/tags/venv.md
```diff
@@ -1,21 +1,20 @@
 **Virtual Environments**
 
-Virtual environments are isolated Python environments, which make it easier to keep your system clean and manage dependencies. By default, when activated, only libraries and scripts installed in the virtual environment are accessible.
+Virtual environments are isolated Python environments, which make it easier to keep your system clean and manage dependencies. By default, when activated, only libraries and scripts installed in the virtual environment are accessible, allowing dependency conflicts to be resolved, and your program requirements to be found easily.
```
bast0006, line 3: I think the wording here could be a little better. Perhaps `preventing cross-project dependency conflicts and easy isolation of requirements`?

bot/resources/tags/venv.md
```diff
@@ -1,21 +1,20 @@
...
 
-PowerShell: `.venv\Scripts\Activate.ps1`
-Command Prompt: `.venv\Scripts\activate.bat`
-Bash: `source .venv/bin/activate`
+**Windows** (PowerShell): `.venv\Scripts\Activate.ps1`
+or (Command Prompt): `.venv\Scripts\activate.bat`
+**MacOS / Linux** (Bash): `source .venv/bin/activate`
+
+Note: On Windows, you may need to change the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) first. This is only required once:
+`Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser`
```
bast0006, line 14: I really do think this should be at the very bottom. As it is it somewhat interrupts the flow of the text like an overly large () in the middle of a paragraph.

bast0006 (changes requested):

bot/resources/tags/venv.md
Zedeldi: That's a lot cleaner - done :+1:

bot/resources/tags/venv.md
```diff
@@ -0,0 +1,20 @@
...
+
+To create a new virtual environment, you can use the standard library `venv` module: `python3 -m venv .venv` (replace `python3` with `python` or `py` on Windows)
+
+Then, to activate the new virtual environment:
+
+**Windows** (PowerShell): `.venv\Scripts\Activate.ps1`
+or (Command Prompt): `.venv\Scripts\activate.bat`
+**MacOS / Linux** (Bash): `source .venv/bin/activate`
+
+Note: On Windows, you may need to change the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) first. This is only required once:
```
wookie184, line 13:
```suggestion
Note: When using Windows PowerShell, you may need to change the [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies) first. This is only required once:
```
This doesn't apply to cmd, so we should specify it's PowerShell only. I also like bast's suggestion of moving this to the bottom

bot/resources/tags/venv.md
```diff
@@ -0,0 +1,20 @@
...
+
+Then, to activate the new virtual environment:
+
+**Windows** (PowerShell): `.venv\Scripts\Activate.ps1`
+or (Command Prompt): `.venv\Scripts\activate.bat`
+**MacOS / Linux** (Bash): `source .venv/bin/activate`
+
+Packages can then be installed to the virtual environment using `pip`, as normal.
+
+For more information, take a read of the [documentation](https://docs.python.org/3/library/venv.html). Your IDE, for example [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) or [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html), likely has `venv` integration.
```
wookie184, line 15: I think the bit about other IDEs should be reworded. I think it should be more clear that the links relate to the corresponding editors' virtual environment documentation rather than just information about the editors itself. I also think it should be specifically recommending looking at your editors venv documentation.

For example, it is common that a user may set up a venv using a guide similar to this, and then be confused when they click the green run button in the IDE and it runs with the system python rather than in the venv.

Something like:
```
If you run your code through your editor you should check it's docs so you can set it up to use your venv, for example, see the [VSCode](link) or [Pycharm](link) docs.
```
Thoughts?

wookie184 (commented):
This is starting to look really good now, nice work.

bot/resources/tags/venv.md
Zedeldi: Yep, I struggled to keep that section short, while keeping it relevant. I think that sums it up nicely.

Just a slight reword:

> If you run code through your editor, check the relevant documentation on how to use your new virtual environment. For example, see the [VSCode](https://code.visualstudio.com/docs/python/environments#_select-and-activate-an-environment) or [PyCharm](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html) docs.

Is that alright?

wookie184: Yeah, that sounds great.

This is maybe slightly more concise again

```
If you run code through your editor, check it's documentation on how to make it use your virtual environment. For example, see the VSCode or PyCharm docs.
```

Anything along these lines or what you suggested is good though

wookie184 (approved):
Looks good to me, thanks!

bast0006 (approved):

dawnofmidnight (approved):

PR #1704, opened 2021-07-31
Add partners to the filtering whitelist

ChrisLovering:
A partner being filtered is highly unlikely to be correct


Akarys42 (approved):
Lgtm, if it breaks, blame the owl

mbaruh (approved):

Issue #1641, opened 2021-06-13
Add flake8-quotes to linting workflow


Numerlor:
Double quotes are currently preferred by the style guide, but aren't enforced in any way in the project. This leaves the project in a bit inconsistent state ands adds more work for reviewers and contributors during the PR process.

The flake8-quotes plugin can be used to raise linting errors if the preferred quote type is not used. From testing, everything seems to work fine except for multi line strings with implicit concat https://github.com/zheller/flake8-quotes/issues/82, but this is easy to ignore and I haven't found any uses in a quick look through this project.

The main deterrent to this change would be the change to the existing codebase, at the time of writing the `main` branch raised
1044 errors, on 773 lines in 71 unique files. 14 of these files and 511 errors are in files under the tests folder.
While it would be a big change, it can be done as a mechanical process and merged quickly after the change if everything is discussed, so the long term disruption should be minimal.

Akarys42:
I'm not sure if we really want to change them all at once, mainly for all the conflicts it will arise. We can just change them as we go on, and perhaps look at this once we only have a few double quotes remaining? 

I am personally not a fan of the rule that enforces double quotes, I use each type of quote depending on the context. Also, this does not automatically change the quotes, which is pretty silly if you ask me, I'd rather have it as a precommit hook.

Poke @Xithrius, @jb3, @HassanAbouelela

Numerlor:
> I'm not sure if we really want to change them all at once, mainly for all the conflicts it will arise. 

From a quick replace I found conflicts when merging #1602, #1568, #1539, #1487 and #836; it affects less PRs than I thought it would, but still a considerable number. Although conflicts coming from here should be mostly trivial to resolve.

> We can just change them as we go on, and perhaps look at this once we only have a few double quotes remaining?

I really doubt this will ever happen, the existing code doesn't get change as much and quotes are easy to miss during reviews.
> I am personally not a fan of the rule that enforces double quotes, I use each type of quote depending on the context.

The plugin will properly choose quotes to avoid escaping etc. in other cases the quotes would be against the style guide so I don't think those should be accommodated.

> Also, this does not automatically change the quotes, [...], I'd rather have it as a precommit hook.

I didn't consider that, but it sounds better as quotes can differ between projects so having it taken care of automatically would be nice. 
[...]

PR #1705, opened 2021-08-01
Add commands to turn automatic review posting on or off

wookie184:
We may want to temporarily stop new nomination reviews from being automatically posted, so this PR creates three new commands, `tp autoreview on`, `tp autoreview off`, and `tp autoreview status`, all with fairly self-explanatory names, I think :)

This uses RedisCache to store the status so it can be persisted over restarts.

### DIFF
--- a/bot/exts/recruitment/talentpool/_cog.py
+++ b/bot/exts/recruitment/talentpool/_cog.py
@@ -18,7 +18,7 @@ from bot.exts.recruitment.talentpool._review import Reviewer
 from bot.pagination import LinePaginator
 from bot.utils import time
 
-AUTOREVIEW_ENABLED_KEY = 'autoreview_enabled'
+AUTOREVIEW_ENABLED_KEY = "autoreview_enabled"
 REASON_MAX_CHARS = 1000
 
 log = logging.getLogger(__name__)
@@ -48,7 +48,7 @@ class TalentPool(WatchChannel, Cog, name="Talentpool"):
         if await self.autoreview_enabled():
             await self.reviewer.reschedule_reviews()
         else:
-            self.log.trace('Not scheduling reviews as autoreview is disabled.')
+            self.log.trace("Not scheduling reviews as autoreview is disabled.")
 
     async def autoreview_enabled(self) -> bool:
         """Return whether automatic posting of nomination reviews is enabled."""
@@ -60,41 +60,41 @@ class TalentPool(WatchChannel, Cog, name="Talentpool"):
         """Highlights the activity of helper nominees by relaying their messages to the talent pool channel."""
         await ctx.send_help(ctx.command)
 
-    @nomination_group.group(name='autoreview', aliases=('ar',), invoke_without_command=True)
+    @nomination_group.group(name="autoreview", aliases=("ar",), invoke_without_command=True)
     @has_any_role(*MODERATION_ROLES)
     async def nomination_autoreview_group(self, ctx: Context) -> None:
         """Commands for enabling or disabling autoreview."""
         await ctx.send_help(ctx.command)
 
-    @nomination_autoreview_group.command(name='enable', aliases=('on',))
+    @nomination_autoreview_group.command(name="enable", aliases=("on",))
     @has_any_role(Roles.admins)
     async def autoreview_enable(self, ctx: Context) -> None:
         """
         Enable automatic posting of reviews.
 
         This will post reviews up to one day overdue. Older nominations can be
-        manually reviewed with `!tp post_review <user_id>`.
+        manually reviewed with the `tp post_review <user_id>` command.
         """
         await self.talentpool_settings.set(AUTOREVIEW_ENABLED_KEY, True)
         await self.reviewer.reschedule_reviews()
-        await ctx.send(':white_check_mark: Autoreview enabled')
+        await ctx.send(":white_check_mark: Autoreview enabled")
 
-    @nomination_autoreview_group.command(name='disable', aliases=('off',))
+    @nomination_autoreview_group.command(name="disable", aliases=("off",))
     @has_any_role(Roles.admins)
     async def autoreview_disable(self, ctx: Context) -> None:
         """Disable automatic posting of reviews."""
         await self.talentpool_settings.set(AUTOREVIEW_ENABLED_KEY, False)
         self.reviewer.cancel_all()
-        await ctx.send(':white_check_mark: Autoreview disabled')
+        await ctx.send(":white_check_mark: Autoreview disabled")
 
-    @nomination_autoreview_group.command(name='status')
+    @nomination_autoreview_group.command(name="status")
     @has_any_role(*MODERATION_ROLES)
     async def autoreview_status(self, ctx: Context) -> None:
         """Show whether automatic posting of reviews is enabled or disabled."""
         if await self.autoreview_enabled():
-            await ctx.send('Autoreview is currently enabled')
+            await ctx.send("Autoreview is currently enabled")
         else:
-            await ctx.send('Autoreview is currently disabled')
+            await ctx.send("Autoreview is currently disabled")
 
     @nomination_group.command(name='watched', aliases=('all', 'list'), root_aliases=("nominees",))
     @has_any_role(*MODERATION_ROLES)
--- a/bot/exts/recruitment/talentpool/_cog.py
+++ b/bot/exts/recruitment/talentpool/_cog.py
@@ -27,6 +27,8 @@ log = logging.getLogger(__name__)
 class TalentPool(WatchChannel, Cog, name="Talentpool"):
     """Relays messages of helper candidates to a watch channel to observe them."""
 
+    # RedisCache[str, bool]
+    # Can contain a single key, "autoreview_enabled", with the value a bool indicating if autoreview is enabled.
     talentpool_settings = RedisCache()
 
     def __init__(self, bot: Bot) -> None:
--- a/bot/exts/recruitment/talentpool/_cog.py
+++ b/bot/exts/recruitment/talentpool/_cog.py
@@ -75,6 +75,10 @@ class TalentPool(WatchChannel, Cog, name="Talentpool"):
         This will post reviews up to one day overdue. Older nominations can be
         manually reviewed with the `tp post_review <user_id>` command.
         """
+        if await self.autoreview_enabled():
+            await ctx.send(":x: Autoreview is already enabled")
+            return
+
         await self.talentpool_settings.set(AUTOREVIEW_ENABLED_KEY, True)
         await self.reviewer.reschedule_reviews()
         await ctx.send(":white_check_mark: Autoreview enabled")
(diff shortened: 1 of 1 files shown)


bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -37,14 +40,61 @@ def __init__(self, bot: Bot) -> None:
...
+    @nomination_autoreview_group.command(name='on')
+    @has_any_role(Roles.admins)
+    async def autoreview_on(self, ctx: Context) -> None:
+        """
+        Turn on automatic posting of reviews.
+
+        This will post reviews up to one day overdue, older nominations can be
+        manually reviewed with `!tp post_review <user_id>`.
+        """
+        await self.talentpool_settings.set('autoreview_enabled', True)
```
MarkKoz, line 77: Use a constant for the key. String literals are prone to programmer errors.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -37,14 +40,61 @@ def __init__(self, bot: Bot) -> None:
...
+        await ctx.send_help(ctx.command)
+
+    @nomination_autoreview_group.command(name='on')
+    @has_any_role(Roles.admins)
+    async def autoreview_on(self, ctx: Context) -> None:
+        """
+        Turn on automatic posting of reviews.
+
+        This will post reviews up to one day overdue, older nominations can be
+        manually reviewed with `!tp post_review <user_id>`.
```
MarkKoz, line 75: This is a comma splice. Use a semicolon or make them separate sentences with a full stop.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -37,14 +40,61 @@ def __init__(self, bot: Bot) -> None:
...
+    @nomination_autoreview_group.command(name='off')
+    @has_any_role(Roles.admins)
+    async def autoreview_off(self, ctx: Context) -> None:
+        """Turn off automatic posting of reviews."""
+        await self.talentpool_settings.set('autoreview_enabled', False)
+        self.reviewer.cancel_all()
+        await ctx.send(':white_check_mark: Autoreview turned off')
+
+    @has_any_role(*MODERATION_ROLES)
+    @nomination_autoreview_group.command(name='status')
```
MarkKoz, line 90: Don't check decorators have to go under the command decorator?

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -37,14 +40,61 @@ def __init__(self, bot: Bot) -> None:
...
+        """Return whether automatic posting of nomination reviews is enabled."""
+        return await self.talentpool_settings.get('autoreview_enabled', True)
 
     @group(name='talentpool', aliases=('tp', 'talent', 'nomination', 'n'), invoke_without_command=True)
     @has_any_role(*MODERATION_ROLES)
     async def nomination_group(self, ctx: Context) -> None:
         """Highlights the activity of helper nominees by relaying their messages to the talent pool channel."""
         await ctx.send_help(ctx.command)
 
+    @nomination_group.group(name='autoreview', invoke_without_command=True)
```
MarkKoz, line 62: `ar` might be a handy alias.

MarkKoz (changes requested):

bot/exts/recruitment/talentpool/_cog.py
wookie184: From a quick check it didn't seem to matter, although makes sense to change it for consistency anyway.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -37,14 +41,61 @@ def __init__(self, bot: Bot) -> None:
...
         """Highlights the activity of helper nominees by relaying their messages to the talent pool channel."""
         await ctx.send_help(ctx.command)
 
+    @nomination_group.group(name='autoreview', aliases=('ar',), invoke_without_command=True)
+    @has_any_role(*MODERATION_ROLES)
+    async def nomination_autoreview_group(self, ctx: Context) -> None:
+        """Commands for enabling or disabling autoreview."""
+        await ctx.send_help(ctx.command)
+
+    @nomination_autoreview_group.command(name='on')
```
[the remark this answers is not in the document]
wookie184: Fair, i'll change the names to `enable` and `disable` and make on and off aliases.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -37,14 +41,61 @@ def __init__(self, bot: Bot) -> None:
...
+    @has_any_role(Roles.admins)
+    async def autoreview_off(self, ctx: Context) -> None:
+        """Turn off automatic posting of reviews."""
+        await self.talentpool_settings.set(AUTOREVIEW_ENABLED_KEY, False)
+        self.reviewer.cancel_all()
+        await ctx.send(':white_check_mark: Autoreview turned off')
+
+    @nomination_autoreview_group.command(name='status')
+    @has_any_role(*MODERATION_ROLES)
+    async def autoreview_status(self, ctx: Context) -> None:
```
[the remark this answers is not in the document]
wookie184: It could be, although I think both ways are equally fine. The current method is similar to how `!defcon` does it, and as it's a command that wont be used very often I think having it show the help by default may be helpful.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -37,14 +41,61 @@ def __init__(self, bot: Bot) -> None:
...
+        await ctx.send_help(ctx.command)
+
+    @nomination_autoreview_group.command(name='enable', aliases=('on',))
+    @has_any_role(Roles.admins)
+    async def autoreview_enable(self, ctx: Context) -> None:
+        """
+        Enable automatic posting of reviews.
+
+        This will post reviews up to one day overdue. Older nominations can be
+        manually reviewed with `!tp post_review <user_id>`.
```
Bluenix2, line 76: Including prefix like this is rather brittle. Can you not refer to it ambiguously?

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -37,14 +41,61 @@ def __init__(self, bot: Bot) -> None:
...
+        await ctx.send(':white_check_mark: Autoreview disabled')
+
+    @nomination_autoreview_group.command(name='status')
+    @has_any_role(*MODERATION_ROLES)
+    async def autoreview_status(self, ctx: Context) -> None:
+        """Show whether automatic posting of reviews is enabled or disabled."""
+        if await self.autoreview_enabled():
+            await ctx.send('Autoreview is currently enabled')
+        else:
+            await ctx.send('Autoreview is currently disabled')
```
Bluenix2, line 97: This could probably be simplified to a ternary (`status = "enabled" if await self.autoreview_enabled() else "disabled"`) and then an f-string.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -5,18 +5,20 @@
...
 from bot.bot import Bot
-from bot.constants import Channels, Emojis, Guild, MODERATION_ROLES, STAFF_ROLES, Webhooks
+from bot.constants import Channels, Emojis, Guild, MODERATION_ROLES, Roles, STAFF_ROLES, Webhooks
 from bot.converters import FetchedMember
 from bot.exts.moderation.watchchannels._watchchannel import WatchChannel
 from bot.exts.recruitment.talentpool._review import Reviewer
 from bot.pagination import LinePaginator
 from bot.utils import time
 
+AUTOREVIEW_ENABLED_KEY = 'autoreview_enabled'
```
Bluenix2, line 21: Python Discord uses double quotes (I won't bother with commenting on every single instance).

MarkKoz (commented):
The commands don't check the current status, causing them to sometimes perform redundant operations. For example, if autoreview is already enabled, then an enable command will still cause all reviews to be rescheduled. What do you think about adding a check to avoid that?

Apart from that, this would be ready for approval from me. My tests have shown everything is working.

bot/exts/recruitment/talentpool/_cog.py
wookie184: Fair enough, can do.

bot/exts/recruitment/talentpool/_cog.py
wookie184: I think the way it is currently looks simpler than a ternary and an f-string would.

bot/exts/recruitment/talentpool/_cog.py
wookie184: True, although the cog (and the whole bot) is so mixed already that i'm not sure it's really worth it, imo if we want to enforce it it should be through a linting workflow, as suggested in https://github.com/python-discord/bot/issues/1641. I guess I can fix this on the bit's I PRd though, will leave the rest of the cog as-is though.

MarkKoz (approved):

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -37,14 +41,69 @@ def __init__(self, bot: Bot) -> None:
...
+        await ctx.send(":white_check_mark: Autoreview disabled")
+
+    @nomination_autoreview_group.command(name="status")
+    @has_any_role(*MODERATION_ROLES)
+    async def autoreview_status(self, ctx: Context) -> None:
+        """Show whether automatic posting of reviews is enabled or disabled."""
+        if await self.autoreview_enabled():
+            await ctx.send("Autoreview is currently enabled")
+        else:
+            await ctx.send("Autoreview is currently disabled")
```
Bluenix2, line 105:
```suggestion
        """Show whether automatic posting of reviews is enabled or disabled."""
        status = "enabled" if await self.autoreview_enabled() else "disabled"
        await ctx.send(f"Autoreview is currently {status}")
```

Bluenix2 (commented):
Should we have punctuation? As in, "`:x: Autoreview is already disabled."` (currently missing).

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -25,6 +27,8 @@
 class TalentPool(WatchChannel, Cog, name="Talentpool"):
     """Relays messages of helper candidates to a watch channel to observe them."""
 
+    talentpool_settings = RedisCache()
```
mbaruh, line 30: Can you give this a description of what the keys and values are, similarly to the ones in modpings and voice_gate

mbaruh (changes requested):

bot/exts/recruitment/talentpool/_cog.py
wookie184: 👍

bot/exts/recruitment/talentpool/_cog.py
mbaruh: Neither seems preferable over the other to me. Your suggestion looks neater but the one liner is maybe slightly less readable.

wookie184:
> 
> 
> Should we have punctuation? As in, "`:x: Autoreview is already disabled."` (currently missing).

The rest of the cog doesn't seem to include full stops, and I don't think it matters too much anyway, so I don't think so, no.

mbaruh (approved):
:shipit:

PR #1708, opened 2021-08-03
Fix TypeError when infraction append is not given a reason

### LINKED ISSUE #1706: TypeError raised when using infraction append incorrectly
!i a l 3d raises the following exception:

```py
TypeError: can only concatenate str (not "NoneType") to str
```

https://github.com/python-discord/bot/blob/e65afc4e83f822f06fa2d25a7eb8c233b1cce71a/bot/exts/moderation/infraction/management.py#L58 makes reason optional

https://github.com/python-discord/bot/blob/e65afc4e83f822f06fa2d25a7eb8c233b1cce71a/bot/exts/moderation/infraction/management.py#L86

assumes that the provided reason is not None. Should be a very straightforwards fix changing `if old_reason is not None` `and reason is not None`.

MarkKoz:
Fix #1706

Not much to say. See the related issue.

### DIFF
--- a/bot/exts/moderation/infraction/management.py
+++ b/bot/exts/moderation/infraction/management.py
@@ -81,7 +81,7 @@ class ModManagement(commands.Cog):
         """
         old_reason = infraction["reason"]
 
-        if old_reason is not None:
+        if old_reason is not None and reason is not None:
             add_period = not old_reason.endswith((".", "!", "?"))
             reason = old_reason + (". " if add_period else " ") + reason
 


wookie184 (approved):
I'm not sure it makes complete sense for this command to accept an empty reason, as the only reason to use it without a reason is if you were changing the duration, in which case you should just use `!infr edit`. It's probably fine to leave it like it is, though.

mbaruh (approved):

PR #1710, opened 2021-08-03
Update Sentry SDK to 1.3

MarkKoz:
It was quite outdated. 1.0's breaking change doesn't affect us. 1.3 doesn't have any features or bug fixes that really benefit us either, but updating it will make sentry.io stop pestering us to update it.


Xithrius (approved):
I say yes.

wookie184 (approved):

PR #1711, opened 2021-08-03
Suppress 403 error when sending DEFCON reject DM

MarkKoz:
Sentry Issue: [BOT-37](https://sentry.io/organizations/python-discord/issues/2437630529)

403 occurs if the user has DMs disabled. The specific code is 50007, but I don't think 403 would be raised for any other reason here, so I didn't check the code explicitly. If someone has more insight on this, please comment.

### DIFF
--- a/bot/exts/moderation/defcon.py
+++ b/bot/exts/moderation/defcon.py
@@ -9,7 +9,7 @@ from typing import Optional, Union
 from aioredis import RedisError
 from async_rediscache import RedisCache
 from dateutil.relativedelta import relativedelta
-from discord import Colour, Embed, Member, User
+from discord import Colour, Embed, Forbidden, Member, User
 from discord.ext import tasks
 from discord.ext.commands import Cog, Context, group, has_any_role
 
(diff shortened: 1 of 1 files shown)


NIRDERIi:
Wouldn't it require to handle discord.HTTPException? Or I totally misunderstood you

wookie184:
> 
> 
> Wouldn't it require to handle discord.HTTPException? Or I totally misunderstood you

`discord.Forbidden` is a subclass of `discord.HTTPException`, httpexception could be any response from the api saying it didn't work, e.g. a random error with the Discord API, while `Forbidden` is more specific, and as in that part of the code we only want to catch disabled DMs, I believe `Forbidden` is correct.

If you were instead referring to replacing the general `Exception` with `discord.HTTPException`, then yeah, that should probably work from what I can see, although the logger logs an exception in that instance anyway, and as the comment says, sending the DM is less important than kicking the user, so we want to avoid the code failing at that point as much as possible.

wookie184 (approved):
I can't think of any other reason the bot should get a `Forbidden` error here. lgtm

mbaruh:
Just for my understading, the issue here is that `log.exception` opens an issue on Sentry? as functionality-wise it seems to be working fine already.

MarkKoz:
> 
> 
> Just for my understanding, the issue here is that `log.exception` opens an issue on Sentry? as functionality-wise it seems to be working fine already.

The issue being fixed is 403 Forbidden errors being logged as exceptions and creating issues on Sentry.

PR #1712, opened 2021-08-03
Correctly delete from cache in ModPings

mbaruh:
* Fixed wrong condition in rescheduler which made the eventual consistency not work.
* Mods now correctly removed from cache on role reapplies.


wookie184 (approved):
Good stuff.

ChrisLovering (approved):

PR #1719, opened 2021-08-04
Reorder user roles in !user command from highest to lowest

Objectivitix:
Minor PR, simply changed the slice

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -244,7 +244,7 @@ class Information(Cog):
         if on_server:
             joined = discord_timestamp(user.joined_at, TimestampFormats.RELATIVE)
             # The 0 is for excluding the default @everyone role,
-            # and the -1 is for ordering the roles from highest to lowest.
+            # and the -1 is for ordering the roles from highest to lowest (instead of the other way around).
             roles = ", ".join(role.mention for role in user.roles[:0:-1])
             membership = {"Joined": joined, "Verified": not user.pending, "Roles": roles or None}
             if not is_mod_channel(ctx.channel):
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -243,6 +243,8 @@ class Information(Cog):
 
         if on_server:
             joined = discord_timestamp(user.joined_at, TimestampFormats.RELATIVE)
+            # The 0 is for excluding the default @everyone role,
+            # and the -1 is for ordering the roles from highest to lowest.
             roles = ", ".join(role.mention for role in user.roles[:0:-1])
             membership = {"Joined": joined, "Verified": not user.pending, "Roles": roles or None}
             if not is_mod_channel(ctx.channel):
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -243,7 +243,7 @@ class Information(Cog):
 
         if on_server:
             joined = discord_timestamp(user.joined_at, TimestampFormats.RELATIVE)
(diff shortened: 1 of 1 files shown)


bot/exts/info/information.py
```diff
@@ -243,7 +243,7 @@ async def create_user_embed(self, ctx: Context, user: FetchedMember) -> Embed:
 
         if on_server:
             joined = discord_timestamp(user.joined_at, TimestampFormats.RELATIVE)
-            roles = ", ".join(role.mention for role in user.roles[1:])
+            roles = ", ".join(role.mention for role in user.roles[:0:-1])
```
ChrisLovering, line 246:
```suggestion
            roles = ", ".join(role.mention for role in reversed(user.roles))
```
I find this easier to read, so it's clear what this is doing to people coming across this code in future.

This also has the benefit of just being a generator, rather than making a whole new list

ChrisLovering (changes requested):

bot/exts/info/information.py
Objectivitix: I agree, but we'll still need a slice for removing the `@everyone` role, right? That's why the `[1:]` was there and that's why I didn't do `[::-1]`, but I did `[:0:-1]`

Objectivitix: If we need a slice, THEN a reversed iterator, I believe everything in a slice is better.

Objectivitix: Perhaps this can be resolved with a comment explaining the slice, so we get the best of both worlds?

ChrisLovering: Ah right, yea that would make sense why the :0 is there then. Could you add a comment to explain that? :D 

Seeing [:0:-1] just looked wrong to me, but it makes sense now. So I think a comment could really help the next person that glances at this and may think the same

Objectivitix: Got it, comment added, resolved!

ChrisLovering (approved):

bot/exts/info/information.py
```diff
@@ -243,7 +243,9 @@ async def create_user_embed(self, ctx: Context, user: FetchedMember) -> Embed:
 
         if on_server:
             joined = discord_timestamp(user.joined_at, TimestampFormats.RELATIVE)
-            roles = ", ".join(role.mention for role in user.roles[1:])
+            # The 0 is for excluding the default @everyone role,
+            # and the -1 is for ordering the roles from highest to lowest.
```
Bluenix2, line 247: This makes it seem like you sort them. Just say that it reverses the order.

Objectivitix: Well, it *is* sorted - from lowest to highest. `-1` makes it highest to lowest. I'll add "instead of the other way around" to the end of the comment to clear any possible confusion.

bot/exts/info/information.py
```diff
@@ -243,7 +243,9 @@ async def create_user_embed(self, ctx: Context, user: FetchedMember) -> Embed:
 
         if on_server:
             joined = discord_timestamp(user.joined_at, TimestampFormats.RELATIVE)
-            roles = ", ".join(role.mention for role in user.roles[1:])
+            # The 0 is for excluding the default @everyone role,
+            # and the -1 is for ordering the roles from highest to lowest (instead of the other way around).
```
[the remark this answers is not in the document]
Objectivitix: Yeah, alright.

Objectivitix:
We might want to squash merge this.

Bluenix2 (approved):
Good enough, thanks!

PR #1725, opened 2021-08-06
added escape markdown in PythonNews

Ryu1845:
Mailing lists embeds don't escape markdown which means Dunder methods are underlined, this PR tries to fix it.
(This is my first PR sorry if I'm making something wrong)


bot/exts/info/python_news.py
```diff
@@ -72,6 +73,12 @@ async def get_webhook_names(self) -> None:
             if mail["name"].split("@")[0] in constants.PythonNews.mail_lists:
                 self.webhook_names[mail["name"].split("@")[0]] = mail["display_name"]
 
+    @staticmethod
+    def escape_markdown(content: str) -> str:
+        """Escape the markdown underlines."""
+        # taken from discord.utils.escape_markdown
+        return re.sub(r'[_\\~|]', lambda match: '\\' + match[0], content, 0, re.MULTILINE)
```
Bluenix2, line 80: If this is copied from `discord.utils.escape_markdown`, why aren't we using it?

Ryu1845: Actually I removed most of the code it's not really the same function anymore. We don't because apparently we want to keep some markdown
https://discord.com/channels/267624335836053506/635950537262759947/873320120494211152

Ryu1845: I should remove this comment

Shivansh-007 (approved):

Numerlor (changes requested):
Why are the other chars being escaped? The docstring only mentions underlines.

With strikethrough I'd just let it apply the Markdown as that was most likely the intention
I guess spoilers are fine although I don't think they've ever shown up, and why are the backslashes being escaped?

Also, please use double quotes as stated in the [style guide](https://pythondiscord.com/pages/guides/pydis-guides/contributing/style-guide/#quote-marks)

bot/exts/info/python_news.py
```diff
@@ -72,6 +73,11 @@ async def get_webhook_names(self) -> None:
             if mail["name"].split("@")[0] in constants.PythonNews.mail_lists:
                 self.webhook_names[mail["name"].split("@")[0]] = mail["display_name"]
 
+    @staticmethod
+    def escape_markdown(content: str) -> str:
+        """Escape the markdown underlines."""
+        return re.sub(r'[_\\~|]', lambda match: '\\' + match[0], content, 0, re.MULTILINE)
```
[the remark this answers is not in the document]
Numerlor: Actually, why is the MULTILINE flag used?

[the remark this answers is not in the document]
Ryu1845: Thank you for your review, I'll change the code to reflect this as I have no objection. I'll keep the lambda for the time being.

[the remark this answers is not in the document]
Ryu1845: > Actually, why is the MULTILINE flag used?

I used discord.py escape markdown util as a basis, I don't think it needs to be there right now

bot/exts/info/python_news.py
```diff
@@ -72,6 +73,11 @@ async def get_webhook_names(self) -> None:
             if mail["name"].split("@")[0] in constants.PythonNews.mail_lists:
                 self.webhook_names[mail["name"].split("@")[0]] = mail["display_name"]
 
+    @staticmethod
+    def escape_markdown(content: str) -> str:
+        """Escape the markdown underlines and spoilers."""
+        return re.sub(r"[_|]", lambda match: "\\" + match[0], content)
```
Bluenix2, line 79: Can we add `*` here too?
```suggestion
        return re.sub(r"[_*|]", lambda match: "\\" + match[0], content)
```

Bluenix2: I understand that it may capture honest markdown such as *italic*, but I would prefer we escape most markdown. If you look at the mailing list through the archive that the embed links you don't get this formatting so we might as well escape it as well?

Ryu1845: I can add it but I was asked to keep it there https://canary.discord.com/channels/267624335836053506/635950537262759947/873320120494211152

Bluenix2: Ah, right. In that case leave it, I'll mark this as resolved.

Bluenix2 (approved):
Looks good now according to what was specified.

Xithrius (approved):
Looks good. We can always update this if we find something new to escape.

Thanks for taking care of this, @Ryu1845!

PR #1729, opened 2021-08-08
"array" tag

swfarnsworth:
This PR would add a tag explaining the difference between lists and arrays and why the two terms shouldn't be used interchangeably.

image


Xithrius (approved):

wookie184:
To be honest i'm not quite sure of the use of this tag, in many online tutorials `array` and `list` are used interchangeably, and I don't see the benefit of making this distinction.

I would probably say most of the time when people say `array` in PyDis they're referring to a Python list, unless they've already explicitly said they're working with numpy.

The code example there is also a bit confusing imo as it suggests the different multiplication behaviour is a feature of a list vs an array, while it's actually a difference between python lists and numpy arrays (e.g. in the stdlib `array` module, multiplication doubles the length like a python list)

swfarnsworth:
We discussed this in a staff channel and don't agree on the intended use case, so I'm going to withdraw this.

Issue #1730, opened 2021-08-08
Monkeypatch discord.py to understand Annotated typehints


Numerlor:
The project currently enforces typehinting, but discord.py uses annotations for its own purposes which makes some annotations on command callbacks invalid typehints.

@MarkKoz proposed in #dev-contrib that converters could inherit from the type they return to alleviate this issue, but this is more limited than normal typehinting (although still better than the current hints which are just wrong and cause warnings during development).

The `typing.Annotated` type allows us to attach metadata do a typehint, where for converters we'd use `typing.Annotated[type_hint, Converter]` where a static checker would understand the argument to be of type `type_hint`. 
This makes typing commands with converters easy to do properly for all tools, but discord.py currently can't handle the Annotated type.
Locally I patched discord.py with https://paste.fuelrats.com/evoyazixaq.py which seemed to work fine from testing with a custom converter and a forward ref

MarkKoz:
Very cool; I like this approach. I'm fine with patching discord.py since the precedent has already been set (implementing root aliases). This patch isn't that bad anyway because it's not really dealing with any internal/private APIs. That being said, we would have to ensure the setter is in sync for all the special cases (like the greedy converter).

How do you think we should name the `typing.Annotated` aliases and the converters? I'm thinking that call converters should be named like `DateConverter` and the alias can be `Date = typing.Annotated[datetime.date, DateConverter]`. I suppose it might still be useful to keep the converter classes public and accessible, though I cannot think of anything off the top of my head that uses them manually.

Numerlor:
> That being said, we would have to ensure the setter is in sync for all the special cases (like the greedy converter).

I don't anticipate any more cases like that as we're really just modifying what's in the loaded params, with the greedy being some kind of an early fail according to a comment in the source. When I looked at the current master branch it looked like it was still the only piece that immediately acts on the changed annotation. But if something does pop up it shouldn't be hard to address

> How do you think we should name the `typing.Annotated` aliases and the converters? I'm thinking that call converters should be named like `DateConverter` and the alias can be `Date = typing.Annotated[datetime.date, DateConverter]`.

That sounds good, most converters are named similarly already so we can replace those with the Annotated type and add the converter suffix to the class. However I think some converters like `UserMentionOrID` could use a rename so the return type is more obvious from the name (e.g.`UserFromMentionOrID`)

> I suppose it might still be useful to keep the converter classes public and accessible, though I cannot think of anything off the top of my head that uses them manually.

[...]

PR #1731, opened 2021-08-09
Add converter typehints

Numerlor:
Adds proper typehinting to converters by assigning the names to the converter return values during type checking.

closes: #1730


bot/converters.py
```diff
@@ -567,6 +643,26 @@ async def convert(self, ctx: Context, arg: str) -> t.Optional[dict]:
...
+    TagContentConverter = str  # noqa: F811
+    SourceConverter = SourceType  # noqa: F811
+    DurationDelta = relativedelta  # noqa: F811
+    Duration = datetime  # noqa: F811
+    OffTopicName = str  # noqa: F811
+    ISODateTime = datetime  # noqa: F811
+    HushDurationConverter = int  # noqa: F811
+    UserMentionOrID = discord.User  # noqa: F811
+    FetchedUser = t.Union[discord.User, discord.Object]  # noqa: F811
+    Infraction = t.Optional[dict]  # noqa: F811
```
MarkKoz, line 664: For what it's worth, `bot.exts.moderation.infractions._utils.Infraction` exists. I don't think being slightly more specific with the dict types really adds much when it comes to JSON though.

bot/converters.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
MarkKoz, line 1: What exactly relies on this?

MarkKoz (commented):
Some places re-annotate the argument to avoid warnings from type checkers. This is no longer necessary. Ideally, all these should be removed.

For example:
https://github.com/python-discord/bot/blob/main/bot/exts/moderation/modpings.py#L90

bot/converters.py
```diff
@@ -128,6 +134,44 @@ async def convert(self, ctx: Context, list_type: str) -> str:
         return list_type
 
 
+class Extension(Converter):
```
MarkKoz, line 137: I suppose it's okay to move these over for consistency's sake, but we should eventually have a discussion on this. I don't think it makes sense to define converters here which are unlikely to be used in multiple places. For example, `Duration` is generic and is indeed used in many places. But `OffTopicName`? That isn't used in more than one place nor will it likely ever be. We don't move cog-specific utlities to the `utils` directory, so why move such converters?

bot/converters.py
Numerlor: The SourceType return typehint in SourceConverter is imported conditionally under TYPE_CHECKING to prevent a cycle

bot/converters.py
Numerlor: I have wondered about this before; with this PR I decided to move them so the type reassignments are all in one place. I agree that it'd make more sense if they were in the modules of the cogs that use them though

MarkKoz:
> 
> 
> Some places re-annotate the argument to avoid warnings from type checkers. This is no longer necessary. Ideally, all these should be removed.
> 
> For example:
> https://github.com/python-discord/bot/blob/main/bot/exts/moderation/modpings.py#L90

Do you think you can go through the code base and remove these? Otherwise, this PR looks good to me.

bot/converters.py
Numerlor: I think it could be worth it with a TypedDict in the infractions package

Numerlor:
> > Some places re-annotate the argument to avoid warnings from type checkers. This is no longer necessary. Ideally, all these should be removed.
> > For example:
> > https://github.com/python-discord/bot/blob/main/bot/exts/moderation/modpings.py#L90
> 
> Do you think you can go through the code base and remove these? Otherwise, this PR looks good to me.

did a regex search with `^ *\w+: [\.\[\]\w]+$`, removed the two it found in the last commit but it seems a bit low so let me know if it missed anything obvious

MarkKoz (approved):

Numerlor:
Noticed `TagContentConverter` was still around even though tags have been file based for a while now so I removed it in 37804ce14c4aeb322ef772b14e7fe55658f80ebf

ChrisLovering (approved):

PR #1732, opened 2021-08-09
Fixed error message to match true value

SuperstalkerX:
When the limit was raised, the error message was not edited to reflect the change.
https://github.com/python-discord/bot/commit/6f45d6896adb3f05962733cec8e5db199def20bc


Xithrius (approved):
Good catch! Thanks for this PR!

RohanJnr (approved):

Issue #1736, opened 2021-08-11
Dynamically insert help categories into DORMANT_MSG


TizzySaurus:
As [suggested by Lemon](https://discord.com/channels/267624335836053506/429409067623251969/874919321283993631) in #community-meta, this will insert the correct category names into the DORMANT_MSG that's sent when a channel is moved to the dormant category.

NIRDERIi:
By the solution you showed in #dev-contrib, it seems ok to me!

PR #1738, opened 2021-08-11
Make DORMANT_MSG Dynamic

TizzySaurus:
Solves #1736

`DORMANT_MSG` now has the category names dynamically set.

### DIFF
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -387,7 +387,14 @@ class HelpChannels(commands.Cog):
         )
 
         log.trace(f"Sending dormant message for #{channel} ({channel.id}).")
-        embed = discord.Embed(description=_message.DORMANT_MSG)
+        available_category = self.bot.get_channel(constants.Categories.help_available)
+        embed = discord.Embed(
+            description=_message.DORMANT_MSG.format(
+                dormant=channel.category.name,
+                available=available_category.name,
+                asking_guide=_message.ASKING_GUIDE_URL
+            )
+        )
         await channel.send(embed=embed)
 
         log.trace(f"Pushing #{channel} ({channel.id}) into the channel queue.")
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -387,10 +387,11 @@ class HelpChannels(commands.Cog):
         )
 
         log.trace(f"Sending dormant message for #{channel} ({channel.id}).")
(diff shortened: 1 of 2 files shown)


NIRDERIi:
Maybe try to somehow get the GuildChannel object and check its category? Not sure why it wouldn't be in cache tho

bot/exts/help_channels/_cog.py
```diff
@@ -387,7 +387,14 @@ async def move_to_dormant(self, channel: discord.TextChannel) -> None:
         )
 
         log.trace(f"Sending dormant message for #{channel} ({channel.id}).")
-        embed = discord.Embed(description=_message.DORMANT_MSG)
+        available_category = self.bot.get_channel(constants.Categories.help_available)
```
Bluenix2, line 390: You should probably check for None and in that case add a log, and default to the old Dormant channels text.

TizzySaurus: I've instead used `utils.channels.try_get_channel`, as suggested by Chrisjl on discord, which will always get the channel

NIRDERIi (approved):
The updated code should be working fine by the looks of it!

bot/exts/help_channels/_cog.py
```diff
@@ -387,7 +387,14 @@ async def move_to_dormant(self, channel: discord.TextChannel) -> None:
         )
 
         log.trace(f"Sending dormant message for #{channel} ({channel.id}).")
-        embed = discord.Embed(description=_message.DORMANT_MSG)
+        available_category = await channel_utils.try_get_channel(constants.Categories.help_available)
+        embed = discord.Embed(
+            description=_message.DORMANT_MSG.format(
+                dormant=channel.category.name,
```
ChrisLovering, line 393: Testing this locally, it seems the channel object here isn't actually updated with the new category when it's moved on the lines above, so the embed reads as below. As you can see, it still references the in use category.

image

ChrisLovering (changes requested):

ChrisLovering (approved):
Re-tested, all works 👌

bot/exts/help_channels/_cog.py
TizzySaurus: Fixed in [this](https://github.com/python-discord/bot/pull/1738/commits/99bfb15037ce7a6f88a403f8c1d2ab36d852874a) commit.

brad90four (approved):
For what it's worth, looks good to me

PR #1740, opened 2021-08-11
Remove added punctuation from reminder

### LINKED ISSUE #1739: Nit: Reminders always have a "." added to the end even if already punctuated
Here, it's simply unconditionally added:

https://github.com/python-discord/bot/blob/eec52390c8102c84e2e9dc8d2e05deaf54e83ce3/bot/exts/utils/reminders.py#L184

It should be fairly simple to check if a string ends with a common list of punctuation (`string.punctuation` ought to do), and in those cases not add the extra `.`.

TizzySaurus:
Closes #1739

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -181,7 +181,7 @@ class Reminders(Cog):
         )
 
         # Let's not use a codeblock to keep emojis and mentions working. Embeds are safe anyway.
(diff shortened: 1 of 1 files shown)


NIRDERIi (approved):
Should work great, nothing should really go wrong

vcokltfre (approved):

PR #1742, opened 2021-08-11
Remove redundant converters and renamed FetchedMember to MemberOrUser

### LINKED ISSUE #1681: apply_infraction assumes the user will have a mention attribute
According to the type annotation, the `user` argument may be a `discord.Object`. This does not have a `mention` attribute. This will lead to an exception when trying to send the confirmation message or log message. I don't remember why `user` can be `discord.Object` but I think it's related to users that aren't in the guild. I haven't tested to see if it's a bug in practice, but this should be looked into.

TizzySaurus:
Closes #1681

Removed converters that were made redundant by discord.py V1.6's change to UserConverter which fetches from API when user not in the cache (server).

Also renames `bot.converters.FetchedMember` to `bot.converters.MemberOrUser` to better represent its value.

Note: removed `proxy_user` stuff since [according to logs](https://grafana.pythondiscord.com/d/CeNBKKfMk/service-logs?orgId=2&refresh=5s&var-job=bot&var-search=Failed+to+fetch+user%2C+returning+a+proxy+instead&from=now-2y&to=now), it has never been used before (no search matches).


vcokltfre (approved):
All seems good to me!

NIRDERIi (approved):
Seems fine to me. Should be functional and operate properly.

bot/exts/moderation/infraction/management.py
```diff
@@ -201,27 +201,21 @@ async def infraction_edit(
     # region: Search infractions
 
     @infraction_group.group(name="search", aliases=('s',), invoke_without_command=True)
-    async def infraction_search_group(self, ctx: Context, query: t.Union[UserMention, Snowflake, str]) -> None:
```
MarkKoz, line 204: Note that it will no longer be possible to look up deleted users by their ID. Is that a use case that needs to be supported?

bot/converters.py
```diff
@@ -568,5 +499,4 @@ async def convert(self, ctx: Context, arg: str) -> t.Optional[dict]:
 
 
 Expiry = t.Union[Duration, ISODateTime]
-FetchedMember = t.Union[discord.Member, FetchedUser]
-UserMention = partial(_snowflake_from_regex, RE_USER_MENTION)
```
MarkKoz, line 572: Are `_snowflake_from_regex` or `RE_USER_MENTION` used anywhere else? If not, delete them too.

MarkKoz (changes requested):

bot/converters.py
ChrisLovering: `_snowflake_from_regex` isn't but it seems `RE_USER_MENTION` is used in `UserMentionOrID` here https://github.com/python-discord/bot/blob/main/bot/converters.py#L475

TizzySaurus: `RE_USER_MENTION` is still used but not `_snowflake_from_regex` so will remove that :+1:

bot/exts/moderation/infraction/management.py
TizzySaurus: Good catch! Discussed this internally in the mod-meta channel and we decided we still want this ability, so I'll revert this change.

TizzySaurus: Should be fixed with [this](https://github.com/python-discord/bot/pull/1742/commits/8ae21a26e4c74314d0726a37e71877f8566fe6a7) commit (also removes the `_snowflake_from_regex` function)

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -115,7 +115,7 @@ async def apply_infraction(
         self,
         ctx: Context,
         infraction: _utils.Infraction,
-        user: UserSnowflake,
+        user: MemberOrUser,
```
MarkKoz, line 118: The following code should be removed from this function

https://github.com/python-discord/bot/blob/8ae21a26e4c74314d0726a37e71877f8566fe6a7/bot/exts/moderation/infraction/_scheduler.py#L168-L174

bot/exts/moderation/infraction/_utils.py
```diff
@@ -75,7 +71,7 @@ async def post_user(ctx: Context, user: UserSnowflake) -> t.Optional[dict]:
 
 async def post_infraction(
         ctx: Context,
-        user: UserSnowflake,
+        user: MemberOrUser,
```
MarkKoz, line 74: These `getattr` can be replaced with normal attribute access

https://github.com/python-discord/bot/blob/8ae21a26e4c74314d0726a37e71877f8566fe6a7/bot/exts/moderation/infraction/_utils.py#L56-L59

bot/exts/moderation/infraction/management.py
```diff
@@ -201,29 +201,26 @@ async def infraction_edit(
...
         infraction_list = await self.bot.api_client.get(
             'bot/infractions/expanded',
             params={'user__id': str(user.id)}
         )
 
-        user = self.bot.get_user(user.id)
-        if not user and infraction_list:
-            # Use the user data retrieved from the DB for the username.
-            user = infraction_list[0]["user"]
-            user = escape_markdown(user["name"]) + f"#{user['discriminator']:04}"
```
MarkKoz, line 223: Is it not better to display this information over just an ID?

bot/exts/moderation/infraction/management.py
```diff
@@ -201,29 +201,26 @@ async def infraction_edit(
...
             await self.search_user(ctx, discord.Object(query))
-        else:
+        elif isinstance(query, str):
             await self.search_reason(ctx, query)
+        else:
+            await self.search_user(ctx, query)
 
     @infraction_search_group.command(name="user", aliases=("member", "id"))
-    async def search_user(self, ctx: Context, user: t.Union[discord.User, proxy_user]) -> None:
+    async def search_user(self, ctx: Context, user: t.Union[MemberOrUser, discord.Object]) -> None:
```
MarkKoz, line 214: Shouldn't this use the `Snowflake` converter too?

MarkKoz (changes requested):
Still just a code review; haven't tested anything locally yet.

bot/exts/moderation/infraction/management.py
TizzySaurus: It needed to be `discord.Object` because that's what I passed in line 207.

Either I keep this as `discord.Object`, or I make it so that L207 passes the int and have the type annotation as `t.Union[MemberOrUser, t.Union[Snowflake, int]]`. Which of these 2 do you prefer?

bot/exts/moderation/infraction/management.py
TizzySaurus: This does still give this information just via an isinstance check instead (see the new [L221](https://github.com/python-discord/bot/blob/remove-redundant-converters/bot/exts/moderation/infraction/management.py#L221)). 

I am about to do a commit to add the `escape_markdown` though as I forgot that, which edits how this works to
```py
        if isinstance(user, (discord.Member, discord.User)):
            user_str = escape_markdown(user.name) + user.discriminator
        else:
            user_str = str(user.id)
```

bot/exts/moderation/infraction/_scheduler.py
TizzySaurus: Have removed this locally :+1:

bot/exts/moderation/infraction/_utils.py
TizzySaurus: Fixed locally :+1:

TizzySaurus (commented):
Have now pushed fixes for all of these bar the type annotation.

bot/exts/moderation/infraction/management.py
MarkKoz: #1731 would fix annotations of converters. Anyway, I suppose the extra validation from `Snowflake` isn't really necessary. I think the group command uses it to disambiguate reasons that look like IDs.

bot/exts/moderation/infraction/management.py
MarkKoz: @TizzySaurus You misunderstood me. The lines I commented were only used when a `discord.Object` was passed. In that case, all it has is an ID, and it sucks to only display an ID. Instead, it relies on the last known information in the database, if it exists. If not, then it can fall back to just an ID.

Furthermore, what you had previously was sufficient when it's a `Member` or `User`. They have already have a `__str__` which does exactly what you changed it to.

TizzySaurus: Right yes, I understand your point now of displaying just id vs name#discrim. I'll add this functionality back.

In regards to the `Member`/`User`'s `__str__` method, it doesn't escape any markdown which is why I put it to this (what it was before I changed anything).

MarkKoz: Wouldn't `escape_markdown(str(user))` be simpler at least?

TizzySaurus: I suppose we'd want something like
```py
        if isinstance(user, (discord.Member, discord.User)):
            user_str = escape_markdown(user.name) + user.discriminator
        else:  # is discord.Object instance
            if infraction_list:  # can get the name#discrim from DB
                user = infraction_list[0]["user"]
                user_str = escape_markdown(user["name"]) + f"#{user['discriminator']:04}"
            else:  # can't get name from DB so just use the id
                user_str = str(user.id)
```? (without the comments)

Comment (MarkKoz) [reply bot/exts/moderation/infraction/management.py:223]: Yes but I believe the first case can be simplified to `escape_markdown(str(user))` (you forgot the `#` anyway).

Comment (TizzySaurus) [reply bot/exts/moderation/infraction/management.py:223]: Right yes, `escape_markdown(str(user))` would be easier so I'll go with that.
```py
        if isinstance(user, (discord.Member, discord.User)):
            user_str = escape_markdown(str(user))
        else:  # is discord.Object instance
            if infraction_list:  # can get the name#discrim from DB
                user = infraction_list[0]["user"]
                user_str = escape_markdown(user["name"]) + f"#{user['discriminator']:04}"
            else:  # can't get name from DB so just use the id
                user_str = str(user.id)
```like this?

Comment (MarkKoz) [reply bot/exts/moderation/infraction/management.py:223]: Yes, that _looks_ right. Test it for good measure before checking it in.

Comment (TizzySaurus) [reply bot/exts/moderation/infraction/management.py:223]: Pushed :+1:

File: bot/exts/moderation/infraction/management.py [hunk 7]
```diff
@@ -201,29 +201,34 @@ async def infraction_edit(
     # region: Search infractions
 
     @infraction_group.group(name="search", aliases=('s',), invoke_without_command=True)
-    async def infraction_search_group(self, ctx: Context, query: t.Union[UserMention, Snowflake, str]) -> None:
+    async def infraction_search_group(self, ctx: Context, query: t.Union[MemberOrUser, Snowflake, str]) -> None:
```
MarkKoz, line 204: It just occurred to me why this didn't originally use the normal converter. The problem is that discord.py's converters also support username and nickname strings (without a discriminator). This makes it impossible to search a reason for a string which happens to be a user's name. In practice, users are rarely, if ever specified this way, so it's preferable to prioritise the reason search.

TizzySaurus: That makes sense.

Would it be more logical to instead use `UserMentionOrId` here, since then that will remove the need of the `Snowflake` converter? We'd likely have to add a try/except around the call to the UserConverter though -- so `UserMentionOrId` becoming something like
```py
    async def convert(self, ctx: Context, argument: str) -> t.Union[discord.User, int]:
        """Convert the `arg` to a `discord.User` or `int` in the case of a deleted account."""
        match = self._get_id_match(argument) or RE_USER_MENTION.match(argument)

        if match is not None:
            try:
                return await super().convert(ctx, argument)
            except commands.UserNotFound:
                return int(argument)
        else:
            raise BadArgument(f"`{argument}` is not a User mention or a User ID.")
```The converter currently isn't used anywhere so this change wouldn't break anything elsewhere.

The only downside to this I can see is an id being passed which is a message id or something and so the passed value wouldn't work for user operations -- although this is already the case for `Snowflake` and there aren't any user operations done in `infraction_search_group` anyway so should be fine). Would just mean the bot says "No infractions could be found for that query." (as it does currently if you pass a message id since AFAIK there's no real way to differentiate between id types).

TizzySaurus: I also just found this so will fix that as well I guess (typehint says it'll be a list but it's actually a tuple)
image

MarkKoz: > Would it be more logical to instead use UserMentionOrId here, since that will remove the need of the Snowflake converter? We'd likely have to add a try/except around the call to the UserConverter though to still allow for searching of deleted accounts etc. -- so UserMentionOrId would become something like:

The `Snowflake` converter has some code that improves the likelihood of a converted input actually being a snowflake. This would further disambiguate snowflakes from reasons that look like snowflakes. However, I'm not sure how often such a long number is included in a reason; it's probably unnecessary to have this extra validation for this command.

That being said, keeping the converters separate would make `UserMentionOrId` more versatile for any future use cases. If it were to fall back to returning the snowflake, then it should return a `discord.Object` instead of an integer.

> Would just mean the bot says "No infractions could be found for that query.", as it does currently if you pass a message id -- no real way to fix this since AFAIK you can't differentiate between p.e a message id and a user id.

Yes, that's just something we have to accept.

TizzySaurus: So are you suggesting using `UserMentionOrId` (not `Snowflake`), but changing the convert method to return a `discord.Object` when the call to UserConverter fails?

MarkKoz: Either that or keeping them separate like `Union[UserMentionOrId, Snowflake, str]` so that `UserMentionOrId` remains versatile. I think the latter would be better. What do you think?

TizzySaurus: Yeah, I think keeping it versatile is better actually. I'll do that :+1:

TizzySaurus: @MarkKoz A bit of a delay because I was away camping but have just [pushed a commit](https://github.com/python-discord/bot/pull/1742/commits/3ce1272c7e556e4b1cb8f330fd2553f20842c061) for this :+1:

MarkKoz: My comment was only on `infraction_search_group`. `search_user` did not need to be updated because it only accepts users; there's no risk of ambiguity between names and reasons.

TizzySaurus: If I don't change `search_user` then I can a lint error on L211 since I'm passing a `UserMentionOrId` to a `MemberOrUser` parameter. In reality `UserMentionOrId` gives a `discord.User` so it isn't actually an issue but the linter doesn't pick up on that.

Am I okay to just ignore this linting error and revert `search_user` back to the `MemberOrUser` converter?

MarkKoz: Yes just ignore it.

MarkKoz: A type checker like mypy isn't used, so it will not fail the CI step. There is a check that a type annotation exists, but it doesn't not enforce type correctness.

TizzySaurus: It has been [committed](https://github.com/python-discord/bot/pull/1742/commits/c88698727cf56b1311e7219ec2dfa06c3a7863ce)

MarkKoz (approved):

TizzySaurus:
I've just squashed a load of commits, would appreciate someone just confirming that everything still looks good (from my quick look-over it does)

mbaruh (changes requested):
This is an outdated comment:
https://github.com/python-discord/bot/blob/4e158b2a64e6f640847f0f7d2f229c96cecb2c46/bot/exts/moderation/watchchannels/bigbrother.py#L97

Otherwise looks great

TizzySaurus:
> This is an outdated comment:
> https://github.com/python-discord/bot/blob/4e158b2a64e6f640847f0f7d2f229c96cecb2c46/bot/exts/moderation/watchchannels/bigbrother.py#L97
> 
> Otherwise looks great

What exactly is this `if` statement here doing? I feel like it's supposed to check the user is in the server and isn't staff, but that's not what it's currently doing (the `and` should be an `or` and instead of `hasattr(user, 'roles')` can use `isinstance(user, discord.User)` (although if we only want to allow people in the server then why have `MemberOrUser` as the typehint and not just `Member`)

MarkKoz:
> I feel like it's supposed to check whether the user is a staff, but would make more sense for the hasattr part to be isinstance(user, discord.Member)

I suppose it's more ducktype-y to check for the attribute instead of the whole type. However, we know the converter can only return one of two types, so it doesn't really matter in this case.

TizzySaurus:
> > I feel like it's supposed to check whether the user is a staff, but would make more sense for the hasattr part to be isinstance(user, discord.Member)
> 
> I suppose it's more ducktype-y to check for the attribute instead of the whole type. However, we know the converter can only return one of two types, so it doesn't really matter in this case.

Yeah, I guess. I'll just pretend I didn't see this and update the comment to say `discord.User` instead of `FetchedUser`.

TizzySaurus:
> This is an outdated comment:
> https://github.com/python-discord/bot/blob/4e158b2a64e6f640847f0f7d2f229c96cecb2c46/bot/exts/moderation/watchchannels/bigbrother.py#L97
> 
> Otherwise looks great

Committed :+1:

mbaruh (approved):

PR #1743, opened 2021-08-11
fix: update urls in site cog

dawnofmidnight:
Some urls in the site cog are currently outdated and lead to the wrong pages. Some of them (like the Resources page) have redirects, while others (like the Asking Good Questions page, do not). This changes the constant to the site base url and updates the rest of the urls to be valid.


wookie184:
A github search for `pythondiscord.com` brings up another case of a redirect link (not in the site cog): https://github.com/python-discord/bot/blob/d5ee28133f8e792526716b07d25cbf6e4faba3a1/bot/exts/help_channels/_message.py#L15

There are also a few other locations where `pythondiscord.com` is used rather than the constant e.g. https://github.com/python-discord/bot/blob/c1c6286fb2225a463533ba17d823b2e6a393f8ad/bot/exts/moderation/infraction/superstarify.py#L21
Do you think these would be worth changing too?

dawnofmidnight:
The `NICKNAME_POLICY_URL` seems to be correct already, but I can fix the other.

wookie184:
> 
> 
> The `NICKNAME_POLICY_URL` seems to be correct already, but I can fix the other.

Yeah that one is the correct URL, just not using the site's name constants. There were a few others like that but I think the `ASKING_GUIDE_URL` is the only one with a redirect url. Just changing that one is fine if you think the others are out of scope, replacing the text with constants isn't particularly important.

Bluenix2 (approved):
Looks good, I see nothing wrong.

mbaruh (approved):

PR #1744, opened 2021-08-11
Remove role pings when a helper vote is posted

ChrisLovering:
We have just deployed thread-bot, which pings these roles inside a thread for better focused discussion. To avoid with duplicated pings, I have removed them from here.


TizzySaurus (approved):
Should work perfectly :+1:

mbaruh (approved):

PR #1745, opened 2021-08-12
fix: Nomination message now checks historic and new style nominations

ChrisLovering:
Previously nomination messages had role pings in them, now they don't as we moved them into a thread.
Due to this, we need to detect both styles of message in the interim of historic nominations existing.

A 'proper' fix for this is to store the nomination message IDs when we post them against the
nomination object in the site api. We are planing to work on this soon, this commit is a
short term fix.

### DIFF
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -33,10 +33,12 @@ MAX_MESSAGE_SIZE = 2000
 # Maximum amount of characters allowed in an embed
 MAX_EMBED_SIZE = 4000
 
-# Regex finding the user ID of a user mention
-MENTION_RE = re.compile(r"<@!?(\d+?)>")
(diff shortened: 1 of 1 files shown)


wookie184 (approved):
Looks good to me

ChrisLovering:
Ran into a bug where it would delete all messages in the channel under some circumstances...investigating a fix

ChrisLovering:
Ok, false alarm. The issue was caused due to me copying the text from another nomination to test, rather than using post_review or get_review. Due to this, the message contained a bunch of spaces that caused the regex to fail. This would also happen with the current system, so the proper fix is to upgrade to storing the message IDs, rather than use regex. This will do for now though.

wookie184 (approved):
Dunno if there's any need for approving again, but I will anyway, lgtm!

Akarys42 (approved):

PR #1749, opened 2021-08-13
Update reminders to reply instead of using a jump url to the origin message

### LINKED ISSUE #1741: Reply to the original message where the reminder was invoked instead of using a jump url
I believe, replying to the original message where the reminder was invoked is more consistent than using a jump url.
For that we would have to modify `jump_url` to be `message_id` instead
```python
        reminder = await self.bot.api_client.post(
            'bot/reminders',
            json={
                'author': ctx.author.id,
                'channel_id': ctx.message.channel.id,
                'jump_url': ctx.message.jump_url,
                'content': content,
                'expiration': expiration.isoformat(),
                'mentions': mention_ids,
            }
        )
```
Although, that would involve a small modification on the site. So that should be discussed more in depth.

D0rs4n:
Closes #1741 
I implemented the changes mentioned, I also added a fallback to use a jump-url if something goes wrong when trying to get the partial message.


bot/exts/utils/reminders.py
```diff
@@ -172,32 +172,39 @@ async def send_reminder(self, reminder: dict, expected_time: datetime = None) ->
...
-        embed.description = f"Here's your reminder: {reminder['content']}"
-
-        if reminder.get("jump_url"):  # keep backward compatibility
-            embed.description += f"\n[Jump back to when you created the reminder]({reminder['jump_url']})"
-
         if expected_time:
             embed.colour = discord.Colour.red()
             embed.set_author(
                 icon_url=Icons.remind_red,
                 name=f"Sorry it should have arrived {time_since(expected_time)} !"
```
Xithrius, line 180: While you're doing this PR, we seem to have a little bug. It's a bit broken at this line (see screenshot below). We should also change the wording to `Sorry, your reminder should have arrived at <time here> !`

Fix this up and we should be good to go!

image

D0rs4n: Yes, it seems like you can't put discord timestamps into the author field, so it seems like we have to go with the second message :(

Xithrius: Unfortunate. Please implement, either way.

Xithrius (approved):
Does the do.

Akarys42 (approved):
Another one in the book, thanks a lot @D0rs4n!

PR #1752, opened 2021-08-16
Help channels: fix initialisation of available channels set

### LINKED ISSUE #1715: Set may be missing some available help channels after initialisation
Sentry issue: [BOT-Z1](https://sentry.io/organizations/python-discord/issues/2354932564/?referrer=github_integration)

If the cog is reloaded while there are less than the maximum amount of available channels, it makes some channels available until the limit is reached. When a channel is made available, it updates the `available_help_channels` set. The `update_available_help_channels()` function will not update this set if it sees that the set already contains elements. This results in only the channels that were just made available being in the set; the set will not contain the channels that were already available when the bot started.

```
KeyError: <TextChannel id=776184670794678303 name='help-mushroom' position=268 nsfw=False news=False category_id=696958401460043776>
...
  File "bot/exts/help_channels/_cog.py", line 469, in on_message
    await self.claim_channel(message)
  File "bot/utils/lock.py", line 108, in wrapper
    log.debug(
  File "bot/utils/lock.py", line 108, in wrapper
    log.debug(
  File "bot/utils/lock.py", line 108, in wrapper
    log.debug(
  File "bot/exts/help_channels/_cog.py", line 150, in claim_channel
    self.available_help_channels.remove(message.channel)

```

MarkKoz:
Fixes #1715

If the cog is reloaded while there are less than the maximum amount of available channels, it makes some channels available until the limit is reached. When a channel is made available, it updates the `available_help_channels` set. The `update_available_help_channels()` function would not update this set if it sees that the set already contains elements. This resulted in only the channels that were just made available being in the set; the set would not contain the channels that were already available when the bot started.

Fix this by unconditionally populating the set, but moving it to `init_available()` so it only happens once.

Also refactor the changes made by #1738 to simplify the implementation.

### DIFF
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -267,9 +267,7 @@ class HelpChannels(commands.Cog):
             for channel in channels[:abs(missing)]:
                 await self.unclaim_channel(channel, closed_on=_channel.ClosingReason.CLEANUP)
 
-        self.available_help_channels = {
-            c for c in self.available_category.channels if not _channel.is_excluded_channel(c)
-        }
+        self.available_help_channels = set(_channel.get_category_channels(self.available_category))
 
         # Getting channels that need to be included in the dynamic message.
         await self.update_available_help_channels()
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -391,13 +391,10 @@ class HelpChannels(commands.Cog):
         )
 
         log.trace(f"Sending dormant message for #{channel} ({channel.id}).")
-        dormant_category = await channel_utils.try_get_channel(constants.Categories.help_dormant)
-        available_category = await channel_utils.try_get_channel(constants.Categories.help_available)
         embed = discord.Embed(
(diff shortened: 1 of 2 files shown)


ChrisLovering (approved):
Yup LGTM!

bot/exts/help_channels/_cog.py
```diff
@@ -267,6 +267,10 @@ async def init_available(self) -> None:
             for channel in channels[:abs(missing)]:
                 await self.unclaim_channel(channel, closed_on=_channel.ClosingReason.CLEANUP)
 
+        self.available_help_channels = {
+            c for c in self.available_category.channels if not _channel.is_excluded_channel(c)
+        }
```
wookie184, line 272: I think you could use `_channel.get_category_channels` here instead
https://github.com/python-discord/bot/blob/4c9405b21341880d9bdc1f0dde02219de9ade005/bot/exts/help_channels/_channel.py#L32-L39

MarkKoz: Yes 052cead

wookie184 (approved):
Thanks

PR #1753, opened 2021-08-16
Delete Webhooks when one is accidentally posted.

### LINKED ISSUE #1747: Delete detected webhooks
I propose that when we detect a webhook that's been posted in the guild, we call `DELETE` on that URL, so that the webhook is deleted.

This is to protect users against themselves, since there is the chance that their message was scraped before we deleted it, and someone now has that webhook url.

Implementing this would be as simple as calling DELETE on any webhook URLs that we detect.

If we do this, we should update the message in the channel to tell them that we did as much.

D0rs4n:
Closes  #1747 

From now on, webhooks that were posted in the chat will be also
deleted from Discord in order to eliminate the risk.

### DIFF
--- a/bot/exts/filters/webhook_remover.py
+++ b/bot/exts/filters/webhook_remover.py
@@ -13,8 +13,8 @@ WEBHOOK_URL_RE = re.compile(r"((?:https?://)?discord(?:app)?\.com/api/webhooks/\
 
 ALERT_MESSAGE_TEMPLATE = (
     "{user}, looks like you posted a Discord webhook URL. Therefore, your "
-    "message has been removed. Your webhook may have been **compromised** so "
-    "please re-create the webhook **immediately**. If you believe this was a "
+    "message has been removed, alongside with your webhook"
+    "you can re-create it if you wish to. If you believe this was a "
     "mistake, please let us know."
 )
 
@@ -32,7 +32,7 @@ class WebhookRemover(Cog):
         """Get current instance of `ModLog`."""
         return self.bot.get_cog("ModLog")
 
-    async def delete_and_respond(self, msg: Message, redacted_url: str) -> None:
+    async def delete_and_respond(self, msg: Message, redacted_url: str, webhook_deleted: bool) -> None:
         """Delete `msg` and send a warning that it contained the Discord webhook `redacted_url`."""
         # Don't log this, due internal delete, not by user. Will make different entry.
         self.mod_log.ignore(Event.message_delete, msg.id)
@@ -44,9 +44,12 @@ class WebhookRemover(Cog):
             return
 
         await msg.channel.send(ALERT_MESSAGE_TEMPLATE.format(user=msg.author.mention))
-
+        if webhook_deleted:
+            delete_state = "The webhook was successfully deleted."
+        else:
+            delete_state = "There was an error when deleting the webhook, it might have already been removed."
         message = (
-            f"{format_user(msg.author)} posted a Discord webhook URL to {msg.channel.mention}. "
+            f"{format_user(msg.author)} posted a Discord webhook URL to {msg.channel.mention}.{delete_state} "
             f"Webhook URL was `{redacted_url}`"
         )
         log.debug(message)
@@ -72,7 +75,12 @@ class WebhookRemover(Cog):
 
         matches = WEBHOOK_URL_RE.search(msg.content)
         if matches:
-            await self.delete_and_respond(msg, matches[1] + "xxx")
+            async with self.bot.http_session.delete(msg.content) as resp:
+                # The Discord API Returns a 204 NO CONTENT repsonse on success.
+                if resp.status == 204:
+                    await self.delete_and_respond(msg, matches[1] + "xxx", True)
+                else:
+                    await self.delete_and_respond(msg, matches[1] + "xxx", False)
 
     @Cog.listener()
     async def on_message_edit(self, before: Message, after: Message) -> None:
--- a/bot/exts/filters/webhook_remover.py
+++ b/bot/exts/filters/webhook_remover.py
@@ -9,7 +9,10 @@ from bot.constants import Channels, Colours, Event, Icons
 from bot.exts.moderation.modlog import ModLog
 from bot.utils.messages import format_user
 
-WEBHOOK_URL_RE = re.compile(r"((?:https?://)?discord(?:app)?\.com/api/webhooks/\d+/)\S+/?", re.IGNORECASE)
+WEBHOOK_URL_RE = re.compile(
+    r"((?:https?:\/\/)?(?:ptb\.|canary\.)?discord(?:app)?\.com\/api\/webhooks\/\d+\/)\S+\/?",
+    re.IGNORECASE
+)
 
 ALERT_MESSAGE_TEMPLATE = (
     "{user}, looks like you posted a Discord webhook URL. Therefore, your "
--- a/bot/exts/filters/webhook_remover.py
+++ b/bot/exts/filters/webhook_remover.py
@@ -16,7 +16,7 @@ WEBHOOK_URL_RE = re.compile(
 
 ALERT_MESSAGE_TEMPLATE = (
     "{user}, looks like you posted a Discord webhook URL. Therefore, your "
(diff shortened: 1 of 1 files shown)


bot/exts/filters/webhook_remover.py
```diff
@@ -13,8 +13,8 @@
 
 ALERT_MESSAGE_TEMPLATE = (
     "{user}, looks like you posted a Discord webhook URL. Therefore, your "
-    "message has been removed. Your webhook may have been **compromised** so "
-    "please re-create the webhook **immediately**. If you believe this was a "
+    "message has been removed, alongside with your webhook"
+    "you can re-create it if you wish to. If you believe this was a "
     "mistake, please let us know."
```
Bluenix2, line 18: This says
> Bluenix, looks like you posted a Discord webhook URL. Therefore, your message has been removed, alongside with your webhook you can re-create it if you wish to. If you believe this was a mistake, please let us know.

You should add a period between "your webhook" and "you can re-create it if you wish to".

bot/exts/filters/webhook_remover.py
```diff
@@ -44,9 +44,12 @@ async def delete_and_respond(self, msg: Message, redacted_url: str) -> None:
...
 
         await msg.channel.send(ALERT_MESSAGE_TEMPLATE.format(user=msg.author.mention))
-
+        if webhook_deleted:
+            delete_state = "The webhook was successfully deleted."
+        else:
+            delete_state = "There was an error when deleting the webhook, it might have already been removed."
         message = (
-            f"{format_user(msg.author)} posted a Discord webhook URL to {msg.channel.mention}. "
+            f"{format_user(msg.author)} posted a Discord webhook URL to {msg.channel.mention}.{delete_state} "
```
Bluenix2, line 52:
```suggestion
            f"{format_user(msg.author)} posted a Discord webhook URL to {msg.channel.mention}. {delete_state} "
```
You don't have a space in `delete_state` so we need to add it here

bot/exts/filters/webhook_remover.py
```diff
@@ -32,7 +32,7 @@ def mod_log(self) -> ModLog:
         """Get current instance of `ModLog`."""
         return self.bot.get_cog("ModLog")
 
-    async def delete_and_respond(self, msg: Message, redacted_url: str) -> None:
+    async def delete_and_respond(self, msg: Message, redacted_url: str, webhook_deleted: bool) -> None:
```
Numerlor, line 35: Making `webhook_deleted` a kwarg only param would make the interface clearer from the caller

bot/exts/filters/webhook_remover.py
```diff
@@ -72,7 +75,12 @@ async def on_message(self, msg: Message) -> None:
 
         matches = WEBHOOK_URL_RE.search(msg.content)
         if matches:
-            await self.delete_and_respond(msg, matches[1] + "xxx")
+            async with self.bot.http_session.delete(msg.content) as resp:
+                # The Discord API Returns a 204 NO CONTENT repsonse on success.
+                if resp.status == 204:
+                    await self.delete_and_respond(msg, matches[1] + "xxx", True)
+                else:
+                    await self.delete_and_respond(msg, matches[1] + "xxx", False)
```
Numerlor, line 83: The connection doesn't need to be held open to send the discord message, we can also get rid of the conditional
```suggestion
            async with self.bot.http_session.delete(msg.content) as resp:
                # The Discord API Returns a 204 NO CONTENT repsonse on success.
                deleted_successfully = resp.status == 204
            await self.delete_and_respond(msg, matches[1] + "xxx", deleted_successfully)
```

bot/exts/filters/webhook_remover.py
```diff
@@ -72,7 +75,12 @@ async def on_message(self, msg: Message) -> None:
 
         matches = WEBHOOK_URL_RE.search(msg.content)
         if matches:
-            await self.delete_and_respond(msg, matches[1] + "xxx")
+            async with self.bot.http_session.delete(msg.content) as resp:
+                # The Discord API Returns a 204 NO CONTENT repsonse on success.
```
Numerlor, line 79:
```suggestion
                # The Discord API Returns a 204 NO CONTENT response on success.
```

Numerlor (changes requested):

bot/exts/filters/webhook_remover.py
D0rs4n: Valid point. Very inefficient, it is.

bot/exts/filters/webhook_remover.py
```diff
@@ -72,7 +75,10 @@ async def on_message(self, msg: Message) -> None:
 
         matches = WEBHOOK_URL_RE.search(msg.content)
         if matches:
-            await self.delete_and_respond(msg, matches[1] + "xxx")
+            async with self.bot.http_session.delete(msg.content) as resp:
```
ChrisLovering, line 78: Won't `msg.content` be the full message here? I think you should be using `matches[1]`, as that's the actual webhook url.

bot/exts/filters/webhook_remover.py
```diff
@@ -13,8 +13,8 @@
 
 ALERT_MESSAGE_TEMPLATE = (
     "{user}, looks like you posted a Discord webhook URL. Therefore, your "
-    "message has been removed. Your webhook may have been **compromised** so "
-    "please re-create the webhook **immediately**. If you believe this was a "
+    "message has been removed, alongside with your webhook."
```
ChrisLovering, line 16:
```suggestion
    "message has been removed, and your webhook has been deleted."
```

ChrisLovering (changes requested):
Could you add a screenshot of the alert message to this Pr so we can see what it looks like in Discord?

bot/exts/filters/webhook_remover.py
ChrisLovering: Ah, it seems the regex doesn't match the webhook secret, just the url and ID. This will need to be changed

D0rs4n: Yes, how foolish of me. But I believe the full url is `matches[0]`

bot/exts/filters/webhook_remover.py
```diff
@@ -72,7 +75,10 @@ async def on_message(self, msg: Message) -> None:
 
         matches = WEBHOOK_URL_RE.search(msg.content)
         if matches:
-            await self.delete_and_respond(msg, matches[1] + "xxx")
+            async with self.bot.http_session.delete(matches[0]) as resp:
```
ChrisLovering, line 78: After testing, this still doesn't work. It seems the regex does not capture the `https://` part of the webhook, this will need to be added back in before you can make a request to delete it.

ChrisLovering (changes requested):

bot/exts/filters/webhook_remover.py
D0rs4n: That's strange, I've tested out the regEx and by using `matches[0]` it worked just fine, it included the whole url.

D0rs4n: Turns out the regex did not match the `https://` part in case of canary urls. The regEx pattern has to be changed.

ChrisLovering (approved):
Tested locally, works well 👍

bot/exts/filters/webhook_remover.py
```diff
@@ -9,12 +9,15 @@
...
+WEBHOOK_URL_RE = re.compile(
+    r"((?:https?:\/\/)?(?:ptb\.|canary\.)?discord(?:app)?\.com\/api\/webhooks\/\d+\/)\S+\/?",
+    re.IGNORECASE
+)
 
 ALERT_MESSAGE_TEMPLATE = (
     "{user}, looks like you posted a Discord webhook URL. Therefore, your "
-    "message has been removed. Your webhook may have been **compromised** so "
-    "please re-create the webhook **immediately**. If you believe this was a "
+    "message has been removed, and your webhook has been deleted."
```
Bluenix2, line 19: I don't understand, I saw the screenshot of this message which looked like there was a space after this period.
```suggestion
    "message has been removed, and your webhook has been deleted. "
```

Bluenix2 (approved):
This looks good now

PR #1756, opened 2021-08-19
Formatting issues on the !server command

ChrisLovering:
This fixes linebreak formatting on the server command.
The `features` part of the embed isn't included in some channels, this leads to there not being a linebreak between voice regions and roles in the embed. By changing it to this, rather than dedent, we specify exact where we want the linebreaks to be.

This also adds comma separators to the large member counts in the embed too.


doublevcodes (approved):
:shipit: LGTM

Akarys42 (approved):
I approve of this pickiness

dawnofmidnight (approved):

PR #1760, opened 2021-08-21
Cross-channel and deleted messages anti-spam

### LINKED ISSUE #622: Utilize a Message Cache for the Antispam Cog
For every new message event the Antispam cog makes a query for the channel's message history in order to build a temporary message cache for the rules to act on: 

https://github.com/python-discord/bot/blob/dd3b2de9bf8a977ad987c2ab9a74d8be6a724173/bot/cogs/antispam.py#L155-L160

While we haven't yet encountered a rate limiting issue with this implementation, it may not be the most ideal approach to make an API call to Discord for every new message on the server.

d.py does provide an internal message cache, but this does not look to be used `.history` calls, as their intent is to get fresh message data. Before considering implementing our own caching system (which could have uses in future enhancements), it would be good to look into the possibility of utilizing this cache, with the API as a fallback.

### LINKED ISSUE #665: Support Anti-Spam Checks Across Channels
There should be an anti-spam filter which checks message history across several channels. This should catch users who send messages in multiple channels which are duplicates and/or sent in quick succession. I'm not sure on the implementation. It seems complicated since history can only be retrieved on a per-channel basis, and we have a lot of channels.

Original message:

> So for context, we just had an incident where someone was auto spamming a channel.  The bot caught the initial spam, gave them the 10 minute mute, as expected.  However, after they were unmuted, the spam continued, at the same rate and enough that it should have triggered a second mute.  It didn't.  This needs to be investigated.

### LINKED ISSUE #1522: Keep deleted messages in cache
We want deleted messages to remain in the cache so that the bot can account for them when applying spam rules.

One way of doing that is monkey-patching this part https://github.com/Rapptz/discord.py/blob/master/discord/state.py#L483-L504 but it should be accompanied by proper tests to prevent any funny business on version bumps.

Another way is to cache deleted messages in a separate cache, but well.. it was already cached.

mbaruh:
Closes #622, closes #665, closes #1522

This PR gives the AntiSpam cog the ability to detect spam of users across channels, and even when their messages (or part of them) were deleted, either by them or automatically because of another filter.

## Implementation
The feature was implemented by making the on_message listener search for messages in a cache, rather than the history of the channel the message was sent in.

### The MessageCache class
The cache used is an instance of a class added in this PR, rather than the one provided by discord.py. This has several advantages:
* The collection of messages to search in can be defined to be significantly smaller, as the cog only cares about the messages sent in the last few seconds.
* discord.py removes deleted messages from the cache, while the new cache implementation does not. This allows easily handling deleted messages in the anti-spam cog.
* Because we don't remove elements from the middle of the sequence, the implementation of the cache could be made such that appending and popping from either side of it, as well as lookup by index and by message ID, can be implemented to run in constant time.

Not all features of the MessageCache (particularly slicing) are currently needed, but I preferred making a complete feature that didn't require too many future alterations, as we were recently discussing doing more things with it.

### Changes to DeletionContext
As a spam event is no longer defined by a single channel, I couldn't use it anymore to identify a deletion context in the `message_deletion_queue`. I chose to identify a spam event by the users causing it instead, allowing additional channels to be added to the deletion context, instead of allowing adding additional members as was before.

This can run into the edge-case of one member becoming irrelevant to the filter while others still are, resulting in another log message being sent, but it's an unlikely edge case since the users should be muted almost immediately, and we're currently not using any multi-member filters in the first place. If new members are causing spam in the same channels in the period of time between the creation of the deletion context, and the alerting, alerting for the new members separately seems acceptable.


bot/utils/message_cache.py
```diff
@@ -0,0 +1,186 @@
...
+        message = self._messages[self._start]
+        del self._message_id_mapping[message.id]
+        self._messages[self._start] = None
+        self._start = (self._start + 1) % self.maxlen
+
+        return message
+
+    def clear(self) -> None:
+        """Remove all messages from the cache."""
+        self._messages: list[t.Optional[Message]] = [None] * self.maxlen
```
Akarys42, line 92:
```suggestion
        self._messages = [None] * self.maxlen
```
This annotation can be removed, but I don't really care that much :P

Akarys42 (approved):
The code is very well written, and the feature works perfectly well. Thanks a lot, Zig, favourite zebra ever!

bot/utils/message_cache.py
```diff
@@ -0,0 +1,184 @@
...
+        Return the message(s) in the index or slice provided.
+
+        This method makes the circular buffer implementation transparent to the user.
+        Providing 0 will return the message at the position perceived by the user to be the beginning of the cache,
+        meaning at `self._start`.
+        """
+        if isinstance(item, int):
+            if item >= len(self) or item < -len(self):
+                raise IndexError("cache index out of range")
+            return self._messages[(item + self._start) % self.maxlen]
```
bast0006, line 128: I think we should comment here that we're relying on how python's modulus treats negative numbers.

-1 % 5 results in +4. This may not be exactly intuitive, and IIRC differs across languages.

tests/bot/utils/test_message_cache.py
```diff
@@ -0,0 +1,214 @@
...
+                    self.assertListEqual(cache[slice_], messages[slice_])
+
+    def test_slicing_with_overfilled_cache(self):
+        """Test if slicing returns the correct messages if the cache was appended with more messages it can contain."""
+        sizes = (5, 10, 55, 101)
+
+        slices = (
+            slice(None), slice(2, None), slice(None, 2), slice(None, None, 2), slice(None, None, 3), slice(-1, 2),
+            slice(-1, 3000), slice(-3, -1), slice(-10, 3), slice(-10, 4, 2), slice(None, None, -1), slice(None, 3, -2),
+            slice(None, None, -3)
```
bast0006, line 192: Can you add a triple negative slice here? `slice(-1, -10, -2)` for example

bot/utils/message_cache.py
```diff
@@ -0,0 +1,197 @@
...
+        if self.newest_first:
+            self._appendleft(message)
+        else:
+            self._appendright(message)
+
+    def _appendright(self, message: Message) -> None:
+        """Add the received message to the end of the cache."""
+        if self._is_full():
+            del self._message_id_mapping[self._messages[self._start].id]
+            self._start = (self._start + 1) % self.maxlen
```
Bluenix2, line 48: Perhaps `_start` could be a property?

bot/utils/message_cache.py
```diff
@@ -0,0 +1,197 @@
...
+            del self._message_id_mapping[self._messages[self._end].id]
+
+        self._start = (self._start - 1) % self.maxlen
+        self._messages[self._start] = message
+        self._message_id_mapping[message.id] = self._start
+
+    def pop(self) -> Message:
+        """Remove the last message in the cache and return it."""
+        if self._is_empty():
+            raise IndexError("pop from an empty cache")
```
Bluenix2, line 67:
```suggestion
            raise IndexError("cannot pop from an empty cache")
```

bot/utils/message_cache.py
```diff
@@ -0,0 +1,197 @@
...
+
+        self._start = 0
+        self._end = 0
+
+    def get_message(self, message_id: int) -> t.Optional[Message]:
+        """Return the message that has the given message ID, if it is cached."""
+        index = self._message_id_mapping.get(message_id, None)
+        return self._messages[index] if index is not None else None
+
+    def update(self, message: Message) -> bool:
```
Bluenix2, line 101: When is this used? This should be a reference to the same Message object, so it will be changed "automatically".

bot/utils/message_cache.py
```diff
@@ -0,0 +1,197 @@
...
+            else:
+                offset = ceil((start + 1) / -step) * -step - start - 1
+                return self._messages[start::step] + self._messages[self.maxlen - 1 - offset:stop:step]
+
+        else:
+            raise TypeError(f"cache indices must be integers or slices, not {type(item)}")
+
+    def __iter__(self) -> t.Iterator[Message]:
+        if self._is_empty():
+            return
```
Bluenix2, line 175: Do we not need to return an empty iterator?

bot/utils/message_cache.py
mbaruh: This is a copy of the deque error message

bot/utils/message_cache.py
mbaruh: Why is the benefit?

bot/exts/filters/antispam.py
```diff
@@ -234,14 +239,18 @@ async def punish(self, msg: Message, member: Member, reason: str) -> None:
...
             # If we have more than one message, we can use bulk delete.
             if len(messages) > 1:
                 message_ids = [message.id for message in messages]
                 self.mod_log.ignore(Event.message_delete, *message_ids)
-                await channel.delete_messages(messages)
+                channel_messages = defaultdict(list)
+                for message in messages:
+                    channel_messages[message.channel].append(message)
+                for channel, messages in channel_messages.items():
+                    await channel.delete_messages(messages)
```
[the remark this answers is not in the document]
mbaruh: I'm grouping the messages per channel to be able to use the `delete_messages` method on all messages of that channel at the same time.

bot/exts/filters/antispam.py
```diff
@@ -69,11 +71,14 @@ async def add(self, rule_name: str, members: Iterable[Member], messages: Iterabl
...
         """Method that takes care of uploading the queue and posting modlog alert."""
-        triggered_by_users = ", ".join(format_user(m) for m in self.members.values())
+        triggered_by_users = ", ".join(format_user(m) for m in self.members)
+        triggered_in_channel = f"**Triggered in:** {self.triggered_in.mention}\n" if len(self.channels) > 1 else ""
+        channels_description = ", ".join(channel.mention for channel in self.channels)
 
         mod_alert_message = (
             f"**Triggered by:** {triggered_by_users}\n"
-            f"**Channel:** {self.channel.mention}\n"
+            f"{triggered_in_channel}"
```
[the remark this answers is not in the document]
mbaruh: The new line is in the value of the variable. If there is only one channel then the string does not appear at all and I don't want a new line in that case.

bot/utils/message_cache.py
mbaruh: This is used in the antispam's on_message_edit. The listener receives two different Message objects.

bot/utils/message_cache.py
mbaruh: Iteration stops when you return. There's no difference between having something to iterate on first and then return, and having nothing to iterate on and then return.

Bluenix2: Ah interesting, doesn't type checkers complain though?

mbaruh: Nothing that I've seen so far. Once you have a `yield` statement the function returns a generator which you use to get the values, and a generator is an iterator.

wookie184 (approved):
I haven't looked through the `MessageCache` implementation too closely, but the rest looks great and seems to work perfectly.

Bluenix2 (approved):
Thank you 🦓. Looks good!

PR #1762, opened 2021-08-22
Metabase sharing

ChrisLovering:
This adds the ability to publicly share metabase questions via the bot, since admins don't have admin access to metabase, but the bot does.

This also refactors the error handling in the metabase cog to avoid copy pasting.


jb3 (approved):

Xithrius (approved):
Very good.

PR #1764, opened 2021-08-22
Fix edge-case of `user.joined_at` being `None`  in userinfo command.

TizzySaurus:
Solves #1763 

Now displays "Unable to get join date" when `user.joined_at` is `None`.


mbaruh (approved):
Epic

Akarys42 (approved):
Let's make sure @mbaruh can sleep tonight. Thanks Tizzy!

PR #1766, opened 2021-08-22
Update time format in reminder message to DAY_TIME from RELATIVE

### LINKED ISSUE #1750: Use DAY_TIME or DATE_TIME  timestamp format for reminders
The reminders display the relative time when the reminder is scheduled, but the relative duration was already given by the user when they scheduled the reminder. It's also unlikely that someone will look at the reminder scheduled message in the past (apart from jumping to it when the reminder is sent) so in most cases the relative timestamp only restates what the user already knows; I think using the DAY_TIME (F) or DATE_TIME (f) would make more sense in this case

This is how they compare to each other, with DATE_TIME being the second format and DAY_TIME the last

TizzySaurus:
Closes #1750 

The message sent when you create a new reminder now uses the DAY_TIME time formatting instead of RELATIVE time format.

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -270,7 +270,7 @@ class Reminders(Cog):
             }
         )
 
-        mention_string = f"Your reminder will arrive {discord_timestamp(expiration, TimestampFormats.RELATIVE)}"
+        mention_string = f"Your reminder will arrive {discord_timestamp(expiration, TimestampFormats.DAY_TIME)}"
 
         if mentions:
             mention_string += f" and will mention {len(mentions)} other(s)"
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -270,7 +270,7 @@ class Reminders(Cog):
             }
         )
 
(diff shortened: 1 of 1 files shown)


bot/exts/utils/reminders.py
```diff
@@ -270,7 +270,7 @@ async def new_reminder(
             }
         )
 
-        mention_string = f"Your reminder will arrive {discord_timestamp(expiration, TimestampFormats.RELATIVE)}"
+        mention_string = f"Your reminder will arrive {discord_timestamp(expiration, TimestampFormats.DAY_TIME)}"
```
Xithrius, line 273: I think `Your reminder will arrive on Thursday, January 1, 1970 1:00 AM` sounds better than `Your reminder will arrive Thursday, January 1, 1970 1:00 AM`

```suggestion
        mention_string = f"Your reminder will arrive on {discord_timestamp(expiration, TimestampFormats.DAY_TIME)}"
```

TizzySaurus: Done :+1:

Bluenix2 (approved):
Simple enough, looks good. Thank you

Xithrius (approved):

PR #1768, opened 2021-08-22
Update reminders command to use `UserMentionOrID` instead of `discord.Member`

### LINKED ISSUE #1767: Reminder command is too greedy
Currently the typehint for the `mentions` argument of the `!remind` command is `Greedy[Mentionable]`, where `Mentionable = t.Union[discord.Member, discord.Role]`.

The issue with have a greedy `discord.Member` is that cases like `!remind 3d fix this" ends up raising the error `fix is not a valid duration string.` since the `3d` gets matched to a member instead of the duration of 3 days.

My proposed solution to this, is simply to instead do `Greedy[t.Union[UserMentionOrID, discord.Role]]`.

There was discussion as to whether we should remove the `discord.Role` part/replace with a `RoleMentionorID` but was decided that the latter was pointless (since invocation would have to be `!remind {role} {duration} {msg}` for a greedy-role match -- people won't put role before duration unless wanting to ping) and that we should keep the `discord.Role` functionality since it works.

TizzySaurus:
Closes #1767.

Makes it so that user mentions for the reminder command have to be mention/id to prevent greediness issues with the `discord.Member` converter.

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -12,7 +12,7 @@ from discord.ext.commands import Cog, Context, Greedy, group
 
 from bot.bot import Bot
 from bot.constants import Guild, Icons, MODERATION_ROLES, POSITIVE_REPLIES, Roles, STAFF_ROLES
-from bot.converters import Duration
+from bot.converters import Duration, UserMentionOrID
 from bot.pagination import LinePaginator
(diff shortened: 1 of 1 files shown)


Bluenix2 (approved):
Looks good, thanks

Xithrius (approved):
Poifect.

PR #1770, opened 2021-08-23
Ignore bot mentions in antispam mentions rule

wookie184:
People may for example list all the bots on this server, this is not a potentially malicious use of pings so there is no reason for it to count towards the threshold for the antispam ping rule.


Akarys42 (approved):

wookie184:
need to fix the tests before merging

mbaruh (approved):

PR #1772, opened 2021-08-23
Remove restriction on snowflake command

### LINKED ISSUE #1759: Remove restriction on snowflake command
Now that the output of the snowflake command is paginated, I see no reason to limit non-staff members to only invoke it with 1 snowflake at a time.

I think we should remove the staff role check in this command and allow anyone to run it with as many snowflakes as needed.

D0rs4n:
Closes #1759


wookie184 (approved):
lgtm

TizzySaurus (approved):
Simple enough. Should work perfectly :+1:

Bluenix2 (approved):
Looks good, thank you

mbaruh (approved):

PR #1776, opened 2021-08-24
Give partners and community members more access

### LINKED ISSUE #1758: Python community & partners get same access as staff
I'd like to propose that everywhere in the bot (barring talent pool) that uses `STAFF_ROLES` as a check is changed to also include both python community and partners as passing that check.

These places would be:
- Duck pond (for seeing if a message was from a staff member)
- Help command (Posts output in channel, rather then being relayed to bot-commands)
- Role list (a command to list all roles)
- Role info (a command to get info on a given role)
- User info (Posts output in channel, rather then being relayed to bot-commands)
- Raw command (Bypasses cooldown, and allows them to run it in any channel)
- List stream (Ignores them in output of list stream, as they have explicit stream permissions like staff anyway
- Ping (Allows them to run it in any channel)
- Reminder (Allow them to run it in any channel)
- Charinfo (Allow them to run it in any channel)

TizzySaurus:
Closes #1758.

Partners and community members can now use commands in any channel as well as bypass cooldowns and a few other restrictions, such as the pings in `!remind` command.


Akarys42 (approved):
Minus the failed lint (you should really set up pre-commit, like, right now), this looks good to me. Thanks!

ChrisLovering (approved):

PR #1777, opened 2021-08-24
Fix antispam looking at wrong messages

mbaruh:


ChrisLovering (approved):

MarkKoz:
This seems rushed. Can you explain the differences between the sets of messages and the reason for this change?

mbaruh:
@MarkKoz I erroneously used the wrong set of messages in the cross-channel antispam PR. The one I used is all messages that the predicate looks at to see if there was a violation. The one that needs to be used is the one I use here which is the actual messages that are part of the offense.

mbaruh:
To put it simply, currently if there's a violation it deletes unrelated messages.

janine9vn (approved):
Didn't do a full test but after looking at this and discussing with you this should now have the intended behavior.

MarkKoz:
Okay this code is confusing because it reassigned `relevant_messages` to be a different set of messages... I was staring at it for 5 minute and this change seemed wrong until I caught that important detail.

MarkKoz:
It doesn't help that this comment is inaccurate since it actually returns 3 items, not 2 

>  If it doesn't, it returns a tuple in the form `(str, Iterable[discord.Member])`

mbaruh:
Yeah it took me a bit to understand what's going on there. But the key point is that the messages that are logged were already `relevant_messages` (line 211), and those were the correct messages, which is probably why it wasn't caught in testing the original PR.

PR #1778, opened 2021-08-25
Add time units and example to docstring of `!remind` and `!remind new`

### LINKED ISSUE #1655: add time formats tooltip to the !remind command
https://imgur.com/a/tYsTD7S
this suggestion is for adding documentation to the tooltip 
suggesting to have a tooltip on which time format is accepted when setting a reminder . 

started with doing` !remind new 12 api ` and it didn't accept the format. (i was guessing what this would do, because otherwise i would check the code on the github to find the formats) , and then i assumed that a `h` format could be used and it worked . 

note:
to quickly get more information in tooltips while on the discord i used `!remind new `and `!remind add` and they also don't provide which formats are accepted

TizzySaurus:
Closes #1655 

As the title says.

Now when the help message for `!remind` / `!remind new` is displayed, it'll also show the supported time units with an example invocation.

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -217,7 +217,20 @@ class Reminders(Cog):
     async def remind_group(
         self, ctx: Context, mentions: Greedy[ReminderMention], expiration: Duration, *, content: str
     ) -> None:
-        """Commands for managing your reminders."""
+        """
+        Commands for managing your reminders.
+
+        The `expiration` duration of `!remind new` supports the following symbols for each unit of time:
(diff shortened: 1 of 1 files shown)


wookie184 (changes requested):
The `edit_reminder_duration` docstring should also be changed to include the new time units.

TizzySaurus:
> The `edit_reminder_duration` docstring should also be changed to include the new time units.

Good catch. Also included it in the `edit_reminder_group` to keep the behaviour consistent with `remind_group`.

PR #1779, opened 2021-08-25
CI: check for dependency licenses

Akarys42:
Since our project is licensed under the MIT License, we can't be using any dependencies in our project. 

This PR adds a step to the CI lint phase that will use pip-licenses to verify that all the installed packages are part of an ALLOWED_LICENSE variable. This variable is currently set to be every license we currently use. 

We opted to use an allowlist instead of a denylist to make sure that new licenses are reviewed by hand and added to that variable.


Akarys42:
> How was it determined which licences are compatible with MIT?

I've looked at choose-a-license.org to see what their requirements are, and they all seemed to be okay with being used with MIT. All of those are very permissive.

*Note that I'm not a lawyer and this is not a legal advice, you know the drill*

MarkKoz:
I looked at [LGPL](https://www.gnu.org/licenses/lgpl-3.0.en.html) and even if it might be compatible with MIT, it does have some extra requirements. See section 4, which states that copies of licence must be included among other things.

Akarys42:
@MarkKoz my understanding here is we are only a consumer of the original library and never modify it, meaning it isn't a derived work and aren't subject to those rules. Although I'm honestly not sure, should we ask anyone about this?

MarkKoz:
> 
> 
> @MarkKoz my understanding here is we are only a consumer of the original library and never modify it, meaning it isn't a derived work and aren't subject to those rules. Although I'm honestly not sure, should we ask anyone about this?

It is a combined work. What I've heard is importing a library is considered to be linking to it. So, see the licence's definition of a combined work:

> A “Combined Work” is a work produced by combining or linking an Application with the Library. The particular version of the Library with which the Combined Work was made is also called the “Linked Version”.

onerandomusername:
As this project uses pre-commit, I feel like it would make sense to add this as a pre-commit hook, using the local configuration options, and setting it to only run on edits to pyproject.toml/poetry.lock. This would help ensure that regressions are discovered before they're ever commited.

onerandomusername:
I must be doing something wrong, as I tried this locally and added fuzzywuzzy as a dependency and this did not flag it at all.

Akarys42:
@MarkKoz alright, so we have two packages that uses LGPL. One of them is flake8-import-order which is quite easy to swap for Isort and would be quite an improvement. 

The other is chardet, a dependency of aiohttp. It is a bit more problematic, but we could dodge it using the accelator lib instead cChardet, which is under the Mozilla license. 

Maybe we could explore that if we want to get rid of them? Another solution would be to simply include a `LICENSE-THIRD-PARTY` file. I honestly surprised LGPL would require that, I don't see it being mentioned anywhere on the Internet, neither do I see any project which include the license. Maybe people are just not following the license.


@onerandomusername we opted to run it only in the CI to facilitate things, as running custom hooks on both Windows and Linux can be a pain. I don't think it matters when it is detected, as the work will have to be re-done either way. Also I don't know how you managed to not make it work, here is an (accidental) working example: https://github.com/python-discord/bot/pull/1779/commits/384d5b4b6b945e1aac3e1dbf82c1faa81cebbcf4. Updating `pip-licenses` added a new dep that weren't part of our set of allowed licenses and the CI correctly rejected it.

ChrisLovering (approved):
This PR is good to catch the licenses that are obviously against our MIT license.

I think we should create an issue to discuss what to do about the grey area licenses, rather than blocking this feature.

PR #1787, opened 2021-08-27
Add support for searching infractions by infraction id

### LINKED ISSUE #1066: Add Search Infraction by ID functionality
Currently, we can only look for an infraction either by the username/id of the person who got the infraction or by the content of the reason.  There is no way to search for an infraction or pull up a specific up by the ID number alone.  

This will likely need tweaking on the site repo as well since the search functionality requires an API call.

TizzySaurus:
Closes #1066.

Can now search infractions by infraction id via `!infraction {id}`, `!infraction search {id}` and `!infraction search id {id}`.

Exact implementation may be tweaked in the future, but got approval from @mbaruh to stick to this for now.


bot/exts/moderation/infraction/management.py
```diff
@@ -233,6 +238,16 @@ async def search_user(self, ctx: Context, user: t.Union[MemberOrUser, discord.Ob
         )
         await self.send_infraction_list(ctx, embed, infraction_list)
 
+    @infraction_search_group.command(name="id", aliases=("i", "infrid", "infr_id"))
+    async def search_id(self, ctx: Context, infr_id: int) -> None:
+        """Search for infraction by the infraction id."""
```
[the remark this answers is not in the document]
TizzySaurus: Right, yes. I will make these two changes shortly.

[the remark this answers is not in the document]
TizzySaurus: [Committed](https://github.com/python-discord/bot/pull/1787/commits/929556f79b2afbe7d5b32fd41e6c54daf2e37ad9) :+1:

bot/exts/moderation/infraction/management.py
```diff
@@ -44,8 +44,10 @@ def infractions_cog(self) -> Infractions:
     # region: Edit infraction commands
 
     @commands.group(name='infraction', aliases=('infr', 'infractions', 'inf', 'i'), invoke_without_command=True)
-    async def infraction_group(self, ctx: Context) -> None:
-        """Infraction manipulation commands."""
+    async def infraction_group(self, ctx: Context, infr_id: int = None) -> None:
```
Bluenix2, line 47: Do we want to make this the correct typing of `t.Optional[int]`?

TizzySaurus: My understanding is that you shouldn't use `Optional` in d.py when it's the last argument. That's what I've been told in the dpy server at least

Bluenix2: Ah right, forgot that discord.py reads these typings. That's good then

mbaruh (approved):
After further consideration of the UX and future development, I think the command should stay at the top level without the `infraction search` and `infraction search id` part.

Sorry Tizzy :pray:

bot/exts/moderation/infraction/management.py
onerandomusername: see here: 
https://discordpy.readthedocs.io/en/stable/ext/commands/commands.html?highlight=optional#typing-optional

mbaruh: That doesn't explain why it can't be the last argument.

Bluenix2: It simply isn't meant for the last argument, but I think it would work. That said though, what will now happen is that if the `IntConverter` fails instead of a converter error we will continue with `infr_id` set to None (as if it wasn't passed).

Is this acceptable?

mbaruh: I see. I suppose that's fine as is then.

bot/exts/moderation/infraction/management.py
```diff
@@ -44,9 +45,25 @@ def infractions_cog(self) -> Infractions:
...
+        """Infraction manipulation commands. If `infr_id` is passed then this command fetches that infraction."""
+        if not infr_id:
+            await ctx.send_help(ctx.command)
+            return
+
+        try:
+            infraction_list = [await self.bot.api_client.get(f"bot/infractions/{infr_id}/expanded")]
+        except ResponseCodeError as e:
+            if e.status == 404:
+                await ctx.send(":x: No matching infraction id could be found")
```
mbaruh, line 58:
```suggestion
                await ctx.send(f":x: No infraction with ID `{infr_id}` could be found.")
```

TizzySaurus: Done :+1:

bot/exts/moderation/infraction/management.py
```diff
@@ -44,9 +45,25 @@ def infractions_cog(self) -> Infractions:
     # region: Edit infraction commands
 
     @commands.group(name='infraction', aliases=('infr', 'infractions', 'inf', 'i'), invoke_without_command=True)
-    async def infraction_group(self, ctx: Context) -> None:
-        """Infraction manipulation commands."""
-        await ctx.send_help(ctx.command)
+    async def infraction_group(self, ctx: Context, infr_id: int = None) -> None:
+        """Infraction manipulation commands. If `infr_id` is passed then this command fetches that infraction."""
+        if not infr_id:
```
mbaruh, line 50: A minor nitpick: this will enter the condition with infr_id == 0. It's not actually important, but more correct to check explicitly for None.

TizzySaurus: True. An infraction id will never be 0 since the count starts at 1, but will make this change nonetheless :+1:

TizzySaurus: Done :+1:

mbaruh (approved):
Amazing 👌

Issue #1530, opened 2021-04-17
New Filters System Specifications


mbaruh:
This issue documents the planned changes to the filters extension. This needs to be done alongside python-discord/site#479.
Minor details may still change. This issue specifies the full extent of the desired changes, but this can be split into several issues and made into a project.

The changes allow for granular control over the behavior of individual filters, and increase the scope of it. All aspects of the extension, including automatic rules, will be controlled from a single entry-point.

# Filter Settings
Each added filter has settings which dictate what happens if a given message matches it. The settings are:
- Does it ping?
  - Which roles? everyone/here/mods/something else
- Does it delete?
- Does it filter reactions?
- Does it filter usernames and nicknames?
- Does it catch DM's?
- Does it ping for DM's?
- Does it DM the user about the rule? (similarly to what happens for invite links / everyone pings)
  - A secondary configuration for the message sent
- Do the filters apply only in specific channels / categories
- Which roles can bypass the filter?
- Does it apply an infraction? warn / mute / ban
- For domains: be able to choose whether it should be exact, or apply for subdomains as well (this should close #1276)

Each filter additionally has a description, as they already do.

Each list type has default settings, while each individual filter can override any setting.

# Filter Types
The current filters are split between four types: guild_invite, file_format, domain_name, filter_token. Each of those is split into a blacklist and a whitelist.

## Renaming
The names should be simpler and more memorable: invites, formats, domains, tokens.

## Whitelisting and Blacklisting
Not every filter type needs to have both a whitelist and a blacklist. Each filter type has its own method of deciding whether action should be taken, taking the whitelist, the blacklist, or both into account.

A blacklist or a whitelist should be created before anything can be added to it.

## Invites
The filters should save the invite which was used to add a guild to the list, unlike the current state in which only the ID is saved.

## Domains
Domains should support path patterns, e.g `*.pythondiscord.com/*/ric*rolls`.

## New Type: Custom Functions
Each filter type has its own method of checking whether a specific filter should be applied. There should be a list which specifies special filters, each with their own way of deciding whether they apply to a given message. Examples of such filters: the token remover, webhook remover, and any of the antispam rules. When an item can and should be added to this list will be clearer in the `Implementation Details` section.

# UI And Utilities
## Search
There should be an option to present all filters in a specific list which match a given input.
For example, if we have the tokens `a\db` and `123`, and we call `!bl search tokens a4b`, we will be presented with a list which contains the filter `a\db`. The listed filters should show each filter's ID (explained below).

[...]

PR #1788, opened 2021-08-28
Only check URL-like objects against domain filters

ChrisLovering:
Previously a message such as 'https://google.com hello! flask.request.method' would be
filtered due to us filtering the url shortener t.me. This commit changes to logic so
that we only check parts of the messages that matched the URL regex against our
blacklist, to avoid these false-positives.


Xithrius (approved):

bot/exts/filters/filtering.py
```diff
@@ -478,15 +478,16 @@ async def _has_urls(self, text: str) -> Tuple[bool, Optional[str]]:
         Second return value is a reason of URL blacklisting (can be None).
         """
         text = self.clean_input(text)
-        if not URL_RE.search(text):
+        matches = URL_RE.findall(text)
+        if not matches:
```
decorator-factory, line 482: You could simplify the code a bit like this:
```py
        domain_blacklist = self._get_filterlist_items("domain_name", allowed=False)
        for match in URL_RE.finditer(text):
            for url in domain_blacklist:
                if url.lower() in match.lower():
                    return True, self._get_filterlist_value("domain_name", url, allowed=False)["comment"]
```
Then you don't need the `if` check and a new local variable.

decorator-factory (changes requested):

bot/exts/filters/filtering.py
decorator-factory: Btw, why return a flag and an `Optional[str]` when you can just return `Optional[str]`? That would also play better with type checkers.

ChrisLovering: Refactored that an pushed it, thanks :D

It seems all of the filters are expected to return a tuple like this. So changing that would require a larger change. As far as I know, #1530 plans to change how to do filtering quite a lot.

decorator-factory: oh i see

PR #1790, opened 2021-08-28
Prevent the usage of names and nicknames when issuing infractions.

Qwerty-133:
It is no more possible to pass arguments such as `Qwerty` to `ban`, `mute`, `kick` etc, as it's error-prone. Only mentions, IDs and usernames (`name#descrim`) are allowed.

This PR has the (positive) side-effect of making `infraction search` and `reminder` work with usernames instead of just mentions and IDs too.

### DIFF
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -495,22 +495,36 @@ class HushDurationConverter(Converter):
         return duration
 
 
-class UserMentionOrID(UserConverter):
-    """
-    Converts to a `discord.User`, but only if a mention or userID is provided.
+def _is_an_unambiguous_user_argument(argument: str) -> bool:
+    """Check if the provided argument is a user mention, user id, or username."""
+    has_id_or_mention = bool(IDConverter()._get_id_match(argument) or RE_USER_MENTION.match(argument))
 
-    Unlike the default `UserConverter`, it doesn't allow conversion from a name or name#descrim.
+    if not has_id_or_mention:
+        if argument[0] == '@':
+            argument = argument[1:]
+
+        # Check to see if the author passed a username (a discriminator exists)
+        if len(argument) > 5 and argument[-5] == '#':
+            return True
+
+    return has_id_or_mention
+
+
+class UnambiguousUser(UserConverter):
+    """
+    Converts to a `discord.User`, but only if a mention, userID or a username is provided.
+
+    Unlike the default `UserConverter`, it doesn't allow conversion from a name.
     This is useful in cases where that lookup strategy would lead to ambiguity.
     """
 
     async def convert(self, ctx: Context, argument: str) -> discord.User:
         """Convert the `arg` to a `discord.User`."""
-        match = self._get_id_match(argument) or RE_USER_MENTION.match(argument)
-
-        if match is not None:
+        if _is_an_unambiguous_user_argument(argument):
             return await super().convert(ctx, argument)
         else:
-            raise BadArgument(f"`{argument}` is not a User mention or a User ID.")
+            raise BadArgument(f"`{argument}` is not a User mention, a User ID or a Username in the format"
+                              " `name#discriminator`.")
 
 
 class Infraction(Converter):
@@ -557,7 +571,7 @@ if t.TYPE_CHECKING:
     OffTopicName = str  # noqa: F811
     ISODateTime = datetime  # noqa: F811
     HushDurationConverter = int  # noqa: F811
-    UserMentionOrID = discord.User  # noqa: F811
+    UnambiguousUser = discord.User  # noqa: F811
     Infraction = t.Optional[dict]  # noqa: F811
 
 Expiry = t.Union[Duration, ISODateTime]
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -11,7 +11,7 @@ import dateutil.tz
 import discord
 from aiohttp import ClientConnectorError
 from dateutil.relativedelta import relativedelta
-from discord.ext.commands import BadArgument, Bot, Context, Converter, IDConverter, UserConverter
+from discord.ext.commands import BadArgument, Bot, Context, Converter, IDConverter, MemberConverter, UserConverter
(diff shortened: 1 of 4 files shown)


bot/converters.py
```diff
@@ -495,22 +495,55 @@ async def convert(self, ctx: Context, argument: str) -> int:
...
+
+    if not has_id_or_mention:
+        if argument[0] == '@':
+            argument = argument[1:]
+
+        # Check to see if the author passed a username (a discriminator exists)
+        if len(argument) > 5 and argument[-5] == '#':
+            return True
+
+    return has_id_or_mention
```
Bluenix2, line 510:
```suggestion
    if has_id_or_mention:
        return has_id_or_mention
    
    if argument[0] == '@':
        argument = argument[1:]

    # Check to see if the author passed a username (a discriminator exists)
    if len(argument) > 5 and argument[-5] == '#':
        return True
```

bot/converters.py
```diff
@@ -495,22 +495,55 @@ async def convert(self, ctx: Context, argument: str) -> int:
...
+        else:
+            raise BadArgument(AMBIGUOUS_ARGUMENT_MSG.format(argument=argument))
+
+
+class UnambiguousMember(MemberConverter):
+    """
+    Converts to a `discord.Member`, but only if a mention, userID or a username is provided.
+
+    Unlike the default `MemberConverter`, it doesn't allow conversion from a name or nickname.
+    This is useful in cases where that lookup strategy would lead to ambiguity.
```
Bluenix2, line 538:
```suggestion
    This is useful in cases where that lookup strategy would lead to too much ambiguity.
```

Bluenix2 (commented):
Did you also adjust tests?

bot/converters.py
Qwerty-133: That wouldn't work as-is (returns None sometimes) but I could instead use this to make the flow better

```py
    has_id_or_mention = ...

    # Check to see if the author passed a username (a discriminator exists)
    argument = argument.remove_prefix('@')
    has_username = len(argument) > 5 and argument[-5] == '#'

    return has_id_or_mention or has_username
```

bot/converters.py
Qwerty-133: > Did you also adjust tests?

By adjusting tests I'm assuming you mean adding new tests for these converters since there aren't any? I didn't do so as UserMentionOrID didn't have any tests - on what basis do we decide to add new tests for something?

Qwerty-133:
> Did you also adjust tests?

By adjusting tests I'm assuming you mean adding new tests for these converters since there aren't any? I didn't do so as UserMentionOrID didn't have any tests - on what basis do we decide to add new tests for something?

Bluenix2:
> > Did you also adjust tests?
> 
> By adjusting tests I'm assuming you mean adding new tests for these converters since there aren't any? I didn't do so as UserMentionOrID didn't have any tests

Oh, in that case I am not sure

> on what basis do we decide to add new tests for something?

That's a question better answered by a core-dev.

HassanAbouelela:
There currently aren’t any specific guidelines on adding tests. If you want to add them, go for it, otherwise don’t worry about it.

bot/converters.py
Qwerty-133: Thanks for the suggestions, I've implemented them

Qwerty-133:
Thanks! I'm not too concerned about testing them so will leave that off for now.

bot/converters.py
```diff
@@ -495,22 +495,51 @@ async def convert(self, ctx: Context, argument: str) -> int:
...
+
+
+class UnambiguousUser(UserConverter):
     """
-    Converts to a `discord.User`, but only if a mention or userID is provided.
+    Converts to a `discord.User`, but only if a mention, userID or a username is provided.
 
-    Unlike the default `UserConverter`, it doesn't allow conversion from a name or name#descrim.
-    This is useful in cases where that lookup strategy would lead to ambiguity.
+    Unlike the default `UserConverter`, it doesn't allow conversion from a name.
```
Bluenix2, line 517: Can we switch the explanation from name vs. username to something else, for many it has the same meaning. Maybe make a note that the discriminator has to be included or similar.

bot/converters.py
```diff
@@ -495,22 +495,51 @@ async def convert(self, ctx: Context, argument: str) -> int:
...
+            return await super().convert(ctx, argument)
+        else:
+            raise BadArgument(AMBIGUOUS_ARGUMENT_MSG.format(argument=argument))
+
+
+class UnambiguousMember(MemberConverter):
+    """
+    Converts to a `discord.Member`, but only if a mention, userID or a username is provided.
+
+    Unlike the default `MemberConverter`, it doesn't allow conversion from a name or nickname.
```
Bluenix2, line 533: See comment above

bot/exts/moderation/infraction/infractions.py
```diff
@@ -10,7 +10,7 @@
 from bot import constants
 from bot.bot import Bot
 from bot.constants import Event
-from bot.converters import Duration, Expiry, MemberOrUser
+from bot.converters import Duration, Expiry, MemberOrUser, UnambiguousMemberOrUser
```
Bluenix2, line 13: Why is `MemberOrUser` still used for bans?
https://github.com/python-discord/bot/blob/2018cb6e5386ead7609b6dcb065abaf30d32fb5a/bot/exts/moderation/infraction/infractions.py#L329-L342
https://github.com/python-discord/bot/blob/2018cb6e5386ead7609b6dcb065abaf30d32fb5a/bot/exts/moderation/infraction/infractions.py#L388-L390

Qwerty-133: That's not the ban command, it's a helper function, here it's used for type-hinting and they'd both do the exact same thing.

bot/converters.py
Qwerty-133: Thanks! Made those changes.

Bluenix2 (approved):

ChrisLovering (approved):

PR #1791, opened 2021-08-29
Allow dmrelay to only be used in mod channels.

Qwerty-133:
DMs between bot and users is private information which mods shouldn't be able to relay in public channels.

### DIFF
--- a/bot/exts/moderation/dm_relay.py
+++ b/bot/exts/moderation/dm_relay.py
@@ -5,6 +5,7 @@ from discord.ext.commands import Cog, Context, command, has_any_role
 
 from bot.bot import Bot
 from bot.constants import Emojis, MODERATION_ROLES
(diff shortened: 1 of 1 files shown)


bot/exts/moderation/dm_relay.py
```diff
@@ -64,7 +65,8 @@ async def dmrelay(self, ctx: Context, user: discord.User, limit: int = 100) -> N
 
     async def cog_check(self, ctx: Context) -> bool:
         """Only allow moderators to invoke the commands in this cog."""
```
mbaruh, line 67: Should update the comment here.

Qwerty-133: Done, thanks!

Xithrius (approved):
Does the do.

ChrisLovering (approved):
Looks good 👍

PR #1792, opened 2021-08-29
Change how the `raw` command is invoked.

Qwerty-133:
```py
    async def raw(self, ctx: Context, *, message: Message, json: bool = False) -> None:
```

This is the current interface, which doesn't allow you to use `!raw msg-link foo` because of the existence of the `json` parameter which is used by `raw json`. The `json` parameter is also included in the help text by default which isn't ideal.

This PR abstracts `raw`s logic away into a helper function to get the better interface.

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -460,11 +460,12 @@ class Information(Cog):
         # remove trailing whitespace
         return out.rstrip()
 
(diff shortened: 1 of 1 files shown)


wookie184 (approved):
Looks good to me 👍

Xithrius (approved):
Perfect.

PR #1794, opened 2021-08-30
Add concurrency rules to all GitHub workflows

ChrisLovering:
This concurrency rule means that workflow actions will be cancelled if a new set of actions are triggered on the same ref. A ref is either a PR or a branch itself.

Doing this means that if someone pushes to a PR multiple times in quick succession, workflows won't get queued up. Instead, only the workflows from the most recent changes will run, as this is all we really care about anyway.

I see the benefits of this as twofold.
1. The author gets faster feedback about the most recent change pushed, rather than have to wait for all previous runs to completed
2. Other contributors don't need to wait for a queue of redundant workflows to finish.

These exact concurrency rules have been live in Sir Lancebot for a few weeks now, and see to be working nicely.


Xithrius (approved):
Tested on Sir-Lance for a while, works perfectly fine.

MarkKoz:
Will this put the main branch into its own concurrency group? Cause that may not be a good idea. The cancellation is premature — it can't actually know if the newer workflow will even succeed. Imagine two different features are pushed to main at around the same time. If the newer workflow cancels the older one, and then the newer one fails, then neither change will get deployed. The older workflow could have succeeded and at least that change could have been deployed rather than nothing at all. Is this a concern?

ChrisLovering:
Yea it does add main as it's own group. I think a deployment failing on push to main isn't too much of a concern, since the checks need to pass in the PR first anyway, so it would be very rare. It's also easily solvable by re-running the failed suite.

I think the change is a net positive as it means there are less suites in the queue, and leads to less re-deployments as you say.

MarkKoz (approved):
> since the checks need to pass in the PR first anyway

okay good point

jchristgit (approved):
Yup. Makes sense to me. Thank you!

PR #1795, opened 2021-08-30
Prevent repetitive stale inventory warnings from erroneously generated inventories

Numerlor:
Some symbols always generate stale inventory warnings because some of the symbols they list are not present on the linked page, presumably because the formatting of the rst file or some other part of the doc build process is wrong. (for example the `environment-variables` symbol on https://www.pygame.org/docs/ref/pygame.html)

This PR adds a counter that's kept in redis and is incremented every time a warning is attempted to be raised for an item, its expire is also set to 3 weeks in the future to clear the counter in case the docs only needed a refresh or were fixed by the maintainers.
If that counter has a value above 2, the stale warning is not sent.

### DIFF
--- a/bot/exts/info/doc/_batch_parser.py
+++ b/bot/exts/info/doc/_batch_parser.py
@@ -16,12 +16,15 @@ from bot.constants import Channels
 from bot.utils import scheduling
 from . import _cog, doc_cache
 from ._parsing import get_symbol_markdown
+from ._redis_cache import StaleItemCounter
 
 log = logging.getLogger(__name__)
 
 
 class StaleInventoryNotifier:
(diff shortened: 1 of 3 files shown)


Numerlor:
I've noticed we didn't raise for status when requesting the html of the page to parse which caused some invalid pages like 404s attempted to be parsed, I set it to true in 23a3e5e.
This should also prevent the cog from trying to parse invalid pages because of misconfigured inventories or completely wrong symbols. They won't hit the the stale warning handler now and will instead be logged as a normal network warning with a full traceback

wookie184 (approved):
I'm not a redis expert but this all looks fine and from my testing seems to work well. Thanks!

PR #1796, opened 2021-08-30
Make the base_url argument in doc set optional

Numerlor:
There have been instances of a doc inventory not working because of the base URL being  misconfigured, this PR makes the URL optional to prevent issues like this from occurring.

For every inventory currently added on the bot, the method of extracting the base URL from the inventory URL by removing the last path segment holds true (the only package where it wasn't true from a recent dump had an invalid base URL to begin with).

At first I thought of making the inventory URL optional and doing this the other way around, but the inventory URL goes through more thorough validation both by the bot and on the user and there is at least one instance (django using `/_objects/` instead `/objects.inv`) where the default objects path is changed and would have to be specified by the user.

should be merged with https://github.com/python-discord/site/pull/571


HassanAbouelela (approved):
Clean, works as expected.

ChrisLovering (approved):
Nice, looks good 👌

PR #1798, opened 2021-08-31
Prevent role hierarchy issues with infractions

### LINKED ISSUE #1780: Suppress exceptions for infraction attempts on mods+
Due to the role hierarchy, the bot cannot infract mods or higher. Currently, this error is caught and handled by the infraction system, but it still logs the exception, causing it to show up on Sentry. A discussion on Discord came to the conclusion that the exception should only be logged if the target user is lower than a mod. The idea is that mods+ is an expected failure, but anything else would be unexpected. The implementation would require hard-coding a condition that the user is not a mod or higher.

If you identify any other functionality that could experience a role hierarchy error, please comment.

Related: #905

TizzySaurus:
Closes #1780.

`log.warning` is no longer called when an infraction/pardon fails due to the target being a Moderator+

I *believe* this is what @MarkKoz was after when they created their issue.

Not sure whether we want to still ping `@Moderators` if the pardon fails for this reason or not, but have left the ping there for now (can remove if requested).


MarkKoz (changes requested):
Did you test this first? It won't work because you're comparing role objects against IDs.

Also, do you think it would be scope creep to create a utility function for this role checking, since a similar pattern is being used elsewhere in the codebase?

TizzySaurus:
After a long discussion on discord it was decided to simply just catch the error raised by using `!starify` on members above the bot in the role hierarchy.

A summary of reason is basically that in the other cases we want to know something went wrong, because there *shouldn't* be an error in any other infractions due to the user being staff (we can successfully ban/kick/mute/etc. staff, including Moderators+, it's only starify we can't as it changes nickname which is hierarchy dependant.)

TizzySaurus:
After many hours dabbling with the tests I figured out how to fix them. *Hopefully* this should all now be ready to merge, just needs the reviews :+1:

Bluenix2 (approved):

bot/exts/moderation/infraction/infractions.py
```diff
@@ -340,6 +344,10 @@ async def apply_ban(
 
         Will also remove the banned user from the Big Brother watch list if applicable.
         """
+        if hasattr(user, 'top_role') and user.top_role >= ctx.me.top_role:
```
jchristgit, line 347: in which cases do we have a `top_role` here? In case this is a member, right? Would it be cleaner to check for `isinstance` here then? That may help some typecheckers.

jchristgit (approved):
I like this. One minor nitpick. Feel free to ignore it completely.

bot/exts/moderation/infraction/infractions.py
TizzySaurus: Yeah, `top_role` is when it's a `discord.Member`. I'm not sure whether I did `hasattr` instead of `isinstance` for a reason or not. I'll check and see - if there's no real reason then `isinstanace` is probably much clearer.

TizzySaurus: Couldn't see a logical reason to stick to `hasattr`, so have pushed a commit changing to `isinstance` :+1:

Thanks for the review :+1:

jb3 (approved):

PR #1804, opened 2021-09-01
Add support for `!infraction last`

### LINKED ISSUE #1803: Add support for `!infraction last`
Extension of #1787.

Will allow the usage of `!infraction last` to view details of the user's most recent infraction made.

Requires changing `infr_id: int` to `infraction: Infraction` and making the `Infraction` converter use the `expanded` endpoint (which I've already checked to ensure won't break anything).

TizzySaurus:
Closes #1803.

Adds `!infraction last`, which will view details of the last infraction the user made. Also changes the `Infraction` converter to use the `expanded` endpoint

### DIFF
--- a/bot/errors.py
+++ b/bot/errors.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
-from typing import Hashable, TYPE_CHECKING
+from typing import Hashable, TYPE_CHECKING, Union
+
+from discord.ext.commands import ConversionError, Converter
+
 if TYPE_CHECKING:
     from bot.converters import MemberOrUser
 
@@ -38,6 +41,20 @@ class InvalidInfractedUserError(Exception):
         self.reason = reason
 
         super().__init__(reason)
+
+
+class InvalidInfraction(ConversionError):
+    """
+    Raised by the Infraction converter when trying to fetch an invalid infraction id.
+
+    Attributes:
+        `infraction_arg` -- the value that we attempted to convert into an Infraction
+    """
+
+    def __init__(self, converter: Converter, original: Exception, infraction_arg: Union[int, str]):
+
+        self.infraction_arg = infraction_arg
+        super().__init__(converter, original)
 
 
 class BrandingMisconfiguration(RuntimeError):
--- a/bot/exts/moderation/infraction/management.py
+++ b/bot/exts/moderation/infraction/management.py
@@ -13,6 +13,7 @@ from discord.utils import escape_markdown
 from bot import constants
 from bot.bot import Bot
 from bot.converters import Expiry, Infraction, MemberOrUser, Snowflake, UnambiguousUser, allowed_strings
+from bot.errors import InvalidInfraction
 from bot.exts.moderation.infraction.infractions import Infractions
 from bot.exts.moderation.modlog import ModLog
 from bot.pagination import LinePaginator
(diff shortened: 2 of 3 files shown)


bot/exts/moderation/infraction/management.py
```diff
@@ -45,25 +44,17 @@ def infractions_cog(self) -> Infractions:
     # region: Edit infraction commands
 
     @commands.group(name='infraction', aliases=('infr', 'infractions', 'inf', 'i'), invoke_without_command=True)
-    async def infraction_group(self, ctx: Context, infr_id: int = None) -> None:
-        """Infraction manipulation commands. If `infr_id` is passed then this command fetches that infraction."""
-        if infr_id is None:
+    async def infraction_group(self, ctx: Context, infraction: Infraction = None) -> None:
```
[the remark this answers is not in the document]
TizzySaurus: We actually don't want this because then the error raised by the `Infraction` converter would get supressed (which means the `cog_command_error` handling wouldn't work).

bot/exts/moderation/infraction/management.py
```diff
@@ -348,13 +339,17 @@ async def cog_check(self, ctx: Context) -> bool:
...
-        """Send a notification to the invoking context on a Union failure."""
+    async def cog_command_error(self, ctx: Context, error: commands.CommandError) -> None:
+        """Handles errors for commands within this cog."""
         if isinstance(error, commands.BadUnionArgument):
             if discord.User in error.converters:
                 await ctx.send(str(error.errors[0]))
                 error.handled = True
 
+        elif isinstance(error, commands.BadArgument):
+            await ctx.send(f":x: Could not find a corresponding infraction for `{error.args[1]}`.")
```
mbaruh, line 350: This assumes that BadArgument will only ever be raised from the infraction converter.

Perhaps this should be a custom error such as `InfractionConversionError` which will inherit from dpy's `ConversionError`

TizzySaurus: Yeah, I've made an `InvalidInfraction` error and decided to also slightly improve the error messages (invalid id and 404). Will push these changes shortly.

example usage

TizzySaurus: Also going to change "...infraction for `{number}`" to `"...infraction with id `{number}`

bot/errors.py
```diff
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
-from typing import Hashable, TYPE_CHECKING
+from typing import Hashable, TYPE_CHECKING, Union
```
Bluenix2, line 3: Doesn't `TYPE_CHECKING` come first?

bot/errors.py
```diff
@@ -40,6 +43,20 @@ def __init__(self, user: MemberOrUser, reason: str = "User infracted is a bot.")
         super().__init__(reason)
 
 
+class InvalidInfraction(ConversionError):
+    """
+    Raised by Infraction converter when fetching an invalid infraction id.
```
Bluenix2, line 48:
```suggestion
    Raised by the Infraction converter when trying to fetch an invalid infraction id.
```

bot/errors.py
TizzySaurus: I don't really understand what you're saying here.

The Pydis style guide states that for imports:
> Each group must be ordered alphabetically, with uppercase modules coming before lowercase.
and that's what I'm doing here? H < T < U so `Hashable, TYPE_CHECKING, Union`, as I did.

bot/errors.py
TizzySaurus: Will adjust this, thanks :+1:

TizzySaurus: Committed :+1:

Bluenix2 (commented):
Tested it simply, and works. Can we just get a comment that the Infraction converter special-cases `last` as getting the last converter from the author?

I thought `last` was a subcommand at first and was confused.

TizzySaurus:
> Tested it simply, and works. Can we just get a comment that the Infraction converter special-cases `last` as getting the last infraction from the author?
> 
> I thought `last` was a subcommand at first and was confused.

Done :+1:

Bluenix2 (approved):
Tested earlier today, works nicely. Thanks Tizzy

Xithrius (approved):

Xithrius:
I've tested this, good stuff. Thanks Tizzy!

PR #1806, opened 2021-09-02
Apply infractions before DMing

### LINKED ISSUE #1805: Apply infraction before DMing the user
The only case where we have to DM the user first is when we apply a ban. Otherwise, if applying the infraction fails for any reason, it will unnecessarily DM the user about it first.

TizzySaurus:
Closes #1805

Bot now attempts to infract before DMing, with the exception of kicks/bans, in order to prevent the user getting sent a message stating they were infracted when the infraction failed (and so they aren't).


bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -226,7 +223,13 @@ async def apply_infraction(
                 log.error(f"Deletion of {infr_type} infraction #{id_} failed with error code {e.status}.")
             infr_message = ""
         else:
+            log.info(f"Applied {purge}{infr_type} infraction #{id_} to {user}.")
             infr_message = f" **{purge}{' '.join(infr_type.split('_'))}** to {user.mention}{expiry_msg}{end_msg}"
+            if infr_type not in ("ban", "kick"):
```
Bluenix2, line 228:
```suggestion
            if infr_type not in {"ban", "kick"}:  # We have already notified the user for these types
```

TizzySaurus: Have committed the comment but leaving as a tuple since that's what's used elsewhere in the codebase.

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -226,7 +223,13 @@ async def apply_infraction(
                 log.error(f"Deletion of {infr_type} infraction #{id_} failed with error code {e.status}.")
             infr_message = ""
         else:
+            log.info(f"Applied {purge}{infr_type} infraction #{id_} to {user}.")
             infr_message = f" **{purge}{' '.join(infr_type.split('_'))}** to {user.mention}{expiry_msg}{end_msg}"
+            if infr_type not in ("ban", "kick"):  # if we haven't already tried to send the DM
```
Bluenix2, line 228:
```suggestion
            if infr_type not in ("ban", "kick"):  # If we haven't already tried to send the DM
```

Bluenix2 (commented):
Really couldn't find anything else, I just need to test this before approving.

bot/exts/moderation/infraction/_scheduler.py
TizzySaurus: From the looks of it we do generally have the first word capitalised so will commit this :+1:

Bluenix2 (approved):
Tested and tried to cause some error cases with missing permission, handles everything correctly.

jb3 (approved):

PR #1811, opened 2021-09-04
Pin platform in Dockerfile

ChrisLovering:
Some of our deps don't have wheels for atm processors.

With Mac's M1 chips becomming more common, we should make it easier for those users to build our environments.


Xithrius (approved):

jchristgit (approved):
Rock solid.

Thanks!

PR #1813, opened 2021-09-05
Declare TalentPool.cache on init of cog

### LINKED ISSUE #1812: AttributeError: 'TalentPool' object has no attribute 'cache'
Sentry Issue: [BOT-1MS](https://sentry.io/organizations/python-discord/issues/2626334845/?referrer=github_integration)

```
AttributeError: 'TalentPool' object has no attribute 'cache'
  File "bot/exts/info/information.py", line 219, in server_info
    embed.add_field(name="Moderation:", value=self.get_extended_server_info(ctx))
  File "bot/exts/info/information.py", line 75, in get_extended_server_info
    talentpool_info = f"Nominated: {len(cog.cache)}\n"

Error executing command invoked by Xithrius#1318: !server
```

ChrisLovering:
Closes #1812

This avoids issues in the server cog trying to access it before it's assigned and refreshed.

I also migrated to the tasks to `scheduling.create_task()` as the created tasks currently don't have any error handling they can hide errors in development until the task object is destroyed (if that occurs at all) which logs the exception. The scheduling alternative attaches a callback which logs exceptions to prevent this.

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -72,7 +72,8 @@ class Information(Cog):
         """Return additional server info only visible in moderation channels."""
         talentpool_info = ""
         if cog := self.bot.get_cog("Talentpool"):
-            talentpool_info = f"Nominated: {len(cog.cache)}\n"
+            num_nominated = len(cog.cache) if cog.cache else 0
+            talentpool_info = f"Nominated: {num_nominated}\n"
 
         bb_info = ""
         if cog := self.bot.get_cog("Big Brother"):
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -72,7 +72,7 @@ class Information(Cog):
         """Return additional server info only visible in moderation channels."""
         talentpool_info = ""
         if cog := self.bot.get_cog("Talentpool"):
-            num_nominated = len(cog.cache) if cog.cache else 0
+            num_nominated = len(cog.cache) if cog.cache else "-"
             talentpool_info = f"Nominated: {num_nominated}\n"
 
         bb_info = ""
--- a/bot/exts/recruitment/talentpool/_cog.py
+++ b/bot/exts/recruitment/talentpool/_cog.py
@@ -2,7 +2,7 @@ import logging
 import textwrap
 from collections import ChainMap, defaultdict
 from io import StringIO
-from typing import Union
+from typing import Optional, Union
 
 import discord
 from async_rediscache import RedisCache
@@ -15,7 +15,7 @@ from bot.constants import Channels, Emojis, Guild, MODERATION_ROLES, Roles, STAF
 from bot.converters import MemberOrUser
 from bot.exts.recruitment.talentpool._review import Reviewer
 from bot.pagination import LinePaginator
-from bot.utils import time
+from bot.utils import scheduling, time
 from bot.utils.time import get_time_delta
 
 AUTOREVIEW_ENABLED_KEY = "autoreview_enabled"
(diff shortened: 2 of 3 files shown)


Numerlor:
Wouldn't setting it to `None` to signal that the cache isn't available be clearer interface? Or would the 0 in the server embed be seen as that

ChrisLovering:
Hm, yea that might be better to do it that way. Will just need to add logic to the server command so it doesn't try to do `len(None)`

Xithrius (approved):
Works well.

wookie184:
Do we want to refresh the cache before checking the length in the server cog? If autoreview is off the cache will not be populated until the cog is interacted with in some way. We could also just populate the cache on loading the cog even if autoreview is off.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -34,12 +34,17 @@ class TalentPool(Cog, name="Talentpool"):
...
         self.api_default_params = {'active': 'true', 'ordering': '-inserted_at'}
-        self.bot.loop.create_task(self.schedule_autoreviews())
+
+        self.initial_refresh_task = scheduling.create_task(self.refresh_cache(), event_loop=self.bot.loop)
+        scheduling.create_task(self.schedule_autoreviews(), event_loop=self.bot.loop)
 
     async def schedule_autoreviews(self) -> None:
         """Reschedule reviews for active nominations if autoreview is enabled."""
         if await self.autoreview_enabled():
+            # Wait for a populated the cache first
```
mbaruh, line 46: Typo

bot/exts/info/information.py
```diff
@@ -72,7 +72,8 @@ def get_extended_server_info(self, ctx: Context) -> str:
         """Return additional server info only visible in moderation channels."""
         talentpool_info = ""
         if cog := self.bot.get_cog("Talentpool"):
-            talentpool_info = f"Nominated: {len(cog.cache)}\n"
+            num_nominated = len(cog.cache) if cog.cache else 0
```
mbaruh, line 75: As it stands if the cache is not ready it will say `0`. I'd rather it gave a non-numerical value like `-` to show that, since 0 nominees is possible.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -50,6 +55,8 @@ async def autoreview_enabled(self) -> bool:
 
     async def refresh_cache(self) -> bool:
         """Updates TalentPool users cache."""
+        # Wait until logged in to ensure bot api client exists
+        await self.bot.on_guild_available()
```
mbaruh, line 59: What is the difference between this and `wait_until_guild_available`?

mbaruh (changes requested):

bot/exts/recruitment/talentpool/_cog.py
ChrisLovering: The difference is that this is the wrong one :) `on_guild_available` is the event that sets the available flag. Nice catch :D

bot/exts/info/information.py
ChrisLovering: Very true, shall fix

ChrisLovering: https://github.com/python-discord/bot/pull/1813/commits/4e61411d2be105ad1e68fa6211d3c95fc3d92b3b

bot/exts/recruitment/talentpool/_cog.py
ChrisLovering: https://github.com/python-discord/bot/pull/1813/commits/4e61411d2be105ad1e68fa6211d3c95fc3d92b3b

bot/exts/recruitment/talentpool/_cog.py
ChrisLovering: https://github.com/python-discord/bot/pull/1813/commits/4e61411d2be105ad1e68fa6211d3c95fc3d92b3b

ChrisLovering:
Before merging, let me know and I'll squash the fixup commit(s)

mbaruh (approved):
Looks good. Do you think the scheduling approach should be used in the rest of the code-base as well? because we use this pattern quite a lot.

PR #1815, opened 2021-09-06
Migrate to Discord 2.0a0 

Akarys42:
Since the Discord.py repository has been archived, we can switch to the latest commit of 2.0a0, knowing no breaking change will occur (still pinned to the commit just in case).

This PR fixes any problem related to the migration:
- New avatar interface
- TZ aware datetimes
- Various internal API changes

The bot will also automatically join new threads and silence cannot be used in threads due to their lack of permissions.
This was a significantly harder migration than Sir Lance, but I have tested every cog individually, we should hopefully not run into any issue.

### DIFF
--- a/bot/exts/moderation/defcon.py
+++ b/bot/exts/moderation/defcon.py
@@ -137,7 +137,7 @@ class Defcon(Cog):
 
                 await self.mod_log.send_log_message(
                     Icons.defcon_denied, Colours.soft_red, "Entry denied",
-                    message, member.avatar_url_as(static_format="png")
+                    message, member.avatar.url
                 )
 
     @group(name='defcon', aliases=('dc',), invoke_without_command=True)
--- a/bot/exts/moderation/modlog.py
+++ b/bot/exts/moderation/modlog.py
@@ -394,7 +394,7 @@ class ModLog(Cog, name="ModLog"):
         await self.send_log_message(
             Icons.user_ban, Colours.soft_red,
             "User banned", format_user(member),
-            thumbnail=member.avatar_url_as(static_format="png"),
+            thumbnail=member.avatar.url,
             channel_id=Channels.user_log
         )
 
@@ -415,7 +415,7 @@ class ModLog(Cog, name="ModLog"):
         await self.send_log_message(
             Icons.sign_in, Colours.soft_green,
             "User joined", message,
-            thumbnail=member.avatar_url_as(static_format="png"),
+            thumbnail=member.avatar.url,
             channel_id=Channels.user_log
         )
 
@@ -432,7 +432,7 @@ class ModLog(Cog, name="ModLog"):
         await self.send_log_message(
             Icons.sign_out, Colours.soft_red,
             "User left", format_user(member),
-            thumbnail=member.avatar_url_as(static_format="png"),
+            thumbnail=member.avatar.url,
             channel_id=Channels.user_log
         )
 
@@ -449,7 +449,7 @@ class ModLog(Cog, name="ModLog"):
         await self.send_log_message(
             Icons.user_unban, Colour.blurple(),
             "User unbanned", format_user(member),
-            thumbnail=member.avatar_url_as(static_format="png"),
+            thumbnail=member.avatar.url,
             channel_id=Channels.mod_log
         )
 
@@ -515,7 +515,7 @@ class ModLog(Cog, name="ModLog"):
             colour=Colour.blurple(),
             title="Member updated",
             text=message,
-            thumbnail=after.avatar_url_as(static_format="png"),
+            thumbnail=after.avatar.url,
             channel_id=Channels.user_log
         )
 
@@ -820,7 +820,7 @@ class ModLog(Cog, name="ModLog"):
             colour=colour,
             title="Voice state updated",
             text=message,
-            thumbnail=member.avatar_url_as(static_format="png"),
+            thumbnail=member.avatar.url,
             channel_id=Channels.voice_log
         )
 
--- a/bot/exts/moderation/silence.py
+++ b/bot/exts/moderation/silence.py
@@ -6,7 +6,7 @@ from datetime import datetime, timedelta, timezone
 from typing import Optional, OrderedDict, Union
 
 from async_rediscache import RedisCache
-from discord import Guild, PermissionOverwrite, TextChannel, VoiceChannel
+from discord import Guild, PermissionOverwrite, TextChannel, Thread, VoiceChannel
 from discord.ext import commands, tasks
 from discord.ext.commands import Context
 from discord.utils import MISSING
@@ -181,6 +181,12 @@ class Silence(commands.Cog):
 
         channel_info = f"#{channel} ({channel.id})"
         log.debug(f"{ctx.author} is silencing channel {channel_info}.")
+
+        # Since threads don't have specific overrides, we cannot silence them individually.
+        # The parent channel has to be muted or the thread should be archived.
+        if isinstance(channel, Thread):
+            await ctx.send(":x: Threads cannot be silenced.")
+            return
 
         if not await self._set_silence_overwrites(channel, kick=kick):
             log.info(f"Tried to silence channel {channel_info} but the channel was already silenced.")
--- a/bot/exts/moderation/silence.py
+++ b/bot/exts/moderation/silence.py
@@ -9,6 +9,7 @@ from async_rediscache import RedisCache
 from discord import Guild, PermissionOverwrite, TextChannel, VoiceChannel
 from discord.ext import commands, tasks
 from discord.ext.commands import Context
+from discord.utils import MISSING
 
 from bot import constants
 from bot.bot import Bot
@@ -47,7 +48,16 @@ class SilenceNotifier(tasks.Loop):
     """Loop notifier for posting notices to `alert_channel` containing added channels."""
 
     def __init__(self, alert_channel: TextChannel):
-        super().__init__(self._notifier, seconds=1, minutes=0, hours=0, count=None, reconnect=True, loop=None)
+        super().__init__(
+            self._notifier,
+            seconds=1,
+            minutes=0,
+            hours=0,
+            count=None,
+            reconnect=True,
+            loop=None,
+            time=MISSING
+        )
         self._silenced_channels = {}
         self._alert_channel = alert_channel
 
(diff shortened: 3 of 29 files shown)


Kronifer (approved):
Looks good!

bot/exts/moderation/silence.py
```diff
@@ -47,7 +48,16 @@ class SilenceNotifier(tasks.Loop):
...
-        super().__init__(self._notifier, seconds=1, minutes=0, hours=0, count=None, reconnect=True, loop=None)
+        super().__init__(
+            self._notifier,
+            seconds=1,
+            minutes=0,
+            hours=0,
+            count=None,
+            reconnect=True,
+            loop=None,
+            time=MISSING
```
Bluenix2, line 59: Huh? Shouldn't `MISSING` be the keyword argument's default already?

bot/exts/utils/bot.py
```diff
@@ -16,6 +17,21 @@ class BotCog(Cog, name="Bot"):
...
+    @Cog.listener()
+    async def on_thread_join(self, thread: Thread) -> None:
+        """
+        Try to join newly created threads.
+
+        Despite the event name being misleading, this is dispatched when new threads are created.
+        We want our bots to automatically join threads in order to answer commands using their prefixes.
+        """
+        if thread.me:
+            # Already in this thread, return early
```
Bluenix2, line 29:
```suggestion
            # We have already joined this thread
```

bot/exts/utils/bot.py
```diff
@@ -16,6 +17,21 @@ class BotCog(Cog, name="Bot"):
     def __init__(self, bot: Bot):
         self.bot = bot
 
+    @Cog.listener()
+    async def on_thread_join(self, thread: Thread) -> None:
+        """
+        Try to join newly created threads.
+
+        Despite the event name being misleading, this is dispatched when new threads are created.
+        We want our bots to automatically join threads in order to answer commands using their prefixes.
```
Bluenix2, line 26: Do know that bots get messages even if they don't join threads

Bluenix2 (commented):
All I could find from the diff, but I'll have to check _what you didn't change_ as well to make sure we do not miss something.

bot/exts/moderation/defcon.py
```diff
@@ -196,7 +203,14 @@ async def unshutdown(self, ctx: Context) -> None:
         role = ctx.guild.default_role
         permissions = role.permissions
 
-        permissions.update(send_messages=True, add_reactions=True, connect=True)
+        permissions.update(
+            send_messages=True,
+            add_reactions=True,
+            create_private_threads=None,
+            create_public_threads=None,
```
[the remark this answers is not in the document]
mbaruh: Yeah I don't think we're planning on enabling it globally atm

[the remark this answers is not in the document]
mbaruh: But considering that, I'm not sure these commands should do anything with thread permissions at all?

bot/exts/moderation/modlog.py
```diff
@@ -515,21 +515,48 @@ async def on_member_update(self, before: discord.Member, after: discord.Member)
...
+
+        # Ignore bots
+        if message.author.bot:
+            return True
+
+        # Look at the parent channel of a thread
+        if isinstance(message.channel, Thread):
+            return message.channel.parent.id in GuildConstant.modlog_blacklist
+
+        return message.channel.id in GuildConstant.modlog_blacklist
```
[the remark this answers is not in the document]
mbaruh: If ` `message.guild` is None then `message.guild.id` will error, but otherwise yeah this should work.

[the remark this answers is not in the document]
mbaruh: The raw version also doesn't check the author.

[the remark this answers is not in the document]
Bluenix2: Ah, that is a good point, so we can change the non-raw one to do so:
```suggestion
    def is_message_blacklisted(self, message: Message) -> bool:
        """Return true if the message is in a blacklisted thread or channel."""
        # Ignore bots or Ms
        if message.author.bot or not message.guild:
            return True

        return self.is_raw_message_blacklisted(message.guild.id, message.channel.id)
```

bot/exts/moderation/defcon.py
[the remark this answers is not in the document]
Bluenix2: It should, you can open threads on messages even if you don't have send message permissions.

And if you have the "Send messages in threads" then you can send messages in that thread.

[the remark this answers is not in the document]
mbaruh: Yes, "Send messages in threads" should be changed, but these two are off by default.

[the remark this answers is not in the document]
Akarys42: I believe that's two perms we want to enable per channel, not globally

bot/exts/moderation/silence.py
Akarys42: > Huh?
That was also my reaction :D but apparently due to how the API changed you have to set it explicitly

bot/exts/moderation/modlog.py
```diff
@@ -751,6 +759,64 @@ async def on_raw_message_edit(self, event: discord.RawMessageUpdateEvent) -> Non
...
+            action = "un-archived"
+            icon = Icons.hash_green
+        else:
+            return
+
+        await self.send_log_message(
+            icon,
+            colour,
+            f"Thread {action}",
+            f"Thread {after.mention} (`{after.id}`) from {after.parent.mention} (`{after.parent.id}`) {action}"
```
Bluenix2, line 792:
```suggestion
            f"Thread {after.mention} (`{after.id}`) from {after.parent.mention} (`{after.parent.id}`) was {action}"
```

bot/exts/moderation/modlog.py
```diff
@@ -751,6 +759,64 @@ async def on_raw_message_edit(self, event: discord.RawMessageUpdateEvent) -> Non
...
+            f"Thread {thread.mention} (`{thread.id}`) from {thread.parent.mention} (`{thread.parent.id}`) deleted"
+        )
+
+    @Cog.listener()
+    async def on_thread_join(self, thread: Thread) -> None:
+        """Log thread creation."""
+        # If we are in the thread already we can most probably assume we already logged it?
+        # We don't really have a better way of doing this since the API doesn't make any difference between the two
+        if thread.me:
+            return
```
Bluenix2, line 811: image

So this is sent when a thread is created, or we get access to a private thread by being mentioned in it.

We don't need this, when this event is received it should be from a thread being created no?

bot/exts/utils/bot.py
```diff
@@ -16,6 +17,20 @@ class BotCog(Cog, name="Bot"):
...
+        Try to join newly created threads.
+
+        Despite the event name being misleading, this is dispatched when new threads are created.
+        """
+        if thread.me:
+            # We have already joined this thread
+            return
+
+        with suppress(Forbidden):
+            await thread.join()
```
Bluenix2, line 32: Is this because we want the bot to appear in the right side?

Akarys42: Yep! It first was to make it listen to messages, but we deemed that as unnecessary. I still find it nice to have the bots in the member list.

bot/exts/moderation/modlog.py
Akarys42: I don't think you are looking at the right event. The API event for [`on_thread_join`](https://discordpy.readthedocs.io/en/master/api.html#discord.on_thread_join) seems to be [`Thread Create`](https://discord.com/developers/docs/topics/gateway#thread-create), which fires both when a new thread becomes accessible by the bot and when the bot actually joins it.

Without this check, we would log the thread twice, once when it is created and the even first fires, and a second time when it actually joins the thread. That's what I could also check during my testings.

ChrisLovering: It also fites when _anyone_ joins a thread.

Bluenix2: Ahhh, I took a screenshot of the right event at first but then mixed it up.

Yes, Thread Update is the one that is sent when people join. But [Thread Create](https://discord.com/developers/docs/topics/gateway#thread-create) is only sent when a thread is made available to the bot.

Akarys42: yes, which I believe is the one used by `on_thread_join`, since its behavior matches that. So we do need to have this check.

Bluenix2: I don't understand, I am still not convinced. Thread Update is the one that gets sent when a thread gets updated, which it does when members join because now the list of joined members change.

Bluenix2 (changes requested):
`discord.User.avatar` is actually `Optional`, you need to change it to `discord.User.display_avatar`.

bot/exts/moderation/defcon.py
mbaruh, line 210: Following the discussion in https://github.com/python-discord/bot/pull/1815#discussion_r711755032 I think these two lines and the matching lines in `shutdown` should be removed. I don't think there even is a None option for global permissions.

mbaruh (changes requested):

bot/exts/moderation/modlog.py
Akarys42: Indeed, but I am not sure how that's relevant? I could confirm through testing that's the behaviour we want

Bluenix2:
I found some datetimes where the timezones haven't been replaced, should they be? See the following Sourcegraph searches:
- [Search for `created_at`](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/python-discord/bot%24%40discord-2.0+created_at&patternType=literal)
- [Search for `edited_at`](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/python-discord/bot%24%40discord-2.0+edited_at&patternType=literal)

Akarys42:
> I found some datetimes where the timezones haven't been replaced, should they be? See the following Sourcegraph searches:
> 
> * [Search for `created_at`](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/python-discord/bot%24%40discord-2.0+created_at&patternType=literal)
> * [Search for `edited_at`](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/python-discord/bot%24%40discord-2.0+edited_at&patternType=literal)

Went over those, they should (hopefully) not cause any issue

PR #1817, opened 2021-09-08
Regex: add a word boundary before .gg

Akarys42:
Before this commit, `an-arbitrary-domain.gg/notaninvite` would trigger the filter. This solve the issue by adding a word boundary before this branch of the pattern.

### DIFF
--- a/bot/utils/regex.py
+++ b/bot/utils/regex.py
@@ -7,7 +7,7 @@ INVITE_RE = re.compile(
     r"discord(?:[\.,]|dot)me|"                        # or discord.me
     r"discord(?:[\.,]|dot)li|"                        # or discord.li
     r"discord(?:[\.,]|dot)io|"                        # or discord.io.
-    r"(?:[\.,]|dot)gg"                                # or .gg/
+    r"(?:\b([\.,]|dot))gg"                                # or .gg/
     r")(?:[\/]|slash)"                                # / or 'slash'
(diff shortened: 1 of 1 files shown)


bot/utils/regex.py
```diff
@@ -7,7 +7,7 @@
     r"discord(?:[\.,]|dot)me|"                        # or discord.me
     r"discord(?:[\.,]|dot)li|"                        # or discord.li
     r"discord(?:[\.,]|dot)io|"                        # or discord.io.
-    r"(?:[\.,]|dot)gg"                                # or .gg/
+    r"(?:\b([\.,]|dot))gg"                                # or .gg/
```
ChrisLovering, line 10:
```suggestion
    r"(?:\b(?:[\.,]|dot))gg"                                # or .gg/
```
We want this to be a non-capturing group, so that it doesn't affect us extracting the actual invite code.

Also, is this trying to enforce that there is whitespace before the `.gg/`? If so, don't we need to use `\B`?

ChrisLovering (changes requested):

bot/utils/regex.py
mbaruh: `\b` is a word boundary, `\B` is the opposite.

mbaruh: The reason this isn't working is because `.` is not a word character, so it makes no sense to use `\b` here. It's possible to instead have a negative lookbehind for `\w`

PR #1818, opened 2021-09-08
Normal type hints. Delete shit-code

SayHelloRoman:


jb3:
No-return is used when the function will never return (e.g. an infinite loop), typehints should be for success cases, you do not generally typehint exceptions.

This should be raised either as an issue or in the #dev-contrib channel where a core developer can approve it. I'm not a huge fan of "shit code" as a PR title and an empty PR body, it's not in line with our contribution guidelines.

lightmanLP:
kinda comedy.

PR #1819, opened 2021-09-09
Add string-formatting tag

janine9vn:
Adds a tag to show the string formatting mini language.

To keep this tag on the shorter side, I'm aiming for a "look at how cool this is!" vs a "let me explain how this works."
So the examples aren't explained in depth, but hopefully it's enticing enough to click on the links and provide an entry point to a larger conversation.

This is how the tag looks:
image


vcokltfre (approved):
Looks good to me, good tag for getting people into using f-strings!

bot/resources/tags/string-formatting.md
```diff
@@ -0,0 +1,27 @@
+**String Formatting Mini Language**
+The String Formatting Language in Python is a powerful way to tailor the display of strings and other data structures. This string formatting mini language works for f-strings and .format().
+
+Take a look at some of these examples!
+```py
+my_num = 2134234523
+print(f"{my_num:,}")
```
brad90four, line 7:
```suggestion
print(f"{my_num = :,}")
```

Numerlor:
Have you considered using a REPL style codeblock for displaying the results next to the expressions that print them? Or was this done to avoid the potential confusion around prompts

bot/resources/tags/string-formatting.md
```diff
@@ -0,0 +1,27 @@
...
+
+my_smaller_num = -30.0532234
+print(f"{my_smaller_num:=09.2f}")
```
brad90four, line 10:
```suggestion
print(f"{my_smaller_num = :=09.2f}")
```

bot/resources/tags/string-formatting.md
```diff
@@ -0,0 +1,27 @@
...
+
+my_str = "Center me!"
+print(f"{my_str:^20}")
```
brad90four, line 13:
```suggestion
print(f"{my_str = :^20}")
```

bot/resources/tags/string-formatting.md
```diff
@@ -0,0 +1,27 @@
...
+
+repr_str = "Spam \t Ham"
+print(f"{repr_str!r}")
```
brad90four, line 16:
```suggestion
print(f"{repr_str = !r}")
```

bot/resources/tags/string-formatting.md
```diff
@@ -0,0 +1,27 @@
...
+```
+Results:
+```
+2,134,234,523
+-00030.05
+      Center me!
+'Spam \t Ham'
```
brad90four, line 23:
```suggestion
my_num = 2,134,234,523
my_smaller_num = -00030.05
my_str =       Center me!
my_repr = 'Spam \t Ham'
```

brad90four (commented):
Looks great to me! Just as a suggestion, you could add "=" to the f-string as well, it will print out the variable name and the value. It is also part of the mini language in f-strings.

But adding it to everything could also make things more confusing to people that don't know about f-strings...so take it for what you want to lol.

janine9vn:
> 
> 
> Have you considered using a REPL style codeblock for displaying the results next to the expressions that print them? Or was this done to avoid the potential confusion around prompts

I have and decided that because this is primarily showcasing final print results I wanted to avoid any potential confusion with some of the REPL environment indicators.

mbaruh:
Having to jump between the two code blocks is a bit confusing. Maybe the result can be shown as a comment near the line as in other tags?

janine9vn:
> 
> 
> Having to jump between the two code blocks is a bit confusing. Maybe the result can be shown as a comment near the line as in other tags?

So I did try it and it ended up looking cluttered and confusing to me. At least for me, I don't necessarily expect the answer to be provided as a comment, so I don't pay that much attention to it. Let me see if I can make the change and illustrate.

image

Numerlor:
If we're giving center as an example I think some characters should be put around the replacement field to show that it's centered (pipes for example).
Showing the repr would be ideal but I think that could conflict with the repr below.

For the comments I think it's nicer to read than the two codeblocks, but the color used for them may be a bit too muted

bot/resources/tags/string-formatting.md
```diff
@@ -0,0 +1,27 @@
+**String Formatting Mini Language**
+The String Formatting Language in Python is a powerful way to tailor the display of strings and other data structures. This string formatting mini language works for f-strings and .format().
```
Bluenix2, line 2: Isn't it `mini-language`?

mbaruh:
Maybe something like this?
image
Also explains what each formatting is supposed to do

Numerlor:
how much vertical space does that introduce? I'm mostly concerned about mobile users where it may take up the whole viewport

bot/resources/tags/string-formatting.md
Bluenix2, line 16: I feel like this is a bad example, it doesn't actually showcase that we're doing the same as `rep()`. Because it outputs exacltly the string you typed.

We should find something that has a different `__repr__` than `__str__`, or at least point out that the quotes were included when we printed it.

bot/resources/tags/string-formatting.md
Bluenix2, line 2: Can you mention that it seems to have come from the usefulness of old-style `%s` formatting?

That's at least my take-away from [PyFormat](pyformat.info)

bot/resources/tags/string-formatting.md
Bluenix2: I disagree with these, it means that we show no examples without it and now someone might think you have to do it.

bot/resources/tags/string-formatting.md
Bluenix2, line 2:
```suggestion
The String Formatting Language in Python is a powerful way to tailor the display of strings and other data structures. This string formatting mini language works for f-strings and `.format()`.
```

Bluenix2 (commented):
Can we also change some examples to be using `.format()` and that it works either way?

bot/resources/tags/string-formatting.md
janine9vn: Why doesn't it showcase this? The `\t` is for tabs and the repr is showcasing that the tab isn't rendered in the repr version.

bot/resources/tags/string-formatting.md
janine9vn: What useful information does that provide to people? I'm not sure how knowing it derives from `%s` is necessarily useful for this tag.

bot/resources/tags/string-formatting.md
Bluenix2: I feel like that can easily be mixed up because the string quite literally does `\t`.

janine9vn:
@Bluenix2 I'm not sure what you're referencing with this statement:
> I disagree with these, it means that we show no examples without it and now someone might think you have to do it.

Also, I almost exclusively prefer the use of f-strings vs `.format()` with the exception of handling newlines in `str.join()`. This tag isn't meant to be a "here's all the ways it could work" but covering the most common use cases/illustrative examples and having people explore for themselves.

bot/resources/tags/string-formatting.md
Bluenix2: There are still uses for `%s`, like with logging when you need lazily evaluated strings, and knowing that you have similar options with `%s` could prove helpful?

That said I am pretty neutral on the matter, but thought I'd bring it up for judgement haha

brad90four:
> 
> 
> @Bluenix2 I'm not sure what you're referencing with this statement:
> 
> > I disagree with these, it means that we show no examples without it and now someone might think you have to do it.
> 
> Also, I almost exclusively prefer the use of f-strings vs `.format()` with the exception of handling newlines in `str.join()`. This tag isn't meant to be a "here's all the ways it could work" but covering the most common use cases/illustrative examples and having people explore for themselves.

I think that was a reply to my commented suggestion, it could add confusion that people think they have to use "=" in the f-string. Maybe just as a separate example is better to show what it does instead of mixing multiple formats together.

Bluenix2:
> @Bluenix2 I'm not sure what you're referencing with this statement:
> 
> > I disagree with these, it means that we show no examples without it and now someone might think you have to do it.

That was in response to @brad90four's comments where they suggested adding ` = ` to all examples.

> Also, I almost exclusively prefer the use of f-strings vs `.format()` with the exception of handling newlines in `str.join()`. This tag isn't meant to be a "here's all the ways it could work" but covering the most common use cases/illustrative examples and having people explore for themselves.

Sounds good!

brad90four (approved):
LGTM

Not too shabby for a speed run right?

PR #1820, opened 2021-09-09
Add metricity to docker-compose

ChrisLovering:
By adding metricity to the compose, we allow it to migrate itself, rather than needing the site to do it.

Defaulting 'skip_metricity' to true means that it will run migrations, but not actually start the bot. This means we don't add another service that needs to run all the time, which could impact some contribs on lower powered hardware.

This needs to be merged at the same time as https://github.com/python-discord/site/pull/588

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,11 @@ services:
       POSTGRES_DB: pysite
       POSTGRES_PASSWORD: pysite
       POSTGRES_USER: pysite
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U pysite"]
+      interval: 2s
+      timeout: 1s
+      retries: 5
 
   redis:
     << : *logging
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -31,6 +36,19 @@ services:
...
+    << : *restart_policy
+    depends_on:
+      postgres:
+        condition: service_healthy
+    image: ghcr.io/python-discord/metricity:latest
+    env_file:
+      - .env
+    environment:
+      DATABASE_URI: postgres://pysite:pysite@postgres:5432/pysite
+      skip_metricity: true  # Delete this to run the metricity bot too
```
Numerlor, line 50: Is there some way of accepting this from .env and defaulting to true or will docker always use the val from here?

If it stays I think it'd be worth mentioning the override file here as changing the value here is not a permanent solution as it's tracked

ChrisLovering: I think this is possible now that I've pushed https://github.com/python-discord/metricity/pull/7/commits/9a5d4ce294c3d0a9d588bc4e1435606690e8c02f will test it tonight

ChrisLovering: https://github.com/python-discord/bot/pull/1820/commits/4d7496d20c867b1d5201e7d60d3c22d18e2a21a3 Much nicer interface, thanks

mbaruh:
Not an actual requested change (unless you want to make it), but having to set `SKIP_METRICITY` to false seems like a double negative 😅
Setting `USE_METRICITY` to true seems more intuitive.

ChrisLovering:
> Not an actual requested change (unless you want to make it), but having to set `SKIP_METRICITY` to false seems like a double negative 😅
> Setting `USE_METRICITY` to true seems more intuitive.

Good point, I've raised https://github.com/python-discord/metricity/pull/8 will update this PR to match

mbaruh (approved):
Works well 👍

Xithrius (approved):
Also yes.

PR #1822, opened 2021-09-10
python-news escape fixes

### LINKED ISSUE #1821: Too strict markdown escape in PythonNews
The bot can be seen escaping underscores in code [in this message](https://discord.com/channels/267624335836053506/704372456592506880/885531093158215791).

The regex (if possible) should be updated to not insert `\` inside codelines or codeblocks such as (in this case) `__eq__`.

On the bright side, it did work for one ocasion I've screenshoted below!
image

TizzySaurus:
Closes #1821.

Adds escapes to post titles and also no longer escapes markdown inside of codeblocks for python news posts.

Haven't been able to test in the context of python-news but have done simulated tests locally which work:
```py
>>> string = """
Hey,

I noticed that writing a class with a classmethod __eq__, it does not work as expected:
```py
class ANY:
    @classmethod
    def __eq__(cls, other):
        return True

assert ANY == 2   # succeeds
assert ANY == 2   # fails
\`\`\`  # backslashes aren't actually there, added here because of GitHub markdown
"""
>>> regex = re.compile(
    r"(?P<codeblock>`.*?`)"  # matches everything within a codeblock
    r"|(?P<markdown>(?<!\\)[_|])",  # matches unescaped `_` and `|`
    re.DOTALL  # required to support multi-line codeblocks
)
>>> indices_of_markdown_outside_codeblocks = {
    m.start()
    for m in re.finditer(regex, string)
    if m.group('markdown')
}
>>> indices_of_markdown_outside_codeblocks
{640, 641, 637, 57, 58, 636, 61, 62}
>>> new_str = regex.sub(
    lambda match: match.group("codeblock") or "\\" + match.group("markdown"),
    content
)
>>> print(new_str)
"""  # quotes added for GitHub markdown
Hey,

I noticed that writing a class with a classmethod \_\_eq\_\_, it does not work as expected:
```py
class ANY:
    @classmethod
    def __eq__(cls, other):
        return True

assert ANY == 2   # succeeds
assert ANY == 2   # fails
\`\`\`  # backslashes aren't actually there, added here because of GitHub markdown
"""  # quotes added for GitHub markdown
```
As we can see, markdown inside of codeblocks no-longer gets escaped. Have done others tests for spoilers/in-line codeblocks/pre-escaped locally and everything works perfectly (thanks to quite a bit of help from @mbaruh).

### DIFF
--- a/bot/exts/info/python_news.py
+++ b/bot/exts/info/python_news.py
@@ -139,7 +139,7 @@ class PythonNews(Cog):
             self.bot.stats.incr("python_news.posted.pep")
 
             if msg.channel.is_news():
-                log.trace("Publishing PEP annnouncement because it was in a news channel")
+                log.trace("Publishing PEP announcement because it was in a news channel")
                 await msg.publish()
 
         # Apply new sent news to DB to avoid duplicate sending
--- a/bot/exts/info/python_news.py
+++ b/bot/exts/info/python_news.py
@@ -119,7 +119,7 @@ class PythonNews(Cog):
 
             # Build an embed and send a webhook
             embed = discord.Embed(
-                title=new["title"],
+                title=self.escape_markdown(new["title"]),
                 description=self.escape_markdown(new["summary"]),
                 timestamp=new_datetime,
                 url=new["link"],
@@ -189,7 +189,7 @@ class PythonNews(Cog):
 
                 # Build an embed and send a message to the webhook
                 embed = discord.Embed(
-                    title=thread_information["subject"],
+                    title=self.escape_markdown(thread_information["subject"]),
                     description=content[:1000] + f"... [continue reading]({link})" if len(content) > 1000 else content,
                     timestamp=new_date,
                     url=link,
--- a/bot/exts/info/python_news.py
+++ b/bot/exts/info/python_news.py
@@ -22,6 +22,14 @@ THREAD_URL = "https://mail.python.org/archives/list/{list}@python.org/thread/{id
 
 AVATAR_URL = "https://www.python.org/static/opengraph-icon-200x200.png"
 
+# By first matching everything within a codeblock,
+# when matching markdown it won't be within a codeblock
+MARKDOWN_REGEX = re.compile(
+    r"(?P<codeblock>`.*?`)"  # matches everything within a codeblock
+    r"|(?P<markdown>(?<!\\)[_|])",  # matches unescaped `_` and `|`
+    re.DOTALL  # required to support multi-line codeblocks
+)
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


bot/exts/info/python_news.py
```diff
@@ -75,8 +75,17 @@ async def get_webhook_names(self) -> None:
 
     @staticmethod
     def escape_markdown(content: str) -> str:
-        """Escape the markdown underlines and spoilers."""
-        return re.sub(r"[_|]", lambda match: "\\" + match[0], content)
+        """Escape the markdown underlines and spoilers that aren't in codeblocks."""
```
Akarys42, line 78: Can we have explanations for this regex? Potentially also move it to `bot/utils/regex.py`.

Akarys42 (changes requested):

bot/exts/info/python_news.py
```diff
@@ -75,8 +75,17 @@ async def get_webhook_names(self) -> None:
...
+        markdown_regex = r"(?:`[^`\\]*(?:\\.[^`\\]*)*`)|([_|])"
+        markdown_outside_codeblocks_indices = [m.span()[0] for m in re.finditer(markdown_regex, content) if m.group(1)]
```
Akarys42, line 80: Seems like I forgot about it, please use a named capture group instead of relying on indices to make the regex more resilient

bot/exts/info/python_news.py
TizzySaurus: I've redesigned this PR quite a bit.

The regex does now have comments explaining each part, and also a comment above where it gets used to explain why the regex works.

In terms of moving it, I decided to instead just make it a constant within the file.

bot/exts/info/python_news.py
TizzySaurus: Addressed as part of the redesign :+1:

Ryu1845:
I made the code to escape the markdown previously so I fixed it but I only just noticed someone else was working on it. Here's my version of the fix, https://github.com/Ryu1845/bot/commit/051fc1559bcd2c7e2a1d3a67332a27a81431049a .

Ryu1845:
I added the escaping to the titles too.

TizzySaurus:
> I made the code to escape the markdown previously so I fixed it but I only just noticed someone else was working on it. Here's my version of the fix, [Ryu1845@5a6dc88](https://github.com/Ryu1845/bot/commit/5a6dc8840fa39cb5d6baedd166a48569f45e26e8) .

After discussion on Discord in dev-contrib and necessary changes etc. this has been committed, thanks :+1:

This PR is now awaiting reviews.

Xithrius (approved):

ChrisLovering (approved):

PR #1831, opened 2021-09-17
Add `CustomLogger` to prevent IDE lint warnings when calling `log.trace`.

TizzySaurus:
Closes bot#1830.

Implements the proposed `CustomLogger` class to fix the lint warnings.

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -9,7 +9,6 @@ the custom configuration. Any settings left
 out in the custom user configuration will stay
 their default values from `config-default.yml`.
 """
-
 import logging
 import os
 from collections.abc import Mapping
@@ -25,7 +24,7 @@ try:
 except ModuleNotFoundError:
     pass
 
-log = logging.getLogger(__name__)
+log = logging.getLogger(__name__)  # use logging.getLogger (not bot.utils.logging.get_logger) to prevent circular import
 
 
 def _env_var_constructor(loader, node):
--- a/bot/log.py
+++ b/bot/log.py
@@ -14,11 +14,27 @@ from bot import constants
 TRACE_LEVEL = 5
 
 
+class CustomLogger(Logger):
+    """Custom implementation of the `Logger` class with an added `trace` method."""
+
+    def trace(self, msg: str, *args, **kwargs) -> None:
+        """
+        Log 'msg % args' with severity 'TRACE'.
+
+        To pass exception information, use the keyword argument exc_info with
+        a true value, e.g.
+
+        logger.trace("Houston, we have an %s", "interesting problem", exc_info=1)
+        """
+        if self.isEnabledFor(TRACE_LEVEL):
+            self._log(TRACE_LEVEL, msg, args, **kwargs)
+
+
 def setup() -> None:
     """Set up loggers."""
     logging.TRACE = TRACE_LEVEL
     logging.addLevelName(TRACE_LEVEL, "TRACE")
-    Logger.trace = _monkeypatch_trace
+    logging.setLoggerClass(CustomLogger)
 
     format_string = "%(asctime)s | %(name)s | %(levelname)s | %(message)s"
     log_format = logging.Formatter(format_string)
@@ -71,19 +87,6 @@ def setup_sentry() -> None:
         ],
         release=f"bot@{constants.GIT_SHA}"
     )
-
-
-def _monkeypatch_trace(self: logging.Logger, msg: str, *args, **kwargs) -> None:
-    """
-    Log 'msg % args' with severity 'TRACE'.
-
-    To pass exception information, use the keyword argument exc_info with
-    a true value, e.g.
-
-    logger.trace("Houston, we have an %s", "interesting problem", exc_info=1)
-    """
-    if self.isEnabledFor(TRACE_LEVEL):
-        self._log(TRACE_LEVEL, msg, args, **kwargs)
 
 
 def _set_trace_loggers() -> None:
--- a/bot/utils/services.py
+++ b/bot/utils/services.py
@@ -6,7 +6,7 @@ from aiohttp import ClientConnectorError
 import bot
 from bot.constants import URLs
 
-log = logging.getLogger(__name__)
+log = logging.getLogger(__name__)  # use logging.getLogger (not bot.utils.logging.get_logger) to prevent circular import
 
 FAILED_REQUEST_ATTEMPTS = 3
 
--- a/bot/__main__.py
+++ b/bot/__main__.py
@@ -1,11 +1,10 @@
-import logging
-
 import aiohttp
 
 import bot
 from bot import constants
 from bot.bot import Bot, StartupError
 from bot.log import setup_sentry
+from bot.utils.logging import get_logger
 
 setup_sentry()
 
@@ -21,7 +20,7 @@ except StartupError as e:
         message = "Could not connect to Redis. Is it running?"
 
     # The exception is logged with an empty message so the actual message is visible at the bottom
-    log = logging.getLogger("bot")
+    log = get_logger("bot")
     log.fatal("", exc_info=e.exception)
     log.fatal(message)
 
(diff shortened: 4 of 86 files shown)


Xithrius:
[here](https://github.com/python-discord/bot/blob/1adae82ebc32fc39636f4ca2d20ec2e2b3c3fbd0/bot/log.py#L17-L30) is the `CustomLogger` class, so anyone reviewing this doesn't have to scroll to find it.

TizzySaurus:
> [Here](https://github.com/python-discord/bot/blob/1adae82ebc32fc39636f4ca2d20ec2e2b3c3fbd0/bot/log.py#L17-L30) is the `CustomLogger` class, so anyone reviewing this doesn't have to scroll to find it.

There's also the [`get_logger` util function](https://github.com/python-discord/bot/blob/1adae82ebc32fc39636f4ca2d20ec2e2b3c3fbd0/bot/utils/logging.py#L7-L9) (entire file is new)

All other changes are removing the old monkeypatch `trace` or updating imports from
```py
import logging

```
to
```py
from bot.utils.logging import get_logger

log = get_logger(__name__)
```

bot/log.py
```diff
@@ -14,11 +14,27 @@
 TRACE_LEVEL = 5
 
 
+class CustomLogger(Logger):
```
kosayoda, line 17: The logger should inherit [.getLoggerClass()](https://docs.python.org/3/library/logging.html?highlight=logging#logging.getLoggerClass) to play nice with other loggers.

bot/log.py
```diff
@@ -14,11 +14,27 @@
...
+        """
+        Log 'msg % args' with severity 'TRACE'.
+
+        To pass exception information, use the keyword argument exc_info with
+        a true value, e.g.
+
+        logger.trace("Houston, we have an %s", "interesting problem", exc_info=1)
+        """
+        if self.isEnabledFor(TRACE_LEVEL):
+            self._log(TRACE_LEVEL, msg, args, **kwargs)
```
kosayoda, line 30: It's better to use the non-internal `.log()` method to do the work for us, it essentially does [what you're doing here but extra](https://github.com/python/cpython/blob/778b07565e38cc94aa90318eb47b9cd09716756a/Lib/logging/__init__.py#L1536).

kosayoda (dismissed):
Nice, I've personally wanted this for a while 👍🏻 

Missing log migration: `bot/exts/recruitment/talentpool/_cog.py`

bot/log.py
TizzySaurus: So to clarify, this would become `self.log(...)`?

bot/log.py
TizzySaurus: Changing this to `class CustomLogger(logging.getLoggerClass())` results in PyCharm giving a warning stating `Unresolved attribute reference 'isEnabledFor' for class 'CustomLogger'` and `Unresolved attribute reference '_log' for class 'CustomLogger'` (same happens for `log` instead of `_log`).

Even [`cast`](https://docs.python.org/3/library/typing.html#typing.cast)ing doesn't fix this (`cast(Logger, logging.getLoggerClass())`)

bot/log.py
kosayoda: Yep, and you pass `*args` instead of `args`

bot/log.py
kosayoda: > Unresolved attribute reference 'isEnabledFor' for class 'CustomLogger'

This seems to be a PyCharm bug, as both `mypy` and `pyright` correctly infers the correct type.

However, `mypy` (but not `pyright`) returns a different error relating to dynamic base classes. The workaround for that is
```python

if typing.TYPE_CHECKING:
    LoggerClass = Logger
else:
    LoggerClass = logging.getLoggerClass()

class CustomLogger(LoggerClass):
```
A nice explanation of why casting and even type variables will not work for this case can be found here: https://github.com/python/mypy/issues/11050
An even longer read (with a guido appearance) can be found here: https://github.com/python/mypy/issues/2477

TizzySaurus: Committed as part of [ca2d4a5](https://github.com/python-discord/bot/pull/1831/commits/1adae82ebc32fc39636f4ca2d20ec2e2b3c3fbd0)

bot/log.py
TizzySaurus: Committed as part of [ca2d4a5](https://github.com/python-discord/bot/pull/1831/commits/1adae82ebc32fc39636f4ca2d20ec2e2b3c3fbd0)

TizzySaurus:
NB: Linting & Test currently failing due to [maintenance on the coveralls API](https://www.moesif.com/public-apis/status/coveralls). Will rerun checks later :+1:

TizzySaurus:
NB: The `getLogger` in  `bot.utils.members` will also need to be migrated once #1837 has been merged.

bot/constants.py
```diff
@@ -25,7 +24,7 @@
 except ModuleNotFoundError:
     pass
 
-log = logging.getLogger(__name__)
+log = logging.getLogger(__name__)  # use logging.getLogger (not bot.utils.logging.get_logger) to prevent circular import
```
wookie184, line 27: I'm not sure this is the best option here, there was some discussion on discord on this starting here https://discord.com/channels/267624335836053506/635950537262759947/894235145362616390, and it seems the conclusion was that the logs here don't work anyway because this is run before logging is set up.

bot/utils/services.py
```diff
@@ -6,7 +6,7 @@
 import bot
 from bot.constants import URLs
 
-log = logging.getLogger(__name__)
+log = logging.getLogger(__name__)  # use logging.getLogger (not bot.utils.logging.get_logger) to prevent circular import
```
wookie184, line 9: Similar to other comment, although I feel we should be able to use the logger here, and using `get_logger` for it would be nice, although i'm not sure what the best way forwards with that is. This problem only seems to happen because this file is imported in the `utils/__init__.py`

wookie184 (commented):
I'm not sure using `log = logging.getLogger(__name__)` is the best solution to the circular import problem, although i'm not sure what is either...

Other than that this looks good.

bot/utils/services.py
TizzySaurus: This has been resolved with the little re-write :+1:

bot/constants.py
TizzySaurus: This has been resolved with the little re-write :+1:

TizzySaurus:
Have fixed the root issue of the circular imports, so now all files should be using `bot.log.get_logger` instead of `logging.getLogger`.

Also added any missing migrations from PRs made after I started this and squashed, fixed the conflicts (all 25), and did a commit for making newlines after imports consistent (part of PyCharm's "Optimize Imports" feature).

This should now be fully read for review @kosayoda @wookie184 :+1:

bot/constants.py
```diff
@@ -123,11 +119,11 @@ def check_required_keys(keys):
                 if lookup is None:
                     raise KeyError(key)
         except KeyError:
-            log.critical(
+            raise (
```
wookie184, line 122: Need an exception class here, `KeyError` would probably be good, So `raise KeyError("message")`. That or we could print and then just `raise`, maybe that would be better as it would keep the context of where the KeyError was raised? Can't remember exactly how these things work

wookie184 (changes requested):
Two small things, other than that it looks good.

bot/constants.py
```diff
@@ -186,8 +182,8 @@ def __getattr__(cls, name):
                 (cls.section, cls.subsection, name)
                 if cls.subsection is not None else (cls.section, name)
             )
-            # Only an INFO log since this can be caught through `hasattr` or `getattr`.
-            log.info(f"Tried accessing configuration variable at `{dotted_path}`, but it could not be found.")
+            # Only an print since this can be caught through `hasattr` or `getattr`.
+            print(f"Tried accessing configuration variable at `{dotted_path}`, but it could not be found.")
```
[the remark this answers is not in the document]
TizzySaurus: Removed :+1:

PR #1834, opened 2021-09-19
Request all users before marking the guild as available

ChrisLovering:
This ensures that the member cache is fully populated before we mark the guild as ready, as this is what a lot of cogs use as indication of all caches being populated."

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -260,15 +260,14 @@ class Bot(commands.Bot):
 
     async def on_guild_available(self, guild: discord.Guild) -> None:
         """
-        Set the internal guild available event when constants.Guild.id becomes available.
+        Set the internal guild available event when constants.Guild.id becomes available and all members are fetched.
 
-        If the cache appears to still be empty (no members, no channels, or no roles), the event
-        will not be set.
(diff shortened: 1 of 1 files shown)


vcokltfre (approved):

mbaruh (dismissed):

onerandomusername:
Before merging, it might be worth seeing how long this takes to chunk on the main server.

onerandomusername:
Because of how long this event takes to run, it could have problems with where that event is waited for. In addition, that event is called every time this guild comes up after an outage. This could be problematic because it would be calling await guild.chunk() multiple times while the other ones are still running. I think this pr could be a major source of problems and should be investigated further

vcokltfre:
On GTA Online, a server with 390K members, this takes about 4 seconds.

onerandomusername:
Interesting, its taken over 11 minutes to chunk in pydis.

vcokltfre:
Ignore previous comment, I was thinking of a different thing we run at startup, testing again now

ChrisLovering:
This is a weird one, this isn't the solution to the problem, as we don't know what the problem is.

After redeploying the bot, the cache is full...

PR #1836, opened 2021-09-20
Fix typos and grammar in tags

### LINKED ISSUE #1832: Tag Content Revision
A number of tags have typos / grammatical issues. This issue aims to resolve these.

For example, the `!windows-path` tag has a typo in the final codeblock, where it states:

> ... Python 3.6 stars ...

Here "stars" should be "starts".

I'll work on this myself over the upcoming days.

TizzySaurus:
Closes #1832.

Fixes typos and grammar in tags, as well as changing a few examples etc. where I felt appropriate (e.g. in `!precedence` I changed to examples where the order is more obviously important).

### DIFF
--- a/bot/resources/tags/async-await.md
+++ b/bot/resources/tags/async-await.md
@@ -2,18 +2,18 @@
 
 Python provides the ability to run multiple tasks and coroutines simultaneously with the use of the `asyncio` library, which is included in the Python standard library.
 
-This works by running these coroutines in an event loop, where the context of which coroutine is being run is switches periodically to allow all of them to run, giving the appearance of running at the same time. This is different to using threads or processes in that all code is run in the main process and thread, although it is possible to run coroutines in threads.
+This works by running these coroutines in an event loop, where the context of the running coroutine switches periodically to allow all the coroutines to run, thus giving the appearance of running at the same time. This is different to using threads or processes in that all code runs in the main process and thread, although it is possible to run coroutines in threads.
 
 To call an async function we can either `await` it, or run it in an event loop which we get from `asyncio`.
 
-To create a coroutine that can be used with asyncio we need to define a function using the async keyword:
+To create a coroutine that can be used with asyncio we need to define a function using the `async` keyword:
 ```py
 async def main():
     await something_awaitable()
 ```
-Which means we can call `await something_awaitable()` directly from within the function. If this were a non-async function this would have raised an exception like: `SyntaxError: 'await' outside async function`
+Which means we can call `await something_awaitable()` directly from within the function. If this were a non-async function, this would have raised the exception `SyntaxError: 'await' outside async function`
 
-To run the top level async function from outside of the event loop we can get an event loop from `asyncio`, and then use that loop to run the function:
+To run the top level async function from outside the event loop, we can get the event loop from `asyncio` and then use it to run the function:
 ```py
 from asyncio import get_event_loop
 
--- a/bot/resources/tags/traceback.md
+++ b/bot/resources/tags/traceback.md
@@ -1,4 +1,4 @@
-Please provide a full traceback to your exception in order for us to identify your issue.
+Please provide the full traceback for your exception in order to help us identify your issue.
 
 A full traceback could look like:
 ```py
@@ -6,13 +6,13 @@ Traceback (most recent call last):
     File "tiny", line 3, in
         do_something()
     File "tiny", line 2, in do_something
-        a = 6 / 0
+        a = 6 / b
 ZeroDivisionError: integer division or modulo by zero
 ```
 The best way to read your traceback is bottom to top.
 
-• Identify the exception raised (e.g. ZeroDivisionError)  
-• Make note of the line number, and navigate there in your program.  
-• Try to understand why the error occurred.  
+• Identify the exception raised (e.g. `ZeroDivisionError`)  
+• Make note of the line number (in this case 2), and navigate there in your program.  
+• Try to understand why the error occurred (in this case because `b` must be `0`).
 
-To read more about exceptions and errors, please refer to the [PyDis Wiki](https://pythondiscord.com/pages/asking-good-questions/#examining-tracebacks) or the [official Python tutorial.](https://docs.python.org/3.7/tutorial/errors.html)
+To read more about exceptions and errors, please refer to the [PyDis Wiki](https://pythondiscord.com/pages/guides/pydis-guides/asking-good-questions/#examining-tracebacks) or the [official Python tutorial](https://docs.python.org/3.7/tutorial/errors.html).
--- a/bot/resources/tags/windows-path.md
+++ b/bot/resources/tags/windows-path.md
@@ -1,10 +1,10 @@
 **PATH on Windows**
 
-If you have installed Python but you forgot to check the *Add Python to PATH* option during the installation you may still be able to access your installation with ease.
+If you have installed Python, but you forgot to check the *Add Python to PATH* option during the installation, you may still be able to access your installation with ease.
 
-If you did not uncheck the option to install the Python launcher then you will find a `py` command on your system. If you want to be able to open your Python installation by running `python` then your best option is to re-install Python.
+If you did not uncheck the option to install the Python launcher then you will find a `py` command on your system. If you want to be able to open your Python installation by running `python`, then your best option is to re-install Python.
 
-Otherwise, you can access your install using the `py` command in Command Prompt. Where you may type something with the `python` command like:
+Otherwise, you can access your installation using the `py` command in Command Prompt, where you may type something with the `python` command such as:
 ```
 C:\Users\Username> python3 my_application_file.py
 ```
@@ -24,7 +24,7 @@ You can also access different versions of Python using the version flag, like so
 C:\Users\Username> py -3.7
 ... Python 3.7 starts ...
 C:\Users\Username> py -3.6
-... Python 3.6 stars ...
+... Python 3.6 starts ...
 C:\Users\Username> py -2
 ... Python 2 (any version installed) starts ...
 ```
--- a/bot/resources/tags/xy-problem.md
+++ b/bot/resources/tags/xy-problem.md
@@ -1,7 +1,7 @@
 **xy-problem**
 
-Asking about your attempted solution rather than your actual problem.
+The XY problem can be summarised as asking about your attempted solution, rather than your actual problem.
 
 Often programmers will get distracted with a potential solution they've come up with, and will try asking for help getting it to work. However, it's possible this solution either wouldn't work as they expect, or there's a much better solution instead.
 
-For more information and examples: http://xyproblem.info/
+For more information and examples, see http://xyproblem.info/
--- a/bot/resources/tags/ytdl.md
+++ b/bot/resources/tags/ytdl.md
@@ -1,4 +1,4 @@
-Per [Python Discord's Rule 5](https://pythondiscord.com/pages/rules), we are unable to assist with questions related to youtube-dl, pytube, or other YouTube video downloaders as their usage violates YouTube's Terms of Service.
+Per [Python Discord's Rule 5](https://pythondiscord.com/pages/rules), we are unable to assist with questions related to youtube-dl, pytube, or other YouTube video downloaders, as their usage violates YouTube's Terms of Service.
 
 For reference, this usage is covered by the following clauses in [YouTube's TOS](https://www.youtube.com/static?gl=GB&template=terms), as of 2021-03-17:
 ```
--- a/bot/resources/tags/zip.md
+++ b/bot/resources/tags/zip.md
@@ -3,7 +3,7 @@ The zip function allows you to iterate through multiple iterables simultaneously
 ```py
 letters = 'abc'
 numbers = [1, 2, 3]
-# zip(letters, numbers) --> [('a', 1), ('b', 2), ('c', 3)]
+# list(zip(letters, numbers)) --> [('a', 1), ('b', 2), ('c', 3)]
 for letter, number in zip(letters, numbers):
     print(letter, number)
 ```


bot/resources/tags/async-await.md
```diff
@@ -2,18 +2,18 @@
 
 Python provides the ability to run multiple tasks and coroutines simultaneously with the use of the `asyncio` library, which is included in the Python standard library.
 
-This works by running these coroutines in an event loop, where the context of which coroutine is being run is switches periodically to allow all of them to run, giving the appearance of running at the same time. This is different to using threads or processes in that all code is run in the main process and thread, although it is possible to run coroutines in threads.
+This works by running these coroutines in an event loop, where the context of the running coroutine switches periodically to allow all the coroutines to run, thus giving the appearance of running at the same time. This is different to using threads or processes in that all code runs in the main process and thread, although it is possible to run coroutines in threads.
```
brad90four, line 5:
```suggestion
This works by running these coroutines in an event loop, where the context of the running coroutine switches periodically to allow all of the coroutines to run, thus giving the appearance of running at the same time. This is different to using threads or processes in that all code runs in the main process and thread, although it is possible to run coroutines in threads.
```

brad90four (commented):
Literally the only thing that I could find, everything else looks solid.

bot/resources/tags/async-await.md
TizzySaurus: PyCharm doesn't like the added `of` here, likely since it's somewhat grammatically redundant, which is why I removed it.

image

brad90four: Sweet, I'm no English major so I'm good with it. If I say it out load I want to use of, but that could just be me. 👍

Bluenix2, line 5:
```suggestion
This works by running these coroutines in an event loop, where the context of the running coroutine switches periodically to allow all other coroutines to run, thus giving the appearance of running at the same time. This is different to using threads or processes in that all code runs in the main process and thread, although it is possible to run coroutines in threads.
```

bot/resources/tags/async-await.md
```diff
@@ -2,18 +2,18 @@
...
+To create a coroutine that can be used with asyncio we need to define a function using the `async` keyword:
 ```py
 async def main():
     await something_awaitable()
 ```
-Which means we can call `await something_awaitable()` directly from within the function. If this were a non-async function this would have raised an exception like: `SyntaxError: 'await' outside async function`
+Which means we can call `await something_awaitable()` directly from within the function. If this were a non-async function, this would have raised the exception `SyntaxError: 'await' outside async function`
 
-To run the top level async function from outside of the event loop we can get an event loop from `asyncio`, and then use that loop to run the function:
+To run the top level async function from outside the event loop, we can get the event loop from `asyncio` and then use it to run the function:
```
Bluenix2, line 16: Can we feature creep and take the time to recommend `asyncio.run()` instead?

bot/resources/tags/traceback.md
```diff
@@ -1,18 +1,18 @@
...
 ZeroDivisionError: integer division or modulo by zero
 ```
 The best way to read your traceback is bottom to top.
 
-• Identify the exception raised (e.g. ZeroDivisionError)  
-• Make note of the line number, and navigate there in your program.  
-• Try to understand why the error occurred.  
+• Identify the exception raised (e.g. `ZeroDivisionError`)  
+• Make note of the line number (in this case 2), and navigate there in your program.  
+• Try to understand why the error occurred (in this case because `b` must be `0`).
```
Bluenix2, line 16:
```suggestion
• Try to understand why the error occurred (in this case probably because `b` is `0`).
```

bot/resources/tags/xy-problem.md
```diff
@@ -1,7 +1,7 @@
 **xy-problem**
 
-Asking about your attempted solution rather than your actual problem.
+The XY problem can be summarised as asking about your attempted solution, rather than your actual problem.
```
Bluenix2, line 3:
```suggestion
The XY problem can be summarised as asking about your attempted solution rather than your actual problem.
```
This PR is supposed to be about improving grammar but adding a `,` there is invalid grammar.

bot/resources/tags/xy-problem.md
```diff
@@ -1,7 +1,7 @@
...
 
 Often programmers will get distracted with a potential solution they've come up with, and will try asking for help getting it to work. However, it's possible this solution either wouldn't work as they expect, or there's a much better solution instead.
 
-For more information and examples: http://xyproblem.info/
+For more information and examples, see http://xyproblem.info/
```
Bluenix2, line 7:
```suggestion
For more information and examples see http://xyproblem.info/
```

bot/resources/tags/windows-path.md
```diff
@@ -1,10 +1,10 @@
 **PATH on Windows**
 
-If you have installed Python but you forgot to check the *Add Python to PATH* option during the installation you may still be able to access your installation with ease.
+If you have installed Python, but you forgot to check the *Add Python to PATH* option during the installation, you may still be able to access your installation with ease.
```
Bluenix2, line 3: I disagree with this change. It may feel more correct but it is in-fact not: [Comma-splice - Wikipedia](https://en.wikipedia.org/wiki/Comma_splice)

To actually fix this we can make the following change:
```suggestion
If you have installed Python but forgot to check the *Add Python to PATH* option during the installation you may still be able to access your installation with ease.
```

bot/resources/tags/windows-path.md
```diff
@@ -1,10 +1,10 @@
...
 
-If you did not uncheck the option to install the Python launcher then you will find a `py` command on your system. If you want to be able to open your Python installation by running `python` then your best option is to re-install Python.
+If you did not uncheck the option to install the Python launcher then you will find a `py` command on your system. If you want to be able to open your Python installation by running `python`, then your best option is to re-install Python.
```
Bluenix2, line 5: I see this repeated a lot, I do not see how these changes lead to better grammar?

bot/resources/tags/windows-path.md
```diff
@@ -1,10 +1,10 @@
...
 
-Otherwise, you can access your install using the `py` command in Command Prompt. Where you may type something with the `python` command like:
+Otherwise, you can access your installation using the `py` command in Command Prompt, where you may type something with the `python` command such as:
```
Bluenix2, line 7:
```suggestion
Otherwise, you can access your installation using the `py` command in Command Prompt. Where you may type something with the `python` command such as:
```
The change regarding the comma did not make any improvement.

bot/resources/tags/traceback.md
TizzySaurus: Even "most likely" isn't correct. We know categorically that `b` *is* `0` from this error. There's no other way for the error to be raised.

I agree that `is` flows better than `must be` though so will change that.

Also going to update the error message since apparently I forgot to do that.

bot/resources/tags/xy-problem.md
TizzySaurus: I don't see how adding the `,` makes it invalid grammar. It's just adding a natural pause in speech.

bot/resources/tags/xy-problem.md
TizzySaurus: Same as above. I don't see how this comma is wrong.

bot/resources/tags/windows-path.md
TizzySaurus: This isn't a comma-splice because it's not joining two separate clauses. This is a subordinate clause (iirc that's the correct grammatical term for it).

I'll remove the second "you" though since the sentence does flow better without it.

bot/resources/tags/windows-path.md
TizzySaurus: As noted above, this is mostly just to add a natural pause in speech

bot/resources/tags/windows-path.md
TizzySaurus: Having `Where` as the next sentence doesn't make any sense to me here. Perhaps a semi-colon would be more appropriate?

bot/resources/tags/async-await.md
TizzySaurus: I'm not at all familiar with `asyncio` so I'd need more info as to what to write.

bot/resources/tags/async-await.md
TizzySaurus: Changed locally :+1:

bot/resources/tags/windows-path.md
Bluenix2: Yeah this one didn't feel "right" fully. "Where" itself may not be the right word?

I don't think I fully understand this sentence either. Is this sentence meant to convey that `py` and `python` are interchangable?

bot/resources/tags/windows-path.md
Bluenix2: I disagree that this improves grammar but if this is a change you would like to make that is fine with me. I will mark all of the comments I made about the added commas as resolved.

bot/resources/tags/traceback.md
Bluenix2: Yup sounds good

bot/resources/tags/windows-path.md
Bluenix2: Cool!

bot/resources/tags/async-await.md
Bluenix2: You can see the [documentation for some more context](https://docs.python.org/3/library/asyncio-task.html#asyncio.run), but I'd recommend something like the following:
> To run the top level async function from outside of the event loop we need to use `asyncio.run()` like this:
> ```py
> import asyncio
> 
> async def main():
>     await something_awaitable()
> 
> asyncio.run(main())

Then in the note after that change the reference of `run_until_complete()` to `asyncio.run()`.

bot/resources/tags/windows-path.md
TizzySaurus: I believe it's saying you can use `py` wherever you'd use `python` in the command line.

Perhaps `Otherwise, you can access your installation using the `py\` command in Command Prompt rather than `python`.

bot/resources/tags/async-await.md
TizzySaurus: Will look into this tonight :+1:

bot/resources/tags/windows-path.md
TizzySaurus: In fact, after re-reading the tag I realise this sentence is rather redundant as the previous sentence covers it, so I'll ignore for now

bot/resources/tags/async-await.md
TizzySaurus: Updated as part of [33b4f42](https://github.com/python-discord/bot/pull/1836/commits/33b4f42b481e6b5e1686f66803dc916595e4b457)

bot/resources/tags/async-await.md
```diff
@@ -2,27 +2,26 @@
 
 Python provides the ability to run multiple tasks and coroutines simultaneously with the use of the `asyncio` library, which is included in the Python standard library.
 
-This works by running these coroutines in an event loop, where the context of which coroutine is being run is switches periodically to allow all of them to run, giving the appearance of running at the same time. This is different to using threads or processes in that all code is run in the main process and thread, although it is possible to run coroutines in threads.
+This works by running these coroutines in an event loop, where the context of the running coroutine switches periodically to allow all other coroutines to run, thus giving the appearance of running at the same time. This is different to using threads or processes in that all code runs in the main process and thread, although it is possible to run coroutines in threads.
```
Bluenix2, line 5:
```suggestion
This works by running these coroutines in an event loop, where the context of the running coroutine switches periodically to allow all other coroutines to run, thus giving the appearance of running at the same time. This is different to using threads or processes in that all code runs in the main process and thread, although it is possible to run coroutines in other threads.
```

bot/resources/tags/async-await.md
```diff
@@ -2,27 +2,26 @@
...
 To call an async function we can either `await` it, or run it in an event loop which we get from `asyncio`.
 
-To create a coroutine that can be used with asyncio we need to define a function using the async keyword:
+To create a coroutine that can be used with asyncio we need to define a function using the `async` keyword:
 ```py
 async def main():
     await something_awaitable()
 ```
-Which means we can call `await something_awaitable()` directly from within the function. If this were a non-async function this would have raised an exception like: `SyntaxError: 'await' outside async function`
+Which means we can call `await something_awaitable()` directly from within the function. If this were a non-async function, this would have raised the exception `SyntaxError: 'await' outside async function`
```
Bluenix2, line 14:
```suggestion
Which means we can call `await something_awaitable()` directly from within the function. If this were a non-async function, it would have raised the exception `SyntaxError: 'await' outside async function`
```
Don't you think there should be `it` here instead?

bot/resources/tags/windows-path.md
```diff
@@ -1,30 +1,17 @@
 **PATH on Windows**
 
-If you have installed Python but you forgot to check the *Add Python to PATH* option during the installation you may still be able to access your installation with ease.
+If you have installed Python but forgot to check the *Add Python to PATH* option during the installation, you may still be able to access your installation with ease.
 
-If you did not uncheck the option to install the Python launcher then you will find a `py` command on your system. If you want to be able to open your Python installation by running `python` then your best option is to re-install Python.
+If you did not uncheck the option to install the Python launcher, then you'll instead have a `py` command which can be used in the same way. If you want to be able to access your Python installation via the `python` command, then your best option is to re-install Python (remembering to tick the `add to PATH` checkbox).
```
Bluenix2, line 5: First we say "Check *Add Python to PATH*" but then say "if you did not uncheck"?

TizzySaurus: The `uncheck` is referring to a separate setting (installing the Python launcher)

bot/resources/tags/async-await.md
TizzySaurus: Good catch :+1:

bot/resources/tags/async-await.md
TizzySaurus: Yep :+1:

bot/resources/tags/windows-path.md
Bluenix2: Huh, I don't even know the difference haha

TizzySaurus: Perhaps we should have the option in asterisks to better highlight it?

Bluenix2: Yeah we can put italic around *Python Installer*?

TizzySaurus: The installer has the setting as `py launcher` so will reword to this and add the asterisks

TizzySaurus: This has been done in [093fdaa](https://github.com/python-discord/bot/pull/1836/commits/093fdaa2b8c89b770d7ff2b7889077dd9e21ad2d)

bot/resources/tags/async-await.md
```diff
@@ -2,27 +2,26 @@
...
 
 async def main():
     await something_awaitable()
 
-loop = get_event_loop()
-loop.run_until_complete(main())
+asyncio.run(main())
 ```
-Note that in the `run_until_complete()` where we appear to be calling `main()`, this does not execute the code in `main`, rather it returns a `coroutine` object which is then handled and run by the event loop via `run_until_complete()`.
+Note that in the `asyncio.run()` where we appear to be calling `main()`, this does not execute the code in `main`, rather it returns a `coroutine` object which is then handled and run by the event loop via `asyncio.run`.
```
Bluenix2, line 25:
```suggestion
Note that in the `asyncio.run()` where we appear to be calling `main()`, this does not execute the code in `main`, rather it creates a `coroutine` object which is then handled and run by the event loop via `asyncio.run`.
```
Would this be an improvement?

TizzySaurus: `returnI personally think it should have `return`.. Perhaps `creates and returns` if you want to mention that the coroutine gets created?

Bluenix2: Yeah I think that's a good compromise. I do think it is important to mention that `main() is not main()`.

PR #1837, opened 2021-09-20
Add get or fetch util

ChrisLovering:
This is now needed, as we have noticed that it's not guaranteed that the member cache will always be fully populated. Leading to members not having their help cooldown roles removed.


bot/exts/moderation/stream.py
```diff
@@ -47,8 +48,9 @@ async def _reload_tasks_from_redis(self) -> None:
         """Reload outstanding tasks from redis on startup, delete the task if the member has since left the server."""
         await self.bot.wait_until_guild_available()
         items = await self.task_cache.items()
+        guild = self.bot.get_guild(Guild.id)
         for key, value in items:
-            member = self.bot.get_guild(Guild.id).get_member(key)
+            member = await members.get_or_fetch(guild, key)
 
             if not member:
```
wookie184, line 55: This section (in the `if not member:`, github wont let me comment on it directly :/) should probably be updated as fetch_member would already have been tried.

wookie184 (changes requested):

bot/exts/moderation/stream.py
ChrisLovering: Ah yea, good catch. Force pushed :)

wookie184 (dismissed):
LGTM

TizzySaurus (commented):
There's a possible missed migration [here](https://github.com/python-discord/bot/blob/add-get-or-fetch-util/bot/exts/utils/reminders.py#L143) (L143 of bot/exts/utils/reminders.py).

Also not sure but should the converter instead catch `discord.ext.commands.MemberNotFound` (the actual error rather than a parent class)?

TizzySaurus:
> I purposefully skipped that since it's an generator, so I'd need to convert it to an async generator, so that cna be done in a different PR.

> This isn't a convertor, this is calling Guild.fetch_member, which doesn't raise that error afaik.

Looks good then. There's also the fact that the equivalent util for channels is called `try_get_channel` rather than `get_or_fetch`. Do we want to change one of these for consistency's sake or just leave it?

ToxicKidz (approved):
LGTM :shipit:

bot/utils/members.py
```diff
@@ -0,0 +1,24 @@
...
+
+log = logging.getLogger(__name__)
+
+
+async def get_or_fetch_member(guild: discord.Guild, member_id: int) -> t.Optional[discord.Member]:
+    """
+    Attempts to get a member from cache, on failure tries to fetch from the API.
+
+    Returns None if member is not part of the guild.
+    """
```
kwzrd, line 14: The [pydis style guide recommends imperative mood](https://pythondiscord.com/pages/guides/pydis-guides/contributing/style-guide/#mood). :smile:

kwzrd (commented):
Have you considered making the helper a bot instance method, having the `guild` argument optional and defaulting to pydis? In some cases, caller code could then have:

```python
bot.get_or_fetch_member(1234)
```

Instead of:

```python
import guild id
import get_or_fetch_member

guild = self.bot.get_guild(guild id)
get_or_fetch_member(guild, member id)
```

I haven't thought about it much, maybe this is a bad idea.

bot/utils/members.py
ChrisLovering: Updated the docstring

TizzySaurus:
I haven't tested anything so won't formally approve, but after a few things discussed on Discord this all looks good now :+1:

wookie184 (approved):
LGTM, tested and seems to work smoothly

PR #1838, opened 2021-09-20
Remove coveralls from lint-test

TizzySaurus:
Exactly as title says.

Removes coveralls from the lint-test stage due to the recent issues with their API.

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -121,13 +121,6 @@ jobs:
       - name: Run tests and generate coverage report
         run: pytest -n auto --cov --disable-warnings -q
 
-      # This step will publish the coverage reports coveralls.io and
-      # print a "job" link in the output of the GitHub Action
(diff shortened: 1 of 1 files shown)


HassanAbouelela (approved):

jb3 (approved):

.github/workflows/lint-test.yml
```diff
@@ -121,13 +121,6 @@ jobs:
       - name: Run tests and generate coverage report
         run: pytest -n auto --cov --disable-warnings -q
```
doublevcodes, line 122:
```suggestion
        run: pytest -n auto --disable-warnings -q
```
Shouldn't the coverage report not be generated now?

doublevcodes (changes requested):

doublevcodes (approved):

wookie184:
Just to note, part of the reason for the removal is that we decided it isn't something we really need, so we wont be re-adding it.

PR #1841, opened 2021-09-21
Allow helpers to edit their own nomination reason

### LINKED ISSUE #1617: Allow helpers to update nomination reasons
Sometimes, you want to add information to a nomination, possibly after getting to know the nominee better. This is currently not possible for helpers.

When a solution involving allowing staff to nominate the same member multiple times was proposed in `#nominations`, @wookie184 pointed out that the `!tp edit` command can be used to edit nomination reasons, but that it is currently restricted to moderators and admins. I propose that helpers should be allowed to use the command to update their own nominations as well.

It might also be worth considering allowing helpers to delete their own nominations (if this is not already possible), if they change their minds on a candidate.

ChrisLovering:
Closes #1617 

This change will allow helpers to run the edit reason command in the Talentpool cog.

To ensure that non-mod helpers can only edit their own reasons the interface for the reason command has been changed slightly.

If nominee_or_nomination_id is a member or user, then the command edits the currently active nomination for that person.
If it's an int, then use this to look up that nomination ID to edit.

If no nominator is specified, assume the invoker is editing their own nomination reason.
Otherwise, edit the reason from that specific nominator.

Raise a permission error if a non-mod staff member invokes this command on a
specific nomination ID, or with an nominator other than themselves.


bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -342,18 +342,40 @@ async def end_nomination_command(self, ctx: Context, user: MemberOrUser, *, reas
...
+            await ctx.send(f":x: Maximum allowed characters for the reason is {REASON_MAX_CHARS}.")
             return
+        if isinstance(target, Member):
+            if nomination := self.cache.get(target.id):
+                nomination_id = nomination["id"]
+            else:
+                await ctx.send("No active nomination found for that member.")
+                return
+        else:
+            nomination_id = target
```
Bluenix2, line 378: How does this work when multiple staff members nominate the same user?

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -342,18 +342,40 @@ async def end_nomination_command(self, ctx: Context, user: MemberOrUser, *, reas
...
+        await self._edit_nomination_reason(ctx, target=nomination_id, actor=actor, reason=reason)
+
+    async def _edit_nomination_reason(
+        self,
+        ctx: Context,
+        target: Union[int, Member],
+        actor: MemberOrUser,
+        reason: str,
+    ) -> None:
+        """Edits a nomination reason in the database after validating the input."""
```
Bluenix2, line 367:
```suggestion
        """Edit a nomination reason in the database after validating the input."""
```

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -342,18 +342,40 @@ async def end_nomination_command(self, ctx: Context, user: MemberOrUser, *, reas
...
     @nomination_edit_group.command(name='reason')
     @has_any_role(*MODERATION_ROLES)
     async def edit_reason_command(self, ctx: Context, nomination_id: int, actor: MemberOrUser, *, reason: str) -> None:
         """Edits the reason of a specific nominator in a specific active nomination."""
+        await self._edit_nomination_reason(ctx, target=nomination_id, actor=actor, reason=reason)
+
+    async def _edit_nomination_reason(
+        self,
+        ctx: Context,
+        target: Union[int, Member],
```
Bluenix2, line 363:
```suggestion
        ctx: Context,
        *,
        target: Union[int, Member],
```
Might as well make it kwarg-only if that is how you use it?

jchristgit (approved):
Yup. Yup. Yup. Yup. Yup.

bot/exts/recruitment/talentpool/_cog.py
ChrisLovering: The `actor` kwarg is what controls which nomination to edit.

IE If you nominated Akarys, I would set the target to Akarys and the actor to you to update your nomination.

This is how we ensure helpers can only edit their own nominations, since we hardcode actor to be ctx.author in the helper command

Bluenix2: But the lookup is done by target ID. Do nominations for the same user have the same ID?

ChrisLovering: Yea, that's right. Each nomination ID contains a list of nomination entries from different actors

mbaruh: Hmm wait, I don't get it.

- Each helper can have one nomination of someone
- A helper can only edit their own nomination of that person

So that means that a nomination is identifiable by the actor and the nominee.
So that removes the need for an ID.

ChrisLovering: Yea, helpers don't need to supply an ID when running !tp edit, since the ID is retrieved on lines 372-374.

My comments above were in an attempt to describe the structure

mbaruh (changes requested):
The implementation looks good on the technical side, but I find the command structure confusing.

It's not clear to me semantically why the command group is the one which accepts the narrower option for a target. In other commands, when a group accepts arguments, it's the opposite - meaning, it accepts the wider definition of a certain argument (e.g `infraction search`).

While this is easier to write as code, technically speaking it's not much more problematic to replicate the same behavior by letting helpers invoke the `edit reason` subcommand and raising the appropriate error if another actor is specified. Or, to make the help embed more friendly to helpers, to do the opposite of what you did, and make the group take the wider definition of a target and let helpers use `tp edit reason` specifically. Still, in that case the functionality of the group command in this way will seem odd to me as the group includes editing end reasons as well.

Also, something that is missing is the ability to raise the help embed for the group. Currently, you make the target mandatory, which raises an error even when you would expect only the help embed to show up.

edit: Thinking about the comment below, I'm probably more in favor of the first approach of keeping the group without arguments, as in the second approach they will see the group taking arguments but not be able to use them?

mbaruh:
Additionally, I'd argue we should allow the `tp` top command be invoked by helpers as well, so that they can see what command are available to them.

ChrisLovering:
I have change this back to one command, and change the interface slightly to allow for both helpers and mods to run the command, with validation being ran to ensure non-mod helpers can't edit other user's commands.

This also improves the interface for mods wanting to edit the reason for the own nomination of a user, by not needing them to look up the nomination ID.

wookie184 (approved):
We could limit this command to only be run in the `nominations` channel by helpers, although not really important. This all seems to work well 👍

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -342,18 +342,75 @@ async def end_nomination_command(self, ctx: Context, user: MemberOrUser, *, reas
...
+        actor: MemberOrUser,
+        reason: str,
+    ) -> None:
+        """Edit a nomination reason in the database after validating the input."""
         if len(reason) > REASON_MAX_CHARS:
-            await ctx.send(f":x: Maxiumum allowed characters for the reason is {REASON_MAX_CHARS}.")
+            await ctx.send(f":x: Maximum allowed characters for the reason is {REASON_MAX_CHARS}.")
             return
+        if isinstance(target, Member):
```
mbaruh, line 406: The command can resolve to a user. What happens if you try to nominate someone off-server?

ChrisLovering: It'll just work, I'll swap this around, so it check isinstance of int and then an else to catch Member & User

ChrisLovering: https://github.com/python-discord/bot/pull/1841/commits/d444ebd8c3f9b10e021ecc94da8bbac482a1b236

mbaruh (approved):
I'm not sure it makes sense to nominate someone outside the server at all, but that's out of scope for this PR. Looks good otherwise 👍

PR #1843, opened 2021-09-23
Update paste-site URL in `!paste`

### LINKED ISSUE #1842: Inconsistency with bot sending the paste-site
It was pointed out by a user via ModMail that the bot is inconsistent when sending the paste-site url. 

`!paste` uses `https://paste.pydis.com` (short), whereas the message sent when you upload a python/text file uses `https://paste.pythondiscord.com` (full).

In fact, from the looks of it all messages (other than `!paste`) which share the paste-site (such as `!e` when output is too long) use the longer form. Due to this, I think `!paste` should be the one changed (since only requires one change).

TizzySaurus:
Closes #1842 

`!paste` now uses `https://paste.pythondiscord.com` instead of `https://paste.pydis.com`.

### DIFF
--- a/bot/resources/tags/paste.md
+++ b/bot/resources/tags/paste.md
@@ -1,6 +1,6 @@
 **Pasting large amounts of code**
 
 If your code is too long to fit in a codeblock in discord, you can paste your code here:
-https://paste.pydis.com/
+https://paste.pythondiscord.com/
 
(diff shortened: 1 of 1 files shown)


Akarys42 (approved):
It was shorter, but I don't really have a strong opinion either way

PR #1844, opened 2021-09-23
Direct users to the appeals server when banned

ChrisLovering:
This is a new appeals process we are trialing. Users who get banned join this server and DM a modmail bot, who relays the message to the main server for mods to discuss.

I have updated the shortening logic to allow for extra information to be included at the end of the embed, while still staying under the limit.


bot/exts/moderation/infraction/_utils.py
```diff
@@ -27,16 +27,18 @@
...
+APPEAL_SERVER_INVITE = "https://discord.gg/WXrCJxWBnm"
 
 INFRACTION_TITLE = "Please review our rules"
-INFRACTION_APPEAL_EMAIL_FOOTER = f"To appeal this infraction, send an e-mail to {APPEAL_EMAIL}"
+INFRACTION_APPEAL_SERVER_FOOTER = f"\n\nTo appeal this infraction, join our [appeals server]({APPEAL_SERVER_INVITE})."
 INFRACTION_APPEAL_MODMAIL_FOOTER = (
-    'If you would like to discuss or appeal this infraction, '
-    'send a message to the ModMail bot'
+    '\n\nIf you would like to discuss or appeal this infraction, '
+    'send a message to the ModMail bot.'
```
TizzySaurus, line 36: Wouldn't this be the "Ban Appeals" bot rather than ModMail? Potentially also means constant should be renamed to `INFRACTION_BAN_APPEAL_BOT_FOOTER` or alike

bot/exts/moderation/infraction/_utils.py
```diff
@@ -170,8 +172,10 @@ async def notify_infraction(
     )
 
     # For case when other fields than reason is too long and this reach limit, then force-shorten string
-    if len(text) > 4096:
-        text = f"{text[:4093]}..."
+    if len(text) > 4096 - LONGEST_EXTRAS:
+        text = f"{text[:4093-LONGEST_EXTRAS]}..."
+
+    text += INFRACTION_APPEAL_SERVER_FOOTER if infr_type.lower() == 'ban' else INFRACTION_APPEAL_MODMAIL_FOOTER
```
TizzySaurus, line 178: Is there a reason this is now being sent in the description rather than footer?

tests/bot/exts/moderation/infraction/test_utils.py
```diff
@@ -139,14 +139,14 @@ async def test_notify_infraction(self, send_private_embed_mock):
                         type="Ban",
                         expires="2020-02-26 09:20 (23 hours and 59 minutes)",
                         reason="No reason provided."
-                    ),
+                    ) + utils.INFRACTION_APPEAL_SERVER_FOOTER,
```
TizzySaurus, line 142: Likewise, is there a reason all the tests now have this under the description rather than footer?

TizzySaurus (commented):
Seems pretty solid so far, just 2 small things

bot/exts/moderation/infraction/_utils.py
ChrisLovering: `INFRACTION_APPEAL_SERVER_FOOTER` is used for bans, `INFRACTION_APPEAL_MODMAIL_FOOTER` is used for everything else.

The reason for this is because banned used cannot DM modmail, as they do not share a server, so they need to join the appeal server to be able to. Whereas other kinds of infractions don't get kicked from the server, so they have access to DM modmail directly.

bot/exts/moderation/infraction/_utils.py
ChrisLovering: This is to allow hyperlinking the guild invite

tests/bot/exts/moderation/infraction/test_utils.py
ChrisLovering: The message isn't in the footer, so it the tests tested that it was, they would fail.

ChrisLovering:
In draft until we have a modmail plugin to control pings.

bot/exts/moderation/infraction/_utils.py
TizzySaurus: Do we currently handle the situation for kicks? Whilst it's rare that we kick instead of mute/ban someone, it does sometimes happen and so this should be handled. Presumably we'd want a different message saying "rejoin when you're ready to follow our rules and coc" or something.

ChrisLovering: Kicks currently also send the `INFRACTION_APPEAL_MODMAIL_FOOTER` we could have another message that explains it more along the lines that you suggest. But that's out of scope for this PR, so should be another issue/PR.

PR #1845, opened 2021-09-24
Handle edge case of `message.author` being a `discord.User` when claiming a help channel.

### LINKED ISSUE #1839: AttributeError: 'User' object has no attribute 'add_roles'
Sentry Issue: [BOT-1N6](https://sentry.io/organizations/python-discord/issues/2661278228/?referrer=github_integration)

```
AttributeError: 'User' object has no attribute 'add_roles'
...
  File "bot/exts/help_channels/_cog.py", line 477, in on_message
    await self.claim_channel(message)
  File "bot/utils/lock.py", line 108, in wrapper
    except discord.Forbidden:
  File "bot/utils/lock.py", line 108, in wrapper
    except discord.Forbidden:
  File "bot/utils/lock.py", line 108, in wrapper
    except discord.Forbidden:
  File "bot/exts/help_channels/_cog.py", line 128, in claim_channel
    await self._handle_role_change(message.author, message.author.add_roles)

```

This happened as the user was banned from the server in time time between sending a message to open the help channel, and this line of code being hit.

TizzySaurus:
Closes #1839.

Adds an `isinstance` check before trying to add cooldown role and DM, to see if the `message.author` is a `discord.User`/`discord.Member`.

Will cause a sentry warning when `message.author` is a `discord.User`, in hopes of helping debug the issue in future (cause of `message.author` being a `discord.User` is unknown -- some theories are posted in #1839. As @ChrisLovering said on discord, the level can be lowered if it becomes too noisy.

### DIFF
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -125,14 +125,19 @@ class HelpChannels(commands.Cog):
         """
         log.info(f"Channel #{message.channel} was claimed by `{message.author.id}`.")
         await self.move_to_in_use(message.channel)
-        await self._handle_role_change(message.author, message.author.add_roles)
+
+        # Handle odd edge case of `message.author` being a `discord.User` (see bot#1839)
+        if isinstance(message.author, discord.User):
+            log.warning("`message.author` is a `discord.User` so not handling role change or sending DM.")
+        else:
+            await self._handle_role_change(message.author, message.author.add_roles)
+
+            try:
+                await _message.dm_on_open(message)
+            except Exception as e:
+                log.warning("Error occurred while sending DM:", exc_info=e)
 
         await _message.pin(message)
-
-        try:
-            await _message.dm_on_open(message)
-        except Exception as e:
-            log.warning("Error occurred while sending DM:", exc_info=e)
 
         # Add user with channel for dormant check.
         await _caches.claimants.set(message.channel.id, message.author.id)


MarkKoz:
Did you go over other possible failure points for this in the cog? It is actually documented that `discord.Message.author` might be a `discord.User` if they left the guild. Maybe start by looking for other uses of that attribute in the help system. Are there any other member attributes that could be a user too?

bot/exts/help_channels/_cog.py
```diff
@@ -125,14 +125,19 @@ async def claim_channel(self, message: discord.Message) -> None:
         """
         log.info(f"Channel #{message.channel} was claimed by `{message.author.id}`.")
         await self.move_to_in_use(message.channel)
-        await self._handle_role_change(message.author, message.author.add_roles)
 
-        await _message.pin(message)
+        # Handle odd edge case of `message.author` being a `discord.User` (see bot#1839)
+        if isinstance(message.author, discord.User):
```
MarkKoz, line 130: Do you think it would make more sense to invert this i.e. check for the type it supports rather than checking for the one it can't support? Technically the author could be _any_ `abc.User`.

TizzySaurus: I did the `User` check first since then the comment directly correlates to the first statement. It doesn't really matter which way round the statement is, and afaik the only possible subclassess of `abc.User` is `User` and `Member` (can't be `ClientUser`, even when message is sent by the bot).

TizzySaurus:
> Did you go over other possible failure points for this in the cog? It is actually documented that `discord.Message.author` might be a `discord.User` if they left the guild. Maybe start by looking for other uses of that attribute in the help system. Are there any other member attributes that could be a user too?

There isn't any other cases where we rely on `message.author` being a `discord.Member`. Other cases just get the `.id`, which will exist regardless. Only exception is L593 of `_cog.py` which is `message.author.send()`. Depending on the cause of the author being `discord.User`, this may or may not be an issue (if it's just caused by cache, then there's still a mutual server, so we can send a message. If it's caused by user leaving, then there's no mutual server, and we can't send a message. The thing is this is handled either way already so not really an issue -- although I guess you could argue we can save an API call)

PR #1854, opened 2021-10-04
Use utcnow() and lower a logging level in patch_typing

ChrisLovering:


wookie184:
For context the reason for the now -> utcnow change is explained here: https://github.com/python-discord/sir-lancebot/pull/892#issuecomment-933795156

wookie184 (approved):
LGTM

Bluenix2 (approved):
Of course we don't get 403s anymore, but this should work considering you changed the time both where it is read and where it is set.

jchristgit (approved):
Thanks, Chris. This is rock solid.

PR #1855, opened 2021-10-05
Improve anti-spam logs

### LINKED ISSUE #1850: Always create a log if filtered messages have attachments.
Currently, when creating a mod alert, the bot will not create a log of the messages if there is only one message involved, and will display it in the alert embed directly instead.

That should not be the case when there are attachements. If there are any, it should always create a new log and link to it instead of trying to display the contents of the message (which might not exist).

Example of an alert with no content (because it's a single message with only attachments):
image

How it looks like when it's several messages:
image

TizzySaurus:
Closes #1850.

Messages triggering antispam which have attachments will now *always* be uploaded to the log site, and messages that are truncated will now be uploaded to the paste site as well as showing the truncated version of the message.

### DIFF
--- a/bot/exts/filters/antispam.py
+++ b/bot/exts/filters/antispam.py
@@ -85,7 +85,7 @@ class DeletionContext:
         # For multiple messages or those with excessive newlines, use the logs API
         if any((
             len(self.messages) > 1,
-            self.messages[0].attachments,
+            len(self.messages[0].attachments) > 0,
             self.messages[0].count('\n') > 15
         )):
             url = await modlog.upload_log(self.messages.values(), actor_id, self.attachments)
--- a/bot/exts/filters/antispam.py
+++ b/bot/exts/filters/antispam.py
@@ -46,6 +46,7 @@ RULE_FUNCTION_MAPPING = {
 @dataclass
 class DeletionContext:
     """Represents a Deletion Context for a single spam event."""
+
     members: frozenset[Member]
     triggered_in: TextChannel
     channels: set[TextChannel] = field(default_factory=set)
--- a/bot/exts/filters/antispam.py
+++ b/bot/exts/filters/antispam.py
@@ -46,7 +46,6 @@ RULE_FUNCTION_MAPPING = {
 @dataclass
 class DeletionContext:
     """Represents a Deletion Context for a single spam event."""
-
     members: frozenset[Member]
     triggered_in: TextChannel
     channels: set[TextChannel] = field(default_factory=set)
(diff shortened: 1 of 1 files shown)


bot/exts/filters/antispam.py
```diff
@@ -83,7 +83,11 @@ async def upload_messages(self, actor_id: int, modlog: ModLog) -> None:
         )
 
         # For multiple messages or those with excessive newlines, use the logs API
-        if len(self.messages) > 1 or 'newlines' in self.rules:
+        if any((
+            len(self.messages) > 1,
+            self.messages[0].attachments,
```
Akarys42, line 88:
```suggestion
            len(self.messages[0].attachments) > 0,
```

Akarys42 (changes requested):
I'd prefer if that check would explicitly use `len()`

bot/exts/filters/antispam.py
TizzySaurus: Fair enough, have committed :+1:

bot/exts/filters/antispam.py
```diff
@@ -83,7 +83,11 @@ async def upload_messages(self, actor_id: int, modlog: ModLog) -> None:
         )
 
         # For multiple messages or those with excessive newlines, use the logs API
-        if len(self.messages) > 1 or 'newlines' in self.rules:
+        if any((
+            len(self.messages) > 1,
+            len(self.messages[0].attachments) > 0,
+            self.messages[0].count('\n') > 15
+        )):
```
wookie184, line 90: `self.messages` is a dictionary mapping message IDs to message objects, so this doesn't work.
I get this error:
```
bot_1        | 2021-10-05 12:13:33 | bot.utils.scheduling | ERROR | Error in task AntiSpam._process_deletion_context(frozenset({<Member id=493839819168808962 name='wookie' discriminator='4017' bot=False nick=None guild=<Guild id=813116921646022716 name="wookie's test server" shard_id=None chunked=True member_count=4>>})) 139705866738800!
bot_1        | Traceback (most recent call last):
bot_1        |   File "/bot/bot/exts/filters/antispam.py", line 299, in _process_deletion_context
bot_1        |     await deletion_context.upload_messages(self.bot.user.id, self.mod_log)
bot_1        |   File "/bot/bot/exts/filters/antispam.py", line 89, in upload_messages
bot_1        |     len(self.messages[0].attachments) > 0,
bot_1        | KeyError: 0
```

wookie184 (changes requested):

wookie184:
looks good other than the linting error

wookie184 (approved):
LGTM

Akarys42 (approved):
Lgtm

PR #1856, opened 2021-10-05
Add redirect message to `!defcon threshold`.

TizzySaurus:
`!defcon threshold` now uses the `@redirect_output` decorator in order to provide a redirect-messsage when command is invoked outside of #defcon.


wookie184 (approved):
I don't really mind as this is a pretty minor thing anyway, but I think just limiting the command to the defcon channel would be a bit nicer. Will approve anyway though as this works I don't think this is something that will be happen much.

TizzySaurus:
> I don't really mind as this is a pretty minor thing anyway, but I think just limiting the command to the defcon channel would be a bit nicer. Will approve anyway though as this works I don't think this is something that will be happen much.

As discussed on discord, this has been implemented :+1:

TizzySaurus:
This has been changed again, this time decision was to simply have the message relay to both `ctx.channel` and #defcon.

bot/exts/moderation/defcon.py
```diff
@@ -208,7 +208,13 @@ def _update_channel_topic(self) -> None:
         scheduling.create_task(self.channel.edit(topic=new_topic))
 
     @defcon_settings.atomic_transaction
-    async def _update_threshold(self, author: User, threshold: relativedelta, expiry: Optional[Expiry] = None) -> None:
+    async def _update_threshold(
+        self,
+        ctx: Context,
+        author: User, threshold:
```
wookie184, line 214: I think `threshold` was meant to be on the next line here :)

Also thoughts on removing the `author` parameter since we pass in ctx now? I can't think we'd ever want to do an updating threshold on someone elses behalf or something like that lol, reckon it'd always be `ctx.author`

TizzySaurus: > I think `threshold` was meant to be on the next line here :)

It was indeed, thank you.

> Also thoughts on removing the `author` parameter since we pass in ctx now?

Yeah, that seems fair enough.

Will commit both of these now :+1:

TizzySaurus:
Fixed the issues with this and squashed down to one commit.

Should be ready for re-review @wookie184 :+1:

wookie184 (approved):
LGTM, thanks :D

PR #1860, opened 2021-10-06
Attempt infraction before DMing

### LINKED ISSUE #1805: Apply infraction before DMing the user

TizzySaurus:
Closes #1805 

A fix for #1806.

This time it won't DM when it's not supposed to :p

### DIFF
--- a/bot/exts/moderation/infraction/_scheduler.py
+++ b/bot/exts/moderation/infraction/_scheduler.py
@@ -149,8 +149,8 @@ class InfractionScheduler:
         else:
             expiry_msg = f" until {expiry}" if expiry else " permanently"
 
-        dm_result = constants.Emojis.failmail
-        dm_log_text = "\nDM: **Failed**"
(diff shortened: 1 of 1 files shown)


TizzySaurus:
> WIth the latest changes, some infraction types will send messages twice. Warns, for instance, get sent in the original block, then get sent again in the modified block.
> 
> We can get around this by adding a conditional opoosite to the `not in (ban, kick)` from the new block, but perhaps we should instead rework this so it's a little less confusing?

For some reason the second part of the if statement was missing. `if not infraction["hidden"]:` should be `if not infraction["hidden"] and infr_type in {"ban", "kick"}`.

That way, only bans and kicks send DMs before trying to apply the infraction (since we can't send after applying as there will be no mutual servers).

PR #1889, opened 2021-10-18
feat: added url parsing to the filter

### LINKED ISSUE #1260: Adding special logic when matching domains (bar.com matches foobar.com)

Kronifer:
Closes #1260

Added some url parsing using urllib.parse to stop false positives. to remove false positives, I parsed the URLs and checked if the netloc was the same. If it wasn't, I passed. I also check if the netloc is prefixed with www. or any other subdomain should people try to circumvent the filter. An example false positive would be deliciouscookies.com triggering cookies.com, a blacklisted url (not linking actual example cause it's NSFW)

### DIFF
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -2,6 +2,8 @@ import asyncio
 import re
 import unicodedata
 from datetime import timedelta
+import urllib.parse
+from datetime import timedelta
 from typing import Any, Dict, List, Mapping, NamedTuple, Optional, Tuple, Union
 
 import arrow
@@ -534,8 +536,14 @@ class Filtering(Cog):
         domain_blacklist = self._get_filterlist_items("domain_name", allowed=False)
         for match in URL_RE.finditer(text):
             for url in domain_blacklist:
-                if url.lower() in match.group(1).lower():
-                    return True, self._get_filterlist_value("domain_name", url, allowed=False)["comment"]
+                blacklisted_parsed = urllib.parse.urlparse(url)
+                url_parsed = urllib.parse.urlparse(match.group(1).lower())
+                if blacklisted_parsed.netloc != "":
+                    if url_parsed.netloc in (f"www.{blacklisted_parsed.netloc}", blacklisted_parsed.netloc):
+                        return True, self._get_filterlist_value("domain_name", url, allowed=False)["comment"]
+                else:
+                    if url_parsed.netloc in (f"www.{blacklisted_parsed.path}", blacklisted_parsed.path):
+                        return True, self._get_filterlist_value("domain_name", url, allowed=False)["comment"]
         return False, None
 
     @staticmethod
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -2,14 +2,13 @@ import asyncio
 import re
 import unicodedata
 from datetime import timedelta
-import urllib.parse
-from datetime import timedelta
 from typing import Any, Dict, List, Mapping, NamedTuple, Optional, Tuple, Union
 
 import arrow
 import dateutil.parser
 import discord.errors
 import regex
+import tldextract
 from async_rediscache import RedisCache
 from dateutil.relativedelta import relativedelta
 from discord import Colour, HTTPException, Member, Message, NotFound, TextChannel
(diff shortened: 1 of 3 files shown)


Kronifer:
> Can you provide examples of false positives? Is there a related issue? 

@mbaruh yes, I'll link the issue and provide examples in the initial description

bot/exts/filters/filtering.py
```diff
@@ -479,7 +480,11 @@ async def _has_urls(self, text: str) -> Tuple[bool, Optional[str]]:
         domain_blacklist = self._get_filterlist_items("domain_name", allowed=False)
         for match in URL_RE.finditer(text):
             for url in domain_blacklist:
-                if url.lower() in match.group(1).lower():
+                blacklisted_parsed = urllib.parse.urlparse(f"https://{url}")
```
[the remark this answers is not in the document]
onerandomusername: note that yarl is open for usage too

Yarl is a direct dependency of aiohttp and is already checked in and used by aiohttp, so if you wanted to use it here, i can't see why not.

bot/exts/filters/filtering.py
```diff
@@ -480,7 +481,9 @@ async def _has_urls(self, text: str) -> Tuple[bool, Optional[str]]:
         domain_blacklist = self._get_filterlist_items("domain_name", allowed=False)
         for match in URL_RE.finditer(text):
             for url in domain_blacklist:
-                if url.lower() in match.group(1).lower():
+                blacklisted_parsed = tldextract.extract(url)
+                url_parsed = tldextract.extract(match.group(1).lower())
```
onerandomusername: line 485 can be before line 483

onerandomusername (changes requested):
As a heads up, poetry.lock got all of the dependencies updated.

To fix this, please be sure you are using poetry 1.1.x 

First revert poetry.lock locally.

Ensure that pyproject.toml is the same, with tldextract in it.

Next, use `poetry lock --no-update`
This will relock poetry without updating all of the dependencies.


However, it may be worth an update seperately to the dependencies, as this updated redis, rapidfuzz, sentry, etc

bot/exts/filters/filtering.py
onerandomusername: how is this already outdated?

bot/exts/filters/filtering.py
```diff
@@ -481,10 +481,11 @@ async def _has_urls(self, text: str) -> Tuple[bool, Optional[str]]:
         domain_blacklist = self._get_filterlist_items("domain_name", allowed=False)
         for match in URL_RE.finditer(text):
             for url in domain_blacklist:
-                blacklisted_parsed = tldextract.extract(url)
-                url_parsed = tldextract.extract(match.group(1).lower())
-                if blacklisted_parsed.registered_domain == url_parsed.registered_domain:
-                    return True, self._get_filterlist_value("domain_name", url, allowed=False)["comment"]
+                if url.lower() in match.group(1).lower():
+                    blacklisted_parsed = tldextract.extract(url.lower())
+                    url_parsed = tldextract.extract(match.group(1).lower())
```
onerandomusername, line 486: L486 can be before L483

onerandomusername (changes requested):

bot/exts/filters/filtering.py
Kronifer: i dont see the need of creating the extra variable unless we know we'll need it.

Kronifer:
> As a heads up, poetry.lock got all of the dependencies updated.
> 
> To fix this, please be sure you are using poetry 1.1.x
> 
> First revert poetry.lock locally.
> 
> Ensure that pyproject.toml is the same, with tldextract in it.
> 
> Next, use `poetry lock --no-update` This will relock poetry without updating all of the dependencies.
> 
> However, it may be worth an update seperately to the dependencies, as this updated redis, rapidfuzz, sentry, etc

this has been deemed not to be a problem as of https://discord.com/channels/267624335836053506/635950537262759947/916168658651324487

onerandomusername:
> this has been deemed not to be a problem as of https://discord.com/channels/267624335836053506/635950537262759947/916168658651324487


After looking into this further, markdownify cannot be updated, as it will nerf the results of the doc command.

onerandomusername:
Fixed that change in GH-2014 👍

onerandomusername (approved):
forgot to come back, approved now

bot/exts/filters/filtering.py
```diff
@@ -481,7 +482,10 @@ async def _has_urls(self, text: str) -> Tuple[bool, Optional[str]]:
         for match in URL_RE.finditer(text):
             for url in domain_blacklist:
                 if url.lower() in match.group(1).lower():
```
mbaruh, line 484: I'd save `url.lower()` and `match.group(1).lower()` into separate variables just because those values are used a couple of times each, but it's not a big deal here.

mbaruh (approved):
Looks great and seems to be working. Thanks!

ChrisLovering (approved):
lgtm :D

Akarys42 (approved):
Proxy approval

PR #1892, opened 2021-10-18
Update username checks

### LINKED ISSUE #1891: Update Username Checks
Currently, we check people's usernames when they send a message on the server, to ensure it isn't inappropriate. I think we should actually check in member join and member update events, as people can still have their nickname be visible on server without ever sending a message. They can also update their usernames after sending a message.

If we do that, we'll also no longer need the check in the message event, just the other two. It should be a relatively straightforward swap I hope.

I'd also like to propose we add something like [confusables](https://pypi.org/project/confusables/), to help deal with people who use ASCII art or slight variations to get around our filters. We've used it before for this purpose using temporary patches, with decent success.

ChrisLovering:
Closes #1891

Currently, we check people's usernames when they send a message on the server, to ensure it isn't inappropriate. To get more coverage, we should actually be checking member join and member update events, as people can still have their nickname be visible on server without ever sending a message. They can also update their usernames after sending a message.

I have also:
- Added a check for the normalised version of the name too, using unicodedata.normalize, to ensure unicode versions of names don't bypass our filters.
- Returned the first bad match we find, rather than collecting them all, so we don't run it against *all* of our filters if not needed. If a single bad match is found, I trust our mods to actually read the user's name
- Moved the `check_send_alert()` call to the first thing we do, as a redis call is cheaper than checking all our filter tokens

### DIFF
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -237,10 +237,14 @@ class Filtering(Cog):
         """Send a mod alert every 3 days if a username still matches a watchlist pattern."""
         # Use lock to avoid race conditions
         async with self.name_lock:
+            # Check if we recently alerted about this user first,
+            # to avoid running all the filter tokens against their name again.
+            if not await self.check_send_alert(member):
+                return
+
             # Check whether the users display name contains any words in our blacklist
             matches = self.get_name_matches(member.display_name)
-
-            if not matches or not await self.check_send_alert(member):
+            if not matches:
                 return
 
             log.info(f"Sending bad nickname alert for '{member.display_name}' ({member.id}).")
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -1,8 +1,8 @@
 import asyncio
 import re
+import unicodedata
 from datetime import timedelta
 from typing import Any, Dict, List, Mapping, NamedTuple, Optional, Tuple, Union
-from unicodedata import normalize
 
 import arrow
 import dateutil.parser
@@ -208,11 +208,12 @@ class Filtering(Cog):
 
     def get_name_match(self, name: str) -> Optional[re.Match]:
         """Check bad words from passed string (name). Return the first match found."""
-        normalised_name = normalize("NFKC", name)
+        normalised_name = unicodedata.normalize("NFKC", name)
+        cleaned_normalised_name = "".join(c for c in normalised_name if not unicodedata.combining(c))
 
-        # Run filters against normalized and original version,
+        # Run filters against normalised, cleaned normalised and the original name,
         # in case we have filters for one but not the other.
-        names_to_check = (name, normalised_name)
+        names_to_check = (name, normalised_name, cleaned_normalised_name)
 
         watchlist_patterns = self._get_filterlist_items('filter_token', allowed=False)
         for pattern in watchlist_patterns:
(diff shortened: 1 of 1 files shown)


wookie184:
Normalizing strings before checking if they match on all filters removes our ability to match specific unnormalized forms. I guess this should be ok, although we should look into our current filters and see what can no longer match. I'm not sure how aggressive the normalization by confusables is.

bot/exts/filters/filtering.py
```diff
@@ -157,6 +158,17 @@ def mod_log(self) -> ModLog:
...
 
+    @Cog.listener()
+    async def on_member_join(self, member: Member) -> None:
+        """Check if the joining member has a bad word in their name."""
+        await self.check_bad_words_in_name(member)
+
+    @Cog.listener()
+    async def on_member_update(self, before: Member, after: Member) -> None:
+        """If the member has updated their name, check if it has a bad word in it."""
+        if before.display_name != after.display_name:
```
Bluenix2, line 169: When the member isn't in cache `before is after` will be True afaik. Do we want to run the check if there's a possiblity that the name or nickname was changed? This isn't very expensive right?

bot/exts/filters/filtering.py
```diff
@@ -179,15 +191,21 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
...
         await self._filter_message(after, delta)
 
-    def get_name_matches(self, name: str) -> List[re.Match]:
-        """Check bad words from passed string (name). Return list of matches."""
-        name = self.clean_input(name)
-        matches = []
+    def get_name_match(self, name: str) -> Optional[re.Match]:
+        """Check bad words from passed string (name). Return the first match found."""
+        normalised_name = unicodedata.normalize("NFKC", name)
+        cleaned_normalised_name = "".join(c for c in normalised_name if not unicodedata.combining(c))
```
Bluenix2, line 197:
```suggestion
        cleaned_normalised_name = "".join([c for c in normalised_name if not unicodedata.combining(c)])
```
List comprehension for that Sonic the hedgehog speed?

onerandomusername (approved):
this looks good to me, keep up the moderation tooling! :shipit: 

(side: I agree with bluenix's comments)

bot/exts/filters/filtering.py
jchristgit: Isn't not building an intermediate list faster?

jchristgit (approved):
Nice use of `unicodedata` 👍

bot/exts/filters/filtering.py
ChrisLovering: Apparently not, just tested in repl, giving str.join a list is about 10-20% faster that giving it a generator

Bluenix2: List comprehensions are faster yes. Without the brackets you are passing an iterable (generator in this case) which Python needs to create, then individually advance and resize the list.

`list(x for x in y)` is not the same as `[x for x in y]`.

That kind of stuff adds up.

bot/exts/filters/filtering.py
ChrisLovering: following a discussion in a mod channel, we decided we don't want to alert on member join/edit, as those members may never have even spoken in our community, so it's not a burden we want to put on the mod team.

bot/exts/filters/filtering.py
ChrisLovering: https://github.com/python-discord/bot/pull/1892/commits/16f389efa239f5847b1553c461bf19725f1f33bd

PR #1910, opened 2021-10-25
Consider parent channels when checking mod channels

ChrisLovering:
This allows commands that check if a channel is a mod channel to also work in threads of that channel


wookie184 (approved):
Works nicely.

PR #1917, opened 2021-10-27
Patreon

### LINKED ISSUE #1765: Shoutout our patreon supporters!
In a recent staff meeting we agreed that we weren't giving enough love to our patreon supporters.

As such, we have agreed that we should have the bot post in #community-meta every time someone new receives one of the patreon roles, along with posting once a month (on the 1st, just after the patreon subscription cycle) with a full list of all of our supporters.

The post when we get a new supporter should include:
 - A mention for the user who has started supporting
 - What tier they are supporting us
 - A link to our patreon page

The monthly post should include:
- A 3 column embed (`embed.add_field()` makes this easy), one for each tier
- The names of all of our current supporters, under the appropriate tier column
- A link to our patreon page

mathstrains21:
Closes #1765 

First off, this is my first pull request so please give me general pull request feedback if you have any :-)

Some decisions have been made on discord. A search of `bot#1765` should find you all the conversations.

#### Some changes to make to test:
- Create 3 new roles, with clear different colours so you can see colours on embeds!
- Put these in your config.yml

There are 2 new commands: `!patrons` and `!patreon`. These send the list of patrons and some patreon info respectively.

#### Logging
I have not included an logging. Please provide feedback about logging.


TizzySaurus (dismissed):
Thanks for the hardwork you've put into this PR!

A few small things in my preliminary review.

wookie184 (dismissed):
I've only tested it quickly so this is mainly a code review (although it seems to work nicely, I like the embeds!)

Firstly, thanks for the PR! The structure of the code seems good and it's been written well.

My main comments here are about repetition, there are couple of places where I think using loops would neaten the code up a bit, making it a bit more concise and extensible. That said, these aren't required changes, i've only mentioned them as something to think about if you want to.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,160 @@
...
+        await self.bot.wait_until_guild_available()
+
+        guild = self.bot.get_guild(constants.Guild.id)
+
+        tier_1_patrons = set(guild.get_role(constants.Roles.patreon_tier_1).members)
+        tier_2_patrons = set(guild.get_role(constants.Roles.patreon_tier_2).members)
+        tier_3_patrons = set(guild.get_role(constants.Roles.patreon_tier_3).members)
+
+        tier_1_patrons = tier_1_patrons - tier_2_patrons - tier_3_patrons
+        tier_2_patrons = tier_2_patrons - tier_3_patrons
```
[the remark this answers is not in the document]
Bluenix2: I agree with Wookie here, with a small comment that basically clarifies what that last line does this is more maintainable.

Something like "Remove members who's already in the 'tier_members' list and have a higher role".. maybe?

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,160 @@
...
+            colour = patreon_tier_3_role.colour
+        elif patreon_tier_2_role in after.roles:
+            new_patreon_tier = 2
+            colour = patreon_tier_2_role.colour
+        elif patreon_tier_1_role in after.roles:
+            new_patreon_tier = 1
+            colour = patreon_tier_1_role.colour
+
+        if not new_patreon_tier > current_patreon_tier:
+            return
```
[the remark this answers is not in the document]
Bluenix2: Will this work correctly (both semantically doing what you want to - and not breaking) considering the `enumerate()` is inside the `reversed()`?

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,163 @@
+import datetime
+import logging
+
+import discord
+from discord.ext import commands, tasks
+
+from bot import constants
+from bot.bot import Bot
+
+log = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
Bluenix2: I think it's still good to keep, since `print()` doesn't work at all if you want to debug something.

config-default.yml
```diff
@@ -291,6 +291,11 @@ guild:
         # Streaming
         video:          764245844798079016
 
+        # Patreon
+        patreon_tier_1: 505040943800516611
+        patreon_tier_2: 743399725914390631
+        patreon_tier_3: 743400204367036520
```
Bluenix2, line 297: Aren't these supposed to be the IDs from the actual Python Discord server?

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,160 @@
...
+            embed_tier_3 = discord.Embed(
+                title="Tier 3 patrons",
+                description="\n".join(tier_3_patrons),
+                colour=guild.get_role(constants.Roles.patreon_tier_3).colour
+            )
+            embed_list.append(embed_tier_3)
+
+        await channel.send(embeds=embed_list)
+
+    @commands.command("patrons")
```
Bluenix2, line 129: There's no check on this, anybody can send this anywhere.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @commands.command("patreon")
+    async def patreon_info(self, ctx: commands.Context) -> None:
+        """Sends patreon info."""
+        await ctx.send(embed=discord.Embed(
+            title="Patreon",
+            description=(
+                "We use the money from Patreon to offer excellent prizes for all of our events. Prizes like t-shirts, "
+                "stickers, microcontrollers that support CircuitPython, or maybe even a mechanical keyboard.\n\nYou can"
+                " read more about how Patreon supports us, or even support us yourself, on our Patreon page "
```
Bluenix2, line 152:
```suggestion
                "stickers, microcontrollers that support CircuitPython, or maybe even a mechanical keyboard.\n\n"

                "You can" read more about how Patreon supports us, or even support us yourself, on our Patreon page "
```

I like to refect empty lines in the code.

Bluenix2, line 152: This is kind of misleading, Patreon the company doesn't support us.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,160 @@
...
+        date = datetime.date.today().day
+
+        if date == 1:
+            await self.send_current_supporters(
+                utils.channels.get_or_fetch_channel(constants.Channels.meta)
+            )
+
+    @commands.command("patreon")
+    async def patreon_info(self, ctx: commands.Context) -> None:
+        """Sends patreon info."""
```
Bluenix2, line 146:
```suggestion
        """Send information about how Python Discord uses Patreon."""
```

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,160 @@
...
+        sending_channel = utils.channel.get_or_fetch_channel(constants.Channels.meta)
+
+        await sending_channel.send(
+            embed=discord.Embed(
+                description=message,
+                colour=colour
+            )
+        )
+
+    async def send_current_supporters(self, channel: discord.TextChannel) -> None:
```
Bluenix2, line 72: Yeah I really would like to see some of the suggestions brought up by Wookie implemented.

Even though it may result in more complicated code, that code will take less lines and as such be easier to read (on top of looking cleaner).

bot/exts/info/patreon.py
mathstrains21: How would I let admins (or whatever role we want) send it in community meta, while the rest of the community can only send in the bot command channels?

config-default.yml
mathstrains21: assuming the ones that chris posted in #1765 are correct, then these are correct

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,162 @@
...
+        """Sends the current list of patreon supporters, sorted by tier level."""
+        await self.send_current_supporters(ctx.channel)
+
+    @tasks.loop(time=datetime.time(hour=17))
+    async def current_monthly_supporters(self) -> None:
+        """A loop running daily to see if it's the first of the month. If so call `self.send_current_supporters()`."""
+        date = datetime.date.today().day
+
+        if date == 1:
+            await self.send_current_supporters(
```
onerandomusername, line 141:
```py
if date != 1:
  return
```
This has the benefit of returning early/guarding ifs, which are easier to read.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+        )
+
+        embed_list.append(embed_patron_info)
+
+        if tier_1_patrons:
+            embed_tier_1 = discord.Embed(
+                title="Tier 1 patrons",
+                description="\n".join(tier_1_patrons),
+                colour=guild.get_role(constants.Roles.patreon_tier_1).colour
```
onerandomusername, line 108: these roles should be an instance attribute.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+
+    async def send_current_supporters(self, channel: discord.TextChannel) -> None:
+        """Send the current list of patreon supporters, sorted by tier level."""
+        await self.bot.wait_until_guild_available()
+
+        guild = self.bot.get_guild(constants.Guild.id)
+
+        tier_1_or_above_patrons = set(guild.get_role(constants.Roles.patreon_tier_1).members)
+        tier_2_or_above_patrons = set(guild.get_role(constants.Roles.patreon_tier_2).members)
```
onerandomusername, line 79: I like using a set here, since it favours no patreon over another, it effectively randomly sorts them.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,162 @@
...
+                title="Tier 3 patrons",
+                description="\n".join(tier_3_patrons),
+                colour=guild.get_role(constants.Roles.patreon_tier_3).colour
+            )
+            embed_list.append(embed_tier_3)
+
+        await channel.send(embeds=embed_list)
+
+    @commands.command("patrons")
+    async def current_supporters_command(self, ctx: commands.Context) -> None:
```
onerandomusername, line 131: this command name of `patrons` vs `patreon` for the below command, will likely be a source of confusion. A different way to go about this may be making the below informational command a command *group*, and making the list a subcommand. 

The docs for doing so are located here: https://discordpy.readthedocs.io/en/master/ext/commands/api.html#group

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+            embed_list.append(embed_tier_2)
+
+        if tier_3_patrons:
+            embed_tier_3 = discord.Embed(
+                title="Tier 3 patrons",
+                description="\n".join(tier_3_patrons),
+                colour=guild.get_role(constants.Roles.patreon_tier_3).colour
+            )
+            embed_list.append(embed_tier_3)
```
onerandomusername, line 126: I was thinking about a way to do this, and think I've come up with a way to do it.

additionally, I think that tier 3 should be before tier 2 and 1, as tier 3 is above tier 2.

Using a dictionary we can match the roles to the members, which ends up like this:
*note: this builds off of storing the patreon roles in a list*
```py
embeds = []
for num, role in reversed(enumerate(patreon_role_list)):
    if not role.members:
       # skipping this role since there are no members
       continue

    # right here the description could be formulated, instead of above

    embeds.append(discord.Embed(
                title=f"Tier {num + 1} patrons",
                description="\n".join(role.members),
                colour=role.colour
            )
    )

if not embeds:
   # there's no embeds, so there's no embed to send
   # yes, this is an extreme edge case to not have any embeds, but should be prepared for it anyhow.

```

onerandomusername (changes requested):
This is really good, honestly, for a first pr, keep it up /gen

bot/exts/info/patreon.py
mathstrains21: Most certainly not the intended reason for this set usage :-)

mathstrains21:
> @mathstrains21 will you be continuing this PR?

Hi! I will be, I just need to find the time to do so! Thanks for the reminder though!

Xithrius:
Thanks for the quick update! Have a good one!

Xithrius:
@mathstrains21 Hey, it's been a while. What's your status on this PR?

Xithrius:
@mathstrains21 how's it going?

wookie184:
I'll finish this one off

wookie184:
will have to open a new PR I think

mathstrains21:
I don't have time to continue, but would you like access to my repo to keep it in one PR?

wookie184:
> I don't have time to continue right now, but would you like access to my repo to keep it in one PR?

That would be helpful, thanks!

bot/exts/info/patreon.py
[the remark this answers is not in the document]
wookie184: Changed, I might add some logging but doesn't hurt to leave it regardless

bot/exts/info/patreon.py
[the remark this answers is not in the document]
wookie184: Turns out it didn't work, I went for a similar approach that does

bot/exts/info/patreon.py
wookie184: Changed, i'd be happy for any more feedback on the wording though, probably still room for improvement.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,162 @@
...
+
+        if not new_patreon_tier > current_patreon_tier:
+            return
+
+        message = (
+            f":tada: {after.mention} just became a **tier {new_patreon_tier}** patron!\n"
+            "[Support us on Patreon](https://pydis.com/patreon)"
+        )
+
+        sending_channel = utils.channel.get_or_fetch_channel(constants.Channels.meta)
```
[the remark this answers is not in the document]
wookie184: Fixed

bot/exts/info/patreon.py
wookie184: I'm not sure it really makes a difference here, I think i'll leave it as is

bot/exts/info/patreon.py
wookie184: I agree it is a bit confusing, i'll think about how best to do this

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,128 @@
...
+        """Send the current list of patreon supporters, sorted by tier level."""
+        guild = self.bot.get_guild(constants.Guild.id)
+
+        embed_list = []
+        for tier, role_id in PATREON_TIERS:
+            role = guild.get_role(role_id)
+
+            # Filter out any members where this is not their highest tier.
+            patrons = [member for member in role.members if get_patreon_tier(member) == tier]
+            patron_names = [f"{patron.mention} ({patron.name}#{patron.discriminator})" for patron in patrons]
```
wookie184, line 85: I'm not sure how well having mentions in embeds will work, given they often don't resolve. Would it be better to make this a normal text message with allowed_mentions to prevent mentioning users. The character limit is much more limiting unless we use 3 separate messages though.

I think I should also add some sort of overflow handling if there are too many users. I'm thinking I might shuffle each tier and pick the first 10 or something.

wookie184: This is how it currently looks on server for me (ignore first embed):

image

Not great. I'll try with a normal text message

minalike: Since you're likely going to remove the mentions, why not list all the users in paragraph form? Then we shouldn't need to pick only a random 10.

ionite34: Should we consider maybe multi-column tables or embeds? Should at least be able to fit in all the patron supporters that way.

img1

img2

bot/exts/info/patreon.py
wookie184: I have limited the command to list patrons to `bot_commands`

bot/exts/info/patreon.py
wookie184: Done

bot/exts/info/patreon.py
wookie184: I've gone with plain text names listed in the embeds, I think that's neatest.

wookie184:
This is now ready for review

bot/exts/info/patreon.py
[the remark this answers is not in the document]
wookie184: I have refactored this.

onerandomusername:
Please remove my request for review.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+        """Send the current list of patreon supporters, sorted by tier level."""
+        guild = self.bot.get_guild(Guild.id)
+
+        embed_list = []
+        for tier, role_id in PATREON_TIERS:
+            role = guild.get_role(role_id)
+
+            # Filter out any members where this is not their highest tier.
+            patrons = [member for member in role.members if get_patreon_tier(member) == tier]
+            patron_names = [f"• {patron.name}#{patron.discriminator}" for patron in patrons]
```
onerandomusername, line 80:
```suggestion
            patron_names = [f"• {str(patron.name)}" for patron in patrons]
```

ionite34 (dismissed):

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+    def __init__(self, bot: Bot) -> None:
+        self.bot = bot
+
+        self.current_monthly_supporters.start()
+
+    @commands.Cog.listener()
+    async def on_member_update(self, before: discord.Member, after: discord.Member) -> None:
+        """Send a message when someone receives a patreon role."""
+        old_patreon_tier = get_patreon_tier(before)
+        new_patreon_tier = get_patreon_tier(after)
```
onerandomusername, line 58: This has an inherent dependency on how the patreon bot works. If the patreon bot decided to take away and giveth the new role in two seperate api requests, this would cease to work. Or, if the patreon bot were to take the old role away and then give the new role, this would also cease to work.

This all hinges on the patreon bot making a single api request, and would lead to funky behaviour if their bot used more than one request--and I believe it does.

onerandomusername (commented):
didn't intend to leave another review but I am.

bot/exts/info/patreon.py
TizzySaurus: I believe you meant `str(patron)`, which since this is an f-string would be the same as just `patron`.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+from bot.constants import Channels, Guild, Roles, STAFF_PARTNERS_COMMUNITY_ROLES
+from bot.decorators import in_whitelist
+from bot.log import get_logger
+from bot.utils.channel import get_or_fetch_channel
+from bot.utils.members import has_role_id
+
+log = get_logger(__name__)
+
+PATREON_INFORMATION = (
+    "Python Discord is not-for-profit and volunteer run, so we rely on Patreon donations to do what we do. "
```
TizzySaurus, line 18: Minor, but "Python Discord is a volunteer led non-profit, so we rely..." sounds slightly better to me

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+    (2, Roles.patreon_tier_2),
+    (1, Roles.patreon_tier_1),
+]
+
+
+def get_patreon_tier(member: discord.Member) -> int:
+    """
+    Get the patreon tier of `member`.
+
+    A patreon tier of 0 indicates the user is not a patreon.
```
TizzySaurus, line 38: Again minor, but this should be "A patreon tier of 0 indicates the user is not a *patron*."

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+
+        self.current_monthly_supporters.start()
+
+    @commands.Cog.listener()
+    async def on_member_update(self, before: discord.Member, after: discord.Member) -> None:
+        """Send a message when someone receives a patreon role."""
+        old_patreon_tier = get_patreon_tier(before)
+        new_patreon_tier = get_patreon_tier(after)
+
+        if not new_patreon_tier > old_patreon_tier:
```
TizzySaurus, line 60: I think `if new_patreon_tier <= old_patreon_tier` reads nicer here.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+    async def send_current_supporters(self, channel: discord.abc.Messageable, automatic: bool = False) -> None:
+        """Send the current list of patreon supporters, sorted by tier level."""
+        guild = self.bot.get_guild(Guild.id)
+
+        embed_list = []
+        for tier, role_id in PATREON_TIERS:
+            role = guild.get_role(role_id)
+
+            # Filter out any members where this is not their highest tier.
+            patrons = [member for member in role.members if get_patreon_tier(member) == tier]
```
TizzySaurus, line 79: Since `PATREON_TIERS` is ordered highest to lowest, that means we iterate highest to lowest.

As such, we could instead store "seen patrons" and check if they're in that list. May be slightly cleaner.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+            return tier
+    return 0
+
+
+class Patreon(commands.Cog):
+    """Cog that shows patreon supporters."""
+
+    def __init__(self, bot: Bot) -> None:
+        self.bot = bot
+
```
TizzySaurus, line 51: I believe we may need to add `await self.bot.wait_until_ready()` here to ensure that the cache will be loaded before running tasks (ensure L72 + L76 etc. will definitely work).

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+            embed = discord.Embed(
+                title=role.name,
+                description="\n".join(patron_names),
+                colour=role.colour
+            )
+            embed_list.append(embed)
+
+        main_embed = discord.Embed(
+            title="Patreon Supporters - Monthly Update" if automatic else "Patreon Supporters",
+            description=(
```
TizzySaurus, line 91: I don't think it's necessary for this to be split over multiple lines, and IMO one line is cleaner.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+                PATREON_INFORMATION +
+                "\n\nThank you to the users listed below who are already supporting us!"
+            ),
+        )
+
+        await channel.send(embeds=(main_embed, *embed_list))
```
TizzySaurus, line 97: According to [the docs](https://discordpy.readthedocs.io/en/latest/api.html#discord.abc.Messageable.send), the embeds kwarg should be a list and not a tuple.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @commands.group("patreon", aliases=("patron",), invoke_without_command=True)
+    async def patreon_info(self, ctx: commands.Context) -> None:
```
TizzySaurus, line 100: Do we want to also limit what channels this can be invoked in? It's a much smaller message than the supporters subcommand, but the default attitude seems to be to limit things to bot-commands (maybe also pygen and/or community-meta here).

TizzySaurus (commented):
A few points, mostly about style so not requesting changes so as to not block the PR with this review.

bot/exts/info/patreon.py
onerandomusername: Yep, edited the suggestion

TizzySaurus: The `str()` part isn't necessary (f-strings automatically call the `__str__()`)

bot/exts/info/patreon.py
wookie184: We wait until ready before loading extensions so it shouldn't matter https://github.com/python-discord/bot-core/blob/127d0efa219d29e15bd71339d0d2517945aed57c/botcore/_bot.py#L125

bot/exts/info/patreon.py
wookie184: It would make a pretty long line. I prefer it as is.

bot/exts/info/patreon.py
wookie184: This already feels pretty clean and explicit to me, keeping track of an extra list would require more code.

bot/exts/info/patreon.py
wookie184: Given this is static text it feels similar to a tag which we allow to be invoked anywhere. I don't feel strongly either way though.

bot/exts/info/patreon.py
wookie184: I don't think this should be an issue. Whatever order things happen there will still only be one member_update that involves the tier increasing, which is all that this checks.

The only slightly unintended case would be if a user moves down a tier (e.g. from 3 to 2), and the bot does this by removing 3 and then adding 2. In this case the message would be sent, although it still wouldn't be incorrect (Just saying they are now a tier 2 patron).

Fixing this would probably require waiting for other role changes to see if they change things, although this could be fairly complicated so I think it would be best to see how what we have currently works out.

bot/exts/info/patreon.py
wookie184: I opened https://github.com/Rapptz/discord.py/issues/8460 as this shouldn't matter and it's type hinted as allowing sequences anyway

bot/exts/info/patreon.py
TizzySaurus: Fair enough :+1:

bot/exts/info/patreon.py
TizzySaurus: It would be less than the 120 char limit (107 chars), but ultimately up to you.

bot/exts/info/patreon.py
TizzySaurus: I don't feel strongly either, and I don't see it being an issue so think it's fine as-is. If it does prove to be an issue we can always add in another PR.

bot/exts/info/patreon.py
```diff
@@ -14,9 +14,9 @@
 log = get_logger(__name__)
 
 PATREON_INFORMATION = (
-    "Python Discord is not-for-profit and volunteer run, so we rely on Patreon donations to do what we do. "
-    "We use the money we get to offer excellent prizes for all of our events. Prizes like t-shirts, "
-    "stickers, microcontrollers that support CircuitPython, or maybe even a mechanical keyboard.\n\n"
+    "Python Discord a volunteer run non-profit organization, so we rely on Patreon donations to do what we do. "
```
TizzySaurus, line 17: I think you accidentally removed the "is" after Python Discord

bot/exts/info/patreon.py
```diff
@@ -14,9 +14,9 @@
...
+    "We use the money we get to offer excellent prizes for all of our events. These include t-shirts, "
+    "stickers, microcontrollers that support CircuitPython, and maybe even mechanical keyboards.\n\n"
```
TizzySaurus, line 19: NIT: "and maybe" --> "and sometimes"

bot/exts/info/patreon.py
wookie184: 👍 I made a couple other small changes to it too

bot/exts/info/patreon.py
wookie184: oops

bot/exts/info/patreon.py
wookie184: I don't think we've ever actually given out a mechanical keyboard, so I removed it from here.

ionite34 (approved):
Wasn't able to test the role assignments on real users, but seem to work fine in manual testing.

bot/exts/info/patreon.py
```diff
@@ -0,0 +1,128 @@
...
+        """Send the current list of patreon supporters, sorted by tier level."""
+        guild = self.bot.get_guild(Guild.id)
+
+        embed_list = []
+        for tier, role_id in PATREON_TIERS:
+            role = guild.get_role(role_id)
+
+            # Filter out any members where this is not their highest tier.
+            patrons = [member for member in role.members if get_patreon_tier(member) == tier]
+            patron_names = [f"• {patron}" for patron in patrons]
```
ChrisLovering, line 79: If this list is empty, imo the embed should contain something like "No patrons in this tier" as opposed to being something like this:
image

ChrisLovering (approved):
Approving as my only comment won't affect us atm, so I'll leave it to you to decide whether to action it before merging.

bot/exts/info/patreon.py
wookie184: That makes sense, done 👍

mathstrains21:
Thanks @wookie184 for finishing this off for me!

Issue #1916, opened 2021-10-27
Missing Reminders


HassanAbouelela:
For a while now, I've been noticing that my longer reminders are usually removed from the database, without being sent, leading to many missing reminders. I've been digging into it for a while now, but there is nothing noteworthy as far as I could tell.

There is nothing in logs, and there is nothing in the bot behavior that should block reminders. The most recent example I have was scheduled for 1 minute after another reminder. The other reminder was delivered, mine wasn't. I won't be linking the examples here because they are all in internal channels.

To help continue debugging this, and because it is so difficult to reproduce, I'm going to modify the reminder task to not delete entries from the DB, and instead mark them as inactive. I'll continue tracking this issue here.

HassanAbouelela:
I've spoken with Zig, and we think it's best if on top of the other PR, we try to decouple the reliance on member objects for this cog where possible. Replacing `member.mention` with manual mentions, and adding a fallback default for role checks are examples of how we could go about that.

wookie184:
I think the original issue here was fixed, though I noticed another way reminders may go missing, which i've opened a new issue for https://github.com/python-discord/bot/issues/3015.

PR #1919, opened 2021-10-28
Fixes Missing Reminders From Empty Caches

HassanAbouelela:
Blocked by python-discord/site#618.
See #1916 for more info. This PR shouldn't close that issue.

I updated the reminders cog to try and get around the issue with missing users in the cache during startup. To try and resolve that issue, I switched to fetching as a fallback if the user wasn't found. To avoid accidentally hammering the discord API with fetch requests at startup, I removed the validity checks from the init function, instead opting to only run the check during the send function. It doesn't cost much to fail the task later.


bot/exts/utils/reminders.py
```diff
@@ -55,32 +56,59 @@ async def reschedule_reminders(self) -> None:
...
 
-    def ensure_valid_reminder(self, reminder: dict) -> t.Tuple[bool, discord.User, discord.TextChannel]:
-        """Ensure reminder author and channel can be fetched otherwise delete the reminder."""
-        user = self.bot.get_user(reminder['author'])
+    async def ensure_valid_reminder(self, reminder: dict) -> t.Tuple[bool, bool, discord.User, discord.TextChannel]:
+        """
+        Ensure reminder author and channel can be fetched otherwise delete the reminder.
+
+        If more than 3 failed attempts at validating this reminder occur, it is deactivated,
+        and the second return value is set to True.
```
MarkKoz, line 72: Use the imperative mood.

bot/exts/utils/reminders.py
```diff
@@ -168,13 +196,35 @@ async def _reschedule_reminder(self, reminder: dict) -> None:
...
+                # No need to cancel the task too; it'll simply be done once this coroutine returns.
+                return
+            else:
+                # Reminder is valid, continue.
+                break
+
+        await self._format_send_reminder(reminder, expected_time, user, channel)
+
+    @lock_arg(LOCK_NAMESPACE, "reminder", itemgetter("id"), raise_error=True)
+    async def _format_send_reminder(
```
MarkKoz, line 220: Can anyone think of a better name?

Also, putting the underscore prefix is unnecessary in cogs. We can work under the assumption that cogs generally will not have their methods accessed externally because the classes aren't meant to be instantiated directly, and there are very few exceptions where `get_cog` is used to get to a method. I realise the underscore is prefix is used inconsistently in cogs, but I suppose if that is to be changed then it needs to start being brought up in reviews.

MarkKoz (changes requested):

onerandomusername:
This is no longer blocked by python-discord/site#618 as its been merged!

jchristgit (approved):
Hassan my beloved

After Mark's comments are addressed, this looks good to me.

Xithrius:
Bumping this PR, would anyone like to take over? This is still in high priority.

MarkKoz:
In hindsight, the DB should have been modified to store the timestamp of the last send attempt rather than a count of failures. The latter approach is flawed because if the bot restarts immediately after failing, it will instantly try again rather than waiting.

wookie184:
Hmmm. Given #2080 should have fixed the original issue, I'm not sure what the benefit of this retry approach is any more.

MarkKoz:
> Hmmm. Given #2080 should have fixed the original issue, I'm not sure what the benefit of this retry approach is any more.

It fixed it by doing away with the author check. This means it still sends the reminder even if the user left the server.

wookie184:
> > Hmmm. Given #2080 should have fixed the original issue, I'm not sure what the benefit of this retry approach is any more.
> 
> It fixed it by doing away with the author check. This means it still sends the reminder even if the user left the server.

Doesn't `fetch_user` still get the user if they're not on the server, so this would do the same? If we use `get_or_fetch_member` then it should only (and always as it isn't reliant on cache) return the member if they're on the server, so the retry logic wouldn't be necessary.

MarkKoz:
@wookie184 Yes, actually. I guess I don't know what the point is. @HassanAbouelela ?

HassanAbouelela:
This PR was written before that one and opted not to do away with the checks. If we don't care about the check anymore, then this is redundant.

MarkKoz:
I'm going to close this then. I don't see any benefit to checking the user if it can't reliably know whether they're in the guild.

Issue #1746, opened 2021-08-12
Updates to nomination review system posting conditions


wookie184:
## Updates to nomination review system posting conditions
There are a few changes we'd like to make to the nominations review system.

### Features we want:
1) A limit to the maximum number of reviews active in the channel at once.
2) A limit on the frequency reviews are posted (e.g. one a day).
3) A check to ensure the user has been active recently, to hold off the review if they haven't and possibly unnominate after a while.

#### General Idea:
These changes don't fit in that easily with the current scheduler based system, so I propose adding a second stage using an "awaiting review" status. So once their 30 days in the scheduler is up, they will be added to this "queue".

We could then have a `discord.ext.tasks.loop` run, maybe every hour, which checks a) that it's been 24 hours since the last vote; and b) that the number of active votes is below the maximum. If these are true then it will try and pick a user to review.

To pick a user, we'll loop over users awaiting review (probably ordered by first nominated) until we find one with recent activity, and then post their review.

#### Details:
"awaiting review" will probably need to be a new field stored on the nomination in postgres, or it may be better to change the current "reviewed" field to have multiple values, "not reviewed", "awaiting review", "being reviewed" (for keeping track of number of active reviews), and "reviewed".

For keeping track of the number of active votes, the message ID(s) of each vote could be stored on the nomination in postgres. When a user is unnominated the bot can use this to automatically delete the message(s) and archive it. 

Storing the message IDs with the user would help for archiving the vote so we have the information we need rather than trying to parse the messages with dodgy regexes to work out which review a message is from.

To handle reviews posted manually, we could replace manual posting with a bot command where you pass the user and the content you want the bot to paste. We *could* try and work out a way for this manual posting to handle multiple messages, but I think it would be easier to accept one and further content can go in a thread (although it wouldn't be archived).

The number of active votes can be calculated using the number of reviews with a "being reviewed" status.
 ___
I would be happy to try implementing these changes if the implementation I have suggested sounds OK, although it would be good if somebody else could handle the site side of things.

Also, feel free to suggest alternative implementations / point out flaws in what i've said, and let me know if i've missed out any details.

wookie184:
Thinking about this a bit more, it would probably make more sense to remove the current scheduler completely and add another check that they've been in the pool long enough along with the check for recent activity.

ChrisLovering:
[...]

PR #1928, opened 2021-11-01
Manage nomination threads in bot

ChrisLovering:
These changes moves the management of nomination threads from [Sir Threadevere](https://github.com/python-discord/sir-threadevere) to the bot. Once this PR is merged, Sir Threadevere can be shutdown.

These changes include creating a thread while posting the nomination, pinging staff inside the thread, and then archiving it once the nomination is concluded.

### DIFF
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -78,14 +78,14 @@ class Reviewer:
 
     async def post_review(self, user_id: int, update_database: bool) -> None:
         """Format the review of a user and post it to the nomination voting channel."""
-        review, reviewed_emoji = await self.make_review(user_id)
+        review, reviewed_emoji, nominee = await self.make_review(user_id)
         if not review:
             return
 
         guild = self.bot.get_guild(Guild.id)
         channel = guild.get_channel(Channels.nomination_voting)
 
-        log.trace(f"Posting the review of {user_id}")
+        log.trace(f"Posting the review of {nominee} ({nominee.id})")
         messages = await self._bulk_send(channel, review)
 
         await pin_no_system_message(messages[0])
@@ -113,14 +113,14 @@ class Reviewer:
             return "", None
 
         guild = self.bot.get_guild(Guild.id)
-        member = await get_or_fetch_member(guild, user_id)
+        nominee = await get_or_fetch_member(guild, user_id)
 
-        if not member:
+        if not nominee:
             return (
                 f"I tried to review the user with ID `{user_id}`, but they don't appear to be on the server :pensive:"
-            ), None
+            ), None, None
 
-        opening = f"{member.mention} ({member}) for Helper!"
+        opening = f"{nominee.mention} ({nominee}) for Helper!"
 
         current_nominations = "\n\n".join(
             f"**<@{entry['actor']}>:** {entry['reason'] or '*no reason given*'}"
@@ -128,7 +128,7 @@ class Reviewer:
         )
         current_nominations = f"**Nominated by:**\n{current_nominations}"
 
-        review_body = await self._construct_review_body(member)
+        review_body = await self._construct_review_body(nominee)
 
         reviewed_emoji = self._random_ducky(guild)
         vote_request = (
@@ -138,7 +138,7 @@ class Reviewer:
         )
 
         review = "\n\n".join((opening, current_nominations, review_body, vote_request))
-        return review, reviewed_emoji
+        return review, reviewed_emoji, nominee
 
     async def archive_vote(self, message: PartialMessage, passed: bool) -> None:
         """Archive this vote to #nomination-archive."""
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -10,7 +10,7 @@ from typing import List, Optional, Union
 
 import arrow
 from dateutil.parser import isoparse
-from discord import Embed, Emoji, Member, Message, NoMoreItems, PartialMessage, TextChannel
+from discord import Embed, Emoji, Member, Message, NoMoreItems, NotFound, PartialMessage, TextChannel
 from discord.ext.commands import Context
 
 from bot.api import ResponseCodeError
@@ -220,10 +220,13 @@ class Reviewer:
         if not nomination_thread:
             log.warning(f"Could not find a thread linked to {message.channel.id}-{message.id}")
             return
-        await nomination_thread.edit(archived=True)
 
         for message_ in messages:
-            await message_.delete()
+            with contextlib.suppress(NotFound):
+                await message_.delete()
+
+        with contextlib.suppress(NotFound):
+            await nomination_thread.edit(archived=True)
 
     async def _construct_review_body(self, member: Member) -> str:
         """Formats the body of the nomination, with details of activity, infractions, and previous nominations."""
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -15,7 +15,7 @@ from discord.ext.commands import Context
 
 from bot.api import ResponseCodeError
 from bot.bot import Bot
-from bot.constants import Channels, Colours, Emojis, Guild
+from bot.constants import Channels, Colours, DEFAULT_THREAD_ARCHIVE_TIME, Emojis, Guild, Roles
 from bot.log import get_logger
 from bot.utils.members import get_or_fetch_member
 from bot.utils.messages import count_unique_users_reaction, pin_no_system_message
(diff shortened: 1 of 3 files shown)


bot/exts/recruitment/talentpool/_review.py
```diff
@@ -41,6 +41,10 @@
     r"(?:<@&\d+> <@&\d+>\n)*?<@!?(\d+?)> \(.+#\d{4}\) for Helper!\n\n\*\*Nominated by:\*\*",
```
Akarys42, line 41: This regex will have to be updated, although I have no idea how it didn't already break

Akarys42 (changes requested):
It would be nice if we could also add a listener on this cog automatically unarchiving threads if they get archived through a timeout and not manually

bot/exts/recruitment/talentpool/_review.py
ChrisLovering: The one I made for threadbot seem to do the job. Does this look better?
```
NOMINATION_MESSAGE_REGEX = re.compile(
    r"<@!?\d+> \((.+)#\d{4}\) for Helper!\n\n\*\*Nominated by:\*\*",
    re.MULTILINE
)
```

ChrisLovering: (ideally when we rewrite this for https://github.com/python-discord/bot/issues/1746 we will store these messages in redis

ChrisLovering: Fixed in https://github.com/python-discord/bot/pull/1928/commits/57288b22e4e42493e2522b86db8a6f55c117d820 as for why it didn't break, this current regex supports both the current style of message, but also the old style before thread-bot. Since there aren't any old-style messages no, we can use the same regex as threadbot does.

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -210,6 +219,13 @@ async def archive_vote(self, message: PartialMessage, passed: bool) -> None:
...
 
+        # Thread channel IDs are the same as the message ID of the parent message.
+        nomination_thread = message.guild.get_thread(message.id)
+        if not nomination_thread:
+            log.warning(f"Could not find a thread linked to {message.channel.id}-{message.id}")
+            return
+        await nomination_thread.edit(archived=True)
+
         for message_ in messages:
             await message_.delete()
```
jchristgit, line 230: Should we move this above the archive logic to make sure the nomination message(s) are deleted even if the thread is lost somewhere?

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -78,14 +81,14 @@ def schedule_review(self, user_id: int) -> None:
 
     async def post_review(self, user_id: int, update_database: bool) -> None:
```
jchristgit, line 82: The `-> None` here is kind of ... wrong. But since it was wrong before, no need to fix it here.

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -36,11 +36,14 @@
...
-# Historic nominations will have 2 role mentions at the start, new ones won't, optionally match for this.
 NOMINATION_MESSAGE_REGEX = re.compile(
-    r"(?:<@&\d+> <@&\d+>\n)*?<@!?(\d+?)> \(.+#\d{4}\) for Helper!\n\n\*\*Nominated by:\*\*",
+    r"<@!?\d+> \((.+)#\d{4}\) for Helper!\n\n\*\*Nominated by:\*\*",
     re.MULTILINE
 )
+if DEBUG_MODE:
+    NOMINATION_THREAD_ARCHIVE_TIME = ThreadArchiveTimes.DAY.value
+else:
+    NOMINATION_THREAD_ARCHIVE_TIME = ThreadArchiveTimes.WEEK.value
```
jchristgit, line 46: This setting should maybe just be part of the config, or at least be computed in `bot/constants.py` to keep configuration logic out of here. What do you think?

jchristgit (commented):
This looks good. I would leave an approval, but I have a few comments which may be nice to address or not.

bot/exts/recruitment/talentpool/_review.py
ChrisLovering: Hmm, I guess we could have a `DEFAULT_THREAD_ARCHIVE_TIME` in constants.py yea.

bot/exts/recruitment/talentpool/_review.py
ChrisLovering: Yea, agreed. we should probably wrap the edits and delete in suppress too

bot/exts/recruitment/talentpool/_review.py
ChrisLovering: How so?

bot/exts/recruitment/talentpool/_review.py
ChrisLovering: https://github.com/python-discord/bot/pull/1928/commits/b8d393e90e550e425327314ddf72c6275718f688

bot/exts/recruitment/talentpool/_review.py
ChrisLovering: https://github.com/python-discord/bot/pull/1928/commits/df38d2817bf20f6eac1cde8fa7f63d741f4a5d74

Akarys42 (changes requested):
I ran into an error when trying to archive
```
pythondiscord-bot-1        | 2021-11-20 17:07:28 | bot | ERROR | Unhandled exception in on_raw_reaction_add.
pythondiscord-bot-1        | Traceback (most recent call last):
pythondiscord-bot-1        |   File "/usr/local/lib/python3.9/site-packages/discord/client.py", line 351, in _run_event
pythondiscord-bot-1        |     await coro(*args, **kwargs)
pythondiscord-bot-1        |   File "/bot/bot/exts/recruitment/talentpool/_cog.py", line 526, in on_raw_reaction_add
pythondiscord-bot-1        |     await self.reviewer.archive_vote(message, emoji == Emojis.incident_actioned)
pythondiscord-bot-1        |   File "/bot/bot/exts/recruitment/talentpool/_review.py", line 171, in archive_vote
pythondiscord-bot-1        |     user_id = int(NOMINATION_MESSAGE_REGEX.search(content).group(1))
pythondiscord-bot-1        | ValueError: invalid literal for int() with base 10: 'Chrisjl'
```

The regex [doesn't seem right](https://regex101.com/r/TT5Etx/1). I'd also remove the `**Nominated by:**` literal, to try to not make manual reviews fail. Also maybe we should properly handle that error and send an error message somewhere?

Akarys42 (approved):
Excellent :shipit:

bot/exts/recruitment/talentpool/_review.py
Akarys42: The annotation seems correct here

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -78,14 +77,14 @@ def schedule_review(self, user_id: int) -> None:
...
-        review, reviewed_emoji = await self.make_review(user_id)
+        review, reviewed_emoji, nominee = await self.make_review(user_id)
         if not review:
             return
 
         guild = self.bot.get_guild(Guild.id)
         channel = guild.get_channel(Channels.nomination_voting)
 
-        log.trace(f"Posting the review of {user_id}")
+        log.trace(f"Posting the review of {nominee} ({nominee.id})")
```
janine9vn: So! This is fun. The addition of `nominee.id` will no longer make this fail silently if a user leaves the guild after they were nominated but before they're up for a vote.

`self.make_review()` returns (string, None, None). `review` is therefore a non-empty string and that short-circuit logic there won't work anymore. `nominee.id` now tries to access the `id` attribute which a NoneType doesn't have.

So, the `if not review` should explicitly check for the nominee, not the review.

janine9vn (approved):
This works well! (After I lost a year of my life debugging emojis).

Approving this since you've already fixed my comments.

Goodbye Sir Threadevere!

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -95,6 +94,12 @@ async def post_review(self, user_id: int, update_database: bool) -> None:
             for reaction in (reviewed_emoji, "\N{THUMBS UP SIGN}", "\N{THUMBS DOWN SIGN}"):
```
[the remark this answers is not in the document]
ChrisLovering: This has been fixed in https://github.com/python-discord/bot/pull/1928/commits/4bde7915630bbd18a767e77b7a7cb54404fb0306

bot/exts/recruitment/talentpool/_review.py
ChrisLovering: Fixed in https://github.com/python-discord/bot/pull/1928/commits/769660b46b1eebedadeccc9d22a7bdf7705befc2

Issue #1903, opened 2021-10-21
Add Consistency Checks


HassanAbouelela:
Every now and then, the help channel cog will desync (mostly due to unavoidable issues from the discord API). This desync takes many forms, most prominent being:

- Decrease in number of help channels available (#1941)
- Desync of dynamic help message in `#how-to-get-help`
- ~~Pins not being removed after a channel is closed~~ Fixed
- Cooldown roles not being removed after a session ends (#1940)

These happen often enough that I think we should spend some effort to get them to automatically figure themselves out. I've divided this issue into three separate sub-problems, one for (mostly) each of the points above.

## Pins
This one is relatively straightforward. We should [get all pins](https://discordpy.readthedocs.io/en/master/api.html#discord.abc.Messageable.pins) when a channel is moved into dormant, and unpin them. I've checked the rate limit for that endpoint, and it appears to be 1 request / 5s / channel, so I think we're fine.

## Cooldown roles
We already keep a list of which authors currently have a channel open, and I believe that does not desync in line with the cooldown roles (though it can be difficult to test). I think that means we can compare the [list of people with the role](https://discordpy.readthedocs.io/en/master/api.html#discord.Role.members), against the list of channel claimants, and remove any outliers. We should probably run this in some sort of periodic task, possibly every hour? That implementation detail can be left up to the author.

## Channels & Channel Names
For the two desync patterns not mentioned yet, a simple cog reload seems to mostly do the trick. It's the go-to method for staff whenever we see a desync, and I estimate we run it bi-weekly right now. It only takes a few seconds (5 according to Akarys), so we should be fine to periodically just reload. Perhaps every hour as well.

@Akarys42 has brought up concerns about redis race conditions, we can discuss those here.

## Closing Remarks
None of the mentioned fixes depend on each other, so we tackle each in separate PRs asynchronously.

I'm going to say that if no one has concerns within the next couple days, I'll be approving this issue for the cooldown role and pin fixes. I'll hold off on the last one till we figure out the race issue, but I'll update the issue once it's approved.

dementati:
The pins issue is solved by #1909 and #1936.

PR #1936, opened 2021-11-05
Unpin all messages when help channel goes dormant

dementati:
As described in #1903, unpinning a help channel question will occasionally fail when a channel goes dormant. This issue was previously fixed in #1909 by querying and unpinning all pinned messages in the help channel when it moves from dormant to available. This likely fixes the issue, but the current unpinning solution is still unnecessarily complex.

Unpinning the question when the channel goes dormant is currently solved by storing the question message ID when the channel is first claimed and retrieving it from there to be unpinned at the appropriate time. An alternative and simpler solution, implemented in this PR, is to query all pinned messages and unpin those found, like what happens when the channel becomes available. This allows us to get rid of the message cache with no known disadvantages.


MarkKoz:
If they're unpinned when the channel becomes available, then why even bother unpinning when it goes dormant?

> This allows us to get rid of the message cache with no known disadvantages.

Getting pinned messages requires an API call to Discord. With the cache, this is avoided. That is basically the only advantage. That being said, this sort of optimisation is not necessary; it was nice to have but it's not worth keeping if it's causing consistency issues.

dementati:
> If they're unpinned when the channel becomes available, then why even bother unpinning when it goes dormant?

Apparently, it's the API itself that bugs out occasionally, and making two passes will decrease the likelihood that there are pins remaining by the time the channel becomes available.

> Getting pinned messages requires an API call to Discord. With the cache, this is avoided. That is basically the only advantage.

That is true.

> it was nice to have but it's not worth keeping if it's causing consistency issues.

As stated above, it's allegedly the API and not the cache that caused the issue. I just prefer the solution in this PR because it reduces code complexity.

Bluenix2 (approved):
Perfect, works much better than the previous behaviour! Another bug squashed..
image

bot/exts/help_channels/_message.py
```diff
@@ -174,8 +174,7 @@ async def notify(channel: discord.TextChannel, last_notification: t.Optional[Arr
 
 async def pin(message: discord.Message) -> None:
     """Pin an initial question `message` and store it in a cache."""
```
onerandomusername, line 176: No longer storing to a cache.

onerandomusername (changes requested):

bot/exts/help_channels/_message.py
MarkKoz, line 176:
```suggestion
    """Pin an initial question `message`."""
```

bot/exts/help_channels/_message.py
```diff
@@ -201,11 +200,8 @@ async def send_available_message(channel: discord.TextChannel) -> None:
 
 async def unpin(channel: discord.TextChannel) -> None:
     """Unpin the initial question message sent in `channel`."""
```
MarkKoz, line 202:
```suggestion
    """Unpin all messages sent in `channel`."""
```

MarkKoz (changes requested):

MarkKoz (commented):
This will allow a `discord.Message` to be passed to `pin_wrapper` instead of a message ID. Thus, the wrapper will be able to call the `discord.Message.pin` and `discord.Message.unpin` functions rather than using the undocumented methods on `HTTPClient`. Would you mind updating the wrapper function too @dementati?

Xithrius:
@dementati What's your status on this PR?

MarkKoz (approved):

ChrisLovering (approved):

PR #1937, opened 2021-11-05
Improve Filter Rerun Logic

### LINKED ISSUE #1934: Don't trigger filters on pin
The filters are re-run on messages that are edited and pinned. Because of the help channel system, This can cause a filter to trigger twice if the opening message contains a trigger.

We should check what causes the filters to trigger on pins and see if there's a way to disable that.

TizzySaurus:
Closes #1934.

Will now only rerun filters in `on_message_edit` if the content or attachments of a message changed. This stops reruns for removal of embeds, change in pin status, etc.

### DIFF
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -189,10 +189,14 @@ class Filtering(Cog):
         """
         Invoke message filter for message edits.
 
-        If there have been multiple edits, calculate the time delta from the previous edit.
+        Also calculates the time delta from the previous edit or when message was sent if there's no prior edits.
         """
-        # We only care about changes to the message contents/attachments, not pin status or embeds etc.
-        if before.content == after.content and before.attachments == after.attachments:
+        # We only care about changes to the message contents/attachments and embed additions, not pin status etc.
+        if all((
+            before.content == after.content,  # content hasn't changed
+            before.attachments == after.attachments,  # attachments haven't changed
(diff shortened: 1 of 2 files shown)


HassanAbouelela (approved):

Qwerty-133:
This would not alert us when an embed has been added to a message, which is why we should only really be skipping the filters if the message was pinned or unpinned.

TizzySaurus:
> This would not alert us when an embed has been added to a message, which is why we should only really be skipping the filters if the message was pinned or unpinned.

We don't want to be alerted of that, since it means we'll get double-pinned for domain filters

Qwerty-133:
We might miss self-bots though.

HassanAbouelela:
This is only for message edits, can you actually edit an embed in?

Qwerty-133:
Yep, you can

TizzySaurus:
> We might miss self-bots though.

I believe we have other filters for that

Qwerty-133:
The reason we got two pings for that mod-alert is because of an oversight in the embed filter. So we should be running our filters on embed additions, I can PR a fix for the embed filter.

TizzySaurus:
> We should be running our filters on embed additions. I can PR a fix for the embed filter.
> 
> #1938 relevant issue.

@Qwerty-133 This has been addressed :+1:

Qwerty-133 (approved):
Everything looks good to me, thanks!

PR #1939, opened 2021-11-05
Dynamic views for command help embeds

mbaruh:
Adds views for commands to navigate groups.
For subcommands, a button is added to show the parent's help embed.
For groups, buttons are added for each subcommand to show their help embeds.

The views are not generated when help is invoked in the context of an error.

The code has gone through several iterations of design, and this is what I arrived at.
Because conversion of callbacks to buttons is done in __subclass_init__ of the view baseclass, I had to create Button subclasses.
Because the normalization of button positions is in the view init and is obfuscated, I had to meddle with discord.py internals a bit in GroupView to make it work. The only way I can think of around all that is to instantiate an instance of `type` on the fly, but that doesn't sound very reader-friendly.

I also don't really like the help formatting functions returning the views, so if anyone has a better idea let me know. At least in the case of a group, I had to do get the view from there because I needed the filtered subcommands. Since the filtering is asynchronous I couldn't do it in the view's init.

https://user-images.githubusercontent.com/18114431/140521230-3923da4b-99d1-4eaf-8c9b-69f8eb35f225.mp4

### DIFF
--- a/bot/exts/backend/error_handler.py
+++ b/bot/exts/backend/error_handler.py
@@ -1,5 +1,4 @@
 import difflib
-import typing as t
 
 from discord import Embed
 from discord.ext.commands import ChannelNotFound, Cog, Context, TextChannelConverter, VoiceChannelConverter, errors
@@ -97,13 +96,14 @@ class ErrorHandler(Cog):
             # MaxConcurrencyReached, ExtensionError
             await self.handle_unexpected_error(ctx, e)
 
-    @staticmethod
-    def get_help_command(ctx: Context) -> t.Coroutine:
+    async def send_command_help(self, ctx: Context) -> None:
         """Return a prepared `help` command invocation coroutine."""
         if ctx.command:
-            return ctx.send_help(ctx.command)
+            self.bot.help_command.context = ctx
+            await ctx.send_help(ctx.command)
+            return
 
-        return ctx.send_help()
+        await ctx.send_help()
 
     async def try_silence(self, ctx: Context) -> bool:
         """
@@ -245,7 +245,6 @@ class ErrorHandler(Cog):
         elif isinstance(e, errors.ArgumentParsingError):
             embed = self._get_error_embed("Argument parsing error", str(e))
             await ctx.send(embed=embed)
-            self.get_help_command(ctx).close()
             self.bot.stats.incr("errors.argument_parsing_error")
             return
         else:
@@ -256,7 +255,7 @@ class ErrorHandler(Cog):
             self.bot.stats.incr("errors.other_user_input_error")
 
         await ctx.send(embed=embed)
-        await self.get_help_command(ctx)
+        await self.send_command_help(ctx)
 
     @staticmethod
     async def handle_check_failure(ctx: Context, e: errors.CheckFailure) -> None:
--- a/bot/exts/info/help.py
+++ b/bot/exts/info/help.py
@@ -1,10 +1,12 @@
+from __future__ import annotations
+
 import itertools
 import re
 from collections import namedtuple
 from contextlib import suppress
-from typing import List, Union
+from typing import List, Optional, Union
 
-from discord import Colour, Embed
+from discord import ButtonStyle, Colour, Embed, Emoji, Interaction, PartialEmoji, ui
 from discord.ext.commands import Bot, Cog, Command, CommandError, Context, DisabledCommand, Group, HelpCommand
 from rapidfuzz import fuzz, process
 from rapidfuzz.utils import default_process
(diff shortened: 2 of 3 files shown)


bot/exts/backend/error_handler.py
```diff
@@ -97,13 +96,14 @@ async def on_command_error(self, ctx: Context, e: errors.CommandError) -> None:
             # MaxConcurrencyReached, ExtensionError
             await self.handle_unexpected_error(ctx, e)
 
-    @staticmethod
-    def get_help_command(ctx: Context) -> t.Coroutine:
+    async def send_command_help(self, ctx: Context) -> None:
         """Return a prepared `help` command invocation coroutine."""
         if ctx.command:
-            return ctx.send_help(ctx.command)
+            self.bot.help_command.context = ctx
```
Bluenix2, line 102: What caused this?

bot/exts/backend/error_handler.py
```diff
@@ -245,7 +245,6 @@ async def handle_user_input_error(self, ctx: Context, e: errors.UserInputError)
         elif isinstance(e, errors.ArgumentParsingError):
             embed = self._get_error_embed("Argument parsing error", str(e))
             await ctx.send(embed=embed)
-            self.get_help_command(ctx).close()
```
Bluenix2, line 248: Dont't we still need to remove the trashcan? Though that may be different..

bot/exts/info/help.py
```diff
@@ -26,6 +28,123 @@
...
+
+        subcommand = self.command
+        if isinstance(subcommand, Group):
+            embed, subcommand_view = await self.help_command.format_group_help(subcommand)
+        else:
+            embed, subcommand_view = await self.help_command.command_formatting(subcommand)
+        await message.edit(embed=embed, view=subcommand_view)
+
+
+class GroupButton(ui.Button):
```
Bluenix2, line 72: This is used for that "Go back" button right?

bot/exts/info/help.py
```diff
@@ -26,6 +28,123 @@
...
+    """
+
+    MAX_BUTTONS_IN_ROW = 5
+    MAX_ROWS = 5
+
+    def __init__(self, help_command: CustomHelpCommand, group: Group, subcommands: list[Command]):
+        super().__init__(help_command, group)
+        # Don't add buttons if only a portion of the subcommands can be shown.
+        if len(subcommands) + len(self.children) > self.MAX_ROWS * self.MAX_BUTTONS_IN_ROW:
+            return
```
Bluenix2, line 137: No trace, debug, log or warning? No error raised to signal this behaviour?

Just silently do nothing.. what benefit does this have?

bot/exts/info/help.py
```diff
@@ -26,6 +28,123 @@
...
+
+        for subcommand in subcommands:
+            presentable_name = subcommand.name.replace("_", " ").title()
```
Bluenix2, line 140: I am not sure if I think this should be done, it feels more like actual buttons that when you press will "append" something rather than selecting a command.

Could just be a matter of getting used to it?

bot/exts/info/help.py
```diff
@@ -26,6 +28,123 @@
...
+            self.children.append(
+                SubcommandButton(help_command, subcommand, label=presentable_name)
+            )
+        # This is required to normalize the button positions.
+        self.__weights = ui.view._ViewWeights(self.children)
```
Bluenix2, line 145: Won't this be done by discord.py if you just use the correct method being `add_item()`?

bot/exts/info/help.py
mbaruh: I'm not sure what you mean here

bot/exts/info/help.py
mbaruh: Like the comment says, that you don't show buttons for only some of the subcommand. If you can't add the menu properly then don't add it.

I can add logging

bot/exts/info/help.py
mbaruh: Yeah

bot/exts/backend/error_handler.py
mbaruh: This is never called, it's closed immediately after creation. This is an artifact left behind from when the function had a slightly different structure.

bot/exts/backend/error_handler.py
mbaruh: When you call the help command directly, the context for the help cog gets updated, but here that doesn't happen anywhere and the context is empty.

I'm not at a computer at the moment, but you'll be able to see the bug that happens if you remove that line and then try to do `!clean &$#`

bot/exts/info/help.py
mbaruh: Hmm yeah I'll try that, thanks

bot/exts/info/help.py
Bluenix2: When I used the help command and the name of one subcommand was "Append" it seemed more to me that when I press that button something will append rather than me viewing the help of the subcommand called "append".

Does that make more sense? I think the `.title()` part is what messes this up for me.

bot/exts/info/help.py
jchristgit: Personally I think we can leave it, since the bot logs plenty of things by default already, and this is a static thing that won't ever change.

bot/exts/info/help.py
jchristgit: I'd agree with removing `.title()`, I had the same thought as @Bluenix2 watching the video posted.

jchristgit (approved):
This is a very very very cool use of buttons. When the open discussions are addressed, this looks good to merge from me 👍

bot/exts/info/help.py
Bluenix2: But since there's *no* logging so it won't even show up in development.

mbaruh: e78c821

bot/exts/info/help.py
mbaruh: 6766fc1

bot/exts/info/help.py
mbaruh: f5f10f8

Bluenix2 (approved):
Works nicely, thanks!

PR #1941, opened 2021-11-07
Check number of available help channels hourly

dementati:
Solves the help channel count inconsistency from #1903.

If number of available help channels is incorrect, reload the HelpChannels cog.

### DIFF
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -7,7 +7,7 @@ from operator import attrgetter
 import arrow
 import discord
 import discord.abc
-from discord.ext import commands
+from discord.ext import commands, tasks
 
 from bot import constants
 from bot.bot import Bot
@@ -88,6 +88,7 @@ class HelpChannels(commands.Cog):
         """Cancel the init task and scheduled tasks when the cog unloads."""
         log.trace("Cog unload: cancelling the init_cog task")
         self.init_task.cancel()
+        self.check_channel_consistency.cancel()
 
         log.trace("Cog unload: cancelling the channel queue tasks")
         for task in self.queue_tasks:
@@ -330,6 +331,7 @@ class HelpChannels(commands.Cog):
         await self.init_available()
         _stats.report_counts()
 
+        self.check_channel_consistency.start()
         log.info("Cog is ready!")
 
     async def move_idle_channel(self, channel: discord.TextChannel, has_task: bool = True) -> None:
@@ -642,3 +644,16 @@ class HelpChannels(commands.Cog):
         else:
             await _caches.help_dm.delete(ctx.author.id)
         await ctx.send(f"{constants.Emojis.ok_hand} {ctx.author.mention} Help DMs {state_str}!")
+
+    @tasks.loop(hours=1)
+    async def check_channel_consistency(self) -> None:
+        """
+        Check number of available help channels periodically.
+
+        If number of available help channels is incorrect, reload cog.
+        """
+        log.trace("Checking channel consistency")
+        available_help_channels = set(_channel.get_category_channels(self.available_category))
+        if len(available_help_channels) != 3:
+            log.trace("Unexpected number of available help channels, reloading cog...")
+            self.bot.reload_extension("bot.exts.help_channels")
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -654,6 +654,6 @@ class HelpChannels(commands.Cog):
         """
         log.trace("Checking channel consistency")
         available_help_channels = set(_channel.get_category_channels(self.available_category))
-        if len(available_help_channels) != 3:
+        if len(available_help_channels) != constants.HelpChannels.max_available:
             log.trace("Unexpected number of available help channels, reloading cog...")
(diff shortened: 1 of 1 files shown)


bot/exts/help_channels/_cog.py
```diff
@@ -642,3 +644,16 @@ async def helpdm_command(
...
+    @tasks.loop(hours=1)
+    async def check_channel_consistency(self) -> None:
+        """
+        Check number of available help channels periodically.
+
+        If number of available help channels is incorrect, reload cog.
+        """
+        log.trace("Checking channel consistency")
+        available_help_channels = set(_channel.get_category_channels(self.available_category))
+        if len(available_help_channels) != 3:
```
Numerlor, line 657: the maximum amount of available channels is configurable in the config, this should use the config value

Numerlor (commented):
Would this help with the issue? I imagine if the count drops the admins will be notified sooner than the one hour period in most cases

bot/exts/help_channels/_cog.py
dementati: Fixed.

dementati:
> Would this help with the issue? I imagine if the count drops the admins will be notified sooner than the one hour period in most cases

I used the one-hour interval mentioned in the original issue, but perhaps that was referring to actually reloading the cog and not just checking the available channel count. Since getting the category channels in this manner doesn't seem to require an API call, I guess we could do it much more frequently at no cost.

bot/exts/help_channels/_cog.py
```diff
@@ -642,3 +644,16 @@ async def helpdm_command(
         else:
             await _caches.help_dm.delete(ctx.author.id)
         await ctx.send(f"{constants.Emojis.ok_hand} {ctx.author.mention} Help DMs {state_str}!")
+
+    @tasks.loop(hours=1)
+    async def check_channel_consistency(self) -> None:
+        """
+        Check number of available help channels periodically.
+
+        If number of available help channels is incorrect, reload cog.
```
Bluenix2, line 653:
```suggestion
        If the number of available help channels is incorrect, reload the cog.
```

bot/exts/help_channels/_cog.py
```diff
@@ -642,3 +644,16 @@ async def helpdm_command(
         else:
             await _caches.help_dm.delete(ctx.author.id)
         await ctx.send(f"{constants.Emojis.ok_hand} {ctx.author.mention} Help DMs {state_str}!")
+
+    @tasks.loop(hours=1)
+    async def check_channel_consistency(self) -> None:
```
onerandomusername, line 649: I believe there is a possible race condition which would extend the amount of bugs.

If a channel is currently being closed while this runs, they would be awaiting an edit coroutine. This would allow the program to switch context and run the waiting task here.

However, the api would still be completing when this task decides to reload the cog, furthering the issue. 

Its a race condition that would make the problem worse, imo. 

A solution with that could be some asyncio.Locks, in order to not let the context switch to within here before a channel finishes closing and returns.

onerandomusername (changes requested):
This doesn't seem like it would fix the problem, and may add more problems in the long run.

bot/exts/help_channels/_cog.py
HassanAbouelela: I've looked into it when we opened the issue, but the cog is pretty resilient. If the closing process is still on the bot side, that shouldn't affect matters since it'll be canceled and resynced on startup.

If it's waiting on an API call from discord, I think it actually does not cancel until it's done, but I may be mistaken. Either way, the worst case scenario if it doesn't is that we'll move one channel twice, but it'll be to the same place both times, so nothing will break.

bot/exts/help_channels/_cog.py
```diff
@@ -642,3 +644,16 @@ async def helpdm_command(
...
+        """
+        Check number of available help channels periodically.
+
+        If number of available help channels is incorrect, reload cog.
+        """
+        log.trace("Checking channel consistency")
+        available_help_channels = set(_channel.get_category_channels(self.available_category))
+        if len(available_help_channels) != constants.HelpChannels.max_available:
+            log.trace("Unexpected number of available help channels, reloading cog...")
+            self.bot.reload_extension("bot.exts.help_channels")
```
MarkKoz, line 659: Why reload the whole cog? Is it not sufficient to just call `init_available()` again?

MarkKoz: In such case, I think you'd need to add a lock so it doesn't interfere with `claim_channel`.

Xithrius:
@MarkKoz Hello. What's your status on this PR?

Thanks.

MarkKoz:
> @MarkKoz Hello. What's your status on this PR?
> 
> Thanks.

I do have some local changes but nothing was competed. I believe I ran into a difficult concurrency problem which led me to writing a custom lock, but I was having trouble achieving what I needed. I'll try to find it this weekend and see if this is something I can continue or not.

MarkKoz:
Work has started on the new help system, and given the amount of work this issue will require, I don't think it makes sense to continue. It's not high priority anyway.

ChrisLovering:
Agreed, lets not spend resources on this.

PR #1943, opened 2021-11-08
Make type-hint tag

jonathan-d-zhang:
@TizzySaurus 

Closes python-discord/meta#154

Let the bikeshedding begin!


TizzySaurus:
@SavagePastaMan 

A few first-glance alterations:

I don't like that we're calling the variable `sum`, thus overriding the built-in `sum()` function. The type hint here also doesn't particularly make sense. Inline assignment type hints like this should only be done when the type isn't obvious from the variables / other type hints.

There's also some small rewording I've done here.

Current proposal:
image

We should probably link to some further resources on type hints (perhaps a PEP?) and also mention static type checkers like mypy which can be used to "enforce" type hints.

I'm going to bed now, will look further into this tomorrow :+1:

*(Apparently my tired brain forgot to add the labels all at once so sorry for the spam there.)

SavagePastaMan:
>  The type hint here also doesn't particularly make sense.

I wanted to include it to show that you can typehints normal variables too, maybe we can find another place to show that? I agree with not naming it `sum` though.

> We should probably link to some further resources on type hints (perhaps a PEP?) and also mention static type checkers like mypy which can be used to "enforce" type hints.

Yes, I agree.

@TizzySaurus

bot/resources/tags/type-hint.md
```diff
@@ -0,0 +1,16 @@
+**Type Hints**
+
+A typehint indicates what type something should be. For example,
+```python
+def add(a: int, b: int) -> int:
+    sum: int = a + b
+    return sum
+```
+In this case, `a` and `b` are expected to be ints, and the function returns an int. We also declare an intermediate variable `sum`, which we indicate to be an int.
```
Bluenix2, line 9: Should probably mention that the `sum` variable can be inferred (or guessed could be a easier word to understand) by whatever static type checker the user is using.

bot/resources/tags/type-hint.md
```diff
@@ -0,0 +1,16 @@
...
+
+It's important to note these are just hints and have no runtime effect. For example,
+```python
+#uh oh
```
Bluenix2, line 13:
```suggestion
# Uh oh!
```

bot/resources/tags/type-hint.md
```diff
@@ -0,0 +1,16 @@
...
+add("hello ", "world")
+```
+This code will run without error, even though it doesn't follow the function's type hints.
```
Bluenix2, line 16:
```suggestion
This code will run without error because strings can also be added together, even though it doesn't follow the function's type hints.
```

Something like this or similar, we should make it clear that it works even though it's the wrong type because the type we passed in allows the same operations - a kind of gateway into duck typing.

bot/resources/tags/type-hint.md
```diff
@@ -0,0 +1,16 @@
+**Type Hints**
+
+A typehint indicates what type something should be. For example,
```
Bluenix2, line 3:
```suggestion
A typehint indicates what type something is expected to be. For example,
```

Would this be an improvement in wording?

Bluenix2 (commented):
🚴‍♂️🚴‍♂️🚴‍♂️💨💨

bot/resources/tags/type-hint.md
minalike: "Type hint(s)" should be two separate words

SavagePastaMan: yes

bot/resources/tags/type-hint.md
```diff
@@ -0,0 +1,19 @@
+**Type Hints**
+
+A type hint indicates what type something is expected to be. For example,
+```python
+def add(a: int, b: int) -> int:
+    return a + b
+```
+In this case, we have a function,`add`, with parameters `a` and `b`. The type hints indicate that the parameters and return type are all integers.
```
TizzySaurus, line 8: Missing a space here after the comma

bot/resources/tags/type-hint.md
```diff
@@ -0,0 +1,19 @@
...
+
+It's important to note these are just hints and have no runtime effect. For example,
+```python
+# Uh oh
+add("hello ", "world")
+```
+This code won't error even though it doesn't follow the function's type hints. It will just concatenate the two strings.
```
TizzySaurus, line 15: I'd add a comma here after the "error" -- "...won't error, even though..."

bot/resources/tags/type-hint.md
```diff
@@ -0,0 +1,19 @@
...
+
+Third party tools like [mypy](http://mypy-lang.org/) can enforce your type hints. Mypy would error in the second example.
```
TizzySaurus, line 17: Picky but should probably be `MyPy` not `mypy`. Same for after the period.

bot/resources/tags/type-hint.md
```diff
@@ -0,0 +1,19 @@
...
+
+For more info about type hints, check out [PEP 484](https://www.python.org/dev/peps/pep-0484/).
```
TizzySaurus, line 19: Not sure if we want a newline before this or not. I think it's better without it.

TizzySaurus (changes requested):
Few minor things

bot/resources/tags/type-hint.md
ichard26, line 19: Not sure if this would be out of scope for this tag but I think it'd be a nice addition to reference https://typing.readthedocs.io/en/latest/index.html as a more general user focused resource on static typing as it'd be next logical step after learning about type-hints IMO.

JelleZijlstra: Yes, there are plans for adding more content like that but it's limited by people's time.

bot/resources/tags/type-hint.md
onerandomusername: Also feels worth linking pyright and other checkers.

bot/resources/tags/type-hint.md
onerandomusername: ^

onerandomusername (changes requested):
Given that type-hinting also exists, it may be worth mentioning that pydis has that channel as well? Not sure.

wookie184:
@jonathan-d-zhang are you interested in continuing this? If not i'd be happy to take over.

jonathan-d-zhang:
@wookie184 oops, I kinda forgot about this yeah, you can take over if you'd like

wookie184:
I've made a couple of changes, feel free to request changes on any of them if you think anything should be reverted to previous wording or changed further

bot/resources/tags/type-hint.md
wookie184: Included it, I also mentioned the `mypy` docs as I think that's probably one of the best places to go for an introduction to type hinting.

bot/resources/tags/type-hint.md
wookie184: I changed to link the docs. I'm not sure mentioning pyright adds much here, and a link I added mentions other type chackers.

bot/resources/tags/type-hint.md
```diff
@@ -0,0 +1,19 @@
...
+It's important to note these are just hints and are not enforced at runtime.
+
+```python
+add("hello ", "world")
+```
+The above code won't error even though it doesn't follow the function's type hints; the two strings will be concatenated as normal.
+
+Third party tools like [mypy](https://mypy.readthedocs.io/en/stable/introduction.html) can validate your code to ensure it is type hinted correctly. This can help you identify potentially buggy code, for example it would error on the second example as our `add` function is not intended to concatenate strings.
+
+[`mypy`s documentation](https://mypy.readthedocs.io/en/stable/builtin_types.html) contains useful information on type hinting, and for more information check out [this documentation page](https://typing.readthedocs.io/en/latest/index.html).
```
ChrisLovering, line 19:
```suggestion
[mypy's documentation](https://mypy.readthedocs.io/en/stable/builtin_types.html) contains useful information on type hinting, and for more information check out [this documentation page](https://typing.readthedocs.io/en/latest/index.html).
```

It's not in a code block on the line above, so shouldn't be here either. Apostrophe to indicate ownership.

ChrisLovering (changes requested):

ichard26 (approved):
STGM.

ChrisLovering (approved):

PR #1944, opened 2021-11-09
Filter the same kind of messages in both message deletion listeners.

Qwerty-133:
The `on_raw_message_delete` listener doesn't filter messages from bots or DMs unlike `on_message_delete`, which leads to unwanted logs.


ChrisLovering (approved):
🚀

Akarys42 (approved):
:shipit:

PR #1952, opened 2021-11-13
YTDL tag update

kkrypt0nn:
As per rule 5:
> Do not provide or request help **on projects** that may break laws, breach terms of services, or are malicious or inappropriate.

The current tag says you should not assist with *questions related to ytdl*. People might ask a simple question about general Python coding within a project breaking YouTube's Terms of Service, and other members provide support for the user, which breaks the rule 5.

### DIFF
--- a/bot/resources/tags/ytdl.md
+++ b/bot/resources/tags/ytdl.md
@@ -1,4 +1,4 @@
-Per [Python Discord's Rule 5](https://pythondiscord.com/pages/rules), we are unable to assist with questions related to youtube-dl, pytube, or other YouTube video downloaders, as their usage violates YouTube's Terms of Service.
+Per [Python Discord's Rule 5](https://pythondiscord.com/pages/rules), we are unable to assist with projects related to youtube-dl, pytube, or other YouTube video downloaders, as their usage violates YouTube's Terms of Service.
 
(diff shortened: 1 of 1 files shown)


JacobMonck:
Which issue is this relating to?

kkrypt0nn:
Not really any, the current tag simply doesn't really say the same as the rules.

JacobMonck:
So normally for PR's you need to make an issue so we can discuss and bounce back ideas and agree on what ends up in the repo. I believe we have some contributing guidelines in the readme of this repo.

JacobMonck:
And personally I don't see what the issue is with this but if we were to change it I would do something like "questions/projects" but that's just my opinion.

Xithrius:
Until the issue gets an approval, this PR shall not be merged.

Kronifer (changes requested):
maybe reword this to specify both questions and projects? Otherwise, I think this should be good.

onerandomusername:
> maybe reword this to specify both questions and projects? Otherwise, I think this should be good.

This edit is currently blocked until a dev actually approves it, see the comment by xith

JacobMonck:
> maybe reword this to specify both questions and projects? Otherwise, I think this should be good.

I agree

Bluenix2 (approved):

PR #1953, opened 2021-11-14
Add #bot-commands to guild features in !server

jb3:
This prevents spam in dev-contrib and dev-core from people trying to
find which Discord feature flags are enabled for Python Discord. It's
not ideal that we have to increase output size in #bot-commands but it
prevents spam in #dev-contrib.

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -178,7 +178,10 @@ class Information(Cog):
 
         # Server Features are only useful in certain channels
         if ctx.channel.id in (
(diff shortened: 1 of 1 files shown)


onerandomusername (approved):

hedyhli (approved):
This was added in #1166 by kosayoda through [suggestion](https://github.com/python-discord/bot/pull/1166#issuecomment-696386773) from Numerlor to remove the clutter and possibly unnecessary information, but I do agree we shouldn't encourage users to spam in #dev-contrib just to see this information.

jchristgit (approved):
Great Elixir code. Thanks!

Xithrius (approved):

PR #1962, opened 2021-11-17
Run black on entire repo

ChrisLovering:
This is just to see what the diff would look like and whether we would want some of these changes.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,5 +15,7 @@ ignore=
     # Docstring Content
     D400,D401,D402,D404,D405,D406,D407,D408,D409,D410,D411,D412,D413,D414,D416,D417
     # Type Annotations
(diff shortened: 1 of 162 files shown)


tox.ini
```diff
@@ -15,5 +15,7 @@ ignore=
     # Docstring Content
     D400,D401,D402,D404,D405,D406,D407,D408,D409,D410,D411,D412,D413,D414,D416,D417
     # Type Annotations
-    ANN002,ANN003,ANN101,ANN102,ANN204,ANN206
+    ANN002,ANN003,ANN101,ANN102,ANN204,ANN206,
```
onerandomusername, line 18: nit: if adding a comma at the end of this line, should they be added for the other lines? (eg L16)

ChrisLovering: Please see the PR description.

PR #1966, opened 2021-11-20
Disable File Logging By Default

HassanAbouelela:
After a vote on discord, it appears almost no one uses file logging, except in exceptional circumstances. I changed the settings so file logging would be disabled by default, and remove the volume from the docker-compose to avoid any problems, and reduce the maintenance burden. I added an environment variable to enable it.

If someone really wants to use file logging, they still can, but it will no longer be the default, nor particularly convenient.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -90,7 +90,6 @@ services:
       context: .
       dockerfile: Dockerfile
     volumes:
-      - ./logs:/bot/logs
       - .:/bot:ro
     tty: true
     depends_on:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -90,7 +90,6 @@ services:
       context: .
       dockerfile: Dockerfile
     volumes:
-      - ./logs:/bot/logs
```
onerandomusername, line 93: Should this remain here even if its optional?

onerandomusername (approved):
LGTM

One question, but not important to approval.

docker-compose.yml
HassanAbouelela: You've commented on the diff that removes it.

ChrisLovering (approved):

Akarys42 (approved):

PR #1970, opened 2021-11-22
slash commands in d.py v2 in python

Rishiraj0100:
This `PR` adds a tag on how to
make `/` commands or `context-menus` in d.py`v2.0a`


Akarys42:
Hello @Rishiraj0100,

Please leave a suggestion on [python-discord/meta](https://pydis.com/meta) before opening a PR for a new tag. Right now we don't want to show what could be perceived as open support to a discord.py library or another.

PR #1976, opened 2021-11-23
Added suppress for NotFound error

git-zephyrus:
Solves BOT-1QH
Solves #1963 
[issue reference](https://github.com/python-discord/bot/issues/1963#issue-1057499336)

### DIFF
--- a/bot/exts/moderation/clean.py
+++ b/bot/exts/moderation/clean.py
@@ -285,7 +285,8 @@ class Clean(Cog):
 
                 if len(to_delete) == 100:
                     # Only up to 100 messages can be deleted in a bulk
-                    await channel.delete_messages(to_delete)
+                    with suppress(NotFound):
+                        await channel.delete_messages(to_delete)
                     deleted.extend(to_delete)
                     to_delete.clear()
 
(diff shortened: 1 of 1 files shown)


bot/exts/moderation/clean.py
```diff
@@ -285,15 +285,17 @@ async def _delete_found(self, message_mappings: dict[TextChannel, list[Message]]
 
                 if len(to_delete) == 100:
                     # Only up to 100 messages can be deleted in a bulk
-                    await channel.delete_messages(to_delete)
+                    with suppress(NotFound):
+                        await channel.delete_messages(to_delete)
```
jchristgit, line 289: I believe in this case we can leave out the `suppress`, as we know it will be a bulk delete per `len(to_delete) == 100`, which doesn't care that the messages are absent.

jchristgit (changes requested):
Thank you for your PR!

I think we can leave out the `suppress` in one of the two spots, I left a comment there. Otherwise, this looks great.

bot/exts/moderation/clean.py
git-zephyrus: I agree.

image
My bad I missed, it was specified there.

ichard26 (approved):
SGTM, thank you!

ChrisLovering (approved):

PR #1978, opened 2021-11-27
Combine `bright_green` and `soft_green` constants to `green`

### LINKED ISSUE #1977: Match the green colors in the help system
`constants.Colours.bright_green` and `constants.Colours.soft_green` both exist, and they're [very similar](https://coolors.co/01d277-68c290). (For reference, `orange` and `soft_orange` are [nowhere near each other](https://coolors.co/e67e22-f9cb54).)

`bright_green` is only used 3 times: twice in the help channel system and once in the modpings cog. `soft_green` is used 19 times across 12 files.

I propose that we make all the green color usages the same, keeping only one and possibly renaming it to `green`.

bsoyka:
This closes #1977 by doing the following:

- Replace `bright_green` with `soft_green`.
- Rename `soft_green` to `green`.

### DIFF
--- a/config-default.yml
+++ b/config-default.yml
@@ -30,11 +30,10 @@ bot:
 style:
     colours:
         blue: 0x3775a8
-        bright_green: 0x01d277
+        green: 0x01d277
         orange: 0xe67e22
         pink: 0xcf84e0
         purple: 0xb734eb
-        soft_green: 0x68c290
         soft_orange: 0xf9cb54
         soft_red: 0xcd6d6d
         white: 0xfffffe
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -249,11 +249,10 @@ class Colours(metaclass=YAMLGetter):
     subsection = "colours"
 
     blue: int
-    bright_green: int
+    green: int
     orange: int
     pink: int
     purple: int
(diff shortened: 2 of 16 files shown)


config-default.yml
```diff
@@ -30,11 +30,10 @@ bot:
 style:
     colours:
         blue: 0x3775a8
-        bright_green: 0x01d277
+        green: 0x01d277
         orange: 0xe67e22
         pink: 0xcf84e0
         purple: 0xb734eb
-        soft_green: 0x68c290
```
TizzySaurus, line 37: I'm inclined to keep this the same name for the sake of easily referencing it with the [discordpy docs](https://discordpy.readthedocs.io/en/master/api.html#discord.Colour). We do this with all the other colours, so I'm not sure this is a good reason to strafe from that.

This is made worse by the fact that it's actually misleading (rather than just confusing having a non-existent colour) since `discord.Colour.green` is actually 0x2ecc71, not 0x68c290 (which are rather different).

TizzySaurus (commented):
You've not missed any migrations etc. just one concern I have.

We should probably do the same thing with [sir-lancebot](https://github.com/python-discord/sir-lancebot/) too.

PR #1979, opened 2021-11-27
Use bright_green for "Currently Helping" DMs

### LINKED ISSUE #1977: Match the green colors in the help system

bsoyka:
Closes #1977 by using the `bright_green` color for "Currently Helping" DM embeds.

(This is a replacement to #1978 after discussion in #dev-contrib.)

### DIFF
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -590,7 +590,7 @@ class HelpChannels(commands.Cog):
             embed = discord.Embed(
                 title="Currently Helping",
                 description=f"You're currently helping in {message.channel.mention}",
-                color=constants.Colours.soft_green,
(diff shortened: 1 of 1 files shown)


TizzySaurus (approved):
Does what it should :+1:

Xithrius (approved):

PR #1980, opened 2021-11-27
Limit the length of the invalid rule indices message

bsoyka:
This closes #1975, which shows that extremely long invalid rule indices errors, as shown below, throw an error if the message is long enough.

image

This PR limits the length of the invalid indices string to 50 characters to avoid errors like the one shown in Sentry.

image

PR #1981, opened 2021-11-27
Small changes to invalid rule indices message shortening

bsoyka:
This is a quick visual fix to add onto #1980. It adds a space before the ellipsis in the placeholder.

image


jchristgit:
This also changes the length from 50 to 75. Was that intentional?

bsoyka:
> This also changes the length from 50 to 75. Was that intentional?

@jchristgit Yes, because similar to how this is done elsewhere in the bot's code, the `:x: Invalid rule indices: ` string is now part of the `shorten` call, which adds about 25 characters.

jchristgit (approved):

ChrisLovering (approved):

PR #1982, opened 2021-11-28
Rename channels.discord_py to discord_bots

bsoyka:
Since this is such a minor change, I'm just PRing it (at the advice of a core dev in `#dev-contrib`). However, there's a slight issue of whether this might affect stats too much, so I'm CCing @jb3 here for an opinion.

(Update: Joe believes this isn't a critical stat, so it should be fine to change.)


ks129 (approved):

TizzySaurus:
Added do-not-merge label in case it messes with stats in a way we don't want.

bot/exts/moderation/slowmode.py
```diff
@@ -16,7 +16,7 @@
 
 COMMONLY_SLOWMODED_CHANNELS = {
     Channels.python_general: "python_general",
-    Channels.discord_py: "discordpy",
+    Channels.discord_bots: "discordpy",
```
ToxicKidz, line 19: We should change the channel name here as well.
```suggestion
    Channels.discord_bots: "discord_bots",
```

ToxicKidz (dismissed):
Everythings looks good, except for one suggestion.

bot/exts/moderation/slowmode.py
onerandomusername: Should probably have a comment why

bsoyka: Originally, it was to avoid messing with stats too much. Just asked @jb3 on Discord and he says we should be good to rename the stat too as it's not critical or checked much, so I'm committing this.

onerandomusername: 👍

onerandomusername (approved):

Xithrius (approved):

PR #1985, opened 2021-12-01
Redirect subscribe command output to bot commands

ChrisLovering:
Instead of silently failing in channels other than bot commands for non-staff, the bot now instead redirects the command output to bot commands and pings the user.

To facilitate this, I had to change the ctx.reply to a ctx.send since the invocation message may be in a different channel.

This shows the output from the command when ran in bot-commands and then below is the output if the user runs the command from another channel.
image

### DIFF
--- a/bot/exts/info/subscribe.py
+++ b/bot/exts/info/subscribe.py
@@ -10,9 +10,9 @@ from discord.interactions import Interaction
 
 from bot import constants
 from bot.bot import Bot
-from bot.decorators import in_whitelist
+from bot.decorators import redirect_output
 from bot.log import get_logger
-from bot.utils import checks, members, scheduling
+from bot.utils import members, scheduling
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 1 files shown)


bsoyka (approved):
Functions as it should, code is great, looks good to me!

ToxicKidz (approved):
LGTM

bot/exts/info/subscribe.py
```diff
@@ -181,18 +186,12 @@ async def subscribe_command(self, ctx: commands.Context, *_) -> None:  # We don'
             row = index // ITEMS_PER_ROW
             button_view.add_item(SingleRoleButton(role, role.role_id in author_roles, row))
 
-        await ctx.reply(
+        await ctx.send(
```
onerandomusername, line 189: this can use the reference, with a fail_if_not_exists IIRC

ChrisLovering: I see this exists https://discordpy.readthedocs.io/en/master/api.html#discord.Message.to_reference but that's for creating message references. So it could be used by adding `reference=ctx.message.to_reference(fail_if_not_exists=False),` to the ctx.send call.

However fail_if_not_exists is just for messages that are deleted/NotFound before the send could happen. The API still raises the error `discord.errors.HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body In message_reference: Cannot reply to a message in a different channel` if the message exists, but is in another channel.

onerandomusername: Huh, that's interesting, and a bit annoying, I guess.

onerandomusername (approved):

bot/exts/info/subscribe.py
ChrisLovering: > 

Agreed :D

jchristgit (approved):

PR #1989, opened 2021-12-02
Clean improvements

mbaruh:
This PR follows observations of how the new clean features were being used.

### Changes:
* All clean commands now use the clean limit (message, time delta, ISO datetime) instead of `traverse`.
* Consequently, `clean all` has been removed as `clean until` now effectively fulfills that role.
* Removes the cache usage argument from the clean commands. The cache will be used if the age of the oldest message requested for cleaning is younger than the oldest message in the cache (meaning, all of the messages we're looking for are cached).
* Along with a small speed improvement in the implementation, when specifying `"*"` for the channels the command will now delete only from public channels for further speed-up.

### Fixes:
* Iterating over the cache was done from the wrong end.
* The file was using the old logger due to a faulty merge.


ChrisLovering (approved):
Some very nice code Zig 👍 I really like the change you made to the cache, should make this much nicer to use :D

bot/exts/moderation/clean.py
```diff
@@ -87,17 +85,16 @@ def mod_log(self) -> ModLog:
...
             bots_only: bool,
             users: Optional[list[User]],
             first_limit: Optional[CleanLimit],
             second_limit: Optional[CleanLimit],
     ) -> None:
         """Raise errors if an argument value or a combination of values is invalid."""
-        # Is this an acceptable amount of messages to traverse?
-        if traverse > CleanMessages.message_limit:
-            raise BadArgument(f"Cannot traverse more than {CleanMessages.message_limit} messages.")
+        if first_limit is None:
```
Shivansh-007, line 95: We could remove the check on line 111 as `first_limit` is now a required argument, so it wouldn't be `None` (after this check).

mbaruh: True

Shivansh-007 (approved):
Very cool 👍🏻

mbaruh:
Just noticed a bug this PR is going to introduce, will fix soon.

ChrisLovering (approved):
Tested latest changes 👍

bot/exts/moderation/clean.py
```diff
@@ -110,23 +107,46 @@ def _validate_input(
...
+            elif isinstance(second_limit, Message):
+                channels = {second_limit.channel}
+            else:
+                channels = {ctx.channel}
+        else:
+            if channels == "*":
+                channels = {
+                    channel for channel in ctx.guild.channels
+                    if isinstance(channel, TextChannel)
+                    # Assume that non-public channels are not needed to optimize for speed.
```
Bluenix2, line 135: I understand why it's done, but what happens if we have "public" channels that are still locked behind a role such as during codejams.

mbaruh: The participants channel stays public, just not writeable. If we get to the point of cleaning a user from the team leaders channel, I think a more focused action will be made anyway, likely including the disqualification of the participant. Specifying such a channel explicitly in that case seems like an acceptable edge case. And if it stops being acceptable we can consider adding another shortcut for actually-all channels.

Bluenix2 (approved):
From my very limited experience with the clean cog... it works! Great thank you my favourite peppa-pig look-alike. Nice work here
high5

PR #1990, opened 2021-12-03
Patch d.py's message convertor to infer channelID from the given context

ChrisLovering:
Discord.py's Message convertor is supposed to infer channelID based on ctx.channel if only a messageID is given. A 'refactor' (linked below) a few weeks before d.py's archival broke this, so that if only a messageID is given to the convertor, it will only find that message if it's in the bot's cache.

### DIFF
--- a/bot/__init__.py
+++ b/bot/__init__.py
@@ -18,6 +18,11 @@ if os.name == "nt":
 
 monkey_patches.patch_typing()
 
+# This patches any convertors that use PartialMessage, but not the PartialMessageConverter itself
+# as library objects are made by this mapping.
+# https://github.com/Rapptz/discord.py/blob/1a4e73d59932cdbe7bf2c281f25e32529fc7ae1f/discord/ext/commands/converter.py#L984-L1004
+commands.converter.PartialMessageConverter = monkey_patches.FixedPartialMessageConverter
+
 # Monkey-patch discord.py decorators to use the Command subclass which supports root aliases.
 # Must be patched before any cogs are added.
 commands.command = partial(commands.command, cls=monkey_patches.Command)
--- a/bot/monkey_patches.py
+++ b/bot/monkey_patches.py
@@ -5,6 +5,7 @@ from discord import Forbidden, http
 from discord.ext import commands
 
 from bot.log import get_logger
+from bot.utils.regex import MESSAGE_ID_RE
 
 log = get_logger(__name__)
 
@@ -50,3 +51,25 @@ def patch_typing() -> None:
             pass
 
     http.HTTPClient.send_typing = honeybadger_type
+
+
+class FixedPartialMessageConverter(commands.PartialMessageConverter):
+    """
+    Make the Message converter infer channelID from the given context if only a messageID is given.
+
+    Discord.py's Message converter is supposed to infer channelID based
+    on ctx.channel if only a messageID is given. A refactor commit, linked below,
+    a few weeks before d.py's archival broke this defined behaviour of the converter.
+    Currently, if only a messageID is given to the converter, it will only find that message
+    if it's in the bot's cache.
+
+    https://github.com/Rapptz/discord.py/commit/1a4e73d59932cdbe7bf2c281f25e32529fc7ae1f
+    """
+
+    @staticmethod
+    def _get_id_matches(ctx: commands.Context, argument: str) -> tuple[int, int, int]:
+        """Inserts ctx.channel.id before calling super method if argument is just a messageID."""
+        match = MESSAGE_ID_RE.match(argument)
+        if match:
+            argument = f"{ctx.channel.id}-{match.group('message_id')}"
+        return commands.PartialMessageConverter._get_id_matches(ctx, argument)
--- a/bot/utils/regex.py
+++ b/bot/utils/regex.py
@@ -12,3 +12,4 @@ INVITE_RE = re.compile(
     r"(?P<invite>[a-zA-Z0-9\-]+)",                # the invite code itself
     flags=re.IGNORECASE
 )
+MESSAGE_ID_RE = re.compile(r'(?P<message_id>[0-9]{15,20})$')


onerandomusername (commented):
Why is this convert**or** when the class itself is named convert**er**?

HassanAbouelela (approved):
I know this doesn't count, but I can't help myself. You gotta do what you gotta do when you see beautiful code.

ChrisLovering:
> Why is this convert**or** when the class itself is named convert**er**?

Because those are the ones I manually typed out, rather than copy-pasted, and I always spell it **or** for some reason, ignoring the difference. Force pushed a fix though.

bot/__init__.py
```diff
@@ -17,6 +17,7 @@
     asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
 
 monkey_patches.patch_typing()
+monkey_patches.patch_message_convertor()
```
onerandomusername, line 20: method no longer exists (e)

ChrisLovering: Pushed

onerandomusername:
> > Why is this convert**or** when the class itself is named convert**er**?
> 
> Because those are the ones I manually typed out, rather than copy-pasted, and I always spell it **or** for some reason, ignoring the difference. Force pushed a fix though.
Yeah, makes sense. Missed one tho :P

bot/monkey_patches.py
```diff
@@ -1,3 +1,5 @@
+import typing as t
+import unittest.mock
```
onerandomusername, line 2: uh, why exactly are we using unittest in production?

onerandomusername: I suppose I'm asking, is there a way to do what we need to do without mocking it?

bot/monkey_patches.py
```diff
@@ -50,3 +52,36 @@ async def honeybadger_type(self, channel_id: int) -> None:  # noqa: ANN001
...
+        try:
+            value = data[key]
+        except KeyError:
+            return None
+        else:
+            return value and int(value)
+
+    def patch(ctx: commands.Context, argument: str) -> t.Callable:
+        """Infer channel ID from context if not explicitly specified."""
+        with unittest.mock.patch("discord.utils._get_as_snowflake") as mock:
```
onerandomusername, line 82: its possible to patch with a provided object, rather than patching with a mock.

we can provide the new= param, https://docs.python.org/3/library/unittest.mock.html?highlight=unittest%20mock#patch

This means that we don't need to use a mock at all, which should save making a useless objects (and all of its internal workings, unittest isn't exactly known for being fast in prod)

bot/monkey_patches.py
```diff
@@ -50,3 +52,36 @@ async def honeybadger_type(self, channel_id: int) -> None:  # noqa: ANN001
...
+    log.debug("Patching Message converter to infer channelID from context.")
+    original = commands.PartialMessageConverter._get_id_matches
+
+    def _get_as_snowflake(data: t.Any, key: str) -> t.Optional[int]:
+        try:
+            value = data[key]
+        except KeyError:
+            return None
+        else:
+            return value and int(value)
```
onerandomusername, line 78: this is identical to what we're mocking, so I don't understand the point?

bot/monkey_patches.py
```diff
@@ -50,3 +52,36 @@ async def honeybadger_type(self, channel_id: int) -> None:  # noqa: ANN001
             pass
 
     http.HTTPClient.send_typing = honeybadger_type
+
+
+def patch_message_converter() -> None:
```
onerandomusername, line 57: nit: this is only PartialMessageConvertor

onerandomusername: nvm, found the relevant line:

https://github.com/Rapptz/discord.py/blob/1a4e73d59932cdbe7bf2c281f25e32529fc7ae1f/discord/ext/commands/converter.py#L386

bot/monkey_patches.py
```diff
@@ -50,3 +52,36 @@ async def honeybadger_type(self, channel_id: int) -> None:  # noqa: ANN001
...
+    """
+    Make the Message converter infer channelID from the given context.
+
+    Discord.py's Message converter is supposed to infer channelID based
+    on ctx.channel if only a messageID is given. A "refactor" (linked below)
+    a few weeks before d.py's archival broke this, so that if only a messageID
+    is given to the converter, it will only find that message if it's in the bot's
+    cache.
+
+    https://github.com/Rapptz/discord.py/commit/1a4e73d59932cdbe7bf2c281f25e32529fc7ae1f
```
onerandomusername, line 67: I don't exactly see what line is patching this, given I can't find an await to fetch messages from the provided channel. If they aren't in the cache, how would not fetching help? Serious question, because I can't exactly see where this solves the problem.

onerandomusername (changes requested):
I'm a bit confused as to how this patch is working, I can't find the line that fixes a message being found if it isn't in the cache?

bot/monkey_patches.py
HassanAbouelela: It's simpler than the alternative, but mocking in the main application is not necessarily a bad thing if used properly, See my other comments on this.

bot/monkey_patches.py
HassanAbouelela: I don't think using mocks here is really a problem. This function is only called when something explicitly calls the message converter, and I struggled to find anything that does to test this function with. Mind you that the message object in context, and the message passed to `on_message` do not go through this. The only thing that goes through this are commands which accept a message as their arg.

bot/monkey_patches.py
HassanAbouelela: No? We're mocking a function which calls this in the middle because we want to have slightly different behavior (see the side effect on mock). We duplicated the base function because we need to call it in the side effect, and this was the easiest way.

bot/monkey_patches.py
onerandomusername: The above method would result in a performance boost from the current code. I feel like mocks are a bit slower than other objects, given their internals.

HassanAbouelela: You're talking performance boosts that will show benefits if we're calling this a 1000 times a second, but that's simply never going to be the case. We may as well do it, but honestly, it should be fine even if we don't.

bot/monkey_patches.py
ChrisLovering: This patches the Message converter, not fetching a message from the API. This is the doc string for Discord.py's message converter https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/converter.py#L373-L386.

The second step there doesn't actually work. It doesn't fall back to the context channel if just a messageID is given, it actually only checks the cache.

Line 84 of this PR is what is actually patching this behaviour, where if _get_as_snowflake doesn't return a channel, we use the channel from the context instead, so that the message converter actually works as intended.

bot/monkey_patches.py
ChrisLovering: Since this function is the one we're mocking, and within the mock we calling it first, before inferring, we can't have the mock side effect call the function it's mocking, so we copied the function instead.

ChrisLovering:
@onerandomusername @HassanAbouelela I think I've arrived at a nicer solution for this, which still works from my testing. If you wanted a simple way to test this, the `!raw` command using a message convertor.

bot/monkey_patches.py
```diff
@@ -50,3 +51,26 @@ async def honeybadger_type(self, channel_id: int) -> None:  # noqa: ANN001
...
+    """
+
+    @staticmethod
+    def _get_id_matches(ctx: commands.Context, argument: str) -> tuple[int, int, int]:
+        """Inserts ctx.channel.id before calling super method if argument is just a messageID."""
+        message_id_regex = re.compile(r'(?P<message_id>[0-9]{15,20})$')
+        match = message_id_regex.match(argument)
+        if match:
+            argument = f"{ctx.channel.id}-{match.group('message_id')}"
+        return commands.PartialMessageConverter._get_id_matches(ctx, argument)
```
onerandomusername, line 76: nit: `super()`?

onerandomusername (changes requested):
After much testing, I can now say that this works, and I like the implementation as well.

bot/monkey_patches.py
```diff
@@ -50,3 +51,26 @@ async def honeybadger_type(self, channel_id: int) -> None:  # noqa: ANN001
             pass
 
     http.HTTPClient.send_typing = honeybadger_type
+
+
+class FixedPartialMessageConverter(commands.PartialMessageConverter):
+    """
+    Make the Message converter infer channelID from the given context if only a messageID is given.
```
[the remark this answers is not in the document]
ChrisLovering: Yea, I guess it's technically both, since the MessageConverter just uses util functions from the PartialMessageConverter.

I'm going to say that this word message refers to the generic message and covers both, don't think it's something that needs changing :D

bot/monkey_patches.py
```diff
@@ -50,3 +51,26 @@ async def honeybadger_type(self, channel_id: int) -> None:  # noqa: ANN001
...
+
+    Discord.py's Message converter is supposed to infer channelID based
```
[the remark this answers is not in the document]
onerandomusername: resolved by https://github.com/python-discord/bot/pull/1990#discussion_r762135937

bot/monkey_patches.py
ChrisLovering: Since this is a staticmethod, I'd likely need to do `return super(commands.PartialMessageConverter,commands.PartialMessageConverter)._get_id_matches(ctx, argument)` for that to work right?

bot/__init__.py
```diff
@@ -18,6 +18,11 @@
 
 monkey_patches.patch_typing()
 
+# This patches any convertors that use PartialMessage, but not the PartialMessageConverter itself
+# as library objects are made by this mapping.
+# https://github.com/Rapptz/discord.py/blob/1a4e73d59932cdbe7bf2c281f25e32529fc7ae1f/discord/ext/commands/converter.py#L984-L990
```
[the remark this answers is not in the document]
ChrisLovering: Fixed in a forcepush

bot/monkey_patches.py
```diff
@@ -50,3 +51,26 @@ async def honeybadger_type(self, channel_id: int) -> None:  # noqa: ANN001
...
+    Currently, if only a messageID is given to the converter, it will only find that message
+    if it's in the bot's cache.
+
+    https://github.com/Rapptz/discord.py/commit/1a4e73d59932cdbe7bf2c281f25e32529fc7ae1f
+    """
+
+    @staticmethod
+    def insert_channel_id(ctx: commands.Context, argument: str) -> tuple[int, int, int]:
+        """Inserts ctx.channel.id before calling super method if argument is just a messageID."""
+        message_id_regex = re.compile(r'(?P<message_id>[0-9]{15,20})$')
```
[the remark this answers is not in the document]
ChrisLovering: yea, lets actually use the regex util file for this :D Fixed in a forcepush

bot/monkey_patches.py
onerandomusername: Ah, yes, yes you would. That would make no sense in this case.

onerandomusername (approved):
finally, LGTM!

jchristgit (approved):

PR #1997, opened 2021-12-08
Reduce threshold before fuzzy matching to 2

ChrisLovering:
Commands such as !ot, !if, !xy are commonly used as shortcuts to their respective tags. We recently upped the threshold before fuzzy matching to 3 characters, which broke these shortcuts.

This commit reduces that threshold down to 2, so users who are used to those commands can still use them.

### DIFF
--- a/bot/exts/info/tags.py
+++ b/bot/exts/info/tags.py
@@ -168,11 +168,11 @@ class Tags(Cog):
         """Get tags with identifiers similar to `tag_identifier`."""
         suggestions = []
 
-        if tag_identifier.group is not None and len(tag_identifier.group) >= 3:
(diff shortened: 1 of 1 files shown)


Numerlor:
Should probably change it on L175 too so it also works as a direct invocation without any arguments (words after it)

ChrisLovering:
> Should probably change it on L175 too so it also works as a direct invocation without any arguments (words after it)

Ah yea, good point. Didn't notice that one. Updated now.

Numerlor (approved):

Akarys42 (approved):

vcokltfre (approved):

PR #1999, opened 2021-12-09
Remove deprecated server voice region

PH-KDX:
Discord's current model for voice regions is setting it per server. Hence, the "Voice region" section in the server info tag will always display as "deprecated". This pull request removes it.


eivl (approved):
I approve, the voice server info is deprecated.

ChrisLovering (approved):

PR #2000, opened 2021-12-09
Strip gotcha tag

Shom770:
This pull request adds a strip-gotcha tag, which represents the "gotcha" that a lot of people face with `strip` where they might expect it to remove a suffix/prefix and that it interprets its argument like a substring when it interprets it like a set of characters.

An example:
```py
>>> "Monty Python".strip(" Python")
"M"
```

<h3>Changes</h3>

- Adds a strip-gotcha tag


Akarys42:
image

jchristgit (approved):
Cool and good tag. Ran into this a bunch of times myself.

vcokltfre:
May be worth noting that removesuffix is a fairly recent addition

Shom770:
> May be worth noting that removesuffix/removeprefix is a fairly recent addition

How about under the doc links it says "Note: `str.removeprefix` and `str.removesuffix` were added in Python 3.9 and are fairly new"

mbaruh:
The embed already says they were added in 3.9 though

onerandomusername (commented):
What about telling the author how to do it if they're using 3.8 or lower?

Shom770:
> The embed already says they were added in 3.9 though

Oh wait, I missed it while reading

Shom770:
> What about telling the author how to do it if they're using 3.8 or lower?

Hmm, that's mentioned in the documentation I believe, and that might make the embed a little long. How about adding that next to the footnote at the bottom?

Issue #1421, opened 2021-02-18
MFA User Token filter


fisher60:
Recently a user posted a valid user token and we realized we do not have a filter for this. I believe it would be simple enough to implement this check and worth it to protect users that, for *whatever reason* make this mistake.

While there is no valid reason I can think of for a user to have their token in their clipboard, I still think it is important to protect our users. This could also potentially aid in moderating selfbotting discussions in the very off chance a user accidentally posts a user token when asking for help.

A far too simple implementation/example of this would be to add something like:
 ```USER_TOKEN_RE = re.compile(r"^([a-z]{3})\.([\w\-=]+)", re.ASCII)``` 
 However this is too simple and needs to be refined.

Xithrius:
@fisher60 would you like to implement this?

bast0006:
We've got rudimentary detection for regular user tokens courtesy of the work I did (a few months ago). However, it only works on regular tokens (of which bot tokens are a part of). Users who have multi-factor authentication on have a different token format. 

As far as I'm aware, they always start with `mfa` although I believe I've sometimes seen that capitalized. But no other letter pattern. We may or may not want to be strict with this? I'm pretty sure they always have two underscores within them as well, and I suspect that corresponds to the original JWT-like format as a divider of some kind but it wasn't immediately obvious to me how it correlated with the regular token format (ID.timestamp.hash).

bast0006:
From a cursory google search (ugh) token grabbers use this regex: `mfa\.[\w-]{84}`.

Akarys42:
We do grab some forms of tokens at the moment (bearer tokens, I think), but not mfa tokens, we should definitely do that.

Xithrius:
@bast0006 was this eventually taken care of, or do we need to make a PR? As far as I saw, we just added another regex filter.

Akarys42:
Sounds like we just need a new regex in the token filter cog, yes.

bast0006:
> @bast0006 was this eventually taken care of, or do we need to make a PR? As far as I saw, we just added another regex filter.

We've got a regex registered as a custom filter, but I think it'd be a good idea to immortalize it within the filter_token cog

Xithrius:
Alright, I'll shove it into that cog.

Akarys42:
This may get replaced by #1530, what do you think @Xithrius ?

Xithrius:
I completely forgot this issue existed, and yes, I think this issue should be replaced by that one in the end.

[...]

PR #2006, opened 2021-12-12
enhancement(filters): use a stricter bot token regex

onerandomusername:
Approval: https://canary.discord.com/channels/267624335836053506/635950537262759947/919203386660884560

Enhances the regex of the token remover to use the same regex that discord itself uses, with a slight modification. The `mfa` section was removed, but depending on an updated #1421, may be implemented. Additionally, the sections were grouped to keep working with the current code.

I kept the existing validation to keep false positives at a minimum. The current code checks the user resolves, the timestamp is valid, and the last section has at least 3 different characters.

As per @jb3, to implement:
- ~~#1421~~ now irrelevant.

Rejected:
- also check the upper bound of a token timestamp, to ensure it was in the past.


onerandomusername:
I'll be fixing this commit history tonight

bot/exts/filters/token_remover.py
```diff
@@ -78,6 +83,9 @@ async def on_message(self, msg: Message) -> None:
         if found_token:
             await self.take_action(msg, found_token)
```
Bluenix2, line 84: We could add a return after this - there's no point in checking for an MFA token.

Bluenix2 (approved):
Works as usual - I can't get access to an MFA token to test though. Thanks for this, I just have one comment that won't affect my review.

bot/exts/filters/token_remover.py
onerandomusername: I think it is worth checking, or I should move checking for an mfa token before checking the other token format.

the reason for checking both was to ensure that a user who happened to share a user and bot token in the same message would not go undetected.

onerandomusername: Basically, with an mfa token those are to alert mods that someone sent it,  and don't let the user know why their messages was deleted intentionally.

however, I should probably go a little further with the implementation and possibly add some configuration as we may not want to ping moderators for mfa tokens

bot/exts/filters/token_remover.py
```diff
@@ -167,7 +219,7 @@ def find_token_in_message(cls, msg: Message) -> t.Optional[Token]:
                 return token
 
         # No matching substring
-        return
+        return None
```
Akarys42, line 222: Why this change?

Akarys42 (approved):
LGTM, just a small style change

bot/exts/filters/token_remover.py
onerandomusername: As the function can return a value, I wanted to make it concretely return None while I was here, rather than implicitly return None.

bot/exts/filters/token_remover.py
Bluenix2: *Reading back my mails, I realize that it might be best for me to respond to this..*

Makes sense yeah, we'll keep it like this then.

bot/exts/filters/token_remover.py
```diff
@@ -116,6 +124,50 @@ async def take_action(self, msg: Message, found_token: Token) -> None:
...
+
+        token = match.group()
+        # since the token was probably valid, we can now delete the message and ping the moderators.
+        # the user is not informed, given the reasoning for deleting the message.
+        with contextlib.suppress(NotFound):
+            await msg.delete()
+
+        log_message = (
+            f"Deleted mfa token sent by {msg.author} in {msg.channel}: "
+            f"{token[:4]}{'x' * len(token[3:-3])}{token[-3:]}"
```
wookie184, line 154:
```suggestion
            f"{token[:4]}{'x' * len(token[4:-3])}{token[-3:]}"
```
Just so the number of `x`s matches how many characters are missing :)

bot/exts/filters/token_remover.py
```diff
@@ -116,6 +124,50 @@ async def take_action(self, msg: Message, found_token: Token) -> None:
...
+        Due to how the mfa tokens work, there is no user information supplied for this token.
+        """
+        # Ignore DMs; can't delete messages in there anyway.
+        if not msg.guild or msg.author.bot:
+            return
+
+        was_valid = False
+        for match in MFA_TOKEN_RE.finditer(msg.content):
+
+            if self.is_maybe_valid_hmac(match.group()):
```
wookie184, line 140: `match.group()` here will include `mfa.` so this will always return true, only the characters after should be passed here.

wookie184 (dismissed):
Just a couple things, could you also make the message match the other token message a bit closer, particularly:
- having the user as a clickable mention and the ID in a codeblock afterwards
- having the channel as a clickable mention
- putting the censored token in a codeblock
- adding the pfp of the user that send the message as a thumbnail

bot/exts/filters/token_remover.py
onerandomusername: https://github.com/python-discord/bot/pull/2006/commits/8fc503305cde2bf55a22a389bfdce1f7b01b6903

wookie184:
This has been inactive for a while so I'll put it up for grabs.

gustavwilliam:
> This has been inactive for a while so I'll put it up for grabs.

What's left to do? Addressing your comment and fixing the merge conflict?

mbaruh:
> > This has been inactive for a while so I'll put it up for grabs.
> 
> What's left to do? Addressing your comment and fixing the merge conflict?

Yeah, merge main, match the output embed to the existing one, and fix the bug

onerandomusername:
As far as I can tell the mfa code is no longer necessary as discord doesn't have mfa tokens like that now.

onerandomusername:
There's also a bit of the matter that this regex is now out of date.

wookie184:
> There's also a bit of the matter that this regex is now out of date. I don't know what the current token length is.

I can't recall any problems with false positives, even with the current regex, so we can afford to be very general with what we match. Could just check that the first and last parts are >= 10 and middle is >= 5 characters in length or something.

Akarys42:
I agree with wookie, I think having extensive matching for the sake of even future-proofing is okay, considering additional checks are being performed

wookie184:
> The only inaccurate part is the last section is too short as it is, but because of how the parsing works, that isn't an issue, i suppose
> 

If that's the case I think we should just have no upper limit on the length of the last section. Otherwise the code is somewhat misleading in how it works.

mbaruh:
Hey @onerandomusername, thanks for your work so far. We needed to get this merged so went ahead and implemented the comments. Sorry the MFA part didn't end up being used 😅

bot/exts/filters/token_remover.py
```diff
@@ -40,7 +40,7 @@
 # The HMAC isn't parsed further, but it's in the regex to ensure it at least exists in the string.
 # Each part only matches base64 URL-safe characters.
 # These regexes were taken from discord-developers, which are used by the client itself.
-TOKEN_RE = re.compile(r"([a-z0-9_-]{23,28})\.([a-z0-9_-]{6,7})\.([a-z0-9_-]{27})", re.IGNORECASE)
+TOKEN_RE = re.compile(r"([\w_-]{10,})\.([\w_-]{5,})\.([\w_-]{10,})", re.IGNORECASE)
```
onerandomusername: way too lenient, this makes *impossible* tokens match.

HassanAbouelela: Way too lenient is not defined by what it can match, but what it will match in practice, which so far has not been an issue. This PR made the regex a little more flexible in terms of false-positives, without increasing the possibility of false-negatives or creating something that has to be modified in the future.

PR #2001, opened 2021-12-10
Modpings Improvements

Shivansh-007:
Fixes BOT-1RG
Fixes BOT-1RD

Closes #2002 
Closes #2003 

### Changes:

- Add a better datetime converter that allows passing time in 12/24 hour format, the previous method of parsing `start` and `end` time is removed (`8h`, `16h`, etc).
- If your modpings schedule is over and you even had a modpings off which got over in modpings off time period, the role won't be added to you as your schedule is over.
- Make `modpings schedule` message more verbose.

### Fixes:

- Make `16` hours as the max off time i.e. make `8` hours the minimum on time.
- Since discord.py moved to timezone aware datetimes, the discord timestamp converter failed, therefore giving no confirmation message, this has now been fixed.
- `add_roles` is a method on a member object, thus calling `add_roles` on a user object would fail, this has now been fixed and it uses `get_or_fetch_member` utility function.

PR #2008, opened 2021-12-12
fix: check for the modteam role before reassigning the user's moderation role

### LINKED ISSUE #2007: modpings: role is assigned even if the user is no longer a mod
### Description
There are no checks if a user is still a mod when reassigning the mod role.

If a moderator is demoted, the modpings cog will still readd the moderator role, if the user had it scheduled to be readded later.

image

### Proposed Solution(s)

1: Check if the user has the modteam role immediately before reassigning the ping role.
2: Sync the mod pings on all role edits, and on load.

The best solution may be a combination of the above.

onerandomusername:
closes #2007

### DIFF
--- a/bot/exts/moderation/modpings.py
+++ b/bot/exts/moderation/modpings.py
@@ -116,16 +116,25 @@ class ModPings(Cog):
             log.trace(f"Skipping adding moderator role to mod with ID {mod.id} - found in pings off cache.")
         else:
             log.trace(f"Applying moderator role to mod with ID {mod.id}")
-            await mod.add_roles(self.moderators_role, reason="Moderator scheduled time started!")
+            await self.reapply_role(mod, reason="Moderator scheduled time started!")
 
         log.trace(f"Sleeping for {work_time} seconds, worktime for mod with ID {mod.id}")
         await asyncio.sleep(work_time)
         await self.remove_role_schedule(mod, work_time, schedule_start)
 
-    async def reapply_role(self, mod: Member) -> None:
+    async def reapply_role(self, mod: Member, *, reason: str = None) -> None:
         """Reapply the moderator's role to the given moderator."""
+        if self.guild.get_role(Roles.mod_team) not in mod.roles:
+            log.info(
+                f"Preventing {mod} from being assigned the moderator role"
+                "since they are missing the mod team role."
+            )
+            # the code that calls this method either prechecks the moderator's roles
+            # or it doesn't matter since this was called in a task
+            # the logging statement above is enough to ensure that this was caught
+            return
         log.trace(f"Re-applying role to mod with ID {mod.id}.")
-        await mod.add_roles(self.moderators_role, reason="Pings off period expired.")
+        await mod.add_roles(self.moderators_role, reason=reason or "Pings off period expired.")
         await self.pings_off_mods.delete(mod.id)
 
     @group(name='modpings', aliases=('modping',), invoke_without_command=True)
@@ -184,9 +193,7 @@ class ModPings(Cog):
             await ctx.send(":question: You already have the role.")
             return
 
-        await mod.add_roles(self.moderators_role, reason="Pings off period canceled.")
-
-        await self.pings_off_mods.delete(mod.id)
+        await self.reapply_role(mod, reason="Pings off period canceled.")
 
         # We assume the task exists. Lack of it may indicate a bug.
         self._role_scheduler.cancel(mod.id)


jchristgit (approved):

bot/exts/moderation/modpings.py
```diff
@@ -116,16 +116,25 @@ async def add_role_schedule(self, mod: Member, work_time: int, schedule_start: d
...
+            await self.reapply_role(mod, reason="Moderator scheduled time started!")
 
         log.trace(f"Sleeping for {work_time} seconds, worktime for mod with ID {mod.id}")
         await asyncio.sleep(work_time)
         await self.remove_role_schedule(mod, work_time, schedule_start)
 
-    async def reapply_role(self, mod: Member) -> None:
+    async def reapply_role(self, mod: Member, *, reason: str = None) -> None:
         """Reapply the moderator's role to the given moderator."""
+        if self.guild.get_role(Roles.mod_team) not in mod.roles:
```
mbaruh, line 127: As far as I can see you have no guarantee that `self.guild` was already set. A possible solution is to set the value in `reschedule_modpings_schedule` as in `reschedule_roles`.

mbaruh (changes requested):

bot/exts/moderation/modpings.py
ChrisLovering: The other option is to go with the async init_cog pattern we have in a few other places, where we create a task to init the cog within the cog init and then await that task in all the functions that rely on things like this.

onerandomusername: e1a580d6

bot/exts/moderation/modpings.py
```diff
@@ -50,11 +50,17 @@ def __init__(self, bot: Bot):
             name="mod-pings-reschedule",
             event_loop=self.bot.loop,
         )
+        self.bot.loop.create_task(self.fetch_guild())
```
mbaruh, line 53: Might be a nitpick, but this seems like a task that should come before the other two. Also notice that the task is still hanging on cog unload.

onerandomusername: I think the task can be hanging? It's not a looping task, it's one and done. Not entire sure how I would solve this.

onerandomusername: https://github.com/python-discord/bot/pull/2008/commits/188635730fb65bf8ebd6d7f1af154bfbf5b815ad

bot/exts/moderation/modpings.py
```diff
@@ -51,10 +52,15 @@ def __init__(self, bot: Bot):
...
-        """Reschedule moderators role re-apply times."""
+    async def fetch_guild(self) -> None:
+        """Fetch the guild."""
         await self.bot.wait_until_guild_available()
         self.guild = self.bot.get_guild(Guild.id)
+
+    async def reschedule_roles(self) -> None:
+        """Reschedule moderators role re-apply times."""
+        if not self.guild:
+            await self.fetch_guild()
```
ChrisLovering, line 63: Just make this `await self.fetch_guild_task`

onerandomusername: TIL

onerandomusername: 1dbedb2e

bot/exts/moderation/modpings.py
```diff
@@ -116,16 +121,26 @@ async def add_role_schedule(self, mod: Member, work_time: int, schedule_start: d
...
         log.trace(f"Sleeping for {work_time} seconds, worktime for mod with ID {mod.id}")
         await asyncio.sleep(work_time)
         await self.remove_role_schedule(mod, work_time, schedule_start)
 
-    async def reapply_role(self, mod: Member) -> None:
+    async def reapply_role(self, mod: Member, *, reason: str = None) -> None:
         """Reapply the moderator's role to the given moderator."""
+        await self.fetch_guild_task
+        if self.guild.get_role(Roles.mod_team) not in mod.roles:
+            log.info(
```
TizzySaurus, line 134: I think this should be upgraded to a `log.warning`, since it means we likely need to `int e` remove them from the modpings.

bot/exts/moderation/modpings.py
```diff
@@ -116,16 +121,26 @@ async def add_role_schedule(self, mod: Member, work_time: int, schedule_start: d
...
+                f"Preventing {mod} from being assigned the moderator role"
+                "since they are missing the mod team role."
+            )
+            # the code that calls this method either prechecks the moderator's roles
+            # or it doesn't matter since this was called in a task
```
TizzySaurus, line 139: Should probably have a `.` at the end of this line (meaning next line will need a capital letter)

bot/exts/moderation/modpings.py
```diff
@@ -116,16 +121,26 @@ async def add_role_schedule(self, mod: Member, work_time: int, schedule_start: d
...
-    async def reapply_role(self, mod: Member) -> None:
+    async def reapply_role(self, mod: Member, *, reason: str = None) -> None:
         """Reapply the moderator's role to the given moderator."""
+        await self.fetch_guild_task
+        if self.guild.get_role(Roles.mod_team) not in mod.roles:
+            log.info(
+                f"Preventing {mod} from being assigned the moderator role"
+                "since they are missing the mod team role."
+            )
+            # the code that calls this method either prechecks the moderator's roles
```
TizzySaurus, line 138: Nitpick but comments should generally start with a capital letter

bot/exts/moderation/modpings.py
```diff
@@ -51,10 +52,14 @@ def __init__(self, bot: Bot):
             event_loop=self.bot.loop,
         )
 
-    async def reschedule_roles(self) -> None:
-        """Reschedule moderators role re-apply times."""
+    async def fetch_guild(self) -> None:
+        """Fetch the guild."""
```
TizzySaurus, line 56: Maybe specify that it's PyDis we're getting? Probably not needed but could be nice. 

I'd also add that we assign it to `self.guild`.

"Fetch the Python Discord guild and assign to \`self.guild\`."

bot/exts/moderation/modpings.py
```diff
@@ -51,10 +52,14 @@ def __init__(self, bot: Bot):
...
         await self.bot.wait_until_guild_available()
         self.guild = self.bot.get_guild(Guild.id)
+
+    async def reschedule_roles(self) -> None:
+        """Reschedule moderators role re-apply times."""
```
TizzySaurus, line 61: Could this perhaps be reworded a bit as this doesn't make much sense to me.

TizzySaurus (changes requested):
A few nitpicks. Main one I'd like to see implemented is the `log.warning`

bot/exts/moderation/modpings.py
onerandomusername: I believe this is out of scope of this pr, and should be changed in GH-2001

bot/exts/moderation/modpings.py
onerandomusername: https://github.com/python-discord/bot/pull/2008/commits/1266c5321253da1817d1550efb8008000901a64d

bot/exts/moderation/modpings.py
onerandomusername: https://github.com/python-discord/bot/pull/2008/commits/1266c5321253da1817d1550efb8008000901a64d

bot/exts/moderation/modpings.py
onerandomusername: https://github.com/python-discord/bot/pull/2008/commits/1266c5321253da1817d1550efb8008000901a64d

bot/exts/moderation/modpings.py
onerandomusername: https://github.com/python-discord/bot/pull/2008/commits/1266c5321253da1817d1550efb8008000901a64d

bot/exts/moderation/modpings.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, bot: Bot):
         )
 
     async def fetch_guild(self) -> None:
-        """Fetch the guild."""
+        """Fetch the configured constants.Guild.id guild object and assign to self.guild."""
```
TizzySaurus, line 56: Really picky but variables should go in backquotes to indicate that it's a variable (this would apply to both `constants.Guild.id` and `self.guild`.

TizzySaurus (commented):
Haven't tested anything, but aside from the backquote comment the code looks good :+1:

bot/exts/moderation/modpings.py
TizzySaurus: Yeah, agreed. Have done left a comment on that PR :+1:

bot/exts/moderation/modpings.py
```diff
@@ -116,16 +121,27 @@ async def add_role_schedule(self, mod: Member, work_time: int, schedule_start: d
...
         log.trace(f"Sleeping for {work_time} seconds, worktime for mod with ID {mod.id}")
         await asyncio.sleep(work_time)
         await self.remove_role_schedule(mod, work_time, schedule_start)
 
-    async def reapply_role(self, mod: Member) -> None:
+    async def reapply_role(self, mod: Member, *, reason: str = None) -> None:
         """Reapply the moderator's role to the given moderator."""
+        await self.fetch_guild_task
+        if self.guild.get_role(Roles.mod_team) not in mod.roles:
+            log.warning(
```
ChrisLovering, line 134: I don't think this needs to raise a sentry issue.

Either it can be just an info log, or it could send a message to Channels.mod_meta.

ChrisLovering: Thinking on this more, we should remove them from the cache if they don't have the mod role too, otherwise this will just be hit on the next time the cog starts.

IMO this would make sense to be info log, remove from cache & message in mod-meta

onerandomusername: I think that is resolved with the rest of the enhancements in #2001.

onerandomusername: I can change this down to a info for now, and then in 2001 that can be changed.

bot/exts/moderation/modpings.py
```diff
@@ -37,19 +37,25 @@ def __init__(self, bot: Bot):
         self.bot = bot
         self._role_scheduler = Scheduler("ModPingsOnOff")
         self._modpings_scheduler = Scheduler("ModPingsSchedule")
+        self.fetch_guild_task = self.bot.loop.create_task(self.fetch_guild())
```
ChrisLovering, line 40: This pattern of making a task and awaiting it in functions isn't needed anymore, since d.py added async cog_load loading. As it guarantees that the cog_load function is finished before the cog accepts any command/events.

So this task, an all usages of it can be safely deleted.

ChrisLovering (changes requested):

bot/exts/moderation/modpings.py
```diff
@@ -122,16 +128,27 @@ async def add_role_schedule(self, mod: Member, work_time: int, schedule_start: d
...
+        if self.guild.get_role(Roles.mod_team) not in mod.roles:
+            log.info(
+                f"Preventing {mod} from being assigned the moderator role"
+                "since they are missing the mod team role."
+            )
+            # The code that calls this method either prechecks the moderator's roles,
+            # or it doesn't matter since this was called in a task.
+            # The logging statement above is enough to ensure that this was caught
+            # as with a logging level of WARNING, it will create a sentry issue.
+            return
```
wookie184, line 149: Should we delete them from `pings_off_mods` before returning? Does it matter?

wookie184 (changes requested):

wookie184:
Is this still necessary? I think the issue was fixed in https://github.com/python-discord/bot/commit/2b2dc95f2aa2ac6606b51be183861f2663be3b3e

TizzySaurus:
> Is this still necessary? I think the issue was fixed in [2b2dc95](https://github.com/python-discord/bot/commit/2b2dc95f2aa2ac6606b51be183861f2663be3b3e)

Without actually testing anything, it does indeed look like this issue has been fixed (which would mean this can be closed).

wookie184:
Since it's a bit more robust to have the check  in the same place that the role is applied I think we could do this anyway.

I'll take this over and finish it off.

PR #2013, opened 2021-12-14
Add text indicating when user fetched by message

### LINKED ISSUE #2012: Add context to user command when run against messageID
Our `!user` command currently supports a messageID arg, and returns the user embed for that message's author. This can lead to some confusion when the user running the command doesn't know about this feature, and then tries to use that ID with another command that doesn't support it.

The proposed solution for this is to highlight when the user command is using a messageID, as this goes against expectations, in the embed title.

EG for this embed, the title would be **Chrisjl#2655 (From Message ID)**.
image

TizzySaurus:
Closes #2012.

`!user` now highlights in the embed title when the user command is using a message to retrieve the user information.

Also removed a redundant assignment.

image


codephile1221 (approved):
I did not test it but looks good to me.

TizzySaurus:
Have just realised that the tests on this failed. I'll update the tests (hopefully) tonight.

TizzySaurus:
I realised tonight that I never got round to fixing the tests so have just done so with the help of @HassanAbouelela and @onerandomusername :+1:

This PR now just needs a core-dev approval and it can be merged.

onerandomusername (approved):
lgtm

bot/exts/info/information.py
```diff
@@ -256,6 +256,9 @@ async def create_user_embed(self, ctx: Context, user: MemberOrUser) -> Embed:
             name = f"{user.nick} ({name})"
         name = escape_markdown(name)
 
+        if passed_as_message:
+            name += " - From Message"
```
Xithrius, line 260:
```suggestion
            name += " (From Message ID)"
```

TizzySaurus: I decided to use a hyphen instead of brackets since it looked weird when the user has a nickname

Xithrius: Ah, that makes sense.

Xithrius (approved):

PR #2014, opened 2021-12-15
chore: ensure the docs command will look good on later versions of markdownify

onerandomusername:
…rkdownify

as of markdownify 0.7.1 it now converts hr tags natively
however, what it converts to is not supported by discord markdown
to get around this, we just return an empty string, as per current behavior

Current behavior:
image

Newer version of markdownify
image

With the patch:
image

Offending diff:
https://github.com/matthewwithanm/python-markdownify/compare/0.7.0...0.7.1


Numerlor:
Why is this necessary? The version of markdownify is pinned to exactly 0.6.1

onerandomusername:
I-

I think it would have been worth noticing that.

Kronifer (approved):
Looks good!

PR #2016, opened 2021-12-16
Include message counts in all channels

ChrisLovering:
Thanks to a recent database maintenance (https://pythondiscord.freshstatus.io/incident/139811) getting a member's message count is far cheaper. So there is no longer a reason to reserve it for mod channels.

This also reverts this commit https://github.com/python-discord/bot/commit/1872bb12766bf3c56284ef614b5ab22166b488e5 as the site will always return activity blocks now (as of https://github.com/python-discord/site/pull/636).


onerandomusername (approved):
lgtm

ChrisLovering:
Site PR has been merged, this can now be merged safely

bot/exts/info/information.py
```diff
@@ -420,13 +420,8 @@ async def user_messages(self, user: MemberOrUser) -> Tuple[Union[bool, str], Tup
             if e.status == 404:
                 activity_output = "No activity"
```
Bluenix2, line 421: Uh oh, because of the code below it won't this try to iterate the string?
```python
            activity_output = "\n".join(
                f"{name}: {metric}" for name, metric in zip(["Messages", "Activity blocks"], activity_output)
            )
```

Shouldn't it be `activity_output = ["No activity"]`? Have we never hit this codepath?

ChrisLovering: ??? that .join is within the else block, so will never get hit if this line is.

Akarys42 (approved):

Xithrius:
As far as I can remember, one of the reasons that we didn't want to include message counts in the `!user` command was to not "gamify" the number.

I feel like there's a chance this could increase spam, but also help those with good and bad intentions how close they are to getting voice verified.

Xithrius (approved):

PR #2019, opened 2021-12-17
Better off topic tags

### LINKED ISSUE #2018: Re-order off-topic channels in !ot embed
### Amendment type

Content Update

### Content type

Tag

### Please provide the location of the content

!ot

### What amendment do you think should be made to this content, and why?

`!ot` is very often used to re-direct users to off-topic channels when their discussions are out of scope for whatever channel they are in. Currently the channels in the embed are listed in numerical order: 0, 1, 2. 

Users then head over to `#ot0` and continue their conversation, often while there are existing ongoing conversations already in the channel. `#ot0` already gets a lot of traffic compared to `#ot1` and `#ot2`. The multiple overlapping conversations would be mitigated if members used the other 2 channels more.

My proposal was to have the below embed prioritize the off-topic channel with the oldest activity first, and most recent activity last. Other staff have suggested we may just reverse the current order to: 2, 1, 0.

Personally when directing others to off-topic I pick and mention whichever one has no recent activity, instead of using `!ot`

Screenshot of current embed:
image

minalike:
Closes #2018 

As [discussed](https://discord.com/channels/267624335836053506/551789653284356126/921435381331791923) in staff channels, we are modifying the original `!ot` embed to only list #ot2, since the tag's primary use is to redirect users to an off-topic channel. By only listing #ot2 the hope is that there will be less interruptions to the very traffic-heavy #ot0 channel.

image

A new tag called `!off-topic-names` is introduced to serve the content of the original tag. Channels are listed in the original order, with a brief explanation of the nightly name changes, and a link to the off-topic etiquette page.

image


minalike:
to quote kat:

> I think it's shown a beneficial change, even from the perspective of people new to our server and off topics. Rather than posting in a busy channel and getting quickly drowned out, their message is now given more space and isn't likely to interrupt and ongoing convo.
Looking at percentages of where users will post first messages in off-topic:
ot0: 76% -> 71%
ot1: 13.5% -> 14.22%
ot2: 10.22% -> 14.67%
ot2 even had more messages posted it than ot1 for this latest stat pull. That indicates to me that the embed is successfully redirecting folks. So why not just clarify the embed further, only link ot2, and take our improvements for a marginal change.

bot/resources/tags/off-topic-names.md
```diff
@@ -0,0 +1,10 @@
+**Off-topic channels**
+
+There are three off-topic channels:
+• <#291284109232308226>
+• <#463035241142026251>
+• <#463035268514185226>
+
+The channel names change every night at midnight UTC and are fun meta references to jokes or conversations that happened on the server.
```
dawnofmidnight, line 8:
```suggestion
The channel names change every night at midnight UTC and are often fun meta references to jokes or conversations that happened on the server.
```
Just a single suggestion, I don't really have an opinion on the tags themselves.

janine9vn (approved):
Looks good! I would suggest committing Dawn's suggestion because I like the wording of "often", but it's not a blocker so I'm going to go ahead an approve this.

Otherwise, as a note, this does mean that `!off-topic` will fuzzy-match to the longer embed and not `!ot`, which may *at first* seem like inconsistent behavior. But actually, no one really uses the full `!off-topic` tag invocation. The usage for the past month is almost entirely `!ot`, so I'm fully onboard with changing the tag name to `!ot`.

I would like to add an issue for adding an `alt-names` key to tag metadata, but that's not required for this PR.

PR #2020, opened 2021-12-18
feat: `!remind` end time

### LINKED ISSUE #1924: Allow passing end time to `!remind`
This issue proposes allowing users to specify the end time of a reminder as the duration, instead of a relative time (e.g. `!remind "01-11-2021 08:00" example` instead of `!remind 16h38M example`).

This would mean the user doesn't have to do the maths of "how far away is the time I want", but would also mean we'd have to come up with some sort of parsing for the time, potentially with support for `!remind 08:00 example` (send reminder whenever the next occurrence of 8am is) and/or `!remind "01-11-2021 08:00UTC+1" example` (specifying timezone instead of defaulting to UTC).

onerandomusername:
closes #1924

cherry-picked this commit from #2001  https://github.com/python-discord/bot/pull/2001/commits/f8ea3e884128883381d8b4d37ff94e62f0400e36

This is still a draft, existing to be able to share my progress.

### DIFF
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -10,11 +10,12 @@ import dateutil.tz
 import discord
 from aiohttp import ClientConnectorError
 from dateutil.relativedelta import relativedelta
-from discord.ext.commands import BadArgument, Bot, Context, Converter, IDConverter, MemberConverter, UserConverter
+from discord.ext.commands import BadArgument, Context, Converter, IDConverter, MemberConverter, UserConverter
 from discord.utils import escape_markdown, snowflake_time
 
 from bot import exts
 from bot.api import ResponseCodeError
+from bot.bot import Bot
 from bot.constants import URLs
 from bot.errors import InvalidInfraction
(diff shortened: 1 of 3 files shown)


TizzySaurus:
I've been looking into this and we need to wait for [bot#2001](https://github.com/python-discord/bot/pull/2001) to be merged, as a part of this will use the `DayDuration` converter that was implemented. 

Due to this, I'm going to mark this PR as stalled until that's merged.

onerandomusername:
This is more stalled as the reminder command is currently disabled due to bugs.

MarkKoz:
@onerandomusername No longer disabled since #2080. Do you plan to continue working on this?

wookie184:
I'm going to close this due to lack of activity and since it still seems quite far from being done. More discussion is needed on what formats should be accepted before continuing with this idea (this currently only accepts hh:mm or just hh which is quite different to what the example in the issue proposed).

PR #2021, opened 2021-12-18
More Message Link Unfurls 

### LINKED ISSUE #1964: Unfurl Discord links in more places
In #1446 we made the bot unfurl and message links found in reported incidents. This sends a webhook with details on the message referenced.

This is *incredibly* useful, especially for people on Android, where Discord links are completely broken.

I would like to request that this feature is extended to at least *all* staff channels, with the option to whitelist additional channels/categories if needed.

As a brief spec, the bot should look for Discord links in all messages in staff channels, for each link found the bot should send an embed similar to below.

image

See the linked PR for reference on how this was done. We should likely make this logic a shared util.

Shivansh-007:
Closes #1964 

Screenshot from 2021-12-18 17-18-06

### DIFF
--- a/bot/exts/utils/unfurl_message_links.py
+++ b/bot/exts/utils/unfurl_message_links.py
@@ -2,7 +2,7 @@ import discord
 from discord.ext.commands import Cog
 
 from bot.bot import Bot
-from bot.constants import Guild
+from bot.constants import Channels, Guild
 from bot.utils.channel import is_staff_channel
 from bot.utils.messages import extract_message_links, make_message_link_embed
 
@@ -19,6 +19,7 @@ class UnfurlMsgLinks(Cog):
         if (
             not message.guild
             or message.guild.id != Guild.id
+            or message.channel.id != Channels.incidents
             or not is_staff_channel(message.channel)
         ):
             return
--- a/bot/exts/utils/unfurl_message_links.py
+++ b/bot/exts/utils/unfurl_message_links.py
@@ -19,7 +19,7 @@ class UnfurlMsgLinks(Cog):
         if (
             not message.guild
             or message.guild.id != Guild.id
-            or message.channel.id != Channels.incidents
+            or message.channel.id == Channels.incidents
             or not is_staff_channel(message.channel)
         ):
             return
--- a/bot/exts/utils/unfurl_message_links.py
+++ b/bot/exts/utils/unfurl_message_links.py
@@ -1,0 +1,34 @@
+import discord
+from discord.ext.commands import Cog
+
+from bot.bot import Bot
+from bot.constants import Guild
+from bot.utils.channel import is_staff_channel
+from bot.utils.messages import extract_message_links, make_message_link_embed
+
+
+class UnfurlMsgLinks(Cog):
+    """Unfurl message links and send embeds containing information about the linked message."""
+
+    def __init__(self, bot: Bot) -> None:
+        self.bot = bot
+
+    @Cog.listener()
+    async def on_message(self, message: discord.Message) -> None:
+        """Unfurl messages linked in the message if any and send the message link embed."""
+        if (
+            not message.guild
+            or message.guild.id != Guild.id
+            or not is_staff_channel(message.channel)
+        ):
+            return
+
+        if message_links := await extract_message_links(message):
+            ctx = await self.bot.get_context(message)
+            embeds = [await make_message_link_embed(ctx, msg) for msg in message_links]
+            await message.reply(embeds=[embed for embed in embeds if embed])
+
+
+def setup(bot: Bot) -> None:
+    """Load the UnfurlMsgLinks cog."""
+    bot.add_cog(UnfurlMsgLinks(bot))
--- a/bot/utils/messages.py
+++ b/bot/utils/messages.py
@@ -6,12 +6,13 @@ from io import BytesIO
 from typing import Callable, List, Optional, Sequence, Union
 
 import discord
-from discord.ext.commands import Context
+from discord.ext.commands import Context, MessageConverter, MessageNotFound
 
 import bot
-from bot.constants import Emojis, MODERATION_ROLES, NEGATIVE_REPLIES
+from bot.constants import Channels, Emojis, MODERATION_ROLES, NEGATIVE_REPLIES, Roles
 from bot.log import get_logger
 from bot.utils import scheduling
+from bot.utils.regex import DISCORD_MESSAGE_LINK_RE
 
 log = get_logger(__name__)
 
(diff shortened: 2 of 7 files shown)


bot/exts/utils/unfurl_message_links.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self, bot: Bot) -> None:
+        self.bot = bot
+
+    @Cog.listener()
+    async def on_message(self, message: discord.Message) -> None:
+        """Unfurl messages linked in the message if any and send the message link embed."""
+        if (
+            not message.guild
+            or message.guild.id != Guild.id
+            or message.channel.id != Channels.incidents
```
onerandomusername, line 22: this should be `==`

onerandomusername (changes requested):
looks pretty good, tested and found a small bug.

bot/exts/utils/unfurl_message_links.py
onerandomusername: Shivansh gave me permissions to push, so its fixed in https://github.com/python-discord/bot/pull/2021/commits/37098dfa7138bbfa7734983496bf7810692d1d1b

onerandomusername (approved):
lgtm

bot/utils/messages.py
```diff
@@ -238,3 +239,102 @@ async def send_denial(ctx: Context, reason: str) -> discord.Message:
...
+    """Create an embedded representation of the discord message link."""
+    embed = None
+
+    try:
+        message: discord.Message = await MessageConverter().convert(ctx, message_link)
+    except MessageNotFound:
+        mod_logs_channel = ctx.bot.get_channel(Channels.mod_log)
+        last_100_logs: list[discord.Message] = await mod_logs_channel.history(limit=100).flatten()
+
+        for log_entry in last_100_logs:
```
Bluenix2, line 287:
```suggestion
        async for log_entry in mod_logs_channel.history(limit=100):
```
There's no need to flatten it and then iterate through it.

tests/bot/utils/test_messages.py
```diff
@@ -25,3 +27,86 @@ def test_sub_clyde(self):
...
+            "http://ptb.discord.com/channels/884793126525489203/884793128073195592/921717920869003294",
+            "http://canary.discordapp.com/channels/884793126525489203/884793128073195592/921717920869003294",
+        )
+
+        for link in test_links:
+            with self.subTest(token=link):
+                results = regex.DISCORD_MESSAGE_LINK_RE.fullmatch(link)
+                self.assertIsNotNone(
+                    results, f"Valid link '{link}' was not matched by the regex"
+                )
```
Bluenix2, line 66: Shouldn't `results[0] == link`?

tests/bot/utils/test_messages.py
```diff
@@ -25,3 +27,86 @@ def test_sub_clyde(self):
...
+            "https://discord.com/channels/884793126525489203/884793128073195592000/921717920869003294",
+            "http:/discord.com/channels/884793126525489203/884793128073195592/92171792086900329400",
+            "http://ptb.discord.com/channels/884793126525489203/884793128073195592000/921717920869003294",
+            "http://canary.discord.app/channels/884793126525489203/884793128073195592/921717920869003294",
+        )
+
+        for link in test_links:
+            with self.subTest(token=link):
+                results = regex.DISCORD_MESSAGE_LINK_RE.findall(link)
+                print(results)
```
Bluenix2, line 80: You may have forgotten to remove this print.

bot/exts/utils/unfurl_message_links.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @Cog.listener()
+    async def on_message(self, message: discord.Message) -> None:
+        """Unfurl messages linked in the message if any and send the message link embed."""
+        if (
+            not message.guild
+            or message.guild.id != Guild.id
+            or message.channel.id == Channels.incidents
+            or not is_staff_channel(message.channel)
+        ):
```
bast0006, line 24: Add a comment here describing the intended behavior?

Unfurl links only if the message containing the link:
- isn't in dms
- is not within the python discord
- is not within the incidents channel (handled in the incidents cog)
- and the message is in a staff channel


Also missing from here is the ability to configure other channels (which isn't required but would be amazing if we could get a channel whitelist and add staff channels to it rather than need to edit the code each time)

bot/exts/utils/unfurl_message_links.py
```diff
@@ -0,0 +1,35 @@
...
+            return
+
+        if message_links := await extract_message_links(message):
+            ctx = await self.bot.get_context(message)
+            embeds = [await make_message_link_embed(ctx, msg) for msg in message_links]
+            await message.reply(embeds=[embed for embed in embeds if embed])
```
bast0006, line 30: Would this be cleaner as a regular for loop?

bot/utils/messages.py
```diff
@@ -238,3 +239,102 @@ async def send_denial(ctx: Context, reason: str) -> discord.Message:
...
+    # Truncate text to a maximum of 300 characters
+    if len(text) > 300:
+        text = text[:300]
+
+    # Limit to a maximum of three lines
+    text = "\n".join(text.split("\n", maxsplit=3)[:3])
+
+    # If it is a single word, then truncate it to 50 characters
+    if text.find(" ") == -1:
+        text = text[:50]
```
bast0006, line 260: " " not in text

bot/utils/messages.py
```diff
@@ -238,3 +239,102 @@ async def send_denial(ctx: Context, reason: str) -> discord.Message:
...
+
+    # Remove extra whitespaces from the `text`
+    text = text.strip()
```
bast0006, line 263: This step should come first

bot/utils/messages.py
bast0006: This also fails to account for nonstandard or newline delimited content. Example:

123456789012345678901234567890
123456789012345678901234567890
123456789012345678901234567890

is 60 characters, but this filter will truncate it to 50. I don't think we should do anything fancy here--it's very unlikely someone will link a message with nonbreaking spaces in it--but I think we should account for the newline case.

bot/utils/messages.py
```diff
@@ -238,3 +239,102 @@ async def send_denial(ctx: Context, reason: str) -> discord.Message:
...
+    # Add placeholder if the text was shortened
+    if len(text) < original_length:
+        text = f"{text}..."
+
+    return text
+
+
+async def extract_message_links(message: discord.Message) -> Optional[list[str]]:
+    """This method extracts discord message links from a message object."""
+    return [msg_link[0] for msg_link in DISCORD_MESSAGE_LINK_RE.findall(message.content)]
```
bast0006, line 274: [0] here doesn't really make it clear what we're grabbing. We can remove this by fixing the message link re. See that other comment.

bot/utils/regex.py
```diff
@@ -12,4 +12,11 @@
     r"(?P<invite>[a-zA-Z0-9\-]+)",                # the invite code itself
     flags=re.IGNORECASE
 )
+
 MESSAGE_ID_RE = re.compile(r'(?P<message_id>[0-9]{15,20})$')
+
+DISCORD_MESSAGE_LINK_RE = re.compile(
+    r"(https?:\/\/(?:(ptb|canary|www)\.)?discord(?:app)?\.com\/channels\/"
```
bast0006, line 19: There's a missing ?: before the `ptb`:

(https?:\/\/(?:( __?:__ ptb|canary|www)\.)?dis

That turns this into a two-group capture regex. As we don't have a use (afaict) here for that second half, we should add the ?: to simplify the interface. This will change the return result of .findall() to a simple list of entries.

tests/bot/utils/test_messages.py
```diff
@@ -25,3 +27,86 @@ def test_sub_clyde(self):
...
+                )
+
+    def test_extract_invalid_message_links(self):
+        """Test message link detection regex."""
+        test_links = (
+            "https://discord.com/channels/884793126525489203/884793128073195592000/921717920869003294",
+            "http:/discord.com/channels/884793126525489203/884793128073195592/92171792086900329400",
+            "http://ptb.discord.com/channels/884793126525489203/884793128073195592000/921717920869003294",
+            "http://canary.discord.app/channels/884793126525489203/884793128073195592/921717920869003294",
+        )
```
bast0006, line 75: Currently the DISCORD_MESSAGE_LINK_RE does not match links to direct messages (for example: `https://discord.com/channels/@me/759105648322936863/911741334019264583`). We should clarify this/comment it at least here or at the regex itself. For the link unfurl feature, it's good that it doesn't match them.

bast0006 (changes requested):

Xithrius:
Per [this message](https://canary.discord.com/channels/267624335836053506/635950537262759947/940807906264219678) from the author, this PR is now up for grabs.

Xithrius:
@bast0006 perhaps you could take over this PR?

mbaruh:
This PR is not ready, but the source branch no longer exists. Will need to be re-implemented in a different PR.

PR #2023, opened 2021-12-18
Try an exact match with only the name when a tag with a group is searched

Numerlor:
Currently, a tag is directly returned when it has an exact match even when there are matched fuzzy alternatives, but this does not attempt the exact match on only the name if the tag has a group.  This causes issues if the user invokes a tag with some text after it which is considered as the tag's name.

The PR adds an additional search with only the tag's name without a group when the first lookup of the original identifier fails

### DIFF
--- a/bot/exts/info/tags.py
+++ b/bot/exts/info/tags.py
@@ -275,6 +275,11 @@ class Tags(Cog):
         ]
 
         tag = self.tags.get(tag_identifier)
+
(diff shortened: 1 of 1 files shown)


janine9vn (approved):
Works like a charm!

I tested this with a couple of different combos of tags + text after including with a tag group (dpy tag group) and everything works as expected.

jb3 (approved):

PR #2031, opened 2021-12-29
Incident archive msg improvements

### LINKED ISSUE #1059: Incidents: show report datetime in archive embed
As it is, the archive embeds in `#incidents-archive` show the datetime the report was actioned in both the message send time `A` the archive embed footer `B`.

archives

This means that we're archiving duplicate information and failing to archive the datetime that the incident was reported.

The proposal is to adjust the embed to show `incident.created_at` instead of `utcnow()`. This would be a trivial fix here:

https://github.com/python-discord/bot/blob/65e2c2946ddd4d2dcc6b00ec68c0b456048840db/bot/cogs/moderation/incidents.py#L91-L95

**Pros**

* We would be archiving a possibly useful piece of information, which can be used to calculate interesting stats such as response time w.r.t. time of day or whether the report contains a `@Moderators` pingu.

**Cons**

* I think the embed would be somewhat confusing. It'd be a lot more logical to have `A` and `B` switched around, so that the mod's identity is shown next to the time the report was actioned, and the helper's next to the report time. That sadly isn't possible though. Neither is it possible to have a custom footer format, e.g. `Actioned by X, Reported on: Y` as we're using the timestamp field, which will convert the utc timestamp to the client's local timezone (as it does with the message sent time, so they are in sync).
* If multiple incidents reported by the same user are actioned in succession, the Discord GUI won't show the send datetime for each, but will show the embeds as if they were sent together. This means that we will only see when each incident was reported, but not when actioned. Both pieces of information can still be easily retrieved when accessing the `Message` object programmatically, e.g. via int e though.

**Conclusion**
As shown above the technical side of things is trivial but we should discuss whether this is a change that we'd want. It'd make a lot of sense to archive both datetimes, but using the embed footer doesn't seem very elegant to me for the reason outlined above.

We originally discussed this with @TizzySaurus.

TizzySaurus:
Closes #1059 

Incident archive messages now show when the incident was made and no longer displays when the incident was actioned as it's the same as the message timestamp.

Also removed some redundant escapes in the discord message link regex.


kosayoda (commented):
I like the idea of the proposed UI, but I think it can be improved. Thoughts?

Suggestion:
image

- [x] Simplified time of report (less clutter, straight to the point)
- [x] Relative action time in footer (per @MarkKoz [here](https://github.com/python-discord/bot/issues/1059#issuecomment-670072938))
- [x] Absolute action time as footer timestamp

<details>
<summary>crude suggestion diff</summary>

```diff
diff --git a/bot/exts/moderation/incidents.py b/bot/exts/moderation/incidents.py
index bd9e5b88..f1ad6a59 100644
--- a/bot/exts/moderation/incidents.py
+++ b/bot/exts/moderation/incidents.py
@@ -1,18 +1,20 @@
 import asyncio
 import re
+from datetime import datetime, timezone
 from enum import Enum
 from typing import Optional
 
 import discord
 from async_rediscache import RedisCache
 from discord.ext.commands import Cog, Context, MessageConverter, MessageNotFound
+from dateutil.relativedelta import relativedelta
 
 from bot.bot import Bot
 from bot.constants import Channels, Colours, Emojis, Guild, Roles, Webhooks
 from bot.log import get_logger
 from bot.utils import scheduling
 from bot.utils.messages import format_user, sub_clyde
-from bot.utils.time import TimestampFormats, discord_timestamp
+from bot.utils.time import TimestampFormats, discord_timestamp, humanize_delta
 
 log = get_logger(__name__)
 
@@ -90,17 +92,19 @@ async def make_embed(incident: discord.Message, outcome: Signal, actioned_by: di
     """
     log.trace(f"Creating embed for {incident.id=}")
 
+    relative_action_time = humanize_delta(
+        relativedelta(datetime.now(timezone.utc), incident.created_at), max_units=1
+    )
     if outcome is Signal.ACTIONED:
         colour = Colours.soft_green
-        footer = f"Actioned by {actioned_by}"
+        footer = f"Actioned by {actioned_by} after {relative_action_time}"
     else:
         colour = Colours.soft_red
-        footer = f"Rejected by {actioned_by}"
+        footer = f"Rejected by {actioned_by} after {relative_action_time}"
 
-    day_timestamp = discord_timestamp(incident.created_at, TimestampFormats.DATE)
-    time_timestamp = discord_timestamp(incident.created_at, TimestampFormats.TIME)
+    datetime_timestamp = discord_timestamp(incident.created_at, TimestampFormats.DATE_TIME)
     relative_timestamp = discord_timestamp(incident.created_at, TimestampFormats.RELATIVE)
-    reported_on_msg = f"__*Reported {day_timestamp} at {time_timestamp} ({relative_timestamp}).*__"
+    reported_on_msg = f"Reported {datetime_timestamp} ({relative_timestamp})."
 
     # If the description will be too long (>4096 total characters), truncate the incident content
     if len(incident.content) > (allowed_content_chars := 4096-len(reported_on_msg)-2):  # -2 for the newlines
@@ -110,6 +114,7 @@ async def make_embed(incident: discord.Message, outcome: Signal, actioned_by: di
 
     embed = discord.Embed(
         description=description,
+        timestamp=datetime.utcnow(),
         colour=colour,
     )
     embed.set_footer(text=footer, icon_url=actioned_by.display_avatar.url)
```

</details>
cc issue creator @kwzrd

TizzySaurus:
>  Relative action time in footer (per @MarkKoz here)

I discussed this with one of the core devs (I believe in a mod channel) and we decided we don't want this statistic, as it's often not particularly accurate - people forgetting to action etc.

> Absolute action time as footer timestamp

For this I refer back to this PR description: "no longer displays when the incident was actioned as it's the same as the message timestamp." This isn't needed.


>  Simplified time of report (less clutter, straight to the point)
I do agree with this point; it definitely looks much nicer.

kosayoda:
> we decided we don't want this statistic

Sure, makes sense.

> when multiple reports by the same user are actioned.

In compact mode, if I want to see the date an incident was actioned, I would have to scroll to the first incident of that day, so adding the timestamp to the footer helps. Although, in practice the action time shouldn't be too distant from the report time, so the information is technically in the embed already.

TizzySaurus:
> Although, in practice the action time shouldn't be too distant from the report time, so the information is technically in the embed already.

I mean there are multiple instances where it's several hours after, simply because we're unfortunately not a 24/7 taskforce. Then there's the odd case of it being days later (due to being delayed by a ModMail thread etc.).

I think supported with the beginning of your message, that's enough rationale for re-adding the timestamp.

bot/exts/moderation/incidents.py
```diff
@@ -97,9 +97,19 @@ async def make_embed(incident: discord.Message, outcome: Signal, actioned_by: di
         colour = Colours.soft_red
         footer = f"Rejected by {actioned_by}"
 
+    day_timestamp = discord_timestamp(incident.created_at, TimestampFormats.DATE)
+    time_timestamp = discord_timestamp(incident.created_at, TimestampFormats.TIME)
```
Bluenix2, line 101: Is there any reasoning behind using this over the option that shows both?

TizzySaurus: I believe I did it to be able to customise the format (add extra words between date and time) but in hindsight it's rather unnecessary, so have change to use the DATETIME option.

TizzySaurus:
This latest commit is untested as I'm on holiday, so have added the "do not merge" tag to make sure it doesn't get merged before confirming it all works as planned.

TizzySaurus:
Have tested and everything works as expected:

image

As such, this is now ready for review :smile:

bot/exts/moderation/incidents.py
```diff
@@ -97,10 +98,20 @@ async def make_embed(incident: discord.Message, outcome: Signal, actioned_by: di
...
 
+    reported_timestamp = discord_timestamp(incident.created_at)
+    relative_timestamp = discord_timestamp(incident.created_at, TimestampFormats.RELATIVE)
+    reported_on_msg = f"*Reported {reported_timestamp} ({relative_timestamp}).*"
+
+    # If the description will be too long (>4096 total characters), truncate the incident content
+    if len(incident.content) > (allowed_content_chars := 4096-len(reported_on_msg)-2):  # -2 for the newlines
+        description = incident.content[:allowed_content_chars-3] + f"...\n\n{reported_on_msg}"
+    else:
+        description = incident.content + f"\n\n{reported_on_msg}"
```
wookie184, line 109: 4096 characters is quite a lot, do you think we should limit it to something smaller, maybe 2000 or even just 1000 characters, so it's less bulky?

wookie184 (approved):
LGTM. Seems like there are some errors in the tests that need fixing though

bot/exts/moderation/incidents.py
TizzySaurus: I went for 4096 simply because that's the Discord maximum. If we'd prefer to lower it then I'm happy to.

TizzySaurus:
> LGTM. Seems like there are some errors in the tests that need fixing though

Turns out I forgot to push the commit where I fixed the tests 😅

shenanigansd (approved):
LGTM

bot/exts/moderation/incidents.py
wookie184: I guess since it was already like that and there wasn't an issue, it's probably fine. We can always change it later if we need to

wookie184 (approved):

PR #2036, opened 2022-01-01
Infraction mod-log improvements

### LINKED ISSUE #2034: Add infraction ID to the #mod-log entry when the infraction is edited
> Member: @name [ID]
> Actor: @Zig
> Edited by: @Zig
> Previous expiry: in [duration]
> New expiry: in [duration]
> Today at 12:00 PM

Current entries look like this, which can lead to confusion when moderators search the log by infraction ID.

TizzySaurus:
Closes #2034.

Improvements:
- Updated the text in a `log.trace` to be less misleading
- Infraction edit mod-logs now display the infraction id
- Infraction applied mod-logs now have the previously missing colon (`:`)
- Edited some code to use the defined infraction id variables, rather than re-indexing the dictionary.

The below image shows the mod-logs for applying a mute, editing the mute reason & duration, and finally removing the mute:

### DIFF
--- a/bot/exts/moderation/infraction/_scheduler.py
+++ b/bot/exts/moderation/infraction/_scheduler.py
@@ -223,7 +223,7 @@ class InfractionScheduler:
                 failed = True
 
         if failed:
-            log.trace(f"Deleted infraction {infraction['id']} from database because applying infraction failed.")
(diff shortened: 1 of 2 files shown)


kosayoda (approved):
LGTM, works as expected.

ChrisLovering (approved):

PR #2039, opened 2022-01-05
Disallow Editing Infraction Durations Into The Past

### LINKED ISSUE #2032: Don't allow editing an active infraction's expiration into the past
`!infraction edit` can take an ISO datetime to edit the expiration time, but there's no guard to prevent it from specifying a past date, immediately pardoning the infraction as a result.

The same goes for `!infraction append`.

ShakyaMajumdar:
Closes #2032


bot/exts/moderation/infraction/management.py
```diff
@@ -150,6 +150,10 @@ async def infraction_edit(
             request_data['expires_at'] = None
             confirm_messages.append("marked as permanent")
         elif duration is not None:
+            now_datetime = datetime.now(duration.tzinfo) if duration.tzinfo else datetime.utcnow()
```
Akarys42, line 153: Shouldn't the duration timezone always be UTC?

mbaruh, line 153: Notice that `datetime.utcnow()` is actually timezone unaware (yeah, I know), which could become a problem since you wouldn't be able to compare the timezone-unaware `now_datetime` to the timezone-aware `duration`. In practice this will never happen since `duration` should always have timezone information.

Furthermore, even if `duration`'s timezone could be something other than UTC, there's no need to take it into account when creating `now_datetime`, since comparing two datetimes in different timezones will cause the library to make the appropriate conversions.

So all in all it's fine to always have `now_datetime = datetime.now(timezone.utc)` when `timezone` is imported from `datetime`. Or just use `arrow.utcnow()` which is timezone aware.

mbaruh (changes requested):
Forgot to make the above a requested change

bot/exts/moderation/infraction/management.py
ShakyaMajumdar: I'm not sure. I copied that line from the [scheduler](https://github.com/python-discord/bot/blob/main/bot/utils/scheduling.py#L73).

VirdanTheBurden: The duration timezone is always UTC. The `Expiry` typehint defines `ISODateTime` and `Duration`, which both convert their respective inputs into datetime objects within the UTC+0 timezone.

[ISODateTime](https://github.com/python-discord/bot/blob/3824ddbddff8c3191632c0479f0f594985a55b32/bot/converters.py#L425-L432)
[Duration](https://github.com/python-discord/bot/blob/3824ddbddff8c3191632c0479f0f594985a55b32/bot/converters.py#L348)
[Expiry](https://github.com/python-discord/bot/blob/3824ddbddff8c3191632c0479f0f594985a55b32/bot/converters.py#L596)

ShakyaMajumdar: Then ` now_datetime = datetime.utcnow()` should be enough, if I follow correctly?

VirdanTheBurden: It should be enough. An alternative would be `now_datetime = datetime.now(timezone.utc)`, as the `datetime` module says that [`datetime.now(timezone.utc)` is preferred over `datetime.utcnow()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) when it comes to creating `datetime` objects that represent UTC times.

ShakyaMajumdar:
Hmm, looks like its not just for edits, setting a past time as the initial infraction duration seems to be allowed too
image
image

VirdanTheBurden (approved):
There's nothing wrong with both `_scheduler.py` and `management.py` from what I see.

Xithrius:
@ShakyaMajumdar Greetings. Please resolve these the file conflicts. Thanks!

MaskDuck (approved):
gud

minalike (commented):
I can't comment much on the code, but I can say it does work as expected in testing. Maybe @VirdanTheBurden can take another look since new code has been pushed since your last review.

bot/decorators.py
```diff
@@ -236,3 +238,31 @@ async def wrapped(*args, **kwargs) -> t.Any:
             return await func(*args, **kwargs)
         return wrapped
     return decorator
+
+
+def ensure_duration_in_future(duration_arg: function.Argument) -> t.Callable:
```
MarkKoz, line 243: How about `ensure_future_duration` as the name? It's slightly shorter.

I also humbly suggest to make this more generic (currently the name and error message tie it too close to infractions). The name could then be `ensure_future_timestamp`.

bot/decorators.py
```diff
@@ -236,3 +238,31 @@ async def wrapped(*args, **kwargs) -> t.Any:
...
+    """
+    Ensure the duration argument is in the future.
+
+    If the condition fails, a warning is sent to the invoking context.
```
MarkKoz, line 247: Imperative mood rather than passive voice is more consistent.
```suggestion
    If the condition fails, send a warning to the invoking context.
```

bot/decorators.py
```diff
@@ -236,3 +238,31 @@ async def wrapped(*args, **kwargs) -> t.Any:
...
+    """
+    def decorator(func: types.FunctionType) -> types.FunctionType:
+        @command_wraps(func)
+        async def wrapper(*args, **kwargs) -> t.Any:
+            bound_args = function.get_bound_args(func, args, kwargs)
+            target = function.get_arg_value(duration_arg, bound_args)
+
+            ctx = function.get_arg_value(1, bound_args)
+
+            if isinstance(target, datetime) and target < arrow.utcnow():
```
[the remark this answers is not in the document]
MarkKoz: I think it's because the duration converter returns a `datetime` and `Arrow` does not subclass it.

I'm not sure why the type is being checked though, especially since it's silently ignored. The converter always returns a `datetime`, and this decorator is only intended to be used with that type. Thus, if the type doesn't match then it's a programmer error that should not just be ignored.

One could argue that the type check can still be used to raise a clearer error message but I argue that it would subvert duck typing. That could be especially pertinent in this case since it's conceivable that this decorator will be used with an `Arrow` parameter in the future. As I mentioned earlier, it is not a subclass of `datetime` despite having an identical interface for this purpose.

If the duration can be None then explicitly check for None instead of doing it this way.

MarkKoz, line 262: Subtle thing but using less than instead of less than equal is misleading when it's supposed to ensure it's in the _future_.

MarkKoz (changes requested):

bot/decorators.py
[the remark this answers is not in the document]
ShakyaMajumdar: The failure isn't very silent
image

I suppose I could check whether it's not one of None and str instead of whether it's a datetime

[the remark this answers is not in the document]
MarkKoz: Oh, right cause you call the original function and it picks up the error.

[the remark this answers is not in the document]
ShakyaMajumdar: How about `isinstance(duration, (datetime, Arrow))` for future proofing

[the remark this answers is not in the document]
MarkKoz: > I suppose I could check whether it's not one of None and str instead of whether it's a datetime

How about 
```py
try:
    if (target < arrow.utcnow()):
        ...
except TypeError:
    func(...)
```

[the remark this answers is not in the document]
MarkKoz: Actually this would be a bit better since it's a narrower scope for the caught exception

```py
try:
    is_future = target > arrow.utcnow())
except TypeError:
    func(...)
else:
    if not is_future:
        await ctx.send("!!!!")
```

bot/decorators.py
ShakyaMajumdar: resolved in 574adc3

bot/decorators.py
ShakyaMajumdar: resolved in 574adc3

bot/decorators.py
ShakyaMajumdar: resolved in 61feabb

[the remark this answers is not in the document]
ShakyaMajumdar: resolved in 61feabb

MarkKoz (changes requested):
You forgot superstarify but otherwise it looks good and works according to my tests.

MarkKoz (approved):

PR #2040, opened 2022-01-07
Mentioning the user when a nomination 'fails'

Shom770:
Currently, when someone has already nominated the user, it simply says that the nomination already exists, but doesn't mention the user.
The same goes for when someone edits the nomination reason for a user, it doesn't mention the user itself just the fact that the nomination reason changed.

It's helpful to know who the user is because most of the time you won't know the user by their ID which most people use for nominating.

Permission to open a PR directly:
https://discord.com/channels/267624335836053506/822920136150745168/928694255734632478


bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -438,7 +438,7 @@ async def _edit_nomination_reason(
             json={"actor": actor.id, "reason": reason}
         )
         await self.refresh_cache()  # Update cache
-        await ctx.send(":white_check_mark: Successfully updated nomination reason.")
+        await ctx.send(f":white_check_mark: Successfully updated nomination reason for {actor.mention}")
```
ChrisLovering, line 441: Isn't `target` the person being nominated here, with `actor` being the person nominating?

ChrisLovering (changes requested):

bot/exts/recruitment/talentpool/_cog.py
Shom770: Agh, skimmed through and I thought actor would be the person being nominated. Fixed it.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -438,7 +440,7 @@ async def _edit_nomination_reason(
             json={"actor": actor.id, "reason": reason}
         )
         await self.refresh_cache()  # Update cache
-        await ctx.send(":white_check_mark: Successfully updated nomination reason.")
+        await ctx.send(f":white_check_mark: Successfully updated nomination reason for {target_mention}")
```
ChrisLovering, line 443:
```suggestion
        await ctx.send(f":white_check_mark: Successfully updated nomination reason for {target.mention}")
```

ChrisLovering (changes requested):

bot/exts/recruitment/talentpool/_cog.py
ChrisLovering: Ah wait, I see what this is doing now, let me think.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -406,9 +406,11 @@ async def _edit_nomination_reason(
             return
         if isinstance(target, int):
             nomination_id = target
+            target_mention = (await self.bot.fetch_user(target)).mention
```
ChrisLovering, line 409: When `target` is an int, it is a nomination_id, not a user_id, so this will never work.
it might be easier to do this using the `nomination` variable, that is the response from `await self.bot.api_client.get(f"bot/nominations/{nomination_id}")`. Since that returns a [nomination object](https://github.com/python-discord/site/blob/main/pydis_site/apps/api/models/bot/nomination.py#L7-L37), which contains a [User object](https://github.com/python-discord/site/blob/ecf794b87a5fa8026e5cfd2f996a4ade143a9042/pydis_site/apps/api/models/bot/user.py#L18) as one of it's keys

ChrisLovering (changes requested):

bot/exts/recruitment/talentpool/_cog.py
Shom770: I see, thanks!

Shom770: I grabbed `nomination['user']` and then ran `await self.bot.fetch_user` given the ID, and it seems to work well now.

One quick question - I know that fetch_user launches an API request so would we want to pull the user from the cache first and if that fails, run an API request on the ID? or is the user object already existing within the nomination object, specifically `nomination['user']`

ChrisLovering: Is there a need to fetch the user at all? `nomination['user']` should have the id in it, so you can just do <@id> manually.

Shom770: Oh wait, I forgot how mentions were formatted in the first place. Thanks!

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -438,7 +438,10 @@ async def _edit_nomination_reason(
             json={"actor": actor.id, "reason": reason}
         )
         await self.refresh_cache()  # Update cache
-        await ctx.send(":white_check_mark: Successfully updated nomination reason.")
+        await ctx.send(
+            f":white_check_mark: Successfully updated nomination reason"
```
codephile1221, line 442: How about extending this behaviour of mentioning the nominee to cases where the nomination is inactive ([line 427](https://github.com/python-discord/bot/blob/762dca11064d2487fa0c7fc3be4284775e19c293/bot/exts/recruitment/talentpool/_cog.py#L427), might need a slight rewording in that case) and when the actor has not nominated the user ([line 431](<https://github.com/python-discord/bot/blob/762dca11064d2487fa0c7fc3be4284775e19c293/bot/exts/recruitment/talentpool/_cog.py#L431>))?

codephile1221 (changes requested):

Xithrius:
@Shom770 Wold you like this PR to go up for grabs if you're not able to work on it anymore?

Shom770:
> @Shom770 Wold you like this PR to go up for grabs if you're not able to work on it anymore?

yeah sure, i've been really busy with school 😅

Xithrius:
Very well. Thanks for the quick response!

HassanAbouelela (approved):

ChrisLovering (approved):

PR #2041, opened 2022-01-08
Restrict allowed mentions for !eval results

bsoyka:
This restricts allowed mentions for `!eval` results (which were previously just the moderation roles) to only allow mentioning the user invoking the command.


HassanAbouelela (approved):
Thanks

ChrisLovering (approved):

ChrisLovering (changes requested):
Actually, changing this until I see what the test fixes are like

HassanAbouelela (approved):

ChrisLovering (approved):

PR #2043, opened 2022-01-09
Add bot-core as a dependency

ChrisLovering:
This PR adds bot-core as a dependency, using a commit hash to pin to a specific version.

While bot core is still quite rudimentary in what it is doing, adding it in early allows us to start adding features and actually using them within the bot.

Currently bot-core only has the regex helpers in it, so this PR moves bot to use the regex available from bot core wherever it can.

This also moves the message ID regex, that's only used once, to inside file that uses it. So that we can delete the regex.py file.

### DIFF
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -9,6 +9,7 @@ import dateutil.parser
 import dateutil.tz
 import discord
 from aiohttp import ClientConnectorError
+from botcore.regex import DISCORD_INVITE
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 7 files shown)


gamingbuddhist (approved):
Everything looks good switching the regex util code location to bot-core and making it a dependency. Did not test functionality but everything looks 1 for 1.

wookie184 (approved):
looks good, and works 👌 

Looking forward to more botcore in the future 😄

PR #2046, opened 2022-01-12
seperate tag dataclasses from tag cog file

### LINKED ISSUE #2022: Incorrect handling of tags by the `!source` command
image

onerandomusername:
### Relevant Issues

closes #2022

### Description

Move the dataclasses that are used by the source cog out of the tag cog file in order to fix the current import issue.

See https://github.com/python-discord/bot/issues/2022#issuecomment-997541569 for a longer explanation.


brad90four (approved):
Looking through the code it looks good to me, haven't tested locally.

Numerlor:
If going with this approach, I think it would make more sense to make the tags extension a package with a separate module for the cog and the non-cog classes instead of moving the classes to utils

onerandomusername:
> If going with this approach, I think it would make more sense to make the tags extension a package with a separate module for the cog and the non-cog classes instead of moving the classes to utils

Resolved as of https://github.com/python-discord/bot/pull/2046/commits/95de0f353012ae17dbfc7eea6fa71c18db93ac07

bot/exts/info/source.py
```diff
@@ -65,7 +65,7 @@ def get_source_link(self, source_item: SourceType) -> Tuple[str, str, Optional[i
 
         # Handle tag file location differently than others to avoid errors in some cases
         if not first_line_no:
-            file_location = Path(filename).relative_to("bot/")
+            file_location = str(filename)
```
Bluenix2, line 68: What was the point of this change?

onerandomusername: See this comment: https://github.com/python-discord/bot/issues/2022#issuecomment-997258029

Bluenix2 (changes requested):
This does not fully fix #2022. If I reload the source cog and then reload the tags cog I can still reproduce the error:
```
AttributeError: 'TagIdentifier' object has no attribute 'qualified_name'
```
The issue is that even though the utils are in another file, that file *also* gets reloaded. Applying this diff to the bot will show that when reloading tags, the utils also gets reloaded:
```diff
diff --git a/bot/exts/info/tags/_utils.py b/bot/exts/info/tags/_utils.py
index d321dd81..d55174e4 100644
--- a/bot/exts/info/tags/_utils.py
+++ b/bot/exts/info/tags/_utils.py
@@ -14,6 +14,7 @@ from bot import constants
 from bot.log import get_logger

 log = get_logger(__name__)
+log.warn('TAGS COG UTILS HAVE BEEN RELOADED')
 
 __all__ = (
     "REGEX_NON_ALPHABET",
```
````
python-discord-bot-bot-1        | 2022-02-28 18:07:00 | bot | INFO | Cog loaded: BotSource
python-discord-bot-bot-1        | 2022-02-28 18:07:00 | bot.exts.utils.extensions | DEBUG | Extension successfully reloaded: `bot.exts.info.source`.
python-discord-bot-bot-1        | 2022-02-28 18:07:04 | bot.exts.info.tags._utils | WARNING | TAGS COG UTILS HAVE BEEN RELOADED
python-discord-bot-bot-1        | 2022-02-28 18:07:05 | bot | INFO | Cog loaded: Tags

Nipa-Code:
Hello!

I'm curios if there are any plans of continuing to work on this again?

wookie184:
I'm going to close this PR as:
- From bluenix's comment it seems this approach doesn't fix the issue (and I don't think this refactor is worth doing otherwise)
- There are quite a few conflicts
- It's gone a long time without any activity

PR #2048, opened 2022-01-16
Shorten the `TXT_EMBED_DESCRIPTION` message

swfarnsworth:
The previous version of the message gave different instructions for each possible reason that their message was deleted, which was rarely read in entirety and was sometimes confusing. This version simply instructs the user to use the paste bin.

### DIFF
--- a/bot/exts/filters/antimalware.py
+++ b/bot/exts/filters/antimalware.py
@@ -18,14 +18,8 @@ PY_EMBED_DESCRIPTION = (
 
 TXT_LIKE_FILES = {".txt", ".csv", ".json"}
 TXT_EMBED_DESCRIPTION = (
-    "**Uh-oh!** It looks like your message got zapped by our spam filter. "
-    "We currently don't allow `{blocked_extension}` attachments, "
-    "so here are some tips to help you travel safely: \n\n"
-    "• If you attempted to send a message longer than 2000 characters, try shortening your message "
-    "to fit within the character limit or use a pasting service (see below) \n\n"
-    "• If you tried to show someone your code, you can use codeblocks \n(run `!code-blocks` in "
-    "{cmd_channel_mention} for more information) or use a pasting service like: "
-    f"\n\n{URLs.site_schema}{URLs.site_paste}"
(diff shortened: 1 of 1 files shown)


bot/exts/filters/antimalware.py
```diff
@@ -18,14 +18,8 @@
...
-    "**Uh-oh!** It looks like your message got zapped by our spam filter. "
-    "We currently don't allow `{blocked_extension}` attachments, "
-    "so here are some tips to help you travel safely: \n\n"
-    "• If you attempted to send a message longer than 2000 characters, try shortening your message "
-    "to fit within the character limit or use a pasting service (see below) \n\n"
-    "• If you tried to show someone your code, you can use codeblocks \n(run `!code-blocks` in "
-    "{cmd_channel_mention} for more information) or use a pasting service like: "
-    f"\n\n{URLs.site_schema}{URLs.site_paste}"
+    "You either uploaded a `{blocked_extension}` file or entered a message that was too long. "
+    f"Please use our [{URLs.site_schema}{URLs.site_paste}](paste bin) instead."
```
wookie184, line 22:
```suggestion
    f"Please use a [{URLs.site_schema}{URLs.site_paste}](paste bin) instead."
```
-2 characters (/s)
Not really important but might make it clear they can use any pastebin, and the link to ours is just an option

wookie184 (approved):
Good change

bot/exts/filters/antimalware.py
swfarnsworth: Am I about to bikeshed a determiner?

If a link comes after a definite determiner, like "our", I would expect it to link to a specific thing. If it follows an indefinite determiner, I would expect it to link to information about what that thing is in general.

We can change the whole sentence to "You can use our paste bin instead".

wookie184: That's fair. You can change it to your suggestion there or leave it as is. (`our` could also be part of the link to emphasise the link is to our paste bin even more :D)

Xithrius (approved):
I like this. Thanks!

PR #2050, opened 2022-01-17
Fix pastebin hyperlink to use correct markdown syntax

Nipa-Code:
PR to fix pastebin hyperlink, it looked like this before. Hyperlink is now formatted to correct format.
image


HassanAbouelela (approved):
Looks good, thanks!

codephile1221 (approved):
LGTM.

Issue #1849, opened 2021-09-29
Create new voice-ban command and change the current voice-ban's name to voice-mute


MrHemlock:
Recently we've been having a problem with people joining and leaving the voice chat over and over and over.  I'm wanting to have a new command made that will not only prevent them from being verified (for as long as the join ban is active) as well as prevent them from rejoining the channel.

In addition to that, we'd need to rename the current `voiceban` command as `voicemute` to better reflect what they do.  

The implementation would likely be similar to how we have the mute command.

JacobMonck:
So just to make sure I understand this you want to rename the current command to `voicemute` and this new one `voiceban`? But besides the naming of the command I agree that this is useful and I would be happy to implement it if you don't have the time to.

MrHemlock:
Current command that we have will be named `voicemute`
New command will be named `voiceban`

jchristgit:
As part of our core developer meeting, we have decided that in order to have the site pull request taken care of for now, @ChrisLovering will implement a minimal patch in the bot to adjust the infraction type name in the backend to match the new name in the site, which will be merged together with the site pull request.

The main functionality of creating a new voice ban command will be implemented separately, but this way we keep work on the site project to a minimum (e.g. conflict resolution, backlog).

ChrisLovering:
The previous behaviour has now been migrated over to `voice_mute`. `voice_ban` currently has stub commands, so who ever picks up this ticket show implement those as described above.

TizzySaurus:
@MrHemlock is this a feature that's still wanted? If so, I'd be happy to work on it (assuming @JacobMonck no longer wants to).

Eggetxd:
Hi! Me and my friends want to implement the `!voiceban` command in this issue! could you please assign me?

jb3:
Hi, after speaking with a few Moderators we feel this is no longer a priority issue and functionality we want, most moderation action that would've required this can just be handled via other infraction types we already have.

PR #2052, opened 2022-01-21
Migrate to voice_mute over voice_ban

ChrisLovering:
Works towards https://github.com/python-discord/bot/issues/1849

This PR changes everything related to voice_bans to instead use voice_mute so that the bot is in line with the new site API.
As such, this should be merged at the same time as https://github.com/python-discord/site/pull/608.
I've done it this way so that the site PR doesn't need to be held up until the whole voice_ban implementation is completed.

This also adds stub commands for voice_ban , as they are useful for moderators during the change over from voice_ban to voice_mute, to remind them that the command has been renamed.

Reviewing may be easier by reviewing individual commits, rather than the full diff.


jchristgit (approved):
Thank you Chris. This is great!

ks129 (approved):

Issue #1960, opened 2021-11-17
Improvements to `helpdm` and a feature request for help channel notifications.


Qwerty-133:
This issue consists of two parts:

1. I believe we can improve `helpdm`'s embeds by making changes such as:
   - Adding a mention to the claimant
   - Setting their profile picture as the thumbnail
   - Adding the condensed text that the claimant themselves receive
   - Updating the embed when the help channel has been closed

2. While `helpdm` sometimes comes in handy when juggling between multiple help channels, what I would like more is to have the option to be notified for messages that I might have missed from the claimant. In terms of implementation, getting a DM from the bot if I haven't sent a message in the help channel 5 minutes after the claimant's message is what I had in mind.

wookie184:
I agree with all points made, and think they would make the helpdm system much more useful.

Were you intereseted in implementing any of this @Qwerty-133? (regardless, as a note if anyone would like to implement just part of this that would be fine as there's a couple of different things here)

Qwerty-133:
I'll start working on `helpdm` for now.

minalike:
With our migration to forum channels this week we are now using threads (posts) instead of channels, and closed/dormant channels now get referred to as `#deleted-channel`.

I'd like to suggest we add the channel name as text in the embed somewhere as well.

PR #2056, opened 2022-01-23
Add more information to helpdm embeds.

Qwerty-133:
Implements part 1 of #1960

### DIFF
--- a/bot/exts/help_channels/_caches.py
+++ b/bot/exts/help_channels/_caches.py
@@ -29,7 +29,9 @@ dynamic_message = RedisCache(namespace="HelpChannels.dynamic_message")
 # RedisCache[discord.User.id, bool]
 help_dm = RedisCache(namespace="HelpChannels.help_dm")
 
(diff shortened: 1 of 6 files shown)


MarkKoz (changes requested):
I'd rather see most, if not all the help dm code in the `_messages.py` file or some new module rather than in the cog's file. The cog already has quite a lot going on, and we should try to avoid making it even harder to read.

Xithrius:
@Qwerty-133 What's your status on this PR?

Thanks!

Qwerty-133:
Hey @Xithrius! I'll be finishing this up soon, thanks for the reminder

Xithrius:
@Qwerty-133 Hey, been a while. Thoughts on reviving this PR?

Thanks!

wookie184:
Since this is quite old and there are a lot of conflicts since this was made before we moved to forum help channels, i'll close this. It probably also needs some more discussion about what parts are still relevant/need changing on the issue before continuing.

PR #2061, opened 2022-01-25
setuptools use stdlib distutils over embedded

ChrisLovering:
This is caused by an upstream issue with setuptools 60.* (via virtualenv) changeing the default to using the setuptools-embedded distutils rather than the stdlib distutils, which breaks within pip's isolated builds.

This is explained quite well here https://github.com/pre-commit/pre-commit/issues/2178#issuecomment-1002163763

This also fixes a small inconsistency with our docker-compose file.


MarkKoz (approved):

ToxicKidz (approved):

HassanAbouelela (approved):
What's up with the restart policy Chris. I want clean PRs Chris.

Issue #2060, opened 2022-01-23
State claimer on help channels


aphkyle:
### What feature are you proposing?

Edit the "Available help channel" embed 
![image](https://user-images.githubusercontent.com/81857274/150662919-9f9389e4-34fb-4568-b635-8ba0b4079c12.png)
to
![image](https://user-images.githubusercontent.com/81857274/150663135-f653a7f5-3750-4594-bce6-255cd483401d.png)

### What is your justification for the proposed feature?

I've been visiting some help channels,
I found that some claimers delete their messages. Then, someone else who isn't the claimant came and asked questions. Adding this feature to make helpers know who the claimer is.

minalike:
Below is one example of a scenario that can potentially be avoided
![image](https://user-images.githubusercontent.com/75038675/151014378-d5b0228c-4b95-4193-a4e1-c3aab663ef60.png)

minalike:
Below is an example of a scenario that can potentially be avoided, or resolved a bit more quickly. This was a perfectly civil resolution, but sometimes these situations can also spiral.

![image](https://user-images.githubusercontent.com/75038675/151014799-038c41a1-8287-44e0-ad97-6d263391fb36.png)

ChrisLovering:
Myself, @HassanAbouelela and @Qwerty-133 discussed this in #dev-contrib around [here](https://canary.discord.com/channels/267624335836053506/635950537262759947/935571836576612399) and came to the conclusion that sending an embed after pinning the opener message would be better than editing the available one.

This is in-line with what we do for the available and dormant embeds, and also leads to better UX, especially if the opener message is long, since there's little use editing a message that's off the screen already.

The embed that gets sent should just have a title with the person who claimed it, maybe an thumbnail that shows the same emoji and the in use help channel; category does.

ChrisLovering:
For anyone wanting to pick this up, this is the function that manages what gets executed when a channel is claimed.
https://github.com/python-discord/bot/blob/main/bot/exts/help_channels/_cog.py#L104

PR #2062, opened 2022-01-27
Add embed message mentioning help channel claimant

minalike:
Closes python-discord/bot#2060

This PR adds a message by the bot stating who the help channel claimant is.
image

This is my first code contribution to Python Discord and my first time working with discord.py so during your review it is safe to assume that I know nothing.

I'm not sure if the embed should be sent before or after the channel is moved. But I would like it to be sent at least before the claimant's message is pinned.


ToxicKidz (approved):
Works well and looks good to me, thanks for the change!

Issue #2068, opened 2022-02-01
Add Moderator pings for filtered usernames when they are active


minalike:
Currently, moderators receive a non-pinging alert in `#mod-alerts` when a username containing a filtered word sends a message. The moderators would like to add a Moderators ping to this alert.

https://github.com/python-discord/bot/blob/2f64912bf4fd0d8bf1f55cf97cb4d4aef5ea81a3/bot/exts/filters/filtering.py#L236-L268

We also want to add the same ping for when such a user joins a voice channel. However, this type of alert is not currently in the code and would require another listener.

TizzySaurus:
Looking over the file, there's also some inconsistencies in imports which could be addressed in the same PR:

[L690](https://github.com/python-discord/bot/blob/2f64912bf4fd0d8bf1f55cf97cb4d4aef5ea81a3/bot/exts/filters/filtering.py#L690) uses the [`from discord import NotFound`](https://github.com/python-discord/bot/blob/2f64912bf4fd0d8bf1f55cf97cb4d4aef5ea81a3/bot/exts/filters/filtering.py#L15) import, whilst [L356](https://github.com/python-discord/bot/blob/2f64912bf4fd0d8bf1f55cf97cb4d4aef5ea81a3/bot/exts/filters/filtering.py#L356) uses the [`import discord.errors`](https://github.com/python-discord/bot/blob/2f64912bf4fd0d8bf1f55cf97cb4d4aef5ea81a3/bot/exts/filters/filtering.py#L9) import.

I'd personally solely use the `discord.errors` variant, meaning `NotFound` can be removed from the import on L15. Should probably also remove `HTTPException` in favour of `discord.errors.HTTPException` too, for consistency.

PR #2069, opened 2022-02-01
Ping Moderators On Bad Words In Username, Also Checks For Bad Words When Joining Or Switching Voicechannels

GDWR:
closing #2068

Enjoy the gitmojis 🎉

### DIFF
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -12,7 +12,7 @@ import tldextract
 from async_rediscache import RedisCache
 from botcore.regex import DISCORD_INVITE
 from dateutil.relativedelta import relativedelta
-from discord import Colour, HTTPException, Member, Message, NotFound, TextChannel, VoiceState, Forbidden
+from discord import Colour, Forbidden, HTTPException, Member, Message, NotFound, TextChannel, VoiceState
 from discord.ext.commands import Cog
 from discord.utils import escape_markdown
 
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -209,10 +209,9 @@ class Filtering(Cog):
         await self._filter_message(after, delta)
 
     @Cog.listener()
-    async def on_voice_state_update(self, member: Member, before: VoiceState, after: VoiceState) -> None:
-        """Checks for bad words in usernames when users initially joins a voice channel."""
-        if after.channel and not before.channel:
-            await self.check_bad_words_in_name(member)
+    async def on_voice_state_update(self, member: Member, *_) -> None:
+        """Checks for bad words in usernames when users join, switch or leave a voice channel."""
+        await self.check_bad_words_in_name(member)
 
     def get_name_match(self, name: str) -> Optional[re.Match]:
         """Check bad words from passed string (name). Return the first match found."""
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -208,8 +208,8 @@ class Filtering(Cog):
         await self._filter_message(after, delta)
 
     @Cog.listener()
-    async def on_voice_state_update(self, member: Member, before: VoiceState, after: VoiceState):
-        """When a member initially joins a voice channel or switches to a new one, check for a bad name in username."""
+    async def on_voice_state_update(self, member: Member, before: VoiceState, after: VoiceState) -> None:
+        """Checks for bad words in usernames when users join or change voice channels."""
         if after.channel and not before.channel:
             await self.check_bad_words_in_name(member)
 
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -6,13 +6,14 @@ from typing import Any, Dict, List, Mapping, NamedTuple, Optional, Tuple, Union
 
 import arrow
 import dateutil.parser
-import discord
 import regex
 import tldextract
 from async_rediscache import RedisCache
 from botcore.regex import DISCORD_INVITE
 from dateutil.relativedelta import relativedelta
-from discord import Colour, Forbidden, HTTPException, Member, Message, NotFound, TextChannel, VoiceState
+from discord import (
+    ChannelType, Colour, Embed, Forbidden, HTTPException, Member, Message, NotFound, TextChannel, VoiceState
+)
 from discord.ext.commands import Cog
 from discord.utils import escape_markdown
 
@@ -63,14 +64,14 @@ AUTO_BAN_REASON = (
 )
 AUTO_BAN_DURATION = timedelta(days=4)
 
-FilterMatch = Union[re.Match, dict, bool, List[discord.Embed]]
+FilterMatch = Union[re.Match, dict, bool, List[Embed]]
 
 
 class Stats(NamedTuple):
     """Additional stats on a triggered filter to append to a mod log."""
 
     message_content: str
-    additional_embeds: Optional[List[discord.Embed]]
+    additional_embeds: Optional[List[Embed]]
 
 
 class Filtering(Cog):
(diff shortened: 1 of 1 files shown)


bot/exts/filters/filtering.py
```diff
@@ -207,6 +207,12 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
             delta = relativedelta(after.edited_at, before.edited_at).microseconds
         await self._filter_message(after, delta)
 
+    @Cog.listener()
+    async def on_voice_state_update(self, member: Member, before: VoiceState, after: VoiceState) -> None:
+        """Checks for bad words in usernames when users join or change voice channels."""
```
TizzySaurus, line 212: This docstring isn't accurate. The `if after.channel and not before.channel` won't trigger when a member changes voice channels. It only triggers when a user joins a voice channel and wasn't in a voice channel previously.

If you want the described behaviour, you'd probably want `if after.channel and after.channel != before.channel`, although I personally think the current behaviour is better.

TizzySaurus (changes requested):
Main point is just a minor issue with the docstring.

Side notes:
- [L418](https://github.com/python-discord/bot/blob/fdb702d3e186a6d5020c9b3cd32c2f0588542c56/bot/exts/filters/filtering.py#L418) uses `discord.Message` despite `Message` being imported separately (we should probably change to `Message`).

- If we change above, the only remaining usage of `discord.xxx` is `discord.Embed` and `discord.ChannelType`. If we wanted to we could remove these in favour of a `from` import, saving having to need `import discord`.

Up to you whether you'd like to implement the side notes or not, but I'd like to see the main point resolved.

bot/exts/filters/filtering.py
GDWR: good point, I'll clarify in the docstring

"""Checks for bad words in usernames when users initially joins a voice channel."""

GDWR:
> Main point is just a minor issue with the docstring.
> 
> Side notes:
> 
>     * [L418](https://github.com/python-discord/bot/blob/fdb702d3e186a6d5020c9b3cd32c2f0588542c56/bot/exts/filters/filtering.py#L418) uses `discord.Message` despite `Message` being imported separately (we should probably change to `Message`).
> 
>     * If we change above, the only remaining usage of `discord.xxx` is `discord.Embed` and `discord.ChannelType`. If we wanted to we could remove these in favour of a `from` import, saving having to need `import discord`.
> 
> 
> Up to you whether you'd like to implement the side notes or not, but I'd like to see the main point resolved.

Would cause flake8 to complain about line length, and I think the following code block would be uglier
```py

from discord import (
  ChannelType,
  Colour,
  ...,
)
```


I cannot find any examples in the repo of how it would like to be handled when the imports exceed the line length. And I don't want to be the one to figure it out.

TizzySaurus:
> https://github.com/python-discord/bot/blob/fdb702d3e186a6d5020c9b3cd32c2f0588542c56/bot/exts/filters/filtering.py#L418

There's [multiple instances](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/python-discord/bot%24+import+%28&patternType=literal) of doing something along the lines of
```py
from discord import (
    ChannelType, Colour, ...
)```which would fit within the 120 lines, so that would be the approach should you want to implement this.

File: bot/exts/filters/filtering.py (hunk 1 above)
Comment (TizzySaurus) [reply bot/exts/filters/filtering.py]: The "joins" should technically be singular, but I think that's too small to be worth fixing. Thanks for updating :+1:

**Comment by GDWR** [comment]
> > I cannot find any examples in the repo of how it would like to be handled when the imports exceed the line length. And I don't want to be the one to figure it out.
> 
> There's [multiple instances](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/python-discord/bot%24+import+%28&patternType=literal) of doing something along the lines of
> 
> ```python
> from discord import (
>     ChannelType, Colour, ...
> )
> ```
> 
> which would fit within the 120 lines, so that would be the approach should you want to implement this.

lol, my search for `from discord (` was a bad choice. I will use this.

**Comment by TizzySaurus** [comment]
Haven't tested anything but the code is looking good :+1:

I'll test this as soon as I get a chance, and leave a formal review after :+1:

File: bot/exts/filters/filtering.py [hunk 2]
```diff
@@ -207,6 +208,12 @@ async def on_message_edit(self, before: Message, after: Message) -> None:
             delta = relativedelta(after.edited_at, before.edited_at).microseconds
         await self._filter_message(after, delta)
 
+    @Cog.listener()
+    async def on_voice_state_update(self, member: Member, before: VoiceState, after: VoiceState) -> None:
+        """Checks for bad words in usernames when users initially joins a voice channel."""
```
Xithrius, line 213: Given it's `on_voice_state_update`, should we be more specific as to when this is triggered?
```suggestion
        """Checks for bad words in usernames when users join or switch voice channels."""
```

GDWR: With the condition being `if after.channel and not before.channel:`, it will only trigger when a member initially joins a voice channel. 

This is because we are checking the `after.channel` is not `None`, meaning they are now in a voice channel.
And we are also checking if `before.channel` was `None`, meaning they was not previously in a voice channel.

This condition isn't very easy to read, should I change it to the following for readability?

```python
if after.channel is not None and before.channel is None:
```

TizzySaurus: > This condition isn't very easy to read, should I change it to the following for readability?
> 
> ```python
> if after.channel is not None and before.channel is None:
> ```

I personally don't really see this as an improvement. I actually find it better as-is.

> Additionally, I've been asked for it to only be triggered on initial joins.

To further expand on this: I requested this since there's no point in triggering when switching voice channels. In order to switch channels, one must first initially join a voice channel -- meaning that the moderators have already been alerted of any bad name. TL;DR; By not checking when switching channels, we prevents duplicate pings.

GDWR: The `check_bad_words_in_name` only pings once every 3 days, unless the docstring is misleading.

TizzySaurus: > The `check_bad_words_in_name` only pings once every 3 days, unless the docstring is misleading.

Right. I didn't word my reasoning very well, but it still applies. It just saves a redundant function call.

GDWR: I am going to make this run on every single `voice_state_update` event, as we are already running this same function for every single message sent. So there shouldn't be any issues with slow down.

One thing we could do, if we are worried about that, is implementing an LRU cache on this [function](https://github.com/python-discord/bot/blob/2fc4f53bd7c3c669fc744ca307163d5c59eaa8a0/bot/exts/filters/filtering.py#L217). We would need to worry about invalidating the cache when the filters change. If we were to go with adding the LRU cache I would want to create a static function to be used for the LRU cache

Xithrius:
Remember to [lint before you push](https://youtu.be/hgI0p1zf31k)!

bot/exts/filters/filtering.py
```diff
@@ -6,13 +6,12 @@
 
 import arrow
 import dateutil.parser
-import discord.errors
```
MaskDuck, line 9: what exactly is the reason for this?

MaskDuck (changes requested):

bot/exts/filters/filtering.py
GDWR: As mentioned [here](https://github.com/python-discord/bot/pull/2069#pullrequestreview-869807491), swapping from this usage of `discord.errors` namespace to exact imports would keep the consistency of how the `discord` package is used across this file.

wookie184 (approved):
Very nice, looks good and works well!

bot/exts/filters/filtering.py
wookie184: Current behaviour seems 👌, will mark this as resolved

ChrisLovering (approved):

PR #2071, opened 2022-02-02
Notify helpers without ping when dormant channels are running low

### LINKED ISSUE #2051: Send a non-pinging message in `#helpers` when we only have 10% help channels available
# Description
We currently have a message in `#helpers` that goes out when all help channels are used and no new ones can be pulled up/created that pings all Helpers.

Unfortunately, this message is a bit too late as all the channels are already taken. 

We'd like to add a non-pinging message when we're at 90% help channel usage, to let people know we're running low on help channels. If helpers are free they can try to clear up some overdue channels before we run out.

We'd like to implement this along with some other solutions that are still being discussed.

minalike:
Closes #2051 

This PR adds a non-pinging notification to `#helpers` to alert staff when we are running low on dormant channels. The threshold for when this notification is triggered is added as a configurable.

image

### DIFF
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -80,6 +80,7 @@ class HelpChannels(commands.Cog):
 
         # Notifications
         self.notify_interval_seconds = constants.HelpChannels.notify_minutes * 60
+        # Using a very old date so that we don't have to use Optional typing.
         self.last_none_remaining_notification = arrow.get('1815-12-10T18:00:00.00000+00:00')
         self.last_running_low_notification = arrow.get('1815-12-10T18:00:00.00000+00:00')
 
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -79,7 +79,7 @@ class HelpChannels(commands.Cog):
         self.name_queue: t.Deque[str] = None
 
         # Notifications
-        self.notify_interval_seconds = (constants.HelpChannels.notify_minutes * 60)
+        self.notify_interval_seconds = constants.HelpChannels.notify_minutes * 60
         self.last_none_remaining_notification = arrow.get('1815-12-10T18:00:00.00000+00:00')
         self.last_running_low_notification = arrow.get('1815-12-10T18:00:00.00000+00:00')
 
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -79,7 +79,6 @@ class HelpChannels(commands.Cog):
         self.name_queue: t.Deque[str] = None
 
         # Notifications
-        self.notify_interval_seconds = constants.HelpChannels.notify_minutes * 60
         # Using a very old date so that we don't have to use Optional typing.
         self.last_none_remaining_notification = arrow.get('1815-12-10T18:00:00.00000+00:00')
         self.last_running_low_notification = arrow.get('1815-12-10T18:00:00.00000+00:00')
@@ -233,29 +232,31 @@ class HelpChannels(commands.Cog):
 
         try:
             channel = self.channel_queue.get_nowait()
-
-            time_since_last_notify_seconds = (arrow.utcnow() - self.last_running_low_notification).seconds
-            within_interval = time_since_last_notify_seconds >= self.notify_interval_seconds
-            if within_interval and self.channel_queue.qsize() <= constants.HelpChannels.notify_running_low_threshold:
-                await _message.notify_running_low(
-                    self.bot.get_channel(constants.HelpChannels.notify_channel),
-                    self.channel_queue.qsize()
-                )
-                self.last_running_low_notification = arrow.utcnow()
-
         except asyncio.QueueEmpty:
             log.info("No candidate channels in the queue; creating a new channel.")
             channel = await self.create_dormant()
 
             if not channel:
                 log.info("Couldn't create a candidate channel; waiting to get one from the queue.")
+                last_notification = await _message.notify_none_remaining(
+                    self.bot.get_channel(constants.HelpChannels.notify_channel),
+                    self.last_none_remaining_notification
+                )
 
-                if (arrow.utcnow() - self.last_none_remaining_notification).seconds >= self.notify_interval_seconds:
-                    await _message.notify_none_remaining(self.bot.get_channel(constants.HelpChannels.notify_channel))
-                    self.last_none_remaining_notification = arrow.utcnow()
-                    self.bot.stats.incr("help.out_of_channel_alerts")
+                if last_notification:
+                    self.last_none_remaining_notification = last_notification
 
-                channel = await self.wait_for_dormant_channel()
+                channel = await self.wait_for_dormant_channel()  # Blocks until a new channel is available
+
+        else:
+            last_notification = await _message.notify_running_low(
+                self.bot.get_channel(constants.HelpChannels.notify_channel),
+                self.channel_queue.qsize(),
+                self.last_running_low_notification
+            )
+
+            if last_notification:
+                self.last_running_low_notification = last_notification
 
         return channel
 
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -246,7 +246,10 @@ class HelpChannels(commands.Cog):
                 channel = await self.wait_for_dormant_channel()  # Blocks until a new channel is available
 
         else:
-            last_notification = await _message.notify_running_low(self.channel_queue.qsize(), self.last_running_low_notification)
+            last_notification = await _message.notify_running_low(
+                self.channel_queue.qsize(),
+                self.last_running_low_notification
+            )
 
             if last_notification:
                 self.last_running_low_notification = last_notification
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -78,7 +78,10 @@ class HelpChannels(commands.Cog):
         self.channel_queue: asyncio.Queue[discord.TextChannel] = None
         self.name_queue: t.Deque[str] = None
 
-        self.last_notification: t.Optional[arrow.Arrow] = None
+        # Notifications
+        self.notify_interval_seconds = (constants.HelpChannels.notify_minutes * 60)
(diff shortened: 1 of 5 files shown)


bot/exts/help_channels/_cog.py
```diff
@@ -229,16 +232,26 @@ async def get_available_candidate(self) -> discord.TextChannel:
...
             channel = self.channel_queue.get_nowait()
+
+            time_since_last_notify_seconds = (arrow.utcnow() - self.last_running_low_notification).seconds
+            within_interval = time_since_last_notify_seconds >= self.notify_interval_seconds
+            if within_interval and self.channel_queue.qsize() <= constants.HelpChannels.notify_running_low_threshold:
+                await _message.notify_running_low(
+                    self.bot.get_channel(constants.HelpChannels.notify_channel),
+                    self.channel_queue.qsize()
+                )
+                self.last_running_low_notification = arrow.utcnow()
```
MarkKoz, line 243: I think it'd be better to put in the `else` of the try-except, since the exception being caught has nothing to do with this code.

Also, what do you think about defining a predicate for this in `_messages.py` to reduce the amount of code in the cog? I'm always concerned that the cog is growing too big and things need to be refactored out of it to stay manageable. Such predicate could probably be used by both this code and the code in the `except` block.

bot/exts/help_channels/_cog.py
```diff
@@ -78,7 +78,10 @@ def __init__(self, bot: Bot):
         self.channel_queue: asyncio.Queue[discord.TextChannel] = None
         self.name_queue: t.Deque[str] = None
 
-        self.last_notification: t.Optional[arrow.Arrow] = None
+        # Notifications
+        self.notify_interval_seconds = (constants.HelpChannels.notify_minutes * 60)
```
MarkKoz, line 82: The parenthesis isn't necessary and I don't feel that it clarifies anything.

bot/exts/help_channels/_message.py
```diff
@@ -124,54 +123,59 @@ async def dm_on_open(message: discord.Message) -> None:
         )
 
 
-async def notify(channel: discord.TextChannel, last_notification: t.Optional[Arrow]) -> t.Optional[Arrow]:
+async def notify_none_remaining(channel: discord.TextChannel) -> None:
```
MarkKoz, line 126: Because `bot.instance` is available, this argument can be omitted. Instead, the function body can be hard-coded to use `constants.HelpChannels.notify_channel` as the channel. I don't see any benefit to allowing arbitrary channels in this case. Thoughts?

bot/exts/help_channels/_cog.py
MarkKoz: You should add stat for this similar to `help.out_of_channel_alerts`.

bot/exts/help_channels/_message.py
```diff
@@ -124,54 +123,59 @@ async def dm_on_open(message: discord.Message) -> None:
...
+    """
+    if not constants.HelpChannels.notify_running_low:
+        return
+
+    log.trace("Notifying about getting close to no dormant channels.")
+
+    await channel.send(
+        f"There are only {number_of_channels_left} dormant channels left. "
+        "Consider participating in some help channels so that we don't run out."
+    )
```
MarkKoz, line 176: This should be in a try-except too for similar reasons as the other notify functions.

MarkKoz (changes requested):

bot/exts/help_channels/_cog.py
```diff
@@ -78,7 +78,10 @@ def __init__(self, bot: Bot):
...
+        self.last_none_remaining_notification = arrow.get('1815-12-10T18:00:00.00000+00:00')
+        self.last_running_low_notification = arrow.get('1815-12-10T18:00:00.00000+00:00')
```
ChrisLovering, line 84: Are these just arbitrarily low dates? If so doing `arrow.get(datetime.min)` might make it more clear what this is trying to do.

GDWR: it is Ada Lovelace's birthday as a little "easter egg". it is just a very old date, so we don't have to use `Optional` typing.

Would you like for me to change it or comment it?

ChrisLovering: Yea, I think just a comment that it's just an arbitrarily old date is fine.

bot/exts/help_channels/_message.py
GDWR: I agree, the notify_channel is already configurable via the `config.yml`

MarkKoz (changes requested):

bot/exts/help_channels/_message.py
```diff
@@ -124,52 +124,93 @@ async def dm_on_open(message: discord.Message) -> None:
...
+
+    log.trace("Notifying about getting close to no dormant channels.")
+
+    channel = bot.instance.get_channel(constants.HelpChannels.notify_channel)
+    if channel is None:
+        log.trace("Did not send notify_running notification as the notification channel couldn't be gathered.")
+
+    try:
+        await channel.send(
+            f"There are only {number_of_channels_left} dormant channels left. "
```
MarkKoz, line 204: Edge case: if `number_of_channels_left` is `1` then it will state "There are only 1", but it should state "There _is_ only 1".

config-default.yml
```diff
@@ -513,19 +513,16 @@ help_channels:
...
-    # Channel in which to send notifications
-    notify_channel: *HELPERS
-
-    # Minimum interval between helper notifications
-    notify_minutes: 15
-
-    # Mention these roles in notifications
-    notify_roles:
+    notify_none_remaining:         true      # Pinging notification for the Helper role when no dormant channels remain
+    notify_none_remaining_roles:             # Mention these roles in the non_remaining notification
```
minalike, line 520: non_remaining > none_remaining

MarkKoz (dismissed):

config-default.yml
minalike: https://github.com/python-discord/bot/pull/2071/commits/75b3a515910f04f295e0b70c208e8383b1783b7d

bot/exts/help_channels/_message.py
```diff
@@ -166,14 +176,28 @@ async def notify_running_low(channel: discord.TextChannel, number_of_channels_le
...
+    log.trace("Notifying about getting close to no dormant channels.")
+    try:
+        await channel.send(
+            f"There are only {number_of_channels_left} dormant channels left. "
+            "Consider participating in some help channels so that we don't run out."
+        )
+    except Exception:
+        # Handle it here cause this feature isn't critical for the functionality of the system.
+        log.exception("Failed to send notification about running low of dormant channels!")
+    finally:
```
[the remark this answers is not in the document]
MarkKoz: @minalike oops I forgot to check this. You didn't actually fix these.

bot/exts/help_channels/_message.py
```diff
@@ -124,52 +124,93 @@ async def dm_on_open(message: discord.Message) -> None:
...
+        * `HelpChannels.notify_minutes`              - minimum interval between notifications
+        * `HelpChannels.notify_none_remaining`       - toggle none_remaining notifications
+        * `HelpChannels.notify_none_remaining_roles` - roles mentioned in notifications
     """
-    if not constants.HelpChannels.notify:
-        return
+    if not constants.HelpChannels.notify_none_remaining:
+        return None
+
+    if (arrow.utcnow() - last_notification).seconds < (constants.HelpChannels.notify_minutes * 60):
```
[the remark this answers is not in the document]
GDWR: I think you are correct, i'll check docs.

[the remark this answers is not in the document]
minalike: https://github.com/python-discord/bot/pull/2071/commits/6c7444d6f61fe201f97673de435a63d4b4f67c7d

bot/exts/help_channels/_message.py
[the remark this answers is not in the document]
GDWR: Yeah, we were playing around with the concept of keeping the finally for the metric only. But this would cause metrics to be sent, even if the message wasn't successfully sent to the helpers. 

Should now be corrected.
[commit](https://github.com/python-discord/bot/pull/2071/commits/54480a1a166511c8b85dbb46403247144138fd24) 

The diff is a little grim, so here are the exact files from the latest commit.
[here](https://github.com/python-discord/bot/blob/6c7444d6f61fe201f97673de435a63d4b4f67c7d/bot/exts/help_channels/_message.py#L209-L214) and [here](https://github.com/python-discord/bot/blob/6c7444d6f61fe201f97673de435a63d4b4f67c7d/bot/exts/help_channels/_message.py#L209-L214)

bot/exts/help_channels/_message.py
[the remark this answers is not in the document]
MarkKoz: Why was it working before then?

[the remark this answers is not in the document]
ChrisLovering: For timedeltas like this, I believe the resolution is days, then seconds. So if the last_notificaiton isn't greater that a day ago, `.seconds` and `.total_secoonds()` would be the same.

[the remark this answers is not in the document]
GDWR: Ah, when I was testing it was in the negative which seems to have major issues. i.e.
```py
>>> import arrow
>>> past = arrow.utcnow()
>>> range = past - arrow.utcnow()
>>> range.seconds
86391
>>> range.total_seconds()
-8.216756
>>>
```

For a positive time it seems to work fine?
```py
>>> import arrow
>>> past = arrow.utcnow()
>>> range = arrow.utcnow() - past
>>> range.total_seconds()
72.76027
>>> range.seconds
72
```

Stuggling to finds docs, probs will have to go source.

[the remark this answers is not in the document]
ChrisLovering: Negative time deltas use the days component to go back to the day in question, then the seconds component to go "up" to the correct time.

So the seconds component will be quite high due to starting from midnight, whereas the total_seconds() function resolves this to just a negative number of seconds.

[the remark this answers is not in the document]
MarkKoz: `total_seconds` seems better in general, so I am fine with using it. Good catch regardless.

MarkKoz (approved):

ToxicKidz (approved):
LGTM :shipit:

PR #2078, opened 2022-02-09
Ensure claimant cache is populated on help channel init

ChrisLovering:
These changes ensure that for every help channel in use, a claimant is present in the redis cache. This is to avoid an issue caused when a user tries to close a channel, but the cache is empty so the author check fails.

This also cancels a help channel claim if we hit a discord 500 error while moving the channel.

### DIFF
--- a/bot/exts/help_channels/_channel.py
+++ b/bot/exts/help_channels/_channel.py
@@ -1,3 +1,4 @@
+import re
 import typing as t
 from datetime import timedelta
 from enum import Enum
@@ -16,6 +17,7 @@ log = get_logger(__name__)
 
 MAX_CHANNELS_PER_CATEGORY = 50
 EXCLUDED_CHANNELS = (constants.Channels.cooldown,)
+CLAIMED_BY_RE = re.compile(r"Channel claimed by <@!?(?P<user_id>\d{17,20})>\.$")
 
 
 class ClosingReason(Enum):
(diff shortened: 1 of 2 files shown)


GDWR (approved):

bot/exts/help_channels/_channel.py
```diff
@@ -157,3 +159,35 @@ async def move_to_bottom(channel: discord.TextChannel, category_id: int, **optio
...
+    async for message in channel.history(limit=1000):
+        if message.author.id != bot.instance.user.id:
+            # We only care about bot messages
+            continue
+        if message.embeds:
+            if _message._match_bot_embed(message, _message.DORMANT_MSG):
+                log.info("Hit the dormant message embed before finding a claimant in %s (%d).", channel, channel.id)
+                break
+            user_id = CLAIMED_BY_RE.match(message.embeds[0].description).group("user_id")
+            await _caches.claimants.set(channel.id, user_id)
```
ToxicKidz, line 186: This should be cast to `int`.
```suggestion
            await _caches.claimants.set(channel.id, int(user_id))
```

ToxicKidz (changes requested):
Looks good, one bug in `ensure_cached_claimant`.

bot/exts/help_channels/_channel.py
ChrisLovering: Nice catch :D

ChrisLovering:
Will squash before merging

ToxicKidz (approved):
Works well :shipit:

PR #2080, opened 2022-02-11
Don't validate reminder author

mbaruh:
Reminder validation relied on the member cache which might not be properly filled, which caused some reminders to not be sent.

This PR removes this part of the validation. This can cause reminders to be sent for users who are no longer in the server, but this seems negligible.

While #1919 might be more thorough, the changes here are very simple and will allow us to have the reminders cog again.

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -66,20 +66,19 @@ class Reminders(Cog):
             else:
                 self.schedule_reminder(reminder)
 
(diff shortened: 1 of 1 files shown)


jb3 (approved):

mbaruh:
Don't be alarmed by the line count, the merge commit brought back the entire file. The actual change is the first commit which is an 8-line change

ChrisLovering (approved):

PR #2084, opened 2022-02-13
Add a cog to bump threads

ChrisLovering:
Quite often we want threads such as event discussions, or moderation discussions to live beyond their maximum of 1 week of auto-archival. This cog allows staff to add a thread to a list that will get 'bumped' back open by the bot when they are auto-archived

### DIFF
--- a/bot/exts/utils/thread_bumper.py
+++ b/bot/exts/utils/thread_bumper.py
@@ -8,7 +8,7 @@ from bot import constants
 from bot.bot import Bot
 from bot.log import get_logger
 from bot.pagination import LinePaginator
-from bot.utils import scheduling
+from bot.utils import channel, scheduling
 
 log = get_logger(__name__)
 
@@ -23,22 +23,50 @@ class ThreadBumper(commands.Cog):
         self.bot = bot
         self.init_task = scheduling.create_task(self.ensure_bumped_threads_are_active(), event_loop=self.bot.loop)
 
+    async def unarchive_threads_not_manually_archived(self, threads: list[discord.Thread]) -> None:
+        """
+        Iterate through and unarchive any threads that weren't manually archived recently.
+
+        This is done by extracting the manually archived threads from the audit log.
+
+        Only the last 200 thread_update logs are checked,
+        as this is assumed to be more than enough to cover bot downtime.
+        """
+        guild = self.bot.get_guild(constants.Guild.id)
+
+        recent_manually_archived_thread_ids = []
+        async for thread_update in guild.audit_logs(limit=200, action=discord.AuditLogAction.thread_update):
+            if getattr(thread_update.after, "archived", False):
+                recent_manually_archived_thread_ids.append(thread_update.target.id)
+
+        for thread in threads:
+            if thread.id in recent_manually_archived_thread_ids:
+                log.info(
+                    "#%s (%d) was manually archived. Leaving archived, and removing from bumped threads.",
+                    thread.name,
+                    thread.id
+                )
+                await self.threads_to_bump.delete(thread.id)
+            else:
+                await thread.edit(archived=False)
+
     async def ensure_bumped_threads_are_active(self) -> None:
         """Ensure bumped threads are active, since threads could have been archived while the bot was down."""
         await self.bot.wait_until_guild_available()
 
+        threads_to_maybe_bump = []
         for thread_id, _ in await self.threads_to_bump.items():
-            if thread := self.bot.get_channel(thread_id):
-                if not thread.archived:
-                    continue
-
             try:
-                thread = await self.bot.fetch_channel(thread_id)
+                thread = await channel.get_or_fetch_channel(thread_id)
             except discord.NotFound:
-                log.info(f"Thread {thread_id} has been deleted, removing from bumped threads.")
+                log.info("Thread %d has been deleted, removing from bumped threads.", thread_id)
                 await self.threads_to_bump.delete(thread_id)
+                continue
+
             if thread.archived:
-                await thread.edit(archived=False)
+                threads_to_maybe_bump.append(thread)
+
+        await self.unarchive_threads_not_manually_archived(threads_to_maybe_bump)
 
     @commands.group(name="bump")
     async def thread_bump_group(self, ctx: commands.Context) -> None:
@@ -100,7 +128,7 @@ class ThreadBumper(commands.Cog):
 
         bumped_threads = [k for k, _ in await self.threads_to_bump.items()]
         if after.id in bumped_threads:
-            await after.edit(archived=False)
+            await self.unarchive_threads_not_manually_archived([after])
 
     async def cog_check(self, ctx: commands.Context) -> bool:
         """Only allow staff & partner roles to invoke the commands in this cog."""
--- a/bot/exts/utils/thread_bumper.py
+++ b/bot/exts/utils/thread_bumper.py
@@ -74,7 +74,7 @@ class ThreadBumper(commands.Cog):
         if not ctx.invoked_subcommand:
             await ctx.send_help(ctx.command)
 
-    @thread_bump_group.command(name="add")
+    @thread_bump_group.command(name="add", aliases=("a",))
     async def add_thread_to_bump_list(self, ctx: commands.Context, thread: t.Optional[discord.Thread]) -> None:
         """Add a thread to the bump list."""
         await self.init_task
@@ -84,6 +84,9 @@ class ThreadBumper(commands.Cog):
                 thread = ctx.channel
             else:
                 raise commands.BadArgument("You must provide a thread, or run this command within a thread.")
+
+        if await self.threads_to_bump.contains(thread.id):
+            raise commands.BadArgument("This thread is already in the bump list.")
 
         await self.threads_to_bump.set(thread.id, "sentinel")
         await ctx.send(f":ok_hand:{thread.mention} has been added to the bump list.")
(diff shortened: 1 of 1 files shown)


bot/exts/utils/thread_bumper.py
```diff
@@ -0,0 +1,112 @@
...
+        If the thread has been archived, and is in the bump list, un-archive it.
+        """
+        await self.init_task
+
+        if not after.archived:
+            return
+
+        bumped_threads = [k for k, _ in await self.threads_to_bump.items()]
+        if after.id in bumped_threads:
+            await after.edit(archived=False)
```
vivekashok1221, line 101: Would manually archived threads be bumped too? I don't think that'd be a desired behaviour.
[Thread.archiver_id](<https://discordpy.readthedocs.io/en/master/api.html?highlight=on_thread_update#discord.Thread.archiver_id>) is `Optional[int]`. I'm assuming it will not be `None` in cases of manual archiving. I've not tested it but I guess you could use that to check if a thread was manually archived.

vivekashok1221: Please correct me if I'm wrong. As perfect as I might seem, I make sometimes make mistakes.

vivekashok1221 (changes requested):
Neat and concise!

bot/exts/utils/thread_bumper.py
ChrisLovering: Hmmm, I've tried this in the on_thread_update event listener, and it doesn't seem to be set at all, on the before or after thread object when manually archiving.

i also can't see where this data would come from, since it's not present on the discord api https://discord.com/developers/docs/resources/channel#thread-metadata-object and in the discord.py source, its tries to pull this key directly form the data from the api.

ChrisLovering: So possibly it used to be there, but isn't anymore? 🤷

vivekashok1221: I've been bamboozled!

> As of June 10th, 2021, this value will always be None since Discord doesn’t provide this information anymore.
This property will be removed in a future version.

[(From Disnake docs)](<https://docs.disnake.dev/en/latest/api.html?highlight=on_thread_update#disnake.Thread.archiver_id>)

I guess we can manually remove the thread from the bump list before archiving it.

ChrisLovering: Ah, that explains it.

The only other option I can think of is to take the last message in the thread, add the auto-archive time to it and see if that is close to the current time, or in the past.

ChrisLovering: Done in https://github.com/python-discord/bot/pull/2084/commits/2d8fa806aff55e5efdbd6b74440acf78c256fdd5

onerandomusername: Hi, saw this and wanted to explain a 100% guaranteed way to know if it was manual or automatic: manual closes will have an audit log entry. If there's no relevant audit log entry then it safe to assume that the thread closed automatically, rather than manual.

ChrisLovering: > Hi, saw this and wanted to explain a 100% guaranteed way to know if it was manual or automatic: manual closes will have an audit log entry. If there's no relevant audit log entry then it safe to assume that the thread closed automatically, rather than manual.

Were you expecting something like this?
https://github.com/python-discord/bot/pull/2084/commits/0ae5ee9c478bd4a627473acce8a46de3965176b9

bot/exts/utils/thread_bumper.py
```diff
@@ -0,0 +1,145 @@
...
+
+        This is done by extracting the manually archived threads from the audit log.
+
+        Only the last 200 thread_update logs are checked,
+        as this is assumed to be more than enough to cover bot downtime.
+        """
+        guild = self.bot.get_guild(constants.Guild.id)
+
+        recent_manually_archived_thread_ids = []
+        for thread_update in guild.audit_logs(limit=200, action=discord.AuditLogAction.thread_update):
```
ToxicKidz, line 38: [`Guild.audit_logs()`](https://discordpy.readthedocs.io/en/stable/api.html?highlight=audit_logs#discord.Guild.audit_logs) is an async iterator.
```suggestion
        async for thread_update in guild.audit_logs(limit=200, action=discord.AuditLogAction.thread_update):
```

ToxicKidz (changes requested):

bot/exts/utils/thread_bumper.py
ChrisLovering: hah, you're right, I had this as async locally before I refactored the function and missed this :(

ChrisLovering: https://github.com/python-discord/bot/pull/2084/commits/ef2717706458de4da1441726a0eed1a42bababe8

bot/exts/utils/thread_bumper.py
```diff
@@ -0,0 +1,144 @@
...
+        for thread_id, _ in await self.threads_to_bump.items():
+            if thread := self.bot.get_channel(thread_id):
+                if not thread.archived:
+                    continue
+
+            try:
+                thread = await self.bot.fetch_channel(thread_id)
+            except discord.NotFound:
+                log.info(f"Thread {thread_id} has been deleted, removing from bumped threads.")
+                await self.threads_to_bump.delete(thread_id)
```
ToxicKidz, line 67: This should return after this line, otherwise thread won't be set, causing an `AttributeError`.

ToxicKidz (changes requested):

bot/exts/utils/thread_bumper.py
ChrisLovering: Yea, will add a continue after this line

ChrisLovering: https://github.com/python-discord/bot/pull/2084/commits/e50715d0fa536041238378da7ca9a67f14243547

bot/exts/utils/thread_bumper.py
```diff
@@ -0,0 +1,142 @@
...
+    @thread_bump_group.command(name="add")
+    async def add_thread_to_bump_list(self, ctx: commands.Context, thread: t.Optional[discord.Thread]) -> None:
+        """Add a thread to the bump list."""
+        await self.init_task
+
+        if not thread:
+            if isinstance(ctx.channel, discord.Thread):
+                thread = ctx.channel
+            else:
+                raise commands.BadArgument("You must provide a thread id, or run this command in a thread.")
```
ToxicKidz, line 86: This might need to be updated.

ToxicKidz (approved):
LGTM, works well, :shipit:, just one comment.

bot/exts/utils/thread_bumper.py
ChrisLovering: Changed this to `You must provide a thread, or run this command within a thread.` during the squash.

bot/exts/utils/thread_bumper.py
```diff
@@ -0,0 +1,142 @@
...
+
+        await self.unarchive_threads_not_manually_archived(threads_to_maybe_bump)
+
+    @commands.group(name="bump")
+    async def thread_bump_group(self, ctx: commands.Context) -> None:
+        """A group of commands to manage the bumping of threads."""
+        if not ctx.invoked_subcommand:
+            await ctx.send_help(ctx.command)
+
+    @thread_bump_group.command(name="add")
```
vivekashok1221, line 77:
```suggestion
    @thread_bump_group.command(name="add", aliases=("a",))
```
How about adding `a` as an alias to `add`? Feel free to ignore this comment because `add` is short as it is, but, personally, I would expect `a` to be an alias for `add` since `r` is an alias for `remove`.

bot/exts/utils/thread_bumper.py
```diff
@@ -0,0 +1,142 @@
...
+        """Add a thread to the bump list."""
+        await self.init_task
+
+        if not thread:
+            if isinstance(ctx.channel, discord.Thread):
+                thread = ctx.channel
+            else:
+                raise commands.BadArgument("You must provide a thread, or run this command within a thread.")
+
+        await self.threads_to_bump.set(thread.id, "sentinel")
```
vivekashok1221, line 88: When trying to add a thread which is already in the bump list, it might be useful to send a message like "Thread is already in the bump list. ".

vivekashok1221 (changes requested):
Two minor suggestions, feel free to ignore if you feel they're not needed.

bot/exts/utils/thread_bumper.py
```diff
@@ -0,0 +1,142 @@
...
+        If the thread has been archived, and is in the bump list, un-archive it.
+        """
+        await self.init_task
+
+        if not after.archived:
+            return
+
+        bumped_threads = [k for k, _ in await self.threads_to_bump.items()]
+        if after.id in bumped_threads:
+            await self.unarchive_threads_not_manually_archived([after])
```
vivekashok1221, line 131: PS: I believe you can do something like this here instead.
```suggestion
        if await bumped_threads.contains(after.id):
            await self.unarchive_threads_not_manually_archived([after])
```

vivekashok1221: You could also use [`RedisCache.contains()`](https://github.com/SebastiaanZ/async-rediscache/blob/6a68850e48cc36f0a60a3ddd12bbf3c908e10380/async_rediscache/types/cache.py#L54-L55) in line 88 if you decide to incorporate [this](https://github.com/python-discord/bot/pull/2084/files#r810018069) suggestion.

vivekashok1221 (changes requested):
Sorry, but I one more thing I wanted to point out  :sweat_smile: .

bot/exts/utils/thread_bumper.py
ChrisLovering: https://github.com/python-discord/bot/pull/2084/commits/3346d71416fdb3223d0c4998f92e420886445fac

bot/exts/utils/thread_bumper.py
ChrisLovering: https://github.com/python-discord/bot/pull/2084/commits/3346d71416fdb3223d0c4998f92e420886445fac

PR #2086, opened 2022-02-14
Fix DM handling for code snippets.

### LINKED ISSUE #2081: Cannot suppress embeds in a DM channel
Sentry Issue: [BOT-2Y7](https://sentry.io/organizations/python-discord/issues/3011204281/?referrer=github_integration)

```
Forbidden: 403 Forbidden (error code: 50003): Cannot execute action on a DM channel
  File "discord/client.py", line 351, in _run_event
    await coro(*args, **kwargs)
  File "bot/exts/info/code_snippets.py", line 254, in on_message
    await message.edit(suppress=True)
  File "discord/message.py", line 1289, in edit
    data = await self._state.http.edit_message(self.channel.id, self.id, **payload)
  File "discord/http.py", line 329, in request
    raise Forbidden(response, data)

```

This seems to be caused by the code snippet cog trying to remove the embed from the user's message, but doesn't have permissions to in DMs.

TizzySaurus:
Closes #2081.

Catches error raised when trying to edit message in DM fails (we still try to edit so we can ensure the message hasn't been deleted).

If the error was raised by a message inside the guild, then we reraise the error since that's unexpected behaviour (would mean that the bot has missing permissions).

Also only checks if we need to send to #bot-commands when the message was sent in the guild.

*NB: Currently untested.*

### DIFF
--- a/bot/exts/info/code_snippets.py
+++ b/bot/exts/info/code_snippets.py
@@ -262,7 +262,7 @@ class CodeSnippets(Cog):
                     raise e
 
             # If we're in a guild, then check if we need to redirect to #bot-commands
-            if destination.guild:
+            if message.guild:
                 if len(message_to_send) > 1000 and message.channel.id != Channels.bot_commands:
                     # Redirects to #bot-commands if the snippet contents are too long
                     await self.bot.wait_until_guild_available()
@@ -273,10 +273,12 @@ class CodeSnippets(Cog):
                         f'Please see {destination.mention} for the full snippet.'
                     )
 
-            await wait_for_deletion(
-                await destination.send(message_to_send),
-                (message.author.id,)
-            )
+                await wait_for_deletion(
+                    await destination.send(message_to_send),
+                    (message.author.id,)
+                )
+            else:
+                await destination.send(message_to_send)
 
 
 def setup(bot: Bot) -> None:
--- a/bot/exts/info/code_snippets.py
+++ b/bot/exts/info/code_snippets.py
@@ -255,16 +255,23 @@ class CodeSnippets(Cog):
             except discord.NotFound:
                 # Don't send snippets if the original message was deleted.
                 return
+            except discord.Forbidden as e:
+                # We still want to send snippets when in DMs, but if we're in guild then
+                # reraise error since that means there's a permissions issue with the bot.
+                if message.guild:
+                    raise e
 
-            if len(message_to_send) > 1000 and message.channel.id != Channels.bot_commands:
-                # Redirects to #bot-commands if the snippet contents are too long
-                await self.bot.wait_until_guild_available()
-                destination = self.bot.get_channel(Channels.bot_commands)
+            # If we're in a guild, then check if we need to redirect to #bot-commands
+            if destination.guild:
+                if len(message_to_send) > 1000 and message.channel.id != Channels.bot_commands:
+                    # Redirects to #bot-commands if the snippet contents are too long
+                    await self.bot.wait_until_guild_available()
+                    destination = self.bot.get_channel(Channels.bot_commands)
 
-                await message.channel.send(
-                    'The snippet you tried to send was too long. '
-                    f'Please see {destination.mention} for the full snippet.'
-                )
+                    await message.channel.send(
+                        'The snippet you tried to send was too long. '
+                        f'Please see {destination.mention} for the full snippet.'
+                    )
 
             await wait_for_deletion(
                 await destination.send(message_to_send),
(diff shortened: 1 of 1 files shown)


bot/exts/info/code_snippets.py
```diff
@@ -255,16 +255,23 @@ async def on_message(self, message: discord.Message) -> None:
...
-                    f'Please see {destination.mention} for the full snippet.'
-                )
+            except discord.Forbidden as e:
+                # We still want to send snippets when in DMs, but if we're in guild then
+                # reraise error since that means there's a permissions issue with the bot.
+                if message.guild:
+                    raise e
+
+            # If we're in a guild, then check if we need to redirect to #bot-commands
+            if destination.guild:
```
ToxicKidz, line 265: [`DMChannel`](https://discordpy.readthedocs.io/en/stable/api.html?highlight=audit_logs#discord.DMChannel) has no `guild` attribute.
```suggestion
            if message.guild:
```

ToxicKidz (changes requested):
There's one line that causes an error, but other than that this looks good.

bot/exts/info/code_snippets.py
```diff
@@ -255,16 +255,23 @@ async def on_message(self, message: discord.Message) -> None:
...
+                    # Redirects to #bot-commands if the snippet contents are too long
+                    await self.bot.wait_until_guild_available()
+                    destination = self.bot.get_channel(Channels.bot_commands)
+
+                    await message.channel.send(
+                        'The snippet you tried to send was too long. '
+                        f'Please see {destination.mention} for the full snippet.'
+                    )
 
             await wait_for_deletion(
```
ToxicKidz, line 276: Only saw this after I pressed the submit review button, but `wait_for_deletion()` only accepts messages that are in guilds. Although, I don't think we should really support code snippets in DMs anyways.

bot/exts/info/code_snippets.py
TizzySaurus: Good catch :+1:

Have fixed.

bot/exts/info/code_snippets.py
TizzySaurus: > I don't think we should really support code snippets in DMs anyways.

I think it's nice to support as many features in DMs as we can. It's relatively easy to support code snippets in DMs, so I think we should.

> `wait_for_deletion()` only accepts messages that are in guilds.

Thanks, have fixed :+1:

HassanAbouelela: > I think it's nice to support as many features in DMs as we can. It's relatively easy to support code snippets in DMs, so I think we should.

I disagree with this, and I believe that’s the general view of the core dev team as well. There is a very, very clear cost (this entire PR) to supporting these features in DMs, which doesn’t seem to balance against the benefits. You could always just… run this feature in bot-spam.

TizzySaurus: Eh, I guess.

I've fixed this now though so we might as well merge it 🤷.

ChrisLovering: I agree with Scale here, we don't want to support things in DMs with the bot. I think the correct solution to this sentry issue is to just add the below conditional to the top of the event listener.
```py
if not message.guild:
    return
```

Bluenix2 (approved):
I do have one comment you can implement if there comes other feedback, it's not a big deal overall. I can't manage to trigger the codepath that sends the ouput to bot-commands, but this fixes the issue listed.

Thank you for this Tizzy!

bot/exts/info/code_snippets.py
```diff
@@ -255,21 +255,30 @@ async def on_message(self, message: discord.Message) -> None:
             except discord.NotFound:
                 # Don't send snippets if the original message was deleted.
                 return
+            except discord.Forbidden as e:
+                # We still want to send snippets when in DMs, but if we're in guild then
+                # reraise error since that means there's a permissions issue with the bot.
+                if message.guild:
+                    raise e
```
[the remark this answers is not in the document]
ToxicKidz: This part was removed as of https://github.com/python-discord/bot/pull/2086/commits/95106690fa32d1cc5578e018ee7c75857d41d8a5, as it removed support of code snippets in DMs and the message will always be sent in a guild.

bot/exts/info/code_snippets.py
ToxicKidz: Done in https://github.com/python-discord/bot/pull/2086/commits/95106690fa32d1cc5578e018ee7c75857d41d8a5.

ChrisLovering (approved):

PR #2087, opened 2022-02-14
Allow duckifying of messages in threads

### LINKED ISSUE #1930: Duckifying doesn't work for messages inside threads.
As the title says.

An example can be found on an internal staff thread [here](https://discord.com/channels/267624335836053506/902258820561137754/902259286527311913). For those that can't view the message, it has the required amount of duck reactions but the bot didn't "duckify" it.

The issue *may* be these lines ([L168-L170](https://github.com/python-discord/bot/blob/main/bot/exts/fun/duck_pond.py#L168-L170)) but can't be sure without further investigation:
```py
        channel = discord.utils.get(self.bot.get_all_channels(), id=payload.channel_id)
        if channel is None:
            return
```

TizzySaurus:
Closes #1930.

Messages in threads can now be duckified!

*NB: Currently untested.*

### DIFF
--- a/bot/exts/fun/duck_pond.py
+++ b/bot/exts/fun/duck_pond.py
@@ -2,7 +2,7 @@ import asyncio
 from typing import Union
 
 import discord
-from discord import Color, Embed, Message, RawReactionActionEvent, TextChannel, errors
+from discord import Color, Embed, Message, RawReactionActionEvent, TextChannel, Thread, errors
 from discord.ext.commands import Cog, Context, command
 
 from bot import constants
@@ -47,14 +47,18 @@ class DuckPond(Cog):
         return False
 
     @staticmethod
-    def is_helper_viewable(channel: TextChannel) -> bool:
+    def is_helper_viewable(channel: Union[TextChannel, Thread]) -> bool:
         """Check if helpers can view a specific channel."""
         guild = channel.guild
         helper_role = guild.get_role(constants.Roles.helpers)
         # check channel overwrites for both the Helper role and @everyone and
         # return True for channels that they have permissions to view.
-        helper_overwrites = channel.overwrites_for(helper_role)
-        default_overwrites = channel.overwrites_for(guild.default_role)
+        if isinstance(channel, Thread):
+            helper_overwrites = channel.permissions_for(helper_role)
+            default_overwrites = channel.permissions_for(guild.default_role)
+        else:
+            helper_overwrites = channel.overwrites_for(helper_role)
+            default_overwrites = channel.overwrites_for(guild.default_role)
         return default_overwrites.view_channel is None or helper_overwrites.view_channel is True
 
     async def has_green_checkmark(self, message: Message) -> bool:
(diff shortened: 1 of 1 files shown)


bot/exts/fun/duck_pond.py
```diff
@@ -165,7 +169,9 @@ async def on_raw_reaction_add(self, payload: RawReactionActionEvent) -> None:
         if not self._payload_has_duckpond_emoji(payload.emoji):
             return
 
-        channel = discord.utils.get(self.bot.get_all_channels(), id=payload.channel_id)
+        await self.bot.wait_until_guild_available()
+        guild = self.bot.get_guild(payload.guild_id)
+        channel = discord.utils.get(guild.text_channels + guild.threads, id=payload.channel_id)
```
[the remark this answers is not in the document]
TizzySaurus: I wasn't aware that this was a thing.

Although looking at the docs, that would also get voice channels etc., not just TextChannel or Thread. Is this something that matters?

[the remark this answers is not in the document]
ToxicKidz: Well, I wouldn't worry about getting reactions inside a voice channel.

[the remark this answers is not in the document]
TizzySaurus: Surely it would break stuff. E.g. on L183 we do `channel.fetch_message`, which would raise an error for a VoiceChannel (voice channels don't have such a method)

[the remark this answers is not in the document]
TizzySaurus: I'll update this to use `Guild.get_channel_or_thread`

bot/exts/fun/duck_pond.py
```diff
@@ -47,14 +47,18 @@ def is_staff(member: MemberOrUser) -> bool:
         return False
 
     @staticmethod
-    def is_helper_viewable(channel: TextChannel) -> bool:
+    def is_helper_viewable(channel: Union[TextChannel, Thread]) -> bool:
```
ChrisLovering, line 50: `permissions_for` works on all `GuildChannel` objects, and also takes into account default role permissions, so this whole function can be simplified quite a lot. The `channel` type hint can also be changed to `discord.GuildChannel`, since this funciton is available on all GuildChannel objects.

```py
def is_helper_viewable(channel: discord.GuildChannel) -> bool:
"""Return whether helpers can view the given channel."""
    helper_role = guild.get_role(constants.Roles.helpers)
    return channel.permissions_for(helper_role).view_channel
```

ChrisLovering (changes requested):

bot/exts/fun/duck_pond.py
ToxicKidz: Although, the only channel you can get reactions from is `TextChannel` and `Thread` (also `PrivateChannel`, but we do a check for the guild), so the current typehint is also correct.

ChrisLovering: Yea, that's true. Although looking at the file, this function is only actually used once. So imo this function can be deleted and [this condition](https://github.com/python-discord/bot/blob/main/bot/exts/fun/duck_pond.py#L172-L174) can just be changed to
```py
helper_role = guild.get_role(constants.Roles.helpers)
if not channel.permissions_for(helper_role).view_channel:
    return
```

Bluenix2: @TizzySaurus I am looking through my notifications and found this PR, are you looking to implement the feedback Chris mentioned here?

ToxicKidz: This is resolved in https://github.com/python-discord/bot/pull/2087/commits/532ce7d7785103c8c67341d8d24d5301f76a8257.

Bluenix2 (approved):
Thank you for this fix Tizzy. Managed to fix a bug and simultaneously reduce code complexity - perfect!

PR #2089, opened 2022-02-15
Feat: add reminder snoozing & show new expiry / mention count when editing reminder

### LINKED ISSUE #2045: Reschedule reminders
Quite often users setup a reminder, and when it gets to that time set another reminder for the same reason. This is usually because the reminder is to get them to do some task, and the user doesn't have time to do that task, so delays it.

We should support this workflow natively within the reminders, rather than having the user create a new reminder.

I propose that when the bot sends the notification for the reminder, the reminder includes some buttons that allows the user to re-schedule the reminder for some set periods. Maybe , 1 hour, 1 day, 1 week and the same period that the original remind was set for.

An example for the last one would be that I set a reminder for 4 hours, so that option would be for another 4 hours.

When the "snoozed" reminder comes up, the bot should link to the original message.

### LINKED ISSUE #2172: Show new timestamp in confirmation embed after editing reminder duration
image

When setting a reminder with the `!remind` command, the embed contains a timestamp which states the time at which the reminder will arrive. Same should be done for the embed you get when you edit a reminder, i.e, a timestamp of the newly edited time should be added to the embed.

Anyone can implement.

TizzySaurus:
Closes #2045.
Closes #2172.

Users can now "snooze" reminders, by pressing the "Snooze Reminder" button and then selecting a snooze duration from the dropdown menu that appears upon doing so.

I've tried to make this as resilient as I can, and have extensively tested it so there should hopefully not be any bugs 🤞. 

Tested the "that's not your reminder to snooze" with help of @LiquidPulsar (thanks!). 

This is my first attempt at using dpy2.0 features, so please feel free to give suggestions as to any possible improvements as well as reviewing the code as-is.

"Snooze Reminder" button:
image

After pressing the button:
image

After selecting snooze duration (both embeds in one message):
image


bot/exts/utils/reminders.py
```diff
@@ -30,6 +32,133 @@
...
+        original_embed = interaction.message.embeds[0]
+        new_embed = discord.Embed(
+            colour=discord.Colour.green(),
+            title=random.choice(POSITIVE_REPLIES),
+            description=f"Successfully snoozed reminder for {selected_duration}. New remind time: {new_end_timestamp}",
+        )
+        new_embed.set_footer(text=f"ID: {self.reminder_id}")
+        try:
+            await interaction.response.edit_message(embeds=[original_embed, new_embed], view=None)
+        except discord.NotFound:
```
onerandomusername, line 110: what exactly is the reason for this?

bot/exts/utils/reminders.py
```diff
@@ -198,17 +338,41 @@ async def send_reminder(self, reminder: dict, expected_time: t.Optional[time.Tim
         jump_url = reminder.get("jump_url")
         embed.description += f"\n[Jump back to when you created the reminder]({jump_url})"
         partial_message = channel.get_partial_message(int(jump_url.split("/")[-1]))
```
onerandomusername, line 340: while we're here, `rsplit("/", 1)[-1]`

bot/exts/utils/reminders.py
```diff
@@ -30,6 +32,133 @@
...
+    def __init__(self, bot: Bot, reminder_id: int, reminder_author_id: int):
+        super().__init__(timeout=30)
+        self.new_expiry: t.Optional[datetime] = None
+        self.reminder_was_snoozed = False
+
+        self.bot = bot
+        self.reminder_id = reminder_id
+        self.interaction_owner_id = reminder_author_id
+
+        self.dropdown: discord.ui.Select = self.children[0]
```
onerandomusername, line 68: Why are we doing this? The defined select can be accessed as `self.select_snooze_duration`.

bot/exts/utils/reminders.py
```diff
@@ -30,6 +32,133 @@
...
+        for duration in self.SNOOZE_DURATIONS:
+            self.dropdown.add_option(label=duration)
```
onerandomusername, line 70: We can add a value here which will be sent by the api. Rather than having to map them all up on receiving them, we can send both what to display and what we want to see to the api.

onerandomusername (commented):
basic overlook

bot/exts/utils/reminders.py
TizzySaurus: I did notice this, but I also don't see a point in doing so. We only ever use the label, so there's not much point in defining a value (especially since not all values are fixed -- for durations >= 1 month).

bot/exts/utils/reminders.py
TizzySaurus: > Why are we doing this? The defined select can be accessed as `self.select_snooze_duration`.

This was simply copied from the code on lancebot:
https://github.com/python-discord/sir-lancebot/blob/main/bot/exts/utilities/epoch.py#L115

bot/exts/utils/reminders.py
TizzySaurus: I believe this may be a left-over try/except from when I was trying to debug a separate error.

The idea is that the message may be deleted before we edit it, but in reality I don't think this would happen.

bot/exts/utils/reminders.py
```diff
@@ -30,6 +32,133 @@
...
+        if interaction.user.id != self.interaction_owner_id:
+            await interaction.response.send_message(":x: This is not your reminder to snooze!", ephemeral=True)
+            return False
+        return True
+
+    @discord.ui.select(placeholder="Select the snooze duration")
+    async def select_snooze_duration(self, _: discord.ui.Select, interaction: discord.Interaction) -> None:
+        """Drop down menu which contains a list of snooze durations one can choose."""
+        selected_duration = interaction.data["values"][0]
+        current_datetime = datetime.now()
```
Bluenix2, line 83: I believe this is naive, you should most likely supply it with a tomezone.

bot/exts/utils/reminders.py
```diff
@@ -30,6 +32,133 @@
...
+        if interaction.user.id != self.parent_view.reminder_author_id:
+            await interaction.response.send_message(":x: This is not your reminder to snooze!", ephemeral=True)
+            return
+
+        snooze_select_view = SnoozeSelectView(
+            bot=self.parent_view.bot,
+            reminder_id=self.parent_view.reminder_id,
+            reminder_author_id=self.parent_view.reminder_author_id
+        )
+        await interaction.response.edit_message(view=snooze_select_view)
```
Bluenix2, line 151: Wouldn't it be easier for this flow to continue as an ephemeral response (this creates a new ephemeral response, which the select menu and that in turn creates a new ephemeral response/edits the ephemeral response)?

Is there perhsps concern among the moderators that it should be clearer to them when a reminder has been snoozed?

bot/exts/utils/reminders.py
```diff
@@ -30,6 +32,133 @@
...
+
+        await snooze_select_view.wait()
+
+        if snooze_select_view.reminder_was_snoozed:
```
Bluenix2, line 155: You can use the return value of `wait()` instead! It returns a boolean which indicates whether the view timed out or was stopped (I don't remember which case had which boolean value though).

bot/exts/utils/reminders.py
```diff
@@ -198,17 +338,41 @@ async def send_reminder(self, reminder: dict, expected_time: t.Optional[time.Tim
...
+            m = await channel.send(
+                content=f"<@{reminder_author_id}> {additional_mentions}",
+                embed=embed,
+                view=snooze_button_view
+            )
+
+        # Before deleting the reminder, see if the user wants to "snooze" it
+        await snooze_button_view.wait()
+
+        if not snooze_button_view.reminder_was_snoozed:
```
Bluenix2, line 364: My comment above applies to this as well. We can use the return value of `wait()`.

bot/exts/utils/reminders.py
ToxicKidz: Is there a point for this? We aren't passing this to dpy methods at all.

Bluenix2: Even if we don't pass it discord.py, if we make it a habit of always using aware datetime objects everywhere then we'll never have issues with aware/naive datetimes anywhere. I think it's a good habit to maintain.

bot/exts/utils/reminders.py
```diff
@@ -30,6 +32,133 @@
...
+        super().__init__(timeout=300)
+        self.new_expiry: t.Optional[datetime] = None
+        self.reminder_was_snoozed = False
+
+        self.bot = bot
+        self.reminder_id = reminder_id
+        self.reminder_author_id = reminder_author_id
+
+        snooze_button = SnoozeButton(self)
+        self.add_item(snooze_button)
```
JacobMonck, line 130: Just a minor nit pick, why not have
```py
snooze_button = SnoozeButton(self)
self.add_item(snooze_button)
```
as
```py
self.add_item(SnoozeButton(self))
```
just seems like an unnecessary variable assignment but please let me know if there's a reason for this.

TizzySaurus: Yep, agreed!

bot/exts/utils/reminders.py
TizzySaurus: Yep! Have updated :+1:

bot/exts/utils/reminders.py
TizzySaurus: I don't see any potential moderation issues, and Zig is okay either way so I'll go ahead and make this ephemeral :+1:

TizzySaurus: Would I have to create a new view to achieve this?

bot/exts/utils/reminders.py
TizzySaurus: Yep, I've fixed this by doing `datetime.now(timezone.utc)`, as per [the docs'](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) recommended way.

bot/exts/utils/reminders.py
TizzySaurus: Upon further inspection, I don't see this? `self.select_snooze_duration` is a method, so doesn't give us the Select menu.

bot/exts/utils/reminders.py
TizzySaurus: Unless anyone else has any issues with this, I think I'll just leave it. From what I can tell there's no harm in having this safety net.

bot/exts/utils/reminders.py
TizzySaurus: Managed to figure it out with @D0rs4n's help. This has been added :+1:

bot/exts/utils/reminders.py
onerandomusername: It is decorated with the `discord.ui.select` decorator which converts it into a Select.

onerandomusername: @TizzySaurus I think you accidental resolved this without fixing it.

TizzySaurus: > It is decorated with the `discord.ui.select` decorator which converts it into a Select.

Right, yes, I missed that. Have fixed :+1:

Xithrius:
Jeez the review interface just completely broke on me, sorry about that.

Xithrius:
Alright whatever I'm doing this manually, just in case this actually didn't work.

@JacobMonck @vivekashok1221 @ToxicKidz @Bluenix2 @onerandomusername please re-review.

Thanks!

henri-climber:
It's been a while since this discussion was active but i really like the idea. 
I just tested everything but for me there seems to be a problem when snoozing a timer for the second time. I get the option to snooze again and when i click it I can also select for how long i want to snooze but then it says "This interaction failed". Did this happen to any of you as well?

TizzySaurus:
> It's been a while since this discussion was active but i really like the idea. I just tested everything but for me there seems to be a problem when snoozing a timer for the second time. I get the option to snooze again and when i click it I can also select for how long i want to snooze but then it says "This interaction failed". Did this happen to any of you as well?
> 
> EDIT: The problem seems to be that once the first reminder is finished and you click snooze the second reminder is automatically deleted.
> When pressing snooze for the first time it deletes the Reminder after 30s.

I'm not able to reproduce this. I get [the trace log from pydis_core](https://github.com/python-discord/bot-core/blob/main/pydis_core/utils/scheduling.py#L193) saying the reminder is being deleted, but it still shows when I use the `remind list` command (so hasn't actually been deleted from the site, only the internal pydis_core tasks) and I can successfully re-snooze (and keep re-snoozing).

> 2023-02-28 18:47:50 | bot.exts.utils.reminders | DEBUG | Deleting reminder #11 (432215309278117890 has been reminded and didn't snooze).

I don't get this log message from the bot code, so don't really know what's going on there. I'll try and look further into it but am not particularly hopeful.

bot/exts/utils/reminders.py
```diff
@@ -35,6 +37,149 @@
...
+            await interaction.response.edit_message(embed=new_embed, view=None)
+        except discord.NotFound:
+            await interaction.message.channel.send(embed=new_embed)
+
+
+class SnoozeButtonView(discord.ui.View):
+    """The reminder's snooze button UI View."""
+
+    def __init__(self, reminders_instance: "Reminders", reminder_dct: dict):
+        super().__init__(timeout=30)  # user has 5 minutes from reminder to snooze it
```
TizzySaurus, line 131: Adding this comment as a reminder to myself to fix this comment/timeout as is appropriate before this gets merged.

bot/exts/utils/reminders.py
```diff
@@ -35,6 +37,149 @@
...
+        "1 day": 24 * 60 * 60,
+        "1 week": 7 * 24 * 60 * 60,
+        "1 month": None,
+        "3 months": None,
+        "6 months": None,
+        "1 year": None
+    }
+
+    def __init__(self, reminders_instance: "Reminders", reminder_dct: dict):
+        super().__init__(timeout=30)
```
TizzySaurus, line 65: The exact appropriate timeout for the drop-down menu is still up-for-debate.

bot/exts/utils/reminders.py
```diff
@@ -35,6 +37,149 @@
 ReminderMention = t.Union[UnambiguousUser, discord.Role]
 
 
+class SnoozeSelectView(discord.ui.View):
+    """The reminder's select dropdown UI View."""
+
+    SNOOZE_DURATIONS: dict[str, t.Optional[int]] = {
```
TizzySaurus, line 43: Exact snooze durations still up-for-debate.

TizzySaurus (commented):
A few comments for consideration before this PR is merged.

Xithrius:

PR #2090, opened 2022-02-15
Fix: `!raw` can now be used in threads

### LINKED ISSUE #2077: !raw command does not work in threads
The `raw` command does not work in threads. It displays the following message even when the user has access to the thread:
image.
I haven't looked very closely into it and I'm rusty with discord.py but the bug could be because `message.channel.members` [[line 473](https://github.com/python-discord/bot/blob/ee5a2166e08805a31dd6141db663ccacfde0b551/bot/exts/info/information.py#L473)] is a list of [`ThreadMember`](https://discordpy.readthedocs.io/en/master/api.html?highlight=channel#discord.ThreadMember)s when the command is invoked in a thread.

TizzySaurus:
Closes #2077.

`!raw` now works for messages inside threads.

Tested w/ help of @LiquidPulsar.

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -470,7 +470,7 @@ class Information(Cog):
 
         If `json` is True, send the information in a copy-pasteable Python format.
         """
-        if ctx.author not in message.channel.members:
+        if not message.channel.permissions_for(ctx.author).read_messages:
             await ctx.send(":x: You do not have permissions to see the channel this message is in.")
             return
 


TizzySaurus:
> `read_message_history` is probably a better option for this check.

Decided I'll just leave it as-is, since I'm pretty sure it won't make a difference within pydis anyway. If someone else thinks I should change this then I will.

Xithrius (approved):

ChrisLovering (approved):

PR #2092, opened 2022-02-16
Remove unnecessary Infraction conversion in clean ban

ChrisLovering:
apply_ban already returns the data exactly how it's needed, so we don't need to call the infraction convertor and use an API call to site.


jchristgit (approved):
thanks

Xithrius:
You didn't lint before you pushed, you must now perish.

Xithrius (approved):
Once test is fixed then we're good to go

PR #2098, opened 2022-02-18
Add user ID in message content for mod-alerts, but not for autobans

### LINKED ISSUE #2088: Show user ID outside embed for `#mod-alerts`
Many moderators moderate from iOS and Android mobile devices. As of today, Discord iOS doesn't allow for copying embed text *at all*, and Android iOS allows copying of embed text as a whole (not selectively).

One of the biggest pain points for moderators is not being able to grab the user's ID from the mod-alert embed, neither from the embed itself nor from the mention of the user (because the mention doesn't resolve properly).

A quick easy fix for this is to simply add the user ID outside the embed as plain text so that moderators can copy/paste that text and research and infract the user easily. There are other proposed ideas such as buttons and such, but that can all be planned later, whereas this is a very quick easy change with a lot of immediate benefit.

I would like the user ID to be added for all mod-alerts, even non-pinging ones (such as for those sent to bot via DM)

image

minalike:
Continued work on #2096 
Further closes #2088 

This PR 
- additionally adds the user ID for antispam alerts
- additionally adds the user ID for username filtering alerts
- removes the user ID for autoban alerts

I'll make an additional commit to rename the `content` in the code block below, since it overlaps with the `content` parameter used a little later on.

I didn't quite understand what lines 98-101 were doing though, and need some guidance on good rename suggestions for `content` 😄 
https://github.com/python-discord/bot/blob/389cb815a379feca1122df12b3492144b76e7749/bot/exts/filters/antispam.py#L93-L101

### DIFF
--- a/bot/exts/filters/antispam.py
+++ b/bot/exts/filters/antispam.py
@@ -103,6 +103,7 @@ class DeletionContext:
             mod_alert_message += content
 
         await modlog.send_log_message(
+            content=", ".join(str(m.id) for m in self.members),  # quality-of-life improvement for mobile moderators
             icon_url=Icons.filtering,
             colour=Colour(Colours.soft_red),
             title="Spam detected!",
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -256,6 +256,7 @@ class Filtering(Cog):
             )
 
             await self.mod_log.send_log_message(
+                content=str(member.id),  # quality-of-life improvement for mobile moderators
                 icon_url=Icons.token_removed,
                 colour=Colours.soft_red,
                 title="Username filtering alert",
@@ -423,9 +424,12 @@ class Filtering(Cog):
             # Allow specific filters to override ping_everyone
             ping_everyone = Filter.ping_everyone and _filter.get("ping_everyone", True)
 
-        # If we are going to autoban, we don't want to ping
+        content = str(msg.author.id)  # quality-of-life improvement for mobile moderators
+
+        # If we are going to autoban, we don't want to ping and don't need the user ID
         if reason and "[autoban]" in reason:
             ping_everyone = False
+            content = None
 
         eval_msg = "using !eval " if is_eval else ""
         footer = f"Reason: {reason}" if reason else None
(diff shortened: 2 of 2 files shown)


minalike:
Will skip the additional commit and my curiosity about the above code is satisfied. Please review!

bot/exts/filters/antispam.py
```diff
@@ -103,6 +103,7 @@ async def upload_messages(self, actor_id: int, modlog: ModLog) -> None:
             mod_alert_message += content
 
         await modlog.send_log_message(
+            content=", ".join(str(m.id) for m in self.members),  # quality-of-life improvement for mobile moderators
```
ChrisLovering, line 106: Make this a set, so the same user_id isn't sent more than once.

```suggestion
            content=", ".join({str(m.id) for m in self.members}),  # quality-of-life improvement for mobile moderators
```

bot/exts/filters/filtering.py
```diff
@@ -423,9 +424,12 @@ async def _send_log(
             # Allow specific filters to override ping_everyone
             ping_everyone = Filter.ping_everyone and _filter.get("ping_everyone", True)
 
-        # If we are going to autoban, we don't want to ping
+        content = str(msg.author.id)  # quality-of-life improvement for mobile moderators
+
+        # If we are going to autoban, we don't want to ping and don't need the user ID
         if reason and "[autoban]" in reason:
             ping_everyone = False
+            content = ""
```
ChrisLovering, line 432: To be explicit.
```suggestion
            content = None
```

ChrisLovering (changes requested):

bot/exts/filters/antispam.py
ChrisLovering: Actually, members is already a frozenset, so this can be ignored.

Akarys42 (approved):

bot/exts/filters/filtering.py
minalike: https://github.com/python-discord/bot/pull/2098/commits/6c561e251be3a4af2af349ba4b86b9ff0f0a9b7b

ChrisLovering (approved):

vivekashok1221 (approved):
lgtm.

PR #2103, opened 2022-02-25
Disnake migration

ChrisLovering:
This migrates the bot to disnake, as per recent discussions.

What this includes:
- Changes all references from discord.py to disnake, including in tests
- Bump version of bot-core so discord.py wasn't required as a sub dep
- Use monkey patches from bot-core
- Fix the single breaking change I could find
    - The internal bot.http.edit_message function now has a required `files` kwarg, which is only used in the help channels cog.
    - Set this to None as we never upload files using that message

This not included in this PR:
- Updating the tags to reference disnake. These were left with the discord module, and can be updated later if needed
- Migrating the bot to use all utils that have been moved to bot-core
    - This would inflate this PR even further, so didn't want to pull it into scope.

P.S this is quite a large PR, so reviewing by the commits may be helpful. I tried to keep the huge rename within the single commit https://github.com/python-discord/bot/pull/2103/commits/27b18baaf785091e4e2d5dcdca3ff18981dfcd44, and didn't include anything else in there.


bot/__init__.py
```diff
@@ -16,16 +15,7 @@
...
-# as library objects are made by this mapping.
-# https://github.com/Rapptz/discord.py/blob/1a4e73d59932cdbe7bf2c281f25e32529fc7ae1f/discord/ext/commands/converter.py#L984-L1004
-commands.converter.PartialMessageConverter = monkey_patches.FixedPartialMessageConverter
-
-# Monkey-patch discord.py decorators to use the Command subclass which supports root aliases.
-# Must be patched before any cogs are added.
-commands.command = partial(commands.command, cls=monkey_patches.Command)
-commands.GroupMixin.command = partialmethod(commands.GroupMixin.command, cls=monkey_patches.Command)
+# Apply all monkey patches from bot core.
+monkey_patches.apply_monkey_patches()
```
MarkKoz, line 19: This API is kind of silly. If there's only going to be one public function, it would have been better to expose it directly from the utils namespace. You stated yourself that we'd probably never want to apply patches on an individual basis; if patches are no longer needed we'd just remove them entirely from the code base.

pyproject.toml
```diff
@@ -9,7 +9,7 @@ license = "MIT"
 python = "3.9.*"
 disnake = "~=2.4"
 # See https://bot-core.pythondiscord.com/ for docs.
-bot-core = {url = "https://github.com/python-discord/bot-core/archive/511bcba1b0196cd498c707a525ea56921bd971db.zip"}
+bot-core = {url = "https://github.com/python-discord/bot-core/archive/5593a0c95e2b4fd9515c8875ea6e51c0ab88eb00.zip"}
```
MarkKoz, line 12: We basically do versioning already for the change log. Do we plan to use git tags? Then I think this URL could use the tag, which would be clearer than a hash.

Bluenix2:
Just as point of order / reminder, please do not force push in this PR as that will mess with the "Viewed" markers to the right.

Bluenix2:
Good work on this by the way, I ran both Pyright and Mypy in the command line with most type checking disabled (or ignored when reading the output) and I could not find any complaints about `discord` or missing imports (although Mypy failed to find the typings for `disnake.ext.commands` which may make this test incomplete).

Bluenix2 (changes requested):
I found these three references to discord.py still, they should probably be renamed as well:
https://github.com/python-discord/bot/blob/91621bc7ef590118e91bbb2e387db5e14128b530/bot/utils/function.py#L106
https://github.com/python-discord/bot/blob/91621bc7ef590118e91bbb2e387db5e14128b530/bot/utils/function.py#L97
https://github.com/python-discord/bot/blob/91621bc7ef590118e91bbb2e387db5e14128b530/bot/utils/function.py#L139

pyproject.toml
ChrisLovering: Yea, good idea. https://github.com/python-discord/bot/pull/2103/commits/3a03e83776db009494f189183189615d616fdb59

bot/__init__.py
Bluenix2: @MarkKoz I think you should open an issue in the bot-core repository then. I agree with you and would like to see this changed, but at the moment this is not something Chris can change in this PR other than doing `from botcore.utils.monkey_patches import apply_monkey_patches`.

Bluenix2:
Thanks for addressing my comments Chris. I have looked at the diff of each file and not yet found any odd changes - I could also not find any differences between discord.py and disnake when migrating **other than the new text-in-voice feature resulting in some methods returning `VoiceChannel`** which may not include all attributes/methods (`send()` and `edit()` would be present I believe). Any issues that may arise from this that come to mind?

Additionally, this pull request is pretty much incompatible with every other pull request currently open. Has there been any thought put into figuring out which ones should be merged before this is merged, and which ones should be merged afterwards (so that we can let those know)?

bot/__init__.py
ChrisLovering: Raised https://github.com/python-discord/bot-core/pull/35 for this comment, good point Mark

ChrisLovering:
>[ ... ] `VoiceChannel`** which may not include all attributes/methods (`send()` and `edit()` would be present I believe). Any issues that may arise from this that come to mind?
Not that I can think of, we also don't use text in voice channels in the server.

> Additionally, this pull request is pretty much incompatible with every other pull request currently open. Has there been any thought put into figuring out which ones should be merged before this is merged, and which ones should be merged afterwards (so that we can let the author of those know to update to disnake)?

I think any PRs that are open when this is merged will just need to be rebased, I don't think there are any PRs open that should block this.

We of course can extend the offer to help anyone with an open PR to resolve conflicts.

ChrisLovering:
Marked as do not merge to allow me to squash fixup commits on this PR before it hits main

bot/__init__.py
ChrisLovering: Fixed in https://github.com/python-discord/bot/pull/2103/commits/9e6989aee80ee21a2f3defa94d9eca36834fa771

bot/utils/checks.py
```diff
@@ -135,7 +135,7 @@ async def predicate(cog: Cog, ctx: Context) -> None:
         if any(role.id in bypass for role in ctx.author.roles):
             return
 
-        # cooldown logic, taken from discord.py internals
+        # cooldown logic, taken from disnake's, internals
```
MarkKoz, line 138: Reviewing every line paid off
```suggestion
        # cooldown logic, taken from disnake's internals
```

MarkKoz (changes requested):
I've looked over all the diffs; looks fine except that one comment. I ran the bot and it started up fine. I did my own search for missed references and couldn't find anything.

bot/utils/checks.py
ChrisLovering: hah, nice spot https://github.com/python-discord/bot/pull/2103/commits/445c4a533eb11e9f9c8d8b438c858f7425775d50

MarkKoz (approved):

mbaruh (approved):
Ran most commands, seems to work well. Thanks!

PR #2107, opened 2022-03-01
fix: Make the help command's button only work for author

### LINKED ISSUE #2104: Limit users interaction on mod embeds
When someone from moderation team uses ``!help sm`` non staff member or user who did not send the command is able to interact with buttons. I think it should be limited to not allow users interaction on mod help embed if they are not the author. 
The paginator wont show up when non staff member uses the `!help sm` command but I think it would make sense that if moderator runs the command in some general channel, the interactions would be limited to author of message.

Link to message: https://discordapp.com/channels/267624335836053506/267624335836053506/947082166020501524
image

an-dyy:
This PR closes #2104 (Limit users interaction on mod embeds)

### DIFF
--- a/bot/exts/info/help.py
+++ b/bot/exts/info/help.py
@@ -121,7 +121,7 @@ class CommandView(ui.View):
             self.children.append(GroupButton(help_command, command, emoji="↩️"))
 
     async def interaction_check(self, interaction: Interaction) -> bool:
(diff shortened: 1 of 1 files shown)


ToxicKidz (approved):
LGTM, thanks for the contribution!

Bluenix2 (approved):
Awesome! This works as expected, thanks for your contribution.

Issue #2076, opened 2022-02-07
Do not ignore invocations of `!voiceverify` in `#voice-verification`


minalike:
Once in a blue moon (that day was today) a user somehow gets the Voice Verified role but the event is not logged by Discord's audit log nor in our `#user-log.`

Currently, the messages in `#voice-verification` that get deleted by the bot are logged in `#message-change-log` with the exception of invocations of the `!voiceverify` command.

We would like to see the bot not ignore these messages, and still log them.

The relevant sections of code (thanks @ChrisLovering)

https://github.com/python-discord/bot/blob/ee5a2166e08805a31dd6141db663ccacfde0b551/bot/exts/moderation/voice_gate.py#L194

https://github.com/python-discord/bot/blob/ee5a2166e08805a31dd6141db663ccacfde0b551/bot/exts/moderation/voice_gate.py#L242-L243

ChrisLovering:
The further this, my opinions are that we should either remove these two mod_log.ingore calls, or we should log.info() these events, so that when investigating we have an avenue to go down that isn't trawl through Discord's audit logs.

ChrisLovering:
In the linked PR I decided to only remove the ignore for the role assignment. I didn't think there was any point in logging the deleted message too.

minalike:
Presently the bot has logic to exclude logging of both the member role update and the `!voiceverify` invocation. While removing the ignore for the role assignment does solve my issue of not knowing when a user had verified previously, I find the current behavior (prior to your PR) very odd and not intuitive. 

Why have code to ignore `!voiceverify` invocations at all? It goes against my expectations as an end user of the bot as a moderator, since we log all other deleted messages in #voice-verification

PR #2111, opened 2022-03-08
Remove modlog ignore for voice verifcation

ChrisLovering:
When a user leaves and rejoins, and then asks for the voice verification role back in modmail, we want a way to see if they have every been given the role.

By removing this modlog ignore, it allows us to check the user-log channel for user updates to see if they were ever assigned the role.

### DIFF
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -10,7 +10,7 @@ from discord.ext.commands import Cog, Context, command
 
 from bot.api import ResponseCodeError
 from bot.bot import Bot
-from bot.constants import Channels, Event, MODERATION_ROLES, Roles, VoiceGate as GateConf
+from bot.constants import Channels, MODERATION_ROLES, Roles, VoiceGate as GateConf
 from bot.decorators import has_no_roles, in_whitelist
 from bot.exts.moderation.modlog import ModLog
 from bot.log import get_logger
@@ -191,7 +191,6 @@ class VoiceGate(Cog):
                 await ctx.channel.send(ctx.author.mention, embed=embed)
             return
 
-        self.mod_log.ignore(Event.member_update, ctx.author.id)
         embed = discord.Embed(
             title="Voice gate passed",
             description="You have been granted permission to use voice channels in Python Discord.",


minalike:
As per the feedback from other mods, can you also remove the modlog ignore deleting `!voiceverify` messages?

MaskDuck (approved):
gud

minalike (changes requested):
Changing my position to block this PR until we also remove the modlog ignore for invocations of !voiceverify 

https://github.com/python-discord/bot/blob/ee5a2166e08805a31dd6141db663ccacfde0b551/bot/exts/moderation/voice_gate.py#L242-L243

My justification is that when users contact the moderation team claiming they meet the requirements but still aren't able to pass voice gate, the first thing a moderator thinks to check is the message logs to see if they've correctly attempted the command. We can infer if they did based on relaying the DM logs with the bot, but those logs are in UTC and it's not as intuitive as simply searching the user ID in the message logs, nor is it guaranteed that a user had DMs open when attempting to verify. 

There's also no good reason I can foresee for ignoring these anyway. And who doesn't love removing code?

ChrisLovering:
> Changing my position to block this PR until we also remove the modlog ignore for invocations of !voiceverify 
> 
> https://github.com/python-discord/bot/blob/ee5a2166e08805a31dd6141db663ccacfde0b551/bot/exts/moderation/voice_gate.py#L242-L243
> 
> My justification is that when users contact the moderation team claiming they meet the requirements but still aren't able to pass voice gate, the first thing a moderator thinks to check is the message logs to see if they've correctly attempted the command. We can infer if they did based on relaying the DM logs with the bot, but those logs are in UTC and it's not as intuitive as simply searching the user ID in the message logs, nor is it guaranteed that a user had DMs open when attempting to verify. 
> 
> There's also no good reason I can foresee for ignoring these anyway. And who doesn't love removing code? 

Yea I can agree with wanting to remove that,but let's not block this PR since it's still useful. I'm not able to make these code changes right now, so someone else can pickup the scope creep once this has been merged.

To that end I'll remove the issue link in the pr description so that it doesn't get closed.

minalike (approved):
So long as #2076 is not closed, then this is fine

D0rs4n (approved):

MaskDuck (changes requested):
as per the gh actions you should remove event import

PR #2114, opened 2022-03-14
Increase watch API call limit to 10,000

jb3:
By default, the site infractions query API will only return 100 records,
but we have more active watches than this meaning that the response
truncated active infractions and so the bigbrother cog was not
initialising with all active watches.

This change ensures that up to 10,000 watches are fetched instead meaning no
infractions will be truncated and missed.


jchristgit (approved):

minalike (approved):
Thanks

PR #2115, opened 2022-03-14
Pre-fetch watched users to avoid sequential API calls

ChrisLovering:
When there are many watched users, many sequential calls to  get_or_fetch_member will be made, slowing down the whole command.

Pre-fetching them with an asyncio.gather will speed this up dramatically, thanks to the power of non-blocking networking.


bot/exts/moderation/watchchannels/_watchchannel.py
```diff
@@ -353,8 +353,16 @@ async def prepare_watched_users_data(
...
+        # Pre-fetch all watched users to avoid sequential API calls
+        # Filter out None to avoid AttributeError from disnake.utils.get()
+        members = await asyncio.gather(*[
+            get_or_fetch_member(ctx.guild, user_id)
+            for user_id, _ in watched_iter
+        ])
+        members = list(filter(None, members))
         for user_id, user_data in watched_iter:
-            member = await get_or_fetch_member(ctx.guild, user_id)
+            member = disnake.utils.get(members, id=user_id)
```
mbaruh, line 365: You're already iterating over the entire list. You can build a dictionary instead and not have to iterate over all members in a loop.

mbaruh (changes requested):

bot/exts/moderation/watchchannels/_watchchannel.py
ChrisLovering: Do you mean something like this, instead of the filter?
`members = {member.id: member for member in members if member is not None}`

mbaruh: yeah

ChrisLovering: makes sense, pushed

bot/exts/moderation/watchchannels/_watchchannel.py
```diff
@@ -353,8 +353,16 @@ async def prepare_watched_users_data(
...
+        # Pre-fetch all watched users to avoid sequential API calls
+        members = await asyncio.gather(*[
+            get_or_fetch_member(ctx.guild, user_id)
+            for user_id, _ in watched_iter
+        ])
+        members = {member.id: member for member in members if member is not None}
+
         for user_id, user_data in watched_iter:
-            member = await get_or_fetch_member(ctx.guild, user_id)
+            member = members.get(user_id)
```
Bluenix2, line 365: This will not contain all `user_id`s from `watched_iter` because members that could not be found will be removed in the line above (`... for member in members if member is not None`). This means that the lines below will fail with a `TypeErro`/`AttributeError` if I understand things correctly.

We need to handle `member` being None here.

Bluenix2 (changes requested):

bot/exts/moderation/watchchannels/_watchchannel.py
```diff
@@ -353,8 +353,16 @@ async def prepare_watched_users_data(
             watched_iter = reversed(watched_iter)
 
         list_data["info"] = {}
+
+        # Pre-fetch all watched users to avoid sequential API calls
+        members = await asyncio.gather(*[
+            get_or_fetch_member(ctx.guild, user_id)
+            for user_id, _ in watched_iter
```
onerandomusername, line 360: Why is this making a possible ton of api requests? Additionally, gathering won't speed this up due to the library's implementation of ratelimit handling, internally each request is made sequentially.

onerandomusername (commented):
There's a few ways to fetch a bunch of members but this does not seem like the most optimal way. 

Consider using https://docs.disnake.dev/en/latest/api.html#disnake.Guild.query_members in order to request multiple members at once. This can fetch upto 100 users at once, which means using https://docs.disnake.dev/en/latest/api.html#disnake.utils.as_chunks in order to have lists of up to 100 users


*While these links go to disnake they are the same in dpy*

bot/exts/moderation/watchchannels/_watchchannel.py
ChrisLovering: Because there is a large majority of users who are not in the bot cache (since they're off-guild) so we need to fetch their user.

Does you suggestion of using Guild.query_members fetch users that aren't in the guild?

bot/exts/moderation/watchchannels/_watchchannel.py
ChrisLovering: This is already handled further down.

bot/exts/moderation/watchchannels/_watchchannel.py
onerandomusername: What do you mean off-guild? Does this mean they are no longer a member in the server?

ChrisLovering:
This approach won't work.

Ideally this would be fixed by the paginator lazy loading the information.

Issue #2132, opened 2022-04-17
Allow `0s` as a valid argument for `!slowmode` command


minalike:
The proper way to remove a slowmode from a channel is to use `!slowmode reset [channel]`

However, many moderators often go for `!slowmode [channel] 0s`

It would be nice to allow this, as it's quite intuitive for most mods. 

The current error:
<img width="398" alt="image" src="https://user-images.githubusercontent.com/75038675/163726420-d4c821ba-0206-4a29-81db-2f2ae98da640.png">

novialriptide:
Can I work on this issue?

D0rs4n:
> Can I work on this issue?

Absolutely, Assigned!
Good luck! If you have any questions, don't hesitate to ask!

PR #2134, opened 2022-04-18
Added an option to use 0s as an argument

novialriptide:
I'm not 100% sure if this was the solution you guys were looking for, but it works.

image

Fix: https://github.com/python-discord/bot/issues/2132


bot/exts/moderation/slowmode.py
```diff
@@ -44,16 +44,21 @@ async def get_slowmode(self, ctx: Context, channel: Optional[TextChannel]) -> No
...
             channel = ctx.channel
 
         # Convert `dateutil.relativedelta.relativedelta` to `datetime.timedelta`
         # Must do this to get the delta in a particular unit of time
-        slowmode_delay = time.relativedelta_to_timedelta(delay).total_seconds()
+        try:
+            delay = await DurationDelta().convert(ctx, delay)
+
+        except BadArgument:
+            delay = relativedelta(seconds=0)
```
mbaruh, line 59: This isn't necessarily the expected behavior. This behavior means that any invalid duration will reset the slowmode, while the behavior outlined in the issue is to just make an exception for "0s".

You could check for the value of the string, or you could make use of the converters and make the delay a union of `DuraitonDelta` and a `Literal["0s"]`

mbaruh (changes requested):

novialriptide:
image
Updated.

minalike:
Can a core dev comment/confirm: I just realized the `!defcon threshold` command supports passing just `0` (without the `s`). Could we add support for `0` to this command as well?

ChrisLovering:
For this, I would suggest instead of using literal, we do the same as what defcon has. https://github.com/python-discord/bot/blob/main/bot/exts/moderation/defcon.py#L163-L178

This Unions the DurationDelta with an int, if it's an int defcon assumed it's a number of days, for this we could assume a number of seconds and then use that

mbaruh:
> For this, I would suggest instead of using literal, we do the same as what defcon has. https://github.com/python-discord/bot/blob/main/bot/exts/moderation/defcon.py#L163-L178
> 
> This Unions the DurationDelta with an int, if it's an int defcon assumed it's a number of days, for this we could assume a number of seconds and then use that

I don't mind, but the reason for this PR is that people are trying to use "0s" specifically

ChrisLovering:
> > For this, I would suggest instead of using literal, we do the same as what defcon has. https://github.com/python-discord/bot/blob/main/bot/exts/moderation/defcon.py#L163-L178
> > This Unions the DurationDelta with an int, if it's an int defcon assumed it's a number of days, for this we could assume a number of seconds and then use that
> 
> I don't mind, but the reason for this PR is that people are trying to use "0s" specifically

Is there a specific reason that the duration delta doesn't all 0s as a valid duration? Would be nicer to solve this issue there then.

mbaruh:
@ChrisLovering durationdelta is used in all time related commands, including infractions, reminders, and clean. A 0 duration makes no sense there and will most likely break a few things.

ChrisLovering:
> @ChrisLovering durationdelta is used in all time related commands, including infractions, reminders, and clean. A 0 duration makes no sense there and will most likely break a few things.

Ah good point, this is probably best then, the alternative is to make a new convertor that accepts 0 delta durations.

bot/exts/moderation/slowmode.py
```diff
@@ -44,16 +44,19 @@ async def get_slowmode(self, ctx: Context, channel: Optional[TextChannel]) -> No
         await ctx.send(f'The slowmode delay for {channel.mention} is {humanized_delay}.')
 
     @slowmode_group.command(name='set', aliases=['s'])
-    async def set_slowmode(self, ctx: Context, channel: Optional[TextChannel], delay: DurationDelta) -> None:
+    async def set_slowmode(self, ctx: Context, channel: Optional[TextChannel],
+                           delay: Union[DurationDelta, Literal["0s", "0seconds"]]) -> None:
```
ChrisLovering, line 48:
```suggestion
    async def set_slowmode(
        self,
        ctx: Context,
        channel: Optional[TextChannel],
        delay : Union[DurationDelta, Literal["0s", "0seconds"]],
    ) -> None:
```
To be consistent with how we format other function signatures in the bot

ChrisLovering (changes requested):
Can't test the code right now, so only have a styling request.

mbaruh (approved):
Looks good, thank!

ChrisLovering (approved):

PR #2137, opened 2022-04-20
Don't try to get modlog cog on defcon cog_load

ChrisLovering:
Instead get the cog at runtime, and deal with the case where it isn't loaded

### DIFF
--- a/bot/exts/moderation/defcon.py
+++ b/bot/exts/moderation/defcon.py
@@ -71,30 +71,17 @@ class Defcon(Cog):
 
         scheduling.create_task(self._sync_settings(), event_loop=self.bot.loop)
 
-    @property
-    def mod_log(self) -> ModLog:
+    async def get_mod_log(self) -> ModLog:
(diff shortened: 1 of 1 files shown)


HassanAbouelela (approved):
I really think the solution here is to either:
1. Move stuff like this into shared util files if possible
2. or make the cog loading order deterministic so we know what we're dealing with.

We should probably do 2 regardless, so we stop running into issues where this doesn't cause problems in testing or CI, but fails in prod. Approving as a stopgap.

kosayoda (changes requested):
Works correctly during testing. In place of the cog loading check, could we also send a message to the defcon channel if the mod log cog isn't loaded by then? That would at least give us a heads up to reload defcon before logs start coming in.

wookie184 (approved):
I believe the issue we were encountering is that the timeout previously was not long enough (so it failed before modlog had had a chance to load). It's possible we might be able to work out a neater solution in the future, but I think this should solve our immediate problems.

PR #2138, opened 2022-04-20
Use the extensions list & util from bot-core

ChrisLovering:
Currently the bot is still using the extensions utils from within the bot, which isn't the actual list of extensions loaded, causing the any interaction with EXTENSIONS to be incorrect.


HassanAbouelela (approved):

wookie184:
The removed file is specified in the codeowners file, should probably be removed.

ChrisLovering:
rebased onto main and pushed code owners changes

wookie184 (approved):
LGTM

PR #2145, opened 2022-04-22
Make modpings rescheduling robust to unfilled cache

mbaruh:
Moderators might not get their pingable role back if the discord.py cache is missing members during rescheduling. This PR makes the rescheduling resolve every entry in the redis cache.

Additionally, this adds a check which will remove entries in the redis cache of former moderators.


ChrisLovering (approved):
Forgot to press submit on this one yesterday 😅

LGTM, just one comment but doesn't really matter in the big picture, since there's a few to change anyway.

wookie184 (approved):
LGTM, will merge this now and can do the botcore change some other time

PR #2148, opened 2022-04-23
Speed Up Sync Cog Loading

HassanAbouelela:
The user syncer was blocking the startup of the sync cog due to having to perform thousands of pointless member fetch requests. This speeds up that process b:
- Increasing the probability that the cache is up-to-date using `Guild.chunked`, and assuming that any found cache data is correct
- Limiting the fetches to members who were in the guild during the previous sync only, meaning we don't fetch a bunch of outdated members

### DIFF
--- a/bot/exts/backend/sync/_cog.py
+++ b/bot/exts/backend/sync/_cog.py
@@ -1,10 +1,11 @@
 import asyncio
+import datetime
 from typing import Any, Dict
 
 from botcore.site_api import ResponseCodeError
 from discord import Member, Role, User
 from discord.ext import commands
-from discord.ext.commands import Cog, Context
+from discord.ext.commands import Cog, Context, errors
 
 from bot import constants
 from bot.bot import Bot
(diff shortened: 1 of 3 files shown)


ChrisLovering (approved):
LGTM 🚀

bot/exts/backend/sync/_cog.py
```diff
@@ -27,6 +28,11 @@ async def cog_load(self) -> None:
         if guild is None:
             return
 
+        log.info("Waiting for guild to be chunked to start syncers.")
+        while not guild.chunked:
+            await asyncio.sleep(10)
+        log.info("Starting syncers.")
```
MarkKoz, line 34: Is this guaranteed to succeed? If not, there should be some time out and the syncer should still run if possible.

MarkKoz (changes requested):

bot/exts/backend/sync/_cog.py
ChrisLovering: The only time this won't succeed is if the member cache isn't populated by Discord.py, which could happen in a very rare case if Discord has gateway issues.

In that case we wouldn't want the user syncer to run at all, as we assume further down if a user isn't in the member cache, they are no longer a member.

MarkKoz (commented):
What if the member was not in the cache and `in_guild = False`. Does this mean the DB won't be able to know that they've rejoined the server (until the next sync where hopefully they are in cache)?

MarkKoz:
What kind of impact could a member being out of sync have on the rest of our features?

HassanAbouelela:
I think the big impact would primarily be on moderation tools. More specifically, the user embed would be off.

There are some other operations that would have problems if the cache was off (banning would raise an error for instance), but that is only in the case where the user is off the guild, but in the cache. I don't think that's an anticipated scenario, the opposite is more likely.

MarkKoz (approved):

bot/exts/backend/sync/_cog.py
HassanAbouelela: Added an exception and timeout in acc16545. It will now log the following after 30 minutes:
> discord.ext.commands.errors.ExtensionFailed: Extension 'Sync' raised an error: RuntimeError: The guild was not chunked in time, not loading sync cog.

PR #2151, opened 2022-04-25
Bump bot core & discord.py

ChrisLovering:
This is so that a real statsd client is actually created, along with getting the latest changes in d.py

None of the documented breaking changes affect us, and some manual testing found nothing.

Notably, one of the commits in this bump dynamically extends the timeout of Guild.chunk() based on the number or members, so it should actually work on our guild now.

This will help the cog_load behaviour of the syncer cog. I didn't want to change this as part of this PR though, to allow us to test the new chunk behaviour first

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -125,7 +125,7 @@ lxml = ["lxml"]
 
 [[package]]
 name = "bot-core"
-version = "6.1.0"
+version = "6.3.2"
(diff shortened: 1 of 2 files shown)


HassanAbouelela (approved):

HassanAbouelela (approved):
Still approved

mbaruh (approved):

PR #2152, opened 2022-04-26
Manually chunk guild if not chunked 20s after startup

ChrisLovering:


bot/exts/backend/sync/_cog.py
```diff
@@ -29,16 +29,20 @@ async def cog_load(self) -> None:
...
+        attempts = 0
+        while True:
+            attempts += 1
+            if guild.chunked:
+                log.info("Guild was found to be chunked after %d attempt(s).", attempts)
+                break
+
+            if attempts == MAX_ATTEMPTS:
+                log.info("Guild not chunked after %d attempts, calling chunk manually.", MAX_ATTEMPTS)
+                await guild.chunk()
```
HassanAbouelela, line 41: Can this fail?

ChrisLovering: The only documented failure is if the members intent isn't enabled, looking though the call stack it doesn't raise anywhere else.

It's also a websocket action, so won't get normal http errors.

HassanAbouelela: Sounds good

HassanAbouelela (approved):

wookie184:
This will actually only sleep for 20s before attempting chunking. (there are 3 checks with 10s wait between then but the first one is at 0s).

Would it also be worth logging some sort of exception if the guild doesn't end up chunked? I don't think there is a reason it shouldn't eventually end up chunked, only that it might take a long time to do so. If it not chunking itself is a common occurence we'd probably want to know about it as it would cause problems elsewhere in the bot.

Also do we know what the behaviour if it was in the process of chunking and then `guild.chunk` is called is?

ChrisLovering:
> This will actually only sleep for 20s before attempting chunking. (there are 3 checks with 10s wait between then but the first one is at 0s).

Ah good point, I'll update PR title, not a big issue though.

> Would it also be worth logging some sort of exception if the guild doesn't end up chunked? I don't think there is a reason it shouldn't eventually end up chunked, only that it might take a long time to do so

Guild.chunk() is blocking until the guild is successfully chunked. I cannot spot any errors that this would raise within the chunk nor ws impl, so I wouldn't not what exception to catch if we wanted to rethrow with a customer exception.

> Also do we know what the behaviour if it was in the process of chunking and then guild.chunk is called is?

If a chunk is in progress, the same future will be waited for, rather than start a new chunking task. https://github.com/Rapptz/discord.py/blob/master/discord/state.py#L1101-L1112

wookie184 (approved):
Thanks for explaining that! This looks good to me

PR #2153, opened 2022-04-27
Add Support For Multiple Banners

HassanAbouelela:
This PR adds support for having multiple banners which rotate in a single event, much like icons currently do. See python-discord/branding#179 for the relevant changes there. They should be merged around the same time (preferably branding first).

There are almost no code changes here as most of the required logic was already implemented, and was standardized to work for both icons and banners. A majority of the diff is renaming "icon" to "asset" and adding an asset type argument to the functions.

For testing purposes, make sure to create an API token as you'll hit the unauthorized request cap almost instantly. You can use my branding fork for testing as well, it has 2 banners under the fallback event:

`https://api.github.com/repos/HassanAbouelela/pydis-branding/contents`

### DIFF
--- a/bot/exts/backend/branding/_cog.py
+++ b/bot/exts/backend/branding/_cog.py
@@ -1,6 +1,7 @@
 import asyncio
 import contextlib
 import random
+import types
 import typing as t
 from datetime import timedelta
 from enum import Enum
@@ -120,10 +121,10 @@ class Branding(commands.Cog):
     # asset has been used in the current rotation.
     _cache_icons = RedisCache()
     _cache_banners = RedisCache()
-    asset_caches = {
+    asset_caches = types.MappingProxyType({
         AssetType.ICON: _cache_icons,
         AssetType.BANNER: _cache_banners
-    }
+    })
 
     # All available event names & durations. Cached by the daemon nightly; read by the calendar command.
     cache_events = RedisCache()
--- a/bot/exts/backend/branding/_cog.py
+++ b/bot/exts/backend/branding/_cog.py
@@ -119,11 +119,9 @@ class Branding(commands.Cog):
     # Icons and banners in current rotation.
     # Keys (str) are download URLs, values (int) track the amount of times each
     # asset has been used in the current rotation.
-    _cache_icons = RedisCache()
-    _cache_banners = RedisCache()
     asset_caches = types.MappingProxyType({
-        AssetType.ICON: _cache_icons,
-        AssetType.BANNER: _cache_banners
+        AssetType.ICON: RedisCache(namespace="Branding.icon_cache"),
+        AssetType.BANNER: RedisCache(namespace="Branding.banner_cache")
     })
 
     # All available event names & durations. Cached by the daemon nightly; read by the calendar command.
--- a/bot/exts/backend/branding/_cog.py
+++ b/bot/exts/backend/branding/_cog.py
@@ -104,19 +104,26 @@ class Branding(commands.Cog):
     """
 
     # RedisCache[
-    #     "daemon_active": bool            | If True, daemon starts on start-up. Controlled via commands.
-    #     "event_path": str                | Current event's path in the branding repo.
-    #     "event_description": str         | Current event's Markdown description.
-    #     "event_duration": str            | Current event's human-readable date range.
-    #     "banner_hash": str               | SHA of the currently applied banner.
-    #     "icons_hash": str                | Compound SHA of all icons in current rotation.
-    #     "last_rotation_timestamp": float | POSIX UTC timestamp.
+    #     "daemon_active": bool                   | If True, daemon starts on start-up. Controlled via commands.
+    #     "event_path": str                       | Current event's path in the branding repo.
+    #     "event_description": str                | Current event's Markdown description.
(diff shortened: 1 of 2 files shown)


HassanAbouelela:
Kwzrd suggested we remove support for standalone banners to simplify the code here, thanks for the idea!

wookie184:
Is it important that we are able to link a banner and icon so they always appear at the same time? I'd guess they probably would end up in sync if they are in the same order in the `banners/` and `icons/` directories in the branding repo, but is there a way of making this more explicit?

HassanAbouelela:
@wookie184 I considered it, but after deliberating on the matter for a while, I didn't think the added complexity of trying to coordinate both was worth the payoff, especially considering the tight deadline we're trying to finish this in.

It can be added in the future, I imagine we can link them based off their name, but more logic would be necessary to properly handle rotating them at the same time (and certain edge cases such as having 2 banners, and 1 icon).

wookie184 (approved):
Tested and all seems good to go.

Was trying to think of a good way of handling matching banners and icons. Had one idea for a layout:
```python
icons
  - icon_green.png
  - icon_red.png
banners
  - banner_green.png
  - banner_red.png
```
To get the assets to use, we'd select a random icon, and then select a random banner in the same group (group is what's after the `_`, or something like that). This would be pretty flexible as you can easily have one icon with many banners or vice versa.
```
icons
  - icon1_green.png
  - icon2_green.png
banners
  - banner_green.png
```
And files with no "group" would just be considered part of the default group, so e.g. `icon.png` would match to `banner.png`.

HassanAbouelela:
@wookie184 That's a pretty good system provided you don't try to deal with combinations of groups, and no groups, such as:
```
icons:
icon_green.png
icon_red.png

banners:
banner_green.png
banner_red.png
banner_other.png
```

In this case, you have the following problem:
```
Day 1: Green icon, green banner
Day 2: Red icon, red banner
Day 3: Other banner, ?? icon.
````
You end up either having to repeat an icon two days in a row, cause a mismatch in the banners and icons, or smartly pick an icon that is not in the upcoming group. This complication is why I chose not to bother with support for this in this PR.

**Comment by wookie184** [comment]
I don't think there would be any need to support groups with only banners or only icons, that's something that could be validated in the branding repo to make sure it doesn't happen,

**Comment by HassanAbouelela** [comment]
I don’t agree. We already have events that don’t have banners, because not every contributor will make one. That would artificially limit us. Additionally, with your proposed system where you can have multiple assets in each “group”, you run into the same problems.

**Comment by wookie184** [comment]
> We already have events that don’t have banners

Do we? I couldn't find any, and it looks like we already enforce events having to have a banner https://github.com/python-discord/branding/blob/3173b230d8e2b0382a0cb19f0966a7100ad88910/events/validation.py#L68-L73

**Comment by HassanAbouelela** [comment]
We have events that don’t have as many banners as icons rather. This is the case with all events that have more than one icon currently, because nothing else is supported yet.

**Comment by wookie184** [comment]
> We have events that don’t have as many banners as icons rather. This is the case with all events that have more than one icon currently, because nothing else is supported yet.

That wouldn't be an issue with what I am proposing. For example, if you had 2 icons and 1 banner, you'd just put them all in a single group. A random icon would be chosen, and a random banner from the corresponding group is chosen (in this case always the same one).

**Comment by HassanAbouelela** [comment]
No the problem is with what I mentioned about having unbalanced groups.
> You end up either having to repeat an icon two days in a row, cause a mismatch in the banners and icons, or smartly pick an icon that is not in the upcoming group.

Ultimately it is solvable. Repeating an asset is one such solution. Which solution is deemed best is probably best left to the actual PR implementing that.

File: bot/exts/backend/branding/_cog.py [hunk 1]
```diff
@@ -104,19 +104,26 @@ class Branding(commands.Cog):
...
-    cache_icons = RedisCache()
+    # Icons and banners in current rotation.
+    # Keys (str) are download URLs, values (int) track the amount of times each
+    # asset has been used in the current rotation.
+    _cache_icons = RedisCache()
+    _cache_banners = RedisCache()
+    asset_caches = {
+        AssetType.ICON: _cache_icons,
+        AssetType.BANNER: _cache_banners
+    }
```
MarkKoz, line 126: Why not put the caches directly in the dictionary? I am pretty sure you can manually provide a namespace string.

MarkKoz: Also, do you think it's worth it to use `MappingProxyType` to make this immutable and prevent programming errors?

MarkKoz (approved):
The code looks fine but I'm not sure how to test it with an actual event. Works fine with the default event though.

bot/exts/backend/branding/_cog.py
HassanAbouelela: You can provide it, but it would require hardcoding the class name to get the same behavior. In retrospect, that probably would've been better. I'll change it.

I'll also look into the mapping proxy.

HassanAbouelela: Pushed the MappingProxy and Explicit namespaces in 90c34292 and a383040c respectively.

MarkKoz: Looks good.

PR #2156, opened 2022-04-29
revert bump to markdownify version, ensure redis cache ttl is set correctly

Numerlor:
The new versions bumped in https://github.com/python-discord/bot/commit/bd13ed68c6b0a33622ae8a6416c67948978bdd38 introduce conversions which cause the doc embeds to be formatted improperly, e.g.

|new markdownify|old markdownify|
|-|-|
|||
|||

I've also removed an invalid task cancel in 4cdaaf429ec27074117ce41c37aa47e72545c778, and ensured that an expire is set after a cache clear in df99fa8d8208c14e65ff623bb220c659d94aed2e


bot/exts/info/doc/_redis_cache.py
```diff
@@ -29,13 +31,20 @@ async def set(self, item: DocItem, value: str) -> None:
...
                 # An expire is only set if the key didn't exist before.
                 # If this is the first time setting values for this key check if it exists and add it to
                 # `_set_expires` to prevent redundant checks for subsequent uses with items from the same page.
-                self._set_expires.add(redis_key)
                 needs_expire = not await connection.exists(redis_key)
 
+            elif time.monotonic() - set_expire > WEEK_SECONDS:
+                # If we got here the key expired in redis and we can be sure it doesn't exist.
+                needs_expire = True
+                del self._set_expires[redis_key]
```
wookie184, line 44: The delete wont do anything currently as the value is reset on the next line

bot/exts/info/doc/_redis_cache.py
```diff
@@ -29,13 +31,20 @@ async def set(self, item: DocItem, value: str) -> None:
         needs_expire = False
 
         with await self._get_pool_connection() as connection:
-            if redis_key not in self._set_expires:
+            set_expire = self._set_expires.get(redis_key)
+            if set_expire is None:
                 # An expire is only set if the key didn't exist before.
                 # If this is the first time setting values for this key check if it exists and add it to
                 # `_set_expires` to prevent redundant checks for subsequent uses with items from the same page.
```
wookie184, line 38: This comment doesn't make sense here any more

wookie184 (changes requested):
This seems alright but i'm not 100% sure on the changes to the `set` function. I may just be misunderstanding though.

If `set` was called after a bot restart with an hour until a key expired, the expiry would not be changed but the `_set_expires` value would be set to a week. If `set` was called a day later, after the key had expired, the key would be set but no expiry would be set. I'm not sure if this would happen with the current workings of `doc` but if it didn't i'm not sure how the changes would help.

Would a more robust change be to replace `needs_expire = not await connection.exists(redis_key)` with checking the ttl is greater than 0 https://redis.io/commands/ttl/ (so we catch keys that didn't exist before and also keys with no expiry set). Then we can put that expiry in `_set_expires`, which seems safer than assuming it's just a week from now. If it wasn't set we can set the expiry and also set it in `_set_expires`.

bot/exts/info/doc/_redis_cache.py
Numerlor: Why not? The bock will execute on the first set of the key like the comment says

wookie184: The top line is fine. Just the next two about `_set_expires` are not relevant specifically to this block any more, as now `_set_expires` is always reset regardless of whether it already existed.

Numerlor:
Thanks, I forgot to take into account the ongoing expires in redis, should be fine now I believe

bot/exts/info/doc/_redis_cache.py
```diff
@@ -49,12 +65,15 @@ async def get(self, item: DocItem) -> Optional[str]:
...
+        pattern = f"{self.namespace}:{package}:*"
+
         with await self._get_pool_connection() as connection:
             package_keys = [
-                package_key async for package_key in connection.iscan(match=f"{self.namespace}:{package}:*")
+                package_key async for package_key in connection.iscan(match=pattern)
             ]
             if package_keys:
                 await connection.delete(*package_keys)
+                self._set_expires = {key for key in self._set_expires if not fnmatch.fnmatchcase(key, pattern)}
```
wookie184, line 76: Just noticed this is changing the dict into a set.

wookie184 (changes requested):
One thing, other than that all looks good.

wookie184 (approved):
All seems good to me. Thanks

HassanAbouelela (approved):

HassanAbouelela:
For the record, I'm highly opposed to the philosophy of not updating our deps because we can, however that's not a reason to block the PR for now since it'll fix the bug. We can work on getting it working in another PR. I suspect no one will pick that up though, so I'll try to get it to it once I have some free time.

Xithrius (approved):

PR #2166, opened 2022-05-02
Simplify PEP cog to use PEP API

wookie184:
There is a new PEP API (https://peps.python.org/api/peps.json) that saves us having to manually scrape files from github, largely simplifying the implementation.

As far as I can tell this was also the only usage of https://github.com/python-discord/bot/blob/5fe110b9a49144480ebca34dec65de91753994ec/bot/utils/caching.py, do we want to remove that given it's no longer used?


ChrisLovering:
> As far as I can tell this was also the only usage of https://github.com/python-discord/bot/blob/5fe110b9a49144480ebca34dec65de91753994ec/bot/utils/caching.py, do we want to remove that given it's no longer used?

I'd vote for removing it yea. We could add it to bot-core for future use if needed too.

bot/exts/info/pep.py
```diff
@@ -1,168 +1,90 @@
...
+                log.warning(
+                    f"Fetching PEP data from PEP API failed with code {resp.status}"
+                )
                 return
-
             listing = await resp.json()
 
-        log.trace("Got PEP URLs listing from GitHub API")
+        for pep_num, pep_info in listing.items():
+            self.peps[int(pep_num)] = pep_info
```
vivekashok1221, line 46: Forgive me for asking but can't you just write these two lines as:
```suggestion
        self.peps = listing
```
The keys of the dictionary will be strings in this case though.

vivekashok1221 (changes requested):
Wonderful work, wookie. I've reviewed and tested, I just have one small question.

bot/exts/info/pep.py
wookie184: I think i'll leave it, as being an int and having this code naming `pep_num` and `pep_info` makes it a little clearer what the structure of `peps` is imo.

vivekashok1221: Alrighty.

vivekashok1221 (approved):
LGTM :star2:

wookie184:
There's no rush to get this merged so we should wait for https://github.com/python/peps/issues/2584 to be resolved.

CAM-Gerlach:
Yeah, I'd suggest waiting if you can, since until #2584 is resolved (hopefully soon...™) the "API" is experimental and undocumented, and the path/endpoint, structure, etc. may change.

bot/exts/info/pep.py
```diff
@@ -1,168 +1,88 @@
...
 
 class PythonEnhancementProposals(Cog):
     """Cog for displaying information about PEPs."""
 
     def __init__(self, bot: Bot):
         self.bot = bot
-        self.peps: Dict[int, str] = {}
-        # To avoid situations where we don't have last datetime, set this to now.
-        self.last_refreshed_peps: datetime = datetime.now()
+        self.peps: dict[int, dict[str, Optional[str]]] = {}
```
shtlrs, line 21: This would be a good place to start using PEP 604 :D

shtlrs (approved):
Looks great, thanks wookie :D
One tiny comment for you :p

wookie184:
I'm going to close this PR for now as it's been stalled for a while, hopefully it can be reused at some point in the future.

PR #2176, opened 2022-05-21
Catch correct exception in clean cog

ChrisLovering:
When attempted to reply to a message, fi the message is not found, Discord.py does not raise a NotFound error, instead it raises the generic HTTPException.

Due to this, we now except this error when trying to reply to the message.


wookie184 (approved):

Xithrius (approved):
Yes.

PR #2178, opened 2022-05-22
Fix !clean users command to handle more than one user

### LINKED ISSUE #2171: Enable `!clean users` to  take multiple users as arguments
`!clean users` is a valid alias for `!clean user` but it can only take one user. The `!clean` command can already handle multiple users, so this is a very minor change.

Anyone can implement

Autonymic:
A minor fix that should close #2171.


wookie184 (approved):
Looks good to me, thanks!

ichard26 (approved):
Looks and works great to me, thank you @Autonymic! Good luck with your devops course!

PR #2179, opened 2022-05-22
Updating and improving clarity in help documentation for clean subcommands

### LINKED ISSUE #2170: Imrpove help text of the !clean subcommands
The `!clean` command has multiple modes, and it can be confusing to use.

- `!clean bots` mentions a `traverse` argument that doesn't exist.
- `!clean regex` has an example of formatting the regex pattern that I think needs more escape characters.
- Most of the commands can take either a message or a time limit. They should make it clear that providing a message limits the search to its channel, and to clean across channels you need to provide a time limit instead.

Autonymic:
This should close #2170.

I have removed the obsolete description about the traverse argument in the clean bots subcommand and fixed as well as added to the description of the clean regex subcommand.

I also improved the explanations for the message or timelimit argument and their relation to the channel argument across all clean subcommands.


bot/exts/moderation/clean.py
```diff
@@ -531,11 +537,21 @@ async def clean_regex(
...
+
+        The `regex` pattern must be provided enclosed in backticks, that will make it appear as a code section.
+
+        For example: \u02CB[0-9]\u02CB, which should appear as `[0-9]`.
+
+        If the `regex` pattern contains spaces, it still needs to be enclosed in double quotes on top of that.
+
+        For example: "\u02CB[0-9]\u02CB", which should appear as "`[0-9]`".
+
+        Do not copy and paste the backticks from the examples, they are special unicode characters that will not work.
```
wookie184, line 554: I think this bit is a bit wordier than it needs to be, I think something like this would suffice (note backticks can be escaped with `\`):
```
        The `regex` pattern must be enclosed in backticks (e.g. \`[0-9]\`). 
        If it contains a space it must also be enclosed in double quotes (e.g. "\`test [0-9]\`").
```
image

I'm not exactly sure what change the original issue was proposing to this part though. Thoughts?

wookie184 (changes requested):
Thanks for doing this, I just have one comment.

bot/exts/moderation/clean.py
Autonymic: The problem is that the intent seems to be to show how to write the regex part of the command, e.g. must be enclosed in backticks `[0-9]`, but the problem is when it appears on discord it just turns into a code block due to the [0-9] being enclosed in ``, potentially making a less knowledgeable user confused.

When I tried escaping the standard backticks with backslash the linter didn't seem to like it, so I just chose to use special unicode characters that won't turn into a code block, even though it may cause confusion if a user tries to copypaste those backticks.

wookie184: Ah, in that case you can escape the backticks explicitly to make the linter happy
```
        The `regex` pattern must be enclosed in backticks (e.g. \\`[0-9]\\`).
        If it contains a space it must also be enclosed in double quotes (e.g. "\\`test [0-9]\\`").
```
(is the same as my screenshot above)

Autonymic: Yeah that seemed to do the trick, thanks! I removed some of the excessive words, so it should be a bit cleaner now.

wookie184 (approved):
Looks good, thanks.

ChrisLovering (approved):
Looks good, thanks :)

PR #2180, opened 2022-05-22
Log more data on failed clean

ChrisLovering:
This adds request data as a sentry breadcrumb when clean cog fails, so we can understand the data that is causing the site api to error.

This also updates the log string to use % based format strings, rather than a mix of f-string and % string which caused the %r to not work.


wookie184 (approved):
Hard to test properlu but looks fine and doesn't seem to break anything

bot/exts/moderation/modlog.py
```diff
@@ -53,24 +55,35 @@ async def upload_log(
...
+                    "deletedmessage_set": deletedmessage_set,
+                }
+            )
+        except ResponseCodeError as e:
+            add_breadcrumb(
+                category="api_error",
+                message=str(e),
+                level="error",
+                data=deletedmessage_set,
+            )
```
MarkKoz, line 85: There's a discrepancy in the documentation of this function: https://getsentry.github.io/sentry-python/api.html?highlight=add_breadcrumb#sentry_sdk.add_breadcrumb vs https://docs.sentry.io/platforms/python/enriching-events/breadcrumbs/#manual-breadcrumbs

I don't know how to test this so let's just try and fix it if it doesn't work.

MarkKoz (approved):

bot/exts/moderation/modlog.py
ChrisLovering: Yea, I was following the latter. I guess we can just find out.

PR #2181, opened 2022-05-28
Add special handling for eval command followed by backticks.

### LINKED ISSUE #2047: !e Command won't work if codeblock is right after
# Problem

image

The code in the above image wont execute

image

The code in the above image will execute

wookie184:
Closes #2047

I think this approach should be fairly robust. I figured this wasn't worth extending to inline codeblocks or other commands.


wookie184:
I've fixed the existing tests, although i'm not convinced trying to write new ones for this is worth it.

MarkKoz (approved):

ichard26 (approved):

PR #2182, opened 2022-05-28
Clean from public active threads when channels is set to *

ChrisLovering:


MarkKoz (changes requested):
Doesn't work — it deletes messages in channels, but does not delete messages in threads. Does the bot have to join the thread first?

MarkKoz (approved):

ichard26 (approved):
After a lot of trial and error and confusion, I can confirm it works.

PR #2189, opened 2022-06-11
Clean improvments

mbaruh:
- Fixed a bug where `clean between` deleted the last message. Clarified in the appropriate command descriptions that the clean limits are exclusive.
- Added a shortcut to a common use case of deleting all recent public messages of a user. Since the `purge` alias was never used for the clean group I repurposed it for this new command.

### DIFF
--- a/bot/exts/moderation/clean.py
+++ b/bot/exts/moderation/clean.py
@@ -182,7 +182,7 @@ class Clean(Cog):
             return first_limit < message.created_at < second_limit
 
         def predicate_after(message: Message) -> bool:
-            """Check if the message is older than the first limit."""
+            """Check if the message is younger than the first limit."""
             return message.created_at > first_limit
 
         predicates = []
--- a/bot/exts/moderation/clean.py
+++ b/bot/exts/moderation/clean.py
@@ -621,7 +621,11 @@ class Clean(Cog):
 
     @command()
     async def purge(self, ctx: Context, users: Greedy[User], age: Optional[Union[Age, ISODateTime]] = None) -> None:
-        """Clean messages of users from all public channels up to a certain message age (10 minutes by default)."""
+        """
+        Clean messages of users from all public channels up to a certain message age (10 minutes by default).
+
+        The age is *exclusive*, meaning that `10s` won't delete a message exactly 10 seconds old.
+        """
         if age is None:
             age = await Age().convert(ctx, "10M")
         await self._clean_messages(ctx, channels="*", users=users, first_limit=age)
--- a/bot/exts/moderation/clean.py
+++ b/bot/exts/moderation/clean.py
@@ -8,7 +8,7 @@ from itertools import takewhile
 from typing import Callable, Iterable, Literal, Optional, TYPE_CHECKING, Union
 
 from discord import Colour, Message, NotFound, TextChannel, Thread, User, errors
-from discord.ext.commands import Cog, Context, Converter, Greedy, group, has_any_role
+from discord.ext.commands import Cog, Context, Converter, Greedy, command, group, has_any_role
 from discord.ext.commands.converter import TextChannelConverter
 from discord.ext.commands.errors import BadArgument
 
@@ -452,7 +452,7 @@ class Clean(Cog):
 
     # region: Commands
 
-    @group(invoke_without_command=True, name="clean", aliases=["clear", "purge"])
+    @group(invoke_without_command=True, name="clean", aliases=("clear",))
     async def clean_group(
         self,
         ctx: Context,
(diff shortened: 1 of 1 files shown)


ChrisLovering (approved):

bot/exts/moderation/clean.py
```diff
@@ -615,6 +619,13 @@ async def clean_cancel(self, ctx: Context) -> None:
         await self._send_expiring_message(ctx, message)
         await self._delete_invocation(ctx)
 
+    @command()
+    async def purge(self, ctx: Context, users: Greedy[User], age: Optional[Union[Age, ISODateTime]] = None) -> None:
+        """Clean messages of users from all public channels up to a certain message age (10 minutes by default)."""
```
MarkKoz, line 624: Is this limit inclusive? It should be documented.

bot/exts/moderation/clean.py
```diff
@@ -615,6 +619,13 @@ async def clean_cancel(self, ctx: Context) -> None:
...
+        if age is None:
+            age = await Age().convert(ctx, "10M")
```
MarkKoz, line 626: I suppose discord.py does not apply decorators to default values? Ideally it would use an actual default, because I believe that gets shown in the help info for the command. Can that still be used somehow?

MarkKoz (changes requested):

bot/exts/moderation/clean.py
ChrisLovering: With age being an offset from the time the command was ran, I'm not sure how we could get this to work with a deco evaluated at startup.

bot/exts/moderation/clean.py
mbaruh: Hmm yeah, I didn't consider it because it would only be relevant if someone took a message, deciphered its exact age from the ID, and used it with the expectation the message would be deleted. Added in 154cd04 for good measure.

bot/exts/moderation/clean.py
mbaruh: Yeah I don't see a good way to do that. I used the same strategy for the default mute duration.

bot/exts/moderation/clean.py
```diff
@@ -179,7 +179,7 @@ def predicate_regex(message: Message) -> bool:
 
         def predicate_range(message: Message) -> bool:
             """Check if the message age is between the two limits."""
-            return first_limit <= message.created_at <= second_limit
+            return first_limit < message.created_at < second_limit
```
MarkKoz, line 182: I'm confused how this makes a difference. The messages are fetched using `Messageable.history()` with the same limits. For `history()`, those limits are already exclusive. I don't understand why `predicate_range` and `predicate_after` are even needed.

mbaruh: Because the search might go through the cache. If history is indeed exclusive then I guess this predicate could be dropped in those cases, but I don't mind another guarantee if it already exists.

MarkKoz: `predicate_after` should also be exclusive then.

mbaruh: Hmm good point. It doesn't cause a bug because the iteration excludes the stopping point.

bot/exts/moderation/clean.py
```diff
@@ -179,11 +179,11 @@ def predicate_regex(message: Message) -> bool:
...
 
         def predicate_after(message: Message) -> bool:
             """Check if the message is older than the first limit."""
```
MarkKoz, line 185: Shouldn't this be "younger" rather than "older"?

MarkKoz (changes requested):
This argument can be None, so it should be annotated as such https://github.com/python-discord/bot/blob/0bbcde6af81405380bbae1eca3556338331b03fa/bot/exts/moderation/clean.py#L244=

mbaruh:
> This argument can be None, so it should be annotated as such https://github.com/python-discord/bot/blob/0bbcde6af81405380bbae1eca3556338331b03fa/bot/exts/moderation/clean.py#L244=

There *is* an issue with the annotation, but it's actually different. `after` is the one which isn't optional, while `before` is. This is enforced by `_validate_input`

MarkKoz:
> > This argument can be None, so it should be annotated as such https://github.com/python-discord/bot/blob/0bbcde6af81405380bbae1eca3556338331b03fa/bot/exts/moderation/clean.py#L244=
> 
> There _is_ an issue with the annotation, but it's actually different. `after` is the one which isn't optional, while `before` is. This is enforced by `_validate_input`

Technically both are optional since they're just passed to `history()`, and that function allows both to be None.

mbaruh:
> Technically both are optional since they're just passed to `history()`, and that function allows both to be None.

yeah but this is about how the function is expected to be used, rather than what it can technically accept

MarkKoz:
> > Technically both are optional since they're just passed to `history()`, and that function allows both to be None.
> 
> yeah but this is about how the function is expected to be used, rather than what it can technically accept

I don't feel strongly about this but if it were me I wouldn't impose artificial constraints on it.

mbaruh:
If I write a function which takes a number and adds 1 to it, I could assign a default value of 0 to the number, but I have no reason to. That's not how I want the function to behave. If I don't supply it a number I want it to yell at me, not because of what the function can do in practice if I allow it, but by my choice.

bot/exts/moderation/clean.py
lemonsaurus:
```suggestion
            """Check if the message is younger than the first limit."""
```

If it should, here's a suggestion for easy peasy one-click fix.

MarkKoz (approved):

PR #2196, opened 2022-06-14
Enhance guild_invite based autobans

ChrisLovering:
This PR includes various improvements to guild_invite based autobans as the strucutre of the filter response is slightly different to other filters, causing bad UX.

In this PR I have auto prepended the guild name to guild invite filters, to ensure the guild name is always present in the reason. This ensures that when looking at the block list, or when notifying users about a new autoban guild_invite filter being added, the guild name is always there.

The message sent when a new autoban filter is added also now includes the filter list type, to make it more obvious at a glance when it would be triggered.

I have also made it so that any type of autoban filter does not ping mods when sending the log to mod-alerts, as they have already been auto-actioned by the bot.

### DIFF
--- a/bot/exts/filters/filter_lists.py
+++ b/bot/exts/filters/filter_lists.py
@@ -55,17 +55,20 @@ class FilterLists(Cog):
         """Add an item to a filterlist."""
         allow_type = "whitelist" if allowed else "blacklist"
 
-        # If this is a server invite, we gotta validate it.
+        # If this is a guild invite, we gotta validate it.
         if list_type == "GUILD_INVITE":
             guild_data = await self._validate_guild_invite(ctx, content)
             content = guild_data.get("id")
 
-            # Unless the user has specified another comment, let's
-            # use the server name as the comment so that the list
-            # of guild IDs will be more easily readable when we
-            # display it.
-            if not comment:
-                comment = guild_data.get("name")
+            # Some guild invites are autoban filters, which require the mod
+            # to set a comment which includes [autoban].
+            # Having the guild name in the comment is still useful when reviewing
+            # filter list, so prepend it to the set comment in case some mod forgets.
+            comment = " - ".join(
+                comment_part
+                for comment_part in (guild_data.get("name", ""), comment)
+                if comment_part
+            )
 
         # If it's a file format, let's make sure it has a leading dot.
         elif list_type == "FILE_FORMAT" and not content.startswith("."):
--- a/bot/exts/filters/filter_lists.py
+++ b/bot/exts/filters/filter_lists.py
@@ -64,9 +64,11 @@ class FilterLists(Cog):
             # to set a comment which includes [autoban].
             # Having the guild name in the comment is still useful when reviewing
             # filter list, so prepend it to the set comment in case some mod forgets.
+            guild_name_part = f'Guild "{guild_data["name"]}"' if "name" in guild_data else None
+
             comment = " - ".join(
                 comment_part
-                for comment_part in (guild_data.get("name", ""), comment)
+                for comment_part in (guild_name_part, comment)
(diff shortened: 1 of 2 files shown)


bot/exts/filters/filter_lists.py
```diff
@@ -55,17 +55,20 @@ async def _add_data(
...
-            # display it.
-            if not comment:
-                comment = guild_data.get("name")
+            # Some guild invites are autoban filters, which require the mod
+            # to set a comment which includes [autoban].
+            # Having the guild name in the comment is still useful when reviewing
+            # filter list, so prepend it to the set comment in case some mod forgets.
+            comment = " - ".join(
+                comment_part
+                for comment_part in (guild_data.get("name", ""), comment)
```
MarkKoz, line 69: A format like `Guild "name here" - comment here` would be even clearer. It's not self-evident what the part before the dash means. For example, a guild named "123"

bild

Quoting the guild name will also help set the boundary between the comment and the guild name. Otherwise, a guild name with a dash in it could lead to further confusion.

bot/exts/filters/filtering.py
```diff
@@ -438,7 +442,7 @@ async def _send_log(
         content = str(msg.author.id)  # quality-of-life improvement for mobile moderators
 
         # If we are going to autoban, we don't want to ping and don't need the user ID
-        if reason and "[autoban]" in reason:
+        if auto_ban:
```
MarkKoz, line 445: Did the previous approach not work for invite filters since it could contain multiple reasons? In that case, wouldn't the type annotation of the `reason` argument for this function be wrong, or are the reasons concatenated before this function is called?

bot/exts/filters/filtering.py
```diff
@@ -387,16 +387,19 @@ async def _filter_message(self, msg: Message, delta: Optional[int] = None) -> No
...
+                        # Also pass this to _send_log so mods are not pinged filter matches that are auto-actioned
                         autoban = reason and "[autoban]" in reason.lower()
                         if not autoban and filter_name == "filter_invites" and isinstance(result, dict):
                             autoban = any(
                                 "[autoban]" in invite_info["reason"].lower()
                                 for invite_info in result.values()
                                 if invite_info.get("reason")
                             )
+
+                        await self._send_log(filter_name, _filter, msg, stats, reason, autoban=autoban)
```
MarkKoz, line 401: Might be better to instead rename the keyword to `autoban` for consistency.
```suggestion
                        await self._send_log(filter_name, _filter, msg, stats, reason, auto_ban=autoban)
```

MarkKoz (changes requested):

bot/exts/filters/filter_lists.py
ChrisLovering: https://github.com/python-discord/bot/pull/2196/commits/af4339e1436bf554188205341a7b82becb573af1

bot/exts/filters/filtering.py
ChrisLovering: https://github.com/python-discord/bot/pull/2196/commits/11f329f6847c0e5f97dac248507bbd2970804f43

bot/exts/filters/filtering.py
ChrisLovering: for guild_invites, reason will always be passed as None.

It's initialised as None here https://github.com/python-discord/bot/blob/main/bot/exts/filters/filtering.py#L338 and then since _has_invites doesn't return a tuple, this conditional means it is never updated.

https://github.com/python-discord/bot/blob/main/bot/exts/filters/filtering.py#L340-L343

MarkKoz (approved):
cool and good

brad90four (approved):
Looks almost as good as the author.

PR #2197, opened 2022-06-17
Fix help embed views

mbaruh:
Due to various breaking changes in the discord.py library the views of the help embed stopped working properly.

- Use `add_item` in `CommandView`. This was done in GroupView but was apparently missed in `CommandView`, and the former (hacky) method no longer works.
- Instead of editing the help message normally (which required to then defer the response), the message edit is made to be the response itself. It seems like the callback would automatically defer the response if none was made, but that no longer happens. Thanks to @onerandomusername for pointing me to the right method.

### DIFF
--- a/bot/exts/info/help.py
+++ b/bot/exts/info/help.py
@@ -57,16 +57,13 @@ class SubcommandButton(ui.Button):
 
     async def callback(self, interaction: Interaction) -> None:
         """Edits the help embed to that of the subcommand."""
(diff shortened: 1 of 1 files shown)


MarkKoz (approved):
I'm not familiar with the interactions API, so I can't be effective in reviewing the code. I can't even find where it's documented. However, these changes do work in practice.

ChrisLovering (approved):

PR #2199, opened 2022-06-19
filter out header links in descriptions

Numerlor:
Continuation of #2187, spotted the paragraph in a description of `not in` 
 didn't think of symbols without signatures when doing the last PR.

Adds the filtering to the description tags.
I've also noticed some things were being filtered with a regex that I glossed over last time, I moved over the source filtering to go through html like headerlinks do; not sure what the backslashes were filtering.

### DIFF
--- a/bot/exts/info/doc/_parsing.py
+++ b/bot/exts/info/doc/_parsing.py
@@ -256,4 +256,9 @@ def get_symbol_markdown(soup: BeautifulSoup, symbol_data: DocItem) -> Optional[s
         signature = get_signatures(symbol_heading)
         description = get_dd_description(symbol_heading)
 
+    for description_element in description:
+        if isinstance(description_element, Tag):
+            for tag in description_element.find_all("a", class_="headerlink"):
+                tag.decompose()
+
     return _create_markdown(signature, description, symbol_data.url).strip()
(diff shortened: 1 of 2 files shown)


bot/exts/info/doc/_parsing.py
```diff
@@ -256,4 +256,9 @@ def get_symbol_markdown(soup: BeautifulSoup, symbol_data: DocItem) -> Optional[s
         signature = get_signatures(symbol_heading)
         description = get_dd_description(symbol_heading)
 
+    for description_element in description:
+        if isinstance(description_element, Tag):
+            for tag in description_element.find_all("a", class_="headerlink"):
```
MarkKoz, line 261: Can't you re-use `_filter_signature_links` in this case?

Numerlor: I suppose it could be used, but I'm not convinced if it'd be worth it. 
Using the filter would only change the find_all call which is already clear enough, while also introducing a bit more work to the filtering and adding an another import for it

MarkKoz (approved):

bot/exts/info/doc/_parsing.py
MarkKoz: Fair enough.

ChrisLovering (approved):

PR #2200, opened 2022-06-28
Ignore auto mod notificaitons

ChrisLovering:
When a message triggers an auto_moderation_action notification, Discord re-writes the author field for the system message to look as if it's from the original author.

This means those messages counted towards our anti-spam filter. This both triggered the filter too early, and also resulted in those auto mod messages being deleted when the filter was hit.

To ignore these messages, d.py had to be bumped to thee version where it added support for auto mod.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -297,21 +297,22 @@ testing = ["pre-commit"]
 
 [[package]]
 name = "fakeredis"
-version = "1.8.1"
+version = "1.7.5"
 description = "Fake implementation of redis API for testing purposes."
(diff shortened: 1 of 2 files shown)


ChrisLovering:
Once approved, we need to merge https://github.com/python-discord/bot-core/pull/93 and make a non-pre release of botcore and update the deps here.

HassanAbouelela (approved):

Senjan21 (approved):
Tested, appears to be working. Thanks!

PR #2205, opened 2022-06-29
Limit the ext cog to 1 action at a time

ChrisLovering:
Now that cog loading actually has lots of logic in it, rather than spawning tasks to do the actual load work, loading cogs can take time.

Previously we gave no feedback to the caller that something was actually happening. This could mean the caller would try to run the command again, possibly getting the cogs into a bad state.

This change both enforces only 1 action at a time, and gives the caller feedback that their action is in progress.

### DIFF
--- a/bot/exts/utils/extensions.py
+++ b/bot/exts/utils/extensions.py
@@ -34,6 +34,7 @@ class Extensions(commands.Cog):
 
     def __init__(self, bot: Bot):
         self.bot = bot
+        self.action_in_progress = False
 
     @group(name="extensions", aliases=("ext", "exts", "c", "cog", "cogs"), invoke_without_command=True)
     async def extensions_group(self, ctx: Context) -> None:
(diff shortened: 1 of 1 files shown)


bot/exts/utils/extensions.py
```diff
@@ -151,17 +147,27 @@ def group_extension_statuses(self) -> t.Mapping[str, str]:
...
 
-    async def batch_manage(self, action: Action, *extensions: str) -> str:
+    async def batch_manage(self, action: Action, ctx: Context, *extensions: str) -> None:
         """
-        Apply an action to multiple extensions and return a message with the results.
+        Apply an action to multiple extensions, giving feedback to the invoker while doing so.
 
         If only one extension is given, it is deferred to `manage()`.
         """
+        if self.action_in_progress:
```
HassanAbouelela, line 156: This will block any extensions from loading/unloading while one is going, do we care? We could make the lock unique to each extension.

ChrisLovering: I'm not sure we care, the majority of these will take <5s, so the chance of an accidental overlap by 2 callers is very small.

If a single caller wanted to load.unload multiple exts then they could pass a list to the cog and it would work just fine.

I'm not sure it's worth the complexity.

HassanAbouelela (approved):

bot/exts/utils/extensions.py
```diff
@@ -34,6 +34,7 @@ class Extensions(commands.Cog):
 
     def __init__(self, bot: Bot):
         self.bot = bot
+        self.action_in_progress: bool = False
```
Xithrius, line 37: I feel like it is unnecessary to add an annotation here.

```suggestion
        self.action_in_progress = False
```

Xithrius (changes requested):
Besides this, looks good.

bot/exts/utils/extensions.py
ChrisLovering: Removed, and I rebased onto main.

Xithrius (approved):

PR #2225, opened 2022-07-17
Use the view clear on timeout feature from bot-core in snekbox

### LINKED ISSUE #2223: Remove the eval view on the timeout
Currently, the buttons aren't getting removed and so on timeout it just gives `interaction failed`
It would be better if they were removed.

ChrisLovering:
This will mean the buttons will be cleared from the response on interaction timeout.

This PR also adds the Python version to the output from snekbox.

I have also made the 3.11 snekbox start container by default, since snekbox uses 3.11 by default, it makes sense for this one to be started by default, and the 3.10 container to be opt-in.

Closes #2223

### DIFF
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -233,19 +233,19 @@ class Snekbox(Cog):
         return code, args
 
     @staticmethod
-    def get_results_message(results: dict, job_name: str) -> Tuple[str, str]:
+    def get_results_message(results: dict, job_name: str, python_version: Literal["3.10", "3.11"]) -> Tuple[str, str]:
         """Return a user-friendly message and error corresponding to the process's return code."""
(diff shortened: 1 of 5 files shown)


ChrisLovering:
Marking as do not merge as the version of bot-core it uses will need to be updated to not be a beta release once https://github.com/python-discord/bot-core/pull/106 is merged.

minalike (approved):
Was able to test on Windows, works great. Ran into some errors testing on Mac that we can attempt to figure out at a later date.

- [x] 3.11 by default
- [x] buttons go away after a few minutes
- [x] message indicates what version of Python is being used

image

ChrisLovering:
I forgot to mention in the PR, but the timeout is defaulted to 180 seconds.
https://github.com/python-discord/bot-core/blob/main/botcore/utils/interactions.py#L26

mbaruh (approved):

PR #2230, opened 2022-07-25
Add nomodule tag

meatballs:
Fixes python-discord/meta#189


bot/resources/tags/nomodule.md
```diff
@@ -0,0 +1,15 @@
...
+
+You can read about Python environments at `!tags environments` and `!tags venv`.
+
+Common causes of this problem include:
+
+• You installed your package using `pip install ...`. It could be that the `pip` command is not pointing the the environment where your code runs. For greater control, you could instead run pip as a module within the python environment you specify:
+```
+python -m pip install <your_package>
+```
+(You may have to specify the full path to the python executable).
```
CaedenPH, line 13: This is slightly ambiguous and may confuse the reader

CaedenPH (commented):
Explains well in detail about the module not found and how to fix but doesn't necessarily mention about what a module actually is. Seems more like a temporary fix rather than teaching the reader what exactly a module is and how to avoid module not found.

Xithrius:
@meatballs Hello. It seems you have a request for change on this PR.

Please resolve it, so we're able to get this merged.

Thanks!

meatballs:
What's the requested change? I can see an opinion but it's not one I agree with.

bot/resources/tags/nomodule.md
MarkKoz: Agreed. It may be best to leave it out unless it can be explained briefly and clearly how "xxx -m pip install ..." means the package is available if using "xxx file.py"/"xxx -m mymodule" to run the code.

MarkKoz:
> Explains well in detail about the module not found and how to fix but doesn't necessarily mention about what a module actually is. Seems more like a temporary fix rather than teaching the reader what exactly a module is and how to avoid module not found.

Tags must be limited in scope. Otherwise they become long and loose their effectiveness. I don't think it's possible to explain what a module is in this tag in addition to what the error is without making the content long and overwhelming. Did you have something in mind that would suggest otherwise?

bot/resources/tags/nomodule.md
```diff
@@ -0,0 +1,15 @@
+**ModuleNotFoundError**
+
+If you've installed a package but you're getting a ModuleNotFoundError, it's likely that the environment where your code is running is different from the one where you did the installation.
```
MarkKoz, line 3: What do you think about briefly clarifying here that the error happens upon importing the purportedly installed package?

ChrisLovering (approved):
lgtm!

Issue #2231, opened 2022-07-25
`!remind edit` bugs and potential reminder enhancements


TizzySaurus:
Further to discussion with @HassanAbouelela in #dev-contrib, I'm creating this issue to outline some potential changes to the `!remind` command that can be made.

__Bugs:__
 - `!remind edit` doesn't validate that the reminder was actually changed before sending the API call and sending success message. 
 Resolving this would mean having to `GET` the reminder being edited so that the appropriate fields can be compared.
- `!remind edit content` doesn't support using a referenced message, but `!remind new` does.

__Enhancements:__
- Update `!remind list` to actually mention the reminder's mentions instead of displaying the `.name` attribute, since it's in an embed (won't ping) and will be more explicit.
- Add support for deleting multiple reminders at once (means changing the typehint from `int` to `Greedy[int]` and then iterating).

@HassanAbouelela suggested that PRs should ideally be limited in the amount of things they add, so I'd do a separate PR for each of the bugs, and then both enhancements in 1 PR since the first is a one-line change.

HassanAbouelela:
Doing the enhancements together sounds fine 👍 

I have a couple concerns wrt the implementation of the other two.

Bug 1. I don't know if it's worth it to do a second API call, when the current behavior doesn't seem to be a problem. I don't mind either way though

Bug 2. Not sure this is a bug, more of a new feature. Off the top of my head, I don't believe there's a one-to-one relationship between a message and a reminder. You can create multiple reminders to one message, and you can have multiple users with multiple reminders to the same message. You can have reminders with no messages. How would replying to a message tie it to the reminder?

TizzySaurus:
> Bug 1. I don't know if it's worth it to do a second API call, when the current behavior doesn't seem to be a problem. I don't mind either way though

It's not really a "problem", more a QOL thing. It just seems weird to have an edit command that doesn't actually verify if it was edited. If we decide it's not worth due to the API call then that's fine with me.

> How would replying to a message tie it to the reminder?
The same way it does for `!remind new`, i.e. use the referenced message's content as the reminder content, or if there's no content, it'll instead set the content to "See referenced message".

If we want different behaviour, then `!remind new` should be updated to reflect the new behaviour as well.

HassanAbouelela:
[...]

PR #2232, opened 2022-07-25
Add ability to delete multiple reminders at once & display mentions in `!remind list`.

TizzySaurus:
Part 1 of #2231.

__Delete Multiple Reminders__
This has been tested with the help of @ichard26 (thanks!).

The behaviour behind the deletion is still the exact same (you can't delete others' reminders unless you're an admin), just has been reformatted to account for the fact that a lock needs to be applied to each individual id. 

There's also some changes to prevent a message being sent for each deletion fail -- we instead send one message after processing all ids.

_All reminders successfully deleted:_
image

_Some reminders successfully deleted:_
image

_No reminders were successfully deleted:_
image

__Display Mentions__
`!remind list` will now display the `.mention` attribute, instead of the `.name` attribute to be clearer.

_Before:_
image

_After:_
image


bot/exts/utils/reminders.py
```diff
@@ -450,35 +453,76 @@ async def edit_reminder(self, ctx: Context, id_: int, payload: dict) -> None:
...
-        await self._send_confirmation(
-            ctx,
-            on_success="That reminder has been deleted successfully!",
-            reminder_id=id_
+    @remind_group.command("delete", aliases=("remove", "cancel"))
+    async def delete_reminder(self, ctx: Context, ids: Greedy[int]) -> None:
+        """Delete one of your active reminders."""
+        deleted_ids = []
+        for id_ in ids:
+            try:
```
wookie184, line 471: Could consider making `ids` a set to remove duplicates.

I also wonder if we'd want an upper limit on the length to prevent spamming the site api (you could currently cause nearly 1000 api requests with `!remind delete 1 1 1 ...`)

wookie184 (changes requested):

bot/exts/utils/reminders.py
TizzySaurus: Yep, agreed :+1:

I've set the limit to 5, since that's high enough to likely not inconvenience the user, but l low enough to still prevent API abuse.

wookie184 (approved):
Thanks

vivekashok1221 (approved):
Reviewed and tested.
LGTM :star2: .

bot/exts/utils/reminders.py
```diff
@@ -348,8 +352,8 @@ async def list_reminders(self, ctx: Context) -> None:
             expiry = time.format_relative(remind_at)
 
             mentions = ", ".join([
-                # Both Role and User objects have the `name` attribute
-                mention.name async for mention in self.get_mentionables(mentions)
+                # Both Role and User objects have the `mention` attribute
+                mentionable.mention async for mentionable in self.get_mentionables(mentions)
```
wookie184, line 356: I just remembered, we don't usually put mentions in embeds as they often don't resolve (just show the ID in the mention rather than the name), so I'm not sure about this change.

wookie184 (changes requested):

bot/exts/utils/reminders.py
TizzySaurus: Fixed, as per my comment on Discord :+1:

wookie184 (approved):

PR #2240, opened 2022-08-03
Updated `purge` to require at least one user

### LINKED ISSUE #2238: Make `users` argument of purge command required
When `user` argument is not provided to the `!purge` command, the greedy convertor returns an empty list. This results in the bot deleting messages of ALL users in the last 10 minutes.
Some kind of a check should be added to prevent this from happening.
Also applicable to the `!clean users` command.

ionite34:
Closes #2238 

## Summary
- Added raising of `BadArgument` when the greedy match of users returns an empty list
- Updated help message

## Demo
purge_demo

### DIFF
--- a/bot/exts/moderation/clean.py
+++ b/bot/exts/moderation/clean.py
@@ -7,7 +7,7 @@ from datetime import datetime
 from itertools import takewhile
 from typing import Callable, Iterable, Literal, Optional, TYPE_CHECKING, Union
 
-from discord import Colour, Embed, Message, NotFound, TextChannel, Thread, User, errors
+from discord import Colour, Message, NotFound, TextChannel, Thread, User, errors
 from discord.ext.commands import Cog, Context, Converter, Greedy, command, group, has_any_role
 from discord.ext.commands.converter import TextChannelConverter
 from discord.ext.commands.errors import BadArgument
--- a/bot/exts/moderation/clean.py
+++ b/bot/exts/moderation/clean.py
@@ -7,7 +7,7 @@ from datetime import datetime
 from itertools import takewhile
 from typing import Callable, Iterable, Literal, Optional, TYPE_CHECKING, Union
 
-from discord import Colour, Message, NotFound, TextChannel, Thread, User, errors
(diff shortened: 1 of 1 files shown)


bot/exts/moderation/clean.py
```diff
@@ -627,12 +627,18 @@ async def clean_cancel(self, ctx: Context) -> None:
     @command()
     async def purge(self, ctx: Context, users: Greedy[User], age: Optional[Union[Age, ISODateTime]] = None) -> None:
         """
-        Clean messages of users from all public channels up to a certain message age (10 minutes by default).
+        Clean messages of `users` from all public channels up to a certain message `age` (10 minutes by default).
 
-        The age is *exclusive*, meaning that `10s` won't delete a message exactly 10 seconds old.
+        Requires 1 or more users to be specified. For channel-based cleaning, use `clean` instead.
+
+        The `age` is *exclusive*, meaning that `10s` won't delete a message exactly 10 seconds old.
```
wookie184, line 634: I don't think this would really matter in practice. A message being *exactly* n seconds long is unlikely given snowflakes have millisecond precision (which I assume is being used?).

Also note that in the help command everything is in italics already, so "exclusive" is actually excluded from being in italics.

wookie184: nvm that was there already 😅

wookie184 (approved):
Thanks!

bot/exts/moderation/clean.py
ionite34: I could also remove that original last line? I agree that it's kind of not needed as it's unlikely that kind of precision matters in a typed command.

Also in the demo image you can see the `* *` causing the weird italic font behavior.

ionite34: Updated in 87de12c

```diff
- The `age` is *exclusive*, meaning that `10s` won't delete a message exactly 10 seconds old.
+ `age` can be a duration or an ISO 8601 timestamp.
```

wookie184 (approved):

minalike (approved):
Tested, thanks for the PR! This will prevent many happy (but bad) accidents 😃

PR #2243, opened 2022-08-05
New generator-iterator tag

Okimii:
Would love all the feedback!

### DIFF
--- a/bot/resources/tags/generators-iterators.md
+++ b/bot/resources/tags/generators-iterators.md
@@ -1,0 +1,45 @@
+**Generators and Iterators**
+
+What is a generator and an iterator and what is the difference between each other?
+
+**Generator**
+
+A generator is a function that may contain a `yield` or sometimes `return` statements, a generator can also be an expression(genexpr).
+```py
+>>> print(num for num in range(11))
+<generator object <genexpr> at 0x0000000000000000>
+```
+
+Well, what does a generator do and what is it any different from a list, for example?
+The generator yields one item at a time and generates items only when in demand. Whereas, in a list, Python reserves memory for the whole list. A generator can be an iterator but not vice versa.
+
+
+**Iterator**
+
+An iterator is any object whose class has `__next__` and `__iter__` methods. An iterator can also be created with the built-in function `iter`
+
+```py
+from typing import NoReturn
+
+
+class Foo:
+    def __init__(self, max: int = 0) -> None:
+        self.max = max
+
+    def __iter__(self) -> "Foo":
+        self.n = 0
+        return self
+
+    def __next__(self) -> int | NoReturn:
+        if self.n <= self.max:
+            result = 2**self.n
+            self.n += 1
+            return result
+        else:
+            raise StopIteration
+```
+The `StopIteration` exception must be raised or the iterator is pronounced broken.
+
+**What is the difference between iterators and generators?**
+
+An iterator requires a `__next__` and `__iter__` methods over a generator that only needs an `__iter__` method.


bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,45 @@
+**Generators and Iterators**
+
+What is a generator and an iterator and what is the difference between each other?
+
+**Generator**
+
+A generator is a function that may contain a `yield` or sometimes `return` statements, a generator can also be an expression(genexpr).
```
ionite34, line 7: This seems problematic, generators don't have to be functions. You can create a generator class by implementing methods defined in `collections.abc.Generator`

bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,45 @@
...
+            self.n += 1
+            return result
+        else:
+            raise StopIteration
+```
+The `StopIteration` exception must be raised or the iterator is pronounced broken.
+
+**What is the difference between iterators and generators?**
+
+An iterator requires a `__next__` and `__iter__` methods over a generator that only needs an `__iter__` method.
```
ionite34, line 45: This is incorrect. Merely implementing `__iter__` does not create a `Generator`, you can test it as such:

```py
import typing

class Foo:
    def __iter__(self):
        return self

print(isinstance(Foo(), typing.Generator))
>> False
```

All Generators are Iterators, the abstract base class for a `Generator` actually inherits `Iterator`. `Generator` actually requires 2 methods to be implemented over an `Iterator`'s `__next__` and `__iter__`, which are `send`, `throw`, and the mix-in methods `close`

More info here: https://docs.python.org/3/library/collections.abc.html

ionite34 (changes requested):
A few notes on Generators not being bound to functions, and fixing the inheritance explanation for `Iterator -> Generator`, instead of the other way around

wookie184:
This tag feels like it's trying to *compare* generators and iterators, which I don't think makes sense. Since a generator (function) returns an iterator, this feels a bit like comparing e.g. a BMW to a Car. Would two separate tags on the topics make more sense?

swfarnsworth:
Tags should be concise and answer a common question. If someone wants to know what a generator is, at a high-enough level that it can be answered concisely, I don't think many of the details here are relevant.

I would rather see a tag that gives an example of a generator created through a function, and an explanation of how the generator maintains its own state between yields (but preferably without the word "state").

Okimii:
> This tag feels like it's trying to _compare_ generators and iterators, which I don't think makes sense. Since a generator (function) returns an iterator, this feels a bit like comparing e.g. a BMW to a Car. Would two separate tags on the topics make more sense?

Well, maybe the wording in the file may lead you to think that its trying to compare an iterator and a generator, im not aiming to, im aiming to explain both concepts, if you have any suggestions especially about the wording, please let me know!

bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,58 @@
...
+    def __next__(self) -> int | NoReturn:
+        if self.n <= self.max:
+            result = 2**self.n
+            self.n += 1
+            return result
+        else:
+            raise StopIteration
+```
+The `StopIteration` exception must be raised or the iterator is pronounced broken.
+
```
ionite34, line 53:
```suggestion
\```py
foo = (x * 2 for x in range(5))
bar = sum(x * 2 for x in range(5))
\```

Note that Generators created using expressions are not functions that return Generators, rather, they are already Generator objects. You can only iterate the contents of `foo` once before it is exhausted.
\```py
list(foo) -> [0, 2, 4, 6, 8]
list(foo) -> []
\```

You can also use `*` to unpack Generators as positional arguments for functions, like any other Iterable object.
\```py
print(*foo) -> 0 2 4 6 8
\```
```
Rationale explained in previous suggestion, here it might be a better use of content to explain some examples on generator and expression usage instead of building custom Iterator classes.

> ### Note: Please unescape the `\` marks before each of the code blocks, I can't find any other way to nest codeblocks inside the suggestion codeblock.

bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,58 @@
...
+**What is the difference between iterators and generators?**
+
+All Generators are Iterators, the abstract base class for a `Generator` actually inherits `Iterator`. `Generator` actually requires 2 methods to be implemented over an `Iterator`'s `__next__` and `__iter__`, which are `send`, `throw`, and the mix-in methods `close`, the only difference of a `Generator` and an `Iterator` are there use cases and implementation.
+
+More info here: [collections.abc](https://docs.python.org/3/library/collections.abc.html)
```
ionite34, line 58:
```suggestion
**More Information**

• Generators and Iterators can also be created as custom classes by implementing their required [methods](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes-1). The preferred way to do this is by extending the respective Abstract Base Classes within `collections.abc`.

• Check out the [Real python article](https://realpython.com/introduction-to-python-generators/) for a more in-depth view of Generators and their use-cases!
```

If explaining the difference at the start, we don't really need this ending last part, consider changing to an additional resource section.

ionite34 (changes requested):
I feel like the proposed tag would be better focused on explanations of the creation of generators and generator expressions instead of custom Iterator classes, proposed a couple of changes in line with this.

swfarnsworth:
Your understanding of Python's internals is impressive, but I feel that this tag focuses on the wrong details, and tags of this length are not likely to be read in their entirety. Consider the target audience (novice Python developers) and the context in which this is intended to be read (likely during a help session where said novice is focused mainly on solving a specific problem). As insightful as it is, I would like to see all discussion of classes and dunder methods removed.

Perhaps you could instead demonstrate how a function that returns a list of X could be refactored to generate X. Remember: for a novice Python developer, what's remarkable about generators is that they "go to sleep" between iterations and "keep all variables", or some phrasing along those lines.

bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,56 @@
...
+
+**Generator Functions**
+
+A generator function can be created by using one or more `yield` statements.
+\```py
+def foo():
+    yield 1
+    yield 2
+\```
+When called, this function returns a Generator. Code within the generator does not execute immediately when a Generator is created.
```
ionite34, line 21:
```suggestion
Generators can be a memory-efficient alternative to lists in many situations.
```

bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,56 @@
...
+list(foo) -> []
+\```
+
+You can also use `*` to unpack Generators as positional arguments for functions, like any other Iterable object.
+\```py
+print(*foo) -> 0 2 4 6 8
+\```
+**More Information**
+
+• Generators and Iterators can also be created as custom classes by implementing their required [methods](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes-1). The preferred way to do this is by extending the respective Abstract Base Classes within `collections.abc`.
```
ionite34, line 54:
```suggestion
• Generators and Iterators can also be created as custom classes by implementing their required [abstract methods](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes-1).
```

Made ABC link more concise

bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,56 @@
...
+Generator results can be accessed by iteration with `for`, or using `next()`. When calling `next()` on a Generator, the function will run until it reaches the next yield, then pause and yield the result. Further code does not execute until the next `yield` is reached. If the Generator is exhausted, a `StopIteration` Exception is raised.
+\```py
+x = foo()
+next(x) -> 1
+next(x) -> 2
+next(x) -> StopIteration
+\```
+**Generator Expressions**
+
+Generators can also be created by `for` clauses inside [expressions](https://docs.python.org/3/glossary.html#term-generator-expression). This is similar to creating a list comprehension but using `()` instead of `[]`. You can use `if` `else` ternary operators as you would in other comprehension forms.
```
ionite34, line 35:
```suggestion
Generators can also be created by `for` clauses inside [expressions](https://docs.python.org/3/glossary.html#term-generator-expression), similar to creating a list comprehension.
```
Made expression introduction more concise.

bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,56 @@
...
+foo()
+<generator object foo at 0x1>
+\```
+Generator results can be accessed by iteration with `for`, or using `next()`. When calling `next()` on a Generator, the function will run until it reaches the next yield, then pause and yield the result. Further code does not execute until the next `yield` is reached. If the Generator is exhausted, a `StopIteration` Exception is raised.
+\```py
+x = foo()
+next(x) -> 1
+next(x) -> 2
+next(x) -> StopIteration
+\```
```
ionite34, line 32:
```suggestion
```

Removed detailed discussion of generator states.

bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,56 @@
...
+**Generator Expressions**
+
+Generators can also be created by `for` clauses inside [expressions](https://docs.python.org/3/glossary.html#term-generator-expression). This is similar to creating a list comprehension but using `()` instead of `[]`. You can use `if` `else` ternary operators as you would in other comprehension forms.
+
+\```py
+foo = (x * 2 for x in range(5))
+bar = sum(x * 2 for x in range(5))
```
ionite34, line 39:
```suggestion
foo = (x * 2 for x in range(5))

for num in foo:
    print(num)
```

Instead of using sum, used print example which might be easier to understand.

bot/resources/tags/generators-iterators.md
```diff
@@ -0,0 +1,56 @@
...
+Note that Generators created using expressions are not functions that return Generators, rather, they are already Generator objects. You can only iterate the contents of `foo` once before it is exhausted.
+\```py
+list(foo) -> [0, 2, 4, 6, 8]
+list(foo) -> []
+\```
+
+You can also use `*` to unpack Generators as positional arguments for functions, like any other Iterable object.
+\```py
+print(*foo) -> 0 2 4 6 8
+\```
```
ionite34, line 51:
```suggestion
```

Removed extra details on generator states and unpacking

ionite34 (changes requested):
Reduced some details on generator internal states and classes to be more beginner-friendly.

Okimii:
Well, our target should be intermediate and beginners. I wouldnt really remove the examples as its better to have more information about such a topic so users dont need to search e.g on Google on how a to make a generator or how yield works, yes i do agree its allot of info but its better to have info in depth over explaining concepts and not their implementation.I would agree the tag is long, but @ionite34 did a good job explaining and wording it better than i did and in my eyes nothing really needs to be explained which is good as he gave allot of sources for example so users can browse or understand better about the topic of this tag.

Okimii:
@ionite34 well done especially on the implementation of examples!

ionite34:
For an additional note, the `\` marks on lines 14, 26, 32, 37 need to be removed for the markdown to be valid.

minalike:
We discussed during today's staff meeting, we would like to see this tag revamped/pivoted to a narrower scope, which also leverages and links to the Real Python article on generators. 

Please let us know if you'd be interested in doing that @Okimii.

We will also work on better defining what criteria we feel a tag should meet to best serve as a teaching tool for a real-time communication platform such as Discord, and also make it a bit clearer that a PR for a tag should not be opened without further discussion on a meta issue first.

Okimii:
@minalike I wouldn't mind revamping the tag, I would love for the staff of pydis to make it better and more defined in the context of details for a better tag for beginners, I can also help with the structure of the tag but im not quite good with wording so I would really need help on that part, either way there's allot of talented people in pydis staff, e.g @ionite34 which has helped allot especially wording wise and the tags structure, so even if my wording isn't the best it can also be corrected which I do appreciate as it does help me.

lemonsaurus:
I'd like to see this shortened down dramatically.
- I don't think we should explain iterators and generators in the same tag. Let's make this one solely about generators.
- Tags should not be articles, and generators is a big topic which will not be sufficiently explained in a tag. In this case, it would be better to explain (in only a handful of words) what it is, how it's commonly used, and then link to a good resource for learning about it. 
- The RealPython article is a fantastic resource, and right now, I see no reason why I would use this tag instead of just linking that article directly. Let's instead make this tag a companion to that article, rather than trying to compete with it.

Okimii:
I love that idea.

PR #2245, opened 2022-08-06
Update autoreview system

wookie184:
Not sure if i'm fully done yet so marking as draft, but I don't mind if anyone has any general feedback.

### DIFF
--- a/bot/exts/recruitment/talentpool/_cog.py
+++ b/bot/exts/recruitment/talentpool/_cog.py
@@ -43,7 +43,7 @@ class TalentPool(Cog, name="Talentpool"):
         await self.refresh_cache()
 
         if await self.autoreview_enabled():
-            await self.autoreview_loop.start()
+            self.autoreview_loop.start()
 
     async def autoreview_enabled(self) -> bool:
         """Return whether automatic posting of nomination reviews is enabled."""
@@ -102,7 +102,7 @@ class TalentPool(Cog, name="Talentpool"):
             return
 
         await self.talentpool_settings.set(AUTOREVIEW_ENABLED_KEY, True)
-        await self.autoreview_loop.start()
+        self.autoreview_loop.start()
         await ctx.send(":white_check_mark: Autoreview enabled.")
 
     @nomination_autoreview_group.command(name="disable", aliases=("off",))
@@ -114,7 +114,7 @@ class TalentPool(Cog, name="Talentpool"):
             return
 
         await self.talentpool_settings.set(AUTOREVIEW_ENABLED_KEY, False)
-        await self.autoreview_loop.stop()
+        self.autoreview_loop.stop()
         await ctx.send(":white_check_mark: Autoreview disabled.")
 
     @nomination_autoreview_group.command(name="status")
--- a/bot/exts/recruitment/talentpool/_cog.py
+++ b/bot/exts/recruitment/talentpool/_cog.py
@@ -6,8 +6,8 @@ from typing import Optional, Union
 import discord
 from async_rediscache import RedisCache
 from botcore.site_api import ResponseCodeError
-from botcore.utils import scheduling
 from discord import Color, Embed, Member, PartialMessage, RawReactionActionEvent, User
+from discord.ext import tasks
 from discord.ext.commands import BadArgument, Cog, Context, group, has_any_role
 
 from bot.bot import Bot
@@ -38,17 +38,12 @@ class TalentPool(Cog, name="Talentpool"):
         self.cache: Optional[defaultdict[dict]] = None
         self.api_default_params = {'active': 'true', 'ordering': '-inserted_at'}
 
-        self.initial_refresh_task = scheduling.create_task(self.refresh_cache(), event_loop=self.bot.loop)
-        scheduling.create_task(self.schedule_autoreviews(), event_loop=self.bot.loop)
+    async def cog_load(self) -> None:
+        """Load user cache and maybe start autoreview loop."""
+        await self.refresh_cache()
 
-    async def schedule_autoreviews(self) -> None:
-        """Reschedule reviews for active nominations if autoreview is enabled."""
         if await self.autoreview_enabled():
-            # Wait for a populated cache first
-            await self.initial_refresh_task
-            await self.reviewer.reschedule_reviews()
-        else:
-            log.trace("Not scheduling reviews as autoreview is disabled.")
+            await self.autoreview_loop.start()
 
     async def autoreview_enabled(self) -> bool:
         """Return whether automatic posting of nomination reviews is enabled."""
@@ -93,15 +88,21 @@ class TalentPool(Cog, name="Talentpool"):
         """
         Enable automatic posting of reviews.
 
-        This will post reviews up to one day overdue. Older nominations can be
-        manually reviewed with the `tp post_review <user_id>` command.
+        A review will be posted when the current number of active reviews is below the limit
+        and long enough has passed since the last review.
+
+        Users will be considered for review if they have been in the talent pool past a
+        threshold time.
+
+        The next user to review is chosen based on the number of nominations a user has,
+        using the age of the first nomination as a tie-breaker (oldest first).
         """
         if await self.autoreview_enabled():
             await ctx.send(":x: Autoreview is already enabled.")
             return
 
         await self.talentpool_settings.set(AUTOREVIEW_ENABLED_KEY, True)
-        await self.reviewer.reschedule_reviews()
+        await self.autoreview_loop.start()
         await ctx.send(":white_check_mark: Autoreview enabled.")
 
     @nomination_autoreview_group.command(name="disable", aliases=("off",))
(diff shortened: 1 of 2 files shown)


bot/exts/recruitment/talentpool/_review.py
```diff
@@ -44,37 +46,93 @@
...
+        Users will only be selected for review if:
+         - They have not already been reviewed.
+         - They have been nominated for longer than `MIN_NOMINATION_TIME`.
+
+        The priority of the review is determined by how many nominations the user has
+        (more nominations = higher priority).
+        For users with equal priority the oldest nomination will be reviewed first.
+        """
+        possible = []
         for user_id, user_data in self._pool.cache.items():
```
ChrisLovering, line 118: Not one for this PR, but we should really type the cached pool with some dataclasses some time, so we don't need to slice everywhere

ChrisLovering (approved):
This looks great, thanks Wookie!

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -142,7 +200,7 @@ async def make_review(self, user_id: int) -> typing.Tuple[str, Optional[Emoji],
 
         reviewed_emoji = self._random_ducky(guild)
         vote_request = (
-            "*Refer to their nomination and infraction histories for further details*.\n"
+            f"*Refer to their nomination and infraction histories for further details.*\n"
```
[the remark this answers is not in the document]
wookie184: thank you, thank you. This important change greatly reduces code size, causing code to be up to 100% of previous speed!

bot/exts/recruitment/talentpool/_review.py
wookie184: Yeah that would be nice

ChrisLovering (approved):

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -28,14 +26,18 @@
...
-# Maximum amount of days before an automatic review is posted.
-MAX_DAYS_IN_POOL = 30
-
 # Maximum amount of characters allowed in a message
 MAX_MESSAGE_SIZE = 2000
 # Maximum amount of characters allowed in an embed
 MAX_EMBED_SIZE = 4000
 
+# Maximum number of active reviews
+MAX_ONGOING_REVIEWS = 4
```
mbaruh, line 35: Are we sure we want this to be 4? 3 is the number we're currently using.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -89,31 +88,42 @@ async def nomination_autoreview_group(self, ctx: Context) -> None:
...
-        This will post reviews up to one day overdue. Older nominations can be
-        manually reviewed with the `tp post_review <user_id>` command.
+        A review will be posted when the current number of active reviews is below the limit
+        and long enough has passed since the last review.
+
+        Users will be considered for review if they have been in the talent pool past a
+        threshold time.
+
+        The next user to review is chosen based on the number of nominations a user has,
+        using the age of the first nomination as a tie-breaker (oldest first).
```
mbaruh, line 103: I wonder how good this is, since there's a potential starvation here for nominees with only one nomination. What I would do is take both parameters and give each of them a weight to create a total score. It would then review the nominee with the highest score.

Just off the top of my head, we could do something like:
number_of_nominations / number_for_currently_most_nominated  +  age / age_of_currently_oldest

This could be weighted for example to give the number of nominations more importance.

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -44,37 +46,93 @@
...
+        (more nominations = higher priority).
+        For users with equal priority the oldest nomination will be reviewed first.
+        """
+        possible = []
         for user_id, user_data in self._pool.cache.items():
-            if not user_data["reviewed"]:
-                self.schedule_review(user_id)
+            if (
+                not user_data["reviewed"]
+                and isoparse(user_data["inserted_at"]) < datetime.now(timezone.utc) - MIN_NOMINATION_TIME
```
mbaruh, line 121: It's not a mistake or anything but I think `now - inserted_at > min_time` is easier to read. Ignore if you want.

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -44,37 +46,93 @@
...
-        """Schedules a single user for review."""
-        log.trace(f"Scheduling review of user with ID {user_id}")
+        # Secondary sort key: creation of first entries on the nomination.
+        possible.sort(key=lambda x: isoparse(x[1]["inserted_at"]))
 
-        user_data = self._pool.cache.get(user_id)
-        inserted_at = isoparse(user_data['inserted_at'])
-        review_at = inserted_at + timedelta(days=MAX_DAYS_IN_POOL)
+        # Primary sort key: number of entries on the nomination.
+        user = max(possible, key=lambda x: len(x[1]["entries"]))
```
mbaruh, line 133: You could do this in one go without sorting with tuples.

mbaruh (changes requested):
A few comments, looks great otherwise.

bot/exts/recruitment/talentpool/_review.py
wookie184: I picked that off the top of my head thinking something like "3 for people to vote on, and 1 waiting to be actioned", and wanted to avoid slower reviews holding others up, but i'm happy to change it to 3. Can always increase depending on how it works out.

mbaruh: Yeah just worried the earliest review will get buried or that the team will get saturated with reviews. Thinking about it I would maybe add a cooldown time after resolving a vote too (whether there's helpering or not), so that we don't get pinged right after resolving a vote.

bot/exts/recruitment/talentpool/_cog.py
mbaruh: > If the talentpool is constantly growing, then so will the age of the oldest nomination in it. If the weighting of the next user picked is dependant on the age and number of nomination, then I don't know how you'd avoid the number of nominations required to be picked would also constantly growing

That's why I normalized the age by the age of the oldest vote. The contribution is capped at 1, and the older the oldest vote, the less of an effect its aging will have on the score.

Meaning, for the formula to give a nomination that isn't the oldest, the gap between the number of nominations needs to be large enough, but it's always capped by the largest number of nominations for any currently nominated user. If the oldest nomination has many nominations it'll be invoked, but if all the old nominations have only a few, and then a newer nomination has many, it has a chance to be invoked. If you multiple the number of nominations in the formula by say 1.5 or 2, it'll be even easier.

mbaruh: And, if a nomination is so old it still dominates the score, then maybe it's time we took a look at it?

mbaruh: The bottom line is, instead of using the age as a tie breaker, you can use it as a regularization parameter, and strike a balance between popularity and age.

mbaruh:
One more thing to consider if we're going to allow automatically starting votes again: We could query metricity to check when is the last message sent by the user. If it's too old the vote won't fare well.

wookie184:
> One more thing to consider if we're going to allow automatically starting votes again: We could query metricity to check when the last message sent by the user is. If it's too old the vote won't fare well.

Yeah, that was part of the original specification in the issue. I still plan on adding that at some point, but I wanted to try and limit the scope for this PR so didn't do it here.

bot/exts/recruitment/talentpool/_cog.py
wookie184: Alright, I'm still not convinced that it's really necessary, but i'll try something.

mbaruh: If you'd rather keep it simple we can put it aside for now. I can show you later on Discord how the scoring will look like with the current list.

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -44,37 +50,111 @@
...
+        The higher the score, the higher the priority for being put up for review should be.
+        """
+        num_entries = len(nomination["entries"])
+        entries_score = num_entries / max_entries
 
-        user_data = self._pool.cache.get(user_id)
-        inserted_at = isoparse(user_data['inserted_at'])
-        review_at = inserted_at + timedelta(days=MAX_DAYS_IN_POOL)
+        nomination_date = isoparse(nomination["inserted_at"])
+        now = datetime.now(timezone.utc)
```
mbaruh, line 154: It's not critical in practice, because of the speed of evaluation and the difference between nomination dates, but `now` should be constant across the nominations for correct behavior.

mbaruh (approved):
Looks good!

wookie184:
whoops, didn't mean to commit those constants changes 😓

bot/exts/recruitment/talentpool/_review.py
wookie184: True, fixed 👍

PR #2249, opened 2022-08-14
Globally decode binary responses from redis to strings

ChrisLovering:
The latest version of redis returns bytes by default, rather than strings.

This was in the 3.10 PR but must have been removed during a rebase conflict.


Numerlor (approved):

mbaruh (approved):

ionite34 (approved):
Appears to work in testing, fixes undecoded byte string regression.

PR #2250, opened 2022-08-15
feat: command for banning compromised accounts

dawnofmidnight:
This PR adds a command to ban compromised users. It is the same as cleanban, but it bans specifically for 4 days and uses the typical ban reason.

Approval was granted [in Discord](https://canary.discord.com/channels/267624335836053506/775412552795947058/1008254956312203314).

### DIFF
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -2,7 +2,7 @@ import asyncio
 import re
 import unicodedata
 import urllib.parse
-from datetime import timedelta
+from datetime import datetime, timedelta
 from typing import Any, Dict, List, Mapping, NamedTuple, Optional, Tuple, Union
 
 import arrow
@@ -413,7 +413,7 @@ class Filtering(Cog):
                             await context.invoke(
                                 context.command,
                                 msg.author,
-                                arrow.utcnow() + AUTO_BAN_DURATION,
+                                datetime.utcnow() + AUTO_BAN_DURATION,
                                 reason=AUTO_BAN_REASON
                             )
 
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -2,7 +2,7 @@ import asyncio
 import re
 import unicodedata
 import urllib.parse
-from datetime import datetime, timedelta
+from datetime import timedelta
 from typing import Any, Dict, List, Mapping, NamedTuple, Optional, Tuple, Union
 
 import arrow
@@ -413,7 +413,7 @@ class Filtering(Cog):
                             await context.invoke(
                                 context.command,
                                 msg.author,
-                                datetime.utcnow() + AUTO_BAN_DURATION,
+                                (arrow.utcnow() + AUTO_BAN_DURATION).datetime,
                                 reason=AUTO_BAN_REASON
                             )
 
--- a/bot/exts/moderation/infraction/infractions.py
+++ b/bot/exts/moderation/infraction/infractions.py
@@ -1,7 +1,7 @@
 import textwrap
 import typing as t
-from datetime import datetime
 
+import arrow
 import discord
 from discord import Member
 from discord.ext import commands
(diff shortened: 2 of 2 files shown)


bot/exts/moderation/infraction/infractions.py
```diff
@@ -151,6 +153,11 @@ async def send(*args, **kwargs) -> None:
         ctx.send = send
         await infr_manage_cog.infraction_append(ctx, infraction, None, reason=f"[Clean log]({log_url})")
 
+    @command()
+    async def compban(self, ctx: Context, user: UnambiguousMemberOrUser) -> None:
+        """Same as cleanban, but specifically for four days and with the ban reason used for compromised accounts."""
+        await self.cleanban(ctx, user, duration=arrow.utcnow() + AUTO_BAN_DURATION, reason=AUTO_BAN_REASON)
```
ChrisLovering, line 159: `arrow.utcnow() + AUTO_BAN_DURATION` results in an arrow object, which does not match the `Expiry` type hint.

This doesn't seem like an issue now, but may be broken in a future change, as an `arrow` object is not marked as supported.

Either we need to add `arrow` as part of the `Expiry` union, or change this to instead pass a datetime to cleanban.

bot/exts/moderation/infraction/infractions.py
```diff
@@ -151,6 +153,11 @@ async def send(*args, **kwargs) -> None:
         ctx.send = send
         await infr_manage_cog.infraction_append(ctx, infraction, None, reason=f"[Clean log]({log_url})")
 
+    @command()
+    async def compban(self, ctx: Context, user: UnambiguousMemberOrUser) -> None:
+        """Same as cleanban, but specifically for four days and with the ban reason used for compromised accounts."""
```
minalike, line 158: Tested, works well. Thank you!
```suggestion
        """Same as cleanban, but specifically for the ban reason and duration used for compromised accounts."""
```

minalike (changes requested):

bot/exts/moderation/infraction/infractions.py
dawnofmidnight: Oh, I copied that from this the following snippet. Perhaps I misunderstood what that does, I'll change that.

https://github.com/python-discord/bot/blob/e1967b58b859578b1702d873cbfa737ea3b2feaa/bot/exts/filters/filtering.py#L416

ChrisLovering: Ah, it's wrong there too then 😅. Feel free to make the same change there too if you like.

bot/exts/moderation/infraction/infractions.py
dawnofmidnight: changed in https://github.com/python-discord/bot/pull/2250/commits/a7234a824f3d4b2d052357d73d23a6e0c5cbb1b6

bot/exts/moderation/infraction/infractions.py
dawnofmidnight: fixed in https://github.com/python-discord/bot/pull/2250/commits/a7234a824f3d4b2d052357d73d23a6e0c5cbb1b6

bot/exts/filters/filtering.py
```diff
@@ -413,7 +413,7 @@ async def _filter_message(self, msg: Message, delta: Optional[int] = None) -> No
                             await context.invoke(
                                 context.command,
                                 msg.author,
-                                arrow.utcnow() + AUTO_BAN_DURATION,
+                                datetime.utcnow() + AUTO_BAN_DURATION,
```
ionite34, line 416:
```suggestion
                                arrow.utcnow().datetime + AUTO_BAN_DURATION,
```

Since the usage of `datetime.utcnow()` without a timezone is [not preferred](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow), and arrow is already imported, it might be better to still create an arrow time and pass the `datetime` property of it.

bot/exts/filters/filtering.py
```diff
@@ -2,7 +2,7 @@
 import re
 import unicodedata
 import urllib.parse
-from datetime import timedelta
+from datetime import datetime, timedelta
```
ionite34, line 5:
```suggestion
from datetime import timedelta
```
If using arrow as per other suggestion then this import would not be needed :)

ionite34 (changes requested):

bot/exts/filters/filtering.py
dawnofmidnight: https://github.com/python-discord/bot/pull/2250/commits/4eaecb32830aa122cd53656de562caf20b79e33d

bot/exts/filters/filtering.py
dawnofmidnight: fixed in https://github.com/python-discord/bot/pull/2250/commits/4eaecb32830aa122cd53656de562caf20b79e33d

ionite34 (approved):
All looks good!

wookie184 (approved):
Works nicely, thanks!

PR #2252, opened 2022-08-15
Correctly initialise redis tests

ChrisLovering:
Calling the cog_load from within the setUp function resulted in interaction with a RedisSession before it was initialised. This wasn't noticed in CI as it only error under certain concurrency timings due to xdist.

To resolve this, we moved the setup and async setup logic to a base class.


HassanAbouelela (approved):
Works for me

ionite34 (approved):
Appears to fix the current xdist parallel test errors, tested with some worker combinations on x64 Windows and ARM macOS

Issue #1565, opened 2021-05-05
Add a reason argument to `!unban` 


Akarys42:
It is pretty common for us to run the `!unban` command with a reason at the end, that is currently just silently ignored, causing a loss of information on the long term. The command should accept this argument and append `**Unbanned:** {reason}` to the end of the active ban infraction.

janine9vn:
This is actually a duplicate issue! #718

Akarys42:
> This is actually a duplicate issue! #718

#718 is about adding another field to the infraction for the unban reason, but it will have to wait until the api rewrite. This issue propose a temporary solution of simply adding the unban reason at the end of the infraction reason.

ToxicKidz:
I would like to implement this

Xithrius:
@ToxicKidz What's your status?

minalike:
Would it be possible to also add the unbanned actor and timestamp?

VirdanTheBurden:
It's alright if I take this issue, yes?

VirdanTheBurden:
I'm just taking it. Sorry, I really wanna fix it!

ChrisLovering:
If we are going to be storing the unban actor and timestamp, those should be added as fields against the infraction in site.

PR #2259, opened 2022-08-17
feat: add reason argument to pardon commands

VirdanTheBurden:
Closes issue #1565.

Due to the nature of the bot internals, it would be messy to only change the unban command to take a reason argument. Therefore, all pardon commands that call the method `pardon_infraction()` provided by the `InfractionScheduler` cog will have a reason argument and append the pardon reason to the infraction through `InfractionScheduler.deactivate_infraction()`'s HTTP patch request to the moderation API when the infraction is marked as non-active.

If anyone would like to see a demo of the pardon at work, here's a [video](https://youtu.be/i6iB3nTx-FQ) I've uploaded to YouTube.

The format of the pardoned text is still up for debate, as what I've done was a placeholder.

### DIFF
--- a/bot/exts/moderation/infraction/_scheduler.py
+++ b/bot/exts/moderation/infraction/_scheduler.py
@@ -444,7 +444,7 @@ class InfractionScheduler:
             # Mark infraction as inactive in the database.
             log.trace(f"Marking infraction #{id_} as inactive in the database.")
 
-            data = {"active": False, "reason": ""}
+            data = {"active": False}
 
             if pardon_reason is not None:
                 # Append pardon reason to infraction in database.
--- a/bot/exts/moderation/infraction/_scheduler.py
+++ b/bot/exts/moderation/infraction/_scheduler.py
@@ -447,6 +447,7 @@ class InfractionScheduler:
             data = {"active": False}
 
             if pardon_reason is not None:
+                data["reason"] = ""
                 # Append pardon reason to infraction in database.
                 if (punish_reason := infraction["reason"]) is not None:
                     data["reason"] = punish_reason + " | "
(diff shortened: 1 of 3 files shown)


VirdanTheBurden:
Oops, it seems that I've forgotten about tests - working on that now!

bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -434,9 +443,19 @@ async def deactivate_infraction(
...
+                if (punish_reason := infraction["reason"]) is not None:
+                    data["reason"] = punish_reason + " | "
+
+                data["reason"] += f"Pardoned: {pardon_reason}"
+
             await self.bot.api_client.patch(
                 f"bot/infractions/{id_}",
-                json={"active": False}
+                json=data
             )
```
wookie184, line 459: Hmm, now if `pardoned_reason` is none it'll try to update `reason` to be an empty string.

wookie184 (changes requested):

bot/exts/moderation/infraction/_scheduler.py
VirdanTheBurden: Fixed - removed the "reason" field from data so it'll just have `{"active": False}` if `pardon_reason` is `None`

wookie184: Now it'll error if `data["reason"] += f"Pardoned: {pardon_reason}"` is hit without reason being set first (if the reason is currently `None`) 😅 

I think setting `data["reason"] = ""` first in `if pardon_reason is not None:` would fix it.

VirdanTheBurden: Added.

wookie184 (approved):
Works nicely, thanks!

ChrisLovering (approved):

Issue #2108, opened 2022-03-01
Allow invoking rule command with keywords rather than rule number


RyugaXhypeR:
maybe link some keywords to rules.. like the rule-6 is related to advertising, so maybe we can just type !rule ad.

shtlrs:
Would it be best to have this implemented on a keyword basis? Meaning that we have a list of particular keywords that map to a particular rule?
Or would it be better to use NLP here in order to try to identify the rule that a particular user is trying to mention, and then maybe give suggestions ?

wookie184:
> Would it be best to have this implemented on a keyword basis? Meaning that we have a list of particular keywords that map to a particular rule?
> Or would it be better to use NLP here in order to try to identify the rule that a particular user is trying to mention, and then maybe give suggestions ?

I think having keywords would be best, it should be simpler, as well as being more predictable and giving us more control.

With that we would need some way of discovering the keywords that work for each rule. We could include it in each rule invocation, or have a separate command that lists them.

ChrisLovering:
IMO regardless of what we do, we should keep the ability to reference based on numbers, on top of whatever we do here.

The rule command has existed for quite a while that a lot of staff and regulars are very used to be able to invoke based on an index.

In terms of implementation, this is trivial, as we would just need to leave the greedy int convertor, and add an optional string param.

shtlrs:
@ChrisLovering I don't think the point was to change the fundamentals of how the command works. Listing a rule by its number would remain the same.
What we're trying to do is to make it slightly easier for people who are unacquainted with the rule numbers to find them easily.

@wookie184 I definitely agree with you on that. Because we don't want it to get too clunky and get lost in finding keywords as well.

I'll try to spin up a draft PR for this and discuss implementation details there when needed.

mbaruh:
One aspect that is lost here is that the current rules command is dynamic. The rules are fetched from the API. The suggested approach is probably ok, but will require hardcoding keywords, which will need to be changed/rearranged if the rules ever change. How much of an issue would be to take a keyword and check whether it appears only in a particular rule?

wookie184:
> How much of an issue would it be to take a keyword and check whether it appears only in a particular rule? In terms of getting undesirable results

Not so much undesirable results, but it would stop us having stuff like `!rule tos` if the rule just contains "terms of service", or `!rule advert` if the rule contains `advertisment`.

> The rules are fetched from the API

[...]

PR #2261, opened 2022-08-19
Add support to fetch rules via keywords

shtlrs:
⚠️  Note to core devs: This must be merged with https://github.com/python-discord/site/pull/770 ⚠️

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -774,9 +774,3 @@ ERROR_REPLIES = [
     "Noooooo!!",
     "I can't believe you've done this",
 ]
-
-DEFAULT_RULES_DESCRIPTION = (
-    "The rules and guidelines that apply to this community can be found on"
-    " our [rules page](https://www.pythondiscord.com/pages/rules). We expect"
-    " all members of the community to have read and understood these."
-)
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -774,3 +774,9 @@ ERROR_REPLIES = [
     "Noooooo!!",
     "I can't believe you've done this",
 ]
+
+DEFAULT_RULES_DESCRIPTION = (
+    "The rules and guidelines that apply to this community can be found on"
+    " our [rules page](https://www.pythondiscord.com/pages/rules). We expect"
+    " all members of the community to have read and understood these."
+)
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -397,6 +397,7 @@ class Categories(metaclass=YAMLGetter):
     # 2021 Summer Code Jam
     summer_code_jam: int
 
+
 class Channels(metaclass=YAMLGetter):
     section = "guild"
     subsection = "channels"
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -526,15 +526,15 @@ class Information(Cog):
         Rule numbers and keywords can be sent in any order.
         """
         rules_embed = Embed(title="Rules", color=Colour.og_blurple(), url="https://www.pythondiscord.com/pages/rules")
-        keywords, rules = [], []
+        keywords, rule_numbers = [], []
 
         for word in args:
             try:
-                rules.append(int(word))
+                rule_numbers.append(int(word))
             except ValueError:
                 keywords.append(word.lower())
 
-        if not rules and not keywords:
+        if not rule_numbers and not keywords:
             # Neither rules nor keywords were submitted. Return the default description.
             rules_embed.description = (
                 "The rules and guidelines that apply to this community can be found on"
@@ -548,11 +548,11 @@ class Information(Cog):
         full_rules = await self.bot.api_client.get("rules", params={"link_format": "md"})
 
         # Remove duplicates and sort the rule indices
-        rules = sorted(set(rules))
+        rule_numbers = sorted(set(rule_numbers))
         # Remove duplicate keywords
         keywords = set(keywords)
 
-        invalid = ", ".join(str(index) for index in rules if index < 1 or index > len(full_rules))
+        invalid = ", ".join(str(index) for index in rule_numbers if index < 1 or index > len(full_rules))
 
         if invalid and not keywords:
             await ctx.send(shorten(":x: Invalid rule indices: " + invalid, 75, placeholder=" ..."))
@@ -560,7 +560,7 @@ class Information(Cog):
 
         final_rules = set()
 
-        for pick in rules:
+        for pick in rule_numbers:
             self.bot.stats.incr(f"rule_uses.{pick}")
             final_rules.add(
                 f"**{pick}.** {full_rules[pick - 1][0]}\n\n"
@@ -578,7 +578,7 @@ class Information(Cog):
                     f"You can also invoke this rule with the following keywords: "
                     f"{', '.join(full_rules[pick][1])}")
 
-        if not rules and not final_rules:
+        if not rule_numbers and not final_rules:
             # This would mean that only keywords where used and no match for them was found
             await ctx.send(
                 f"There are currently no rules that correspond to keywords: **{', '.join(keywords)}**.\n"
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -529,16 +529,17 @@ class Information(Cog):
         keywords, rule_numbers = [], []
 
         full_rules = await self.bot.api_client.get("rules", params={"link_format": "md"})
-        available_keywords = set()
+        keyword_to_rule_number = dict()
 
-        for _, rule_keywords in full_rules:
-            available_keywords = available_keywords | set(rule_keywords)
+        for rule_number, (_, rule_keywords) in enumerate(full_rules, start=1):
+            for rule_keyword in rule_keywords:
+                keyword_to_rule_number[rule_keyword] = rule_number
 
         for word in args:
             try:
                 rule_numbers.append(int(word))
             except ValueError:
-                if (kw := word.lower()) not in available_keywords:
+                if (kw := word.lower()) not in keyword_to_rule_number:
                     break
                 keywords.append(kw)
 
@@ -568,10 +569,7 @@ class Information(Cog):
         final_rule_numbers.extend(rule_numbers)
 
         for keyword in keywords:
-            for rule_number, rule in enumerate(full_rules, start=1):
-                if keyword in rule[1]:
-                    final_rule_numbers.append(rule_number)
-                    break
+            final_rule_numbers.append(keyword_to_rule_number.get(keyword))
 
         final_rules = []
         final_rule_numbers = set(final_rule_numbers)
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -558,7 +558,9 @@ class Information(Cog):
         # Remove duplicate keywords and preserve the order of initial keywords
         keywords = list(dict.fromkeys(keywords))
 
-        invalid = ", ".join(str(index) for index in rule_numbers if index < 1 or index > len(full_rules))
+        invalid = ", ".join(
+            str(rule_number) for rule_number in rule_numbers
+            if rule_number < 1 or rule_number > len(full_rules))
 
         if invalid:
             await ctx.send(shorten(":x: Invalid rule indices: " + invalid, 75, placeholder=" ..."))
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -560,12 +560,9 @@ class Information(Cog):
 
         if not rules and not final_rules:
             # This would mean that we didn't find a match for the used keyword
-            rules_embed.description = (
+            await ctx.send(
                 f"There are currently no rules that correspond to keyword: {keyword}."
-                " If you think it should be added, please ask our admins and they'll take care of the rest."
-            )
-
-            await ctx.send(embed=rules_embed)
+                "If you think it should be added, please ask our admins and they'll take care of the rest.")
             return
 
         await LinePaginator.paginate(final_rules, ctx, rules_embed, max_lines=3)
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -19,7 +19,7 @@ from bot.errors import NonExistentRoleError
 from bot.log import get_logger
 from bot.pagination import LinePaginator
 from bot.utils import time
-from bot.utils.channel import is_mod_channel, is_staff_channel
+from bot.utils.channel import get_or_fetch_channel, is_mod_channel, is_staff_channel
 from bot.utils.checks import cooldown_with_role_bypass, has_no_roles_check, in_whitelist_check
 from bot.utils.members import get_or_fetch_member
 
@@ -580,9 +580,12 @@ class Information(Cog):
 
         if not rules and not final_rules:
             # This would mean that only keywords where used and no match for them was found
+            dev_contrib_channel = await get_or_fetch_channel(constants.Channels.dev_contrib)
+            meta_channel = await get_or_fetch_channel(constants.Channels.meta)
             await ctx.send(
-                f"There are currently no rules that correspond to keywords: {[', '.join(keywords)]}."
-                "If you think it should be added, please ask our admins and they'll take care of the rest.")
+                f"There are currently no rules that correspond to keywords: **{', '.join(keywords)}**.\n"
+                f"If you think it should be added, please suggest it in either "
+                f"{meta_channel.mention} or {dev_contrib_channel.mention}")
             return
 
         await LinePaginator.paginate(final_rules, ctx, rules_embed, max_lines=3)
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -8,7 +8,7 @@ from typing import Any, DefaultDict, Mapping, Optional, Tuple, Union
 import rapidfuzz
 from botcore.site_api import ResponseCodeError
 from discord import AllowedMentions, Colour, Embed, Guild, Message, Role
-from discord.ext.commands import BucketType, Cog, Context, Greedy, Paginator, command, group, has_any_role
+from discord.ext.commands import BucketType, Cog, Context, Paginator, command, group, has_any_role
 from discord.utils import escape_markdown
 
 from bot import constants
@@ -518,12 +518,23 @@ class Information(Cog):
         await self.send_raw_content(ctx, message, json=True)
 
     @command(aliases=("rule",))
-    async def rules(self, ctx: Context, rules: Greedy[int], keyword: Optional[str]) -> None:
-        """Provides a link to all rules or, if specified, displays specific rule(s)."""
-        rules_embed = Embed(title="Rules", color=Colour.og_blurple(), url="https://www.pythondiscord.com/pages/rules")
-        keyword = keyword.lower()
+    async def rules(self, ctx: Context, *args: Optional[str]) -> None:
+        """
+        Provides a link to all rules or, if specified, displays specific rules(s).
 
-        if not rules and not keyword:
+        It accepts either rule numbers or particular keywords that map to a particular rule.
+        Rule numbers and keywords can be sent in any order.
+        """
+        rules_embed = Embed(title="Rules", color=Colour.og_blurple(), url="https://www.pythondiscord.com/pages/rules")
+        keywords, rules = [], []
+
+        for word in args:
+            if word.isdigit():
+                rules.append(int(word))
+            else:
+                keywords.append(word.lower())
+
+        if not rules and not keywords:
             # Neither rules nor keywords were submitted. Return the default description.
             rules_embed.description = (
                 "The rules and guidelines that apply to this community can be found on"
@@ -538,30 +549,39 @@ class Information(Cog):
 
         # Remove duplicates and sort the rule indices
         rules = sorted(set(rules))
+        # Remove duplicate keywords
+        keywords = set(keywords)
 
         invalid = ", ".join(str(index) for index in rules if index < 1 or index > len(full_rules))
 
-        if invalid:
+        if invalid and not keywords:
             await ctx.send(shorten(":x: Invalid rule indices: " + invalid, 75, placeholder=" ..."))
             return
 
-        for rule in rules:
-            self.bot.stats.incr(f"rule_uses.{rule}")
+        final_rules = set()
 
-        if rules:
-            final_rules = tuple(f"**{pick}.** {full_rules[pick - 1][0]}\n\n"
-                                f"You can also invoke this rule with the following keywords: "
-                                f"{', '.join(full_rules[pick -1][1])}" for pick in rules)
-        else:
-            final_rules = tuple(f"**{pick + 1}** {full_rules[pick][0]}\n\n"
-                                f"You can also invoke this rule with the following keywords: "
-                                f"{', '.join(full_rules[pick][1])}" for pick, rule in enumerate(full_rules)
-                                if keyword in rule[1])
+        for pick in rules:
+            self.bot.stats.incr(f"rule_uses.{pick}")
+            final_rules.add(
+                f"**{pick}.** {full_rules[pick - 1][0]}\n\n"
+                f"You can also invoke this rule with the following keywords: "
+                f"{', '.join(full_rules[pick -1][1])}")
+
+        for keyword in keywords:
+            for pick, rule in enumerate(full_rules):
+                if keyword not in rule[1]:
+                    continue
+
+                self.bot.stats.incr(f"rule_uses.{pick + 1}")
+                final_rules.add(
+                    f"**{pick + 1}.** {full_rules[pick][0]}\n\n"
+                    f"You can also invoke this rule with the following keywords: "
+                    f"{', '.join(full_rules[pick][1])}")
 
         if not rules and not final_rules:
-            # This would mean that we didn't find a match for the used keyword
+            # This would mean that only keywords where used and no match for them was found
             await ctx.send(
-                f"There are currently no rules that correspond to keyword: {keyword}."
+                f"There are currently no rules that correspond to keywords: {[', '.join(keywords)]}."
                 "If you think it should be added, please ask our admins and they'll take care of the rest.")
             return
 
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -3,7 +3,7 @@ import pprint
 import textwrap
 from collections import defaultdict
 from textwrap import shorten
-from typing import Any, DefaultDict, List, Mapping, Optional, Tuple, Union
+from typing import Any, DefaultDict, Mapping, Optional, Set, Tuple, Union
 
 import rapidfuzz
 from botcore.site_api import ResponseCodeError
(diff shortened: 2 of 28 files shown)


shtlrs:
This PR aims to add support to [this](https://github.com/python-discord/bot/issues/2108) feature.

It also depends on [this](https://github.com/python-discord/site/pull/770) PR.

bot/exts/info/information.py
```diff
@@ -518,12 +518,12 @@ async def json(self, ctx: Context, message: Message) -> None:
         await self.send_raw_content(ctx, message, json=True)
 
     @command(aliases=("rule",))
-    async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
+    async def rules(self, ctx: Context, rules: Greedy[int], keyword: Optional[str]) -> None:
         """Provides a link to all rules or, if specified, displays specific rule(s)."""
         rules_embed = Embed(title="Rules", color=Colour.og_blurple(), url="https://www.pythondiscord.com/pages/rules")
```
shtlrs, line 524: I will make the keyword lowercase to make this rule case insensitive

bot/exts/info/information.py
```diff
@@ -518,12 +518,12 @@ async def json(self, ctx: Context, message: Message) -> None:
         await self.send_raw_content(ctx, message, json=True)
 
     @command(aliases=("rule",))
-    async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
+    async def rules(self, ctx: Context, rules: Greedy[int], keyword: Optional[str]) -> None:
```
MarkKoz, line 521: Because the command currently supports multiple rule integers, I believe it would be intuitive to also support multiple keywords. Ideally, something like `!rules 1 keyword 2 keyword2` would work. However, the current command allows invocations like `!rules 1 2 3 some text here`, and it ignores the "some text here". I'm not sure if this could still be supported given my above proposal.

Need to gather more opinions on what the interface should be.

bot/exts/info/information.py
```diff
@@ -547,7 +547,21 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
+            final_rules = tuple(f"**{pick}.** {full_rules[pick - 1][0]}" for pick in rules)
+        else:
+            final_rules = tuple(f"**{pick + 1}** {full_rules[pick][0]}" for pick, rule in enumerate(full_rules)
+                                if keyword in rule[1])
+
+        if not rules and not final_rules:
+            # This would mean that we didn't find a match for the used keyword
+            rules_embed.description = (
+                f"There are currently no rules that correspond to keyword: {keyword}."
+                " If you think it should be added, please ask our admins and they'll take care of the rest."
```
MarkKoz, line 560: This seems too trivial of a matter to require admin attention. I think it should direct users to one of `#dev-contrib`, `#meta`, or creating an issue on GH. @python-discord/admins thoughts?

bot/exts/info/information.py
```diff
@@ -547,7 +547,21 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
+            )
+
+            await ctx.send(embed=rules_embed)
```
MarkKoz, line 563: For consistency, this error should not be sent as an embed because the invalid indices message doesn't use one:

https://github.com/python-discord/bot/blob/ddd829fc235b0484064fce0c4924e980659ec2bb/bot/exts/info/information.py#L543-L545

bot/exts/info/information.py
```diff
@@ -518,12 +518,12 @@ async def json(self, ctx: Context, message: Message) -> None:
         await self.send_raw_content(ctx, message, json=True)
 
     @command(aliases=("rule",))
-    async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
+    async def rules(self, ctx: Context, rules: Greedy[int], keyword: Optional[str]) -> None:
         """Provides a link to all rules or, if specified, displays specific rule(s)."""
```
MarkKoz, line 522: Once we finalise the desired interface for this command, please update the docstring to mention that keywords can be used. Keep in mind this docstring appears in the help information for the command.

bot/exts/info/information.py
```diff
@@ -547,7 +547,21 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
         for rule in rules:
             self.bot.stats.incr(f"rule_uses.{rule}")
 
-        final_rules = tuple(f"**{pick}.** {full_rules[pick - 1]}" for pick in rules)
+        if rules:
+            final_rules = tuple(f"**{pick}.** {full_rules[pick - 1][0]}" for pick in rules)
+        else:
+            final_rules = tuple(f"**{pick + 1}** {full_rules[pick][0]}" for pick, rule in enumerate(full_rules)
+                                if keyword in rule[1])
```
MarkKoz, line 554: If both integers and keywords are supplied, I think it should search for both. What you wrote ignores keywords if an integer is supplied.

MarkKoz (changes requested):

bot/exts/info/information.py
shtlrs: @MarkKoz Yes I agree that it's not intuitive, but it's also not intuitive IMO to invoke it like this `!rules 1 kw1 2 kw2` and also the greedy pattern matching won't work here.

As for when it comes to detecting multiple keywords, we could do that sure, so let's just gather opinions on the interface like you said.

bot/exts/info/information.py
shtlrs: I agree, I just wanted to put it there as placeholder at first, and the response if it was fitting. But definitely too trivial of a matter.

bot/exts/info/information.py
shtlrs: @MarkKoz  Yes, true. And that was done on purpose after the discussion that @wookie184, others and I have had.
You can read that in issue #2108. 
The purpose is to be able to fetch rules on keywords as a secondary option, but if someone wants to use rule numbers, then it gets processed as a priority. Otherwise it can either conflict or confuse the person to use both.

bot/exts/info/information.py
shtlrs: Noted, thanks !

bot/exts/info/information.py
MarkKoz: Perhaps `!rules 1 kw1 2 kw2` was a poor choice of an example. What I mean is that the keywords and integers should be accepted in any order/grouping. So `kw 1 2 kw 3 kw kw kw 4 5 6 7 kw kw kw 8` would work. Hopefully that's a better example.

bot/exts/info/information.py
MarkKoz: I'm not strongly against it as is. I don't think using both is inherently confusing, but it is likely a niche use case. It may become confusing only because we're trying to allow arbitrary text after the command and it may pick up a word as a keyword that was not intended to be a keyword.

bot/exts/info/information.py
shtlrs: I'm not sure if that's one of the things that we wanted to do, e.g. drop the priority.
I think that having a pre-known order is also a good thing.

But of course, this just a personal opinion, we could collect more feedback about this to agree on the interface.

bot/exts/info/information.py
ChrisLovering: Yea, I think raising it in #community-meta makes sense here. Allows staff that don't watch #dev-contrib to give opinions too.

bot/exts/info/information.py
shtlrs: @MarkKoz After some thought, I figured that it wouldn't hurt to have them in random order & changed the implementation.
Can you have a look now please ?

bot/exts/info/information.py
shtlrs: Done.

bot/exts/info/information.py
shtlrs: Here's a preview
image

bot/exts/info/information.py
```diff
@@ -537,17 +549,44 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
+                self.bot.stats.incr(f"rule_uses.{pick + 1}")
+                final_rules.add(
+                    f"**{pick + 1}.** {full_rules[pick][0]}\n\n"
+                    f"You can also invoke this rule with the following keywords: "
+                    f"{', '.join(full_rules[pick][1])}")
+
+        if not rules and not final_rules:
+            # This would mean that only keywords where used and no match for them was found
+            dev_contrib_channel = await get_or_fetch_channel(constants.Channels.dev_contrib)
+            meta_channel = await get_or_fetch_channel(constants.Channels.meta)
```
MarkKoz, line 584: You could do `f"....... <#{constants.Channels.meta}> ....."` instead to avoid importing the function (and a potential API call if it's not in the cache for some reason).

bot/exts/info/information.py
```diff
@@ -537,17 +549,44 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
+        for keyword in keywords:
+            for pick, rule in enumerate(full_rules):
+                if keyword not in rule[1]:
+                    continue
+
+                self.bot.stats.incr(f"rule_uses.{pick + 1}")
+                final_rules.add(
+                    f"**{pick + 1}.** {full_rules[pick][0]}\n\n"
+                    f"You can also invoke this rule with the following keywords: "
+                    f"{', '.join(full_rules[pick][1])}")
```
MarkKoz, line 579: I haven't seen what this looks like, but I imagine it would add a lot of clutter to the output. I don't think it's necessary. Anyone have thoughts? If you could share a screenshot of the output, that would also help.

bot/exts/info/information.py
```diff
@@ -537,17 +549,44 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
+            self.bot.stats.incr(f"rule_uses.{pick}")
+            final_rules.add(
+                f"**{pick}.** {full_rules[pick - 1][0]}\n\n"
+                f"You can also invoke this rule with the following keywords: "
+                f"{', '.join(full_rules[pick -1][1])}")
+
+        for keyword in keywords:
+            for pick, rule in enumerate(full_rules):
+                if keyword not in rule[1]:
+                    continue
```
MarkKoz, line 573: As I stated in a previous comment, I do have concerns about supporting multiple keywords and also allowing users to write sentences after the command. To avoid unintentional interpretation of words as keywords, I think this argument parser should stop when it encounters an invalid keyword. I believe this will require the keywords to not be converted into a set.

bot/exts/info/information.py
```diff
@@ -537,17 +549,44 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
 
         # Remove duplicates and sort the rule indices
         rules = sorted(set(rules))
```
MarkKoz, line 551: Do you think renaming this to "indices", "numbers", or something similar would make the distinction between this and the keywords more clear?

MarkKoz, line 551: Sorting should happen after all rules, including those from keywords, are found.

bot/exts/info/information.py
```diff
@@ -518,12 +518,24 @@ async def json(self, ctx: Context, message: Message) -> None:
         await self.send_raw_content(ctx, message, json=True)
 
     @command(aliases=("rule",))
-    async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
-        """Provides a link to all rules or, if specified, displays specific rule(s)."""
+    async def rules(self, ctx: Context, *args: Optional[str]) -> None:
+        """
+        Provides a link to all rules or, if specified, displays specific rules(s).
```
MarkKoz, line 523:
```suggestion
        Provides a link to all rules or, if specified, displays specific rule(s).
```

bot/exts/info/information.py
```diff
@@ -518,12 +518,24 @@ async def json(self, ctx: Context, message: Message) -> None:
...
+        Rule numbers and keywords can be sent in any order.
+        """
         rules_embed = Embed(title="Rules", color=Colour.og_blurple(), url="https://www.pythondiscord.com/pages/rules")
+        keywords, rules = [], []
 
-        if not rules:
-            # Rules were not submitted. Return the default description.
+        for word in args:
+            if word.isdigit():
+                rules.append(int(word))
```
MarkKoz, line 533: `isdigit` is not appropriate because it also considers a bunch of Unicode characters valid, but `int()` will not work with them. The best way is to use `int()` within a try-except catching the `ValueError`.

MarkKoz (changes requested):

wookie184 (changes requested):
There are a couple of cases that have come up that I didn't consider in my original proposed specification https://github.com/python-discord/bot/issues/2108#issuecomment-1219378065. Here are some updated cases, and how I think they should work (all still open to discussion).

`!rule 1 coc 2` - Show rule 1 and 2 (coc is rule 1, shouldn't repeat it)
`!rule 1 hello 2 coc` - Show rule 1 (only)
`!rule 999 coc` - Error about invalid indices
`!rule coc 999` - Error about invalid indices
`!rule hello 999` - Show default rules message.
`!rule 999 coc 998 hello 997` - Could say that 999 and 998 are both invalid indices, although may be easier to just say that `999` was invalid.
`!rule coc hello 999` - Show coc rule (only).
`!rule hello 1 2 coc` - Show default rules message.

bot/exts/info/information.py
wookie184: Allowing keywords anywhere is an interesting idea, as it would allow things like `!rule that breaks the tos of xyz`. That said, I think I agree and prefer the more conservative way of stopping after an invalid keyword, I could possibly be convinced otherwise though.

shtlrs:
> There are a couple of cases that have come up that I didn't consider in my original proposed specification [#2108 (comment)](https://github.com/python-discord/bot/issues/2108#issuecomment-1219378065). Here are some updated cases, and how I think they should work (all still open to discussion).
> 
> `!rule 1 coc 2` - Show rule 1 and 2 (coc is rule 1, shouldn't repeat it) `!rule 1 hello 2 coc` - Show rule 1 (only) `!rule 999 coc` - Error about invalid indices `!rule coc 999` - Error about invalid indices `!rule hello 999` - Show default rules message. `!rule 999 coc 998 hello 997` - Could say that 999 and 998 are both invalid indices, although may be easier to just say that `999` was invalid. `!rule coc hello 999` - Show coc rule (only). `!rule hello 1 2 coc` - Show default rules message.

Here's how I see it:
!rule 1 coc 2 -> behaves just like you said
!rule 1 hello 2 coc -> Ignores the `hello` keyword
!rule 999 coc -> Do we really want to be fault intolerant here and show the Error about invalid indices whereas we could match the `coc` kw ?
!rule coc 999 -> Same comment as the one just before
!rule hello 999 -> Don't you think that by showing the default rules here, it would conflict with how you imagined we handle the `!rule coc 999` case ? 
!rule 999 coc 998 hello 997 -> I guess this falls back to how fault tolerant we want to be.
!rule coc hello 999 -> I totally agree with you
!rule hello 1 2 coc -> I think this should display rules 1 and 2 ( since coc maps to rule 1 )

To resume, here's how I imagined it:

- Always truncate the number of keywords we would process to a value `n`
- If there are only rule numbers, display the corresponding ones ( and if there are invalid indices, we ignore them)
- If all the indices are wrong, we display the invalid indices message
- If there are valid rule numbers & valid keywords, display them all ( and avoid duplicates of course)
- If there valid rule numbers, invalid ones, valid keywords and invalid ones, we only send out the valid ones.
- If both indices and keywords are invalid, we show a message where we say that neither the indices nor the keywords are valid.

wookie184:
> Always truncate the number of keywords we would process to a value `n`

Could you clarify what you mean by this?

Are you saying you think `!rule hello coc` should match `coc` but `!rule long message without any keywords but then coc` shouldn't? If so, I think that whatever value for `n` we pick would be quite arbitrary, and make the behaviour somewhat confusing and unpredictable.

shtlrs:
> > Always truncate the number of keywords we would process to a value `n`
> 
> Could you clarify what you mean by this?
> 
> Are you saying you think `!rule hello coc` should match `coc` but `!rule long message without any keywords but then coc` shouldn't? If so, I think that whatever value for `n` we pick would be quite arbitrary, and make the behaviour somewhat confusing and unpredictable.

Humm you're right, the behavior would be somewhere unpredictable based on the number `n` we choose.
But just out of a curiosity, why are we not considering going through them all ? 
Is it because we're considering the case where someone would type a lot of text ?

wookie184:
There are a couple of reasons I can for only matching at the start:
- Acts more like other commands, in that any text after the command invocation is ignored.
- Easier to understand how the command works (e.g. someone wouldn't have to guess which word in someones message is causing the rule to appear)
- Avoids false positives (e.g. `!rule coc While this isn't an advert, it *does* break our code of conduct` won't match `advert`)
- If we match through the whole message people would probably expect more words to work (e.g. plurals, `!rule don't post adverts here`).

Matching throughout has the advantage of letting people include the keyword in their sentence (e.g. `!rule please stop posting that advert`), although I don't think that outweighs the above reason when you can just do e.g. `!rule ad Please stop posting that`

shtlrs:
> There are a couple of cases that have come up that I didn't consider in my original proposed specification [#2108 (comment)](https://github.com/python-discord/bot/issues/2108#issuecomment-1219378065). Here are some updated cases, and how I think they should work (all still open to discussion).
> 
> `!rule 1 coc 2` - Show rule 1 and 2 (coc is rule 1, shouldn't repeat it) `!rule 1 hello 2 coc` - Show rule 1 (only) `!rule 999 coc` - Error about invalid indices `!rule coc 999` - Error about invalid indices `!rule hello 999` - Show default rules message. `!rule 999 coc 998 hello 997` - Could say that 999 and 998 are both invalid indices, although may be easier to just say that `999` was invalid. `!rule coc hello 999` - Show coc rule (only). `!rule hello 1 2 coc` - Show default rules message.

This is done.

bot/exts/info/information.py
```diff
@@ -533,21 +553,43 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
+                    final_rule_numbers.append(pick + 1)
+                    break
+
+        if not rule_numbers and not final_rule_numbers:
+            # This would mean that only keywords where used and no match for them was found
+            await ctx.send(
+                f"There are currently no rules that correspond to keywords: **{', '.join(keywords)}**.\n"
+                f"If you think it should be added, please suggest it in either "
+                f"<#{constants.Channels.meta}> or <#{constants.Channels.dev_contrib}>")
+            return
```
wookie184, line 584: I don't think this code is reachable, in which case it can be removed.

bot/exts/info/information.py
```diff
@@ -533,21 +553,43 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
+
+        final_rules = []
+        final_rule_numbers.sort()
+        final_rule_numbers = set(final_rule_numbers)
```
wookie184, line 588: sets are not ordered, so you should remove duplicates and *then* sort.

(in practice they will most likely end up ordered anyway because of how integers are hashed, although this isn't something to rely on)

bot/exts/info/information.py
```diff
@@ -533,21 +553,43 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
-        for rule in rules:
-            self.bot.stats.incr(f"rule_uses.{rule}")
+        final_rule_numbers = []
+        final_rule_numbers.extend(rule_numbers)
+
+        for keyword in keywords:
+            for pick, rule in enumerate(full_rules):
+                if keyword in rule[1]:
+                    final_rule_numbers.append(pick + 1)
+                    break
```
wookie184, line 576: I think it's slightly neater here to use the `start` argument of enumerate, and I think `rule_number` is a more descriptive name than `pick`.
```suggestion
        for keyword in keywords:
            for rule_number, rule in enumerate(full_rules, 1):
                if keyword in rule[1]:
                    final_rule_numbers.append(rule_number)
                    break
```

Where you create `available_keywords` above you could instead consider making a `keyword_to_rule_number` dictionary, which you could still use `in` on as you do for `available_keywords`, but would also allow you simplify this a little.
```python
for keyword in keywords:
    if rule_number := keyword_to_rule_number.get(keyword):
        final_rule_numbers.append(rule_number)
```
Not a requirement though

bot/exts/info/information.py
```diff
@@ -533,21 +553,43 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
             await ctx.send(embed=rules_embed)
             return
 
-        full_rules = await self.bot.api_client.get("rules", params={"link_format": "md"})
-
         # Remove duplicates and sort the rule indices
-        rules = sorted(set(rules))
+        rule_numbers = sorted(set(rule_numbers))
+        # Remove duplicate keywords and preserve the order of initial keywords
+        keywords = list(dict.fromkeys(keywords))
```
wookie184, line 559: This isn't strictly necessary as you remove duplicates and sort later in `final_rule_numbers`

wookie184 (changes requested):
Behaviour seems perfect now, thanks! I have a couple comments

wookie184 (approved):
LGTM, thanks!

bot/exts/info/information.py
```diff
@@ -533,21 +554,30 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
-            self.bot.stats.incr(f"rule_uses.{rule}")
+        final_rule_numbers = []
+        final_rule_numbers.extend(rule_numbers)
+
+        for keyword in keywords:
+            final_rule_numbers.append(keyword_to_rule_number.get(keyword))
+
+        final_rules = []
+        final_rule_numbers = set(final_rule_numbers)
+        final_rule_numbers = sorted(list(final_rule_numbers))
```
MarkKoz, line 576: Using `list()` is redundant.

bot/exts/info/information.py
```diff
@@ -533,21 +554,30 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
             await ctx.send(shorten(":x: Invalid rule indices: " + invalid, 75, placeholder=" ..."))
             return
 
-        for rule in rules:
-            self.bot.stats.incr(f"rule_uses.{rule}")
+        final_rule_numbers = []
+        final_rule_numbers.extend(rule_numbers)
+
+        for keyword in keywords:
+            final_rule_numbers.append(keyword_to_rule_number.get(keyword))
```
MarkKoz, line 572: Avoid using `get()` if you don't need to rely on its default value. Otherwise, it may mislead readers into thinking that sometimes the key may not exist. Furthermore, the code below assumes everything in `final_rule_numbers` is a valid index. If a bug was somehow introduced that caused a key to not exist in `keyword_to_rule_number`, the resulting error would be clearer if it originated from this point than from somewhere below.

bot/exts/info/information.py
```diff
@@ -533,21 +554,30 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
+
+        final_rules = []
+        final_rule_numbers = set(final_rule_numbers)
+        final_rule_numbers = sorted(list(final_rule_numbers))
 
-        final_rules = tuple(f"**{pick}.** {full_rules[pick - 1]}" for pick in rules)
+        for rule_number in final_rule_numbers:
+            self.bot.stats.incr(f"rule_uses.{rule_number}")
+            final_rules.append(f"**{rule_number}.** {full_rules[rule_number - 1][0]}")
```
MarkKoz, line 580: Thoughts on doing it like this instead?

```py
final_rules = []
final_rule_numbers = {keyword_to_rule_number[keyword] for keyword in keywords}
final_rule_numbers.update(rule_numbers)

for rule_number in sorted(final_rule_numbers):
    ...
```

MarkKoz (changes requested):
Functionally, everything is fine. However, I have some suggestions to improve the code.

bot/exts/info/information.py
```diff
@@ -533,21 +554,24 @@ async def rules(self, ctx: Context, rules: Greedy[int]) -> None:
...
             return
 
-        for rule in rules:
-            self.bot.stats.incr(f"rule_uses.{rule}")
+        final_rules = []
+        final_rule_numbers = {keyword_to_rule_number[keyword] for keyword in keywords}
+        final_rule_numbers.update(rule_numbers)
 
-        final_rules = tuple(f"**{pick}.** {full_rules[pick - 1]}" for pick in rules)
+        for rule_number in final_rule_numbers:
```
MarkKoz, line 572: You forgot this.
```suggestion
        for rule_number in sorted(final_rule_numbers):
```

MarkKoz (changes requested):

tests/bot/exts/info/test_information.py
```diff
@@ -619,3 +619,34 @@ async def test_return_none_if_one_rule_number_is_invalid(self):
                 self.ctx.send.call_args,
                 unittest.mock.call(shorten(":x: Invalid rule indices: " + invalid, 75, placeholder=" ...")))
             self.assertEqual(None, final_rule_numbers)
+
+    async def test_return_correct_rule_numberstest_return_correct_rule_numbers(self):
```
TizzySaurus, line 623: Is there a reason this name is duplicated?

tests/bot/exts/info/test_information.py
```diff
@@ -619,3 +619,34 @@ async def test_return_none_if_one_rule_number_is_invalid(self):
...
+            ((), None),
+            (("hello", "2", "second"), None),
+            (("hello", "999"), None),
+        ]
+
+        description = (
+            "The rules and guidelines that apply to this community can be found on"
+            " our [rules page](https://www.pythondiscord.com/pages/rules). We expect"
+            " all members of the community to have read and understood these."
+        )
```
MarkKoz, line 646: Define a constant instead and share that with the test. This avoids duplication.

tests/bot/exts/info/test_information.py
```diff
@@ -619,3 +619,34 @@ async def test_return_none_if_one_rule_number_is_invalid(self):
...
                 unittest.mock.call(shorten(":x: Invalid rule indices: " + invalid, 75, placeholder=" ...")))
             self.assertEqual(None, final_rule_numbers)
+
+    async def test_return_correct_rule_numberstest_return_correct_rule_numbers(self):
+
+        test_cases = [
+            (("1", "2", "first"), {1, 2}),
+            (("1", "hello", "2", "second"), {1}),
+            (("second", "third", "unknown", "999"), {2, 3})
+        ]
```
MarkKoz, line 629: When there are multiple test cases, use the `subTest` context manager to isolate the asserts.

MarkKoz (changes requested):

bot/constants.py
```diff
@@ -774,3 +774,9 @@ class ThreadArchiveTimes(Enum):
     "Noooooo!!",
     "I can't believe you've done this",
 ]
+
+DEFAULT_RULES_DESCRIPTION = (
+    "The rules and guidelines that apply to this community can be found on"
+    " our [rules page](https://www.pythondiscord.com/pages/rules). We expect"
+    " all members of the community to have read and understood these."
+)
```
MarkKoz, line 782: Sorry for the back-and-forth and not being clearer about this. The constant should be defined in `information.py`. The `bot/constants.py` file is meant for global constants i.e. constants shared by many modules. Tests don't really fit the criteria of "shared by many modules".

MarkKoz (changes requested):

MarkKoz (approved):

PR #2267, opened 2022-09-01
Improve reporting on autoban filter lists

ChrisLovering:
These changes have been discussed within the mod team.

This PR changes autoban filter alerts to be sent to #filter-log instead of #mod-meta. This is with the aim to reduce the amount of spam in #mod-meta.

On top of that, this PR adds a new command to generate a report of all autoban filters added in the past 7 days. This report is also scheduled to be sent to #mod-meta once a week at 6pm UTC.

## Screenshots
Empty state:
image

Small number:
image

Large number of filters:
image

### DIFF
--- a/bot/exts/filters/filter_lists.py
+++ b/bot/exts/filters/filter_lists.py
@@ -117,9 +117,9 @@ class FilterLists(Cog):
                 )
             raise
 
-        # If it is an autoban trigger we send a warning in #mod-meta
+        # If it is an autoban trigger we send a warning in #filter-log
         if comment and "[autoban]" in comment:
-            await self.bot.get_channel(Channels.mod_meta).send(
+            await self.bot.get_channel(Channels.filter_log).send(
                 f":warning: Heads-up! The new `{list_type}` filter "
                 f"`{content}` (`{comment}`) will automatically ban users."
             )
--- a/bot/exts/filters/filter_lists.py
+++ b/bot/exts/filters/filter_lists.py
@@ -1,13 +1,13 @@
 import re
 from typing import Optional
 
+import discord
 from botcore.site_api import ResponseCodeError
-from discord import Colour, Embed
 from discord.ext.commands import BadArgument, Cog, Context, IDConverter, group, has_any_role
 
 from bot import constants
 from bot.bot import Bot
-from bot.constants import Channels
+from bot.constants import Channels, Colours
 from bot.converters import ValidDiscordServerInvite, ValidFilterListType
 from bot.log import get_logger
 from bot.pagination import LinePaginator
@@ -179,9 +179,9 @@ class FilterLists(Cog):
 
         # Build the embed
         list_type_plural = list_type.lower().replace("_", " ").title() + "s"
-        embed = Embed(
+        embed = discord.Embed(
             title=f"{allow_type.title()}ed {list_type_plural} ({len(result)} total)",
-            colour=Colour.blue()
+            colour=Colours.blue
         )
         log.trace(f"Trying to list {len(result)} items from the {list_type.lower()} {allow_type}")
 
--- a/bot/exts/filters/filter_lists.py
+++ b/bot/exts/filters/filter_lists.py
@@ -1,9 +1,13 @@
+import datetime
 import re
+from collections import defaultdict
 from typing import Optional
 
+import arrow
 import discord
 from botcore.site_api import ResponseCodeError
-from discord.ext.commands import BadArgument, Cog, Context, IDConverter, group, has_any_role
+from discord.ext import tasks
+from discord.ext.commands import BadArgument, Cog, Context, IDConverter, command, group, has_any_role
 
 from bot import constants
 from bot.bot import Bot
@@ -13,6 +17,7 @@ from bot.log import get_logger
 from bot.pagination import LinePaginator
 
 log = get_logger(__name__)
+WEEKLY_REPORT_ISO_DAY = 3  # 1=Monday, 7=Sunday
 
 
 class FilterLists(Cog):
@@ -33,6 +38,7 @@ class FilterLists(Cog):
     async def cog_load(self) -> None:
         """Add the valid FilterList types to the docstrings, so they'll appear in !help invocations."""
         await self.bot.wait_until_guild_available()
+        self.weekly_autoban_report_task.start()
 
         # Add valid filterlist types to the docstrings
         valid_types = await ValidFilterListType.get_valid_types(self.bot)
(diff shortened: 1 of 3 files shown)


bot/exts/filters/filter_lists.py
```diff
@@ -287,10 +293,58 @@ async def deny_sync(self, ctx: Context) -> None:
...
+    async def weekly_autoban_report_task(self) -> None:
+        """Trigger autoban report to be sent if it is the desired day of the week (WEEKLY_REPORT_ISO_DAY)."""
+        if arrow.utcnow().isoweekday() != WEEKLY_REPORT_ISO_DAY:
+            return
+
+        await self.send_weekly_autoban_report()
+
+    async def send_weekly_autoban_report(self, channel: discord.abc.Messageable = None) -> None:
+        """Send a list of autobans added in the last 7 days to #mod-meta."""
+        seven_days_ago = arrow.utcnow().shift(days=-7)
```
wookie184, line 311: Given the actual time this is run could differ by up to 18 hours each week, I think it would be possible that any autobans added in (up to) the first 18 hours after a message *may* not appear in the next message.

We could make this `-8` days (or -7 and 18 hours), if we'd rather have repeats than missed ones. I don't mind either way. I don't think it's worth a more complex solution to 'fix' this though, as it seems pretty minor.

mbaruh: Where do the 18 hours come from?

ChrisLovering: https://github.com/python-discord/bot/pull/2267/files#diff-2d7d975ffaf01fcfb495567a4238c5d0e5fee119591ad115a83ea6ebf5471677R301 means that it will run at 6pm UTC every day. Rather than some 18 hour offset.

https://discordpy.readthedocs.io/en/latest/ext/tasks/index.html#discord.ext.tasks.loop

wookie184: I swear I got confused on the exact same thing on a different PR recently 🤦

Ignore what I said 😅

bot/exts/filters/filter_lists.py
```diff
@@ -287,10 +293,58 @@ async def deny_sync(self, ctx: Context) -> None:
         """Syncs both allowlists and denylists with the API."""
         await self._sync_data(ctx)
 
+    @command(name="filter_report")
+    async def force_send_weekly_report(self, ctx: Context) -> None:
+        """Send a list of autobans added in the last 7 days to #mod-meta."""
+        await self.send_weekly_autoban_report()
```
wookie184, line 299: Would it make more sense to send this to the channel the command was invoked in?

bot/exts/filters/filter_lists.py
```diff
@@ -287,10 +293,58 @@ async def deny_sync(self, ctx: Context) -> None:
...
+                created_at = arrow.get(filter_details["created_at"])
+                updated_at = arrow.get(filter_details["updated_at"])
+                # Default to empty string so that the in check below doesn't error on None type
+                comment = filter_details["comment"] or ""
+                if max(created_at, updated_at) > seven_days_ago and "[autoban]" in comment:
+                    line = f"`{filter_content}`: {comment}"
+                    added_autobans[f"**{filter_type} {allow_type}**"].append(line)
+
+        # Nicely format the output so each filter list type is grouped
+        lines = [f"**Autoban filters added since {seven_days_ago.format('YYYY-MM-DD')}**"]
```
wookie184, line 331: Could use a discord timestamp here, I don't mind though.

bot/exts/filters/filter_lists.py
ChrisLovering: hah, yes. I even added the `channel` arg to `send_weekly_autoban_report` to support this, I just never updated this line to send ctx.channel

ChrisLovering: [python-discord/bot@`86aaf02` (#2267)](https://github.com/python-discord/bot/pull/2267/commits/86aaf02213eb92a49be6e4b7852e815ddf1c0272)

wookie184 (approved):
👌 👌 👌

bot/exts/filters/filter_lists.py
```diff
@@ -287,10 +293,58 @@ async def deny_sync(self, ctx: Context) -> None:
         """Syncs both allowlists and denylists with the API."""
         await self._sync_data(ctx)
 
+    @command(name="filter_report")
+    async def force_send_weekly_report(self, ctx: Context) -> None:
+        """Send a list of autobans added in the last 7 days to #mod-meta."""
```
vivekashok1221, line 298: This docstring is technically inaccurate since it can send the report in any channel (you probably pasted this over from another method).
Also. we should probably add a check to ensure that the command cannot be run in public channels.

bot/exts/filters/filter_lists.py
```diff
@@ -287,10 +293,58 @@ async def deny_sync(self, ctx: Context) -> None:
...
+            allow_type = "Allow list" if allow.lower() == "true" else "Deny list"
+
+            for filter_content, filter_details in filters.items():
+                created_at = arrow.get(filter_details["created_at"])
+                updated_at = arrow.get(filter_details["updated_at"])
+                # Default to empty string so that the in check below doesn't error on None type
+                comment = filter_details["comment"] or ""
+                if max(created_at, updated_at) > seven_days_ago and "[autoban]" in comment:
+                    line = f"`{filter_content}`: {comment}"
+                    added_autobans[f"**{filter_type} {allow_type}**"].append(line)
```
vivekashok1221, line 328: if `allow_type` is `"Allow list"`, why would we autoban based on it? This question is probably not even related to the PR, it's just that I feel like I'm missing something :sweat_smile: .

vivekashok1221 (changes requested):
Good stuff. 

And we also gotta remember to update the config file for the staff test server (I don't mind doing it).

bot/exts/filters/filter_lists.py
ChrisLovering: Updated the docstring in [`f2d05d5` (#2267)](https://github.com/python-discord/bot/pull/2267/commits/f2d05d554f012598bcd42b4888400ced6513cb0d). There is a cog wide check which ensures commands are only ran in mod channel (which is why we can't modify filters elsewhere)

bot/exts/filters/filter_lists.py
ChrisLovering: Hah, technically autobans can be added to allow lists, which in itself seems wrong.

ChrisLovering:
Rebased onto main (hence the forcepush)

bot/exts/filters/filter_lists.py
vivekashok1221: The check only prevents non-mods from running the command as far as I can see, it doesn't impose any channel restriction.

ChrisLovering: Good point, [`2142d82` (#2267)](https://github.com/python-discord/bot/pull/2267/commits/2142d824272c738c3d4ac9f20dd59cc4f02e345e)

bot/exts/filters/filter_lists.py
vivekashok1221: True.

But you know what, we can leave this part as it is so that we can identify if an autoban is added to an allow-list.
It might be better to add a check to not allow autoban to be added to an allow-list in the first place but after a quick glance I don't think the bot will autoban based on allow list.
It is possible that a mod might accidentally add a filter to the allow list instead of blacklist and the report will help identify such cases (assuming we miss the original plain message alert sent immediately on adding an autoban filter).

vivekashok1221 (approved):
LGTM :star2:

PR #2268, opened 2022-09-01
Use venvs with poetry in Dockerfile

ChrisLovering:
This is due to a regression in poetry 1.2.0, see https://github.com/HassanAbouelela/actions/pull/7

Without this change, CI is currently failing to install deps, due to us using user-based installs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,26 @@
 FROM --platform=linux/amd64 python:3.10-slim
 
 # Set pip to have no saved cache
-ENV PIP_NO_CACHE_DIR=false \
-    POETRY_VIRTUALENVS_CREATE=false
+ENV PIP_NO_CACHE_DIR=1 \
+    PIP_DISABLE_PIP_VERSION_CHECK=on \
+    POETRY_VERSION=1.1.15 \
+    POETRY_HOME="/opt/poetry" \
+    POETRY_VIRTUALENVS_IN_PROJECT=true \
+    POETRY_NO_INTERACTION=1 \
+    INSTALL_DIR="/opt/dependencies" \
+    APP_DIR="/bot"
 
+ENV PATH="$POETRY_HOME/bin:/$INSTALL_DIR/.venv/bin:$PATH"
 
-# Install poetry
-RUN pip install -U poetry
+RUN apt-get update \
+  && apt-get -y upgrade \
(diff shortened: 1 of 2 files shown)


HassanAbouelela (approved):

D0rs4n (approved):

ionite34 (approved):
CI code looks good.

shenanigansd (approved):
Docker image builds successfully

wookie184 (approved):
Can't say I know fully what's happening but the checks on this PR aren't failing, which is better than what we have on main :P

D0rs4n (approved):

Dockerfile
```diff
@@ -2,16 +2,25 @@ FROM --platform=linux/amd64 python:3.10-slim
 
 # Set pip to have no saved cache
 ENV PIP_NO_CACHE_DIR=false \
-    POETRY_VIRTUALENVS_CREATE=false
+    PIP_DISABLE_PIP_VERSION_CHECK=on \
+    POETRY_VERSION=1.1.15 \
+    POETRY_HOME="/opt/poetry" \
+    POETRY_VIRTUALENVS_IN_PROJECT=true \
+    POETRY_NO_INTERACTION=1 \
+    INSTALL_DIR="/opt/dependancies" \
```
imsofi, line 10: `dependancies` seems like a misspelling?

ChrisLovering: Thanks for your keen eye. This has been fixed.

docker-compose.yml
```diff
@@ -103,7 +103,7 @@ services:
       context: .
       dockerfile: Dockerfile
     volumes:
-      - .:/bot:ro
+      - ./bot:/bot/bot:ro
```
HassanAbouelela, line 106: Why was this change made, don’t we still want the other files and directories such as tests and configs?

HassanAbouelela (changes requested):

docker-compose.yml
ChrisLovering: This was made because I was putting the venv inside APP_DIR, causing venv on host to overwrite it with this volume.

It's not needed now though.

ChrisLovering: Dropped the commit that added this

HassanAbouelela (approved):

PR #2269, opened 2022-09-02
Allow evaling code from context menus

### LINKED ISSUE #2270: Allow evaluating code via a context menu?
PR: #2269 
I think it would be handy if you could write out your code, and then right click and eval the code. This also works with other people's code, in case you're curious on the result.

The actual implementation shouldn't be hard either, the `run_job` function is already there, all that needs to be done is have it moved into the context menu callback.

Robin5605:
Closes #2270 
Pretty simple change. None of the internals have been tweaked. A new context menu with type 3 has been added that will allow for users to eval their or someone else's code (this is new) with the press of a button.

### DIFF
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -9,9 +9,12 @@ from typing import Literal, Optional, Tuple
 
 from botcore.utils import interactions
 from botcore.utils.regex import FORMATTED_CODE_REGEX, RAW_CODE_REGEX
-from discord import AllowedMentions, HTTPException, Interaction, Message, NotFound, Reaction, User, enums, ui
+from discord import (
+    AllowedMentions, HTTPException, Interaction, Message, NotFound, Object, Reaction, User, app_commands, enums, ui
+)
 from discord.ext.commands import Cog, Command, Context, Converter, command, guild_only
 
+from bot import constants
 from bot.bot import Bot
 from bot.constants import Categories, Channels, MODERATION_ROLES, Roles, URLs
 from bot.decorators import redirect_output
@@ -158,6 +161,14 @@ class Snekbox(Cog):
 
     def __init__(self, bot: Bot):
         self.bot = bot
+
+        # The `app_commands.context_menu` decorator does not support cogs
+        self.ctx_menu = app_commands.ContextMenu(
+            name="Run code",
+            callback=self.eval_ctx_menu,
+        )
+        self.bot.tree.add_command(self.ctx_menu)
+
         self.jobs = {}
 
     def build_python_version_switcher_view(
@@ -480,6 +491,13 @@ class Snekbox(Cog):
                 break
             log.info(f"Re-evaluating code from message {ctx.message.id}:\n{code}")
 
+    async def eval_ctx_menu(self, interaction: Interaction, message: Message) -> None:
+        """Same functionality as `eval_command` but invoked using a context menu."""
+        fake_ctx: Context = await Context.from_interaction(interaction)
+        python_version = "3.11"
+        code = await CodeblockConverter.convert(fake_ctx, message.content)
+        await self.run_job("eval", fake_ctx, python_version, "\n".join(code))
+
     @command(name="eval", aliases=("e",), usage="[python_version] <code, ...>")
     @guild_only()
     @redirect_output(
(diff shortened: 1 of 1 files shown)


mbaruh:
We should discuss some limitations here. We wouldn't want, for example, for someone to eval a message from 10 hours ago, especially in help channels. That would just make the bot message pop out of nowhere without context. So what time limit do we set here?

Another is limiting a user to eval only their own messages, and posting an ephemeral message if they try otherwise.

Since the input message might not have the `!eval` command invocation, it might be a better UI to reply to the input message (should it still do that for normal invocations?).

mbaruh (changes requested):
We should discuss some limitations here. We wouldn't want, for example, for someone to eval a message from 10 hours ago, especially in help channels. That would just make the bot message pop out of nowhere without context. So what time limit do we set here?

Another is limiting a user to eval only their own messages, and posting an ephemeral message if they try otherwise.

Since the input message might not have the `!eval` command invocation, it might be a better UI to reply to the input message (should it still do that for normal invocations?).

Robin5605:
> We wouldn't want, for example, for someone to eval a message from 10 hours ago, especially in help channels

You're right, we wouldn't. We could probably add some config option for this in `config.yml` or similar, but I think 2 minutes seems reasonable. As for the message popping up without any context, not necessarily. You can see which user ran the command, and you can click on it to be taken back to the original message that eval was ran on.

> Since the input message might not have the !eval command invocation, it might be a better UI to reply to the input message (should it still do that for normal invocations?).

This is another slight design issue. Due to the nature of context menus, it can technically be ran on _any_ message, even ones that don't have any code blocks or any code at all. We have a few options here:
- Parse for code block, and if there is none, just use the message content as code to evaluate. If the command is ran on a random message, e.g "Hello people of Python Discord! I'm new to this community 😄 ", the bot would try to eval this and come up with a syntax error. 
- We can forcefully look for a code block, and throw an error if there is none. Something along the lines of "this message doesn't have a code block"

> Another is limiting a user to eval only their own messages, and posting an ephemeral message if they try otherwise.

I'm not really sure on this one. I don't see any immediate downsides if someone were to try to eval someone else's code. The result would be ephemeral, and no damage is done. 
 
If I'm missing anything important here, please do let me know. Cheers 🎉

mbaruh:
> I'm not really sure on this one. I don't see any immediate downsides if someone were to try to eval someone else's code. The result would be ephemeral, and no damage is done.

What do you mean by "the result would be ephemeral"? Anyone can see the output right?

bot/exts/utils/snekbox.py
```diff
@@ -568,3 +586,7 @@ def predicate_emoji_reaction(ctx: Context, reaction: Reaction, user: User) -> bo
 async def setup(bot: Bot) -> None:
     """Load the Snekbox cog."""
     await bot.add_cog(Snekbox(bot))
+
+    guild = Object(id=constants.Guild.id)
```
shtlrs, line 590: I think that this code should go away once [this PR](https://github.com/python-discord/bot-core/pull/171) is completed

bot/exts/utils/snekbox.py
```diff
@@ -480,6 +491,13 @@ async def run_job(
                 break
             log.info(f"Re-evaluating code from message {ctx.message.id}:\n{code}")
 
+    async def eval_ctx_menu(self, interaction: Interaction, message: Message) -> None:
+        """Same functionality as `eval_command` but invoked using a context menu."""
+        fake_ctx: Context = await Context.from_interaction(interaction)
+        python_version = "3.11"
```
shtlrs, line 497: Don't we keep track of the python version that we usually use for evaluation ? 
Having this here can lead to us forgetting about it once we update the py version

shtlrs (changes requested):
Looks good, just 2 small things to consider

wookie184:
I think we still need to decide if this is something we want, and how it should work, so i'll mark this as stalled until there's a consensus on the issue, or on the server/in a staff meeting.

Robin5605:
> What do you mean by "the result would be ephemeral"? Anyone can see the output right?

@mbaruh: No, the resultant message would be only visible to the person who ran the command. So it shouldn't add any clutter for anybody else in the channel as well.

Robin5605:
As for the viability of the feature itself - I'm not entirely sure it's really an improvement over the existing `!eval` command. There are limitations and I don't see a context menu fixing a major problem that the `!eval` command already has. I would not be averse to just closing the issue.

PR #2275, opened 2022-09-11
Removed "redis_ready" from additional_spec_asyncs in MockBot

AlexZasorin:
"redis_ready" attribute was removed from Bot in fc05849 so it no longer needs to be mocked. It was [suggested](https://discord.com/channels/267624335836053506/635950537262759947/1018530171436937236) that this change was small enough to not warrant creating an issue.


ionite34 (approved):
Seems fine

PR #2283, opened 2022-09-28
1379 tranform urls upon rich embed filter application

### LINKED ISSUE #1379: Adjust rich_embeds filter so it doesn't catch edge case
Right now, discord converts twitter links when applying the standard twitter embed treatment. Our check for rich embeds determines whether an embed is autogenerated or not by checking if the embed link is present within the source message here https://github.com/python-discord/bot/blob/fa02180246e6cf1245d5705b36a2056ed449ef6b/bot/exts/filters/filtering.py#L520, but this does not account for discords transformation of mobile.twitter.com links to twitter.com links, and thus a message with mobile.twitter.com in it is falsely detected as a rich embed for the watchlist.

This is probably fixable by detecting if the link goes to twitter.com and there is a link to mobile.twitter.com with the same parameters. That may get a little tricky, but twitter links aren't complex links so perhaps simple string operations will suffice.

We may also want to consider writing an "apply_link_transformations" function that will extract and transform relative links within a message to compare against the embed.

shtlrs:
closes #1379

The current implementation allows to collect the URLs from the sent messages and then apply a transformation on them.

The transformation here is basically a removal of the subdomain (if it exists) because of the way Discord renders some URLs (Twitter's in particular)

### DIFF
--- a/bot/exts/filters/filtering.py
+++ b/bot/exts/filters/filtering.py
@@ -23,9 +23,11 @@ from bot.constants import Bot as BotConfig, Channels, Colours, Filter, Guild, Ic
 from bot.exts.events.code_jams._channels import CATEGORY_NAME as JAM_CATEGORY_NAME
 from bot.exts.moderation.modlog import ModLog
 from bot.log import get_logger
+from bot.utils.helpers import remove_subdomain_from_url
 from bot.utils.messages import format_user
 
 log = get_logger(__name__)
+
 
 # Regular expressions
 CODE_BLOCK_RE = re.compile(
@@ -583,7 +585,7 @@ class Filtering(Cog):
         """
         text = self.clean_input(text)
 
-        # Remove backslashes to prevent escape character aroundfuckery like
+        # Remove backslashes to prevent escape character around fuckery like
         # discord\.gg/gdudes-pony-farm
         text = text.replace("\\", "")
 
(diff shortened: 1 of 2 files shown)


shtlrs:
@bast0006, @scragly The floor is yours.

bot/exts/filters/filtering.py
```diff
@@ -649,6 +651,12 @@ async def _has_rich_embed(msg: Message) -> Union[bool, List[Embed]]:
             for embed in msg.embeds:
                 if embed.type == "rich":
                     urls = URL_RE.findall(msg.content)
+                    final_urls = set(urls)
+                    # This is due to way discord renders relative urls in Embdes
+                    # if we send the following url: https://mobile.twitter.com/something
+                    # Discord renders it as https://twitter.com/something
+                    for url in urls:
+                        final_urls.add(remove_subdomain_from_url(url))
                     if not embed.url or embed.url not in urls:
```
ChrisLovering, line 660: You've created a variable `final_urls` and are adding urls to it, but then the check is still `if ... embed.url not in urls`, so does this even do anything?

Also, we have the tldextract lib installed in the project, which is better at separating subddomains from urls than urllib and then assuming the first 2 parts are the subdomain (which isn't always the case)/.

shtlrs: Shoot. The first implementation wasn't in a set and I was adding them in the original list, but I wanted to remove dups.
My bad, I'll fix that.

As for the `tldextract` one, i'll have a look at it right away

ChrisLovering (approved):
Looks good and works well.

mbaruh (approved):

PR #2284, opened 2022-10-03
Fix all non-awaited bot.remove_cog calls

### LINKED ISSUE #2281: Async function remove_cog not awaited
If you take a look at this line
https://github.com/python-discord/bot/blob/71fbd5aa7ef8fdf3547dbae845cf7e6b9bc08277/bot/exts/moderation/watchchannels/_watchchannel.py#L131

You'll see that the remove_cog async func isn't awaited.

shtlrs:
closes #2281

We currently have some instances of the bot.remove_cog async function that is not awaited. 
This PR updates all those calls to be awaited

### DIFF
--- a/bot/exts/filters/antispam.py
+++ b/bot/exts/filters/antispam.py
@@ -156,7 +156,7 @@ class AntiSpam(Cog):
                 colour=Colour.red()
             )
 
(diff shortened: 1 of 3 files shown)


wookie184 (approved):
Thanks :)

PR #2285, opened 2022-10-04
Add slicing tag

Keyacom:
This file adds a tag with an explanation on sequence slicing.


HassanAbouelela:
Hello, thanks for writing this up! Is there an issue for this tag?

Keyacom:
I wrote this because after I helped a person in a help channel, I thought that it would be useful to write a slice syntax tag, even if there is no such issue about it in python-discord/meta.

The conversation:


The message where I mentioned the slice syntax (link):
https://discord.com/channels/267624335836053506/776184661902491678/1026550769358487573

HassanAbouelela:
Alright, for now you can keep this PR open, but please open an issue on the meta repository detailing what it is you think should be added and why. Linking to this PR is a good idea to explain the what.

bot/resources/tags/slicing.md
```diff
@@ -0,0 +1,31 @@
...
+[1, 2, 3, 4]
+>>> l[::2]
+[1, 3]
+```
+Using `some_list[::-1]` is the same as `list(reversed(some_list))`. Just like in regular sequence subscriptions, negative integers may be used.
+
+**Notes**
+• If the start index is greater than the end index, the resulting sequence will be empty.
+• The number of items before applying the step can be calculated as `n = j - i`.
+• The number of items after applying the step is `n / k`, rounded down, but cannot be less than 1, unless `n` is exactly 0.
```
wookie184, line 31: I'm not sure these are really necessary to mention

wookie184 (commented):
Thanks for writing this, the structure of this tag is good, although I think some of the information isn't necessary. It's best to keep tags short and snappy, otherwise people wont read them.

I've tried condensing the tag down a bit, could you take a look and see what you think and maybe merge both our versions togther?
<pre>
---
aliases: ["slice", "seqslice", "seqslicing", "sequence-slice", "sequence-slicing"]
embed:
    title: "Sequence slicing"
---
*slicing* is a way of accessing a part of a sequence by specifying a start index, stop index, and step to take elements at.

The syntax for slicing is: `sequence[start:stop:step]`. Each argument can be left out to use a default value, which is `1` for `step`, and is each end of the sequence (index `0` and `len(sequence)`), for `start` and `stop`.

Like sequence indexing, negative integers for the `start` and `stop` will take the nth index *from the end* of the sequence. A negative `step` value will reverse the sequence.

**Examples**
```py
>>> l = [1, 2, 3, 4]
>>> l[2:]
[3, 4]
>>> l[:2]
[1, 2]
>>> l[::-1]
[4, 3, 2, 1]
>>> l[:]
[1, 2, 3, 4]
>>> l[::2]
[1, 3]
```
```py
>>> s = 'abcDeFgHi'
>>> s[-2:2:-2]
'HFD'
>>> s[100:]
''
```
</pre>

Keyacom:
> Thanks for writing this, the structure of this tag is good, although I think some of the information isn't necessary. It's best to keep tags short and snappy, otherwise people wont read them.
> 
> I've tried condensing the tag down a bit, could you take a look and see what you think and maybe merge both our versions togther?

Yes, I consider that feasible, but keep in mind that sentences must start with a capital letter. In line 6:
```diff
- *slicing* is a way of accessing a part of a sequence by specifying a start index, stop index, and step to take elements at.
+ *Slicing* is a way of accessing a part of a sequence by specifying a start index, stop index, and step to take elements at.
```

Other than that, I'm blown away at your proposal and I'm definitely accepting it.

bot/resources/tags/slicing.md
Keyacom: By now, even I don't think they are necessary.

bot/resources/tags/slicing.md
```diff
@@ -0,0 +1,31 @@
...
+[4, 3, 2, 1]
+>>> l[:]
+[1, 2, 3, 4]
+>>> l[::2]
+[1, 3]
+```
+Using `some_list[::-1]` is the same as `list(reversed(some_list))`. Just like in regular sequence subscriptions, negative integers may be used.
+
+**Notes**
+• If the start index is greater than the end index, the resulting sequence will be empty.
```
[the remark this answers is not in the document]
Keyacom: Woops, I just didn't notice that.

bot/resources/tags/slicing.md
```diff
@@ -0,0 +1,31 @@
+---
+aliases: ["slice", "seqslice", "seqslicing", "sequence-slice", "sequence-slicing"]
+embed:
+    title: "Sequence slicing"
+---
+You're trying to get a part of a string, list, or another sequence object, but you don't want to manually increment and concatenate? There comes the need to *slice* it.
+
+There is a special syntax that can be used to slice a given `some_seq` sequence: `some_seq[i:j:k]`, where `i` is the starting index, `j` is the end index, and `k` is the step, i.e. every how many items should one be kept (the first one is always kept). `i`, `j`, and `k` all must be integers. If any of these values are missing, they're assumed as `some_seq[0:len(some_seq):1]`.
```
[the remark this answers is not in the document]
Keyacom: Another thing I didn't notice.

wookie184:
> Yes, I consider that feasible, but keep in mind that sentences must start with a capital letter. In line 6:
> ```diff
> - *slicing* is a way of accessing a part of a sequence by specifying a start index, stop index, and step to take elements at.
> + *Slicing* is a way of accessing a part of a sequence by specifying a start index, stop index, and step to take elements at.
> ```
> 
> Other than that, I'm blown away at your proposal and I'm definitely accepting it.

Thanks :) And yeah that capital letter should be fixed.

Would you be able to commit the changes to this PR so others can review it and we can hopefully get it merged?

swfarnsworth:
Thank you for taking the initiative to create this tag. I think slicing is a great candidate for having a tag, and that we've been without one for too long.

We need to be very intentional about what and how much information is conveyed through a tag. In that light, there's a lot that I'd like to see different about your proposed tag. Let me propose this alternative:

<pre>
---
aliases: ["slice", "seqslice", "seqslicing", "sequence-slice", "sequence-slicing"]
embed:
    title: "Sequence slicing"
---
*slicing* is a way of accessing a part of a sequence by specifying a start, stop, and step. As with normal indexing, negative numbers can be used to count backwards.

**Examples**
```py
>>> letters = ['a', 'b', 'c', 'd', 'e', 'f', 'g']
['a', 'b', 'c', 'd', 'e', 'f', 'g']
>>> letters[2:]  # from element 2 to the end
['c', 'd', 'e', 'f', 'g']
>>> letters[:4]  # up to element 4
['a', 'b', 'c', 'd']
>>> letters[3:5]  # elements 3 and 4 -- the right bound is not included
['d', 'e']
>>> letters[2:-1:2]  # Every other element between 2 and the last
['c', 'e']
>>> letters[::-1]  # The whole list in reverse
['g', 'f', 'e', 'd', 'c', 'b', 'a']
>>> words = "Hello world!"
>>> words[2:7]  # Strings are also sequences
"llo w"
```
</pre>

I changed the elements from integers to strings so that the distinction between indices and elements would be especially clear to beginners.

While it's interesting that slices can return empty sequences rather than raising an IndexError, I don't think beginners are likely to discover this or find it confusing if they do. I also think that's too much detail for a tag.

I don't think knowing about the behavior of `[:]` or `[::]` is useful, since `.copy()` is always preferred.

Keyacom:
@swfarnsworth Your approach could probably be the best one yet.

Keyacom:
Note about the potentially confusing commit summary (because I can't edit it):
> \- Fixed the REPL example (assignments never return)

Assignment expressions (`a := b`) do return.

bot/resources/tags/slicing.md
```diff
@@ -0,0 +1,26 @@
...
+embed:
+    title: "Sequence slicing"
+---
+*Slicing* is a way of accessing a part of a sequence by specifying a start, stop, and step. As with normal indexing, negative numbers can be used to count backwards.
+
+**Examples**
+```py
+>>> letters = ['a', 'b', 'c', 'd', 'e', 'f', 'g']
+>>> letters
+['a', 'b', 'c', 'd', 'e', 'f', 'g']
```
swfarnsworth, line 12: Since this provides no new information, it can be omitted.

swfarnsworth (changes requested):
I have no other requests other than removing the two lines indicated. Thank you for integrating my suggestions, and again for taking the initiative to create this tag!

bot/resources/tags/slicing.md
Keyacom: I just removed that.

PR #2305, opened 2022-10-22
Store time of last vote in redis to prevent vote triggering early

### LINKED ISSUE #2301: Nominations: Fix logic for when to send a new vote
Currently if a vote is archived within a day of being sent, a new vote can be sent immediately, even though a day hasn't passed since that vote was archived.

We should fix this by using redis to cache the date of when the most recent nomination was sent. Then we'll check if it's been a day since *this* date, even if the nomination was archived.

wookie184:
Previously if a vote was sent and ended within a day a new one could be sent immediately.

Closes #2301


ChrisLovering (approved):

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -180,6 +188,9 @@ async def post_review(self, nomination: Nomination) -> None:
         )
         message = await thread.send(f"<@&{Roles.mod_team}> <@&{Roles.admins}>")
 
+        now = datetime.now(tz=timezone.utc)
```
CaedenPH, line 191:
```suggestion
        now = datetime.utcnow()
```

ChrisLovering: that would be incorrect, as it creates a naive datetime object, with the current time offset to UTC. Whereas, what we want here is the current time with UTC as the timezone info.

See the warning in the docs. https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

mbaruh (approved):

PR #2307, opened 2022-10-22
Talentpool Cleanup

### LINKED ISSUE #2306: Nominations: Clean up
I'd like to:
- Remove the cache of nominations, as we refresh it almost every time before using it anyway so it's pretty useless. It's also a potential for bugs in places we forget to keep it updated.
- Convert the site response of nominations into dataclasses or pydantic models. This means we don't have to use dictionary accessing everywhere and makes my IDE happy because it knows the types. (I think pydantic would probably be easiest as it handles dictionary->model conversion automatically)
- Remove support for manually posting a review/manually specifying review contents in the voting channel. We rely on message contents being what the bot sent in a couple places so I think this could cause issues even currently. It's also not really necessary as you can still trigger votes manually and then comment any extra details in the thread.

wookie184:
Closes #2306. Tested and ready for review!

Sorry for the big diff, it's not that bad I promise.

Marked as high priority as this blocks other changes to talentpool stuff.


bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -436,22 +371,18 @@ async def _edit_nomination_reason(
             else:
                 raise
 
-        if not nomination["active"]:
-            await ctx.send(f":x: <@{nomination['user']}> doesn't have an active nomination.")
+        if not nomination.active:
```
shtlrs, line 374: Question: Will this path ever be executed ?
Since we now pass `active` as `True` when bringing nominations

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -121,45 +119,46 @@ async def get_user_for_review(self) -> Optional[int]:
...
+                possible.append(nomination)
 
         if not possible:
             log.debug("No users ready to review.")
             return None
 
-        oldest_date = min(isoparse(x[1]["inserted_at"]) for x in possible)
-        max_entries = max(len(x[1]["entries"]) for x in possible)
+        oldest_date = min(nom.inserted_at for nom in possible)
+        max_entries = max(len(nom.entries) for nom in possible)
```
shtlrs, line 137: Suggestion: could `possible_nominations` and `possible_nomination` be a better fit as names ?

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -35,45 +36,20 @@ class TalentPool(Cog, name="Talentpool"):
...
-        self.api_default_params = {'active': 'true', 'ordering': '-inserted_at'}
-
+        self.api = NominationAPI(bot.api_client)
+        self.reviewer = Reviewer(bot, self.api)
         # This lock lets us avoid cancelling the reviewer loop while the review code is running.
         self.autoreview_lock = asyncio.Lock()
 
     async def cog_load(self) -> None:
-        """Load user cache and maybe start autoreview loop."""
-        await self.refresh_cache()
```
shtlrs, line 47: Question: I'm curious, why did we remove the cache ?

shtlrs (commented):
I'm no expert in this particular feature, but it LGTM !

bot/exts/recruitment/talentpool/_cog.py
wookie184: It was refreshed pretty much every time it was accessed which made it pretty pointless, as if you make an API call every time anyway it defeats the point of a cache. There were also a bunch of times where it was refreshed even though it didn't need to be, so i'd expect this PR actually makes *less* API calls now.

It would be possible to reimplement it properly (in `_api.py` so it's more contained), although it's not a very high-throughput cog so it's not really necessary and is just extra code that needs to be maintaned and could have bugs.

The reason it was there in the first place is because it used to be a watchchannel (https://github.com/python-discord/bot/blob/main/bot/exts/moderation/watchchannels/_watchchannel.py) which tldr meant it needed to access api info for every message sent on the server so a cache was necessary.

bot/exts/recruitment/talentpool/_cog.py
wookie184: If `target` is an `int` then it gets the nomination straight from that and it can be inactive, rather than if a member/user is passed in which case it does that search for active nominations only.

bot/exts/recruitment/talentpool/_api.py
```diff
@@ -0,0 +1,110 @@
...
+    async def edit_nomination(
+        self,
+        nomination_id: int,
+        *,
+        end_reason: str | None = None,
+        active: bool | None = None,
+        reviewed: bool | None = None,
+    ) -> Nomination:
+        """
+        Edit a nomination entry.
```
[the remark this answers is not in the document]
wookie184: Good spot, yeah, I'll fix that.

bot/exts/recruitment/talentpool/_cog.py
shtlrs: Ah I see now, thanks !

bot/exts/recruitment/talentpool/_cog.py
shtlrs: And do we have metrics of number of api calls / duplicated calls ?

bot/exts/recruitment/talentpool/_cog.py
wookie184: I noticed that bit and another place could be refactored a bit to make it a bit neater and remove a potential API call, so did that.

bot/exts/recruitment/talentpool/_api.py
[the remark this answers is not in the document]
wookie184: Fixed

bot/exts/recruitment/talentpool/_cog.py
wookie184: We have stats for requests the site gets by status code over time, but not where they're coming from afaik.

shtlrs (approved):

shtlrs (approved):
Thanks !

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -273,53 +230,30 @@ async def _nominate_user(self, ctx: Context, user: MemberOrUser, reason: str) ->
...
+                case (400, {"actor": _}):
                     await ctx.send(f":x: You have already nominated {user.mention}.")
-
-                return
-            else:
-                resp.raise_for_status()
-
-        self.cache[user.id] = response_data
+                    return
+            raise
```
ionite34, line 247: Should there be an explicit exception raised instead of the plain `raise` usages here?
Currently if this does raise, it would trace the following:
```py
RuntimeError: No active exception to reraise
```
which might cause confusion later. And since we're matching a specific 400 code, it's pretty trivial for these matches to be non-exhaustive and hit the raise.

ionite34 (changes requested):

bot/exts/recruitment/talentpool/_cog.py
wookie184: It's still in the `except` block so should just reraise that exception, right?

ChrisLovering (approved):
Looks great wookie, should be much easier to maintain now that it has typing!

Nice use of match case too.

bot/exts/recruitment/talentpool/_cog.py
ionite34: It does, I got confused 👍

ionite34 (approved):
Looks good

wookie184:
Thanks for the reviews :)

PR #2309, opened 2022-10-23
Add rule keywords discoverability

### LINKED ISSUE #2280: Add discoverability of keywords that point to particular rules
This PR: [Add support to fetch rules via keywords](https://github.com/python-discord/bot/pull/2261) allows us now to invoke particular rules based on keywords that map to that rule.

The **problem** is that there is currently no way to discover what these keywords are other than looking at the [source code](https://github.com/shtlrs/python-discord-site/blob/7fd81a80049763f1dc71815da27d65ce4344a191/pydis_site/apps/api/views.py#L133) to look for these keywords.

So I think having a new command, whose name can be agreed upon later, that lists these keywords would be helpful.

**Notes**
* Let's call the command **rule_keywords** for now.
* Output format is just for demo purposes and will be discussed

Here's how I see this working.

1. We can fetch keywords of **one rule** on the rule number.
So for example, if I Invoke the following command `!rule 1`, and the bot sends out the rule, and if I'm curious about what could the possible keywords for that rule are, I can invoke it likewise:
`!rule_keywords 2`
This would trigger the following response
image

2. We can allow fetching keywords of **multiple rules at once** like this: `!rule 1 2 3`
The output will be the same as the previous one, only just **paginated** ( one rule per page )

3. If someone invokes it with no particular number, i guess the choice here is to either list them all or none at all

4.  I first thought of the possibility of checking whether a kw exists or not, but that's unnecessary since someone can just try invoking a rule with that keyword. So I think this **shouldn't be an option**

@wookie184, @MarkKoz  what do you think ?

shtlrs:
Closes #2280

The purpose of this PR is to add the available keywords per rule upon consulting the help docs of the `rules` command


bot/exts/info/information.py
```diff
@@ -578,7 +589,12 @@ async def rules(self, ctx: Context, *args: Optional[str]) -> Optional[Set[int]]:
...
 
+    async def cog_load(self) -> None:
+        """Carry out cog asynchronous initialisation."""
+        await self._set_rules_command_help()
+
 
 async def setup(bot: Bot) -> None:
     """Load the Information cog."""
     await bot.add_cog(Information(bot))
+    log.info("Cog loaded: Information")
```
wookie184, line 600: This is already done in a different place (here https://github.com/python-discord/bot-core/blob/db0ff48a574f5b466f3e8dc25cc32c3a9bfc1370/botcore/_bot.py#L155), so shouldn't be needed here.

bot/exts/info/information.py
```diff
@@ -523,6 +523,17 @@ async def json(self, ctx: Context, message: Message) -> None:
...
         await self.send_raw_content(ctx, message, json=True)
 
+    async def _set_rules_command_help(self) -> None:
+        help_string = f"{self.rules.help}\n\n"
+        help_string += "__Available keywords per rule__:\n\n"
+
+        full_rules = await self.bot.api_client.get("rules", params={"link_format": "md"})
+
+        for index, (_, keywords) in enumerate(full_rules, start=1):
+            help_string += f"**Rule n°{index}**: {', '.join(keywords)}\n\n"
```
wookie184, line 533: I don't think we need the `n°` bit, just e.g. `Rule 1` would be fine as that's how we style it elsewhere (in the `#rules` channel on the server)

wookie184 (changes requested):
Nice, this seems like a good solution.

I'm wondering if the styling might look a bit better without empty lines between rules, thoughts?

bot/exts/info/information.py
shtlrs: Oh, that's strange because I searched in the logs and I didn't find it.
Maybe because I was having issues in the first iteration.
Removed !

bot/exts/info/information.py
shtlrs: Alrighty ! :D

shtlrs:
> I'm wondering if the styling might look a bit better without empty lines between rules, thoughts?

Humm, I thought it's too jammed, here's an overview
image

shtlrs:
@wookie184 have you had the chance to look at the format you requested?

wookie184:
> @wookie184 have you had the chance to look at the format you requested? 

I think it's alright and doesn't look too jammed, although I don't really mind. Let's leave it as is for now though and see what the next reviewer thinks.

wookie184 (approved):
LGTM!

ChrisLovering (approved):

PR #2310, opened 2022-10-25
Fix for rules greedy parsing freezing

ionite34:
Command works the same as the original specifications, bug is fixed for now.

img

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -524,13 +524,16 @@ class Information(Cog):
         await self.send_raw_content(ctx, message, json=True)
 
     @command(aliases=("rule",))
-    async def rules(self, ctx: Context, *args: Optional[str]) -> Optional[Set[int]]:
+    async def rules(self, ctx: Context, arg: Optional[str]) -> Optional[Set[int]]:
         """
         Provides a link to all rules or, if specified, displays specific rule(s).
 
         It accepts either rule numbers or particular keywords that map to a particular rule.
         Rule numbers and keywords can be sent in any order.
         """
+        # Temporary fix for discord.py greedy string quote conversion bug
+        args = (arg or "",)
+
         rules_embed = Embed(title="Rules", color=Colour.og_blurple(), url="https://www.pythondiscord.com/pages/rules")
         keywords, rule_numbers = [], []
 
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -531,9 +531,6 @@ class Information(Cog):
         It accepts either rule numbers or particular keywords that map to a particular rule.
         Rule numbers and keywords can be sent in any order.
         """
-        # Note: *args cannot be used due to a discord.py bug
-        # causing infinite loops during greedy string parsing.
-
         rules_embed = Embed(title="Rules", color=Colour.og_blurple(), url="https://www.pythondiscord.com/pages/rules")
         keywords, rule_numbers = [], []
 
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -546,8 +546,10 @@ class Information(Cog):
                 try:
                     rule_numbers.append(int(word))
                 except ValueError:
-                    if (kw := word.lower()) in keyword_to_rule_number:
-                        keywords.append(kw)
+                    # Stop on first invalid keyword/index to allow for normal messaging after
+                    if (kw := word.lower()) not in keyword_to_rule_number:
+                        break
+                    keywords.append(kw)
 
         if not rule_numbers and not keywords:
             # Neither rules nor keywords were submitted. Return the default description.
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -542,7 +542,7 @@ class Information(Cog):
                 keyword_to_rule_number[rule_keyword] = rule_number
 
         if args:
-            for word in args.split():
+            for word in args.split(maxsplit=100):
                 try:
                     rule_numbers.append(int(word))
                 except ValueError:
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -524,7 +524,7 @@ class Information(Cog):
         await self.send_raw_content(ctx, message, json=True)
 
     @command(aliases=("rule",))
-    async def rules(self, ctx: Context, arg: Optional[str]) -> Optional[Set[int]]:
+    async def rules(self, ctx: Context, args: Optional[str]) -> Optional[Set[int]]:
         """
         Provides a link to all rules or, if specified, displays specific rule(s).
 
@@ -532,7 +532,8 @@ class Information(Cog):
         Rule numbers and keywords can be sent in any order.
         """
         # Temporary fix for discord.py greedy string quote conversion bug
-        args = (arg or "",)
+        if not args:
+            args = ("",)
 
         rules_embed = Embed(title="Rules", color=Colour.og_blurple(), url="https://www.pythondiscord.com/pages/rules")
         keywords, rule_numbers = [], []
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -524,22 +524,15 @@ class Information(Cog):
         await self.send_raw_content(ctx, message, json=True)
 
     @command(aliases=("rule",))
-    async def rules(self, ctx: Context, args: Optional[str]) -> Optional[Set[int]]:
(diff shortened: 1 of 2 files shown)


MarkKoz (approved):

bot/exts/info/information.py
```diff
@@ -524,13 +524,17 @@ async def json(self, ctx: Context, message: Message) -> None:
...
+    async def rules(self, ctx: Context, args: Optional[str]) -> Optional[Set[int]]:
         """
         Provides a link to all rules or, if specified, displays specific rule(s).
 
         It accepts either rule numbers or particular keywords that map to a particular rule.
         Rule numbers and keywords can be sent in any order.
         """
+        # Temporary fix for discord.py greedy string quote conversion bug
+        if not args:
+            args = ("",)
```
MarkKoz, line 536: Shouldn't this be an empty tuple?

MarkKoz (changes requested):

bot/exts/info/information.py
MarkKoz:
```py
if args is None:
    args = ()
else:
    args = (args,)
```

Second case is important too since the rest of the code expects `args` to be an iterable of strings.

tests/bot/exts/info/test_information.py
```diff
@@ -614,7 +614,7 @@ async def test_return_none_if_one_rule_number_is_invalid(self):
                     str(rule_number) for rule_number in extracted_rule_numbers
                     if rule_number < 1 or rule_number > len(self.full_rules))
 
-                final_rule_numbers = await self.cog.rules(self.cog, self.ctx, *raw_user_input)
+                final_rule_numbers = await self.cog.rules(self.cog, self.ctx, raw_user_input)
```
MarkKoz, line 617: It's incorrect to merely not unpack it. The command expects one string now, not a tuple.

MarkKoz (changes requested):

onerandomusername:
Here is a hotpatch if you're interested: 
```py
from disnake.ext.commands.view import StringView
from disnake.ext.commands.errors import ArgumentParsingError

_original_get_quoted_word = StringView.get_quoted_word

def get_quoted_word(self:StringView):
    try:
        _original_get_quoted_word(self)
    except ArgumentParsingError:
        raise RuntimeError

StringView.get_quoted_word = get_quoted_word
```

bot/exts/info/information.py
ionite34: > Shouldn't this be an empty tuple?

Fixed in latest push



> I think it should be
> 
> ```python
> if args is None:
>     args = ()
> else:
>     args = (args,)
> ```
> 
> Second case is important too since the rest of the code expects `args` to be an iterable of strings.

Have changed args to be derived from ctx.message instead

tests/bot/exts/info/test_information.py
ionite34: With the new ctx inspection this seems to be the easiest way to keep the tests working. It would be more elegant to rewrite the tests to use the ctx object, but in any case, the tests aren't currently effectively testing the convert portions anyways? (Hence the issue). So I think the test rewrite that also tests the converter can come later?

MarkKoz:
Err yes, that would probably be better. Trying to make the command work for both actual and test inputs is kinda wack, and properly modifying the test seems like a non-trivial amount of effort.

MarkKoz:
There seems to be code that handles `ArgumentParsingError`, so indisriminantely raising `RuntimeError` in its place seems like it could mess up the parsing of other things

https://github.com/Rapptz/discord.py/blob/01915fbc095a72ffd2231c72b7c0892a377a477c/discord/ext/commands/core.py#L698-L705

onerandomusername:
> There seems to be code that handles `ArgumentParsingError`, so indisriminantely raising `RuntimeError` in its place seems like it could mess up the parsing of other things
> 
> https://github.com/Rapptz/discord.py/blob/01915fbc095a72ffd2231c72b7c0892a377a477c/discord/ext/commands/core.py#L698-L705

Yep, it would. Oops.

ionite34:
> There seems to be code that handles `ArgumentParsingError`, so indisriminantely raising `RuntimeError` in its place seems like it could mess up the parsing of other things
> 
> [Rapptz/discord.py@`01915fb`/discord/ext/commands/core.py#L698-L705](https://github.com/Rapptz/discord.py/blob/01915fbc095a72ffd2231c72b7c0892a377a477c/discord/ext/commands/core.py#L698-L705)

Yeah the monkey patch seems to no longer infinitely loop but breaks string parsing in most other places as well:

image

Is there something else we can do for patching?

tests/bot/exts/info/test_information.py
ionite34: Removed the hacks for much simpler fix in [`71fdd6c` (#2310)](https://github.com/python-discord/bot/pull/2310/commits/71fdd6cfd4e5353be61a46a2ab629835238122fc)

ionite34:
New much simpler fix in [`71fdd6c` (#2310)](https://github.com/python-discord/bot/pull/2310/commits/71fdd6cfd4e5353be61a46a2ab629835238122fc) as discovered by @ChrisLovering in https://canary.discord.com/channels/267624335836053506/675756741417369640/1034385695940296704

bot/exts/info/information.py
```diff
@@ -524,13 +524,16 @@ async def json(self, ctx: Context, message: Message) -> None:
...
+    async def rules(self, ctx: Context, *, args: Optional[str]) -> Optional[Set[int]]:
         """
         Provides a link to all rules or, if specified, displays specific rule(s).
 
         It accepts either rule numbers or particular keywords that map to a particular rule.
         Rule numbers and keywords can be sent in any order.
         """
+        # Note: *args cannot be used due to a discord.py bug
+        # causing infinite loops during greedy string parsing.
+
```
ChrisLovering, line 536: I don't think this comment is needed, as *args in a command shouldn't be a thing that's used, and having a comment here doesn't stop someone adding it to a new command, that can only be done in PRs.

bot/exts/info/information.py
```diff
@@ -541,13 +544,14 @@ async def rules(self, ctx: Context, *args: Optional[str]) -> Optional[Set[int]]:
...
-                    break
-                keywords.append(kw)
+        if args:
+            for word in args.split():
+                try:
+                    rule_numbers.append(int(word))
+                except ValueError:
+                    if (kw := word.lower()) not in keyword_to_rule_number:
+                        break
+                    keywords.append(kw)
```
ChrisLovering, line 554: I know you didn't add this code in this PR, but it feel wrong? Surely we don't want to break and stop processing the rest of the args if just one of them doesn't match. I think this would be better, so even if one doens't match, we still grab out the rest.
```suggestion
                    if (kw := word.lower()) in keyword_to_rule_number:
                        keywords.append(kw)
```

ChrisLovering (changes requested):

bot/exts/info/information.py
ionite34: I'm thinking maybe it's to do with time spent in checking messages if someone decides to do a `rules` with 4000 chars or something? Would it be more sensible to impose a parsing limit of, say 50 words?

ChrisLovering (approved):

wookie184 (changes requested):
I think the change in behaviour should be reverted. If we want to change it it can be done in a separate PR after discussing it, but it's out of scope for this PR given the original behaviour was intended

ChrisLovering:
> I think the change in behaviour should be reverted. If we want to change it it can be done in a separate PR after discussing it, but it's out of scope for this PR given the original behaviour was intended

Ah right, we should definitely add a comment there explaining that use case then, as it currently just looks wrong to anyone who isn't aware of that.

onerandomusername:
> New much simpler fix in [`71fdd6c` (#2310)](https://github.com/python-discord/bot/pull/2310/commits/71fdd6cfd4e5353be61a46a2ab629835238122fc) as discovered by @ChrisLovering in https://canary.discord.com/channels/267624335836053506/675756741417369640/1034385695940296704

would you please provide the context for this link that isn't public?

shtlrs:
@ionite34  @ChrisLovering here's the - original [issue](https://github.com/python-discord/bot/issues/2108) where you'll find more details and context in the discussion / PR

wookie184 (approved):
Tested, lgtm

PR #2315, opened 2022-10-28
Pass args to the Python version switch button

### LINKED ISSUE #2299: timeit does not evaluate setup code when using the version switcher button
When using the timeit command with a setup block and clicking the "Run in 3.10/3.11" button, any variables previously defined in the setup are no longer present, typically producing NameErrors in the resulting job:

<p float="left">
  
  
</p>

It appears that the [`PythonVersionSwitcherButton`](https://github.com/python-discord/bot/blob/e453a1648111a93765b6d0b0647f9b32072d316c/bot/exts/utils/snekbox.py#L119) does not store the commandline arguments used in the initial execution which leaves out the setup code that's meant to be evaluated with the job. Passing these arguments when the button gets constructed should be a feasible solution.

ks129:
Closes #2299 

Pass arguments to new execution when switching the Python version.

### DIFF
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -125,7 +125,8 @@ class PythonVersionSwitcherButton(ui.Button):
         version_to_switch_to: Literal["3.10", "3.11"],
         snekbox_cog: "Snekbox",
         ctx: Context,
-        code: str
+        code: str,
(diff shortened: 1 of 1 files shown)


wookie184 (approved):
Tested and seems to fix the bug, thanks 👍

ChrisLovering (approved):

PR #2316, opened 2022-10-29
Ask for confirmation when banning members with elevated roles

### LINKED ISSUE #2314: Safeguards against accidentally banning staff members
To avoid any more situations like these:
image

Propose to add:

>### :warning:  This action will apply **ban** to a staff member, `@user` (User Name #8405). To confirm, rerun the command with `--force`

Robin5605:
Closes #2314 

Created a subclass of `botcore.utils.Interactions.ViewWithUserAndRoleCheck` that will be sent when a moderator attempts to ban staff (helper role or above) in `bot/exts/moderation/infraction/_utils.py`. Only the `!ban` command author and moderators are allowed to interact with this view. If cancelled, the view is deleted (NOTE: not the entire message). Similarly, if the timeout (10 seconds) is reached, the view will be deleted (once again, not the entire message). If confirmed, the user will be banned as usual.

Waiting for input:
image

After timeout **or** cancel:
image


bot/exts/moderation/infraction/infractions.py
```diff
@@ -480,6 +474,45 @@ async def apply_ban(
...
+                )
+
+            elif confirmation_view.value is False:
+                log.trace(
+                    "Attempted ban of user %s by moderator %s cancelled due to manual cancel.",
+                    str(user),
+                    str(ctx.author),
+                )
+
+            return None
```
CaedenPH, line 504: Am I missing something? You're `returning` if the value is `True` aka even if the staff confirms the ban it will exit the function

bot/exts/moderation/infraction/_utils.py
```diff
@@ -269,3 +272,43 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+            allowed_roles=allowed_roles,
+            timeout=timeout,
+            message=message
+        )
+
+        self.value: t.Optional[bool] = None
+
+    async def send(self, channel: TextChannel, *, message_content: str) -> None:
+        """Send this view to the given channel with the given message content."""
+        self.message = await channel.send(message_content, view=self)
```
CaedenPH, line 299: Why do you need to assign the message to `self.message`? 
It doesn't seem used to me

bot/exts/moderation/infraction/_utils.py
```diff
@@ -269,3 +272,43 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        allowed_users: t.Sequence[int],
+        allowed_roles: t.Sequence[int],
+        timeout: t.Optional[float] = 180,
+        message: t.Optional[Message] = None
+    ) -> None:
+        super().__init__(
+            allowed_users=allowed_users,
+            allowed_roles=allowed_roles,
+            timeout=timeout,
+            message=message
```
CaedenPH, line 292: What does this do? You don't pass `message` in any scenario and I can't see the purpose of it.
I also don't see it being used

bot/exts/moderation/infraction/_utils.py
Robin5605: The superclass uses it in their `on_timeout` function: https://github.com/python-discord/bot-core/blob/main/botcore/utils/interactions.py#L65-L70
```py
    async def on_timeout(self) -> None:
        """Remove the view from ``self.message`` if set."""
        if self.message:
            with contextlib.suppress(NotFound):
                # Cover the case where this message has already been deleted by external means
                await self.message.edit(view=None)
```

bot/exts/moderation/infraction/infractions.py
Robin5605: Right. Logical fallacy on my part. Will fix ASAP

bot/exts/moderation/infraction/_utils.py
CaedenPH: > What does this do? You don't pass `message` in any scenario and I can't see the purpose of it. I also don't see it being used

I see the point of `self.message` now, however you don't need to pass it into the `super` class
```suggestion
```

Robin5605: You're absolutely right. I was debating this myself, but I decided to just leave it and have a 1:1 between arguments of the parent class (`ViewWithUserAndRoleCheck`) and the subclass (`StaffBanConfirmationView`). It really can't hurt and it's there in case we ever need it if we decide to re-use the same class elsewhere (we might for other commands, like cleanban, warnings, mutes, etc)

bot/exts/moderation/infraction/_utils.py
```diff
@@ -269,3 +272,43 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        )
+
+        self.value: t.Optional[bool] = None
+
+    async def send(self, channel: TextChannel, *, message_content: str) -> None:
```
CaedenPH, line 297:
```suggestion
    async def send(self, channel: TextChannel, message_content: str) -> None:
```
In my opinion `message_content` doesn't need to be keyword
We can see what other people think as well

bot/exts/moderation/infraction/infractions.py
```diff
@@ -480,6 +474,47 @@ async def apply_ban(
...
+        if isinstance(user, Member) and any(role.id in constants.STAFF_PARTNERS_COMMUNITY_ROLES for role in user.roles):
+            confirmation_view = _utils.StaffBanConfirmationView(
+                allowed_users=(ctx.author.id,),
+                allowed_roles=constants.MODERATION_ROLES,
+                timeout=10,
+            )
+
+            await confirmation_view.send(
+                ctx.channel,
+                message_content="That user is a Helper. Are you sure you want to ban them?",
```
Preocts, line 486: Probably want to use a general warning verbiage here as `STAFF_PARTNERS_COMMUNITY_ROLES` is more than just Helpers. I like that we warn for all of them.

```suggestion
                message_content=f"{user} has an elevated role. Are you sure you want to ban them?",
```

ionite34 (dismissed):
Tested, LGTM. Also, I dedicate this PR in honor of @AbooMinister25

bot/exts/moderation/infraction/_utils.py
```diff
@@ -269,3 +272,44 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        """Send this view to the given channel with the given message content."""
+        self.message = await channel.send(message_content, view=self)
+
+    @discord.ui.button(label="Confirm", style=ButtonStyle.green)
+    async def confirm(self, interaction: Interaction, button: Button) -> None:
+        """Callback coroutine that is called when the "confirm" button is pressed."""
```
shtlrs, line 303: Question: Can't the code used in `confirm` and `cancel` be moved to  a "private" method instead ? 
It's so similair, and the only thing changing is the value
```py
async __set_value(self, interaction: Interaction, button: Button, value: bool):
        await interaction.response.defer()
        self.value = value
        await super().on_timeout()
        self.stop()
   
@discord.ui.button(label="Cancel", style=ButtonStyle.red)
    async def cancel(self, interaction: Interaction, button: Button) -> None:
        """Callback coroutine that is called when the "cancel" button is pressed."""
        await self.__set_value(intraction, button, False)
 ```

File: bot/exts/moderation/infraction/_utils.py [hunk 7]
```diff
@@ -269,3 +272,44 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        await interaction.response.defer()
```
ChrisLovering, line 304: Is this defer needed?

bot/exts/moderation/infraction/_utils.py
```diff
@@ -269,3 +272,44 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        self.value = True
+        await super().on_timeout()
+        self.stop()
+
+    @discord.ui.button(label="Cancel", style=ButtonStyle.red)
+    async def cancel(self, interaction: Interaction, button: Button) -> None:
+        """Callback coroutine that is called when the "cancel" button is pressed."""
+        await interaction.response.defer()
```
ChrisLovering, line 312: Is this defer needed?

bot/exts/moderation/infraction/_utils.py
```diff
@@ -269,3 +272,44 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        self.value = False
```
ChrisLovering, line 313:
```suggestion
        self.confirmed = True
``` To make it more obvious what this instance var is doing.

File: bot/exts/moderation/infraction/infractions.py [hunk 10]
```diff
@@ -480,6 +474,47 @@ async def apply_ban(
         async def action() -> None:
             await ctx.guild.ban(user, reason=reason, delete_message_days=purge_days)
 
+        if isinstance(user, Member) and any(role.id in constants.STAFF_PARTNERS_COMMUNITY_ROLES for role in user.roles):
+            confirmation_view = _utils.StaffBanConfirmationView(
+                allowed_users=(ctx.author.id,),
```
ChrisLovering, line 479: having this allowed_users line is likely superfluous, since only MODERATION_ROLES can run this command inthe first place.

bot/exts/moderation/infraction/infractions.py
```diff
@@ -480,6 +474,47 @@ async def apply_ban(
...
+                allowed_roles=constants.MODERATION_ROLES,
+                timeout=10,
+            )
+
+            await confirmation_view.send(
+                ctx.channel,
+                message_content=f"{user} has an elevated role. Are you sure you want to ban them?",
+            )
```
ChrisLovering, line 487:
```suggestion
            message = await ctx.send(
                "{user} has an elevated role. Are you sure you want to ban them?",
                view=confirmation_view
            )
            confirmation_view.message = message
```

This way would be consistent with howt he view is used in snekbox, and removes the need for both the init and the send functions on the subclass StaffBanConfirmationView

ChrisLovering (changes requested):

bot/exts/moderation/infraction/_utils.py
Robin5605: The only reason I've included it is because setting the `self.value` to `None` by default. If I understand how subclassing works in Python, once you include an `__init__` of your own in a subclass, it overrides the parent's `__init__`, yes?

bot/exts/moderation/infraction/_utils.py
```diff
@@ -269,3 +272,44 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        super().__init__(
+            allowed_users=allowed_users,
+            allowed_roles=allowed_roles,
+            timeout=timeout,
+            message=message
+        )
+
+        self.value: t.Optional[bool] = None
+
+    async def send(self, channel: TextChannel, *, message_content: str) -> None:
```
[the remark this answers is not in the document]
Robin5605: None, really. More of a "convenience" thing than anything, if it's even any more convenient than doing it the other way 😛

bot/exts/moderation/infraction/_utils.py
Robin5605: It definitely could! Though in this case I'm not 100% sure if it's necessary to do so. It would be relatively little effort to extract the common logic into it's own function, but only 2 functions are sharing this logic.

bot/exts/moderation/infraction/_utils.py
Robin5605: Yes. We're not directly responding to button's interaction, but sending a separate message. If we want the ban to be a response of the button interaction, we'd have to pass it into the `apply_ban` function, which would need to have updated tests and update all references. I didn't think it to be worth it.

bot/exts/moderation/infraction/_utils.py
Robin5605: See above

bot/exts/moderation/infraction/_utils.py
ChrisLovering: Yea, but when we do send the seperate message, we don't use the interaction to do so right?

bot/exts/moderation/infraction/_utils.py
shtlrs: I wouldn't call it necessary of course.

The decision is totally up to you, thanks!

bot/exts/moderation/infraction/_utils.py
Robin5605: No, we don't. The message is just sent to the channel as a regular message. IIRC the interaction would "time out" even though we sent a message - since as far as Discord cares we didn't respond to their interaction

wookie184:
The linter error needs fixing

bot/exts/moderation/infraction/_utils.py
ChrisLovering: AH, I see what is going on. Discord.py usually deals with the response to the callback automatically, but since you're calling self.timeout() manually, that s stopping the interaction, so Discord.y does not respond to Discord to say the interaction was handled.

I think what we want to do here is
```python
await interaction.message.edit(view=None)
self.stop()
```
At the end of both button callbacks You could even edit the embed at the same time for feedback to the person who interacted with the button. EG Update it to cancelled,/confirmed, or even delete the bot message and send an ephemeral message.

bot/exts/moderation/infraction/infractions.py
Robin5605: Seems like `allowed_users` is a require argument, so I don't think any harm can come of it by just leave it

wookie184:
This works and the code looks good, although I have two possible suggestions:
- When the interaction times out the buttons should be removed, to avoid the "this interaction failed" message.
- There should be an acknowledgement that the ban was cancelled when clicking "cancel", e.g. editing the message or sending a followup message, to make it clear what happened.

Robin5605:
Sounds great. I'll implement that.

wookie184:
> When the interaction times out the buttons should be removed, to avoid the "this interaction failed" message.

I just realised that i'm not actually sure how this would be implemented, so it's probably fine to leave as is unless you have a simple idea.

ChrisLovering:
> > When the interaction times out the buttons should be removed, to avoid the "this interaction failed" message.
> 
> I just realised that i'm not actually sure how this would be implemented, so it's probably fine to leave as is unless you have a simple idea.

You can use [this view](https://bot-core.pythondiscord.com/main/output/pydis_core.utils.interactions.html#pydis_core.utils.interactions.ViewWithUserAndRoleCheck) from bot-core
```
view = pydis_core.utils.interactions.ViewWithUserAndRoleCheck(
    allowed_users = list_of_users,
    allowed_roles = list_of_roles,
    timeout = timeout,
)
view.message = await ctx.send("foo", view=view)
```

bot/exts/moderation/infraction/_utils.py
Preocts: There doesn't seem to be a pattern available to emit a yes/no response from the interaction. Only a success/timeout.

Unless we want to wrap the entire infraction step in a method, structure it as a callback, and hand it to the buttons for this interaction; I'm not seeing a way around toting the boolean flag `value` around. To complicate that hand-off, it looks like the result of the infraction call needs to be returned again from [the calling method here](https://github.com/python-discord/bot/blob/63c44a47c3e8bf15acf5d6babc2466664da6d63b/bot/exts/moderation/infraction/infractions.py#L528)

ChrisLovering: I realise I never responded to this. What I meant was we don't need this init function specifically, as there's no need to set `confirmed` to None.

when we wait for this view we can do `timed_out = await confirmation_view.wait()` then check `if timed_out` first, meaning that will cover the case where `confirmed` isn't set.

Robin5605:
Removed the unnecessary `__init__` as per @ChrisLovering 's suggestion and made the bot delete the buttons after the timeout as per @wookie184 's suggestion. Thanks!

bot/exts/moderation/infraction/_utils.py
```diff
@@ -280,3 +283,25 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
         return False
+
+
+class StaffBanConfirmationView(interactions.ViewWithUserAndRoleCheck):
+    """The confirmation view that is sent when a moderator attempts to ban a staff member."""
+
+    @discord.ui.button(label="Confirm", style=ButtonStyle.red)
+    async def confirm(self, interaction: Interaction, button: Button) -> None:
+        """Callback coroutine that is called when the "confirm" button is pressed."""
+        await interaction.response.send_message("Applied infraction.")
```
ChrisLovering, line 294: I don't think we need this message, since feedback is managed by the ban command itself.

image

bot/exts/moderation/infraction/_utils.py
```diff
@@ -280,3 +283,25 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        await interaction.message.edit(view=None)
```
ChrisLovering, line 295: Ah, this should be getting done by bot-core automatically, but I notice that it only gets done on timeout. I'll update bot-core so that it works on stop too, could you remove this line and the similar one in `cancel()`?

ChrisLovering (changes requested):
Just some minor changes

bot/exts/moderation/infraction/_utils.py
Robin5605: Should I also remove the response that says "Cancelled infraction" if the invoker hits the "Cancel" button?

bot/exts/moderation/infraction/_utils.py
Robin5605: @ChrisLovering I've tested this locally with the changes from https://github.com/python-discord/bot-core/pull/184 and it works, save having to change the `.stop` to be asynchronous. Might the change in https://github.com/python-discord/bot-core/pull/184 break this functionality anywhere else in the codebase?

bot/exts/moderation/infraction/_utils.py
ChrisLovering: Nah, the cancel makes sense as it is, it was just eh double feedback message from the confirm that looks strange to me.

bot/exts/moderation/infraction/_utils.py
ChrisLovering: Good point. I've just updated that PR to make `stop()` synchronous to avoid this issue.

ChrisLovering (dismissed):

bot/exts/moderation/infraction/_utils.py
```diff
@@ -280,3 +283,23 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
         return False
+
+
+class StaffBanConfirmationView(interactions.ViewWithUserAndRoleCheck):
+    """The confirmation view that is sent when a moderator attempts to ban a staff member."""
+
+    @discord.ui.button(label="Confirm", style=ButtonStyle.red)
+    async def confirm(self, interaction: Interaction, button: Button) -> None:
+        """Callback coroutine that is called when the "confirm" button is pressed."""
+        await interaction.response.defer() # `ViewWithuserAndRoleCheck` view will remove the buttons
```
wookie184, line 294: When I press confirm or cancel the buttons seem to remain on the embed, so i'm not sure that this is working. `ViewWithuserAndRoleCheck` only seems to remove them when the embed times out, but I think we also want to remove them when a button is pressed.

wookie184 (changes requested):

bot/exts/moderation/infraction/_utils.py
ichard26: Fixed by merging in `main` so the latest version of bot core is used which does implement the automatic view removal.

bot/exts/moderation/infraction/infractions.py
```diff
@@ -509,6 +503,34 @@ async def apply_ban(
...
+            timed_out = await confirmation_view.wait()
+            if timed_out:
+                log.trace(f"Attempted ban of user {user} by moderator {ctx.author} cancelled due to timeout.")
+                return None
+
+            if confirmation_view.confirmed is False:
+                log.trace(f"Attempted ban of user {user} by moderator {ctx.author} cancelled due to manual cancel.")
+                return None
+
+        infraction = await _utils.post_infraction(ctx, user, "ban", reason, active=True, **kwargs)
```
wookie184, line 528: Moving this below
```
        if reason:
            reason = textwrap.shorten(reason, width=512, placeholder="...")
```
causes a change in behaviour.

I assume the behaviour we want is for the full message to be sent to the database, but the reason to be trimmed in the audit log (I assume because there's a limit there), but now it's also trimmed in the database.

bot/exts/moderation/infraction/_utils.py
```diff
@@ -298,3 +300,23 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+
+class StaffBanConfirmationView(interactions.ViewWithUserAndRoleCheck):
+    """The confirmation view that is sent when a moderator attempts to ban a staff member."""
+
+    @discord.ui.button(label="Confirm", style=ButtonStyle.red)
+    async def confirm(self, interaction: Interaction, button: Button) -> None:
+        """Callback coroutine that is called when the "confirm" button is pressed."""
+        await interaction.response.defer()
+
+        self.confirmed = True
```
wookie184, line 313: I think it would be nice to initialise `self.confirmed` to `False` in the `__init__`. It shouldn't make a difference with the current logic but it makes the code a bit easier to reason about as the attribute is always defined.

wookie184 (changes requested):
Maybe we could send a reply indicating if the view timed out, like there is if it's cancelled manually, though I don't really mind.

bot/exts/moderation/infraction/infractions.py
wookie184: We should modify the tests to catch this too https://github.com/python-discord/bot/blob/36a9ac91f51a3507d1609c899ccf54bb4c0f3dc6/tests/bot/exts/moderation/infraction/test_infractions.py#L16

You can do that here if you want otherwise I can open an issue for it.

ichard26: See #2955.

ichard26:
Screenshots of the new behaviour:

image

image

bot/exts/moderation/infraction/_utils.py
```diff
@@ -298,3 +300,26 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
             "The user either could not be retrieved or probably disabled their DMs."
         )
         return False
+
+
+class StaffBanConfirmationView(interactions.ViewWithUserAndRoleCheck):
+    """The confirmation view that is sent when a moderator attempts to ban a staff member."""
+
+    confirmed = True
```
ChrisLovering, line 308: Could we flip this, and make the default `False`, setting it to True when confirmed.

Shouldn't be an issue now, it jsut means that it fails safe if we have a bug around her ein teh future.

ChrisLovering (changes requested):

bot/exts/moderation/infraction/_utils.py
ichard26: I'm a big fan of defensive programming, stgm. Thanks!

ChrisLovering (approved):

bot/exts/moderation/infraction/_utils.py
```diff
@@ -298,3 +300,26 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        await interaction.response.defer()
+        self.stop()
+
+    @discord.ui.button(label="Cancel", style=ButtonStyle.green)
+    async def cancel(self, interaction: Interaction, button: Button) -> None:
+        """Callback coroutine that is called when the "cancel" button is pressed."""
+        await interaction.response.send_message("Cancelled infraction.")
+        self.stop()
+
+    async def on_timeout(self) -> None:
```
MarkKoz, line 323: Good place to use the `typing.override` decorator

https://docs.python.org/3/library/typing.html#typing.override

bot/exts/moderation/infraction/_utils.py
```diff
@@ -298,3 +300,26 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
             "The user either could not be retrieved or probably disabled their DMs."
         )
         return False
+
+
+class StaffBanConfirmationView(interactions.ViewWithUserAndRoleCheck):
+    """The confirmation view that is sent when a moderator attempts to ban a staff member."""
+
+    confirmed = False
```
MarkKoz, line 308: Shouldn't this be an instance attribute? Maybe this is fine if the instance is not re-used, but it's not clear this class's instances are not intended to be re-used nor would we have a good way to enforce that.

bot/exts/moderation/infraction/_utils.py
```diff
@@ -298,3 +300,26 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
             "The user either could not be retrieved or probably disabled their DMs."
         )
         return False
+
+
+class StaffBanConfirmationView(interactions.ViewWithUserAndRoleCheck):
+    """The confirmation view that is sent when a moderator attempts to ban a staff member."""
```
MarkKoz, line 306: This view isn't inherently only for bans - it looks like it could work with any infraction if we wanted it to. I think this name could be more generic.

Also, should this be moved to its own module (e.g. a module for all-things views) - are we putting too much stuff into "utils"?

bot/exts/moderation/infraction/infractions.py
```diff
@@ -495,20 +495,41 @@ async def apply_ban(
             log.trace("Old tempban is being replaced by new permaban.")
             await self.pardon_infraction(ctx, "ban", user, send_msg=is_temporary)
 
+        # If user has an elevated role (staff, partner, or community), require
+        # confirmation before banning.
+        if isinstance(user, Member) and any(role.id in constants.STAFF_PARTNERS_COMMUNITY_ROLES for role in user.roles):
+            confirmation_view = _utils.StaffBanConfirmationView(
```
MarkKoz, line 501: Thoughts on refactoring this whole `if` block into its own function? This `apply_ban` function is growing large.

MarkKoz (approved):
Just some nits

bot/exts/moderation/infraction/_utils.py
wookie184: We're not on 3.12 yet, so can't use it yet (we were blocked until very recently waiting for dependencies to update).

bot/exts/moderation/infraction/_utils.py
ichard26: I'm not sure if making it an instance attribute nets us anything as the confirmed flag wouldn't be reset to false either way? But sure, why not.

ichard26:
I'm not really happy with how long this PR and review revisions has taken; I sincerely hope this is close to be able to be merged. (Albeit if you have any more nits, please send them over.)

bot/exts/moderation/infraction/_utils.py
MarkKoz: I didn't correctly word what I was thinking. If it's a class attribute, then the state is persisted across all instances. This means `confirmed` will remain True for subsequent command invocations, right? 

You're correct that as an instance attribute, the instance cannot be re-used because it's never reset. But I think an instance attribute solves the above case.

MarkKoz (approved):

bot/exts/moderation/infraction/_utils.py
ichard26:
```python
ichard26@asus-ubuntu:~$ cat test.py
class A:
    flag = False

a = A()
a2 = A()
a.flag = True
print(a.flag, a2.flag)
ichard26@asus-ubuntu:~$ python test.py
True False
```
As far as I understand, unless the class attribute is directly modified via the class, state won't persist across all instances.

MarkKoz: Okay yeah, that makes sense now that I see it. Thanks for clearing that up - was not intuitive to me at a glance.

wookie184 (approved):
All good, thanks for working on this.

PR #2319, opened 2022-11-06
Link previous nomination threads to a user nomination's history

### LINKED ISSUE #2304: Nominations: Link previous nomination threads
It would be nice if we could add links to previous nomination threads to both !tp history and newly created nomination threads. As it stands, we have to search for the username in the discord thread history (and this isn't particularly great UX nor reliable). 

Storing the thread ID for a nomination sounds like a good first step (and probably worth committing immediately even before other features are complete?), then searching through threads by username to attempt to rediscover old threads. Since usernames are changeable, we might also want to detect "desynced" talentpool history, where we know there have been X previous nominations, but only found less-than-X threads. 

We could also search through threads by time--we know when previous nominations have _ended_, but we don't know when they start, so we can use this also to help figure out which threads might be relevant by `archive_timestamp` or `last_pin_timestamp`, the latter of which might actually be extremely reliable (given reviews are auto-pinned in their thread).

shtlrs:
**Note to core devs**:
This [PR](https://github.com/python-discord/site/pull/792) needs to be merged before this one

Closes #2304 

As requested by @bast0006, we want to be able to go back easily in time and look for all the nomination threads of a particular user.

This currently links all the previous threads (as long as they haven't been deleted) to the nomination history of a particular user.
It also lists the thread ids in the `.md` file when getting a review of a user.

**Preview**:

**History command**
image

**Get review command**

image

**UPDATE**

After the below mentioned issue discovered by wookie when it comes to cached threads, we decided to go with the thread's `jump_url` instead

Here's a preview
image


bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -489,13 +489,20 @@ async def _nomination_to_string(self, nomination: Nomination) -> str:
         entries_string = "\n\n".join(entries)
 
         start_date = time.discord_timestamp(nomination.inserted_at)
+        thread = await self.bot.fetch_channel(nomination.thread_id)
+        if thread:
+            thread_mention = thread.mention
+        else:
+            thread_mention = "Thread hasn't been created yet or has been deleted"
```
shtlrs, line 496: This could simply become: "Thread not found".
All opinions are welcome

shtlrs (commented):
One small suggestion to be considered once opinions have been given.

wookie184:
After some testing i've found a bit of an issue, it seems discord will not render thread mentions if they aren't in the user's cache and they will show as `#deleted-channel`:

image.


I think the best option would be to just show the `jump_url` of the thread instead as they work fine.

shtlrs:
> After some testing i've found a bit of an issue, it seems discord will not render thread mentions if they aren't in the user's cache and they will show as `#deleted-channel`:
> 
> image.
> 
> I think the best option would be to just show the `jump_url` of the thread instead as they work fine.

That's strange, is there a way to clear the cache so I can test this out ? 
Also, I'll try with the jump_url

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -489,13 +489,21 @@ async def _nomination_to_string(self, nomination: Nomination) -> str:
         entries_string = "\n\n".join(entries)
 
         start_date = time.discord_timestamp(nomination.inserted_at)
+
+        if nomination.thread_id:
+            thread = await self.bot.fetch_channel(nomination.thread_id)
```
wookie184, line 494: I think we need to handle if this is `None` as well (which could be the case if the thread was set but then deleted).

Could we also use the `get_or_fetch_channel` utility function (I forget where it's defined, something like `bot.utils.channel`). This would avoid an API call if the thread is cached.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -489,13 +489,21 @@ async def _nomination_to_string(self, nomination: Nomination) -> str:
...
+            thread_jump_url = f'[Jump to thread!]({thread.jump_url})'
+        else:
+            thread_jump_url = "**Thread hasn't been created yet or has been deleted**"
```
wookie184, line 497: If the thread was deleted `thread_id` would still be set, so I think we can just say `*Not created*` here, so it reads something like `Vote thread: *Not Created*`

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -433,11 +433,15 @@ async def _previous_nominations_review(self, member: Member) -> Optional[str]:
 
         nomination_times = f"{num_entries} times" if num_entries > 1 else "once"
         rejection_times = f"{len(history)} times" if len(history) > 1 else "once"
+        nomination_vote_threads = ", ".join(
+            [f"<#{nomination.thread_id}>" if nomination.thread_id else '' for nomination in history]
```
wookie184, line 437: I think we need to use the jump URL here too. It's a bit annoying as it's quite long and we can't use the link syntax because it's not in an embed, but I think it should be fine.

wookie184 (changes requested):
Haven't tried it out yet but looks good, just a couple comments.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -489,13 +490,22 @@ async def _nomination_to_string(self, nomination: Nomination) -> str:
         entries_string = "\n\n".join(entries)
 
         start_date = time.discord_timestamp(nomination.inserted_at)
+
+        thread = None
+
+        if nomination.thread_id:
+            thread = await get_or_fetch_channel(nomination.thread_id)
+
+        thread_jump_url = f'[Jump to thread!]({thread.jump_url})' if thread else "*Not created*"
```
wookie184, line 499: We should catch the error if the channel isn't found. Could write it something like this:

```suggestion
        thread_jump_url = "*Not created*"
        if nomination.thread_id:
            try:
                thread = await get_or_fetch_channel(nomination.thread_id)
            except discord.HTTPException:
                thread_jump_url = "*Not found*"
            else:
                thread_jump_url = f'[Jump to thread!]({thread.jump_url})'
```

or

```suggestion
        thread_jump_url = "*Not created*"
        if nomination.thread_id:
            with contextlib.suppress(discord.HTTPException):
                thread = await get_or_fetch_channel(nomination.thread_id)
                thread_jump_url = f'[Jump to thread!]({thread.jump_url})'
```

(neither tested)

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -433,11 +434,20 @@ async def _previous_nominations_review(self, member: Member) -> Optional[str]:
...
         nomination_times = f"{num_entries} times" if num_entries > 1 else "once"
         rejection_times = f"{len(history)} times" if len(history) > 1 else "once"
+        threads = [await get_or_fetch_channel(nomination.thread_id) for nomination in history]
+
+        nomination_vote_threads = ", ".join(
+            [
+                f"[Thread-{thread_number}]({thread.jump_url})" if thread else ''
+                for thread_number, thread in enumerate(threads, start=1)
+            ]
+        )
```
wookie184, line 444: We should also catch the error if the channel isn't found here. Also, we can't use the `[]()` syntax for links as this isn't sent in an embed.

wookie184 (changes requested):

bot/exts/recruitment/talentpool/_review.py
wookie184: Still this 
> Also, we can't use the `[]()` syntax for links as this isn't sent in an embed.

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -433,11 +435,28 @@ async def _previous_nominations_review(self, member: Member) -> Optional[str]:
 
         nomination_times = f"{num_entries} times" if num_entries > 1 else "once"
         rejection_times = f"{len(history)} times" if len(history) > 1 else "once"
+        thread_jump_urls = []
+
+        for nomination in history:
+            try:
+                thread = await get_or_fetch_channel(nomination.thread_id)
```
wookie184, line 442: Should check for it being none first
```suggestion

        for nomination in history:
            if nomination.thread_id is None:
                continue
            try:
                thread = await get_or_fetch_channel(nomination.thread_id)
```

wookie184 (changes requested):
Have tested, one more thing and this looks good 👍

wookie184 (approved):
Thanks

Xithrius (approved):
LGTM

Xithrius:
Since [this](https://github.com/python-discord/site/pull/792) PR that this one depends on has been merged, we can now merge this one as well.

wookie184:
I'm confused, it seems like half of the PR was removed in this merge commit https://github.com/python-discord/bot/pull/2319/commits/f2c138fd87afd42c21325a55d1f57a1d29452813, so what was merged is incorrect

PR #2320, opened 2022-11-10
Update intents tag to better reflect discord.py's API changes

### LINKED ISSUE #2313: Update d.py intents tag to clarify that privileged intents require activation in d.py 2.0
Raise by `Roie#0089` here: <https://discord.com/channels/267624335836053506/267624335836053506/1035515307818754129>

As of Discord.py 2.0, privileged intents are now required to be explicitly activated.
We would like to add something in the tag body to help clarify this - for example -
> Starting from version 2.0, all intents must now be passed into the bot constructor as a required keyword argument intents and privileged intents must also be explicitly enabled on the developer portal.

thegamecracks:
To resolve #2313, this PR adds a short disclaimer near the top of the intents tag indicating that discord.py v2.0.0 requires its users to specify intents.
The starting paragraphs have also been revised to provide a clearer distinction between Discord's standard and privileged intents.
The code snippet now demonstrates the `message_content` intent since it is (generally) a requirement when writing prefix commands, especially with `commands.Bot`.

### DIFF
--- a/bot/resources/tags/intents.md
+++ b/bot/resources/tags/intents.md
@@ -1,19 +1,17 @@
 **Using intents in discord.py**
 
-Intents are a feature of Discord that tells the gateway exactly which events to send your bot. By default discord.py has all intents enabled except for `Members`, `Message Content`, and `Presences`. These are needed for features such as `on_member` events, to get access to message content, and to get members' statuses.
+Intents are a feature of Discord that tells the gateway exactly which events to send your bot. Various features of discord.py rely on having particular intents enabled. Since discord.py v2.0.0, this has become **mandatory** for developers to define in their code.
 
-To enable one of these intents, you need to first go to the [Discord developer portal](https://discord.com/developers/applications), then to the bot page of your bot's application. Scroll down to the `Privileged Gateway Intents` section, then enable the intents that you need.
+There are *standard* intents and *privileged* intents. The current privileged intents are `Presences`, `Server Members`, and `Message Content`. To use one of the privileged intents, you have to first enable them in the [Discord Developer Portal](https://discord.com/developers/applications). Go to the `Bot` page of your application, scroll down to the `Privileged Gateway Intents` section, and enable the privileged intents that you need.
 
-Next, in your bot you need to set the intents you want to connect with in the bot's constructor using the `intents` keyword argument, like this:
-
+Afterwards in your code, you need to set the intents you want to connect with in the bot's constructor using the `intents` keyword argument, like this:
 ```py
 from discord import Intents
 from discord.ext import commands
 
 intents = Intents.default()
-intents.members = True
+intents.message_content = True
 
 bot = commands.Bot(command_prefix="!", intents=intents)
 ```
-
 For more info about using intents, see the [discord.py docs on intents](https://discordpy.readthedocs.io/en/latest/intents.html), and for general information about them, see the [Discord developer documentation on intents](https://discord.com/developers/docs/topics/gateway#gateway-intents).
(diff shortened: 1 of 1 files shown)


bot/resources/tags/intents.md
```diff
@@ -1,19 +1,17 @@
...
 
-To enable one of these intents, you need to first go to the [Discord developer portal](https://discord.com/developers/applications), then to the bot page of your bot's application. Scroll down to the `Privileged Gateway Intents` section, then enable the intents that you need.
-
-Next, in your bot you need to set the intents you want to connect with in the bot's constructor using the `intents` keyword argument, like this:
+There are *standard* intents and *privileged* intents. The current privileged intents are `Presences`, `Server Members`, and `Message Content`. To use one of the privileged intents, you have to first enable them in the [Discord Developer Portal](https://discord.com/developers/applications). Go to the `Bot` page of your application, scroll down to the `Privileged Gateway Intents` section, and enable the privileged intents that you need.
 
+Afterwards in your code, you need to set the intents you want to connect with in the bot's constructor using the `intents` keyword argument, like this:
```
shtlrs, line 7: We might want to add the fact that the privileged intents are initialized to `False` by default here, to make sure no ones expects it to be the other way around.

bot/resources/tags/intents.md
```diff
@@ -1,19 +1,17 @@
...
+Intents are a feature of Discord that tells the gateway exactly which events to send your bot. Various features of discord.py rely on having particular intents enabled. Since discord.py v2.0.0, this has become **mandatory** for developers to define in their code.
 
-To enable one of these intents, you need to first go to the [Discord developer portal](https://discord.com/developers/applications), then to the bot page of your bot's application. Scroll down to the `Privileged Gateway Intents` section, then enable the intents that you need.
-
-Next, in your bot you need to set the intents you want to connect with in the bot's constructor using the `intents` keyword argument, like this:
+There are *standard* intents and *privileged* intents. The current privileged intents are `Presences`, `Server Members`, and `Message Content`. To use one of the privileged intents, you have to first enable them in the [Discord Developer Portal](https://discord.com/developers/applications). Go to the `Bot` page of your application, scroll down to the `Privileged Gateway Intents` section, and enable the privileged intents that you need.
```
shtlrs, line 5: We don't have access to the standard intents in the developer portal. What do you think about mentioning that ? We'd want to have people ask the question "Why don't I see these standard intents, am I doing something wrong ?"

We could say the the ones we can define in the portal are the privileged ones.
And maybe, if we want, we could add the link to a detailed technical list of these intents.

shtlrs (changes requested):
This looks good, thank you !
I just think that there are some subtle details we could add to make it more explicit.

bot/resources/tags/intents.md
thegamecracks: Maybe it would help to have a code comment explaining what .default() does?
```py
# Enable all standard intents and the `message_content` intent
intents = Intents.default()
intents.message_content = True
```

bot/resources/tags/intents.md
thegamecracks: Yeah, pointing to the [discord.Intents](https://discordpy.readthedocs.io/en/stable/api.html#intents) reference is a good idea. I'll link to it in the first paragraph.

I would think that how it's written right now would sufficiently imply that standard intents aren't part of the dev portal, but I can add a short sentence to clarify that.

shtlrs (approved):
Thanks @thegamecracks!

wookie184 (approved):
Looks good to me, could you fix the merge conflict?

wookie184:
Thanks!

Issue #1835, opened 2021-09-20
Put `!environments` and `!venv` in a tag group


TizzySaurus:
We currently have two tags, which both refer to virtual environments. These are `!environments` and `!venv`.

However, these two tags serve difference purposes (what a venv is and how to use one, respectively), so deleting one isn't really an option and they're too long to merge into a single tag.

The proposed solution for this (suggested by @Akarys42 which I agree with), is to wait for tag groups (#1663) is merged and then put these within a virtual environments tag.

PR #2322, opened 2022-11-11
Edited several tags

Keyacom:
- `environments.md`:
  - Added `envs` alias
  - Moved `Python Environments` to the embed's title field
  - Unwrapped `sys.executable` since it's a REPL example
- `identity.md`: Converted the example with lists to a snippet from a REPL session.
- `inline.md`: Added info about including a backtick in an inline code snippet.
- `listcomps.md`: `if` in list comprehensions is an expression, not a statement, so I changed that.
- `local-file.md`: Short explanation of `rb`, changed modal verb in a sentence
- `slicing.md`: The word "slicing" at the start is now in bold, not in italics.
- `traceback.md`: Added a line that appears since newer Python versions in tracebacks.
- `venv.md`:
  - Added `virtualenv` alias
  - Moved `Virtual Environments` to the embed's title
  - PowerShell is no longer called Windows PowerShell, since its core is cross-platform now. Instead, mentioned that it's about PowerShell in Windows.
  - `-Scope CurrentUser` will make the change apply just for the user, so mentioned that it's required *once per user*.
  - Changed the PowerShell snippet to make it have syntax highlighting

I felt like changing it since I saw what I could improve in `venv.md`, but decided to check more tags as well.

### DIFF
--- a/bot/resources/tags/listcomps.md
+++ b/bot/resources/tags/listcomps.md
@@ -10,7 +10,7 @@ Using list comprehensions can make this both shorter and more readable. As a lis
 >>> [n ** 2 for n in range(5)]
 [0, 1, 4, 9, 16]
 ```
-List comprehensions also get an `if` statement:
+List comprehensions also get an `if` expression:
 ```python
 >>> [n ** 2 for n in range(5) if n % 2 == 0]
 [0, 4, 16]
--- a/bot/resources/tags/listcomps.md
+++ b/bot/resources/tags/listcomps.md
@@ -10,8 +10,8 @@ Using list comprehensions can make this both shorter and more readable. As a lis
 >>> [n ** 2 for n in range(5)]
 [0, 1, 4, 9, 16]
 ```
-List comprehensions also get an `if` expression:
-```python
+List comprehensions also get an `if` clause:
+```py
 >>> [n ** 2 for n in range(5) if n % 2 == 0]
 [0, 4, 16]
 ```
--- a/bot/resources/tags/slicing.md
+++ b/bot/resources/tags/slicing.md
@@ -3,7 +3,7 @@ aliases: ["slice", "seqslice", "seqslicing", "sequence-slice", "sequence-slicing
 embed:
     title: "Sequence slicing"
 ---
-*Slicing* is a way of accessing a part of a sequence by specifying a start, stop, and step. As with normal indexing, negative numbers can be used to count backwards.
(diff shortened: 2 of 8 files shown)


shtlrs (approved):
Neat!
Thanks :D

bot/resources/tags/listcomps.md
```diff
@@ -10,7 +10,7 @@ Using list comprehensions can make this both shorter and more readable. As a lis
 >>> [n ** 2 for n in range(5)]
 [0, 1, 4, 9, 16]
 ```
-List comprehensions also get an `if` statement:
+List comprehensions also get an `if` expression:
```
CaedenPH, line 13: Why this change?

Keyacom: Statements must be separated by newlines or semicolons, that's why.

MarkKoz: Neither statement nor expression seem like appropriate descriptions since the `if` is not a standalone thing in the context of a comprehension. The [documentation](https://docs.python.org/3/reference/expressions.html#displays-for-lists-sets-and-dictionaries) calls them "`if` clauses".

Keyacom: I could modify so it's "clause", not "expression" nor "statement".

bot/resources/tags/slicing.md
```diff
@@ -3,7 +3,7 @@ aliases: ["slice", "seqslice", "seqslicing", "sequence-slice", "sequence-slicing
 embed:
     title: "Sequence slicing"
 ---
-*Slicing* is a way of accessing a part of a sequence by specifying a start, stop, and step. As with normal indexing, negative numbers can be used to count backwards.
+**Slicing** is a way of accessing a part of a sequence by specifying a start, stop, and step. As with normal indexing, negative numbers can be used to count backwards.
```
wookie184, line 6: I think italics was fine here personally, is there a reason for this change?

wookie184 (commented):
Thanks for this, the changes look good.

Side note: I didn't realise we have two tags for virtual environments, maybe we should combine them into one (not in this PR)

bot/resources/tags/slicing.md
Keyacom: Consistency with other tags.

CaedenPH (approved):

shtlrs:
> Side note: I didn't realise we have two tags for virtual environments, maybe we should combine them into one (not in this PR)

I think that's what Tizzy meant by opening #1835

wookie184 (approved):

PR #2323, opened 2022-11-12
Create an admin task when votes are 2 weeks old

### LINKED ISSUE #2226: Create an admin task when votes are 2 weeks old
## Description
We want a way to notify admins when helper votes are 2 weeks old, so that they can step in and get it over the line.

This should be done via a call to GitHub API to create an issue. The github org name, repo name, and the auth token should be constants to allow the to be different in dev & prod.

## Suggested implementation

Have a discord.py task that runs daily to check if there are any votes that meet the stale threshold, if so create the issue and add an emoji to that vote.

The emoji will allow us to make sure we don't create duplicate issues for the same vote.

shtlrs:
closes #2226

The idea is to fetch active nominations that are more than 2 weeks old, and create admin issues for them in Github so that staff will do the follow up.

As soon as a vote is tracked, we add a "🎫" reaction (which can change) to it to make sure we don't create duplicate issues for it, instead of keeping track in another way


shtlrs:
@ChrisLovering Here's a draft PR for #2226.

The only missing part is the Github issue part since i'm not quite aware of the repo/project structure for admins.
I suppose it needs to be configured from out `config.yml` file ?

HassanAbouelela:
Fwiw, you can target one PR to another branch (i.e it would look something like `main <- nomination history PR branch <- this PR branch`).

What that would do is basically try applying the changes from the final branch (3) in this chain to the middle branch (2), so when you check out the 3rd branch, it would have all the changes from branch 2 as well, but the PR itself will only have the changes of branch 3.

For the admin repo, there's an issue template which Chris can probably fetch for ya. The template body is completely unmodified at issue creation, we just set the nominated user as the issue title. You can use [this API](https://docs.github.com/en/rest/issues/issues#create-an-issue) to create the issue (as far as I can tell the API doesn't expose a way to create an issue from a template), which you'll need a token with access for. Simply add that as an environment variable in the default-config file ([example](https://github.com/python-discord/bot/blob/main/config-default.yml#L8)), and the devops team will worry about getting it set, just ping us to do so when the PR is ready.

Also please have a sensible default behavior for development environments where that token is unlikely to be set/exist.

shtlrs:
@HassanAbouelela Thanks for the tip, I didn't know that was feasible.

I see that we already have a keys section in our config, and we have a GitHub key as well but I think we'll need a new one because of permission levels, so I'll add a new key for it.

As for the default behavior, I have just thought of a **warning** log and an early return when that value of that key is null.

shtlrs:
A couple of points to discuss when I publish this PR:
1. Upon creating an issue and when a member is not found, should we delete the nomination ? 
2. Should we update our contribution guide to detail the new config that needs to be added in case someone wants to use/test the feature?

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -111,6 +118,81 @@ async def autoreview_status(self, ctx: Context) -> None:
...
+            for nomination in await self.api.get_nominations(active=True)
+            if (now - nomination.inserted_at).days >= OLD_NOMINATIONS_THRESHOLD_IN_DAYS
+        ]
+        return nominations
+
+    async def _filter_out_tracked_nominations(
+            self,
+            nominations: List[Nomination]
+    ) -> List[Tuple[Nomination, discord.Message]]:
+        """Filter the forgotten nominations that are still untracked in GitHub."""
```
ChrisLovering, line 145: This isn't just a filter, since it returns more data that it was given. This should be described in the docstring, and maybe change the function name to make it clear it isn't just a filter

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -111,6 +118,81 @@ async def autoreview_status(self, ctx: Context) -> None:
...
+        return untracked_nominations
+
+    async def _track_vote_in_github(self, nomination: Nomination) -> bool:
+        """
+        Adds an issue in GitHub to track dormant vote.
+
+        Returns True when the issue has been created, False otherwise.
+        """
+        if not GithubAdminRepo.token:
+            log.warn(f"No token for the {GithubAdminRepo.name} repository was provided, skipping issue creation.")
```
ChrisLovering, line 181:
```suggestion
            log.warning(f"No token for the {GithubAdminRepo.name} repository was provided, skipping issue creation.")
```
warn is a deprecated alias to warning

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -111,6 +118,81 @@ async def autoreview_status(self, ctx: Context) -> None:
...
+            return False
+
+        url = f"https://api.github.com/repos/{GithubAdminRepo.owner}/{GithubAdminRepo.name}/issues"
+        headers = {
+            "Accept": "application/vnd.github.v3+json",
+            "Authorization": f"Bearer {GithubAdminRepo.token}"
+        }
+        member = await get_or_fetch_member(self.bot.get_guild(Guild.id), nomination.user_id)
+        data = {"title": f"Nomination review needed. Id: {nomination.id}. User: {member.name}"}
```
ChrisLovering, line 190: A link to the nomination message in discord could be useful for the body of this issue.

ChrisLovering (changes requested):

ChrisLovering:
> A couple of points to discuss when I publish this PR:
> 
>     1. Upon creating an issue and when a member is not found, should we delete the nomination ?
> 
>     2. We need to update our contribution guide to detail the new config that needs to be added in case someone wants to use/test the feature, right ?
> 
> 
> The issue template/content/title is still to be discussed

1. Not in this PR, that can be discussed in a separate issue.
2. Yea, that can be another PR to site after this is merged

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -111,6 +118,81 @@ async def autoreview_status(self, ctx: Context) -> None:
...
+    async def track_forgotten_nominations(self) -> None:
+        """Track active nominations who are more than 2 weeks old."""
+        old_nominations = await self._get_forgotten_nominations()
+        nomination_details = await self._filter_out_tracked_nominations(old_nominations)
+        for nomination, nomination_vote_message in nomination_details:
+            issue_created = await self._track_vote_in_github(nomination)
+            if issue_created:
+                await nomination_vote_message.add_reaction(FLAG_EMOJI)
+
+    async def _get_forgotten_nominations(self) -> List[Nomination]:
```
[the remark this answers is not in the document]
ChrisLovering: This isn't resolved.

bot/exts/recruitment/talentpool/_cog.py
ChrisLovering: This isn't resolved.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -111,6 +118,81 @@ async def autoreview_status(self, ctx: Context) -> None:
...
+
+            thread = await get_or_fetch_channel(nomination.thread_id)
+            if not thread:
+                # Thread was deleted
+                continue
+
+            starter_message = thread.starter_message
+            if not starter_message:
+                # Starter message will be null if it's not cached
+                starter_message = await self.bot.get_channel(Channels.nomination_voting).fetch_message(thread.id)
```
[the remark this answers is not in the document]
ChrisLovering: This isn't resolved.

bot/exts/recruitment/talentpool/_cog.py
ChrisLovering: This isn't resolved.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -111,6 +118,81 @@ async def autoreview_status(self, ctx: Context) -> None:
...
+        if not GithubAdminRepo.token:
+            log.warn(f"No token for the {GithubAdminRepo.name} repository was provided, skipping issue creation.")
+            return False
+
+        url = f"https://api.github.com/repos/{GithubAdminRepo.owner}/{GithubAdminRepo.name}/issues"
+        headers = {
+            "Accept": "application/vnd.github.v3+json",
+            "Authorization": f"Bearer {GithubAdminRepo.token}"
+        }
+        member = await get_or_fetch_member(self.bot.get_guild(Guild.id), nomination.user_id)
```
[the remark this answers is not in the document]
ChrisLovering: This isn't resolved.

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -433,11 +434,20 @@ async def _previous_nominations_review(self, member: Member) -> Optional[str]:
...
         nomination_times = f"{num_entries} times" if num_entries > 1 else "once"
         rejection_times = f"{len(history)} times" if len(history) > 1 else "once"
+        threads = [await get_or_fetch_channel(nomination.thread_id) for nomination in history]
+
+        nomination_vote_threads = ", ".join(
+            [
+                f"[Thread-{thread_number}]({thread.jump_url})" if thread else ''
+                for thread_number, thread in enumerate(threads, start=1)
+            ]
+        )
```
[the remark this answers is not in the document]
shtlrs: And now that my previously mentioned PR has been merged, these changes no longer exist
Merge happened here 59319b4dd38d6fa49d3eb135604b2c1977c31b03

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -112,6 +118,93 @@ async def autoreview_status(self, ctx: Context) -> None:
...
+            if issue_created:
+                await thread.starter_message.add_reaction(FLAG_EMOJI)
+
+    async def _get_forgotten_nominations(self) -> list[Nomination]:
+        """Get active nominations that are more than 2 weeks old."""
+        now = arrow.utcnow()
+        nominations = [
+            nomination
+            for nomination in await self.api.get_nominations(active=True)
+            if (now - nomination.inserted_at).days >= OLD_NOMINATIONS_THRESHOLD_IN_DAYS
```
ChrisLovering, line 137: `nomination.inserted_at` is the datetime the nomination was created, rather than put up for review.

We want to use the date that the nomination message was posted up for review for this calculation.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -112,6 +118,93 @@ async def autoreview_status(self, ctx: Context) -> None:
...
+
+    async def _track_vote_in_github(self, nomination: Nomination, vote_jump_url: str | None = None) -> bool:
+        """
+        Adds an issue in GitHub to track dormant vote.
+
+        Returns True when the issue has been created, False otherwise.
+        """
+        if not GithubAdminRepo.token:
+            log.warning(f"No token for the {GithubAdminRepo.name} repository was provided, skipping issue creation.")
+            return False
```
ChrisLovering, line 188: this check can be done in the init function so that the loop isn't even started if this secret set

ChrisLovering (changes requested):

bot/exts/recruitment/talentpool/_cog.py
shtlrs: Done in 33415daa

bot/exts/recruitment/talentpool/_cog.py
shtlrs: You're right.

This is fixed in babf90649df6acb5e8ebda28e15b4e4b1ee9e2d2 by relying on the thread_id to turn into into a valid datetime object & use it for comparison

shtlrs:
Closing in favor of #2553

PR #2324, opened 2022-11-13
Make snekbox 3.11 stable default, with 3.10 as an option

ChrisLovering:
The manifests in https://github.com/python-discord/kubernetes/pull/144 should be applied before this PR is merged.

3.11 is now the default version for snekbox, so we no longer need a special cased 3.11 service. This 3.10 service has been created so that we can retain the version swapping capiblities for the next time we want to do something similar.

### DIFF
--- a/tests/bot/exts/utils/test_snekbox.py
+++ b/tests/bot/exts/utils/test_snekbox.py
@@ -29,7 +29,7 @@ class SnekboxTests(unittest.IsolatedAsyncioTestCase):
 
         self.assertEqual(await self.cog.post_job("import random", "3.10"), "return")
         self.bot.http_session.post.assert_called_with(
-            constants.URLs.snekbox_eval_api,
+            constants.URLs.snekbox_310_eval_api,
             json={"input": "import random"},
             raise_for_status=True
         )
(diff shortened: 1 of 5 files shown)


HassanAbouelela (approved):
I had just updated my config chris!

tests/bot/exts/utils/test_snekbox.py
```diff
@@ -29,7 +29,7 @@ async def test_post_job(self):
 
         self.assertEqual(await self.cog.post_job("import random", "3.10"), "return")
         self.bot.http_session.post.assert_called_with(
-            constants.URLs.snekbox_eval_api,
+            constants.URLs.snekbox_310_eval_api,
```
mbaruh, line 32: I'm not sure this is the intended behavior of the test. If the default version is now 3.11 then it makes more sense to change the tests to use 3.11.

mbaruh (changes requested):
I'm not convinced we should keep 3.10 as an option, if 3.11 is stable and has wheels for all of the libraries we use. I'm ok with removing the button without removing the code for switching between versions.

mbaruh:
> 3.11 was already deployed (and we don’t need wheels since we have gcc in the snekbox image)

yeah the bottom line is that if we can have all of the usual libraries available then I don't think we need to have 3.10 as an option. We had two versions only because we wanted to promote 3.11 before its release.

ChrisLovering:
With https://github.com/python-discord/snekbox/pull/175 around the corner, and the recent config changes, I don't think updating this PR is worth the effort.

PR #2325, opened 2022-11-15
Update return.md

swfarnsworth:
Made the tag more brief without any substantial changes to its overall approach.


bot/resources/tags/return.md
```diff
@@ -1,35 +1,32 @@
 **Return Statement**
 
-When calling a function, you'll often want it to give you a value back. In order to do that, you must `return` it. The reason for this is because functions have their own scope. Any values defined within the function body are inaccessible outside of that function.
-
-*For more information about scope, see `!tags scope`*
+A value created inside a function can't be used outside of it unless you `return` it.
```
CaedenPH, line 3: Personally, I think this change makes it harder to understand what exactly a return statement is. My opinion is to revert this change

bot/resources/tags/return.md
```diff
@@ -1,35 +1,32 @@
...
 None
 ```
 **Things to note**  
-• `print()` and `return` do **not** accomplish the same thing. `print()` will only print the value, it will not be accessible outside of the function afterwards.  
-• A function will return `None` if it ends without reaching an explicit `return` statement.  
-• When you want to print a value calculated in a function, instead of printing inside the function, it is often better to return the value and print the *function call* instead.  
-• [Official documentation for `return`](https://docs.python.org/3/reference/simple_stmts.html#the-return-statement)  
+• `print()` and `return` do **not** accomplish the same thing. `print()` will show the value, and then it will be gone.  
+• A function will return `None` if it ends without a `return` statement.  
+• When you want to print a value from a function, it's best to return the value and print the *function call* instead, like `print(square(5))`.
```
CaedenPH, line 32: I think the official documentation link should be kept

CaedenPH, line 32:
```suggestion
• When you want to print a value from a function, it's best to return the value and print the *function call* instead, for example, `print(square(5))`.  
```

bot/resources/tags/return.md
swfarnsworth: The target audience for this tag are those who are so unfamiliar with programming that they do not know what a return statement is or does, and who may not be proficient in English. The expected use case is to resolve confusion when beginners print the result of the function's computation instead of returning it.

While there might be beginners who expect variables defined in the function to be written to the global scope (especially if there is already a variable with the same name), I think that needs to be a separate discussion, and there is indeed a separate scope tag. But I do not think dedicating an extra line to "Run this other command to get even more information" will inspire anyone to do that, and would instead cost vertical space towards eyes glazing over.

bot/resources/tags/return.md
swfarnsworth: I do not think anyone will actually click that link, and I do not think the target audience would find it helpful if they did. The official docs contains formal statements about the semantics of `return` wrt exception handling and generators, which I think would confuse the target audience, if anything.

swfarnsworth: I am indifferent to this change, but I would prefer "instead; for example," instead of "instead, for example".

camcaswell: I think "like" is fine, and prefer anything that is shorter

camcaswell: I agree wtih Steele; the more basic a topic, the less useful a technical explanation will be to a beginner

bot/resources/tags/return.md
camcaswell: This sentence alone might not explain how `return` works, but it does point a beginner in the right direction. If they want something from inside a function, this motivates them to learn how `return` works.

camcaswell (approved):
lgtm

bot/resources/tags/return.md
shtlrs: I agree with @camcaswell, the point is to simplify usage for early beginners as much as possible.
And beginners usually have a sense of "hesitation" and/or "fear" when reading what something does, and we want to help build confidence & understanding in the simplest way.
This should help them get started just fine.

bot/resources/tags/return.md
```diff
@@ -1,35 +1,32 @@
...
 def square(n):
-    return n*n
+    return n * n
 ```
-If we wanted to store 5 squared in a variable called `x`, we could do that like so:
+If we wanted to store 5 squared in a variable called `x`, we would do:
 `x = square(5)`. `x` would now equal `25`.
 
 **Common Mistakes**
 ```py
```
shtlrs, line 14: Might help to enumerate the common mistakes here

bot/resources/tags/return.md
```diff
@@ -1,35 +1,32 @@
...
 >>> print(x)
 None
 ```
 **Things to note**  
-• `print()` and `return` do **not** accomplish the same thing. `print()` will only print the value, it will not be accessible outside of the function afterwards.  
-• A function will return `None` if it ends without reaching an explicit `return` statement.  
-• When you want to print a value calculated in a function, instead of printing inside the function, it is often better to return the value and print the *function call* instead.  
-• [Official documentation for `return`](https://docs.python.org/3/reference/simple_stmts.html#the-return-statement)  
+• `print()` and `return` do **not** accomplish the same thing. `print()` will show the value, and then it will be gone.  
+• A function will return `None` if it ends without a `return` statement.
```
shtlrs, line 31: I suggest to turn this into: `A function will ... if it doesn't have a `return` statement in it`
It doesn't necessarily need to end with it (even though it's not a good practice)

bot/resources/tags/return.md
```diff
@@ -1,35 +1,32 @@
 **Return Statement**
 
-When calling a function, you'll often want it to give you a value back. In order to do that, you must `return` it. The reason for this is because functions have their own scope. Any values defined within the function body are inaccessible outside of that function.
-
-*For more information about scope, see `!tags scope`*
+A value created inside a function can't be used outside of it unless you `return` it.
 
 Consider the following function:
```
shtlrs, line 5: How about we add a "scenario" here, which would be something like:
"Let's suppose we want to calculate the square of a given number and we want to use it later on. It would require having a function that calculates it then `return`s it to us for later usage"

bot/resources/tags/return.md
shtlrs, line 32:
```suggestion
• When you want to print a value from a function, it's best to return the value and print the *result of the function call* instead, like `print(square(5))`.  
```

shtlrs (changes requested):
Thanks for simplifying this!
I felt a couple of comments, let me know what you think :D

bot/resources/tags/return.md
swfarnsworth: What change are you suggesting specifically?

bot/resources/tags/return.md
swfarnsworth: I picked this wording carefully, because a function can have a `return` statement and still have an implicit `return None` if the return statement is never executed (ex, if the only return statement is in an if block that isn't entered). But we can hand-wave that away if others find my proposed wording confusing.

bot/resources/tags/return.md
shtlrs: I'm suggesting to make the separation of the common mistakes clearer by saying sth like:
- Common mistake 1:
...
- Common mistake 2:
...

bot/resources/tags/return.md
swfarnsworth: It's important that the tag be as brief as possible. What value do you think this additional text contributes when the code examples are already there?

bot/resources/tags/return.md
shtlrs: I agree with that, I just am afraid people will think that the return statement can only be at the end

bot/resources/tags/return.md
shtlrs: For total beginners, code might not be an "obvious" thing. And I believe having context helps them better position themselves & understand it better

bot/resources/tags/return.md
swfarnsworth: I don't think these extra words have added value given that an example immediately follows.

shtlrs: Sure, no need to change it

swfarnsworth:
Could we potentially merge this today? I've seen this confusion come up twice since the new help system was rolled out less than 24 hours ago, and I think the new version of this message would be helpful.

bot/resources/tags/return.md
wookie184: I think the current way is good enough and more concise.

PR #2339, opened 2022-11-26
Help forum improvements

ChrisLovering:
This PR has a few improvements to the recently added help forum system.

Each commit is an atomic change. As such, reviewing by commit will make this PR easier to review.

For the auto archive logic, I decided to keep the archive logic simple, and just go for 30 minutes since last message
This is instead of the hybrid of 30 mins + 10 depending on who sent the last message that was used previously.
The reasoning for using the hybrid approach previously was due to us running out of channels frequently, since this is no longer a problem, I decided to keep the logic simple.

### DIFF
--- a/bot/exts/help_channels/_cog.py
+++ b/bot/exts/help_channels/_cog.py
@@ -28,7 +28,12 @@ class HelpForum(commands.Cog):
 
     def __init__(self, bot: Bot):
         self.bot = bot
-        self.help_forum_channel_id = constants.Channels.help_system_forum
+        self.help_forum_channel: discord.ForumChannel = None
+
+    async def cog_load(self) -> None:
+        """Schedule the auto-archive check for all open posts."""
+        log.trace("Initialising help forum cog.")
+        self.help_forum_channel = self.bot.get_channel(constants.Channels.help_system_forum)
 
     async def close_check(self, ctx: commands.Context) -> bool:
         """Return True if the channel is a help post, and the user is the claimant or has a whitelisted role."""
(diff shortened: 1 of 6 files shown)


mbaruh (approved):
Thanks, looks good as far as I can see

bot/exts/help_channels/_cog.py
```diff
@@ -113,33 +129,48 @@ async def rename_help_post(self, ctx: commands.Context, *, title: str) -> None:
...
 
         await self.post_with_disallowed_title_check(thread)
-        await _channel.help_thread_opened(thread)
+        await _channel.help_post_opened(thread)
+
+        delay = min(constants.HelpChannels.deleted_idle_minutes, constants.HelpChannels.idle_minutes) * 60
+        self.scheduler.schedule_later(
+            delay,
+            thread.id,
+            _channel.maybe_archive_idle_post(thread, self.scheduler)
```
sid-maddy, line 153: I think we should set `has_task=False` here, since there won't be an event in the `scheduler` for a new post and so `scheduler.cancel` will raise a `ValueError`

ChrisLovering: The code you just quoted is adding this coro as a task to the scheduler

mbaruh: The scheduler will have a task for the post - the one scheduled in these lines.

sid-maddy: Ohh, right. Gotcha.

sid-maddy (approved):

PR #2341, opened 2022-11-26
Allow members to self assign roles through a persistent view

### LINKED ISSUE #2332: Permanent roles view
Our current setup is that users must run `!subscribe` in bot-commands in order to bring up a view of all self-assignable roles.

We should create a permanent view in `#roles` with a single button which will essentially run the command (with an ephemeral message).

This is more in line with how it works in other communities, with a better UI than running a command, and arguably better discoverability.

The current `!subscribe` command should remain as-is for the time being.

shtlrs:
Closes #2332 

As stated by Zig, subscribing to our public roles is currently not super visible since it revolves around executing the `!subscribe` command that usually goes by unnoticed.

To make this easier, more visible & to align with other communities have, we decided to add a persistent view in the `#roles` channel that allows people to subscribing to their chosen roles by a click of a button.

### DIFF
--- a/bot/exts/info/subscribe.py
+++ b/bot/exts/info/subscribe.py
@@ -151,7 +151,7 @@ class SingleRoleButton(discord.ui.Button):
         self.style = self.REMOVE_STYLE if self.assigned else self.ADD_STYLE
         self.label = self.LABEL_FORMAT.format(action="Remove" if self.assigned else "Add", role_name=self.role.name)
         try:
-            await interaction.message.edit(view=self.view)
+            await self.view.anchor_message.edit(view=self.view)
         except discord.NotFound:
             log.debug("Subscribe message for %s removed before buttons could be updated", interaction.user)
             self.view.stop()
@@ -184,6 +184,7 @@ class ShowAllSelfAssignableRolesButton(discord.ui.Button):
         view = prepare_self_assignable_roles_view(interaction, self.assignable_roles)
         message = await interaction.followup.send(
             view=view,
+            ephemeral=True
         )
         # Keep reference of the message that contains the view to be deleted
         view.anchor_message = message
--- a/bot/exts/info/subscribe.py
+++ b/bot/exts/info/subscribe.py
@@ -183,7 +183,7 @@ class Subscribe(commands.Cog):
 
     SELF_ASSIGNABLE_ROLES_MESSAGE = (
         f"Hi there {GREETING_EMOJI},"
-        "\nWe have self-assignable roles for server updates an events!"
+        "\nWe have self-assignable roles for server updates and events!"
         "\nClick the button below to toggle them:"
     )
 
--- a/bot/exts/info/subscribe.py
+++ b/bot/exts/info/subscribe.py
@@ -214,7 +214,7 @@ class Subscribe(commands.Cog):
 
         placeholder_message_view_tuple = await self._fetch_or_create_self_assignable_roles_message()
         self_assignable_roles_message, self_assignable_roles_view = placeholder_message_view_tuple
-        self.__attach_persistent_roles_view(self_assignable_roles_message, self_assignable_roles_view)
+        self._attach_persistent_roles_view(self_assignable_roles_message, self_assignable_roles_view)
 
     @commands.cooldown(1, 10, commands.BucketType.member)
     @commands.command(name="subscribe", aliases=("unsubscribe",))
@@ -250,7 +250,7 @@ class Subscribe(commands.Cog):
         placeholder_message = await roles_channel.send(self.SELF_ASSIGNABLE_ROLES_MESSAGE, view=view)
         return placeholder_message, view
 
-    def __attach_persistent_roles_view(
+    def _attach_persistent_roles_view(
             self,
             placeholder_message: discord.Message,
             persistent_roles_view: discord.ui.View | None = None
--- a/bot/exts/info/subscribe.py
+++ b/bot/exts/info/subscribe.py
@@ -161,7 +161,7 @@ class ShowAllSelfAssignableRolesButton(discord.ui.Button):
     def __init__(self, assignable_roles: list[AssignableRole]):
         super().__init__(
             style=discord.ButtonStyle.success,
-            label="Show available roles",
+            label="Show all self assignable roles",
             custom_id=self.CUSTOM_ID,
             row=1
         )
@@ -181,7 +181,7 @@ class ShowAllSelfAssignableRolesButton(discord.ui.Button):
 class Subscribe(commands.Cog):
     """Cog to allow user to self-assign & remove the roles present in ASSIGNABLE_ROLES."""
 
-    SELF_ASSIGNABLE_ROLES_MESSAGE = "Click on this button to earn all the currently available server roles"
+    SELF_ASSIGNABLE_ROLES_MESSAGE = "Click on this button to show all self assignable roles"
 
     def __init__(self, bot: Bot):
         self.bot = bot
--- a/bot/exts/info/subscribe.py
+++ b/bot/exts/info/subscribe.py
@@ -74,11 +74,17 @@ class RoleButtonView(discord.ui.View):
 
     anchor_message: discord.Message
 
-    def __init__(self, member: discord.Member):
+    def __init__(self, member: discord.Member, assignable_roles: list[AssignableRole]):
         super().__init__(timeout=DELETE_MESSAGE_AFTER)
         # We can't obtain the reference to the message before the view is sent
         self.anchor_message = None
         self.interaction_owner = member
+
+        author_roles = [role.id for role in member.roles]
+
+        for index, role in enumerate(assignable_roles):
+            row = index // ITEMS_PER_ROW
+            self.add_item(SingleRoleButton(role, role.role_id in author_roles, row))
 
     async def interaction_check(self, interaction: Interaction) -> bool:
         """Ensure that the user clicking the button is the member who invoked the command."""
@@ -173,7 +179,7 @@ class AllSelfAssignableRolesView(discord.ui.View):
     async def show_all_self_assignable_roles(self, interaction: Interaction, button: discord.ui.Button) -> None:
         """Sends the original subscription view containing the available self assignable roles."""
         await interaction.response.defer()
-        view = prepare_self_assignable_roles_view(interaction, self.assignable_roles)
+        view = RoleButtonView(interaction.user, self.assignable_roles)
         message = await interaction.followup.send(
             view=view,
             ephemeral=True
@@ -229,8 +235,7 @@ class Subscribe(commands.Cog):
     )
     async def subscribe_command(self, ctx: commands.Context, *_) -> None:  # We don't actually care about the args
         """Display the member's current state for each role, and allow them to add/remove the roles."""
-        view = prepare_self_assignable_roles_view(ctx, self.assignable_roles)
-
+        view = RoleButtonView(ctx.author, self.assignable_roles)
         message = await ctx.send(
             "Click the buttons below to add or remove your roles!",
             view=view,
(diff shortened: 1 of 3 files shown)


bot/exts/info/subscribe.py
```diff
@@ -178,22 +233,73 @@ async def cog_load(self) -> None:
...
+) -> discord.ui.View:
+    """Prepares the view containing the self assignable roles before its sent."""
+    author = trigger_action.author if isinstance(trigger_action, commands.Context) else trigger_action.user
+    author_roles = [role.id for role in author.roles]
+    button_view = RoleButtonView(member=author)
+    button_view.anchor_message = trigger_action.message
+
+    for index, role in enumerate(assignable_roles):
+        row = index // ITEMS_PER_ROW
+        button_view.add_item(SingleRoleButton(role, role.role_id in author_roles, row))
```
mbaruh, line 295: So I know that this is copied from existing code, but I think that the button creation should be done inside `RoleButtonView`'s init. It already takes the author, and then it can make the buttons for itself. You then also don't need this function at all.

config-default.yml
```diff
@@ -235,6 +235,9 @@ guild:
         # Watch
         big_brother_logs:   &BB_LOGS        468507907357409333
 
+        # Roles
```
mbaruh, line 238: There are no other roles channels, so you can maybe call this section `# Information`

mbaruh (approved):
Looks pretty good! this is purely a code review, will test later.

mbaruh (changes requested):
misclick 😅

bot/exts/info/subscribe.py
```diff
@@ -178,22 +233,73 @@ async def cog_load(self) -> None:
...
+        message = await ctx.send(
             "Click the buttons below to add or remove your roles!",
-            view=button_view,
-            delete_after=DELETE_MESSAGE_AFTER,
+            view=view,
         )
+        # Keep reference of the message that contains the view to be deleted
+        view.anchor_message = message
+
+    async def __search_for_self_assignable_roles_message(self) -> discord.Message:
```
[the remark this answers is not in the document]
shtlrs: Understood, I just didn't feel like "exposing" it.
Fixed in 0a60b5d4200219233ad5164da3ae73c5f43955e7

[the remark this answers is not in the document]
shtlrs: Understood, I just didn't feel like "exposing" it.
Fixed in 0a60b5d4200219233ad5164da3ae73c5f43955e7

bot/exts/info/subscribe.py
```diff
@@ -133,15 +151,50 @@ async def update_view(self, interaction: Interaction) -> None:
...
         except discord.NotFound:
             log.debug("Subscribe message for %s removed before buttons could be updated", interaction.user)
             self.view.stop()
 
 
+class AllSelfAssignableRolesView(discord.ui.View):
+    """A persistent view that'll hold one button allowing interactors to toggle all available self-assignable roles."""
+
+    def __init__(self):
+        super(AllSelfAssignableRolesView, self).__init__(timeout=None)
```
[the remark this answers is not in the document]
shtlrs: Fixed in 28dac62fb325e0f09404945645589a89af19bfdd

config-default.yml
shtlrs: Fair enough, renamed here 72b2f803a6bb915c4bb07d3a0cd3f4d141d46431

bot/exts/info/subscribe.py
```diff
@@ -133,15 +151,50 @@ async def update_view(self, interaction: Interaction) -> None:
...
+
+
+class ShowAllSelfAssignableRolesButton(discord.ui.Button):
```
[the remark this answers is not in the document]
shtlrs: That makes more sense indeed, done here 24c36e5bcc8f87c045242c88b6b5baa8d12110c0

bot/exts/info/subscribe.py
```diff
@@ -133,15 +151,50 @@ async def update_view(self, interaction: Interaction) -> None:
...
+    """A button that, when clicked, sends a view a user can interact with to assign one or all the available roles."""
+
+    CUSTOM_ID = "gotta-claim-them-all"
```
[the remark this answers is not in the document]
shtlrs: Well, I tried ahahaha
Fixed here a6fa4b1b7ce4dcff8c0ecfc4459df471ec97cde9

bot/exts/info/subscribe.py
```diff
@@ -133,15 +151,50 @@ async def update_view(self, interaction: Interaction) -> None:
...
+            ephemeral=True
+        )
+        # Keep reference of the message that contains the view to be deleted
+        view.anchor_message = message
+
+
 class Subscribe(commands.Cog):
     """Cog to allow user to self-assign & remove the roles present in ASSIGNABLE_ROLES."""
 
+    SELF_ASSIGNABLE_ROLES_MESSAGE = "Click on this button to show all self assignable roles"
```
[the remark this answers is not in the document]
shtlrs: I did that here c630e1b9c2ca624dd272d634cd24704c11bea953, let me know what you think

bot/exts/info/subscribe.py
shtlrs: Good eye !  
This has been done here 37b43a6c04af0498d96bd28ec544973d2e653941

bot/exts/info/subscribe.py
```diff
@@ -178,22 +233,73 @@ async def cog_load(self) -> None:
...
+
+        The message is searched for/created upon loading the Cog.
+
+        Parameters
+        __________
+            :param placeholder_message: The message that will hold the persistent view allowing
+            users to toggle the RoleButtonView
+        """
+        view = AllSelfAssignableRolesView()
+        view.add_item(ShowAllSelfAssignableRolesButton(self.assignable_roles))
```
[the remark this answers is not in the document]
shtlrs: Indeed, done here 6c6b2bc358228404978c855d18c788e85fb4433a
If you think the naming I used needs to be improved, please feel free to suggest ideas

bot/exts/info/subscribe.py
```diff
@@ -133,15 +151,50 @@ async def update_view(self, interaction: Interaction) -> None:
...
+            custom_id=self.CUSTOM_ID,
+            row=1
+        )
+        self.assignable_roles = assignable_roles
+
+    async def callback(self, interaction: Interaction) -> t.Any:
+        """Sends the original subscription view containing the available self assignable roles."""
+        await interaction.response.defer()
+        view = prepare_self_assignable_roles_view(interaction, self.assignable_roles)
+        message = await interaction.followup.send(
```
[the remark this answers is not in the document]
shtlrs: Main reason was my bad understanding of the interaction API, editing that view was failing since it was referencing the wrong message. And I thought I needed to workaround this by keeping a ref to the original message.

Seems like using `interaction.response.edit_message` works well
Fixed in both 7118c51b9b62b2a5ce0e745e4832b17f52901235 and 7976bc2af0fe49ddaebc63242b581b9eecb66132

bot/exts/info/subscribe.py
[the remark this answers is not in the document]
shtlrs: And in 91a869e4a23fdee8c6407678ca178e324d6184f3 as well

onerandomusername:
I know I'm late but what about a button that shows an ephemeral view which takes into consideration what roles the user already has, so they can select all of the roles they want in one single go?

shtlrs:
But that's similar to what we already have, isn't it?  or did I understand that wrong?

The ones they have are highlighted in red, and the message states that just fine, e.g. Whether the role will be added or removed.

onerandomusername:
> But that's similar to what we already have, isn't it? or did I understand that wrong?
> 
> The ones they have are highlighted in red, and the message states that just fine, e.g. Whether the role will be added or removed.
> 
> And how would they select them all in one go?

This is what I have for roles in my guilds.

image

This implementation uses a slash command to show the select menu, but it could be anything: like a button in the \#roles channel which shows this select. 

This is nicer from a UX perspective IMO because it shows which roles the user already has, and while its true the button view *does* communicate this fact, the button view is a bit confusing and does not allow for descriptions.

shtlrs:
Hummm, i guess this could be discussed in another issue.
If it is going to be implemented, I think it should be done in another separate PR.

The current goal is to keep the same behavior for the subscribe command and add more visibility by '' pinning '' it in the roles channel.

I like the idea, and if approved, i will be more than happy to do it.

bot/exts/info/subscribe.py
```diff
@@ -133,15 +148,43 @@ async def update_view(self, interaction: Interaction) -> None:
...
+
+
 class Subscribe(commands.Cog):
     """Cog to allow user to self-assign & remove the roles present in ASSIGNABLE_ROLES."""
 
+    GREETING_EMOJI = ":wave:"
+
+    SELF_ASSIGNABLE_ROLES_MESSAGE = f"Hi there {GREETING_EMOJI}," \
+                                    "\nDid you know we had various self-assignable roles for server updates an events?"\
+                                    "\nClick the button below to toggle them."
```
mbaruh, line 186:
```suggestion
    SELF_ASSIGNABLE_ROLES_MESSAGE = (
        f"Hi there {GREETING_EMOJI},"
        "\nWe have self-assignable roles for server updates an events!"
        "\nClick the button below to toggle them:"
    )
```

mbaruh (approved):
Looks great, just a few nitpicks

https://github.com/python-discord/bot/blob/91a869e4a23fdee8c6407678ca178e324d6184f3/bot/exts/info/subscribe.py#L101
Let's get rid of that dot at the end while we're at it.

bot/exts/info/subscribe.py
shtlrs: Oh, that's a good one.
Done in c55b9ffb

bot/exts/info/subscribe.py
```diff
@@ -133,15 +148,45 @@ async def update_view(self, interaction: Interaction) -> None:
...
+
+
 class Subscribe(commands.Cog):
     """Cog to allow user to self-assign & remove the roles present in ASSIGNABLE_ROLES."""
 
+    GREETING_EMOJI = ":wave:"
+
+    SELF_ASSIGNABLE_ROLES_MESSAGE = (
+        f"Hi there {GREETING_EMOJI},"
+        "\nWe have self-assignable roles for server updates an events!"
```
Robin5605, line 186: I'm not sure what this sentence is supposed to mean entirely. Perhaps it could be reworded to be concise?

bot/exts/info/subscribe.py
```diff
@@ -178,22 +226,58 @@ async def cog_load(self) -> None:
...
+
+        If the initial message isn't found, a new one will be created.
+        This message will always be needed to attach the persistent view to it
+        """
+        roles_channel: discord.TextChannel = await get_or_fetch_channel(constants.Channels.roles)
+
+        async for message in roles_channel.history(limit=30):
+            if message.content == self.SELF_ASSIGNABLE_ROLES_MESSAGE:
+                log.debug(f"Found self assignable roles view message: {message.id}")
+                return message, None
```
Robin5605, line 248: Is there a reason we are returning `None` for the `discord.ui.View`? Couldn't we just use [`discord.ui.View.from_message`](https://discordpy.readthedocs.io/en/stable/interactions/api.html?highlight=ui%20button#discord.ui.View.from_message)?

Robin5605 (changes requested):
Nothing major. Thanks for the PR! I haven't tested it either, so I'm hoping we didn't letting anything slip by haha 😄

bot/exts/info/subscribe.py
shtlrs: That's supposed to be an "and", it's a typo.
Does it still appear unclear to you ?

bot/exts/info/subscribe.py
shtlrs: We can't really recreate the view using `from_message`, as this persistent view needs to be instantiated with the self assignable roles, and `from_message` cannot pick that up.

Robin5605 (approved):
Great job, thanks! I have no more suggestions 😃

PR #2347, opened 2022-11-27
include Thread & PrivateChannel in get_or_fetch_channel return types

### LINKED ISSUE #2346: Improve return types of bot.utils.channel.get_or_fetch_channel
Currently, `get_or_fetch_channel` is only typed to return an instance of `GuildChannel`, which isn't always true since it can also return a  `Thread` or a `PrivateChannel`

The purpose of this issue is to improve the return type hints.

shtlrs:
Closes #2346

All in issue


wookie184 (approved):
Seems the same as what discord.py types `get_channel` and `fetch_channel` as so should be good, thanks.

ChrisLovering (approved):

PR #2350, opened 2022-11-29
Rename print-return to return-gif

janine9vn:
This renames `!print-return` to `!return-gif` based on some feedback from Helpers. 

It also adds `!print-return` and `!return-jif` as aliases.


shenanigansd (approved):
"jif" makes me sad

swfarnsworth (approved):
I'm fine with `return-jif` for the meme, even though /gɪf/ is the only pronunciation I will accept.

PR #2351, opened 2022-11-30
Don't count bot messages as an answer in a thread

wookie184:
Not tested, as I don't have forum channels on my test server, but I think this is correct (`on_message` does include bot messages, right?).


wookie184:
I looked into this as I noticed that the unanswer rate was 0 on grafana in the past 12 hours, but there do seem to be some unanswered questions.

ChrisLovering (approved):

mbaruh (approved):

Issue #2331, opened 2022-11-24
Idea: New command


Kingrobotic1000:
I thing this bot should have a new command called `!under`.
`!under` should show the rules for how underscores work in python. it would be very beneficial for new programmers to learn how to use underscores when naming vars/functions/classes.

wookie184:
This sounds like a good idea for a tag. I don't think we have a similar one already.

It should list the following styles:
- `__name__`
- `_name`
- `__name`
- `name_`
- `_`

Giving a very short explanation of each (and potentially linking to a resource with more information if relevant).

DivyanshuBist:
Hi, I would Like to resolve this issue

wookie184:
Closed by https://github.com/python-discord/bot/pull/2354

PR #2354, opened 2022-12-04
added under command Issue #2331

DivyanshuBist:
Added Resources and References for `!under` command as suggested in Issue #2331

### DIFF
--- a/bot/resources/tags/under.md
+++ b/bot/resources/tags/under.md
@@ -1,0 +1,36 @@
+## UseCases of Underscore in Python
+- **Dunders** Double Underscore(Dunder) methods are special methods defined in a class that are invoked implicitly. example- `__name__`,`__init__`,`__repr__`,`__str__`                       
+Use `!dunder-methods` command to know more about these methods
+- **Single leading Underscore**  in front of a variable, a function, or a method name means that these objects are used internally.Remember that these objects can be accessed outside or other script. example-`_name`
+- **Single Trailing Underscores** are used to avoid naming conflict while using a variable which is a reserved keyword.
+example- using a variable name as `class` will produce an error,to avoid this conflict, you can add a trailing underscore as a naming convention i.e. `class_`.
+-  **Double leading underscores** Double leading underscores are typically used for name mangling.
+Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses.
+- **Store Expression Value** Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable.example-
+```python
+>>>1+1
+    2
+>>> _
+    2
+>>> _ + 2
+    4
+```
+-  **Ignore Values** Underscore can be used to ignore values and as a variable in looping.example- 
+ ```python
+>>> a, *_, b = (3,5,2,6,2)
+>>> print(a,b,_)
+     3 2 [5, 2, 6]
+>>> for _ in range(2):  #variable in looping
+          print(_)
+       0
+       1
+```
+-  **Improve Readability** - can be used with numeric literals to improve readability of long numbers.example-
+ ```python
+>>>1_00_000
+    100000
+```
+References-
+- [Towards Data Science](https://towardsdatascience.com/whats-the-meaning-of-single-and-double-underscores-in-python-3d27d57d6bd1)
+- [GeeksforGeeks](https://www.geeksforgeeks.org/underscore-_-python/)
+- [Data Camp](https://www.datacamp.com/tutorial/role-underscore-python)
--- a/bot/resources/tags/under.md
+++ b/bot/resources/tags/under.md
@@ -1,41 +1,24 @@
-**Use cases of Underscore in Python**
+**Meanings of Underscores in Identifier Names**
 
-- **Dunders**: Double Underscore (Dunder) methods are special methods defined in a class that are invoked implicitly. For example, `__name__`,`__init__`,`__repr__`, and `__str__`.
-Use the `!dunder-methods` tag to know more about them.
-- **Single Leading Underscores**:  They're used in front a variable, a function or a method name, which indicates that these objects are meant to be used internally, such as `_name`. However, these objects **can** still be accessed from outside.
-- **Single Trailing Underscores**: They are used to avoid naming conflict while making a new variable whose name might conflict with a reserved keyword or package name, etc. For example, using `class` as a variable name will produce an error; and to avoid this conflict, you can add a trailing underscore to it, i.e. `class_`.
--  **Double Leading Underscores**: Double leading underscores are typically used for "Name Mangling". Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses. More details regarding this topic can be found [here](https://www.geeksforgeeks.org/name-mangling-in-python/).
-- **Store Expression Value**: Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable. Example:
+• `__name__`: Used to implement special behaviour, such as the `+` operator for classes with the `__add__` method. [More info](https://dbader.org/blog/python-dunder-methods)
+• `_name`: Indicates that a variable is "private" and should only be used by the class or module that defines it.
+• `name_`: Used to avoid naming conflicts. For example, as `class` is a keyword, you could call a variable `class_` instead.
+• `__name`: Causes the name to be "mangled" if defined inside a class. [More info](https://docs.python.org/3/tutorial/classes.html#private-variables)
+
+A single underscore, **`_`**, has multiple uses:
+• To indicate an unused variable, e.g. in a for loop if you don't care which iteration you are on
 ```python
->>> _ # Using underscore without any previously executed expression or value assigned to underscore.
-    NameError: name '_' is not defined
->>> 1 + 1  # evaluating expression.
+for _ in range(10):
+    print("Hello World")
+```
+•  In the REPL, where the previous result will assigned to the variable `_`
+```python
+>>> 1 + 1  # Evaluated and stored in `_`
     2
->>> _        # Prints the value of the last executed expression.
-    2
->>> _ + 2
-    4
->>> _ = 3  # Once a value has explicitly been assigned to it, underscore won't store the value of last executed expression anymore.
->>> 5 + 9
-    14
->>> _
-    3
+>>> _ + 3  # Take the previous result and add 3
+    5
 ```
--  **Ignore Values**: We previously stated that underscores can also be used for variable assignments, but they are most commonly used to store values that will not be needed. Therefore,they are also known as `throwaway variable`.
- ```python
->>> a, *_, b = (3, 5, 2, 6, 2) # Ignore values when unpacking.
->>> print(a, b, _)
-    3 2 [5, 2, 6]
->>> for _ in range(10):  # Here the index is ignored since we only want to execute the same function multiple times.
-        do_something()
-```
--  **Improve Readability**: They can also be used with numeric literals to improve readability of long numbers.
- ```python
->>> 1_00_000
-    100000
-```
+• In integer literals, e.g. `x = 1_500_000` can be written instead of `x = 1500000` to improve readability
+• As a [wildcard in match/case statements](https://peps.python.org/pep-0636/#adding-a-wildcard)
 
-**References**:
-- [Towards Data Science](https://towardsdatascience.com/whats-the-meaning-of-single-and-double-underscores-in-python-3d27d57d6bd1)
-- [GeeksforGeeks](https://www.geeksforgeeks.org/underscore-_-python/)
-- [Data Camp](https://www.datacamp.com/tutorial/role-underscore-python)
+See also ["Reserved classes of identifiers"](https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers) in the Python docs, and [this more detailed guide](https://dbader.org/blog/meaning-of-underscores-in-python).
--- a/bot/resources/tags/under.md
+++ b/bot/resources/tags/under.md
@@ -1,5 +1,5 @@
 ## UseCases of Underscore in Python
-- **Dunders** Double Underscore(Dunder) methods are special methods defined in a class that are invoked implicitly. example- `__name__`,`__init__`,`__repr__`,`__str__`  
+- **Dunders** Double Underscore(Dunder) methods are special methods defined in a class that are invoked implicitly. example- `__name__`,`__init__`,`__repr__`,`__str__`
 Use `!dunder-methods` command to know more about these methods
 - **Single leading Underscore**  in front of a variable, a function, or a method name means that these objects are used internally.Remember that these objects can be accessed outside or other script. example-`_name`
 - **Single Trailing Underscores** are used to avoid naming conflict while using a variable which is a reserved keyword.
--- a/bot/resources/tags/under.md
+++ b/bot/resources/tags/under.md
@@ -1,4 +1,5 @@
-## UseCases of Underscore in Python
+**UseCases of Underscore in Python**
+
 - **Dunders** Double Underscore(Dunder) methods are special methods defined in a class that are invoked implicitly. example- `__name__`,`__init__`,`__repr__`,`__str__`
 Use `!dunder-methods` command to know more about these methods
 - **Single leading Underscore**  in front of a variable, a function, or a method name means that these objects are used internally.Remember that these objects can be accessed outside or other script. example-`_name`
--- a/bot/resources/tags/under.md
+++ b/bot/resources/tags/under.md
@@ -2,7 +2,7 @@
 
 - **Dunders**: Double Underscore (Dunder) methods are special methods defined in a class that are invoked implicitly. For example, `__name__`,`__init__`,`__repr__`, and `__str__`.
 Use `!dunder-methods` command to know more about these methods
-- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. Remember that these objects can be accessed outside or other script. 
+- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. Remember that these objects can be accessed outside or other script.
 - **Single Trailing Underscores**: Avoid naming conflict while using a variable which is a reserved keyword. Using a variable name as `class` will produce an error. To avoid this conflict, you can add a trailing underscore as a naming convention i.e. `class_`.
 -  **Double leading underscores**: Double leading underscores are typically used for name mangling.
 Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses.
--- a/bot/resources/tags/under.md
+++ b/bot/resources/tags/under.md
@@ -1,12 +1,12 @@
 **Use cases of Underscore in Python**
 
 - **Dunders**: Double Underscore (Dunder) methods are special methods defined in a class that are invoked implicitly. For example, `__name__`,`__init__`,`__repr__`, and `__str__`.
-Use `!dunder-methods` command to know more about these methods
+Use the `!dunder-methods` tag to know more about these methods
 - **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. Remember that these objects can be accessed outside or other script.
-- **Single Trailing Underscores**: Avoid naming conflict while using a variable which is a reserved keyword. Using a variable name as `class` will produce an error. To avoid this conflict, you can add a trailing underscore as a naming convention i.e. `class_`.
+- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. However, remember that these objects **can** be accessed outside.
 -  **Double leading underscores**: Double leading underscores are typically used for name mangling.
-Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses.
-- **Store Expression Value** Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable.example-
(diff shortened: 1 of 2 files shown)


bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
+## UseCases of Underscore in Python
```
Xithrius, line 1: This doesn't work in Discord embeds. It'll show up as the following:
image

```suggestion
**UseCases of Underscore in Python**

```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+- **Dunders** Double Underscore(Dunder) methods are special methods defined in a class that are invoked implicitly. example- `__name__`,`__init__`,`__repr__`,`__str__`
```
Xithrius, line 2:
```suggestion
- **Dunders**: Double Underscore (Dunder) methods are special methods defined in a class that are invoked implicitly. For example, `__name__`,`__init__`,`__repr__`, and `__str__`.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+Use `!dunder-methods` command to know more about these methods
+- **Single leading Underscore**  in front of a variable, a function, or a method name means that these objects are used internally.Remember that these objects can be accessed outside or other script. example-`_name`
```
Xithrius, line 4:
```suggestion
- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. Remember that these objects can be accessed outside or other script. 
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+- **Single Trailing Underscores** are used to avoid naming conflict while using a variable which is a reserved keyword.
+example- using a variable name as `class` will produce an error,to avoid this conflict, you can add a trailing underscore as a naming convention i.e. `class_`.
```
Xithrius, line 6: I think this should be on a single line.

```suggestion
- **Single Trailing Underscores**: Avoid naming conflict while using a variable which is a reserved keyword. Using a variable name as `class` will produce an error. To avoid this conflict, you can add a trailing underscore as a naming convention i.e. `class_`.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+-  **Double leading underscores** Double leading underscores are typically used for name mangling.
+Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses.
+- **Store Expression Value** Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable.example-
+```python
+>>>1+1
```
Xithrius, line 11: Keep the spacing consistent.

```suggestion
>>> 1 + 1
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+    2
+>>> _ + 2
+    4
+```
+-  **Ignore Values** Underscore can be used to ignore values and as a variable in looping.example-
+ ```python
+>>> a, *_, b = (3,5,2,6,2)
+>>> print(a,b,_)
+     3 2 [5, 2, 6]
+>>> for _ in range(2):  #variable in looping
```
Xithrius, line 23: There should be a space after the `#`.

```suggestion
>>> for _ in range(2):  # variable in looping
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+- **Store Expression Value** Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable.example-
+```python
+>>>1+1
+    2
+>>> _
+    2
+>>> _ + 2
+    4
+```
+-  **Ignore Values** Underscore can be used to ignore values and as a variable in looping.example-
```
Xithrius, line 18:
```suggestion
-  **Ignore Values**: An underscore can be used to ignore values and as a variable in looping.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+ ```python
+>>> a, *_, b = (3,5,2,6,2)
+>>> print(a,b,_)
```
Xithrius, line 21: Let's keep spaces between commas consistent.

```suggestion
>>> a, *_, b = (3, 5, 2, 6, 2)
>>> print(a, b, _)
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+     3 2 [5, 2, 6]
+>>> for _ in range(2):  #variable in looping
+          print(_)
+       0
+       1
+```
+-  **Improve Readability** - can be used with numeric literals to improve readability of long numbers.example-
```
Xithrius, line 28:
```suggestion
-  **Improve Readability**: they can also be used with numeric literals to improve readability of long numbers.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+ ```python
+>>>1_00_000
```
Xithrius, line 30:
```suggestion
>>> 1_00_000
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+    100000
+```
+References-
```
Xithrius, line 33:
```suggestion

**References**:
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
+## UseCases of Underscore in Python
+- **Dunders** Double Underscore(Dunder) methods are special methods defined in a class that are invoked implicitly. example- `__name__`,`__init__`,`__repr__`,`__str__`
+Use `!dunder-methods` command to know more about these methods
+- **Single leading Underscore**  in front of a variable, a function, or a method name means that these objects are used internally.Remember that these objects can be accessed outside or other script. example-`_name`
+- **Single Trailing Underscores** are used to avoid naming conflict while using a variable which is a reserved keyword.
+example- using a variable name as `class` will produce an error,to avoid this conflict, you can add a trailing underscore as a naming convention i.e. `class_`.
+-  **Double leading underscores** Double leading underscores are typically used for name mangling.
```
Xithrius, line 7:
```suggestion
-  **Double leading underscores**: Double leading underscores are typically used for name mangling.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,36 @@
...
+Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses.
+- **Store Expression Value** Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable.example-
```
Xithrius, line 9: I don't think it's necessary to say that we're using an underscore, since that is what this tag is about.

```suggestion
- **Store Expression Value**: Store the value of last expression in an interpreter and can be used as a variable.
```

Xithrius (changes requested):
Hello! Thank you for contributing to this tag. Below is a bunch of changes, but they're all very small, such as adding a space/newline or removing `example-` (I didn't think it's necessary).

Here is what the current tag looks like. I'll post another screenshot once changes are made.

image

bot/resources/tags/under.md
```diff
@@ -0,0 +1,37 @@
+**UseCases of Underscore in Python**
```
vivekashok1221, line 1:
```suggestion
**Use cases of Underscore in Python**
```

vivekashok1221 (changes requested):

bot/resources/tags/under.md
```diff
@@ -0,0 +1,37 @@
+**Use cases of Underscore in Python**
+
+- **Dunders**: Double Underscore (Dunder) methods are special methods defined in a class that are invoked implicitly. For example, `__name__`,`__init__`,`__repr__`, and `__str__`.
+Use `!dunder-methods` command to know more about these methods
+- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. Remember that these objects can be accessed outside or other script.
```
CaedenPH, line 5:
```suggestion
- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. However, remember that these objects **can** be accessed outside.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,37 @@
...
+- **Single Trailing Underscores**: Avoid naming conflict while using a variable which is a reserved keyword. Using a variable name as `class` will produce an error. To avoid this conflict, you can add a trailing underscore as a naming convention i.e. `class_`.
+-  **Double leading underscores**: Double leading underscores are typically used for name mangling.
+Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses.
+- **Store Expression Value** Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable.example-
```
CaedenPH, line 9:
```suggestion
- **Store Expression Value** Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable. Example:
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,37 @@
+**Use cases of Underscore in Python**
+
+- **Dunders**: Double Underscore (Dunder) methods are special methods defined in a class that are invoked implicitly. For example, `__name__`,`__init__`,`__repr__`, and `__str__`.
+Use `!dunder-methods` command to know more about these methods
```
CaedenPH, line 4:
```suggestion
Use the `!dunder-methods` tag to know more about these methods
```

CaedenPH (commented):
Looks good, but could use some rephrasing in certain areas

bot/resources/tags/under.md
```diff
@@ -0,0 +1,37 @@
...
+- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. Remember that these objects can be accessed outside or other script.
+- **Single Trailing Underscores**: Avoid naming conflict while using a variable which is a reserved keyword. Using a variable name as `class` will produce an error. To avoid this conflict, you can add a trailing underscore as a naming convention i.e. `class_`.
+-  **Double leading underscores**: Double leading underscores are typically used for name mangling.
+Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses.
```
shtlrs, line 8: I think it would be helpful to link some article that explains it better here

bot/resources/tags/under.md
shtlrs, line 9: It is indeed correct that it can be used as a variable, but I think we should state that we usually only use it for variables that are not needed from return statements, iterations, etc.
I know you've mentioned later on, but here's how I think it should be
"Underscores can also be used for variable assignments, but they are generally used to store values that will not be needed."
Then you can give your detailed examples (i see you do that later down the tag) where you say how they're using in return statements, and maybe in a loop, but stating the **context** for each example.
We should always keep in mind that these examples need to be as understandable as they can be, and we especially want to reduce frustrations for total beginners.
Let me know if it's not clear enough and I'll gladly assist you

shtlrs, line 9: Maybe "python shell" is better than interpreter here ? We can wait for other people's opinions here as well

bot/resources/tags/under.md
```diff
@@ -0,0 +1,37 @@
...
+>>> _ + 2
+    4
+```
+-  **Ignore Values**: An underscore can be used to ignore values and as a variable in looping.
+ ```python
+>>> a, *_, b = (3, 5, 2, 6, 2)
+>>> print(a, b, _)
+     3 2 [5, 2, 6]
+>>> for _ in range(2):  # variable in looping
+          print(_)
```
shtlrs, line 24: Looks like you have one extra indentation level here, can you fix that please ?

shtlrs (changes requested):
Looks good indeed, thanks for this ! 

I've left some small comments that I think should be addressed.

bot/resources/tags/under.md
```diff
@@ -0,0 +1,35 @@
+**Use cases of Underscore in Python**
+
+- **Dunders**: Double Underscore (Dunder) methods are special methods defined in a class that are invoked implicitly. For example, `__name__`,`__init__`,`__repr__`, and `__str__`.
+Use the `!dunder-methods` tag to know more about these methods
+- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. Remember that these objects can be accessed outside or other script.
+- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. However, remember that these objects **can** be accessed outside.
```
shtlrs, line 6: Seems you have the "Single leading underscore" duplicated.
I think you were referring to "Single trailing underscore", right ? Description needs fixing as well

shtlrs, line 6: I think it would be better if we phrase it like this:
These mean that they're used in front a variable, a function or a method name, which indicates that these objects are meant to be used internally, such as `_name`. However, these objects **can still** be accessed from outside

We could also benefit from saying that the usage of these conventions is to remediate to the lack of "access modifiers" in Python, we'll let others give their opinion on this as well.

bot/resources/tags/under.md
```diff
@@ -0,0 +1,35 @@
...
+-  **Double leading underscores**: Double leading underscores are typically used for name mangling.
+Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses.[GeeksforGeeks](https://www.geeksforgeeks.org/name-mangling-in-python/)
```
shtlrs, line 8: I think  you should say something like: More details & examples about name mangling can be found [here](https://www.geeksforgeeks.org/name-mangling-in-python/).

bot/resources/tags/under.md
```diff
@@ -0,0 +1,35 @@
...
+- **Store Expression Value** Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable. Example:
+```python
+>>> 1 + 1
+    2
+>>> _
+    2
+>>> _ + 2
+    4
+```
+-  **Ignore Values**: Underscores can also be used for variable assignments, but they are generally used to store values that will not be needed.It is known as `throwaway variable`
```
shtlrs, line 18:
```suggestion
-  **Ignore Values**: Underscores can also be used for variable assignments, but they are generally used that way to store values that will not be needed. It is known as `throwaway variable`
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,35 @@
+**Use cases of Underscore in Python**
+
+- **Dunders**: Double Underscore (Dunder) methods are special methods defined in a class that are invoked implicitly. For example, `__name__`,`__init__`,`__repr__`, and `__str__`.
+Use the `!dunder-methods` tag to know more about these methods
+- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. Remember that these objects can be accessed outside or other script.
+- **Single leading Underscore**:  In front of a variable, a function, or a method name means that these objects are used internally, such as `_name`. However, remember that these objects **can** be accessed outside.
+-  **Double leading underscores**: Double leading underscores are typically used for name mangling.
+Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses.[GeeksforGeeks](https://www.geeksforgeeks.org/name-mangling-in-python/)
+- **Store Expression Value** Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable. Example:
```
shtlrs, line 9: You still didn't explain that if this variable has been assigned once, it won't allow the latest evaluated value from the REPL anymore.
Example
```python
>>> 1 + 1
2
>>> 2 + _
4
>>> _
4
>>> _ = 12
>>> 1 + 1
2
>>> 2 + _
14
>>>
```

shtlrs (changes requested):

bot/resources/tags/under.md
```diff
@@ -0,0 +1,41 @@
+**Use cases of Underscore in Python**
+
+- **Dunders**: Double Underscore (Dunder) methods are special methods defined in a class that are invoked implicitly. For example, `__name__`,`__init__`,`__repr__`, and `__str__`.
+Use the `!dunder-methods` tag to know more about these methods.
```
shtlrs, line 4:
```suggestion
Use the `!dunder-methods` tag to know more about them.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,41 @@
...
+- **Single leading Underscore**:  They're used in front a variable, a function or a method name, which indicates that these objects are meant to be used internally, such as `_name`. However, these objects **can** still be accessed from outside.
```
shtlrs, line 5: Nit: Single Leading Underscores

bot/resources/tags/under.md
```diff
@@ -0,0 +1,41 @@
...
+- **Single Trailing Underscores**: They are used to avoid naming conflict while using a variable, which is a reserved keyword. Using a variable name as `class` will produce an error. To avoid this conflict, you can add a trailing underscore as a naming convention i.e. `class_`.
```
shtlrs, line 6:
```suggestion
- **Single Trailing Underscores**: They are used to avoid naming conflict while making a new variable whose name might conflict with a reserved keyword or package name, etc. For example, using `class` as a variable name will produce an error; and to avoid this conflict, you can add a trailing underscore to it, i.e. `class_`.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,41 @@
...
+-  **Double leading underscores**: Double leading underscores are typically used for Name Mangling. Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses. More details regarding Name Mangling can be found [here](https://www.geeksforgeeks.org/name-mangling-in-python/).
```
shtlrs, line 7:
```suggestion
-  **Double Leading Underscores**: Double leading underscores are typically used for "Name Mangling". Name mangling is a process by which the interpreter changes the attribute name to avoid naming collisions in subclasses. More details regarding this topic can be found [here](https://www.geeksforgeeks.org/name-mangling-in-python/).
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,41 @@
...
+- **Store Expression Value**: Underscore(**_**) is used to store the value of last expression in an interpreter and can be used as a variable. Example:
+```python
+>>> _ # Using underscore without any previously executed expression or value assigned to underscore
+    NameError: name '_' is not defined
+>>> 1 + 1  # evaluating expression
+    2
+>>> _        # prints value of last executed expression
```
shtlrs, line 14:
```suggestion
>>> _   # Prints the value of the last executed expression
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,41 @@
...
+    2
+>>> _ + 2
+    4
+>>> _ = 3  # Once assigned any value,Underscore doesn't store the value of last executed expression anymore
```
shtlrs, line 18:
```suggestion
>>> _ = 3  # Once a value has explicitly been assigned to it, underscore won't store the value of last executed expression anymore.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,41 @@
...
+>>> 5 + 9
+    14
+>>> _
+    3
+```
+-  **Ignore Values**: Underscores can also be used for variable assignments, but they are generally used that way to store values that will not be needed. It is known as `throwaway variable`.
```
shtlrs, line 24:
```suggestion
-  **Ignore Values**: We previously stated that underscores can also be used for variable assignments, but they are most commonly used that way to store values that will not be needed. It is known as `throwaway variable`.
```

bot/resources/tags/under.md
```diff
@@ -0,0 +1,41 @@
...
+ ```python
+>>> a, *_, b = (3, 5, 2, 6, 2) # Ignore values when unpacking
+>>> print(a, b, _)
+    3 2 [5, 2, 6]
+>>> for _ in range(10):  # Ignoring the Index
```
shtlrs, line 29:
```suggestion
>>> for _ in range(10):  # Here the index is ignored since we only want to execute the same function multiple times
```

shtlrs (changes requested):
Sorry if this is taking a bit long with the review.

It is important that we maintain good quality and make this as explicit and as understandable as we could for everyone.

These should be the last changes I would request of you.

Thanks ! :D

shtlrs (approved):
Looks good to me, thank you ! :D

wookie184 (changes requested):
IMO this tag is currently much too long, I had to zoom out to even fit it on my screen: 
image

I've had a go trying to condense this down. I would even consider removing the code blocks but it may not be necessary.
image


    **Meanings of Underscores in Identifier Names**

    • `__name__`: Used to implement special behaviour, such as the `+` operator for classes with the `__add__` method. [More info](https://dbader.org/blog/python-dunder-methods)
    • `_name`: Indicates that a variable is "private" and should only be used by the class or module that defines it.
    • `name_`: Used to avoid naming conflicts. For example, as `class` is a keyword, you could call a variable `class_` instead.
    • `__name`: Causes the name to be "mangled" if defined inside a class. [More info](https://docs.python.org/3/tutorial/classes.html#private-variables)

    A single underscore, **`_`**, has multiple uses:
    • To indicate an unused variable, e.g. in a for loop if you don't care which iteration you are on
    ```python
    for _ in range(10):
        print("Hello World")
    ```
    •  In the REPL, where the previous result will assigned to the variable `_`
    ```python
    >>> 1 + 1  # Evaluated and stored in `_`
        2
    >>> _ + 3  # Take the previous result and add 3 
        5
    ```
    • In integer literals, e.g. `x = 1_500_000` can be written instead of `x = 1500000` to improve readability
    • As a [wildcard in match/case statements](https://peps.python.org/pep-0636/#adding-a-wildcard)

    See also ["Reserved classes of identifiers"](https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers) in the Python docs, and [this more detailed guide](https://dbader.org/blog/meaning-of-underscores-in-python).

PR #2357, opened 2022-12-07
Update modlog for threads

ChrisLovering:
With some discussion with mods, it was agreed that logging thread creation wasn't needed, for the same reasons we do not log message creations.

Thread edits & deletes are still wanted, but mod-log doesn't seem like the right place for it. Since it serves a similar purpose to our message logs, we decided to place it there instead.


jchristgit (approved):
Awesome pull request!!!

wookie184:
Do we want to move thread name edits too? I don't really mind

MarkKoz (approved):

PR #2372, opened 2022-12-31
Display jump url to message invoking an infraction in infraction log

### LINKED ISSUE #2329: Add link to infaction message in infraction log
A message link to where an infraction was issued could provide better context regarding an infraction. This feature has been suggested multiple times in #mod-meta and #mod-tools.
If the infraction was issued in a modmail category, the infraction log field for this feature will instead display "Infraction invoked in ModMail channel" or something similar.  In future, the modmail log link for the thead can be displayed.
 

This feature will also require PR to the site repo since a new field for the jump url has to be added to the infraction model.

(Zig has approved this issue in a private channel).

vivekashok1221:
Closes #2329 .

Related site issue: python-discord/site#801
Related site PR: python-discord/site#824

screenshots:

Infraction log:
image

#mod-logs embed: 
image


bot/exts/moderation/infraction/_scheduler.py
```diff
@@ -273,6 +274,7 @@ async def apply_infraction(
                 Member: {messages.format_user(user)}
                 Actor: {ctx.author.mention}{dm_log_text}{expiry_log_text}
                 Reason: {reason}
+                Jump url: {jump_url_text}
```
shtlrs, line 277:
```suggestion
                Jump URL: {jump_url_text}
```

For consistency with the one in `management.py

bot/exts/moderation/infraction/_utils.py
```diff
@@ -94,6 +95,11 @@ async def post_infraction(
 
     current_time = arrow.utcnow()
 
+    if is_in_category(ctx.channel, Categories.modmail):
+        jump_url_text = "Infraction issued in a ModMail channel."
+    else:
+        jump_url_text = f"[Click here]({ctx.message.jump_url})"
```
shtlrs, line 101: I guess infraction messages would never be deleted, would they ? 
otherwise that'll be a blank jump, but i think it's unlikely for something like that to hapen.

Because what I'm thinking is, if an admin or mod expect to find the infraction message, but they don't, it could potentially be confusing.

So storing message id then fetching it would give more context, but again, i think it's almost impossible to happen

shtlrs (changes requested):
A couple of nit changes here and there but I haven't tested this yet.

bot/exts/moderation/infraction/_utils.py
shtlrs: @vivekashok1221 After some testing, and when I tried deleting the message, the jump URL will take you to the latest message sent in that channel, which is confusing.

Again, this is a very unlikely case to happen so it might not even be worth dealing with

vivekashok1221: Back when we discussed this feature in our internal mod channel, we agreed that it was okay to leave dead links as is because, you're right, it's not something that happens often.

shtlrs: Well, in that case, I'll resolve this then.
Thanks !

bot/exts/moderation/infraction/_scheduler.py
vivekashok1221: Will do, thanks :+1: .

bot/exts/moderation/infraction/management.py
```diff
@@ -432,6 +437,7 @@ def infraction_to_string(self, infraction: t.Dict[str, t.Any]) -> str:
             Duration: {duration}
             Actor: <@{infraction["actor"]["id"]}>
             ID: `{infraction["id"]}`
+            Jump Url: {jump_url_text}
```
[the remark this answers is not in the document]
vivekashok1221: Done (https://github.com/python-discord/bot/pull/2372/commits/ed59a9fd38f8bcfcb81efaf66f86e7ebc6661665).

bot/exts/moderation/infraction/_scheduler.py
vivekashok1221: Done (https://github.com/python-discord/bot/pull/2372/commits/ed59a9fd38f8bcfcb81efaf66f86e7ebc6661665).

wookie184 (approved):
LGTM

bot/exts/moderation/infraction/_utils.py
```diff
@@ -94,6 +95,14 @@ async def post_infraction(
 
     current_time = arrow.utcnow()
 
+    if any(
+        is_in_category(ctx.channel, category)
+        for category in (Categories.modmail, Categories.appeals, Categories.appeals2)
+    ):
+        jump_url = "Infraction issued in a ModMail channel."
```
ChrisLovering, line 102: Either something has a jump link or it doesn't, storing a string in here too means we can't have a validator on the db that it's a valid link.

IMO we should have a validator on site that requires it to be formatted as a link, or null.

bot/exts/moderation/infraction/management.py
```diff
@@ -420,6 +421,13 @@ def infraction_to_string(self, infraction: t.Dict[str, t.Any]) -> str:
         else:
             dm_sent_text = "Yes" if dm_sent else "No"
 
+        if jump_url == "":
+            # Infraction was issued prior to jump urls being stored in the database.
+            jump_url = "N/A"
+        elif "discord.com" in jump_url:
+            jump_url = f"[Click here.]({jump_url})"
+            # Else, infraction was issued in ModMail category.
```
ChrisLovering, line 429: This would be simplified if we went with the "it's either a link or null"

bot/exts/moderation/infraction/_utils.py
vivekashok1221: I could be misremembering but Ive read somewhere (django docs or stackoverflow) that using blank strings is recommended by django ORM. I can look for the link later.

vivekashok1221: Maybe it's just a convention.

bot/exts/moderation/infraction/management.py
vivekashok1221: Yeah, I did think of that.

bot/exts/moderation/infraction/_utils.py
vivekashok1221: And regarding the validator, I could use URLField instead of CharField because it has URL validation.

ChrisLovering: A URLField that can be blank/null sounds good to me

Robin5605:
Have we considered a URL button for the embed? Or would that be too big?

vivekashok1221:
> Have we considered a URL button for the embed? Or would that be too big?

Personally, I don't think it's necessary. I'm open to opinions though.

shtlrs:
That would make two buttons at most in the same message when listing infractions. Then we  lose context of which infraction does that button refer to, no? Which might become confusing.

vivekashok1221:
I will fix the tests as soon as I get some time but I'm not sure why the tests didn't fail during my earlier pushes. Might be random and may even disappear on re-running the tests- nevertheless, will try fixing.

Passes locally btw
```bash
poetry run task test
=============================== test session starts ================================
platform linux -- Python 3.10.9, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/vivek/Programs/python/contrib/pydis/bot, configfile: pyproject.toml
plugins: subtests-0.9.0, xdist-3.0.2, cov-4.0.0
gw0 [473] / gw1 [473] / gw2 [473] / gw3 [473] / gw4 [473] / gw5 [473] / gw6 ok / gw7gw0 [473] / gw1 [473] / gw2 [473] / gw3 [473] / gw4 [473] / gw5 [473] / gw6 [473] / gw0 [473] / gw1 [473] / gw2 [473] / gw3 [473] / gw4 [473] / gw5 [473] / gw6 [473] / gw7 [473]
............................................................................ [ 16%]
............................................................................ [ 32%]
............................................................................ [ 48%]
............................................................................ [ 64%]
............................................................s............... [ 80%]
............................................................................ [ 96%]
.................                                                            [100%]
========================= 472 passed, 1 skipped in 20.26s ==========================
```

shtlrs (approved):
🧙

PR #2374, opened 2023-01-01
Make the infraction message mention @ModMail

swfarnsworth:
Previously, the footer for non-ban infraction messages would say to DM the ModMail bot. This commit makes "ModMail" a clickable mention of the bot.

image

When this footer was introduced, we decided not to make it a mention, in case the recipient did not have the ModMail bot account in their user cache (which would cause it to render as the account ID, ie "send a message to 683001325440860340"). Chris has determined that because the bot is at the top of the user list, it seems to always be in the cache for any guild member, even without visiting the guild or viewing the user list.

### DIFF
--- a/bot/exts/moderation/infraction/_utils.py
+++ b/bot/exts/moderation/infraction/_utils.py
@@ -31,12 +31,13 @@ RULES_URL = "https://pythondiscord.com/pages/rules"
 Infraction = t.Dict[str, t.Union[str, int, bool]]
 
 APPEAL_SERVER_INVITE = "https://discord.gg/WXrCJxWBnm"
+MODMAIL_ACCOUNT_ID = "683001325440860340"
 
 INFRACTION_TITLE = "Please review our rules"
 INFRACTION_APPEAL_SERVER_FOOTER = f"\nTo appeal this infraction, join our [appeals server]({APPEAL_SERVER_INVITE})."
 INFRACTION_APPEAL_MODMAIL_FOOTER = (
     '\nIf you would like to discuss or appeal this infraction, '
-    'send a message to the ModMail bot.'
+    f'send a message to <@{MODMAIL_ACCOUNT_ID}>.'
 )
 INFRACTION_AUTHOR_NAME = "Infraction information"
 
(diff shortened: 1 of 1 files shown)


ChrisLovering (approved):

bot/exts/moderation/infraction/_utils.py
```diff
@@ -31,12 +31,13 @@
...
 
 APPEAL_SERVER_INVITE = "https://discord.gg/WXrCJxWBnm"
+MODMAIL_ACCOUNT_ID = "683001325440860340"
 
 INFRACTION_TITLE = "Please review our rules"
 INFRACTION_APPEAL_SERVER_FOOTER = f"\nTo appeal this infraction, join our [appeals server]({APPEAL_SERVER_INVITE})."
 INFRACTION_APPEAL_MODMAIL_FOOTER = (
     '\nIf you would like to discuss or appeal this infraction, '
-    'send a message to the ModMail bot.'
+    f'send a message to <@{MODMAIL_ACCOUNT_ID}>.'
```
vivekashok1221, line 40: How about having the mention inside a pair of parentheses instead?
```suggestion
    f'send a message to the ModMail bot (<@{MODMAIL_ACCOUNT_ID}>).'
```

vivekashok1221 (changes requested):

bot/exts/moderation/infraction/_utils.py
CaedenPH, line 40: What is the difference when having `<@{MODMAIL_ACCOUNT_ID}>` opposed to `<@!{MODMAIL_ACCOUNT_ID}>`?

ChrisLovering: Since this is sent in a DM channel, the bot doesn't have a nickname, so the ! will do nothing right?

swfarnsworth: image

This is how `<@!683001325440860340>` rendered for me in my DMs with the Python bot. But <@683001325440860340>` renders the same way.

wookie184: I don't mind either way, thoughts @swfarnsworth?

vivekashok1221: I don't really mind either way too, for the  record.

swfarnsworth:
For whatever this is worth: I just tested this in my test server by warning a dummy account that hasn't viewed the main guild in a long time, and the ModMail bot was not in that account's user cache until I clicked through to the guild.

vivekashok1221 (approved):
LGTM :star2: .

wookie184:
> For whatever this is worth: I just tested this in my test server by warning a dummy account that hasn't viewed the main guild in a long time, and the ModMail bot was not in that account's user cache until I clicked through to the guild.

Ah interesting, it probably is good that we mention it's the modmail bot then 👍

wookie184 (approved):
Thanks

PR #2375, opened 2023-01-03
Allow passing ful channel objects to check if the channel is ignored

ChrisLovering:
This means that the is_channel_ignored fucntion can be used with on_*_delete events that have the channel object, but can't be retreived fromthe cache by the is_channel_ignored function


wookie184 (approved):

CaedenPH (approved):
Code seems fine, but where would this be necessary?

shtlrs (approved):
LGTM!
I didn't test this, but it looks pretty straightforward

PR #2378, opened 2023-01-12
Add questions tag

meatballs:
Because I find myself pointing people to [Asking Good Questions](https://www.pythondiscord.com/pages/guides/pydis-guides/asking-good-questions/) regularly.

### DIFF
--- a/bot/resources/tags/questions.md
+++ b/bot/resources/tags/questions.md
@@ -1,0 +1,3 @@
+**Asking Good Questions**
+
+The [Asking Good Questions](https://www.pythondiscord.com/pages/guides/pydis-guides/asking-good-questions/) page on our website contains guidance on how to ask a question so that you get help as quickly and effectively as possible.
--- a/bot/resources/tags/questions.md
+++ b/bot/resources/tags/questions.md
@@ -1,3 +1,5 @@
-**Asking Good Questions**
-
+---
+embed:
+    title: "Asking Good Questions"
+---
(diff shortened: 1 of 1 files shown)


CaedenPH (commented):
Just my opinion, but this seems a little bare. Perhaps some other information could be useful

meatballs:
What do you think it needs? The page it links to has comprehensive detail and I didn't want to just repeat any of that.

brodycritchlow (approved):
Don't see any point in adding anymore. The link provides enough detail. Looks good

Robin5605 (approved):
Short, concise, and links to a more detailed article. A perfect combination for tags. LGTM :+1:

bot/resources/tags/questions.md
```diff
@@ -0,0 +1,3 @@
+**Asking Good Questions**
```
shtlrs, line 1: This needs to change to comply with the news tags slash command
#2419 is your friend

shtlrs (changes requested):

HassanAbouelela:
Blocked for further discussion, please open an issue in python-discord/meta.

mbaruh:
Staff consensus in today's meeting is that we don't want to go forward with it, largely for the same reasons we removed the old !ask tag. The link doesn't clarify what exactly is wrong with the question asked, which is very context specific, and has a high chance of being used passive-aggressively by being dumped at people (sometimes several times) with no further explanation, as we have seen with the old tag.

PR #2384, opened 2023-01-22
Make sure interaction_owner is an instance attribute

mbaruh:
If it's a class attribute then the interaction owner of all current views will be the last person to view the buttons.


Robin5605 (approved):
Should the `interaction_owner` docstring still be kept around? Theoretically speaking one could still access it via `role_button_view_instance.interaction_owner` given that `role_button_view_instance` is an instance of `RoleButtonView`

PR #2386, opened 2023-01-23
Update help post messages

### LINKED ISSUE #1295: Improving dormant help channel embed
Following the successful implementation of an updated available help channel embed, I think it's time that we update the dormant help channel embed as well. There are two main reasons and goals for wanting to update this:
- Improving consistency
- Improving the dormant message
  - Telling the user what to do if the question wasn't answered

### Current help channel embeds
Here is the new available help channel embed:
Screenshot 2020-11-21 at 01 14 50

This is the current dormant embed:
Screenshot 2020-11-21 at 01 15 41

### Implementation
I'm currently considering the following styling:
- A red embed
- Uses a red `x` icon

ChrisLovering:
Closes #1295 

Preview:
image


wookie184 (approved):
lgtm

MarkKoz (approved):
It'd be more appropriate to call it a "post" than a "channel", but we can revisit that in the future if we want. I don't think it's that significant either way.

PR #2388, opened 2023-01-25
Add new inplace tag

### LINKED ISSUE #2360: `!in-place` tag
A fairly common problem asked in the help channels involves in-place operations like `list.sort()` which return `None`. A simple tag could be used to explain the difference in behaviour between in-place and out-of-place functions (with examples such as `sorted()` vs `list.sort()`) and how common issues involving them can be fixed.

brodycritchlow:
Closes #2360 

Would love to hear feedback on if the tag explanation is good or not as it is my first time contributing.


bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,23 @@
...
+In programming, there are two types of operations: "out of place" and "in place". An "out of place" operation creates a new object, leaving the original object unchanged. An "in place" operation modifies the original object, without creating a new one. These operations return None explicitly.
+
+A common example of these different concepts is seen in the use of the methods `list.sort()` and sorted(...) in Python. Using `list.sort()` will modify the original list and return None, so attempting to access an element of the list after calling `sort()` will result in an error. 
+
+For example, the following code will result in an error:
+
+```py
+a_list = [3, 1, 2]
+a_new_list = a_list.sort() # This will be None
+print(a_new_list[1]) # This will error
```
CaedenPH, line 12:
```suggestion
print(a_new_list[1]) # This will error because it is empty
```
Is this part necessary?

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,23 @@
...
+```
+
+On the other hand, using `sorted(...)` will return a new sorted list, leaving the original list unchanged. This means that if you expect the original list to be sorted, you will be disappointed with the result. For example, the following code will print 3 instead of 1:
```
CaedenPH, line 15:
```suggestion
On the other hand, using the function `sorted(...)` will return a new sorted list, leaving the original list unchanged. This means that if you expect the original list to be sorted, you will be disappointed with the result. For example, the following code will print 3 instead of 1:
```

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,23 @@
...
+
+```py
+a_list = [3, 1, 2]
+sorted(a_list)
+print(a_list[0]) # You may expect 1, but it will print 3
+```
+
+To avoid these errors and unexpected results, it is required to assign the result of sorted(...) to a new variable and use list.sort() method in the original list. This way, the original list will be sorted and the new list will be created with the sorted elements.
```
CaedenPH, line 23:
```suggestion
To avoid these errors and unexpected results, it is required to assign the result of `sorted(...)` to a new variable and use `list.sort()` method in the original list. This way, the original list will be sorted and the new list will be created with the sorted elements.
```

CaedenPH (approved):
Looks good to me, but I'm sure others with have things to say

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,17 @@
+**Out of Place** and **In Place**
+
+In programming, there are two types of operations: "out of place" operations creates a new object, leaving the original object unchanged. "in place" operations modifies the original object, without creating a new one. These operations return None explicitly.
```
dawnofmidnight, line 3:
```suggestion
In programming, there are two types of operations: "out of place" operations create a new object, leaving the original object unchanged. "in place" operations modify the original object, without creating a new one. These operations return None explicitly.
```

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,17 @@
...
+
+A common example of these different concepts is seen in the use of the methods `list.sort()` and `sorted(...)`. Using `list.sort()` and attempting to access an element of the list will result in an error. 
+
+```py
+a_list = [3, 1, 2]
+a_new_list = a_list.sort() # This will be None
+print(a_new_list[1]) # This will error because it is NoneType and not a list
+
+a_list = [3, 1, 2]
```
dawnofmidnight, line 12: It may be a bit clearer to use a new name here.

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,17 @@
...
+sorted(a_list)
+print(a_list[0]) # You may expect 1, but it will print 3
+```
+
```
dawnofmidnight, line 16: This empty line can be removed, Discord's rendering auto-inserts one after code blocks.

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,17 @@
...
+To avoid these errors and unexpected results, it is required to assign the result of `sorted(...)` to a new variable and use `list.sort()` method in the original list. This way, the original list will be sorted and the new list will be created with the sorted elements.
```
dawnofmidnight, line 17: I'm not sure how I like "it is required" here, as well as the focus on lists. My proposal here would something along the lines of:

> To avoid these errors and unexpected results, you should either use an in-place operation (`sorted(...)`) and assign it to a variable or use an out-of-place operation (`list.sort()`) without assignment.

The wording can obviously vary, but I'd prefer to see something general like that.

dawnofmidnight (changes requested):
Thanks! There are a couple wording changes that I think would be beneficial for clarity here, but this is overall pretty good.

bot/resources/tags/in-place.md
brodycritchlow: For both or what exact list name

bot/resources/tags/in-place.md
brodycritchlow: Seems good to me, I chose required because I couldn't really think of anything else in the moment.

bot/resources/tags/in-place.md
brodycritchlow: Ah i see, resolved

bot/resources/tags/in-place.md
n0Oo0Oo0b: I think variable names here should be more in-line with what the person was trying to do, so I would call them `unsorted_list` and `sorted_list` for the first example and `list_to_sort` in the second one.

I also think the way you show the unexpected behavior (with indexing) is overcomplicated, something like
```py
print(a_new_list)  # Outputs None. Where did the list go?
```
and
```py
print(outofplace_list)  # The list still isn't sorted. Why?
```
would make it easier to understand for beginners.

brodycritchlow: @n0Oo0Oo0b Seems like a good idea, updated.

n0Oo0Oo0b (changes requested):
I still think the list variable names could still be improved (not necessarily to what I suggested before) because having inplace/outofplace in the names suggest that it's a property of the lists, not the functions that you run.

brodycritchlow:
> I still think the list variable names could still be improved (not necessarily to what I suggested before) because having inplace/outofplace in the names suggest that it's a property of the lists, not the functions that you run.

What would be some better names, I mean they could be generically named `a_list`, and `another_list`.

n0Oo0Oo0b:
I think `unsorted_list`/`sorted_list` for the first example and `list_to_sort` for the second will be fine, probably with a `# Wrong:` line at the top as well to show that it isn't the correct way to use the functions

n0Oo0Oo0b (approved):
Looks good, approved.

brodycritchlow:
@dawnofmidnight availabe to rereview?

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,19 @@
...
+
+In programming, there are two types of operations: "out of place" operations create a new object, leaving the original object unchanged. "in place" operations modify the original object, without creating a new one. These operations return None explicitly.
+
+A common example of these different concepts is seen in the use of the methods `list.sort()` and `sorted(...)`. Using `list.sort()` and attempting to access an element of the list will result in an error. 
+
+```py
+# WRONG:
+
+unsorted_list = [3, 1, 2]
+sorted_list = inplace_list.sort() # This will be None
```
dawnofmidnight, line 11:
```suggestion
sorted_list = unsorted_list.sort() # This will be None
```

dawnofmidnight (changes requested):
This looks mostly good to me, apart from a small fix and the proposed re‐wording in my previous review.

bot/resources/tags/in-place.md
brodycritchlow: 👍, updated.

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,19 @@
...
+
+In programming, there are two types of operations: "out of place" operations create a new object, leaving the original object unchanged. "in place" operations modify the original object, without creating a new one. These operations return None explicitly.
+
+A common example of these different concepts is seen in the use of the methods `list.sort()` and `sorted(...)`. Using `list.sort()` and attempting to access an element of the list will result in an error. 
+
+```py
+# WRONG:
+
+unsorted_list = [3, 1, 2]
+sorted_list = unsorted_list.sort() # This will be None
```
Ibrahim2750mi, line 11: [PEP 257](https://peps.python.org/pep-0008/#inline-comments) 
> An inline comment is a comment on the same line as a statement. Inline comments should be separated by at least two spaces from the statement. They should start with a # and a single space.

Should be changed to
```python
sorted_list = unsorted_list.sort()  # This will be None
```

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,19 @@
+**Out of Place** and **In Place**
```
Ibrahim2750mi, line 1: You should add it as embed title according to this #2419

Ibrahim2750mi (changes requested):

brodycritchlow:
@dawnofmidnight are there still changes you would like me to do, I think I have resolved all of them.

Ibrahim2750mi (approved):
LGTM!

bot/resources/tags/in-place.md
```diff
@@ -0,0 +1,22 @@
+---
+embed:
+    title: "Out of place and in place"
+---
+
+In programming, there are two types of operations: "out of place" operations create a new object, leaving the original object unchanged. "in place" operations modify the original object, without creating a new one. These operations return None explicitly.
```
dawnofmidnight, line 6:
```suggestion
In programming, there are two types of operations: "out of place" operations create a new object, leaving the original object unchanged. "in place" operations modify the original object without creating a new one, and return `None` explicitly.
```

brodycritchlow:
Sorry, Im a little slow at seeing the reviews.

dawnofmidnight (approved):
Thanks for entertaining all these changes! I know I've made a lot of comments 😅

PR #2389, opened 2023-01-26
Fix typo in help channel opening embed

shenanigansd:


MarkKoz (approved):
This has caught me off before. However, there is a grammatically correct interpretation as-is, though it isn't the first one I come to naturally. It can be read as "If you got one" -> "if you got an error message" -> "if you got an error message when you ran the code"

MarkKoz:
Ah well we have #2386 so there probably is no point in merging this. It'll just create conflicts and the impact of the issue this solves is minimal.

dawnofmidnight (approved):
:shipit:

Issue #551, opened 2019-10-23
Ability to temporarily ignore user/watchword pairs in #mod-alerts


dementati:
Occasionally a single user will very frequently cause false positive watchlist triggers, which results in constant moderator pings. This can for example happen if pasted code contains identifiers from a different language that happens to be homonymous with a watchlist word (the Danish word "fag" translates to "subject" in English). It would be nice if it was possible to temporarily disable the watchlist for a particular user and a particular word. I'm imagining a command that takes a watchlist word, a user ID and an expiry time. Perhaps the user ID can be optional, to disable the watchlist for all users temporarily. Maybe the expiry can be optional with a reasonable default.

Akarys42:
Superseded by #1530

PR #2390, opened 2023-01-28
New Filtering System

### LINKED ISSUE #1018: Race condition between anti-spam and anti-malware
An anti-spam `DeletionContext` will re-upload attachments to the `#attachment-log` channel. The anti-malware cog will delete attachments that are blacklisted. This sometimes results in a 404 when anti-spam is trying to upload the attachments:

Sentry Issue: [BOT-6D](https://sentry.io/organizations/python-discord/issues/1738255436/?referrer=github_integration)

```
NotFound: 404 Not Found (error code: 0): asset not found
  File "bot/utils/messages.py", line 81, in send_attachments
    await attachment.save(file, use_cached=True)
  File "discord/message.py", line 124, in save
    data = await self.read(use_cached=use_cached)
  File "discord/message.py", line 166, in read
    data = await self._http.get_from_cdn(url)
  File "discord/http.py", line 235, in get_from_cdn
    raise NotFound(resp, 'asset not found')

Failed to re-upload attachment <REDACTED> from message <REDACTED> with status 404.
```

### LINKED ISSUE #1530: New Filters System Specifications

### LINKED ISSUE #2266: Add triggering filter to alert embed title
The current alert embed for a triggered filter puts the filter type in the main body, where it's easily missed - especially for those of us who have higher-res screens.

Having the triggering filter in the embed's title will give instant access to that information at a glance. Below image is a mockup of what I propose:

image

I am not precious about the exact formatting of the title, as long as the information is in there.

mbaruh:
This pull request implements the new filtering system as described in #1530.

For more information about the database and API changes made to support this, check out the site PR: https://github.com/python-discord/site/pull/861

> **Warning** This PR must be merged at the same time as the site PR.

closes #1530.

## Overview
The main idea is that there is a list of filters each deciding whether they apply to the given content.
For example, there can be a filter that decides it will trigger when the content contains the string "lemon".

* There are several types of filters, and two filters of the same type differ by their content.
For example, filters of type "token" search for a specific token (provided as regex) inside the provided string.
One token filter might look for the string "lemon", while another will look for the string "joe".

* Each filter has a set of settings that decide when it triggers (e.g. in which channels), and what happens if it does (e.g. delete the message).
Filters of a specific type can have additional settings that are special to them.

* A list of filters is contained within a filter list.
The filter list gets content to filter, and dispatches it to each of its filters.
It takes the answers from its filters and returns a unified response (e.g. if at least one of the filters says it should be deleted, then the filter list response will include it).

* A filter list has the same set of possible settings, which act as defaults.
If a filter in the list doesn't define a value for a setting (meaning it has a value of None), it will use the value of the containing filter list.

* The cog receives "filtering events". For example, a new message is sent.
It creates a "filtering context" with everything a filtering list needs to know to provide an answer for what should be done.
For example, if the event is a new message, then the content to filter is the content of the message, embeds if any exist, etc.

* The cog dispatches the event to each filter list, gets the result from each, compiles them, and takes any action dictated by them.
For example, if any of the filter lists want the message to be deleted, then the cog will delete it.

## Main Features
* Renames the filter lists to have simpler and more memorable names: token, domain, invite, extension.
* Each filter has a set of settings dictating when it should trigger, and what should happen if it does:
  *  `bypass_roles` - A list of role IDs or role names. Users with these roles will not trigger the filter.
  * `enabled` - A boolean field. Setting it to False allows disabling the filter without deleting it entirely.
  * `filter_dm` - A boolean field. If True, the filter can trigger for messages sent to the bot in DMs.
  * `disabled_channels` - A list of channel IDs or channel names. The filter will not trigger in these channels even if the category is expressly enabled.
  * `disabled_categories` - A list of category IDs or category names. The filter will not trigger in these categories.
  * `enabled_channels` - A list of channel IDs or channel names. The filter can trigger in these channels even if the category is disabled or not expressly enabled.
  * `enabled_categories` - A list of category IDs or category names. If the list is not empty, filters will trigger only in channels of these categories, unless the channel is expressly disabled.
  * `send_alert` - A boolean. If all filters triggered set this to False, no mod-alert will be created.
  * `remove_context` - A boolean field. If True, the filter being triggered will cause the offending context to be removed. An offending message will be deleted, while an offending nickname will be superstarified.
  * `guild_pings` - A list of role IDs/role names/user IDs/user names/here/everyone. If a mod-alert is generated for a filter triggered in a public channel, these will be pinged.
  * `dm_pings` - A list of role IDs/role names/user IDs/user names/here/everyone. If a mod-alert is generated for a filter triggered in DMs, these will be pinged.
  * `infraction_type` - The type of infraction to issue when the filter triggers, or 'NONE'. If two infractions are triggered for the same message, the harsher one will be applied (by type or duration).
  * `infraction_duration` - How long the infraction should last for in seconds. 0 for permanent.
  * `infraction_reason` - The reason delivered with the infraction.
  * `infraction_channel` - The channel ID in which to invoke the infraction (and send the confirmation message). If 0, the infraction will be sent in the context channel. If the ID otherwise fails to resolve, it will default to the mod-alerts channel.
  * `dm_content` - The contents of a message to be DMed to the offending user.
  * `dm_embed` - The contents of the embed to be DMed to the offending user.
* The antimalware and antispam cogs become filter lists and are treated as regular parts of the filtering system.
* The ability to search for filters using an example message or the values of their settings.
* Identifying filters by their ID rather than what they filter (previously one had to write the exact regex of the filter in order to delete it).
* The ability to edit filters after they were added.
* All filters are checked against a message (or another event), instead of only triggering the first one found as before (closes #1018).
* Allows adding the offending URL to the DM sent.
* Actions (additions, editing, deletions, searches) can be done via a UI (using embeds and Discord message components).

Additionally closes #2266 as the formatting of the alert is significantly changed and should address the issue described.

## Deferred Features
Some features are purposefully not addressed in this PR, as it is already very big, and they are not directly related to the scope of the PR or are minor:
* A command to temporarily whitelist users (#551).
* Not applying the regex to new invites - adding a new invite should allow specifying only the invite code, and the moderator shouldn't have to give the whole URL.
* Thread name filter - the new system is able to accommodate filtering the names of threads.
* Convert to hybrid/slash commands.
* Add a pardon button to mod alerts when an automatic infraction is applied.
* An idea raised was to log all filter additions/changes/deletions in #filter-log.
* Be able to list all settings together with their description through some argument to the `!filter settings` command, instead of the current feature which either lists all of them without descriptions, or shows a specific one with its description.


bot/exts/filtering/_filter_lists/filter_list.py
```diff
@@ -0,0 +1,309 @@
...
+            message = f"#{triggers[0].id} (`{triggers[0].content}`)"
+            if triggers[0].description:
+                message += f" - {triggers[0].description}"
+            messages = [message]
+        else:
+            messages = [f"{filter_.id} (`{filter_.content}`)" for filter_ in triggers]
+        return messages
+
+    def __hash__(self):
+        return hash(id(self))
```
ionite34, line 156: Why are we hashing memory identity of the instance here? Shouldn't the instance `id` be used instead?

bot/exts/filtering/_filter_lists/filter_list.py
```diff
@@ -0,0 +1,309 @@
...
+def list_type_converter(argument: str) -> ListType:
+    """A converter to get the appropriate list type."""
+    argument = argument.lower()
+    for list_type, list_aliases in aliases:
+        if argument in list_aliases or argument in map(past_tense, list_aliases):
+            return list_type
+    raise BadArgument(f"No matching list type found for {argument!r}.")
+
+
+# AtomicList and its subclasses must have eq=False, otherwise the dataclass deco will replace the hash function.
```
ionite34, line 50: Why does these have to be hashable in any case? I'm not sure how it makes sense for something to not be comparable (no `__eq__`) but to be hashable.

bot/exts/filtering/_filters/antispam/role_mentions.py
```diff
@@ -0,0 +1,42 @@
...
+    name = "role_mentions"
+    events = (Event.MESSAGE,)
+    extra_fields_type = ExtraRoleMentionsSettings
+
+    async def triggered_on(self, ctx: FilterContext) -> bool:
+        """Search for the filter's content within a given context."""
+        earliest_relevant_at = arrow.utcnow() - timedelta(seconds=self.extra_fields.interval)
+        relevant_messages = list(takewhile(lambda msg: msg.created_at > earliest_relevant_at, ctx.content))
+        detected_messages = {msg for msg in relevant_messages if msg.author == ctx.author}
+        total_recent_mentions = sum(len(msg.role_mentions) for msg in relevant_messages)
```
ionite34, line 36: This stuff seems kind of repetitive (I think there was at least 4 or 5 usages across the filters), perhaps we can provide some superclass instance function to do this or some default implementation?

bot/exts/filtering/_settings_types/actions/ping.py
```diff
@@ -0,0 +1,47 @@
...
+
+    async def action(self, ctx: FilterContext) -> None:
+        """Add the stored pings to the alert message content."""
+        mentions = self.guild_pings if not ctx.channel or ctx.channel.guild else self.dm_pings
+        new_content = " ".join([resolve_mention(mention) for mention in mentions])
+        ctx.alert_content = f"{new_content} {ctx.alert_content}"
+
+    def __or__(self, other: ActionEntry):
+        """Combines two actions of the same type. Each type of action is executed once per filter."""
+        if not isinstance(other, Ping):
```
ionite34, line 44: Why does the type hint for `other` not match the guard condition for `NotImplemented` here?

bot/exts/filtering/_settings.py
```diff
@@ -0,0 +1,232 @@
...
+    @property
+    def overrides(self) -> dict[str, Any]:
+        """Return a dictionary of overrides across all entries."""
+        return reduce(operator.or_, (entry.overrides for entry in self.values() if entry), {})
+
+    def copy(self: TSettings) -> TSettings:
+        """Create a shallow copy of the object."""
+        copy = self.__class__({})
+        copy.update(super().copy())  # Copy the internal dict.
+        return copy
```
ionite34, line 110: Is this different from `copy.copy(self)`?

bot/exts/filtering/_ui/filter_list.py
```diff
@@ -0,0 +1,271 @@
...
+
+        self.embed.clear_fields()
+        new_view = self.copy()
+
+        try:
+            if isinstance(interaction_or_msg, discord.Interaction):
+                await interaction_or_msg.response.edit_message(embed=self.embed, view=new_view)
+            else:
+                await interaction_or_msg.edit(embed=self.embed, view=new_view)
+        except discord.errors.HTTPException:  # Various errors such as embed description being too long.
```
ionite34, line 148: Should we have some embed length trimming / limitation?

bot/exts/filtering/_utils.py
```diff
@@ -0,0 +1,222 @@
...
+    if override_is_sequence != default_is_sequence:  # One is a sequence and the other isn't.
+        return False
+    if override_is_sequence:
+        if len(override) != len(default):
+            return False
+        return all(str(item1) == str(item2) for item1, item2 in zip(set(override), set(default)))
+    return str(override) == str(default)
+
+
+def starting_value(type_: type[T]) -> T:
```
ionite34, line 122: We should probably avoid accessing `__args__` and `__origin__` directly and use `typing.get_args` and `typing.get_origin` instead, since these apis seem to almost always change between versions, https://canary.discord.com/channels/267624335836053506/891788761371906108/1071456037363789865. 

Checking `if get_origin(type_) in (typing.Union, types.UnionType):` seems it could work here?

ionite34 (changes requested):
Very preliminary few notes

bot/exts/filtering/_settings_types/actions/ping.py
mbaruh: This is something I wasn't very sure about. I didn't want to change the signature of the overridden method. Is changing the type hint considered common?

bot/exts/filtering/_filter_lists/filter_list.py
mbaruh: I tried to make this correct beyond my current usage of it. For example, if I use AtomicList objects as dict keys, I don't want to assume what else will be put in that dict.

bot/exts/filtering/_filter_lists/filter_list.py
mbaruh: Because I'm using the class' instances as dict keys https://github.com/python-discord/bot/blob/dc0ce05ff1406954f5006366b2899c4bde1967ce/bot/exts/filtering/filtering.py#L912

There's no logic to it being hashable but not comparable beyond not needing to compare.

I don't mind returning `hash(("AtomicList", self.id))` or something of that sort, though I'm not sure how it's preferable.

bot/exts/filtering/_settings.py
mbaruh: No, it is not 👍

bot/exts/filtering/_ui/filter_list.py
mbaruh: Fair enough, I'll admit I was being lazy here.

bot/exts/filtering/_utils.py
mbaruh: Good point

bot/exts/filtering/_settings_types/actions/ping.py
decorator-factory: @mbaruh If you want to require the same class, use `typing.Self` in the parent and child classes instead of `ActionPing`.

Also, it's still a good idea to check for `isinstance(other, Ping)` because the right operand might implement `__ror__`...

decorator-factory: Actually, why is this a dunder method and not just a normal method? Dunder methods have quite a lot of hariballish semantics to them (like `NotImplemented`).

bot/exts/filtering/_filter_context.py
```diff
@@ -0,0 +1,62 @@
...
+class FilterContext:
+    """A dataclass containing the information that should be filtered, and output information of the filtering."""
+
+    # Input context
+    event: Event  # The type of event
+    author: User | Member | None  # Who triggered the event
+    channel: TextChannel | Thread | DMChannel | None  # The channel involved
+    content: str | Iterable  # What actually needs filtering
+    message: Message | None  # The message involved
+    embeds: list = field(default_factory=list)  # Any embeds involved
```
decorator-factory, line 34: Why are some of the generics here (like `list`) not parameterized? This is a chonky object, I'd at least want to know what's in it

bot/exts/filtering/_settings_types/actions/infraction_and_notification.py
```diff
@@ -0,0 +1,191 @@
...
+    def convert_infraction_name(cls, infr_type: str | Infraction) -> Infraction:
+        """Convert the string to an Infraction by name."""
+        if isinstance(infr_type, Infraction):
+            return infr_type
+        return Infraction[infr_type.replace(" ", "_").upper()]
+
+    async def action(self, ctx: FilterContext) -> None:
+        """Send the notification to the user, and apply any specified infractions."""
+        # If there is no infraction to apply, any DM contents already provided in the context take precedence.
+        if self.infraction_type is None and (ctx.dm_content or ctx.dm_embed):
```
decorator-factory, line 121: `is None` -> `== Infraction.NONE

bot/exts/filtering/_settings_types/actions/ping.py
mbaruh: `typing.Self` is 3.11, the bot uses 3.10.

Being able to use `|` for this operation just seemed nice here 😅

bot/exts/filtering/_filters/antispam/role_mentions.py
mbaruh: This will have to assume that every antispam filter has an extra setting called `interval` (holds now, but isn't logically guaranteed), and the attribute being summed is different every time, so you either exclude the fourth line, or send a function to fetch the right value from each message. I'll think about it, but at first glance trying to generalize it seems clumsy.

bot/exts/filtering/_settings_types/actions/ping.py
decorator-factory: > `typing.Self` is 3.11, the bot uses 3.10.

Import it from `typing_extensions`

>  seemed nice

Between nicity and simplicity I would surely pick simplicity :)

mbaruh: Fair point. When would you say it would be worth using the dunder then?

decorator-factory: Not sure actually. It is pretty nice if you're making some super general data structure, like a `set` or `enum.IntFlag`. IMO dunders are useful for writing more complex DSL-like expressions, like `foo | (bar & baz - (fizz | buzz))`.

Although Clojure programmers are fine without infix notation at all :shrug: 
```clj
(union foo
  (difference
    (intersection bar baz)
    (union fizz buzz)))
```

bot/exts/filtering/_settings.py
mbaruh: 12e1e37

bot/exts/filtering/_utils.py
mbaruh: 333bbf6

bot/exts/filtering/_ui/filter_list.py
mbaruh: The issue was actually in a different file, the comment here was just a copy paste. b67e849

bot/exts/filtering/_filter_context.py
mbaruh: 2bd9d23

bot/exts/filtering/_settings_types/actions/infraction_and_notification.py
mbaruh: 7184925

bot/exts/filtering/_filter_context.py
```diff
@@ -0,0 +1,62 @@
...
+    before_message: Message | None = None
+    message_cache: MessageCache | None = None
+    # Output context
+    dm_content: str = field(default_factory=str)  # The content to DM the invoker
+    dm_embed: str = field(default_factory=str)  # The embed description to DM the invoker
+    send_alert: bool = field(default=False)  # Whether to send an alert for the moderators
+    alert_content: str = field(default_factory=str)  # The content of the alert
```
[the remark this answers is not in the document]
mbaruh: ddcf2c6

bot/exts/filtering/_filter_lists/antispam.py
```diff
@@ -0,0 +1,183 @@
...
+        super().__init__(filtering_cog)
+        self.message_deletion_queue: dict[Member, DeletionContext] = dict()
+
+    def get_filter_type(self, content: str) -> type[UniqueFilter] | None:
+        """Get a subclass of filter matching the filter list and the filter's content."""
+        try:
+            return antispam_filter_types[content]
+        except KeyError:
+            if content not in self._already_warned:
+                log.warn(f"An antispam filter named {content} was supplied, but no matching implementation found.")
```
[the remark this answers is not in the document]
mbaruh: I always mix them up 4d51cad

bot/exts/filtering/_filter_lists/filter_list.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def _create_filter(self, filter_data: dict, defaults: Defaults) -> T | None:
+        """Create a filter from the given data."""
+        try:
+            content = filter_data["content"]
+            filter_type = self.get_filter_type(content)
+            if filter_type:
+                return filter_type(filter_data, defaults)
+            elif content not in self._already_warned:
+                log.warn(f"A filter named {content} was supplied, but no matching implementation found.")
```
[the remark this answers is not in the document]
mbaruh: 4d51cad

bot/utils/message_cache.py
```diff
@@ -89,16 +95,21 @@ def clear(self) -> None:
...
         self._end = 0
 
-    def get_message(self, message_id: int) -> t.Optional[Message]:
+    def get_message(self, message_id: int) -> Message | None:
         """Return the message that has the given message ID, if it is cached."""
         index = self._message_id_mapping.get(message_id, None)
         return self._messages[index] if index is not None else None
 
-    def update(self, message: Message) -> bool:
+    def get_message_metadata(self, message_id: int) -> dict:
```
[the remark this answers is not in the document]
mbaruh: 9b764e5

bot/exts/info/codeblock/_cog.py
```diff
@@ -135,8 +133,6 @@ def should_parse(self, message: discord.Message) -> bool:
             not message.author.bot
             and self.is_valid_channel(message.channel)
             and has_lines(message.content, constants.CodeBlock.minimum_lines)
-            and not TokenRemover.find_token_in_message(message)
-            and not WEBHOOK_URL_RE.search(message.content)
```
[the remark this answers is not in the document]
mbaruh: I knew there were usages I forgot to bring back 😅 bd5c0f8

bot/exts/filtering/_filter_context.py
decorator-factory: What about these
```py
    content: str | Iterable  # What actually needs filtering
    action_descriptions: list = field(default_factory=list)  # What actions were taken
```

mbaruh: I somehow missed `action_descriptions`, but for `content` the type of Iterable depends on the filter list. It might be more useful here to expand the comment.

mbaruh: 659320e

bot/exts/filtering/_filter_lists/antispam.py
```diff
@@ -0,0 +1,191 @@
...
+        if not webhook:
+            return
+
+        ctx, *other_contexts = self.contexts
+        new_ctx = FilterContext(ctx.event, ctx.author, ctx.channel, ctx.content, ctx.message)
+        new_ctx.action_descriptions = reduce(
+            add, (other_ctx.action_descriptions for other_ctx in other_contexts), ctx.action_descriptions
+        )
+        # It shouldn't ever come to this, but just in case.
+        if descriptions_num := len(new_ctx.action_descriptions) > 20:
```
wookie184, line 165:
```suggestion
        if (descriptions_num := len(new_ctx.action_descriptions)) > 20:
```

```
In [1]: if x := 5 > 3:
   ...:     print(x)
   ...:
True
```

bot/exts/filtering/_filter_lists/antispam.py
```diff
@@ -0,0 +1,191 @@
...
+
+        webhook = antispam_list.filtering_cog.webhook
+        if not webhook:
+            return
+
+        ctx, *other_contexts = self.contexts
+        new_ctx = FilterContext(ctx.event, ctx.author, ctx.channel, ctx.content, ctx.message)
+        new_ctx.action_descriptions = reduce(
+            add, (other_ctx.action_descriptions for other_ctx in other_contexts), ctx.action_descriptions
+        )
```
wookie184, line 163: When spamming a bunch I noticed I can get a response like this, would it make sense to remove duplicates? (ignore location of comment, not sure where this would be done).

image

bot/exts/filtering/filtering.py
```diff
@@ -0,0 +1,1424 @@
...
+                        description = self.loaded_filter_settings[filter_name][filter_setting_name][0]
+            if description is None:
+                await ctx.send(f":x: There's no setting type named {setting_name!r}.")
+                return
+            embed = Embed(colour=Colour.blue(), description=description)
+            embed.set_author(name=f"Description of the {setting_name} setting")
+            await ctx.send(embed=embed)
+
+    @filter.command(name="match")
+    async def f_match(
```
wookie184, line 591: This command doesn't seem to detect discord invites (that aren't whitelisted or blacklisted, but still trigger a notification).

bot/exts/filtering/filtering.py
```diff
@@ -0,0 +1,1424 @@
...
+        embed.description = f"`{filter_.content}`"
+        if description:
+            embed.description += f" - {description}"
+        embed.set_author(
+            name=f"Filter {filter_id} - {filter_list[list_type].label}".title())
+        embed.set_footer(text=(
+            "Field names with an asterisk have values which override the defaults of the containing filter list. "
+            f"To view all defaults of the list, "
+            f"run `{constants.Bot.prefix}filterlist describe {list_type.name} {filter_list.name}`."
+        ))
```
wookie184, line 523: I feel like it would be nicer to *only* display the overrides in the UI, seeing all the default values feels mostly unnecessary and a bit overwhelming.

An alternative is if it could also be more clever in only showing relevant values, so no empty lists/null values (unless they're overrides). E.g. the `enabled_*` and `disabled_*` lists could be hidden if they're empty, and `infraction_reason`, `infraction_duration`, `infraction_channel` could also be hidden if they're null.

bot/exts/filtering/_settings_types/actions/infraction_and_notification.py
```diff
@@ -0,0 +1,203 @@
...
+        ),
+        "dm_content": "The contents of a message to be DMed to the offending user.",
+        "dm_embed": "The contents of the embed to be DMed to the offending user."
+    }
+
+    dm_content: str
+    dm_embed: str
+    infraction_type: Infraction
+    infraction_reason: str
+    infraction_duration: float
```
wookie184, line 108: Passing the infraction duration is quite confusing, it would be nice if it is taken/displayed using the normal infraction duration format e.g. `1d`, currently you just get a message saying "Could not process the input value for infraction_duration.".

Another option would be if the "How long the infraction should last for in seconds. 0 for permanent." info could be put in the label of the modal.

wookie184 (commented):
Started playing around with it, this is *super* cool!!

bot/exts/filtering/_filter_lists/antispam.py
mbaruh: In this case, would something like `deleted all, muted, deleted X 9` be ok?

bot/exts/filtering/_filter_lists/antispam.py
mbaruh: lol

bot/exts/filtering/filtering.py
mbaruh: Yes, the goal is to find filters. In a previous iteration I just made it spit out the same messages that would appear in the mod alert, but this seemed more useful. Should it be reverted?

bot/exts/filtering/_filter_lists/antispam.py
wookie184: Only noticed because I was wondering why it was "±19" when I got this output lol
image

bot/exts/filtering/_filter_lists/antispam.py
wookie184: Yeah sounds good

bot/exts/filtering/filtering.py
mbaruh: I'm a bit conflicted about this. It feels like seeing the full context is important to avoid unexpected behavior. For example having a category disabled by default, and then not figuring out why this isn't working. Or the filter deleting messages by default, and not remembering to switch it off because the value is not in front of you.

The alternative is possible.. I'll have to play with it to see if it makes sense.

bot/exts/filtering/filtering.py
wookie184: Hmm, it probably does make more sense as-is actually, maybe the help text could be clarified though.

bot/exts/filtering/_settings_types/actions/infraction_and_notification.py
mbaruh: I wish I could put it in the modal, but the lengths of everything are pretty limiting. The title and any labels are up to 45 chars, placeholder strings are up to 100. I can make the description the default value but then people will need to delete it every time they want to input a value, which seems bad. I'll see if I can come up with something decent.

bot/exts/filtering/_filter_lists/antispam.py
mbaruh: [Fix condition](https://github.com/python-discord/bot/pull/2390/commits/0df028aebbb8fcfd8f6793bd84ddcca0ebf1f16f)

bot/exts/filtering/_filter_lists/antispam.py
mbaruh: [Group identical action descriptions in antispam](https://github.com/python-discord/bot/pull/2390/commits/7a1a553b67ff7fc300e9b457f8750efbf28ecf96)

bot/exts/filtering/filtering.py
mbaruh: [Correct filter match docstring](https://github.com/python-discord/bot/pull/2390/commits/a62993b44cfc0f2c3c5d746005b9a716dea82add)

bot/exts/filtering/_settings_types/actions/infraction_and_notification.py
mbaruh: This should do it
[Support custom value representation if filtering UI](https://github.com/python-discord/bot/pull/2390/commits/5850002237293be3ba0c2f9fd3ce707303a47679)

wookie184: Much better, thanks!

bot/exts/filtering/filtering.py
wookie184: Hmm, I don't think some settings being missing for the falsey example is too confusing, but lets leave this for now. We can always change things once we have more experience with it in use.

wookie184 (approved):
Played around with this some more and didn't notice anything urgent. Very good!!!

shtlrs (changes requested):
Looks & works great!
A couple of comments + small bugs were identified, will retest once addressed

bot/exts/filtering/_settings_types/actions/infraction_and_notification.py
```diff
@@ -0,0 +1,203 @@
...
+
+            try:
+                await ctx.author.send(dm_content, embed=dm_embed)
+                ctx.action_descriptions.append("notified")
+            except Forbidden:
+                ctx.action_descriptions.append("failed to notify")
+
+        if self.infraction_type != Infraction.NONE:
+            alerts_channel = bot_module.instance.get_channel(Channels.mod_alerts)
+            if self.infraction_channel:
```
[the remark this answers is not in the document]
mbaruh: `ctx.channel` will be used for a special value in `self.infraction_channel` (0), which is described in `!filter setting infraction_channel`.

bot/exts/filtering/_settings_types/actions/infraction_and_notification.py
```diff
@@ -0,0 +1,203 @@
...
+                other.infraction_duration is not None and self.infraction_duration > other.infraction_duration
+            ):
+                result = self.copy()
+            else:
+                result = other.copy()
+                other = self
+
+        # If the winner has no message but the loser does, copy the message to the winner.
+        result_overrides = result.overrides
+        if "dm_content" not in result_overrides and "dm_embed" not in result_overrides:
```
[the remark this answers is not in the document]
mbaruh: Yes

bot/exts/filtering/_settings.py
```diff
@@ -0,0 +1,233 @@
...
+
+    A filter doesn't have to have its own settings. For every undefined setting, it falls back to the value defined in
+    the filter list which contains the filter.
+    """
+
+    entry_type = T
+
+    _already_warned: set[str] = set()
+
+    @abstractmethod
```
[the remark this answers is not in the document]
mbaruh: Because the class doesn't count as abstract unless it has at least one abstract method, even if it inherits from ABC.

bot/exts/filtering/_filter_lists/extension.py
```diff
@@ -0,0 +1,122 @@
...
+
+PY_EMBED_DESCRIPTION = (
+    "It looks like you tried to attach a Python file - "
+    f"please use a code-pasting service such as {URLs.site_schema}{URLs.site_paste}"
+)
+
+TXT_LIKE_FILES = {".txt", ".csv", ".json"}
+TXT_EMBED_DESCRIPTION = (
+    "You either uploaded a `{blocked_extension}` file or entered a message that was too long. "
+    f"Please use our [paste bin]({URLs.site_schema}{URLs.site_paste}) instead."
```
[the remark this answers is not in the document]
mbaruh: [d596366](https://github.com/python-discord/bot/pull/2390/commits/d5963661ee6f720a1b5120010856185e03d5fce7)

bot/exts/filtering/_filter_lists/extension.py
```diff
@@ -0,0 +1,122 @@
...
+from bot.exts.filtering._filters.filter import Filter
+from bot.exts.filtering._settings import ActionSettings
+
+if typing.TYPE_CHECKING:
+    from bot.exts.filtering.filtering import Filtering
+
+
+PY_EMBED_DESCRIPTION = (
+    "It looks like you tried to attach a Python file - "
+    f"please use a code-pasting service such as {URLs.site_schema}{URLs.site_paste}"
```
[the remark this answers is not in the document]
mbaruh: [d596366](https://github.com/python-discord/bot/pull/2390/commits/d5963661ee6f720a1b5120010856185e03d5fce7)

bot/exts/filtering/FILTERS-DEVELOPMENT.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Overview
+The main idea is that there is a list of filters each deciding whether they apply to the given content.
+For example, there can be a filter that decides it will trigger when the content contains the string "lemon".
+
+There are several types of filters, and two filters of the same type differ by their content.
+For example, filters of type "token" search for a specific token inside the provided string.
+One token filter might look for the string "lemon", while another will look for the string "joe".
+
+Each filter has a set of settings that decide when it triggers (e.g. in which channels), and what happens if it does (e.g. delete the message).
```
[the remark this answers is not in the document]
mbaruh: [Code style and clarifications](https://github.com/python-discord/bot/pull/2390/commits/990d716ea2e3532e305460d1aef88f9cba211614)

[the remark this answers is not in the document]
mbaruh: [Code style and clarifications](https://github.com/python-discord/bot/pull/2390/commits/990d716ea2e3532e305460d1aef88f9cba211614)

bot/exts/filtering/_settings.py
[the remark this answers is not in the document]
mbaruh: I added a clarifying comment in [Code style and clarifications](https://github.com/python-discord/bot/pull/2390/commits/990d716ea2e3532e305460d1aef88f9cba211614)

bot/exts/filtering/_filter_lists/antispam.py
```diff
@@ -0,0 +1,197 @@
...
+        # Additional messages are necessarily part of the deletion.
+        ctx.upload_deletion_logs = False
+        self.message_deletion_queue[ctx.author].add(ctx, triggers)
+
+        current_actions = sublist.merge_actions(triggers)
+        # Don't alert yet.
+        current_actions.pop("ping", None)
+        current_actions.pop("send_alert", None)
+
+        new_infraction = current_actions["infraction_and_notification"].copy()
```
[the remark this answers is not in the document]
mbaruh: [Code style and clarifications](https://github.com/python-discord/bot/pull/2390/commits/990d716ea2e3532e305460d1aef88f9cba211614)

bot/exts/filtering/_filter_lists/antispam.py
```diff
@@ -0,0 +1,197 @@
...
+            )
+            current_actions["infraction_and_notification"] = new_infraction
+            self.message_deletion_queue[ctx.author].current_infraction = new_infraction.infraction_type
+        else:
+            current_actions.pop("infraction_and_notification", None)
+
+        # Provide some message in case another filter list wants there to be an alert.
+        return current_actions, ["Handling spam event..."], {ListType.DENY: triggers}
+
+    def _create_deletion_context_handler(self, context_id: Member) -> Callable[[FilterContext], Coroutine]:
```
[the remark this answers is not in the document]
mbaruh: [Code style and clarifications](https://github.com/python-discord/bot/pull/2390/commits/990d716ea2e3532e305460d1aef88f9cba211614)

bot/exts/filtering/_filter_lists/extension.py
```diff
@@ -0,0 +1,122 @@
...
+        if ctx.event != Event.SNEKBOX:  # Don't post the embed if it's a snekbox response.
+            if ".py" in not_allowed:
+                # Provide a pastebin link for .py files.
+                ctx.dm_embed = PY_EMBED_DESCRIPTION
+            elif txt_extensions := {ext for ext in TXT_LIKE_FILES if ext in not_allowed}:
+                # Work around Discord auto-conversion of messages longer than 2000 chars to .txt
+                cmd_channel = bot.instance.get_channel(Channels.bot_commands)
+                ctx.dm_embed = TXT_EMBED_DESCRIPTION.format(
+                    blocked_extension=txt_extensions.pop(),
+                    cmd_channel_mention=cmd_channel.mention
```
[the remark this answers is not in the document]
mbaruh: [Code style and clarifications](https://github.com/python-discord/bot/pull/2390/commits/990d716ea2e3532e305460d1aef88f9cba211614)

bot/exts/filtering/_filter_lists/extension.py
```diff
@@ -0,0 +1,122 @@
...
+)
+
+TXT_LIKE_FILES = {".txt", ".csv", ".json"}
+TXT_EMBED_DESCRIPTION = (
+    "You either uploaded a `{blocked_extension}` file or entered a message that was too long. "
+    f"Please use our [paste bin]({URLs.site_schema}{URLs.site_paste}) instead."
+)
+
+DISALLOWED_EMBED_DESCRIPTION = (
+    "It looks like you tried to attach file type(s) that we do not allow ({blocked_extensions_str}). "
```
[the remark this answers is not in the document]
mbaruh: [Code style and clarifications](https://github.com/python-discord/bot/pull/2390/commits/990d716ea2e3532e305460d1aef88f9cba211614)

bot/exts/filtering/_settings_types/validations/channel_scope.py
```diff
@@ -0,0 +1,70 @@
...
+
+from bot.exts.filtering._filter_context import FilterContext
+from bot.exts.filtering._settings_types.settings_entry import ValidationEntry
+
+
+class ChannelScope(ValidationEntry):
+    """A setting entry which tells whether the filter was invoked in a whitelisted channel or category."""
+
+    name: ClassVar[str] = "channel_scope"
+    description: ClassVar[str] = {
```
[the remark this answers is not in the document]
mbaruh: [Code style and clarifications](https://github.com/python-discord/bot/pull/2390/commits/990d716ea2e3532e305460d1aef88f9cba211614)

bot/exts/filtering/_settings_types/actions/infraction_and_notification.py
```diff
@@ -0,0 +1,203 @@
...
+            if dm_embed:
+                dm_embed = Embed(description=dm_embed.format(**formatting), colour=Colour.og_blurple())
+            else:
+                dm_embed = None
+
+            try:
+                await ctx.author.send(dm_content, embed=dm_embed)
+                ctx.action_descriptions.append("notified")
+            except Forbidden:
+                ctx.action_descriptions.append("failed to notify")
```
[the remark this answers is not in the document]
mbaruh: I think I'll leave it for now. 99% of the time it will be the user having DMs off.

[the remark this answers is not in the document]
shtlrs: Okay, sounds good to me.

shtlrs (approved):
The moderation lead strikes again, extremely beautiful work.

Like wookie said, there isn't urgent/alarming that I think needs to be addressed here.
I have played around with it and tried to test all the features that we rely on, and all seems to work fine after the small bugs we've found.

Thank you !

PR #2393, opened 2023-01-31
Added dummy training_example_logger into bot.py

MikeDX:
Please do not merge this as it is just an example.


MikeDX:
Code did not meet standard, rejecting.

PR #2400, opened 2023-02-13
Reusable status embed workflow

shtlrs:
The purpose of this PR is to make the `status_embed` workflow reusable.

**Problem**
We currently have the same `status_embed` workflow file scattered & duplicated across our `bot`, `site`, `sir-robin`, `pixels`, `king-arthur` and `sir-lancebot` repositories.
A solution is to make the workflow reusable, and port it over as part of the organization's workflows, from where it can be accessed & invoked by all of our repos.

**Note**
- This is currently just a test, and the workflow file will later be moved to the organization's level.
- The discord webhook token would also need to be ported at the organization's level, for it to be accessed from all workflows


.github/workflows/lint-test.yml
```diff
@@ -12,6 +12,9 @@ concurrency:
 
 jobs:
   lint-test:
+    continue-on-error: true
```
HassanAbouelela, line 15: I don't know we should necessarily do this. A lot of our actions have a very linear flow of steps that depend on each other:
Install dependencies -> setup requirements and environment -> do something

In theory it doesn't matter if one step fails and causes the next to fail instead of be skipped, but it doesn't seem appropriate for some of the more critical actions such as deployments and uploading images.

I believe you can force a certain step to always run without forcing the entire workflow to continue though, [example](https://github.com/python-discord/bot/blob/main/.github/workflows/lint-test.yml#L84).

shtlrs: The problem is that we don't know which step would fail, and since the second job needs the first one, it should always run otherwise the second might not run? (Not sure of this tbh, I'll test it)

The `continue-on-error` cannot be part of the job that calls the reusable workflow as well, that's why I put it there

shtlrs: I just tested it, the `call_status_embed` doesn't trigger without having `continue-on-error` set to true.
It also won't when i move that key to the reusable workflow's file

HassanAbouelela: Could you clarify, I don't really follow. Where would the on-error go, and what is the second step. Why does it matter knowing which fails, don't we just care that we run this step if something fails?

HassanAbouelela (approved):
Great PR from the mind of Devops' own @shtlrs. LGTM

.github/workflows/lint-test.yml
shtlrs: This has been discussed internally, but for people who might bump into this, here's what has been discussed:

We had the `continue-on-error` option set because we always need the first job to run entirely in order for the second job to trigger and that's because the second depends on the first.

However, it would be a waste to run future steps knowing they would fail if the previous ones did.

The solution for this was to add an `if` statement that will always make the second one run.

PR #2402, opened 2023-02-14
reply to user when using !eval/!timeit command instead of pinging

### LINKED ISSUE #2382: Reply to !eval message instead of pinging
It was discussed in [#community-meta](https://discord.com/channels/267624335836053506/429409067623251969/1064000890798870558) if the `!eval` command should reply to the original command instead of sending a separate message with a ping. Thoughts on this? I'd be willing to PR if accepted.

e1pupper:
Closes #2382 

Just needed to change ctx.send to ctx.reply and remove the author.mention.

### DIFF
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -472,7 +472,7 @@ class Snekbox(Cog):
                 response = await self.send_job(ctx, python_version, code, args=args, job_name=job_name)
             except LockedResourceError:
                 await ctx.reply(
-                    f"You've already got a job running - "
+                    "You've already got a job running - "
                     "please wait for it to finish!"
                 )
                 return
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -471,8 +471,8 @@ class Snekbox(Cog):
             try:
                 response = await self.send_job(ctx, python_version, code, args=args, job_name=job_name)
             except LockedResourceError:
-                await ctx.send(
-                    f"{ctx.author.mention} You've already got a job running - "
+                await ctx.reply(
+                    f"You've already got a job running - "
                     "please wait for it to finish!"
                 )
                 return
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -345,7 +345,7 @@ class Snekbox(Cog):
                 output, paste_link = await self.format_output(results["stdout"])
 
             icon = self.get_status_emoji(results)
-            msg = f"{ctx.author.mention} {icon} {msg}.\n\n```\n{output}\n```"
+            msg = f"{icon} {msg}.\n\n```\n{output}\n```"
             if paste_link:
                 msg = f"{msg}\nFull output: {paste_link}"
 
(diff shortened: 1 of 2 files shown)


n0Oo0Oo0b:
> bot/issues/2382
> 
> Just needed to change ctx.send to ctx.reply and remove the author.mention.

What if the response gets redirected to a different channel? `@redirect_output()` doesn't change the behavior of `ctx.reply()` so the result won't be redirected when it should.
It may be worth considering having `@redirect_output()` add an extra bool attribute (maybe `redirected`?) for whether or not the output was redirected or maybe even replacing the `ctx.reply()` function to send a message to the redirected channel instead (not sure if replacing `ctx.reply()` would be a good idea though).

e1pupper:
> > bot/issues/2382
> > Just needed to change ctx.send to ctx.reply and remove the author.mention.
> 
> What if the response gets redirected to a different channel? `@redirect_output()` doesn't change the behavior of `ctx.reply()` so the result won't be redirected when it should. It may be worth considering having `@redirect_output()` add an extra bool attribute (maybe `redirected`?) for whether or not the output was redirected or maybe even replacing the `ctx.reply()` function to send a message to the redirected channel instead (not sure if replacing `ctx.reply()` would be a good idea though).

I just looked into the `@redirect_output()` i think it's also sensibile to change those to `.reply()` as well

```py
            if ping_user:
                await ctx.send(f"Here's the output of your command, {ctx.author.mention}")
            scheduling.create_task(func(self, ctx, *args, **kwargs))

            message = await old_channel.send(
                f"Hey, {ctx.author.mention}, you can find the output of your command here: "
                f"{redirect_channel.mention}"
            )
```

wookie184 (changes requested):
We need to handle the case where the message is deleted before the bot sends the response. Currently an error is raised:
image
To reproduce, run `!e import time;time.sleep(5)` and immediately delete your message.

There are three main options I can think of for handling this case:
1. Catch the error and send no response
1. Catch the error and resend the message with a ping instead of a reply
1. Make it so the error isn't raised and the message is just sent without being a reply (and without a ping). (I can't remember exactly how you do this, something to do with https://discordpy.readthedocs.io/en/stable/api.html?highlight=reply#discord.Message.to_reference with `fail_if_not_exists=False`).

I don't have much of a preference given this is an edge case.

ChrisLovering:
> I just looked into the @redirect_output() i think it's also sensibile to change those to .reply() as well

You can only reply to messages in the same channel, so if this message gets redirected to another channel (such as bot-commands in this case) the message in bot-commands you can no longer reply to the original message.

e1pupper:
> We need to handle the case where the message is deleted before the bot sends the response. Currently an error is raised: image To reproduce, run `!e import time;time.sleep(5)` and immediately delete your message.
> 
> There are three main options I can think of for handling this case:
> 
> 1. Catch the error and send no response
> 2. Catch the error and resend the message with a ping instead of a reply
> 3. Make it so the error isn't raised and the message is just sent without being a reply (and without a ping). (I can't remember exactly how you do this, something to do with https://discordpy.readthedocs.io/en/stable/api.html?highlight=reply#discord.Message.to_reference with `fail_if_not_exists=False`).
> 
> I don't have much of a preference given this is an edge case.

i see im gonna see if i can work on this in the weekend sorry for the late response

Robin5605:
So, what's the planned behavior for redirected output? Since those can't reply to the original message.
Perhaps a link button in the original `!eval` that links to the message in `#bot-commands`, and a link button in the result message sent in `#bot-commands`, that links back to the original `!eval` command?

Robin5605:
Oh, and also, @e1pupper had noticed us that they will not be able to work on this. I'd be willing to take this up.

wookie184:
> I'd be willing to take this up.

Sounds good, i've assigned you

wookie184:
I'll take this over. If the response is redirected I'll just keep the old behaviour of mentioning the user.

wookie184:
Since the conflicts are due to the entire snekbox.py file being split up, as this PR is fairly small, I'm just going to start over on a new branch.

Issue #2394, opened 2023-02-01
Migrating !tag to slash commands


Canttuchdiz:
Migrating **!tag** to a slash command. Allowing for autocomplete, and a subcommand **/tag help** or **/tags** listing all tags. That list could also be ephemeral, so the user can see all the tags. Also, if possible, make the main command **/tag**, and the param would be autocomplete.

JacobCoffee:
I believe this is a duplicate of #2362

ChrisLovering:
> I believe this is a duplicate of #2362

I suggested splitting out the issue in this way, see my recent comment in the issue for context.

> and a subcommand /tag help or /tags listing all tags. That list could also be ephemeral, so the user can see all the tags.

We could make the tag name argument optional, so if not given it lists all the tags in a paginated embed, like it does currently.

> Also, if possible, make the main command /tag, and the param would be autocomplete.

I agree with this, we should use fuzzy matching for auto complete logic, similar to the current tag suggestion autocomplete, so that users can use the same short-hand names for tags as they currently do.

Ibrahim2750mi:
can I be assigned this?

Ibrahim2750mi:
Also the reactions on the embed needs to be changed to buttons right?
![2023-02-14_03-03](https://user-images.githubusercontent.com/74553450/218579591-c7f4fd95-91f0-409c-ae27-6879914fda4d.png)

wookie184:
I'm not sure if the intention of this issue is to replace the direct invocation of tags (e.g. through `!code`), or just the actual `!tag` command.

I think replacing the actual `!tag` command makes sense, but not removing the direct invocation. It does not seem like an improvement to go from `!code` to `/tag get code`, pressing enter, then having to press enter again after it autocompletes to `/tag get codeblock`, all with a bunch of discord UI popping up in the process. Having `/tag` with autocomplete would be useful for finding a tag if you're not sure if exists, but I think we should also keep the faster shorthand.

Ibrahim2750mi:
I completely agree with what wookie said, I asked in the discord server but was advised it is a full migration.

Also @wookie184 your review comments makes sense now, sorry for misunderstanding.

Robin5605:
I'm also in favor of what wookie said. Seems like a downgrade to remove to !tag functionality. Ideally the slash command would increase visibility

mbaruh:
Yeah let's keep the direct usage for now.

PR #2403, opened 2023-02-14
Migrated `!tags` command to slash command `/tag`

Ibrahim2750mi:
#2394

+ `/tag list` lists all the accessible tags
+ `/tag get <tag_name>` empty `<tag_name>` returns the same output as `/tag list`
+ removed the `search` sub-group

Screenshots:

### DIFF
--- a/bot/exts/backend/error_handler.py
+++ b/bot/exts/backend/error_handler.py
@@ -27,7 +27,8 @@ class ErrorHandler(Cog):
         """
         Add checks for the `get_command_ctx` function here.
 
-        Use discord.utils to run the checks.
+        The command code style is copied from discord.ext.commands.Command.can_run itself.
+        Append checks in the checks list.
         """
         checks = []
         predicates = checks
--- a/bot/exts/backend/error_handler.py
+++ b/bot/exts/backend/error_handler.py
@@ -1,7 +1,8 @@
 import copy
 import difflib
+import typing as t
 
-from discord import Embed
+from discord import Embed, Interaction
 from discord.ext.commands import ChannelNotFound, Cog, Context, TextChannelConverter, VoiceChannelConverter, errors
 from pydis_core.site_api import ResponseCodeError
 from sentry_sdk import push_scope
@@ -20,6 +21,10 @@ class ErrorHandler(Cog):
 
     def __init__(self, bot: Bot):
         self.bot = bot
+
+    @staticmethod
+    async def _can_run(_: Interaction) -> bool:
+        return False
 
     def _get_error_embed(self, title: str, body: str) -> Embed:
         """Return an embed that contains the exception."""
@@ -159,7 +164,7 @@ class ErrorHandler(Cog):
             return True
         return False
 
-    async def try_get_tag(self, ctx: Context) -> None:
+    async def try_get_tag(self, interaction: Interaction, can_run: t.Callable[[Interaction], bool] = False) -> None:
         """
         Attempt to display a tag by interpreting the command name as a tag name.
 
@@ -168,27 +173,28 @@ class ErrorHandler(Cog):
         the context to prevent infinite recursion in the case of a CommandNotFound exception.
         """
         tags_get_command = self.bot.get_command("tags get")
+        tags_get_command.can_run = can_run if can_run else self._can_run
         if not tags_get_command:
             log.debug("Not attempting to parse message as a tag as could not find `tags get` command.")
             return
 
-        ctx.invoked_from_error_handler = True
+        interaction.invoked_from_error_handler = True
 
         log_msg = "Cancelling attempt to fall back to a tag due to failed checks."
         try:
-            if not await tags_get_command.can_run(ctx):
+            if not await tags_get_command.can_run(interaction):
                 log.debug(log_msg)
                 return
         except errors.CommandError as tag_error:
             log.debug(log_msg)
-            await self.on_command_error(ctx, tag_error)
+            await self.on_command_error(interaction, tag_error)
             return
 
-        if await ctx.invoke(tags_get_command, argument_string=ctx.message.content):
+        if await interaction.invoke(tags_get_command, tag_name=interaction.message.content):
             return
 
-        if not any(role.id in MODERATION_ROLES for role in ctx.author.roles):
-            await self.send_command_suggestion(ctx, ctx.invoked_with)
+        if not any(role.id in MODERATION_ROLES for role in interaction.user.roles):
+            await self.send_command_suggestion(interaction, interaction.invoked_with)
 
     async def try_run_eval(self, ctx: Context) -> bool:
         """
--- a/bot/exts/backend/error_handler.py
+++ b/bot/exts/backend/error_handler.py
@@ -2,7 +2,7 @@ import copy
 import difflib
 import typing as t
 
-from discord import Embed, Interaction
+from discord import Embed, Interaction, utils
 from discord.ext.commands import ChannelNotFound, Cog, Context, TextChannelConverter, VoiceChannelConverter, errors
 from pydis_core.site_api import ResponseCodeError
 from sentry_sdk import push_scope
@@ -23,8 +23,19 @@ class ErrorHandler(Cog):
         self.bot = bot
 
     @staticmethod
-    async def _can_run(_: Interaction) -> bool:
-        return False
+    async def _can_run(ctx: Context) -> bool:
+        """
+        Add checks for the `get_command_ctx` function here.
+
+        Use discord.utils to run the checks.
+        """
+        checks = []
+        predicates = checks
+        if not predicates:
+            # Since we have no checks, then we just return True.
+            return True
+
+        return await utils.async_all(predicate(ctx) for predicate in predicates)
 
     def _get_error_embed(self, title: str, body: str) -> Embed:
         """Return an embed that contains the exception."""
@@ -164,7 +175,7 @@ class ErrorHandler(Cog):
             return True
         return False
 
-    async def try_get_tag(self, interaction: Interaction, can_run: t.Callable[[Interaction], bool] = False) -> None:
+    async def try_get_tag(self, ctx: Context, can_run: t.Callable[[Interaction], bool] = False) -> None:
         """
         Attempt to display a tag by interpreting the command name as a tag name.
 
@@ -172,29 +183,30 @@ class ErrorHandler(Cog):
         by `on_command_error`, but the `invoked_from_error_handler` attribute will be added to
         the context to prevent infinite recursion in the case of a CommandNotFound exception.
         """
-        tags_get_command = self.bot.get_command("tags get")
-        tags_get_command.can_run = can_run if can_run else self._can_run
-        if not tags_get_command:
-            log.debug("Not attempting to parse message as a tag as could not find `tags get` command.")
+        tags_cog = self.bot.get_cog("Tags")
+        if not tags_cog:
+            log.debug("Not attempting to parse message as a tag as could not find `Tags` cog.")
             return
+        tags_get_command = tags_cog.get_command_ctx
+        can_run = can_run if can_run else self._can_run
 
-        interaction.invoked_from_error_handler = True
+        ctx.invoked_from_error_handler = True
 
         log_msg = "Cancelling attempt to fall back to a tag due to failed checks."
         try:
-            if not await tags_get_command.can_run(interaction):
+            if not await can_run(ctx):
                 log.debug(log_msg)
                 return
         except errors.CommandError as tag_error:
             log.debug(log_msg)
-            await self.on_command_error(interaction, tag_error)
+            await self.on_command_error(ctx, tag_error)
             return
 
-        if await interaction.invoke(tags_get_command, tag_name=interaction.message.content):
+        if await tags_get_command(ctx, ctx.message.content):
             return
 
-        if not any(role.id in MODERATION_ROLES for role in interaction.user.roles):
-            await self.send_command_suggestion(interaction, interaction.invoked_with)
+        if not any(role.id in MODERATION_ROLES for role in ctx.author.roles):
+            await self.send_command_suggestion(ctx, ctx.invoked_with)
 
     async def try_run_eval(self, ctx: Context) -> bool:
         """
--- a/bot/exts/backend/error_handler.py
+++ b/bot/exts/backend/error_handler.py
@@ -1,8 +1,7 @@
 import copy
 import difflib
-import typing as t
 
-from discord import Embed, Interaction, utils
+from discord import Embed, Member
 from discord.ext.commands import ChannelNotFound, Cog, Context, TextChannelConverter, VoiceChannelConverter, errors
 from pydis_core.site_api import ResponseCodeError
 from sentry_sdk import push_scope
@@ -21,22 +20,6 @@ class ErrorHandler(Cog):
 
     def __init__(self, bot: Bot):
         self.bot = bot
-
-    @staticmethod
-    async def _can_run(ctx: Context) -> bool:
-        """
-        Add checks for the `get_command_ctx` function here.
-
-        The command code style is copied from discord.ext.commands.Command.can_run itself.
-        Append checks in the checks list.
-        """
-        checks = []
-        predicates = checks
-        if not predicates:
-            # Since we have no checks, then we just return True.
-            return True
-
-        return await utils.async_all(predicate(ctx) for predicate in predicates)
 
     def _get_error_embed(self, title: str, body: str) -> Embed:
         """Return an embed that contains the exception."""
@@ -176,7 +159,7 @@ class ErrorHandler(Cog):
             return True
         return False
 
-    async def try_get_tag(self, ctx: Context, can_run: t.Callable[[Interaction], bool] = False) -> None:
+    async def try_get_tag(self, ctx: Context) -> None:
         """
         Attempt to display a tag by interpreting the command name as a tag name.
 
@@ -189,25 +172,28 @@ class ErrorHandler(Cog):
             log.debug("Not attempting to parse message as a tag as could not find `Tags` cog.")
             return
         tags_get_command = tags_cog.get_command_ctx
-        can_run = can_run if can_run else self._can_run
+
+        maybe_tag_name = ctx.invoked_with
+        if not maybe_tag_name or not isinstance(ctx.author, Member):
+            return
 
         ctx.invoked_from_error_handler = True
-
-        log_msg = "Cancelling attempt to fall back to a tag due to failed checks."
         try:
-            if not await can_run(ctx):
-                log.debug(log_msg)
+            if not await self.bot.can_run(ctx):
+                log.debug("Cancelling attempt to fall back to a tag due to failed checks.")
                 return
-        except errors.CommandError as tag_error:
-            log.debug(log_msg)
-            await self.on_command_error(ctx, tag_error)
-            return
 
-        if await tags_get_command(ctx, ctx.message.content):
-            return
+            if await tags_get_command(ctx, maybe_tag_name):
+                return
 
-        if not any(role.id in MODERATION_ROLES for role in ctx.author.roles):
-            await self.send_command_suggestion(ctx, ctx.invoked_with)
+            if not any(role.id in MODERATION_ROLES for role in ctx.author.roles):
+                await self.send_command_suggestion(ctx, maybe_tag_name)
+        except Exception as err:
+            log.debug("Error while attempting to invoke tag fallback.")
+            if isinstance(err, errors.CommandError):
+                await self.on_command_error(ctx, err)
+            else:
+                await self.on_command_error(ctx, errors.CommandInvokeError(err))
 
     async def try_run_eval(self, ctx: Context) -> bool:
         """
--- a/bot/exts/info/tags.py
+++ b/bot/exts/info/tags.py
@@ -8,7 +8,7 @@ from typing import Literal, NamedTuple, Optional, Union
 
 import discord
 import frontmatter
-from discord import Embed, Interaction, Member, User, app_commands
+from discord import Embed, Interaction, Member, app_commands
 from discord.ext.commands import Cog, Context
 
 from bot import constants
@@ -91,7 +91,7 @@ class Tag:
         embed.description = self.content
         return embed
 
-    def accessible_by(self, member: Member | User) -> bool:
+    def accessible_by(self, member: Member) -> bool:
         """Check whether `member` can access the tag."""
         return bool(
             not self._restricted_to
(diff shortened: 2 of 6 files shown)


bot/exts/info/tags.py
```diff
@@ -138,6 +140,15 @@ def __init__(self, bot: Bot):
...
         self.initialize_tags()
+        self.bot.tree.copy_global_to(guild=discord.Object(id=GUILD_ID))
+
+    tag_group = app_commands.Group(name="tag", description="...")
+    # search_tag = app_commands.Group(name="search", description="...", parent=tag_group)
+
+    @Cog.listener()
+    async def on_ready(self) -> None:
+        """Called when the cog is ready."""
+        await self.bot.tree.sync(guild=discord.Object(id=GUILD_ID))
```
wookie184, line 151: I don't think any of the syncing code should be necessary, since we do that automatically now (https://github.com/python-discord/bot-core/pull/171)

bot/exts/info/tags.py
```diff
@@ -362,8 +292,8 @@ def accessible_tags_in_group(self, group: str, user: discord.Member) -> list[str
             if identifier.group == group and tag.accessible_by(user)
         )
 
-    @tags_group.command(name="get", aliases=("show", "g"), usage="[tag_group] [tag_name]")
-    async def get_command(self, ctx: Context, *, argument_string: Optional[str]) -> bool:
+    @tag_group.command(name="get")
+    async def get_command(self, interaction: discord.Interaction, *, tag_name: Optional[str]) -> bool:
```
wookie184, line 296:
```suggestion
    async def get_command(self, interaction: discord.Interaction, *, name: Optional[str]) -> bool:
```

Looks a bit neater in the command, since it's obvious it's the tag name so no need to repeat tag. (will require updates in the rest of the function ofc)

bot/exts/info/tags.py
```diff
@@ -138,6 +140,15 @@ def __init__(self, bot: Bot):
         self.bot = bot
         self.tags: dict[TagIdentifier, Tag] = {}
         self.initialize_tags()
+        self.bot.tree.copy_global_to(guild=discord.Object(id=GUILD_ID))
+
+    tag_group = app_commands.Group(name="tag", description="...")
+    # search_tag = app_commands.Group(name="search", description="...", parent=tag_group)
```
wookie184, line 146:
```suggestion
```
Can this be removed?

tests/helpers.py
```diff
@@ -479,6 +479,26 @@ def __init__(self, **kwargs) -> None:
...
 
 
+class MockInteraction(CustomMockMixin, unittest.mock.MagicMock):
+    """
+    A MagicMock subclass to mock Interaction objects.
+
+    Instances of this class will follow the specifications of `discord.Interaction`
+    instances. For more information, see the `MockGuild` docstring.
+    """
+    # spec_set = context_instance
```
wookie184, line 489: Could try and make an instance for `Interaction` like was done with context, although if the tests work as is it's probably fine to just remove this. I can't remember exactly what it does :P

wookie184 (changes requested):
I'm currently getting an error when running this if there are too many matches, we should probably trim the results in these cases:
```
bot-bot-1          | 2023-02-19 15:45:18 | asyncio | ERROR | Task exception was never retrieved
bot-bot-1          | future: <Task finished name='CommandTree-invoker' coro=<CommandTree._from_interaction.<locals>.wrapper() done, defined at /opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/app_commands/tree.py:1089> exception=HTTPException('400 Bad Request (error code: 50035): Invalid Form Body\nIn data.choices: Must be 25 or fewer in length.')>
bot-bot-1          | Traceback (most recent call last):
bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/app_commands/tree.py", line 1091, in wrapper
bot-bot-1          |     await self._call(interaction)
bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/app_commands/tree.py", line 1238, in _call
bot-bot-1          |     await command._invoke_autocomplete(interaction, focused, namespace)
bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/app_commands/commands.py", line 931, in _invoke_autocomplete
bot-bot-1          |     await interaction.response.autocomplete(choices)
bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/interactions.py", line 966, in autocomplete
bot-bot-1          |     await adapter.create_interaction_response(
bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/webhook/async_.py", line 220, in request
bot-bot-1          |     raise HTTPException(response, data)
bot-bot-1          | discord.errors.HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body
bot-bot-1          | In data.choices: Must be 25 or fewer in length.
```

I think we should try to avoid any cases of "The application did not respond" errors on the client. Another case where this can happen is where no match can be found e.g. `/tag get fionferonfero`

tests/helpers.py
Ibrahim2750mi: The tests work fine, so I will just remove the comment

bot/exts/info/tags.py
```diff
@@ -373,38 +303,68 @@ async def get_command(self, ctx: Context, *, argument_string: Optional[str]) ->
...
+            )
+        else:
+            await interaction.response.send_message(embed=Embed(description="**There are no tags!**"))
+        return True
+
 
 async def setup(bot: Bot) -> None:
     """Load the Tags cog."""
     await bot.add_cog(Tags(bot))
+    await bot.tree.sync(guild=discord.Object(id=GUILD_ID))
```
Ibrahim2750mi, line 370: >I don't think any of the syncing code should be necessary, since we do that automatically now (https://github.com/python-discord/bot-core/pull/171)

@wookie184 should this also be removed?

bot/exts/info/tags.py
```diff
@@ -373,38 +303,68 @@ async def get_command(self, ctx: Context, *, argument_string: Optional[str]) ->
         Returns True if a message was sent, or if the tag is on cooldown.
         Returns False if no message was sent.
```
Ibrahim2750mi, line 304: >I think we should try to avoid any cases of "The application did not respond" errors on the client. Another case where this can happen is where no match can be found e.g. /tag get fionferonfero

@wookie184 Should it always return `True` now as an embed has to be sent either ways or should it return `False` when no match for a tag is found?

bot/exts/info/tags.py
Ibrahim2750mi: Even after removing this, the bot works fine. So I guess its best to remove it.

Ibrahim2750mi:
> I'm currently getting an error when running this if there are too many matches, we should probably trim the results in these cases:
> 
> ```
> bot-bot-1          | 2023-02-19 15:45:18 | asyncio | ERROR | Task exception was never retrieved
> bot-bot-1          | future: <Task finished name='CommandTree-invoker' coro=<CommandTree._from_interaction.<locals>.wrapper() done, defined at /opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/app_commands/tree.py:1089> exception=HTTPException('400 Bad Request (error code: 50035): Invalid Form Body\nIn data.choices: Must be 25 or fewer in length.')>
> bot-bot-1          | Traceback (most recent call last):
> bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/app_commands/tree.py", line 1091, in wrapper
> bot-bot-1          |     await self._call(interaction)
> bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/app_commands/tree.py", line 1238, in _call
> bot-bot-1          |     await command._invoke_autocomplete(interaction, focused, namespace)
> bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/app_commands/commands.py", line 931, in _invoke_autocomplete
> bot-bot-1          |     await interaction.response.autocomplete(choices)
> bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/interactions.py", line 966, in autocomplete
> bot-bot-1          |     await adapter.create_interaction_response(
> bot-bot-1          |   File "/opt/poetry/cache/virtualenvs/bot-TFcQMFAJ-py3.10/lib/python3.10/site-packages/discord/webhook/async_.py", line 220, in request
> bot-bot-1          |     raise HTTPException(response, data)
> bot-bot-1          | discord.errors.HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body
> bot-bot-1          | In data.choices: Must be 25 or fewer in length.
> ```

Fixed.

bot/exts/info/tags.py
wookie184: I believe we should be able to remove this too, yeah, since there's nothing specific to the cog.

bot/exts/info/tags.py
wookie184: We need to be careful to ensure that we don't break normal tag invocation (with just the name of the tag, e.g. `!code`). This is handled in the error handler here https://github.com/python-discord/bot/blob/main/bot/exts/backend/error_handler.py#L162

For that it makes sense to give no result if a tag wasn't found, although if a user is using the `tag get` command explicitly I think they should always get a response (e.g. saying "no tag found").

I'm not sure what the best way to make this change is, it might require changes to the error handler (maybe making it use a helper function which is also used by the get command, but the get command can also handle no tags found).

bot/exts/backend/error_handler.py
```diff
@@ -159,7 +164,7 @@ async def try_silence(self, ctx: Context) -> bool:
             return True
         return False
 
-    async def try_get_tag(self, ctx: Context) -> None:
+    async def try_get_tag(self, interaction: Interaction, can_run: t.Callable[[Interaction], bool] = False) -> None:
```
shtlrs, line 167: I don't think a callable should default to False, should it ?

bot/exts/backend/error_handler.py
```diff
@@ -168,27 +173,28 @@ async def try_get_tag(self, ctx: Context) -> None:
...
             log.debug("Not attempting to parse message as a tag as could not find `tags get` command.")
             return
 
-        ctx.invoked_from_error_handler = True
+        interaction.invoked_from_error_handler = True
 
         log_msg = "Cancelling attempt to fall back to a tag due to failed checks."
         try:
-            if not await tags_get_command.can_run(ctx):
+            if not await tags_get_command.can_run(interaction):
```
shtlrs, line 185: I don't think the try/except should belong here, but in the can_run function instead.
It should be a simple "yes or no" question, and the `CommandError` should be handled inside and return False upon that

shtlrs: Just to be clear, I now this not related to your work, but it's only an improvement suggestion

bot/exts/info/tags.py
shtlrs: @Ibrahim2750mi I don't see this as resolved, the commented line is still there

bot/exts/info/tags.py
shtlrs: As a general advice, **nothing** should be done in the `on_ready` event listener, as it could get triggered multiple times, and syncing is an expensive task.

Instead, you should use the `cog_load` hook for things that need to happen once post instantiation of the cog.

2. This also hasn't been resolved it seems

bot/exts/info/tags.py
```diff
@@ -362,8 +292,8 @@ def accessible_tags_in_group(self, group: str, user: discord.Member) -> list[str
             if identifier.group == group and tag.accessible_by(user)
         )
 
-    @tags_group.command(name="get", aliases=("show", "g"), usage="[tag_group] [tag_name]")
-    async def get_command(self, ctx: Context, *, argument_string: Optional[str]) -> bool:
+    @tag_group.command(name="get")
```
shtlrs, line 295: Is there a reason for removing the aliases ? Or aren't they allowed in app commands?

bot/pagination.py
```diff
@@ -191,7 +191,7 @@ def _split_remaining_words(self, line: str, max_chars: int) -> t.Tuple[str, t.Op
     async def paginate(
         cls,
         lines: t.List[str],
-        ctx: Context,
+        ctx: t.Union[Context, discord.Interaction],
```
shtlrs, line 194:
```suggestion
        ctx: Context | discord.Interaction,
```

this can be done for the rest by ditching the typing import as well, if you want :p
Example: `list[str]` will work too

shtlrs (changes requested):

bot/exts/backend/error_handler.py
shtlrs: Actually, never mind me.
I just realized that this is not part of the lib.
My bad, sorry for that

bot/exts/info/tags.py
Ibrahim2750mi: The reason for using aliases was for ease of user but now as slash commands display all the commands it would kind of get clustered seeing one command take up all the space in the suggestions pop up, like eg: just writing `t` will take 6(3 for get, 3 for list) space. Also when just typing the user can see the commands and can now, even autofill the subgroup easily. And this is the reason discord has removed support for aliases for slash commands.

bot/exts/info/tags.py
Ibrahim2750mi: I think it gets my question unanswered, my question was that earlier the `tag` text command function returned `False` when no embed was sent but as of now the `tag` slash command sends an embed either way(if the tag was found and if the tag wasn't found). So, should the function now return `True` if an embed is sent for no tag found or should it return `False` when a tag is not found. 

Also about the error handler, I have already made changes to it i.e migrated it to handle slash command, thats why the tests even work.

bot/exts/backend/error_handler.py
```diff
@@ -168,27 +173,28 @@ async def try_get_tag(self, ctx: Context) -> None:
         the context to prevent infinite recursion in the case of a CommandNotFound exception.
         """
         tags_get_command = self.bot.get_command("tags get")
+        tags_get_command.can_run = can_run if can_run else self._can_run
```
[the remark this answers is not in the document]
Ibrahim2750mi: Can `lambda` be coroutine?

bot/exts/backend/error_handler.py
```diff
@@ -168,27 +173,28 @@ async def try_get_tag(self, ctx: Context) -> None:
         the context to prevent infinite recursion in the case of a CommandNotFound exception.
         """
         tags_get_command = self.bot.get_command("tags get")
```
Ibrahim2750mi, line 175: Change this to get the actual `tag` command as `self.bot.get_command` doesn't returns `app_commands`

bot/exts/backend/error_handler.py
```diff
@@ -168,27 +173,28 @@ async def try_get_tag(self, ctx: Context) -> None:
...
                 log.debug(log_msg)
                 return
         except errors.CommandError as tag_error:
             log.debug(log_msg)
-            await self.on_command_error(ctx, tag_error)
+            await self.on_command_error(interaction, tag_error)
             return
 
-        if await ctx.invoke(tags_get_command, argument_string=ctx.message.content):
+        if await interaction.invoke(tags_get_command, tag_name=interaction.message.content):
```
Ibrahim2750mi, line 193: Use the `callback` property of the `app_command` to invoke the `tag` command.

bot/exts/info/tags.py
Ibrahim2750mi: Also wookie the `!code` is already broken as the tag command now uses `discord.Interaction` and the ctx captured with the message containing `!code` has its interaction attribute return `None`, which means the `tag` command can't be invoked. A possible work around for this would be to have `get_command_ctx` in `tags.py` and `try_get_tag_ctx` in `error_handler.py` or we have to make the `get_command` ugly by doing `isinstance` everywhere.

wookie184: (sorry i'm sure you know this already, I just wanted to summarise to ensure we're on the same page).

The error handler is used as the individual tag names aren't registered as commands (e.g. `!code`), so we catch any CommandNotFound errors, and in those cases try and match the name given to a tag (this also allows us to do fuzzy matching that doesn't interfere with other tags).

Currently the `!tag get <name>` command, and direct `!<name>` invocation are pretty much exactly the same and use the same code. This has a few issues:
- It means the behaviour has to be the same (e.g. with behaviour if no tag is found, with `!tag get` we'd want to say that but with `!<name>` we'd want to be silent)
- It is now an issue since `/tag get <name>` is an interaction while `!<name>` isn't.

Because of this, as you say, it would probably be easiest to make the code for the direct invocation separate. There are a couple of different ways we could do this. One would be to have a `def search_tags_by_name(self, name: str)` function in the `Tags` cog, this can potentially be used by the interaction code and also in the error handler (by getting the cog using `bot.get_cog`), which could then handle what to do with the tags found itself. (that's not a requirement that you implement it like that at all, you can see what works best as you're doing it)

Ibrahim2750mi: Finally we are on the same page👍🏽

bot/exts/backend/error_handler.py
Ibrahim2750mi: Not need this now as we are bringing back normal invocation through ctx.

wookie184 (changes requested):
Looking good, thanks!

I wonder whether it would be better to just have a `/tag <name>` command, since currently `/tag get` with no arguments is the same as `/tag list`, so i'm not sure there's any point in having two commands. One would be shorter to type and simplify the code a little.

Ibrahim2750mi:
> I wonder whether it would be better to just have a `/tag <name>` command, since currently `/tag get` with no arguments is the same as `/tag list`, so i'm not sure there's any point in having two commands. One would be shorter to type and simplify the code a little.

I had a conversation about this on discord with @ChrisLovering, he suggested these making the commands like that.

wookie184 (approved):
This is great, thanks!

bot/exts/backend/error_handler.py
```diff
@@ -167,28 +167,33 @@ async def try_get_tag(self, ctx: Context) -> None:
...
-        if not tags_get_command:
-            log.debug("Not attempting to parse message as a tag as could not find `tags get` command.")
+        tags_cog = self.bot.get_cog("Tags")
+        if not tags_cog:
+            log.debug("Not attempting to parse message as a tag as could not find `Tags` cog.")
             return
+        tags_get_command = tags_cog.get_command_ctx
 
-        ctx.invoked_from_error_handler = True
+        maybe_tag_name = ctx.invoked_with
```
shtlrs, line 176: Nitpick:
```suggestion
        tag_name = ctx.invoked_with
```
It's better this way IMO, and then you can check whether it exists or not like you'e doing in the following line.

bot/exts/info/tags.py
```diff
@@ -91,7 +91,7 @@ def embed(self) -> Embed:
         embed.description = self.content
         return embed
 
-    def accessible_by(self, member: discord.Member) -> bool:
+    def accessible_by(self, member: Member | User) -> bool:
```
shtlrs, line 94: I think this can be kept to `Member`
Since the invoker needs to be a member of our server if they want to execute the command, so it won't resolve to an instance of `User`

bot/exts/info/tags.py
```diff
@@ -313,14 +233,14 @@ async def get_tag_embed(
...
-                if not tag.on_cooldown_in(ctx.channel)
+                if not tag.on_cooldown_in(channel)
             )
             return Embed(
                 title="Did you mean ...",
                 description=suggested_tags_text
             )
 
-    def accessible_tags(self, user: Member) -> list[str]:
+    def accessible_tags(self, user: Member | User) -> list[str]:
```
shtlrs, line 243: The type union comment should also apply here & everywhere else

bot/exts/info/tags.py
```diff
@@ -354,16 +274,45 @@ def tag_sort_key(tag_item: tuple[TagIdentifier, Tag]) -> str:
...
         )
 
-    @tags_group.command(name="get", aliases=("show", "g"), usage="[tag_group] [tag_name]")
-    async def get_command(self, ctx: Context, *, argument_string: Optional[str]) -> bool:
+    async def get_command_ctx(
+        self,
+        ctx: Context,
+        name: str
+    ) -> bool:
+        """Made specifically for `error_handler.py`, See `get_command` for more info."""
```
shtlrs, line 290: Docstrings should explain what the method does in a nutshell, you can also elaborate as much as you want when it comes to **why** it was created & **where** it will be needed.

bot/pagination.py
```diff
@@ -228,7 +228,10 @@ async def paginate(
         current_page = 0
 
         if not restrict_to_user:
-            restrict_to_user = ctx.author
+            if isinstance(ctx, discord.Interaction):
+                restrict_to_user = ctx.user
+            else:
```
shtlrs, line 233: I think this can boil down to 
```suggestion
if not restrict_to_user:
    restrict_to_user = getattr(ctx, "user", ctx.author)
```

shtlrs (changes requested):
Just a few nitpicks, other than that, it looks good !

bot/exts/backend/error_handler.py
wookie184: The following line only checks that it's not `None` (which should always be the case, but there's no harm in a check and it makes it so the type is correct later). It still might not be a tag name throughout it's usage, since this is run for *all* failed commands.

shtlrs: Ahein, I though the `maybe` prefix was then validated by the following line, The name is still bugging me a bit though :p

bot/pagination.py
wookie184: I don't think that would work as `Interaction` doesn't have an author, so it would raise an error (as all the function arguments are evaluated before the function is called)

shtlrs: Ho, you're actually right.
The way it works didn't actually match my thoughts, sorry for that

bot/exts/info/tags.py
Ibrahim2750mi: I mean it would be just copying the docstring of `get_command` and also I have seen these types of docstring referring to other method for more info in the codebase, some examples are in `helpers.py`. So I think that answers the **why** and I have already included the **where** it will be used.

So shall I still change it?

bot/pagination.py
onerandomusername: > I don't think that would work as `Interaction` doesn't have an author, so it would raise an error (as all the function arguments are evaluated before the function is called)

```suggestion
if not restrict_to_user:
    restrict_to_user = getattr(ctx, "user", None) or ctx.author
```

not the best but it would work with shtlrs idea.

bot/exts/backend/error_handler.py
Ibrahim2750mi: @shtlrs is it ok, if I resolve this now?

shtlrs (approved):
Looks great, thank you!

bot/exts/backend/error_handler.py
shtlrs: It's a trivial matter so I guess we could leave it like that.

PR #2415, opened 2023-02-25
Fix timeit commands with backticks after command name

wookie184:
In theory the list of allowed commands shouldn't be necessary, but I'd prefer to keep it to make the functionality more robust.

### DIFF
--- a/tests/bot/exts/backend/test_error_handler.py
+++ b/tests/bot/exts/backend/test_error_handler.py
@@ -47,7 +47,7 @@ class ErrorHandlerTests(unittest.IsolatedAsyncioTestCase):
         )
         self.cog.try_silence = AsyncMock()
         self.cog.try_get_tag = AsyncMock()
-        self.cog.try_run_eval = AsyncMock(return_value=False)
+        self.cog.try_run_fixed_codeblock = AsyncMock(return_value=False)
 
         for case in test_cases:
             with self.subTest(try_silence_return=case["try_silence_return"], try_get_tag=case["called_try_get_tag"]):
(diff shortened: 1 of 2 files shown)


shtlrs (approved):
Didn't test this, but it doesn't look like the changes will break anything.
Other than that, LGTM !

brodycritchlow (approved):
Looks fine to me, don't see anything that could break.

tests/bot/exts/backend/test_error_handler.py
```diff
@@ -47,7 +47,7 @@ async def test_error_handler_command_not_found_error_not_invoked_by_handler(self
         )
         self.cog.try_silence = AsyncMock()
         self.cog.try_get_tag = AsyncMock()
-        self.cog.try_run_eval = AsyncMock(return_value=False)
+        self.cog.try_run_fixed_codeblock = AsyncMock(return_value=False)
```
Ibrahim2750mi, line 50: `self.cog.try_run_fixed_codeblock` is not used in the test and even after removing it the tests run fine.
Although I do understand that it is used in `self.cog.on_command_error` but even after removing it the tests run fine, so is it really necessary to have this?

wookie184: I guess it still works because it would hit the line `if new_ctx.command is None:` which would probably be True. I think it makes sense to leave it though, since this test isn't meant to test that command so it makes the test more contained.

Another alternative would be to make it so `try_run_fixed_codeblock` is actually tested with both True and False, but I don't think it's worth it, since the test is already pretty complicated compared to the very simple code that it's actually testing.

Ibrahim2750mi (approved):

ChrisLovering (approved):

PR #2417, opened 2023-02-26
Add line to `!code` about the pastebin

swfarnsworth:
Something I've observed a lot is people running the `!code` command to tell users about markdown blocks, only for them to immediately reply saying "my code is too long". This PR adds a new line to the `!code` tag telling users about the pastebin.

For the sake of brevity, the added line says nothing about how to use the pastebin. If this PR is merged, and users who attempt to use the pastebin after seeing this tag end up finding it confusing, we can revisit the content of this tag.

image


wookie184 (approved):
Seems sensible, i'd say this can close https://github.com/python-discord/meta/issues/115 too.

ChrisLovering (approved):

Issue #2255, opened 2022-08-16
`!eval` breaks when using stdin


ZeroIntensity:
when trying to use stdin with the `eval` command, the bot breaks and gives an arbitrary response:
![image](https://user-images.githubusercontent.com/49501366/184989324-bfcf3fc2-6c3d-4cea-8871-4b98416292c2.png)
maybe there's some way to redirect a discord message to stdin when prompted?
if not, the bot should at least give some sort of nicer error when trying to use stdin rather than the `EOFError`

wookie184:
We could check if `EOFError` is in the result (as well as maybe checking that "input" is on their code and the status code indicated an error), and if so add a message after the codeblock saying that `input` doesn't work with the bot.

Making input work would be more difficult, it would require changes to snekbox (the service used to run the code), and it's also not obvious how passing the input in discord would work.

ZeroIntensity:
perhaps this could be brought up on the snekbox repo then?

wookie184:
We'd probably only implement taking stdin in snekbox if we wanted to have it for the bot, and I'm not convinced it's worth the effort given it's usually pretty simple to just replace the `input` with a fixed string.

I think that's a separate issue that could be opened on this repo though. Even if we did support `input` users would get an EOFError if they forgot to supply it, so improving that response would make sense anyway.

henri-climber:
Hello, 
not sure if should ask in this issue or in #2420.
I'd like to work on improving the error message. If I understood things right the plan is to just improve the error message and not integrate the possibilty of using input().

Ibrahim2750mi:
> I'd like to work on improving the error message. If I understood things right the plan is to just improve the error message and not integrate the possibilty of using input().

Well I thought to implement the issue myself after it got approved but now will try to find something else.

henri-climber:
> 

Oh, sure sorry. I didn't want to take it away from you.

mbaruh:
Given #2422 I think it's fine to mark this as complete. This can be re-opened/a new issue opened if there's a concrete suggestion that would intuitively allow using `input`, but now there's a note in the output indicating what the error means.

PR #2422, opened 2023-03-01
Add warning message when using `input`

### LINKED ISSUE #2420: Better error message for `!eval` when using input
Currently snekbox returns this error message

Which is not descriptive enough to show that we don't explicitly support `input()` and it can be just replaced by a string.

#2255
> I think that's a separate issue that could be opened on this repo though. Even if we did support input users would get an EOFError if they forgot to supply it, so improving that response would make sense anyway.

Ibrahim2750mi:
Closes #2420
Closes (MAYBE) #2255

> We could check if EOFError is in the result (as well as maybe checking that "input" is on their code and the status code indicated an error), and if so add a message after the codeblock saying that input doesn't work with the bot.

No need to check if input is in their as the [docs](https://docs.python.org/3/library/exceptions.html#EOFError) say `EOFError` is only raised when
> the [input()](https://docs.python.org/3/library/functions.html#input) function hits an end-of-file condition (EOF) without reading any data

Screenshots:

### DIFF
--- a/bot/exts/utils/snekbox.py
+++ b/bot/exts/utils/snekbox.py
@@ -344,8 +344,14 @@ class Snekbox(Cog):
                 log.trace("Formatting output...")
                 output, paste_link = await self.format_output(results["stdout"])
 
+            warning_message = ""
+            # 33 is the length of the error message. It is done to make sure the last line of output contains
+            # the error and the error is not manually printed by the author with a syntax error.
+            if "EOFError: EOF when reading a line" in output[-33:] and results['returncode'] == 1:
+                warning_message += ":warning: Note: `input` is not supported by the bot :warning:\n\n"
+
             icon = self.get_status_emoji(results)
-            msg = f"{ctx.author.mention} {icon} {msg}.\n\n```\n{output}\n```"
+            msg = f"{ctx.author.mention} {icon} {msg}.\n\n{warning_message}```\n{output}\n```"
             if paste_link:
                 msg = f"{msg}\nFull output: {paste_link}"
 


bot/exts/utils/snekbox.py
```diff
@@ -344,8 +344,14 @@ async def send_job(
                 log.trace("Formatting output...")
                 output, paste_link = await self.format_output(results["stdout"])
 
+            warning_message = ""
+            # 33 is the length of the error message. It is done to make sure the last line of output contains
+            # the error and the error is not manually printed by the author with a syntax error.
+            if "EOFError: EOF when reading a line" in output[-33:] and results['returncode'] == 1:
```
onerandomusername, line 350: should also sanity check the user had "input" in their code, IMO.

onerandomusername: Or the message should be altered to include stdin.

wookie184: As mentioned in the PR description, the docs say EOFError is specific to `input`, so i'm not sure what use case this would catch.

Ibrahim2750mi: This thought didn't cross my mind earlier what if someone raises the `EOFError`

Ibrahim2750mi: This thought didn't cross my mind earlier what if someone raises the `EOFError`, we should probably add a check for this.

wookie184: I don't think we need to add a check, I don't know why someone would raise the error manually, and even if they do the warning isn't incorrect, just unnecessary.

bot/exts/utils/snekbox.py
```diff
@@ -344,8 +344,14 @@ async def send_job(
...
+                warning_message += ":warning: Note: `input` is not supported by the bot :warning:\n\n"
```
wookie184, line 351:
```suggestion
            # the error and the error is not manually printed by the author with a syntax error.
            if results["stdout"].rstrip().endswith("EOFError: EOF when reading a line") and results["returncode"] == 1:
                warning_message += ":warning: Note: `input` is not supported by the bot :warning:\n\n"
```
I'd suggest:
- Double quotes for consistency
- Using `results["stdout"]` instead of `output` so the warning is still displayed if the error doesn't fit in the message.
- Using `endswith` so we don't need to hardcode the constant.

Ibrahim2750mi:
Also wookie thanks for reviewing

bot/exts/utils/snekbox.py
```diff
@@ -344,8 +344,15 @@ async def send_job(
                 log.trace("Formatting output...")
                 output, paste_link = await self.format_output(results["stdout"])
 
+            warning_message = ""
+
+            # This is done to make sure the last line of output contains the error
+            # and the error is not manually printed by the author with a syntax error.
+            if results["stdout"].rstrip().endswith("EOFError: EOF when reading a line") and results["returncode"] == 1:
+                warning_message += ":warning: Note: `input` is not supported by the bot :warning:\n\n"
+
```
onerandomusername, line 353: should this not be a part of get_results_message?

Ibrahim2750mi: Then we would aslo need to probably return the `warning_message` as well, which doesn't suits the function to return a string specifically for one outcome when it can be better sorted elsewhere.

mbaruh (approved):
A false positive here would be
```py
!e raise ValueError("EOFError: EOF when reading a line")
```
Not sure this is worth handling though. If you tried to find `input` in the eval input you would also have a false positive in the form of someone shadowing the built-in, as beginners often do.

Ibrahim2750mi:
> A false positive here would be
> 
> ```python
> !e raise ValueError("EOFError: EOF when reading a line")
> ```

I also realised this and commented it in the first review there wookie recommended to leave this as no user would do this.

> Not sure this is worth handling though. If you tried to find `input` in the eval input you would also have a false positive in the form of someone shadowing the built-in, as beginners often do.

Just doing ` if "input" in code:` is a grave mistake same goes with raise. If we check that there is `input` or `raise` in the code we need to do it with `ast` and not comparing it just with string.

Robin5605:
Sort of along the same lines as #2255 , how feasible would it be to add a modal with a `TextInput` that will be fed into `input()`?

Ibrahim2750mi:
> Sort of along the same lines as #2255 , how feasible would it be to add a modal with a `TextInput` that will be fed into `input()`?

I don't get whats the use of adding any input, `eval` command is not there to build an user interface. Any use of user input can be replaced be hard coded.

wookie184 (approved):
Works nicely, thanks!

PR #2432, opened 2023-03-03
Bump python version to 3 11

shtlrs:
This updates the python version that the bot runs on to 3.11

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Install Python Dependencies
         uses: HassanAbouelela/actions/setup-python@setup-python_v1.4.0
         with:
-          python_version: '3.10'
+          python_version: '3.11'
 
       # Check all of our non-dev dependencies are compatible with the MIT license.
       # If you added a new dependencies that is being rejected,
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 ghcr.io/chrislovering/python-poetry-base:3.10-slim
+FROM --platform=linux/amd64 ghcr.io/chrislovering/python-poetry-base:3.11-slim
 
 # Define Git SHA build argument for sentry
 ARG git_sha="development"
--- a/poetry.lock
+++ b/poetry.lock
@@ -347,18 +347,88 @@ files = [
 
 [[package]]
 name = "charset-normalizer"
-version = "2.1.1"
+version = "3.1.0"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 category = "main"
 optional = false
-python-versions = ">=3.6.0"
+python-versions = ">=3.7.0"
 files = [
-    {file = "charset-normalizer-2.1.1.tar.gz", hash = "sha256:5a3d016c7c547f69d6f81fb0db9449ce888b418b5b9952cc5e6e66843e9dd845"},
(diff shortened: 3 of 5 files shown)


.github/workflows/lint-test.yml
```diff
@@ -31,11 +31,11 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install Python Dependencies
-        uses: HassanAbouelela/actions/setup-python@setup-python_v1.3.1
+        uses: HassanAbouelela/actions/setup-python@setup-python_v1.4.0
         with:
           # Set dev=true to install flake8 extensions, which are dev dependencies
-          dev: true
-          python_version: '3.10'
+          python_version: '3.11'
```
shtlrs, line 37: Actually, we might need to preserve the dev input we provide, I'll confim in a bit

wookie184:
The test failure looks interesting, seems to be here:
https://github.com/python-discord/bot/blob/395e058c52ebc718f7c0c9acec710c71f45d7b38/tests/_autospec.py#L53-L54
not surprising since it seems like that was intended to be a private method.

Maybe replacing it with `pkgutil.resolve_name` will fix it? (https://github.com/python/cpython/pull/18544)

shtlrs:
> The test failure looks interesting, seems to be here:
> 
> https://github.com/python-discord/bot/blob/395e058c52ebc718f7c0c9acec710c71f45d7b38/tests/_autospec.py#L53-L54
> 
> 
> not surprising since it seems like that was intended to be a private method.
> Maybe replacing it with `pkgutil.resolve_name` will fix it? ([python/cpython#18544](https://github.com/python/cpython/pull/18544))

I had actually forgotten about this PR.
I'll try to look into the option you proposed.

one other option we had discussed the other day was monkey patching it by including it in our testing utils.

We'd likely want to avoid that if there's another way that can do this out of the box.

MarkKoz:
> The test failure looks interesting, seems to be here:
> 
> https://github.com/python-discord/bot/blob/395e058c52ebc718f7c0c9acec710c71f45d7b38/tests/_autospec.py#L53-L54
> 
> not surprising since it seems like that was intended to be a private method.
> 
> Maybe replacing it with `pkgutil.resolve_name` will fix it? ([python/cpython#18544](https://github.com/python/cpython/pull/18544))

Yeah, this is intended to be like [this](https://github.com/python/cpython/blob/3.11/Lib/unittest/mock.py#L1662-L1665). That's where I got the original code from, and it used to use that internal method.

MarkKoz (approved):

wookie184 (approved):
LGTM, tested and seems to work.

Two test warnings are raised due to this:
```
bot-hu2rt1Uv-py3.11\Lib\site-packages\feedparser\encodings.py:29: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
    import cgi
```
The issue tracking that is https://github.com/kurtmckee/feedparser/issues/330, it seems it was already fixed (https://github.com/kurtmckee/feedparser/commit/ed8c762286aa1d8a248a6900190bd02c01ac2f8d) and is just waiting on a release.
```
bot-hu2rt1Uv-py3.11\Lib\site-packages\discord\player.py:28: DeprecationWarning: 'audioop' is deprecated and slated for removal in Python 3.13
    import audioop
```
The usage seems to be here: https://github.com/Rapptz/discord.py/blob/bb7668f8a58ba4b8161edeb77f8936ff807d6537/discord/player.py#L632

Given neither of them are issues currently I'm not sure they need to block this. I'll let you merge since you may need to relock first.

jchristgit (approved):

PR #2438, opened 2023-03-04
Migrate from role-based mutes to native timeouts

mbaruh:
> **Note** to the core-devs, once this is ready please leave the merging to me.

## Changes

- Makes use of the native timeout instead of adding the Muted role.
- Renames all references to the "mute" infraction to "timeout", except in command aliases for ease of transition.
- Maintains support for the old functionality (pardoning users with the muted role, applying timeout to users who rejoin and are not yet timed out because they originally had the role). This can be removed (the relevant parts are marked with TODOs) after there are no longer users with the old mute.


bot/constants.py
```diff
@@ -493,7 +493,7 @@ class Roles(metaclass=YAMLGetter):
 
     contributors: int
     help_cooldown: int
-    muted: int
+    muted: int  # TODO remove when no longer relevant.
```
shtlrs, line 496: Is this still of relevance because of the potentially existing mutes ?
We could still rename it to "Timed out" to  reflect that someone is timed out in discord, because i think it's quite handy to have that indicated in the user's roles

bot/exts/moderation/infraction/infractions.py
```diff
@@ -27,37 +29,45 @@
...
     category = "Moderation"
     category_description = "Server moderation tools."
 
     def __init__(self, bot: Bot):
-        super().__init__(bot, supported_infractions={"ban", "kick", "mute", "note", "warning", "voice_mute"})
+        super().__init__(bot, supported_infractions={"ban", "kick", "timeout", "note", "warning", "voice_mute"})
 
         self.category = "Moderation"
-        self._muted_role = discord.Object(constants.Roles.muted)
+        self._muted_role = discord.Object(constants.Roles.muted)  # TODO remove when no longer relevant.
```
shtlrs, line 45: If you approve my suggestion from above to keep the role, then we'd better rename this

bot/exts/moderation/infraction/infractions.py
```diff
@@ -214,15 +224,26 @@ async def tempmute(
...
         if duration is None:
             duration = await Duration().convert(ctx, "1h")
-        await self.apply_mute(ctx, user, reason, duration_or_expiry=duration)
+        else:
+            now = arrow.utcnow()
+            if isinstance(duration, relativedelta):
+                duration += now
+            if duration > now + MAXIMUM_TIMEOUT_DAYS:
+                await ctx.send(f":x: A timeout cannot be longer than {MAXIMUM_TIMEOUT_DAYS.days} days.")
+                return
```
shtlrs, line 241: Would capping it to `MAXIMUM_TIMEOUT_DAYS` be a good idea here ? (-1 minute since it's exclusive like you said)

bot/exts/moderation/infraction/infractions.py
```diff
@@ -402,10 +423,13 @@ async def action() -> None:
             # Skip members that left the server
             if not isinstance(user, Member):
                 return
+            duration_or_expiry = kwargs["duration_or_expiry"]
```
shtlrs, line 426: Are you getting this from kwarg to avoid altering the signature ?

It's just that I think that if it's something that's cohesive with the command, it's worth changing the signature along with it

shtlrs (commented):
Looks great, just a couple of nitpicks

bot/constants.py
mbaruh: Yeah it's just for the existing mutes. As a moderator it will be easy to see that they're timed out (for example you will be able to remove their timeout in the context menu). Their messages will appear grayed out in chat with a red indicator.

bot/exts/moderation/infraction/infractions.py
mbaruh: I think the idea was to make the `apply_x` functions all have the same signature. It's not necessarily the best, but I'm not sure I want to start messing with it.

bot/exts/moderation/infraction/infractions.py
mbaruh: image
The people have spoken

bot/exts/moderation/infraction/infractions.py
shtlrs: Hummm, I guess we could leave it then, I don't think apply_timeout would be used somewhere else so that's probs fine

bot/exts/moderation/infraction/infractions.py
shtlrs: Awesome :D

bot/constants.py
shtlrs: Sounds good

bot/exts/moderation/infraction/infractions.py
mbaruh: 8cfd1b2

bot/exts/moderation/infraction/infractions.py
```diff
@@ -27,37 +30,49 @@
     from bot.exts.moderation.watchchannels.bigbrother import BigBrother
 
 
+MAXIMUM_TIMEOUT_DAYS = timedelta(days=28)
+TIMEOUT_CAP_MESSAGE = (
```
shtlrs, line 34: Suggestion:
This can be further improved by adding more context through mentioning which user was being infracted

shtlrs (approved):
Looks great, just one tiiiny suggestion

jchristgit (approved):
Thanks!

shtlrs (approved):
Re-tested & works even better :D

PR #2439, opened 2023-03-05
Make help showable through button on command error message.

### LINKED ISSUE #126: Consistent Error Responses
Some errors use plain messages (e.g. general error handlers and moderation cogs) while others use embeds, sometimes even including [humorous replies](https://github.com/python-discord/bot/blob/master/bot/constants.py#L585) (e.g. reminders and clean cogs). Errors don't necessarily refer to `CommandError` here - it could be _any_ error that is being handled, even ones that aren't exceptions.

Embeds have more formatting options, allow room for those humorous replies, and are more prominent in the sense that 1. bigger = more noticeable and 2. can't be confused for a regular user's message. On the other hand, plain messages have less clutter around them and are simpler to create in code (may be a moot point if a generic utility function is added to create the embeds).

* Should the plain/embed be mutually exclusive? In other words, should everything only use one of them to stay consistent, or should some things be allowed use the other? If not, which should be used by what?
* What style(s) should be used to stay consistent? Should the humorous replies always be used? Should plain messages always be prefixed by emojis?
* Can and should a generic function be written to create these error messages? #674 is doing that for embeds.

All the areas where errors are being handled should be identified so that it is clear what changes need to be made and where.

**Originally posted by Kingsley McDonald:**

many error responses in the `Moderation` cog aren't sent as an embed. instead, the messages are sent with a very simple (and ugly) :x:. all occurrences of this should be changed to be consistent with the rest of the bot's responses.

wookie184:
closes #126

Notes:
- Help button can be used by anyone
- Delete button can be used by author and moderators
- I would have used https://bot-core.pythondiscord.com/main/output/pydis_core.utils.interactions.html#pydis_core.utils.interactions.ViewWithUserAndRoleCheck, but it seems like that can only lock *all* buttons. I couldn't see a way of overriding the check method to change that behaviour.
- Using the method from the custom help command is a bit ugly, but I wanted to keep this change minimal and it should be fine.

Before:
image

After:
image

And when clicking "Help":
image


Ibrahim2750mi (commented):
LGTM!

bot/exts/backend/error_handler.py
```diff
@@ -15,6 +18,51 @@
...
+        self.add_item(self.delete_button)
+
+    @discord.ui.button(label='Help', style=ButtonStyle.primary)
+    async def help_button(self, interaction: Interaction, button: discord.ui.Button) -> None:
+        """Send an ephemeral message with the contents of the help command."""
+        await interaction.response.send_message(embed=self.help_embed, ephemeral=True)
+
+    async def on_timeout(self) -> None:
+        """Remove the view from `self.message` if set."""
+        if self.message:
```
shtlrs, line 61: I'm curious, will we ever fall into this case where `self.message` is None ? 
The timer will start when the view is sent, and it cannot timeout in the meantime (I tried putting timeout to 1 second and an asyncio.sleep(5) before sending it :P just to verify), so view.message will always be present since that's not something that can fail.

If what I'm saying is true, then the `if self.message` check can be removed

shtlrs (changes requested):
Looks good & works well.

The timeout comment is the one I think should be addressed the most.

bot/exts/backend/error_handler.py
```diff
@@ -15,6 +18,51 @@
...
+            with contextlib.suppress(NotFound):
+                await self.message.edit(view=None)
```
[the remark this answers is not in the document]
wookie184: The docs suggest that the default value is 180 (https://discordpy.readthedocs.io/en/stable/interactions/api.html?highlight=discord%20ui%20view#discord.ui.View), and iirc it worked fine when I was testing.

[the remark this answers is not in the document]
shtlrs: I could have confused the sections I've read then. 180 seems fine indeed

bot/exts/backend/error_handler.py
```diff
@@ -15,6 +18,51 @@
 log = get_logger(__name__)
 
 
+class CheckedDeleteMessageButton(DeleteMessageButton):
+    """Button to delete a message, restricted to the message owner and moderators."""
+
+    def __init__(self, owner: User | Member):
+        super().__init__()
+        self.owner = owner
```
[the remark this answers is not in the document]
shtlrs: Fair enough

bot/exts/backend/error_handler.py
wookie184: I think you're probably correct that it shouldn't ever be `None` with the current code. I mainly put the check there because it's more obviously correct since `self.message` has the type `Message | None`, and it doesn't do any harm to check.

shtlrs: Sure sure, which is why I said that I was only curious to know. Not that there's an issue with the way it is or anything

bot/exts/backend/error_handler.py
```diff
@@ -15,6 +18,51 @@
...
+
+    async def callback(self, interaction: Interaction) -> None:
+        """Delete the message if the user is the owner of the message or a moderator."""
+        user = interaction.user
+        if (
+            user.id == self.owner.id or
+            (
+                isinstance(user, Member) and
+                any(user.get_role(role) for role in MODERATION_ROLES)
```
mbaruh, line 35: Out of scope for this PR probably, but we do this check in various places in the codebase. I wonder if there should be a single function that does it to un-ugly-fy it.

mbaruh (changes requested):

bot/exts/backend/error_handler.py
wookie184: Yeah, possibly something that could go in bot-core. I'll leave it out of this PR though

bot/exts/backend/error_handler.py
```diff
@@ -15,6 +18,53 @@
...
+
+    def __init__(self, help_embed: Embed, owner: User | Member):
+        super().__init__()
+        self.help_embed = help_embed
+        self.message: Message | None = None
+
+        self.delete_button = CheckedDeleteMessageButton(owner)
+        self.add_item(self.delete_button)
+
+    @discord.ui.button(label='Help', style=ButtonStyle.primary)
```
wookie184, line 56:
```suggestion
    @discord.ui.button(label="Help", style=ButtonStyle.primary)
```

shtlrs (approved):
Thanks !

bot/exts/backend/error_handler.py
```diff
@@ -15,6 +18,53 @@
...
+            )
+        ):
+            await super().callback(interaction)
+        else:
+            await interaction.response.send_message(
+                "You can only delete responses to your own command invocations!", ephemeral=True
+            )
+
+
+class HelpEmbedView(discord.ui.View):
```
ChrisLovering, line 45: If you subclass [pydis_core.utils.interactions.ViewWithUserAndRoleCheck](https://bot-core.pythondiscord.com/main/output/pydis_core.utils.interactions.html#pydis_core.utils.interactions.ViewWithUserAndRoleCheck) instead, you can pass it allowed_users and allowed_roles for owner & mods, this will allow you to drop the subclass of DeleteMessageButton entirely and use it directly.

It will also let you remove the on_timeout handler, since you can set `.message` on the view and it will delete on timeout automatically.

ChrisLovering (changes requested):

bot/exts/backend/error_handler.py
wookie184: I think I tried to use that but couldn't find a way to limit the delete button to invoker & mods while still allowing anyone to use the help button.

ChrisLovering: I mentioned that in an edit to me last message in the last paragraph, I'm not entirely sure that's needed.

If we completely set on needing that, then we could use that view from bot-core to handle the view itself, and just pass the `@everyone` role as the allowed roles, then the delete button can retain your callback code.

This then uses the timeout and stop code from bot core, which also covers things like if the message being editing is in a closed thread, which would error under this implementation

bot/exts/backend/error_handler.py
ChrisLovering: Added https://github.com/python-discord/bot-core/pull/194 that can help with this, but doesn't need to be part of this PR.

bot/exts/backend/error_handler.py
wookie184: I found a way to override the interaction check to let everyone use the help command https://github.com/python-discord/bot/pull/2439/commits/c053afe871543c79de241e137d2ab91b581b3ab4#diff-9a635f11cfc93b25d2c440dc75d5f492d682fd9770feebc4e5216000929b9ab9R30-R42, a little hacky but I think it's nicer than having to have the whole deletion button logic. Thoughts?

ChrisLovering (approved):

PR #2446, opened 2023-03-06
Migrate `!source` to `/source` (app_commands)

### LINKED ISSUE #2022: Incorrect handling of tags by the `!source` command

### LINKED ISSUE #2424: Make `!source` a slash command
I think the `!source` command should be made a slash command because, similarly to the reason why we made `!tags` a slash command, it would make it easier to look up the name of a cog without having to leave the Discord channel if you don't exactly remember the name offhand.
Note: I'm fine with someone else implementing this if approved, I don't really have the time right now to contribute at all

Ibrahim2750mi:
Closes #2022
Previously the error prompted because isinstance was returning False when an TagIdentifier object was compared to itself. Instead of checking that, we check if the source object issubclass of commands.Cog and pass the TagIdentifier to else.

Closes #2424
Screenshots:


bot/exts/info/source.py
```diff
@@ -83,12 +96,12 @@ async def build_embed(self, source_object: SourceType) -> Optional[Embed]:
         elif isinstance(source_object, commands.Command):
             description = source_object.short_doc
             title = f"Command: {source_object.qualified_name}"
-        elif isinstance(source_object, TagIdentifier):
-            title = f"Tag: {source_object}"
-            description = ""
-        else:
+        elif issubclass(type(source_object), commands.Cog):
```
Ibrahim2750mi, line 99: Closes #2022

Previously the error prompted because `isinstance` was returning False when an `TagIdentifier` object was compared to itself. Instead of checking that, we check if the source object `issubclass` of commands.Cog and pass the TagIdentifier to else.

bot/exts/info/source.py
```diff
@@ -19,18 +19,31 @@ class BotSource(commands.Cog):
...
         self.bot = bot
 
-    @commands.command(name="source", aliases=("src",))
-    async def source_command(self, ctx: commands.Context, *, source_item: SourceConverter = None) -> None:
+    @app_commands.command(name="source")
+    async def source_command(
+        self,
+        interaction: Interaction,
+        *,
+        source_item: app_commands.Transform[SourceType, SourceTrasnformer] = None
```
Ibrahim2750mi, line 27: This parameter doesn't suits in the UI well, please suggest a new name.

shtlrs: Maybe `target_utility` ?

bot/converters.py
```diff
@@ -267,37 +268,52 @@ async def convert(self, ctx: Context, arg: str) -> int:
...
             f"Unable to convert '{escaped_arg}' to valid command{', tag,' if show_tag else ''} or Cog."
         )
+        return escaped_arg
+
+    async def autocomplete(self, interaction: discord.Interaction, value: str) -> list[app_commands.Choice[str]]:
+        """Autocompleter for `/source` command."""
+        all_cogs = interaction.client.cogs
+        tags_cog = all_cogs.get("Tags", None)
+        names = [*all_cogs.keys(), *interaction.client.all_commands.keys()]
+        if tags_cog:
```
shtlrs, line 308: This currently doesn't suggest the `tag` command itself when I do `/source tag`
Instead, it only returns the cog

bot/exts/info/source.py
```diff
@@ -19,18 +19,30 @@ class BotSource(commands.Cog):
...
-            await ctx.send(embed=embed)
+            await interaction.response.send_message(embed=embed)
             return
 
-        embed = await self.build_embed(source_item)
-        await ctx.send(embed=embed)
+        embed = None
+        ephemeral = False
+        if isinstance(source_item, str):
+            description = f"**Unable to convert '{source_item}' to valid command, tag, or Cog.**"
```
shtlrs, line 40: Maybe staying consistent with capitalization is better here

bot/exts/info/source.py
```diff
@@ -83,12 +95,12 @@ async def build_embed(self, source_object: SourceType) -> Optional[Embed]:
...
-        else:
+        elif issubclass(type(source_object), commands.Cog):
             title = f"Cog: {source_object.qualified_name}"
             description = source_object.description.splitlines()[0]
+        else:
+            title = f"Tag: {source_object}"
+            description = ""
 
         embed = Embed(title=title, description=description)
         embed.add_field(name="Source Code", value=f"[Go to GitHub]({url})")
```
shtlrs, line 106: This currently doesn't work for Tags because it returns the wrong path

shtlrs: image

Notice how the `bot` directory is missing, which results in a 404 when clicking on the guild.

Also, would be great it we could stay consitent with slashes instead of anti slashes for paths

shtlrs (changes requested):
Looks good overall, just small things that need fixing

bot/exts/info/source.py
Ibrahim2750mi: Fixed in https://github.com/python-discord/bot/pull/2446/commits/758a99417ea102edd726fd8f7167a4dbeaade556

bot/exts/info/source.py
Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/525fb54aed18526ed54e995e6e5bab6721c49baf

bot/exts/info/source.py
Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/525fb54aed18526ed54e995e6e5bab6721c49baf

bot/exts/info/source.py
```diff
@@ -1,62 +1,75 @@
...
+        embed = await self.build_embed(cog_command_or_tag) if not embed else embed
+        await interaction.response.send_message(embed=embed, ephemeral=ephemeral)
+
+    def get_source_link(self, source_object: SourceType) -> Tuple[str, str, Optional[int]]:
         """
-        Build GitHub link of source item, return this link, file location and first line number.
+        Build GitHub link of source object, return this link, file location and first line number.
 
-        Raise BadArgument if `source_item` is a dynamically-created object (e.g. via internal eval).
+        Raise BadArgument if `source_object` is a dynamically-created object (e.g. via internal eval).
```
Ibrahim2750mi, line 52: Raise `CommandInvokeError` instead.

bot/exts/info/source.py
```diff
@@ -77,18 +90,21 @@ async def build_embed(self, source_object: SourceType) -> Optional[Embed]:
...
         url, location, first_line = self.get_source_link(source_object)
 
-        if isinstance(source_object, commands.HelpCommand):
+        if isinstance(source_object, HelpCommand):
             title = "Help Command"
             description = source_object.__doc__.splitlines()[1]
-        elif isinstance(source_object, commands.Command):
+        elif isinstance(source_object, Command):
             description = source_object.short_doc
             title = f"Command: {source_object.qualified_name}"
```
shtlrs, line 98: If you want, we can change this to
```suggestion
            title = f"Text command: {source_object.qualified_name}"
```

bot/converters.py
```diff
@@ -267,37 +268,60 @@ async def convert(self, ctx: Context, arg: str) -> int:
...
             return cog
 
-        cmd = ctx.bot.get_command(argument)
+        cmd = interaction.client.get_command(cog_command_or_tag)
         if cmd:
             return cmd
 
-        tags_cog = ctx.bot.get_cog("Tags")
+        slash_cmd = interaction.client.tree.get_command(cog_command_or_tag)
+        slash_cmd = slash_cmd if isinstance(slash_cmd, app_commands.Command) else None
```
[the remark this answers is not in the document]
Ibrahim2750mi: Alright, I will add the `app_commands.Group`, not so sure about ContextMenu, but if you say so I will add it.

bot/exts/info/source.py
Ibrahim2750mi: Thanks for noticing for that, I didn't notice it.

Ibrahim2750mi: According to discord-dev website they are called `Prefix-based commands`, shall I change it to that or this?

shtlrs: I don't think most people would know them under that name, so a text command is better.
Let's see what others have to say about it though

Ibrahim2750mi: I am gonna go with prefix-based commands because the name explains the usage better than text command. As slash commands are also text commands.

bot/converters.py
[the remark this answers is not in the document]
Ibrahim2750mi: I am currently just adding `app_commands.Group` and leaving `app_commands.ContextMenu`.

bot/exts/info/source.py
Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/a1820fa6836d8a0857d93672af9e7738625deafe

bot/converters.py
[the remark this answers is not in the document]
Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/a1820fa6836d8a0857d93672af9e7738625deafe

bot/exts/info/source.py
Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/4601217e7c241f101b1cdf8b9779e4536e4dd4b5

bot/exts/backend/error_handler.py
```diff
@@ -98,6 +106,42 @@ async def on_command_error(self, ctx: Context, e: errors.CommandError) -> None:
...
+    async def on_app_command_error(self, interaction: Interaction, e: app_commands.AppCommandError) -> None:
+        """
+        Error handler for app commands.
+
+        See on_command_error for detailed description.
+        """
+        await interaction.response.defer(ephemeral=True)
+
+        debug_message = (
+            f"Command {interaction.command} invoked by {interaction.user} with error "
```
shtlrs, line 119: Use `command.qualified_name` here as well

bot/exts/backend/error_handler.py
```diff
@@ -98,6 +106,42 @@ async def on_command_error(self, ctx: Context, e: errors.CommandError) -> None:
...
+        if isinstance(e, app_commands.TransformerError):
+            log.debug(debug_message)
+            self.bot.stats.incr("errors.transformer_error")
+            embed = self._get_error_embed("Transformer error", e.__cause__)
+        elif isinstance(e, app_commands.CommandInvokeError):
+            log.debug(debug_message)
+            self.bot.stats.incr("errors.appcommand_invoke_error")
+            embed = self._get_error_embed("App command invoke error", e.original)
+        else:
+            cause = e.__cause__
```
shtlrs, line 132: The `cause` here is not really guaranteed to have value.
For example, if I remove the part where it checks it's a TranformerError & throw one explicitly, the cause will be None.

What you can do instead is check whether it's none or not, otherwise use the original.

shtlrs (changes requested):

bot/converters.py
```diff
@@ -267,37 +268,60 @@ async def convert(self, ctx: Context, arg: str) -> int:
...
         )
+        return escaped_arg
+
+    async def autocomplete(self, interaction: discord.Interaction, value: str) -> list[app_commands.Choice[str]]:
+        """Autocompleter for `/source` command."""
+        all_cogs = interaction.client.cogs
+        tags_cog = all_cogs.get("Tags", None)
+        names = [*all_cogs.keys(), *interaction.client.all_commands.keys()]
+        names += [
+            slsh_cmd.qualified_name
```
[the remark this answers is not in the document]
Ibrahim2750mi: FYI that would result in flake8 saying line character limit reached 122 > 120.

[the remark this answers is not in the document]
shtlrs:
```py
names += [
            slash_cmd.qualified_name
            for slash_cmd in interaction.client.tree.get_commands()
            if not isinstance(slash_cmd, app_commands.ContextMenu)
        ]
 ```

File: bot/converters.py [hunk 12]
```diff
@@ -267,37 +268,60 @@ async def convert(self, ctx: Context, arg: str) -> int:
...
             show_tag = False
         else:
-            identifier = TagIdentifier.from_string(argument.lower())
+            identifier = TagIdentifier.from_string(cog_command_or_tag.lower())
             if identifier in tags_cog.tags:
                 return identifier
-        escaped_arg = escape_markdown(argument)
+        escaped_arg = escape_markdown(cog_command_or_tag)
 
         raise BadArgument(
```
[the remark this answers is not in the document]
Ibrahim2750mi: It would be handled by the error handler because any error raised in the transformer raises `TransformerError`. The reason I did this was because I wanted to pass the error message to the error handler through the `cause` dunder. Rasining `TransformerError` will only add complexity as you can't pass your own custom error message to it.

bot/converters.py
```diff
@@ -267,37 +268,60 @@ async def convert(self, ctx: Context, arg: str) -> int:
...
-            f"Unable to convert '{escaped_arg}' to valid command{', tag,' if show_tag else ''} or Cog."
+            f"Unable to convert '{escaped_arg}' to valid command{', tag' if show_tag else ''} or cog."
         )
+        return escaped_arg
```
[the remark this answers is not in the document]
Ibrahim2750mi: Will be removed, thanks for reminding.

bot/exts/backend/error_handler.py
```diff
@@ -98,6 +106,42 @@ async def on_command_error(self, ctx: Context, e: errors.CommandError) -> None:
...
+        await interaction.response.defer(ephemeral=True)
+
+        debug_message = (
+            f"Command {interaction.command} invoked by {interaction.user} with error "
+            f"{e.__class__.__name__}: {e}"
+        )
+
+        if isinstance(e, app_commands.TransformerError):
+            log.debug(debug_message)
+            self.bot.stats.incr("errors.transformer_error")
```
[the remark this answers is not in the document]
Ibrahim2750mi: I also don't know much about the namespace, are you sure?

bot/exts/backend/error_handler.py
Ibrahim2750mi: `e.original` is only present for `app_commands.CommandInvokeError` iirc. Also the cause is the original statement that caused the error, eg:

```py
# this happens internally somewhere(not in a transformer or a command)
raise ValueError("a is smaller than b")
```

The cause will be returned as "a is smaller than b".


The only times it retuns `None` is when you explicitly raise an app_command error.

bot/exts/backend/error_handler.py
```diff
@@ -98,6 +106,42 @@ async def on_command_error(self, ctx: Context, e: errors.CommandError) -> None:
...
+            embed = self._get_error_embed("Transformer error", e.__cause__)
+        elif isinstance(e, app_commands.CommandInvokeError):
+            log.debug(debug_message)
+            self.bot.stats.incr("errors.appcommand_invoke_error")
```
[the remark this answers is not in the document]
Ibrahim2750mi: Oh, I added the `app` prefix because then it will overshadow the `command_invoke_error` used somewhere in the error handler.

bot/converters.py
```diff
@@ -267,37 +268,60 @@ async def convert(self, ctx: Context, arg: str) -> int:
...
             return cmd
 
-        tags_cog = ctx.bot.get_cog("Tags")
+        slash_cmd = interaction.client.tree.get_command(cog_command_or_tag)
+        slash_cmd = slash_cmd if not isinstance(slash_cmd, app_commands.ContextMenu) else None
+        if slash_cmd:
+            return slash_cmd
+
+        tags_cog = interaction.client.get_cog("Tags")
         show_tag = True
```
shtlrs, line 293: it looks like show_tag isn't used anywhere, can we remove it then ?

Ibrahim2750mi: It is see https://github.com/python-discord/bot/pull/2446/files#diff-a39b0dd807dd61d8677c3fd3c20b409abf4c036c6dbdab15288bd592c3b843ebR304

bot/exts/backend/error_handler.py
Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/13aa29a6ed1069660b7d20aa15222d953a8bf59b

bot/converters.py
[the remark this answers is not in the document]
Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/13aa29a6ed1069660b7d20aa15222d953a8bf59b

bot/exts/backend/error_handler.py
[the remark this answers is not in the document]
Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/13aa29a6ed1069660b7d20aa15222d953a8bf59b

bot/converters.py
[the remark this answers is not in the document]
shtlrs: Oh, I didn't know that the command would be enclosed with `TransformerError.

Yeah `TransformerError` doesn't allow you to do that much, even though you could bend around it by passing the error messafe to the `opt_type` parameter.

I just don't like the fact that we'd be throwing a test command error from a slash one.

bot/exts/backend/error_handler.py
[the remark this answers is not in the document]
shtlrs: It's fine to leave it as `error.` or now

bot/converters.py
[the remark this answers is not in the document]
Ibrahim2750mi: I don't know if theres a better way maybe we can raise `ValueError` instead?

bot/exts/backend/error_handler.py
shtlrs: Sure, I was referring to the case when we'd raise an app command error explicitly. As I said before, I don't like raising a TextCommand error from within a slash command context

The cause should be present when an exception if raised from another.

```py
try:
    raise ValueError("initial")
except ValueError as e:
    raise KeyError("second") from e
```

when you catch that later, 

```py
try:
    ...
except KeyError as e:
    print(e.__cause__)
    # this will print: initial
````

bot/converters.py
shtlrs: Oh sorry for that, it got mixed up when i changed to code to try exception related stuff

bot/exts/backend/error_handler.py
Ibrahim2750mi: I don't think app command errors are meant to be raised explicitly, I maybe wrong.

Sorry, I don't understand the second part of your comment. Can you elaborate?

shtlrs: Let's go with raising a ValueError, which will then become TransformerError, and catch it here

Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/6059dd077468b1338109327793400469ed96096e

bot/converters.py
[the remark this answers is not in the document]
Ibrahim2750mi: Updated in https://github.com/python-discord/bot/pull/2446/commits/6059dd077468b1338109327793400469ed96096e

bot/converters.py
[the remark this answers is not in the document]
Ibrahim2750mi: This is resolved right?

Ibrahim2750mi:
I have also added unit tests for error handler so someone in future doesn't break them, or if they they will get notified.

shtlrs:
Hey @Ibrahim2750mi, I'm sorry this took a while.
Here's a couple of things I noticed
* It currently have duplicates for some command groups, like infractions
image
image
 
The second & third options point to the same command group.


* We don't have autocomplete for subcommands of a certain group, for example the `edit` subcommand of the `infraction group`

image
However, it still works when i force that value, meaning when I do write `infraction edit` and press enter, it fetches it corretly
image

Xithrius:
@Ibrahim2750mi Greetings. Please fix the conflicts contained within this PR.

Thanks!

Canttuchdiz:
I'd like to be assigned.

Xithrius:
@Canttuchdiz Hello, will you be continuing updates to this PR?

wookie184:
I think @hedyhli's proposed solution to https://github.com/python-discord/bot/issues/2022 is a better approach. Along with this, as this has quite a few merge conflicts, is stale, and there are still some existing bugs, I'm going to close the PR and suggest anyone wanting to work on #2424 makes a new one.

PR #2452, opened 2023-03-08
Make tag slash command guild only

### LINKED ISSUE #2450: Disable slash commands in DMs
So far we've had https://github.com/python-discord/bot/blob/bf8f8f4c1f9522a942c88ca69a2d48427d2bbc28/bot/exts/backend/security.py#L21
Which makes sure that prefix commands are run in the server and not in DMs.

This doesn't seem to work for slash commands so we're getting bugs for various features which don't work in DMs.

We should either have a check which handles slash commands, or find some other way to deal with it.

shtlrs:
Closes #2450 

This disables the tag slash command in the DMs.

This allows us to have flexibility when it comes to selecting which command to enable in DMs or not instead of having a check that get clunky depending on special conditions.


Ibrahim2750mi (approved):
LGTM!

shenanigansd (approved):

ChrisLovering (approved):

PR #2458, opened 2023-03-09
Add botstrap depency group and task

shtlrs:
Adds a dependency group for the boostrapping script and a task to launch it

For some reason, these were lost during the merge.


HassanAbouelela (commented):
Just move the dependency into the main dev group. It doesn’t need to be on its own for basically all devs, and it would save us effort trying to guide people to install it.

ChrisLovering (approved):

Ibrahim2750mi (approved):

PR #2459, opened 2023-03-09
Update nested delimiter to double underscore

shtlrs:
This fixed the way pydantic tries to pickup nested values when the attributes have underscores in them.

Without this, a case like the following with a delimiter of `_`

```py
class Webhook(BaseModel):
    id: int
    channel: Optional[int]


class _Webhooks(EnvConfig):
    EnvConfig.Config.env_prefix = "webhooks_"

    big_brother: Webhook = Webhook(id=123456789, channel=Channels.big_brother)
```

and an `.env.server` like this
```text
webhooks_big_brother_id=987654321
```

Will make pydantic think that it'll need to map to `Webhooks.big.brother.id`, which breaks the value

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -258,7 +258,7 @@ class ThreadArchiveTimes(Enum):
 
 class Webhook(BaseModel):
     id: int
-    channel: Optional[int]
+    channel: int
 
 
 class _Webhooks(EnvConfig):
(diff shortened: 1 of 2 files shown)


Ibrahim2750mi (approved):

Ibrahim2750mi:
It works!

PR #2467, opened 2023-03-12
Create python-help forum channel in `botstrap`

### LINKED ISSUE #2466: Add "native" suppot for forum channels in `botstrap`
Discord templates do not include support for community-server based features, such as forum channels, announcement channels, etc...

This makes it not possible to have an out of the box solution to map the help system forum to the `.env.server` file.

The way I propose to do this is, upon launching `botstrap.py`:

1. Check if the server is a community server -> If not: make it one (should be a simple HTTP call that adds the [COMMUNITY FEATURE](https://discord.com/developers/docs/resources/guild#guild-object-guild-features))
2. Check if the guild has a channel with the same name as `Channels.python_help`
2.a. If it does, and that channel is already a forum channel -> do nothing
2.b if it does, and that channel is a text channel -> update it
2.c If it does not, create a forum channel

3. Map the id of that channel to `.env.server`

shtlrs:
Closes #2466 

This now checks if the server is a community & updates it when it's not.
It also provides the announcements & rules channels upon patching the guild since they are needed.

Upgrading a channel from Text to Forum isn't possible, so this basically deletes the text channel if it finds it & makes a new one as a replacement.

### DIFF
--- a/botstrap.py
+++ b/botstrap.py
@@ -22,7 +22,6 @@ ANNOUNCEMENTS_CHANNEL_NAME = "announcements"
 RULES_CHANNEL_NAME = "rules"
 GUILD_FORUM_TYPE = 15
 
-
 if not BOT_TOKEN:
     message = (
         "Couldn't find the `BOT_TOKEN` environment variable. "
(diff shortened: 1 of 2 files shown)


botstrap.py
```diff
@@ -40,14 +46,71 @@ def __init__(self):
...
+        payload["rules_channel_id"] = rules_channel_id_
+        payload["public_updates_channel_id"] = announcements_channel_id_
+        client.patch(f"/guilds/{guild_id}", json=payload)
+
+
+def upgrade_server_to_community(guild_id: int | str, client: DiscordClient) -> None:
+    """
+    Transforms a server into a community one.
+
+    Return true if the server has been correctly upgraded, False otherwise.
```
vivekashok1221, line 78: This docstring doesn't match the function. What is the purpose of this function? It's not used anywhere in this PR.

vivekashok1221 (changes requested):
I haven't tested yet but looks fine upon review (except for the one comment I added).

I will try to test the code today.

botstrap.py
shtlrs: Oh... No it's an old version that just escaped my eyes.
It needs to go away indeed

shtlrs: 00330956f6140476d9e9f633b045921b214b3d37

vivekashok1221: Haha, I thought so. I'll approve after i get a chance to test the code.

vivekashok1221 (approved):
LGTM :star2:

botstrap.py
```diff
@@ -15,6 +15,12 @@
 BOT_TOKEN = os.getenv("BOT_TOKEN", None)
 GUILD_ID = os.getenv("GUILD_ID", None)
 
+COMMUNITY_FEATURE = "COMMUNITY"
+PYTHON_HELP_CHANNEL_NAME = _Channels.__fields__["python_help"].name
+ANNOUNCEMENTS_CHANNEL_NAME = _Channels.__fields__["announcements"].name
+RULES_CHANNEL_NAME = _Channels.__fields__["rules"].name
```
mbaruh, line 21: Maybe I'm missing something, but you're using the channel name to access the... channel name? seems circular.

botstrap.py
```diff
@@ -40,14 +46,61 @@ def __init__(self):
...
 
     @staticmethod
     def _raise_for_status(response: Response) -> None:
         response.raise_for_status()
 
 
+def upgrade_server_to_community_if_necessary(
+        guild_id: int | str,
+        rules_channel_id_: int | str,
+        announcements_channel_id_: int | str,
```
mbaruh, line 60: Just a nitpick, but here and in other places you're adding an underscore after "id". It's only necessary for variables named `id_` since `id` is a built-in.

mbaruh (commented):
Looks good, just a couple of comments

botstrap.py
shtlrs: Actually that was just because they're shadowing variables from the module's namespace and were just bugging me in Pycharm.

Maybe I chose poor names, do you want me to change them?

botstrap.py
shtlrs: You're not missing anything.
It's because I had imagined something different when I was starting up.

But they actually map the same value, so it's pointless.

I'll fix it

shtlrs: 6a782da04193b825220de2a9cefab1983351e134

botstrap.py
mbaruh: It's fine either way, I was just puzzled.

botstrap.py
```diff
@@ -126,6 +179,24 @@ def create_webhook(name: str, channel_id_: int, client: DiscordClient) -> str:
...
+    if PYTHON_HELP_CHANNEL_NAME in all_channels:
+        python_help_channel_id = all_channels[PYTHON_HELP_CHANNEL_NAME]
+        if not is_forum_channel(python_help_channel_id, discord_client):
+            delete_channel(python_help_channel_id, discord_client)
+        else:
+            create_help_channel = False
+
+    if create_help_channel:
+        python_help_channel_name = PYTHON_HELP_CHANNEL_NAME.replace('_', '-')
+        python_help_channel_id = create_forum_channel(python_help_channel_name, GUILD_ID, discord_client)
```
mbaruh, line 198: After this line, `all_channels` should be updated to use the new channel ID, otherwise you get:
image

mbaruh, line 198: Also, why are you not passing the category ID here? you already wrote the function to take one.

mbaruh (changes requested):

mbaruh (commented):
Also I would personally put all of the client-dependent functions inside the client class, but it's more of an opinion.

botstrap.py
```diff
@@ -40,14 +46,61 @@ def __init__(self):
...
     @staticmethod
     def _raise_for_status(response: Response) -> None:
         response.raise_for_status()
 
 
+def upgrade_server_to_community_if_necessary(
+        guild_id: int | str,
+        rules_channel_id_: int | str,
+        announcements_channel_id_: int | str,
+        client: DiscordClient) -> None:
```
mbaruh, line 61:
```suggestion
    guild_id: int | str,
    rules_channel_id_: int | str,
    announcements_channel_id_: int | str,
    client: DiscordClient
) -> None:
```

botstrap.py
```diff
@@ -40,14 +46,61 @@ def __init__(self):
...
+        payload["public_updates_channel_id"] = announcements_channel_id_
+        client.patch(f"/guilds/{guild_id}", json=payload)
+        log.info(f"Server {guild_id} has been successfully updated to a community.")
+
+
+def create_forum_channel(
+        channel_name_: str,
+        guild_id: str,
+        client: DiscordClient,
+        category_id_: int | None = None
```
mbaruh, line 79:
```suggestion
    channel_name_: str,
    guild_id: str,
    client: DiscordClient,
    category_id_: int | None = None
```

mbaruh (changes requested):

botstrap.py
```diff
@@ -126,6 +179,24 @@ def create_webhook(name: str, channel_id_: int, client: DiscordClient) -> str:
 
     config_str += "\n#Channels\n"
 
+    rules_channel_id = all_channels[RULES_CHANNEL_NAME]
+    announcements_channel_id = all_channels[ANNOUNCEMENTS_CHANNEL_NAME]
```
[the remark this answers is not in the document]
shtlrs: We've discussed that in [KeyError on webhooks when running botstrap #2488
](https://github.com/python-discord/bot/issues/2488).

That's mostly a usage error but I don't mind the idea of adding it.

botstrap.py
shtlrs: ae3db03f95eb157db2c1c11787e7c1725d7883e6

botstrap.py
shtlrs: > Also I would personally put all of the client-dependent functions inside the client class, but it's more of an opinion.

Yeah it's better that way, it's just that the original design wasn't intended to have a client like this & drifted away (and my thoughts with it as well :P)
ae3db03f95eb157db2c1c11787e7c1725d7883e6

shtlrs:
> Also I would personally put all of the client-dependent functions inside the client class, but it's more of an opinion.

Yeah it's better that way, it's just that the original design wasn't intended to have a client like this & drifted away (and my thoughts with it as well :P)
ae3db03f95eb157db2c1c11787e7c1725d7883e6

botstrap.py
shtlrs: The category wasn't created in the test server at first, and wasn't tracked.

Idk tbh, but here it is
e4903f80dc409d6ebee7170be8ee5e38c242d81d

shtlrs: 01c2fc14631dff8f84a3695b788317ccb0d02030

botstrap.py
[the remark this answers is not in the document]
mbaruh: It's definitely a usage error but we can still be nice :)
Also Hassan's comment was about this happening when migrating in old servers, which I didn't do: This issue is the result of trying to run it in a blank server.

botstrap.py
shtlrs: I don't know why pycharm does this ...

botstrap.py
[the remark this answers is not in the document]
shtlrs: How does b30475b1fbabbf970dfe1af4801152f70d917080 look ?

botstrap.py
```diff
@@ -33,85 +40,138 @@
...
+        self.name = name
+        super().__init__()
+
+    def __getitem__(self, item: str):
+        try:
+            super().__getitem__(item)
+        except KeyError:
+            log.warning(f"Couldn't find key: {item} in dict: {self.name} ")
+            log.warning("Please make sure to use our template: https://discord.new/zmHtscpYN9E3"
+                        "to make your own copy of the server to guarantee a successful run of botstrap ")
```
mbaruh, line 56: I think I would rather point to the contributing guide than to hardcode the template, but it's not a big deal.

botstrap.py
```diff
@@ -122,10 +182,29 @@ def create_webhook(name: str, channel_id_: int, client: DiscordClient) -> str:
...
     config_str += "\n#Channels\n"
 
+    rules_channel_id = all_channels[RULES_CHANNEL_NAME]
+    announcements_channel_id = all_channels[ANNOUNCEMENTS_CHANNEL_NAME]
+
+    discord_client.upgrade_server_to_community_if_necessary(rules_channel_id, announcements_channel_id)
+
+    create_help_channel = True
+
+    python_help_channel_id = all_channels[PYTHON_HELP_CHANNEL_NAME]
```
mbaruh, line 196: If the server contains a forum channel it won't be in the template, so you can't assume this channel exists. The way it was before was fine IMO.

shtlrs: Euh, idk what happened here tbh.

Oui, i need to bring it back.

botstrap.py
shtlrs: Good point, i will use that.

botstrap.py
shtlrs: dace921097e56dc39d6643f384d1ca08befc9560

botstrap.py
shtlrs: dace921097e56dc39d6643f384d1ca08befc9560

mbaruh (approved):

PR #2478, opened 2023-03-21
Enable sentry profiling

ChrisLovering:
only recording traces half of the time, to not impact performance too much.

### DIFF
--- a/bot/log.py
+++ b/bot/log.py
@@ -98,7 +98,11 @@ def setup_sentry() -> None:
             sentry_logging,
             RedisIntegration(),
         ],
-        release=f"bot@{constants.GIT_SHA}"
+        release=f"bot@{constants.GIT_SHA}",
+        traces_sample_rate=0.5,
+        _experiments={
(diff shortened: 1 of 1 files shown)


wookie184 (approved):
The [docs](https://docs.sentry.io/platforms/python/profiling/) say "The profiles_sample_rate setting is relative to the traces_sample_rate setting.", which I guess would mean we get 25% of profiling events with this?

I also wonder whether this will work properly without hooking into discord.py to create transactions e.g. for messages/commands (https://docs.sentry.io/platforms/python/performance/instrumentation/custom-instrumentation). From what can *guess* from the docs, it may be that this means that 50% of transactions are recorded, and 50% of those transactions are profiled, so if we have no transactions we have no profiling?

Approving this anyway as it should do no harm and merging this is probably the best way of finding out how it works

shenanigansd (approved):

Xithrius:
#2474 merged

PR #2487, opened 2023-03-24
Improve wording in botstrap error

wookie184:
Added a missing space at the start of a new sentence, and removed "likewise" as it's not the correct word in this context. Happy to bikeshed!


shtlrs (approved):
I'm fine with the new looks of it.
I would, when the guide is finished, add a reference to it maybe.

Thank you !

mbaruh (approved):

PR #2489, opened 2023-03-25
Remove usage of `site` config variable

shtlrs:
This defaults `site_paste` to our production pastebin as i don't think no one will ever host their own pastebin service. And this moves it to `BaseURLs`

Same goes for `site_logs_view`, however this one can be overriden if people want to access logs from their local sites. So it'll be included in the guide as an optional section to work with mod logs.

After the previously mentioned changes, `site` and `site_schema` aren't needed anymore, which is why they're removed.
`site_staff` isn't referenced anywhere as well  which is why it's removed.

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -527,8 +527,7 @@ class _BaseURLs(EnvConfig):
     github_bot_repo = "https://github.com/python-discord/bot"
 
     # Site
-    site_api = "site.default.svc.cluster.local/api"
-    site_api_scheme = "http://"
+    site_api = "http://site.default.svc.cluster.local/api"
     site_paste = "https://paste.pythondiscord.com"
 
 
@@ -546,6 +545,7 @@ class _URLs(_BaseURLs):
 
     paste_service: str = "".join([BaseURLs.site_paste, "/{key}"])
     site_logs_view: str = "https://pythondiscord.com/staff/bot/logs"
+
 
 URLs = _URLs()
 
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -527,10 +527,9 @@ class _BaseURLs(EnvConfig):
     github_bot_repo = "https://github.com/python-discord/bot"
 
     # Site
-    site = "pythondiscord.com"
-    site_schema = "https://"
     site_api = "site.default.svc.cluster.local/api"
     site_api_schema = "http://"
+    site_paste = "https://paste.pythondiscord.com"
 
 
 BaseURLs = _BaseURLs()
@@ -545,13 +544,8 @@ class _URLs(_BaseURLs):
     connect_max_retries = 3
     connect_cooldown = 5
 
-    site_staff: str = "".join([BaseURLs.site_schema, BaseURLs.site, "/staff"])
-    site_paste = "".join(["paste.", BaseURLs.site])
-
-    # Site endpoints
-    site_logs_view: str = "".join([BaseURLs.site_schema, BaseURLs.site, "/staff/bot/logs"])
-    paste_service: str = "".join([BaseURLs.site_schema, "paste.", BaseURLs.site, "/{key}"])
-
+    paste_service: str = "".join([BaseURLs.site_paste, "/{key}"])
+    site_logs_view: str = "https://pythondiscord.com/staff/bot/logs"
 
 URLs = _URLs()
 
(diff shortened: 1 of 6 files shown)


bot/exts/recruitment/talentpool/_api.py
```diff
@@ -29,8 +29,8 @@ class Nomination(BaseModel):
 class NominationAPI:
     """Abstraction of site API interaction for talentpool."""
 
-    def __init__(self, site_api: APIClient):
-        self.site_api = site_api
+    def __init__(self, site_api_client: APIClient):
```
wookie184, line 32: IMO the name was clear before so renaming it is unnecessary and will only cause conflicts.

shtlrs: How will it cause conflicts?

The reason I changed it is because when I was renaming the url value assigned to ''site_api'', this one got impacted as well and that's because it doesn't reflect exactly what it does since we clearly have different definitions of whay site_api is.
Plus, it's only a suffix I've added that reflects the class it's representing.

HassanAbouelela: I think this is a fine change, it improves clarity here. I don't think it's too big an issue either way.

bot/constants.py
```diff
@@ -527,10 +527,9 @@ class _BaseURLs(EnvConfig):
     github_bot_repo = "https://github.com/python-discord/bot"
 
     # Site
-    site = "pythondiscord.com"
-    site_schema = "https://"
     site_api = "site.default.svc.cluster.local/api"
-    site_api_schema = "http://"
+    site_api_scheme = "http://"
```
HassanAbouelela, line 531: Let's drop this scheme as well, and just move it directly into `site_api`.

HassanAbouelela (approved):
Good PR, one comment.

bot/exts/recruitment/talentpool/_api.py
wookie184: In general I would avoid renaming variables if not required because:
- It's not related to the PR (the name existed already), and slows down the PR if someone disagrees with the change.
- It causes merge conflicts with other PRs (I have one open I think).
- Variable names are personal preference, I prefer `site_api` which is why I used it when I wrote the code.

I don't think it improves clarity (`site_api.get` reads most fluently and clearly to me), so while I agree it's not a big issue, I don't think the change should be made.

shtlrs: It's not really a matter of personal opinion to me personally, more of a semantics thing.

But sure, it wasn't intentional to get here, it's just that refactoring the original `site_api` changed its value as well, which made me think that not having two same names for different things is worth changing.

I've reverted it here 153ca4a1b31024a170222a18e7f1c7772c6d258f

bot/constants.py
```diff
@@ -545,12 +543,8 @@ class _URLs(_BaseURLs):
...
     connect_cooldown = 5
 
-    site_staff: str = "".join([BaseURLs.site_schema, BaseURLs.site, "/staff"])
-    site_paste = "".join(["paste.", BaseURLs.site])
-
-    # Site endpoints
-    site_logs_view: str = "".join([BaseURLs.site_schema, BaseURLs.site, "/staff/bot/logs"])
-    paste_service: str = "".join([BaseURLs.site_schema, "paste.", BaseURLs.site, "/{key}"])
+    paste_service: str = "".join([BaseURLs.site_paste, "/{key}"])
+    site_logs_view: str = "https://pythondiscord.com/staff/bot/logs"
```
wookie184, line 547: I wonder if it would make sense to just have this as `site` or `site_public` (or similar) in BaseURLs with value `https://pythondiscord.com`, and then the endpoint can be specified in the place where it's used since that doesn't need to be configurable. With that we could use it for linking to the site for things other than logs in the future if we wanted.

wookie184 (approved):
Looks good, one comment, but not required so feel free to merge 👍

bot/constants.py
shtlrs: I'm fine with the idea, it's just that we need to be picky about the names since they expand real quick to become confusing.
We had it at `site` before, and it got people & me [confused](https://discord.com/channels/267624335836053506/1076619789939781775/1084897789739282462) when it came to usage.

I agree with you on the fact that it makes sense to have a base url, and **only** that needs to be configurable, as I also didn't like having to configure that long url if i ever wanted to use site logs.

However, for now, I'd like to keep it simple like this, as it's litteraly only needed for viewing logs, and if we run into the need of reusable, we'll think of a proper approach.

PR #2498, opened 2023-03-30
Eval output improvements

### LINKED ISSUE #2492: Eval output allows too many lines
# The Problem
The [`MAX_OUTPUT_BLOCK_LINES` variable](https://github.com/python-discord/bot/blob/main/bot/exts/utils/snekbox/_cog.py#L81) is set to a value of 10, which means under the [current logic](https://github.com/python-discord/bot/blob/main/bot/exts/utils/snekbox/_cog.py#L267-L278) evals which output more than **10** lines, will be uploaded to the haste site (paste.pydis.com) and a message will be added to the eval output stating it was truncated.

However, since we currently update the lines variable to be `lines[:max_lines+1]`, outputs of 11 lines (such as `for x in range(11): print(x)`) will be uploaded to the pastebin *and* still display all **11** lines (despite the limit being **10**). This also means that the paste content is the same as what's sent to Discord as the eval content.

The code should be changed to instead change lines to `lines[:max_lines]`. This will also break [the logic underneath](https://github.com/python-discord/bot/blob/main/bot/exts/utils/snekbox/_cog.py#L273-L278) (you've shortened so will never be more than `max_lines`) so the if statements will have to be swapped around.

# Current Code
```python
        if len(lines) > 1:
            if line_nums:
                lines = [f"{i:03d} | {line}" for i, line in enumerate(lines, 1)]
            lines = lines[:max_lines+1]  # Limiting to max+1 lines
            output = "\n".join(lines)

        if len(lines) > max_lines:
            truncated = True
            if len(output) >= max_chars:
                output = f"{output[:max_chars]}\n... (truncated - too long, too many lines)"
            else:
                output = f"{output}\n... (truncated - too many lines)"
```
# Proposed Solution (approx.)
```python
    if len(lines) > max_lines:
        truncated = True
        lines = lines[:max_lines]

    if len(lines) > 1:
        if line_nums:
            lines = [f"{i:03d} | {line}" for i, line in enumerate(lines, 1)]
        output = "\n".join(lines)
    
    if truncated:
        if len(output) > max_chars:
            output = f"{output[:max_chars]}\n... (truncated - too long, too many lines)"
        else:
            output = f"{output}\n... (truncated - too many lines)"
```
Notice that the proposed solution sets lines to `lines[:max_lines]` instead of `lines[:max_lines + 1]`. In reality the `if truncated` part can just be slotted into the [pre-existing `if truncated` block](https://github.com/python-discord/bot/blob/main/bot/exts/utils/snekbox/_cog.py#L283-L284)

TizzySaurus:
Closes #2492

## What
As well as correcting the logic for how many lines are displayed in the eval output (see #2492), this PR improves output truncation as follows:

- You can now know an individual line was shortened (but not completely removed) by the inline `...` at the end.
- If there are any completely removed lines, this is indicated by a `...` on a new line after the output.

## Examples:
NB: There have been some minor changes to the truncation logic since these screenshots to ensure the final output is <= `max_chars`, but the actual elipses and truncation messages are the same.
### One line that is too long
Before (not clear what was edited -- whether the line was shortened, or a line completely removed (or both)):
before-one-line-too-long
After (explicitly shows that no line was removed, and that the displayed line was shortened):
after-one-line-too-long
### One line that is too long with a truncated line
Before (shows that something changed, but it's not clear what, and it's not clear that a line was completely removed):
before-one-line-too-long-with-truncated-line
After (explicitly shows that a line was removed and the displayed line was shortened):
after-one-line-too-long-with-truncated-line
### Too many lines (the actual bugfix for #2492)
Before (displays `max_lines + 1` lines instead of `max_lines`):
before-too-many-lines
After (displays `max_lines` lines):
after-too-many-lines
### Too many lines and too long
Before (again not clear *what* was truncated):
before-too-many-lines-and-too-long
After (explicitly shows that a line was removed, and the final displayed line was shortened):
after-too-many-lines-and-too-long
### Multiple lines
Before (shows line number):
before-multiple-lines
After (still shows line number, so didn't break this):
after-multiple-lines
### Single line
Before (doesn't show line number, since there's only one line):
before-single-line
After (still doesn't show line number, so didn't break this):
after-single-line

NB: I'm not sure how much (if any) of the docker-compose changes should actually be part of this PR. Can someone please instruct on that.


wookie184:
This code seems quite complex, given this is balancing quite a lot of different aims/constraints I think we'd benefit from having some tests for this.

I'm also not sure the complexity is really worth it, a simple solution that replaces the current logic and fixes the linked issue would be something like this:
```python
def truncate(text, max_lines, max_chars):
    # Make the truncation message
    lines = text.split("\n")
    reason = []
    if len(text) > max_chars:
        reason.append("too long")
    if len(lines) > max_lines:
        reason.append("too many lines")
    end = f"\n... (truncated - {', '.join(reason)})" if reason else ""

    # Truncate text, leaving space for truncation message if necessary
    text = "\n".join(lines[:max_lines - (1 if end else 0)])
    text = text[:max_chars - len(end)]

    return text + end
```
Which meets the following aims:
1) Never have more than max_chars characters in the final output
2) Never have more than max_lines lines in the final output
3) Have as many lines/chars as possible in the final output

TizzySaurus:
@wookie184 
> 1. Never have more than max_chars characters in the final output
> 2. Never have more than max_lines lines in the final output
> 3. Have as many lines/chars as possible in the final output

For the record, none of these are fulfilled in the old code (that's currently in main). All three of these are fulfilled by my new code, so it does meet the necessary criteria.

> This code seems quite complex

I've added some images showing before vs after for different scenarios to the PR description to make the resulting differences (aside from the aforementioned criteria) clearer). I personally think the extra "complexity" is worthwhile. I've also made the code a bit cleaner since your comment.

> given this is balancing quite a lot of different aims/constraints I think we'd benefit from having some tests for this.

Tests can be added if we decide we want them.

wookie184:
I'm not sure what issue is being solved by having the `...` in different places. Is this solving something that has been an issue in reality? If you want to know what was truncated you can just look at the full output in the pastebin.

Requiring two lines (`...` and then `(truncated ...)`) takes up more space to indicate the same thing, and I think it's also a bit confusing not to have the truncation indicated in a consistent way.

Also, I think the truncated text should be included in the line limit. Consider if the maximum line length was 3, `print(*"abc", sep="\n")` would give
```
a
b
c
```
Currently (from what I can tell from your screenshots), `print(*"abcd", sep="\n")` would output
```
a
b
c
...
(truncated - too many lines)
```
which doesn't seem great, because the full output could have been displayed with one less line of text in total.

TizzySaurus:
> I'm not sure what issue is being solved by having the ... in different places.

It just makes it clear what was truncated *without* having to leave Discord and go to the paste site. It didn't feel like much extra work, so I did it 🤷.

> Requiring two lines (... and then (truncated ...)) takes up more space to indicate the same thing

It doesn't "indicate the same thing" though. It's possible for the truncation reason to be "too long", and to still have lines removed (e.g. `for _ in range(2): print("a" * (max_length // 2))`. Just from "too long" you wouldn't know that lines were removed, which is why I added the `...` line.

> I think it's also a bit confusing not to have the truncation indicated in a consistent way.

It is in a consistent way. The `...` tells you something was truncated. That's consistent across all scenarios.

> Also, I think the truncated text should be included in the line limit.

Yeah, I agree. Have updated the code accordingly.

mbaruh:
Having ellipses in multiple places seems confusing to me and I'm not sure if we have to find a way to address this. In the end the information we had before is that this isn't the full output and that you should open the paste link for the full output, which I think is sufficient.

TizzySaurus:
> @mbaruh Having ellipses in multiple places seems confusing to me

I really don't understand where this confusion comes from. `...` has long been used in the world of UI/UX to indicate that the displayed text has been shortened (and we already had this in the eval output before this PR).

What I've changed is that if the line is shortened, the `...` shows *after* the shortened line, showing it was simply *shortened*, and if the line is completely removed then the `...` shows *in-place* of the line, showing it was *completely removed*.

The `...` tells you *where* it was shortened, as is the job of `...` in the world of UI/UX. Where is the confusion here?

mbaruh:
`...` has long been used in UI yes, however you're taking that symbol and expanding it to potentially appearing several times at the same point, each time for a different reason.

I understand that this may seem obvious to the person who wrote the code, but I purposefully didn't look at the code when trying to understand the examples, and it wasn't obvious what's going on until I did. Your explanation in the latest comment is clear (although again I'm already "tainted" having already read the code), but you're not going to post it for every end-user trying to use snekbox right? This should be something that is immediately obvious without any prior knowledge.

Aside from having two ellipses in the same spot not looking great (in my opinion), I also think it's confusing. While I appreciate you going the extra mile I'd rather focus on addressing what's originally described in the issue.

Xithrius:

PR #2504, opened 2023-04-01
Ping help thread owner if thread closes without engagement.

swfarnsworth:
Add logic to see if anyone other than the OP (or bots) engaged in a help thread when it closes for inactivity. Currently only looks at the last 10 messages. If not, includes a ping to the OP before the embed in the closing message.


swfarnsworth:
image

image

swfarnsworth:
We might also have the message say "@OP it looks like your question wasn't answered. Try reading our guide on asking good questions."

bot/exts/help_channels/_channel.py
```diff
@@ -46,8 +46,19 @@ async def _close_help_post(closed_post: discord.Thread, closing_reason: _stats.C
...
     embed.set_author(name=f"{POST_TITLE} closed", icon_url=CLOSED_POST_ICON_URL)
-
-    await closed_post.send(embed=embed)
+    message = ''
+
+    # Include a ping in the close message if no one else engages, to encourage them
+    # to read the guide for asking better questions
+    if closing_reason == _stats.ClosingReason.INACTIVE and closed_post.owner is not None:
+        participant_ids = {
+            message.author.id async for message in closed_post.history(limit=10, oldest_first=False)
```
ChrisLovering, line 55:
```suggestion
            message.author.id async for message in closed_post.history(limit=100)
```
100 messages is the max that can be retrieved in 1 request. So we might as well do this to cover the case were the OP sent a lot of messages at the end of a session that does have other participants.

`oldest_first=False` is also the default if `after` isn't specified, so that can be removed.

bot/exts/help_channels/_channel.py
```diff
@@ -46,8 +46,19 @@ async def _close_help_post(closed_post: discord.Thread, closing_reason: _stats.C
...
+            if not message.author.bot
+        }
+        if participant_ids == {closed_post.owner_id}:
+            message += closed_post.owner.mention
```
ChrisLovering, line 59:
```suggestion
            message = closed_post.owner.mention
```
iadd isn't needed here since it's just the empty string anyway,

ChrisLovering (changes requested):
> We might also have the message say "@op it looks like your question wasn't answered. Try reading our guide on asking good questions."

I think just pinging like you are now is fine, so long as the embed message is good. If we think the embed message can be improved/condensed then I'm for it.

bot/exts/help_channels/_channel.py
swfarnsworth: I agree that we should up the limit to 100 if we can get all of that in one request, though I think that keeping `oldest_first=False` is worth it for self-documentation purposes.

swfarnsworth:
Note that this PR also changes the opening embed to ping the OP with a warning about pip installing malware.

image

This image just shows what it looks like; the text now reads "Do not pip install anything that isn't related to your question, especially if asked to over DMs."

MarkKoz (approved):
Don't have a strong opinion on whether we should or shouldn't ping, or what the message content should be.

However, code wise this change is nice and simple.

ChrisLovering (approved):

Issue #2009, opened 2021-12-13
Change behavior of `newlines` rule (pending the new filter system)


minalike:
Currently, the `newlines` rule triggers a ping to Moderators, deletes the offending message and also mutes the offending user. 

This is commonly triggered by people seeking help, often in help channels or topical channels.

Let's discuss whether we can:
- prevent bot from muting the user, instead have it reply with a reason for message deletion and a link to our pastebin service
- *not* have the user muted
- *not* ping Moderators 
- *not* relay to #mod-alerts 
- possibly still mute the user if they trigger `newlines` a certain number of times in a given timeframe?

If the above can be implemented, I imagine that discussions and help sessions can continue smoothly without a moderator having to step in to unmute the user, and invoke `!paste` with an explanation.

onerandomusername:
What's the current new line rule, how many lines are needed to trigger the filter? Perhaps it could be changed slightly, to not trigger so easily?

dawnofmidnight:
> What's the current new line rule, how many lines are needed to trigger the filter? Perhaps it could be changed slightly, to not trigger so easily?

I don't think that could be done without making the whole filter pointless, or at least less effective, as it's often over 100 newlines already (the times it's triggered that is, I don't know what the limit).

onerandomusername:
Ah, that makes sense.

dawnofmidnight:
I was able to find this, the limit is already 100 lines, so my point may be invalid. https://github.com/python-discord/bot/blob/a5cfe2ac47077805c205bcaf05eaed5ded5d09f9/config-default.yml#L439-L442

onerandomusername:
> I was able to find this, the limit is already 100 lines, so my point may be invalid.
> 
> https://github.com/python-discord/bot/blob/a5cfe2ac47077805c205bcaf05eaed5ded5d09f9/config-default.yml#L439-L442

Depending on how the messages trigger it, it may be possible to increase max, but keep max_consecutive the same.

onerandomusername:
So it seems like there are still some modifications, and I do understand this a *bit* more, but I think that increasing the max newlines a bit *might* help in the short term.

We could also add a bit more complexity, by calculating the max characters and a percentage of which of those characters were newlines vs whitespace.

minalike:
Example of an [instance](https://discord.com/channels/267624335836053506/696352302386380825/930480677177143298) where it took us 8 minutes to unmute someone that triggered the `newlines` rule while getting help in a help session.

[...]

PR #2505, opened 2023-04-02
Exclude help forum messages from the newlines filter

lxnn:
Just a thought :P

The newlines filter is pretty noisy in the help forum. The frequent false-positives need a lot of moderator attention, and aren't a great experience for new users. Excluding help forum messages from the newline count could be a reasonable stopgap until the new filtering system arrives.

Relevant issue: #2009


brodycritchlow (commented):
I think the bot should still mute if a person sends multiple newline infractions in a row, this way people can't just spam in threads.

mbaruh:
Solvable with the new system

PR #2506, opened 2023-04-02
Remove self-assignable helpers functionality

mbaruh:
Didn't work out unfortunately.


ChrisLovering (approved):

hedyhli (approved):
Kudos to whoever came up with that idea 

This really got me but last year’s PEP was extremely well done!

PR #2512, opened 2023-04-06
Include jump URL in mod log embed for infraction edits

### LINKED ISSUE #2495: Keep jump link in infraction edit logs
image

Currently when an infraction edit is logged the jump URL field is not mentioned.

I think it would be helpful to have it mentioned even on edit logs. I frequently check #mod-log after appending or editing an infraction and find myself looking for the jump link.

vivekashok1221:
Closes #2495 
Screenshot:
image


shtlrs:
It's looking good, I just have one comment before I submit a formal review:
When I edit an infraction twice, first in a `moderators`, then in a `ModMail` channel, I lose context to the first edit.

Which raises 2 questions:
1. Is that something that has happened before ? (Even if it didn't, it can still )
2. Do we want to cover that ?

vivekashok1221:
By context you mean the jump URL, right?
Through this PR, jump URL to the edit context is displayed in the edit log embed- but it is not stored in the database. When you use the `!infraction search`, only the original context is displayed. The assumption is that the context is more likely to be present where the infraction was originally invoked  (although if any offending messages are deleted then :shrug: ).

As of now, there is no indication in the `!infraction search` embed that the infraction reason, expiry, etc. has been edited.  We could/should add message link to  edit contexts to infraction details when/if work on https://github.com/python-discord/site/issues/752 starts.

So, 
>  I lose context to the first edit.
both the edit log embeds will still be in `#mod-log` channel, and the edit context information will not be stored in the database. It will only be included in the log embed that is sent to the #mod-log channel, and then discarded.

mbaruh (approved):
Thanks!

shtlrs (approved):
I had completely forgotten to get back to this.

Tested and approved! Thanks!

PR #2514, opened 2023-04-06
Use the right variables in antispam

mbaruh:


brad90four (approved):
These are good changes, thank you Zig

PR #2520, opened 2023-04-07
Handle off-topic channel names that are invalid for servers in Server Discovery

### LINKED ISSUE #2282: Handle Discord's rejection of off-topic channel names due to disallowed words
Sentry Issue: [BOT-38K](https://sentry.io/organizations/python-discord/issues/3626659695/?referrer=github_integration)

### Summary

Discord must have recently added a new feature which rejects channel names if they contain certain words they don't like. This is a problem for our automated renaming of off-topic channels, which has some names with words that are no longer allowed. The first event of the Sentry issue above was caused by the name "ot0-welcome-and-fuck-you", for example.

### Solutions

There are several ways to mitigate this:

1. Catch the error and keep trying new random channel names until one succeeds. If error code 50035 is too broad, then the code will have to match the error text. The rejected words could also be deleted from the database. Alternatively, we could be notified about rejections so we can rename them to comply with Discord's wishes. For example, replacing "fuck" with "frick".
2. When a new name gets added, attempt to create/rename a temporary channel to test that it's a valid name. Display an error to the command invoker if Discord rejects the name.
3. If we can somehow get and maintain an up-to-date list of disallowed words, check against that before adding new names to the database (and remove existing records which violate it). This would ideally be done by the API, in which case this issue should probably be moved to the python-discord/site.

Solution № 3 is ideal, but I don't think the list is publicly available. Thus, I'm leaning towards № 1. I'd rather not delete the rejected names because:

1. It would just suck to lose names that have accumulated over many years.
2. Discord may revert this feature (or Python Discord is no longer in Server Discovery, which is unlikely I suppose).

Alternative proposals are welcome.

### Other Affected Features

I believe the only other feature that creates/renames channels is the help system, but it uses fruit names, so I think we don't have to worry there. Modmail may suffer from this problem, but it's out of scope for this issue.

### Traceback

```
In name: Contains words not allowed for servers in Server Discovery.
  File "discord/ext/tasks/__init__.py", line 239, in _loop
    await self.coro(*args, **kwargs)
  File "bot/exts/fun/off_topic_names.py", line 65, in update_names
    await channel_0.edit(name=OTN_FORMATTER.format(number=0, name=channel_0_name))
  File "discord/channel.py", line 371, in edit
    payload = await self._edit(options, reason=reason)
  File "discord/abc.py", line 490, in _edit
    return await self._state.http.edit_channel(self.id, reason=reason, **options)
  File "discord/http.py", line 507, in request
    raise HTTPException(response, data)

Unhandled exception in internal background task "update_names".
```

### LINKED ISSUE #2500: Handle failure to edit OT channel names
Sentry Issue: [BOT-3F4](https://python-discord.sentry.io/issues/4050850439/?referrer=github_integration)

```
In name: Contains words not allowed for servers in Server Discovery.
  File "bot/exts/fun/off_topic_names.py", line 66, in update_names
    await channel_1.edit(name=OTN_FORMATTER.format(number=1, name=channel_1_name))

Unhandled exception in internal background task "update_names".
```
The blocked word was "swallow". I guess we should handle an excpetion here.

TizzySaurus:
### Related Issues
Closes #2282 
Closes #2500 

### What
This PR adds re-roll logic for when attempting to rename an off-topic channel to a name that's not allowed for servers in Server Discovery (such as `ot0-welcome-and-fuck-you`).

It will attempt to reroll a maximum of `MAX_RENAME_ATTEMPTS` per off-topic channel, and will send a message detailing any failed renames for a given off-topic channel to the internal mod-meta channel.

The code also deactivates names that are invalid, triggering a `log.info`, and adds logic for handling when the name pool isn't big enough (which triggers a `log.warn`).

### Testing
As for testing, I found that the following works:
```py
class Temp:
    def __init__(self, status, reason):
        self.status = status
        self.reason = reason

if (not channel_indx) or attempt < channel_indx:  # change as appropriate
    raise HTTPException(Temp("status", "reason"), {"code": 50035})
await channel.edit(name=name_after)
```
...and produces the following example message in the mod-meta channel:
Screenshot 2023-08-28 at 22 18 55
To forcefully run the coroutine you can use the `int eval` command, with the following code:
```py
# Reload Cog
await bot.reload_extension("bot.exts.fun.off_topic_names")

# Activate any deactivated off-topic names
names = await self.bot.api_client.get("bot/off-topic-channel-names", params={"active": "false"})
for name in names:
    await self.bot.api_client.patch(f"bot/off-topic-channel-names/{name}", data={"active": True})
    print(f"Activated otn name {name}")

# Call `update_names()` coroutine
cog = bot.get_cog("OffTopicNames")
await cog.update_names()
```


bot/exts/fun/off_topic_names.py
```diff
@@ -51,24 +59,129 @@ async def update_names(self) -> None:
...
+                except HTTPException as e:
+                    # We need to handle code 50035 ("invalid form body"),
+                    # which we get when the new channel name isn't allowed.
+                    #
+                    # For more information see https://github.com/python-discord/bot/issues/2500
+                    if (e.code != 50035):
+                        # The error isn't the one we want to handle so re-raise
+                        raise e
+
+                    failed_renames[channel_indx] = failed_renames.get(channel_indx, []) + [new_channel_name]
```
RohanJnr, line 116: You can use [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) here and set the default to a list.

RohanJnr (changes requested):
Tested and works fine. Gj

RohanJnr:
Also, the `reroll` command does not work if the channel name is not found in the database. This can be a super rare edge case but we can handle it. Under the `re_roll_command` function, you can catch the Exception on the `await self.de_activate_ot_name(ctx, old_ot_name)` and log it. After that the function will continue to rename the channel.

bot/exts/fun/off_topic_names.py
TizzySaurus: 5f3ffc77

TizzySaurus:
> Also, the `reroll` command does not work if the channel name is not found in the database. This can be a super rare edge case but we can handle it. Under the `re_roll_command` function, you can catch the Exception on the `await self.de_activate_ot_name(ctx, old_ot_name)` and log it. After that the function will continue to rename the channel.

That seems unrelated to this work, so should probably go in a separate PR?

RohanJnr:
> separate

Ye sure, I'll open an issue for that later

bot/exts/fun/off_topic_names.py
```diff
@@ -51,24 +60,129 @@ async def update_names(self) -> None:
...
+                except HTTPException as e:
+                    # We need to handle code 50035 ("invalid form body"),
+                    # which we get when the new channel name isn't allowed.
+                    #
+                    # For more information see https://github.com/python-discord/bot/issues/2500
+                    if (e.code != 50035):
+                        # The error isn't the one we want to handle so re-raise
+                        raise e
+
+                    failed_renames[channel_indx] += [new_channel_name]
```
RohanJnr, line 117: You can directly use append here
```suggestion
                    failed_renames[channel_indx].append(new_channel_name)
```

TizzySaurus: Good point, done in bf9f025

RohanJnr (approved):

bot/exts/fun/off_topic_names.py
```diff
@@ -3,9 +3,11 @@
 import difflib
```
shtlrs, line 3: Can you rebase onto main ? 
There are breaking changes since the `get_or_fetch_channel` isn't available anymore, you need to import it from `pydis_core`

bot/exts/fun/off_topic_names.py
```diff
@@ -51,24 +60,129 @@ async def update_names(self) -> None:
...
-        channel_0, channel_1, channel_2 = (self.bot.get_channel(channel_id) for channel_id in CHANNELS)
-
-        await channel_0.edit(name=OTN_FORMATTER.format(number=0, name=channel_0_name))
-        await channel_1.edit(name=OTN_FORMATTER.format(number=1, name=channel_1_name))
-        await channel_2.edit(name=OTN_FORMATTER.format(number=2, name=channel_2_name))
+        channels = [await get_or_fetch_channel(channel_id) for channel_id in CHANNELS]
+        failed_renames = defaultdict(list)
+        successfully_renamed = set()
+        for attempt in range(MAX_RENAME_ATTEMPTS):
+            # Get the necessary amount of new channel names
```
shtlrs, line 67: I don't think that all these comments are necessary, comments should exist to document things that don't jump straight to the eye.

bot/exts/fun/off_topic_names.py
```diff
@@ -51,24 +60,129 @@ async def update_names(self) -> None:
...
-
-        channel_0, channel_1, channel_2 = (self.bot.get_channel(channel_id) for channel_id in CHANNELS)
-
-        await channel_0.edit(name=OTN_FORMATTER.format(number=0, name=channel_0_name))
-        await channel_1.edit(name=OTN_FORMATTER.format(number=1, name=channel_1_name))
-        await channel_2.edit(name=OTN_FORMATTER.format(number=2, name=channel_2_name))
+        channels = [await get_or_fetch_channel(channel_id) for channel_id in CHANNELS]
+        failed_renames = defaultdict(list)
+        successfully_renamed = set()
+        for attempt in range(MAX_RENAME_ATTEMPTS):
```
shtlrs, line 66: I think the approach can be enhanced  a bit. 
We can fetch 3 names (number of attempts) for each channel.
We iterate over each channel, and try to rename it three times, then exit, that way we can increase the chance of a successful rename OP.

```py
number_of_names_to_fetch = MAX_ATTEMPS * len(CHANNELS)
channels = [get_or_fetch_channel(channel) for channel in CHANNELS]
names = iter(api_client.fetch_names(params={"random_items": number_of_names_to_fetch}))

renamed = set()
failed = set()

for channel in channels:
    attempt = 0
    while attempt < MAX_ATTEMPS:
        attempt += 1
        try:
            old_name = channel.name
            name = next(names)
            channel.update(name=name)
            renamed.add(channel.id)
            break
        except:
            # Catch your code here & deactivate the name
            # Reraise if not the right exception
            failed.add(f"- {name}\n") # Leverage the new markdown support to dsplay as a bullet list
            ...

# Simplify the handling of the failed renames.

if failed:
    unchanged = [channel.name for channel in channels if channel.id not in renamed]
    message = f"Couldn't rename the following off-topic channels: {', '.join(unchanged)}" \
              f"The following off-topic names were deactivated:\n{failed}", 
```

shtlrs (changes requested):

bot/exts/fun/off_topic_names.py
```diff
@@ -51,24 +60,129 @@ async def update_names(self) -> None:
...
+            # Get the necessary amount of new channel names
+            remaining_channels = set(channel.id for channel in channels if channel.id not in successfully_renamed)
+            try:
+                num_names_to_fetch = len(remaining_channels)
+                new_channel_names: list = await self.bot.api_client.get(
+                    "bot/off-topic-channel-names", params={"random_items": num_names_to_fetch}
+                )
+                if (num_fetched := len(new_channel_names)) < num_names_to_fetch:
```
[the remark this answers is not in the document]
TizzySaurus: It is strictly possible, however unlikely, so figured it should be accounted for - especially since the handling is relatively simple.

bot/exts/fun/off_topic_names.py
```diff
@@ -51,24 +60,129 @@ async def update_names(self) -> None:
...
-        await channel_2.edit(name=OTN_FORMATTER.format(number=2, name=channel_2_name))
+        channels = [await get_or_fetch_channel(channel_id) for channel_id in CHANNELS]
+        failed_renames = defaultdict(list)
+        successfully_renamed = set()
+        for attempt in range(MAX_RENAME_ATTEMPTS):
+            # Get the necessary amount of new channel names
+            remaining_channels = set(channel.id for channel in channels if channel.id not in successfully_renamed)
+            try:
+                num_names_to_fetch = len(remaining_channels)
+                new_channel_names: list = await self.bot.api_client.get(
```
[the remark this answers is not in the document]
TizzySaurus: c8f79a68

bot/exts/fun/off_topic_names.py
TizzySaurus: Fixed in e296bb56

bot/exts/fun/off_topic_names.py
TizzySaurus: I don't see any real harm in documenting how the code works given the complexity.

bot/exts/fun/off_topic_names.py
TizzySaurus: Rewritten in 2f738ad1. Currently untested, but lmk what you think :)

bot/exts/fun/off_topic_names.py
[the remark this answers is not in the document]
TizzySaurus: The handle for this has changed with the rewrite, but is still there [here](https://github.com/python-discord/bot/pull/2520/files#diff-5e63ba34922b26df25f0145df35b73a5d6816842ded3f0db0be1724ce979d575R76-R85). Also figured it made more sense to send a message to mod_meta than a `log.warn`, so updated to reflect that.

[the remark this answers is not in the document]
vivekashok1221: I agree with keeping the handle since it could save some of my brain cells when, say, I'm running my own instance of the bot in a test server.

bot/exts/fun/off_topic_names.py
shtlrs: That's what I'm saying, that the code can be understood by simply reading it, as the logic is kinda straightforward.

I personally don't think it's complex, we'll see what others have to say about it.

TizzySaurus: Yeah, in hindsight I agree -- these comments were removed as part of the rewrite.

bot/exts/fun/off_topic_names.py
```diff
@@ -51,24 +54,117 @@ async def update_names(self) -> None:
...
+                attempt += 1
+                try:
+                    new_channel_name = next(channel_name_pool)
+                except StopIteration:
+                    mod_meta = await get_or_fetch_channel(self.bot, Channels.mod_meta)
+                    await mod_meta.send(
+                        f":x: The pool of off-topic names ran out whilst attempting to rename {ot_channel.mention}.\n"
+                    )
+
+                    if deactivated_ot_names:
```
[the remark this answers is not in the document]
TizzySaurus: Good point, fixed in 5c16d697.

Xithrius:

PR #2522, opened 2023-04-07
Refine the invite code before filtering it

mbaruh:
Refines the invite to what it is most likely to be. If the refining process changes the invite code, then that means it originally wasn't a valid invite code. So the refining can find that the invite is actually whitelisted, or maybe it is explictly denied and that there's a specific action that needs to be taken.

For exmaple, for `discord.gg/python.` it will ignore the dot at the end.

Forward slashes (after discord.gg/) are included in the invite code because it can confuse Discord, so the behavior of blocking these invites stays unchanged for now.

### DIFF
--- a/bot/exts/filtering/_filter_lists/invite.py
+++ b/bot/exts/filtering/_filter_lists/invite.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 import typing
 
 from discord import Embed, Invite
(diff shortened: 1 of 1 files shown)


ChrisLovering (approved):

bot/exts/filtering/_filter_lists/invite.py
```diff
@@ -18,6 +19,13 @@
     from bot.exts.filtering.filtering import Filtering
 
 
+REFINED_INVITE_CODE = re.compile(
+    r"(?P<invite>[a-zA-Z0-9/]+)"  # The supposedly real invite code.
```
CaedenPH, line 23: Would it be possible to add a comment showing what an example invite code looks like? Difficult to understand what a real invite code looks like

mbaruh: It's really just a random combination of letters and numbers.

brad90four (approved):

bot/exts/filtering/_filter_lists/invite.py
```diff
@@ -18,6 +19,13 @@
...
+    r"(?:[^a-zA-Z0-9/].*)?"       # Ignoring anything that may come after an invalid character.
```
onerandomusername, line 24: That isn't quite how discord link parsing works.

PR #2524, opened 2023-04-10
Fix deleted message logging in mod alert

mbaruh:
The alerting only uploaded the related messages, so if they didn't come including the context message, it wasn't uploaded.


ChrisLovering (approved):

wookie184 (approved):
Looks good.

PR #2526, opened 2023-04-10
Redesign infractions embed

mbaruh:
Redesigns the infractions embed to be more concise and compact.

Before:
image
After:
image

The new embeds has a row of symbols that are explained in the help embeds of the appropriate commands:
image
I expect more symbols to be added in the future once the DB and bot can support them, such as the reason being edited, the infraction being manually pardoned (e.g due to an appeal), and the context link leading to a modmail log.


bot/exts/moderation/infraction/management.py
```diff
@@ -368,94 +391,96 @@ async def send_infraction_list(
...
-            remaining = time.until_expiration(expires_at)
-        else:
-            remaining = "Inactive"
+        symbols = []
+        if not infraction["hidden"] and infraction["dm_sent"] is False:
+            symbols.append(constants.Emojis.failmail)
+        if infraction["hidden"]:
+            symbols.append("🕵️")
+        if inserted_at != infraction["last_applied"]:
+            symbols.append("✏️")
```
ChrisLovering, line 432: Maybe we should have these as constants at the top of the file, used by both the SYMBOLS_GUIDE and here, so if we want to change a symbol it's just in one place

ChrisLovering (approved):

wookie184 (changes requested):
If you timeout and then untimeout a user, it is unclear that the timeout is no longer active in the infraction listing. 

I know for active timeouts it specifically says they are active, but seeing this:
image

still makes me think that it is active, due to the expiry.

wookie184 (approved):
Looks great! I'll let you merge

PR #2528, opened 2023-04-11
Remove dead or unused code

mbaruh:
* Removes the Free constants from two help system ago.
* Removes the mentions rule from the previous filtering system which probably survived a merge conflict.
* Removes the events extension which is now in https://github.com/python-discord/sir-robin.


wookie184 (approved):
Nice!

PR #2536, opened 2023-04-12
Include thread jump url in nomination archive message

### LINKED ISSUE #2494: Add link to archived nomination thread in nomination-voting-archive log embed
Currently after a vote is completed an embed gets sent to the log channel #nomination-voting-archive.

We can add a jump link to the embed linking to the archived discussion thread in #nomination-voting.

Reason: jump links are good.

wookie184:
Closes #2494

### DIFF
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -282,6 +282,7 @@ class Reviewer:
     async def archive_vote(self, message: PartialMessage, passed: bool) -> None:
         """Archive this vote to #nomination-archive."""
         message = await message.fetch()
+        thread = message.channel.get_thread(message.id)
 
         # We consider the first message in the nomination to contain the user ping, username#discrim, and fixed text
         messages = [message]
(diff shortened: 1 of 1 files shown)


bot/exts/recruitment/talentpool/_review.py
```diff
@@ -326,10 +327,12 @@ async def archive_vote(self, message: PartialMessage, passed: bool) -> None:
         result = f"**Passed** {Emojis.incident_actioned}" if passed else f"**Rejected** {Emojis.incident_unactioned}"
         colour = Colours.soft_green if passed else Colours.soft_red
         timestamp = datetime.now(tz=UTC).strftime("%Y/%m/%d")
+        thread_jump = f"[Jump to vote thread]({thread.jump_url})" if thread else "Failed to get thread"
```
mbaruh, line 330: You technically don't need to care about getting the thread or failing to do so. The thread URL is just the message URL without the message ID. I think it will be more consistent that way, because otherwise the thread might not be cached and then you have to do `await guild.fetch_channel`. This is unlikely to come up in testing since you're probably archiving recently created threads.

Alternatively, going further down the function there's a definition for `nomination_thread`, which you can just move up to above its new earliest usage (here).

mbaruh (changes requested):

bot/exts/recruitment/talentpool/_review.py
wookie184: Good point. I made it use the existing variable: https://github.com/python-discord/bot/pull/2536/commits/cf8409463390a2bdb03720bed912ff61a36fa4ae

mbaruh (approved):

shtlrs (approved):
Looks good, thank you.

Issue #1599, opened 2021-05-18
Add issue templates to the repo


Kronifer:
In sir-lancebot, there are currently issue templates for ease of use, and uniform issues. I suggest we either move the sir lancebot issue templates to the repo or create new ones.

dawnofmidnight:
I would like this too, since there have been many times I've had to go copy it from the Sir Lancebot repository. This would make it easier.

HassanAbouelela:
Duplicate of python-discord/organisation#268

Kronifer:
When is that?

HassanAbouelela:
There is currently nothing scheduled. @Xithrius, @Akarys42 thoughts?

Kronifer:
Since I do not have access to that issue, im assuming that it is for issue templates across repos

Akarys42:
I don’t think we reached a consensus on this, although here are my thoughts:

I’m not sure if locking ourselves in with a template issue is a good idea. I think having this freedom of structuring it as we want makes us more efficient and we can assume most of our users know how to write a good issue. It feels also quite bureaucratic.

It isn’t mentioned in the issue but a contrario I’d say copying the PR template over would be a good idea. You don’t need much freeway when writing PRs and it can force everyone to structure it well.

Kronifer:
I'd be fine with just going with the PR template, but I think at least having templates would be a good idea. We could provide a blank template for people to use as well.

Kronifer:
Whats the status on this?

Xithrius:
We're probably not going to add templates, as this repository is meant for more advanced users after coming from sir-lancebot.

Along with that, I agree that there should be no need for a format of PRs in terms of initial comments.

HassanAbouelela:
I've seen it brought up a few times now, and I disagree with the notion that "advanced" users don't find templates helpful.

Templates have a few benefits:
1. They guide beginners about what they should add,
but 2. They also guide any user about what they should add.

You can know what you should add, but still forget something. I also personally find it easier to give the required details, when it's more of a questionnaire than an open-ended response. I'd personally like to see issue templates in this repo as well.

I don't have as strong an opinion about PR templates.

Xithrius:
Should we have issue templates that are more simplified than sir-lance's?

HassanAbouelela:
There are a few things in lance's template that I would like changed, but besides those, I think having the templates be identical would be the easiest choice, both in terms of maintenance and user-friendliness.

Xithrius:
[...]

PR #2540, opened 2023-04-15
Add Tag Issue Template

HassanAbouelela:
People (contribs, staff, and core-devs alike) have often been confused by where to create issues for adding new tags. Creating issues on this repository makes the most sense - issue follows implementation - but it's rarely the correct option. Tag suggestions issues are never about implementation details, but specific tag details such as: do we want this? what form do we want it in? etc.

These questions can not be discussed adequately on this repository as it's difficult to gain traction and get people seeing it. The meta repository is a much better location for it, and hence the [tag suggestion template](https://github.com/python-discord/meta/blob/main/.github/ISSUE_TEMPLATE/tag-request.yaml) exists. This is an attempt at guiding people towards it.

### DIFF
--- a/.github/ISSUE_TEMPLATE/tag-suggestion.yml
+++ b/.github/ISSUE_TEMPLATE/tag-suggestion.yml
@@ -1,0 +1,17 @@
+name: Tag Suggestion
+description: Suggest a new tag.
+title: I DID NOT READ THE INSTRUCTIONS
+labels: []
+assignees: []
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for trying to improve the community! We do not take tag suggestions on this repository, however please submit your suggestion on the [meta](https://github.com/python-discord/meta) repostiory instead.
+  # Mandatory non-markdown body field.
+  - type: input
+    id: data
+    attributes:
+      label: Do not fill this out
+    validations:
+      required: false
--- a/.github/ISSUE_TEMPLATE/tag-suggestion.yml
+++ b/.github/ISSUE_TEMPLATE/tag-suggestion.yml
@@ -7,7 +7,7 @@ body:
   - type: markdown
     attributes:
       value: |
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/tag-suggestion.yml
```diff
@@ -0,0 +1,17 @@
+name: Tag Suggestion
+description: Suggest a new tag.
+title: I DID NOT READ THE INSTRUCTIONS
```
shtlrs, line 3: I would suggest making it clearer, as this currently sounds like an "inside joke". (The fact that people never read the instructions, so we made this)

For example
```suggestion
title: TAG ISSUES GO TO THE META REPO, PLEASE READ THE FOLLOWING INSTRUCTIONS
```

shtlrs (changes requested):
Great idea, just one tiny comment.

wookie184:
I think it would be neater to use `contact_links` here given that's what they're intended for. For examples, see https://github.com/PyCQA/flake8/issues/new/choose and https://github.com/facebook/react/issues/new/choose

wookie184:
Also, looking at https://github.com/facebook/react/issues/new/choose it seems the option to create a blank issue is pretty hidden away. If we want to add a template for this I think we'd have to add ones for normal usage too.

shtlrs:
> I think it would be neater to use `contact_links` here given that's what they're intended for. For examples, see https://github.com/PyCQA/flake8/issues/new/choose and https://github.com/facebook/react/issues/new/choose

Nice ! I didn't know these existed.
In that case, I agree with wookie on the fact that it would look neater if we use `contact_links` for that.

wookie184:
I'm going to close this as it's been stale for a while and I don't think it's close to merging. Given my second comment:

> Also, looking at https://github.com/facebook/react/issues/new/choose it seems the option to create a blank issue is pretty hidden away. If we want to add a template for this I think we'd have to add ones for normal usage too.

I think this would need to go along with https://github.com/python-discord/bot/issues/1599 so I'd consider this something that should be solved as part of that issue.

PR #2542, opened 2023-04-16
Don't collect coverage info by default when running tests

wookie184:
I'm pretty sure in 99% of cases when people are running tests they wont be checking coverage, since we don't take coverage too seriously.

Instead of having `poetry run task test-nocov` the option to not collect coverage info, this makes that default and adds a `poetry run task test-cov` option if coverage info is wanted.

For me this doubles the speed of `poetry run task test` (from ~20 seconds to ~10 seconds).


ChrisLovering (approved):

shtlrs (approved):
We sure like our tests sped up ! :shipit:

PR #2548, opened 2023-04-18
Remove !close command from helpers.

swfarnsworth:
Previously, helpers could use the !close command to put inactive help channels back into circulation. This is no longer necessary now that we use threads.

I confirmed that each of these cases work correctly on my test server:
* Close the thread as the OP with no role
* Not be able to close someone else's thread as a helper
* Close someone else's thread as admin


mbaruh (approved):

bot/constants.py
```diff
@@ -339,7 +339,7 @@ class _HelpChannels(EnvConfig):
     idle_minutes = 30
     deleted_idle_minutes = 5
     # Roles which are allowed to use the command which makes channels dormant
-    cmd_whitelist: list[int] = [Roles.helpers]
+    cmd_whitelist: list[int] = [Roles.moderators, Roles.admins]
```
ChrisLovering, line 342:
```suggestion
    cmd_whitelist = Guild.moderation_roles
```
We should use this instead, or better yet, remove this constant entirely and just use `Guild.moderation_roles` in the cog.

ChrisLovering (changes requested):

ChrisLovering (approved):

PR #2553, opened 2023-04-20
Track forgotten nominations

### LINKED ISSUE #2226: Create an admin task when votes are 2 weeks old

shtlrs:
closes #2226

The idea is to fetch active nominations that are more than 2 weeks old, and create admin issues for them in Github so that staff will do the follow up.

As soon as a vote is tracked, we add a "🎫" reaction (which can change) to it to make sure we don't create duplicate issues for it, instead of keeping track in another way

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -618,6 +618,17 @@ class _Icons(EnvConfig):
 Icons = _Icons()
 
 
+class _GithubModsRepository(EnvConfig):
+    section = "github_mods_"
+
+    owner: str = "python-discord"
+    name: str = "mods"
+    token: str
+
+
+GithubModsRepository = _GithubModsRepository()
+
+
 class _Keys(EnvConfig):
 
     EnvConfig.Config.env_prefix = "api_keys_"
--- a/bot/exts/recruitment/talentpool/_cog.py
+++ b/bot/exts/recruitment/talentpool/_cog.py
@@ -3,15 +3,19 @@ import textwrap
 from datetime import UTC, datetime
 from io import StringIO
 
+import arrow
 import discord
 from async_rediscache import RedisCache
 from discord import Color, Embed, Member, PartialMessage, RawReactionActionEvent, User
 from discord.ext import commands, tasks
 from discord.ext.commands import BadArgument, Cog, Context, group, has_any_role
+from discord.utils import snowflake_time
 from pydis_core.site_api import ResponseCodeError
 
 from bot.bot import Bot
-from bot.constants import Bot as BotConfig, Channels, Emojis, Guild, MODERATION_ROLES, Roles, STAFF_ROLES
+from bot.constants import (
+    Bot as BotConfig, Channels, Emojis, GithubModsRepository, Guild, MODERATION_ROLES, Roles, STAFF_ROLES
+)
 from bot.converters import MemberOrUser, UnambiguousMemberOrUser
 from bot.exts.recruitment.talentpool._api import Nomination, NominationAPI
 from bot.exts.recruitment.talentpool._review import Reviewer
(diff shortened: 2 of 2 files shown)


bot/constants.py
```diff
@@ -618,6 +618,17 @@ class _Icons(EnvConfig):
 Icons = _Icons()
 
 
+class _GithubModsRepository(EnvConfig):
+    section = "github_mods_"
+
+    owner: str = "python-discord"
+    name: str = "mods"
+    token: str
```
RohanJnr, line 626: This PR throws an error when the github token is missing (when it tries it get it from `.env` file as defined in constants). Maybe we can try making this optional. We also need to update the .env guide on the site https://www.pythondiscord.com/pages/guides/pydis-guides/contributing/bot/#appendix-full-env-file-options
```suggestion
    token: str = ""
```

shtlrs: Ha ! Yes.

It didn't before because we weren't using pydantic at that time.

Good catch, i will take care of it!

RohanJnr: we already have `API_KEYS_GITHUB` as optional, is the new `token` constant any different or does it use a different auth approach ?

shtlrs: Github introduced tokens per repo.

So this one will be granular in regards to the repo where tasks will be created

RohanJnr: ah okay, cool

bot/constants.py
```diff
@@ -618,6 +618,17 @@ class _Icons(EnvConfig):
 Icons = _Icons()
 
 
+class _GithubModsRepository(EnvConfig):
+    section = "github_mods_"
+
+    owner: str = "python-discord"
+    name: str = "mods"
```
vivekashok1221, line 625: I very vaguely remember the discussion behind this feature so I just want to double check whether the issue has to be raised on the `python-discord/mods` repo or `python-discord/admins` (?) repo.

Whenever a vote has passed the upvote threshold, the admins add a ticket to the admin tasks tracker** and not `python-discord/mods` repo- which is why I'm asking for clarification on which repo stale reviews should be posted.


**I'm not actually sure what the "admin tasks tracker" is, I always assumed it's a bunch of issues raised on `python-discord/admins` or something, similar to the `python-discord/meta` repo.

vivekashok1221 (changes requested):
This is purely a code review, will test later.

My "changes requested" are more of an attempt to make sure that I understand what's happening clearly.

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -652,6 +663,92 @@ async def _nomination_to_string(self, nomination: Nomination) -> str:
 
         return lines.strip()
 
+    @tasks.loop(hours=72)
+    async def track_forgotten_nominations(self) -> None:
+        """Track active nominations who are more than 2 weeks old."""
```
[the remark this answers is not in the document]
vivekashok1221: Or at least that's how I heard the phrase "active nominations" being used.

bot/constants.py
vivekashok1221: Okay, admin tasks tracker is the `admin-tasks` repo.

vivekashok1221: Yeah, the issue is to be raised on `python-discord/admin-tasks`, Chris just confirmed.

shtlrs:
Closing this as I don't think I will be finishing it.
Not sure mods/admins even need this or will be using it.

If anyone wants to have it, feel free

PR #2554, opened 2023-04-21
Use documented spec extensions in docker compose

ChrisLovering:
The previous method of '<<' no longer works with the latest version of docker compose. This method is what is documented in the [compose spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#extension)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,6 @@ repos:
       - id: check-merge-conflict
       - id: check-toml
       - id: check-yaml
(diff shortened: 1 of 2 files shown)


shtlrs (approved):
Tested on:
docker engine `20.10.23`
docker compose ` v2.15.1`

wookie184 (approved):
It seems like there is a version of the previous syntax that works, but this way seems just as good so lgtm. https://github.com/docker/compose/issues/10411#issuecomment-1488019350

PR #2556, opened 2023-04-22
Do not account for jump urls for voice channels

### LINKED ISSUE #2555: AttributeError: 'NoneType' object has no attribute 'jump_url'
Sentry Issue: [BOT-3FS](https://python-discord.sentry.io/issues/4115261439/?project=2724195&referrer=slack)

The issue seems to be happening at [this line](https://github.com/python-discord/bot/blob/7c8b8e15ae506457da907cc38b675de2b8a89494/bot/exts/moderation/infraction/_utils.py#L104)

I haven't looked much into the exception, but it seems to be induced by a `superstarify` infraction

shtlrs:
Closes #2555 

`on_voice_state_update` events do not have `message` attributes in their context. 

Which is why we'll ignore setting the `jump_url` for infractions triggered by these events.


bot/utils/channel.py
```diff
@@ -46,6 +46,13 @@ def is_in_category(channel: discord.TextChannel, category_id: int) -> bool:
     return getattr(channel, "category_id", None) == category_id
 
 
+def is_voice_channel(
+        channel: discord.TextChannel | discord.DMChannel | discord.Thread | discord.StageChannel
+) -> bool:
+    """Return True if `channel` is a voice one."""
+    return channel.type in [discord.ChannelType.voice, discord.ChannelType.stage_voice]
```
shtlrs, line 53: We might want to account for null channels that we ignore here.
In that case, I'm not sure what value to return 🤔

shtlrs: I guess we can just let it throw, and handle those case by case if they will ever occur.

bot/exts/moderation/infraction/_utils.py
```diff
@@ -98,7 +98,7 @@ async def post_infraction(
     if any(
         is_in_category(ctx.channel, category)
         for category in (Categories.modmail, Categories.appeals, Categories.appeals_2)
-    ):
+    ) or is_voice_channel(ctx.channel):
```
mbaruh, line 101: Wouldn't it make more sense to check if ctx.message is not None?

shtlrs: Yes it would, and that's the first obvious solution.

I just got lost in a tunnel vision of figuring out where it started from and following through.

I'll change is to that, as it's a much much simpler fix.

wookie184: I think we should consider (probably not in this PR) changing the type hints for any functions that can take a `FakeContext`, since just `Context` is misleading and makes it easy to introduce bugs like this.

shtlrs: > I think we should consider (probably not in this PR) changing the type hints for any functions that can take a `FakeContext`, since just `Context` is misleading and makes it easy to introduce bugs like this.

I think the issue now is that we have 3 context classes: `Context`, `FakeContext` and `FilterContext` that are being used for conversion, and `post_infraction` takes either `Context` or `FakeContext`.

In this particular case, I think it's an instance of `FilterContext` that's being converted to `FakeContext` at the level of the `InfractionAndNotification.action` method .

But I do agree that it might need to be cleaned up.

I'll try to tackle this next.

mbaruh: I'm open to a solution that gets rid of FakeContext, but as far as I could see it's a necessity since Context is created from either a message or an interaction.

mbaruh (approved):
Untested

wookie184 (approved):
Looks good

PR #2557, opened 2023-04-22
Map container ports to the host's

shtlrs:
This allows us to access site and other services directly from the host machine.

Which is useful for viewing pg data, or viewing content on site, etc.


ChrisLovering:
> This allows us to access site and other services directly from the host machine.
This behaviour is already possible with the current configuration. Holding this PR until we actually know what the problem is.

shtlrs:
Closing as this issue seems to only happen on my machine specifically.

PR #2566, opened 2023-04-28
Message content intent tag

Canttuchdiz:
Part of [site#695](https://github.com/python-discord/site/issues/695)

Includes the tag for the message content intent, describing use case and importance.


bot/resources/tags/mcintent.md
```diff
@@ -0,0 +1,16 @@
+---
+embed:
+    title: "Message Content Intent"
+---
+
+Discord regulates data being sent to an application via a specific number.
+
+This number represents different permissions A.K.A. intents.
```
Robin5605, line 8: I think the explanation of how the gateway intents bitfield is implemented is necessarily important here - we want to explain *how to do something* rather than *how something works*, if that makes sense. So something like
```suggestion
The discord gateway only dispatches events you subscribe to, which you can configure by using "intents"
```

bot/resources/tags/mcintent.md
```diff
@@ -0,0 +1,16 @@
...
+
+The message content intent is what determines if an app will receive the content of an ``on_message_create`` event.
```
Robin5605, line 10:
```suggestion
The message content intent is what determines if an app will receive the actual content of newly created messages
```

bot/resources/tags/mcintent.md
```diff
@@ -0,0 +1,16 @@
...
+
+Disabling or enabling this feature will allow or disallow this data to be received.
```
Robin5605, line 12: Superfluous after previous line

Robin5605 (changes requested):

bot/resources/tags/mcintent.md
```diff
@@ -0,0 +1,14 @@
+---
+embed:
+    title: "Message Content Intent"
+---
+
+The discord gateway only dispatches events you subscribe to, which you can configure by using "intents"
+
+The message content intent is what determines if an app will receive the actual content of newly created messages
+
+Disabling or enabling this feature will allow or disallow this data to be received.
```
Robin5605, line 10: This line is superflous
```suggestion

```

Robin5605 (changes requested):
Last few changes. I think the codeblock example is probably the most important, since it actually shows *how* users can enable message content intent themselves rather than knowing how it works conceptually

bot/resources/tags/mcintent.md
```diff
@@ -0,0 +1,14 @@
...
+
+Discord has disabled this permission for verified bots (bots verified over 75-100 servers).
+
+Users can request this permission for their bot.
```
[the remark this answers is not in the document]
Canttuchdiz: Fix codeblock.

bot/resources/tags/mcintent.md
```diff
@@ -0,0 +1,15 @@
+---
+embed:
+    title: "Message Content Intent"
+---
+
+The discord gateway only dispatches events you subscribe to, which you can configure by using "intents"
+
+The message content intent is what determines if an app will receive the actual content of newly created messages
```
Robin5605, line 8: I'd just like for this to tie back to the main issue of command handlers not running
```suggestion
The message content intent is what determines if an app will receive the actual content of newly created messages. Without this intent, discord.py wouldn't know which command handler to trigger so none of your commands will work.
```

Robin5605 (changes requested):

bot/resources/tags/mcintent.md
```diff
@@ -0,0 +1,15 @@
+---
+embed:
+    title: "Message Content Intent"
+---
+
+The discord gateway only dispatches events you subscribe to, which you can configure by using "intents"
```
onerandomusername, line 6:
```suggestion
The Discord gateway only dispatches events you subscribe to, which you can configure by using Gateway Intents.
```

n0Oo0Oo0b (approved):

bot/resources/tags/mcintent.md
```diff
@@ -0,0 +1,17 @@
+---
+embed:
+    title: "Message Content Intent"
```
wookie184, line 3:
```suggestion
    title: "Discord Bot Message Content Intent"
```

wookie184 (changes requested):
Could you rename the file to `message-content-intent.md` to make it clearer. You can add an alias so the tag can be invoked with the shorter version (see for example https://github.com/python-discord/bot/blob/main/bot/resources/tags/underscore.md)

Canttuchdiz:
> Could you rename the file to `message-content-intent.md` to make it clearer. You can add an alias so the tag can be invoked with the shorter version (see for example https://github.com/python-discord/bot/blob/main/bot/resources/tags/underscore.md)

bot/resources/tags/message-content-intent.md
```diff
@@ -0,0 +1,18 @@
+---
+aliases: ["mcintent, message_content, message_content_intent"]
```
Robin5605, line 2: Aliases need to be in the same format as Python lists, see [here](https://raw.githubusercontent.com/python-discord/bot/main/bot/resources/tags/f-strings.md) for an example.

Robin5605 (changes requested):

bot/resources/tags/message-content-intent.md
Canttuchdiz, line 2:
```suggestion
aliases: ["mcintent", "message_content", "message_content_intent"]
```

Robin5605 (approved):
Looks good. Thanks!

bot/resources/tags/message-content-intent.md
```diff
@@ -0,0 +1,18 @@
...
+The message content intent is what determines if an app will receive the actual content of newly created messages. Without this intent, discord.py won't be able to detect prefix commands, so prefix commands won't respond.
+
+Privileged intents, such as message content, have to be explicitly enabled from the [Discord Developer Portal](https://discord.com/developers/applications) in addition to being enabled in the code:
+
+```py
+intents = discord.Intents.default() # create a default Intents instance
+intents.message_content = True # enable message content intents
+
+bot = commands.Bot(command_prefix="!", intents=intents) # actually pass it into the constructor
+```
```
n0Oo0Oo0b, line 18: I think mentioning other relevant tags at the end would be helpful
````suggestion
```
For more information on intents, see `/tag intents`. If prefix commands are still not working, see `/tag on-message-event`.
````

n0Oo0Oo0b (changes requested):

bot/resources/tags/message-content-intent.md
n0Oo0Oo0b: Speaking of the intents tag, I've just noticed there is a lot of overlap between it and this, maybe we could differentiate the two a bit more?

Canttuchdiz: Thats a good point. I think the beginning is important to have, but maybe removing the part about the privilaged intents, and linking it would be useful. Also, incorporating the fact that you do have to enable it in the discord developer portal at the beginning. Does that sound good?

n0Oo0Oo0b: That sounds good to me. I think the intents tag can also be improved (primarily the length because it is very wordy at the moment) but it would be better addressed in a separate PR.

PR #2567, opened 2023-04-28
on-message-event tag

n0Oo0Oo0b:
Part of [site#695](https://github.com/python-discord/site/issues/695)

### DIFF
--- a/bot/resources/tags/on-message-event.md
+++ b/bot/resources/tags/on-message-event.md
@@ -1,6 +1,6 @@
 ---
 embed:
-    title: "The `on_message` event"
+    title: "The discord.py `on_message` event"
 ---
 
 Registering the `on_message` event with [`@bot.event`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.event) will override the default behavior of the event. This may cause prefix commands to stop working, because they rely on the default `on_message` event.
--- a/bot/resources/tags/on-message-event.md
+++ b/bot/resources/tags/on-message-event.md
@@ -3,7 +3,7 @@ embed:
     title: "The discord.py `on_message` event"
 ---
 
-Registering the `on_message` event with [`@bot.event`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.event) will override the default behavior of the event. This may cause prefix commands to stop working, because they rely on the default `on_message` event.
+Registering the `on_message` event with [`@bot.event`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.event) will override the default behavior of the event. This may cause prefix commands to stop working, because they rely on the default `on_message` event handler.
 
 Instead, use [`@bot.listen`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.listen) to add a listener. Listeners get added alongside the default `on_message` event, rather than overriding it, so prefix commands can still be invoked as usual:
 ```python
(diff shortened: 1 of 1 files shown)


bot/resources/tags/on-message-event.md
```diff
@@ -0,0 +1,26 @@
+---
+embed:
+    title: "The `on_message` event"
+---
+
+When defining an `on_message` event handler, prefix commands may stop working as it overrides the default behaviour of the `on_message` event.
+
+Instead, use [`@bot.listen`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.listen) to add a listener. Listeners get added alongside the default `on_message` event which allows you to respond to message events and prefix commands at the same time.
```
Canttuchdiz, line 8:
```suggestion
Instead, use [`@bot.listen`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.listen) to add a listener. Listeners get added alongside the default `on_message` event, preventing an override of on_message, and allowing prefix commands to still be invoked.
```

bot/resources/tags/on-message-event.md
```diff
@@ -0,0 +1,26 @@
+---
+embed:
+    title: "The `on_message` event"
+---
+
+When defining an `on_message` event handler, prefix commands may stop working as it overrides the default behaviour of the `on_message` event.
```
Canttuchdiz, line 6:
```suggestion
When registering the `on_message` event, prefix commands may stop working as it overrides the default behaviour of the `on_message` event.
```

wookie184 (changes requested):

bot/resources/tags/on-message-event.md
```diff
@@ -0,0 +1,26 @@
...
+You can also tell discord.py to process commands as usual after you're done processing messages with [`bot.process_commands()`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.process_commands). However, this method isn't recommended as it does not allow you to add multiple `on_message` handlers.
+
+```python
+@bot.event
+async def on_message(message):
+    ...  # do stuff here
+
+    await bot.process_commands(message)
+```
+
```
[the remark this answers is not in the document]
n0Oo0Oo0b: That's what I thought too, I was debating on whether I should delete that entire section but just deleting the example seems like a good choice

bot/resources/tags/on-message-event.md
```diff
@@ -0,0 +1,16 @@
+---
+embed:
+    title: "The discord.py `on_message` event"
+---
+
+Registering the `on_message` event with [`@bot.event`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.event) will override the default behavior of the event. This may cause prefix commands to stop working, because they rely on the default `on_message` event handler.
+
+Instead, use [`@bot.listen`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.listen) to add a listener. Listeners get added alongside the default `on_message` event, rather than overriding it, so prefix commands can still be invoked as usual:
```
Robin5605, line 8: I'd also like to include the fact that you can have more than one listener per event, unlike the event decorator
```suggestion
Instead, use [`@bot.listen`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.listen) to add a listener. Listeners get added alongside the default `on_message` event, rather than overriding it, so prefix commands can still be invoked as usual and you can have more than one handler for each event type (unlike events). Here's an example:
```

bot/resources/tags/on-message-event.md
```diff
@@ -0,0 +1,16 @@
...
+```python
+@bot.listen('on_message')
+async def message_listener(message):
```
Robin5605, line 11: Perhaps we can recommend the traditional usage of `@bot.listen()` where the function name is used as the event:
```suggestion
@bot.listen()
async def on_message(message):
```

Robin5605 (changes requested):

bot/resources/tags/on-message-event.md
Canttuchdiz: This seems a bit redundant at the end talking about having more than one handler for each event. The beginning of the sentence states listeners are added alongside. Though it isn't a big difference, and the information could be helpful.

bot/resources/tags/on-message-event.md
Canttuchdiz: I would provide examples of both implementation methods. 
```suggestion
# Method 1
@bot.listen('on_message')
async def message_listener(message):
    ...  # do stuff here

# Method 2
@bot.listen()
async def on_message(message):
```

bot/resources/tags/on-message-event.md
n0Oo0Oo0b: Maybe we could combine the first and last part?
"Listeners get added alongside the default `on_message` which means you can have multiple handlers for the same event. This means prefix commands can still be invoked as usual. Here's an example:"

Canttuchdiz: I think that is a good way of phrasing it.

wookie184 (approved):
Looks good to me, @Robin5605 would you like to take another look?

Robin5605 (approved):
I'm happy with this. Thanks!

PR #2573, opened 2023-05-01
Update filtering.py to reduce nickname alert spam

lxnn:
This updates the `on_voice_state_update` listener of the filtering cog.

Currently, it calls `self._check_bad_name` where it should call `self._check_bad_display_name`. The difference is that the latter method rate-limits how often a filtering alert is sent for a specific user's display name. Using the former method results in spammy alerts when a user with a bad display name is in a voice channel.


Xithrius (approved):
Although I did not test this, the code itself looks good to me.

wookie184 (approved):
Looks good

PR #2580, opened 2023-05-04
Handle adding reactions to messages from users who have blocked the bot

TizzySaurus:
Completes the bot-side of #2486.

## What
When the bot fails to add a reaction to a message because the user has the bot blocked (e.g. creating an incident, editing an eval invocation, etc.) the bot handles this via a `log.debug` statement, and if a message is passed to the handler then will also send a message to the message's channel in discord (which deletes after 30 seconds -- long enough to give the user a chance to unblock the bot from that message).

## Works For:

- [x] Commands
- [x] Events
- [x] Schedules

## Testing:
To test this, simply block the bot and then either create an incident, or edit an eval command (or anything else that causes the bot to react to your message).

## Example:
Screenshot 2023-05-04 at 23 39 44
Screenshot 2023-05-04 at 23 38 58


bot/bot.py
```diff
@@ -52,11 +56,27 @@ async def setup_hook(self) -> None:
 
     async def on_error(self, event: str, *args, **kwargs) -> None:
         """Log errors raised in event listeners rather than printing them to stderr."""
+        e_type, e_val, e_tb = exc_info()
```
wookie184, line 59: 3.11 has [`sys.exception`](https://docs.python.org/3/library/sys.html#sys.exception), which is nicer to use.

wookie184 (commented):
scheduling is something that is handled by bot-core, so the fix would need to be there. The error handler should probably be ported there first, so we don't have to duplicate this in multiple places. It might make sense to have central error handling that errors from everywhere are handled by, but i've not looked into it so i'm not too sure.

bot/bot.py
```diff
@@ -52,11 +56,27 @@ async def setup_hook(self) -> None:
...
         self.stats.incr(f"errors.event.{event}")
 
         with push_scope() as scope:
             scope.set_tag("event", event)
             scope.set_extra("args", args)
             scope.set_extra("kwargs", kwargs)
 
-            log.exception(f"Unhandled exception in {event}.")
+            formatted_exc = "\n".join(format_exception(e_type, e_val, e_tb))
+            log.exception(f"Unhandled exception in {event}: {formatted_exc}.")
```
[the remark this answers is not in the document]
TizzySaurus: f9054b0

bot/bot.py
TizzySaurus: f9054b0

TizzySaurus: f9054b0

TizzySaurus:
> scheduling is something that is handled by bot-core, so the fix would need to be there. The error handler should probably be ported there first, so we don't have to duplicate this in multiple places.

So are you suggesting to merge this PR as-is, and add the schedule handling to bot-core at a later time? I'd be okay with this, as long as the exceptions raised aren't going to cause sentry alerts etc.

bot/bot.py
```diff
@@ -52,6 +55,21 @@ async def setup_hook(self) -> None:
...
+        e_val = exception()
+
+        if isinstance(e_val, Forbidden):
+            event_to_message_indx = {
+                "on_message": 0,
+                "on_message_edit": 1
+            }
+            message = None
+            if (message_indx_in_args := event_to_message_indx.get(event)) is not None:
+                message = args[message_indx_in_args]
```
ChrisLovering, line 67:
```suggestion
            message = None
            if event in ("on_message", "on_message_edit"):
                message = args[0] if event == "on_message" else args[1]
```
imo this simpler, and much easier to read.

bot/utils/helpers.py
```diff
@@ -41,3 +47,19 @@ def remove_subdomain_from_url(url: str) -> str:
...
     parsed_url = parsed_url._replace(netloc=netloc)
     return parsed_url.geturl()
+
+
+async def try_handle_forbidden(error: Forbidden, message: Message | None = None) -> bool:
+    """Attempts to handle `discord.Forbidden` errors, returning `True` if successful else `False`."""
+    if error.code == 90001:
+        # Occurs when the bot attempted to add a reaction
+        # to a message from a user that has blocked the bot.
+        log.debug("Failed to add reaction to a message since the message author has blocked the bot")
```
ChrisLovering, line 57:
```suggestion
        log.info(
            "Failed to add reaction to message %d-%d since the message author (%d) has blocked the bot",
            message.channel.id,
            message.id,
            message.author.id,
        )
```
Info to make it appear in logs by default, as I think this would be useful. Also including a reference to the message so we can easily figure out where it came from

bot/utils/helpers.py
```diff
@@ -41,3 +47,19 @@ def remove_subdomain_from_url(url: str) -> str:
...
+        if message:
+            await message.channel.send(
+                f":x: {message.author.mention} failed to add reactions to your message as you've blocked me.",
```
ChrisLovering, line 60:
```suggestion
                f":x: {message.author.mention} failed to add reaction(s) to your message as you've blocked me.",
```

bot/utils/helpers.py
```diff
@@ -41,3 +47,19 @@ def remove_subdomain_from_url(url: str) -> str:
...
+                delete_after=30
+            )
+
+        return True
+    return False
```
ChrisLovering, line 65: As discussed in Discord. The return type will need to be updated too
```suggestion
    raise error
```

bot/exts/backend/error_handler.py
```diff
@@ -85,6 +86,11 @@ async def on_command_error(self, ctx: Context, e: errors.CommandError) -> None:
                 await ctx.send(f"{e.original} Please wait for it to finish and try again later.")
             elif isinstance(e.original, InvalidInfractedUserError):
                 await ctx.send(f"Cannot infract that user. {e.original.reason}")
+            elif isinstance(e.original, Forbidden):
+                if await try_handle_forbidden(e.original, ctx.message):
+                    # Forbidden error was handled so return
+                    return
+                await self.handle_unexpected_error(ctx, e.original)
```
ChrisLovering, line 93: As discussed in Discord.

```suggestion
            elif isinstance(e.original, Forbidden):
                try:
                    await try_handle_forbidden(e.original, ctx.message)
                except Forbidden:
                    await self.handle_unexpected_error(ctx, e.original)
```

bot/utils/helpers.py
```diff
@@ -41,3 +47,19 @@ def remove_subdomain_from_url(url: str) -> str:
     netloc = extracted_url.registered_domain
     parsed_url = parsed_url._replace(netloc=netloc)
     return parsed_url.geturl()
+
+
+async def try_handle_forbidden(error: Forbidden, message: Message | None = None) -> bool:
```
ChrisLovering, line 52: This function name, and docstring, should be updated to better reflect what it's doing. Something like `handle_forbidden_from_block`.

ChrisLovering (changes requested):

TizzySaurus:
For the record, @ChrisLovering and I have come up with a solution for handling errors in the schedules (see python-discord/bot-core#177).

bot/utils/helpers.py
ChrisLovering: If we ever find the need to do that then we can come up with a better name at that time, but for now it's needlessly mysterious to the caller as to what it does.

bot/bot.py
TizzySaurus: If we allow for this function to handle codes other than 90001, as was my plan, then we need to account for errors that may be raised from other events too though. Say, `on_guild_leave` or w/e. Which would mean this solution won't work (hence my dict and checking if it's `None`).

bot/utils/helpers.py
TizzySaurus: Fair enough

bot/bot.py
TizzySaurus: e4b2d5e

bot/utils/helpers.py
TizzySaurus: e4b2d5e

bot/utils/helpers.py
TizzySaurus: e4b2d5e

bot/exts/backend/error_handler.py
TizzySaurus: e4b2d5e

bot/utils/helpers.py
TizzySaurus: e4b2d5e

bot/utils/helpers.py
TizzySaurus: e4b2d5e

wookie184:
I don't understand how the bot-core PR would help. Would we have to change all uses of create_task to add an error handler? That doesn't seem great, compared to just handling it there directly.

TizzySaurus:
> I don't understand how the bot-core PR would help. Would we have to change all uses of create_task to add an error handler? That doesn't seem great, compared to just handling it there directly.

Yeah, you're right, in reality we'd want the `handle_forbidden_from_block` function to be in bot-core ,and automatically added in `create_task`. I'll make that change to the bot-core PR :+1:

TizzySaurus:
This is essentially stalled by python-discord/bot-core#177 now. Once that's merged, I can remove `handle_forbidden_from_block` from here, and migrate to importing from bot-core.

TizzySaurus:
python-discord/bot-core#177 has now been merged, so will be continuing on with this.

TizzySaurus:
This PR is now ready for review! :tada:

ChrisLovering (approved):

bot/bot.py
```diff
@@ -48,6 +51,19 @@ async def setup_hook(self) -> None:
...
+        if isinstance(e_val, Forbidden):
+            message = args[0] if event == "on_message" else args[1] if event == "on_message_edit" else None
+            try:
+                await handle_forbidden_from_block(e_val, message)
+            except Forbidden:
+                # Error wasn't handled, so handle below
+                pass
+            else:
+                # Error was handled, so return
+                return
```
Xithrius, line 65: I haven't used contextlib in a bit. Would it be acceptable to do something such as

```
    with contextlib.suppress(Forbidden):
        await handle_forbidden_from_block(e_val, message)
        return
```

wookie184: Yeah, I think that would be good.

The behaviour of `handle_forbidden_from_block` is pretty confusing, so I think it would be good to clear things up with a comment.
```python
    with contextlib.suppress(Forbidden):
        # Attempt to handle the error. This reraises the error if's not due to a block, 
        # in which case the error is suppressed and handled normally. Otherwise, it was
        # handled so return.
        await handle_forbidden_from_block(e_val, message)
        return
```

Xithrius: [56754a8](https://github.com/python-discord/bot/pull/2580/commits/56754a89c5e9ce3f0528c9abb01ef2b1b7e8be75)

wookie184 (approved):
Tested and seems to work nicely.

PR #2614, opened 2023-05-29
Show timestamp in pypi command

### LINKED ISSUE #2613: !pypi should include publish date in response
In many cases, users have been utilizing out of date packages. This would make the !pypi tag substantially more useful in visually parsing whether or not a package could be out of date for the current version of Python. 

https://github.com/python-discord/bot/blob/main/bot/exts/info/pypi.py#L48-L67

Example snippet: 

```py
ver = info["version"]
releasedate = info["release"][ver]["0"]["upload_time"]
embed.description += f"\n Release Date {releasedate}"
```

Should fall within the `if 200` response block. May want to consider prettying up the timestamp a bit to just the date portion.

Robin5605:
Closes #2613 

Show the upload date and time of the requested package in the timestamp field of the embed. Looks similar to this:
image

If there are suggestions to move the timestamp somewhere else within the embed, I'm open to hearing them.

### DIFF
--- a/bot/exts/info/pypi.py
+++ b/bot/exts/info/pypi.py
@@ -1,8 +1,11 @@
 import itertools
 import random
 import re
+import typing
 from contextlib import suppress
+from datetime import datetime
 
+import arrow
 from discord import Embed, NotFound
 from discord.ext.commands import Cog, Context, command
 from discord.utils import escape_markdown
@@ -22,6 +25,12 @@ INVALID_INPUT_DELETE_DELAY = RedirectOutput.delete_delay
 
 log = get_logger(__name__)
 
+def _get_latest_distribution_timestamp(data: dict[str, typing.Any]) -> datetime | None:
+    """Get upload time of last distribution, or `None` if no distributions were found."""
+    if not data["urls"]:
+        return None
+
+    return max(arrow.get(dist["upload_time"]) for dist in data["urls"]).datetime
 
 class PyPi(Cog):
     """Cog for getting information about PyPi packages."""
@@ -62,6 +71,10 @@ class PyPi(Cog):
                         embed.description = escape_markdown(summary)
                     else:
                         embed.description = "No summary provided."
+
+                    upload_time = _get_latest_distribution_timestamp(response_json)
+                    if upload_time:
+                        embed.timestamp = upload_time
 
                     error = False
 


n0Oo0Oo0b:
> If there are suggestions to move the timestamp somewhere else within the embed, I'm open to hearing them.

I think an embed field with title would make it clearer what the date means. This will also let us add more information (such as supported python versions and link to github repo, if they're part of the API response too) to the embed.

shtlrs:
I agree with @n0Oo0Oo0b here, it would be much better to indicate what that date represents here.

A range of the python supported versions is a good idea as well.

Xithrius:
Should we add the license that the package supports as well? I'm unsure how large we want the embed to be for any specific package.

For now, I think the original release date (when the package was created) and supported versions would be great to see.

n0Oo0Oo0b:
About displaying the supported python versions, I think displaying it as a range (e.g. `3.8 - 3.11`) would be nice, but the API only returns individual supported versions (under `response["info"]["classifiers"]`, prefixed with `"Programming Language :: Python ::"`), which would mean some extra processing has to be done. It should also account for packages that don't have support for a version in-between (e.g. 3.8, 3.9 and 3.11), but I think falling back to listing individual versions would be fine since it is quite rare.
There's also `response["info"]["requires_python"]`, but whether it contains any information seems inconsistent from my testing.

bot/exts/info/pypi.py
```diff
@@ -63,6 +72,10 @@ async def get_package_info(self, ctx: Context, package: str) -> None:
                     else:
                         embed.description = "No summary provided."
 
+                    upload_time = _get_latest_distribution_timestamp(response_json)
+                    if upload_time:
+                        embed.timestamp = upload_time
```
Xithrius, line 77: Getting the conversation going again, perhaps something simple like this would work:

```suggestion
                        embed.timestamp = f"Uploaded on {upload_time}"
```

shtlrs: Yes, I like this better, as it doesn't imply the embed's timestamp but rather the version's.

Xithrius:
Talked with Robin, consensus was to put this up for grabs such that the feature would be done faster.

JelmerKorten:
Hey.
Can I take this one?

shtlrs:
Sure, i've assigned you.

Xithrius:
Closing this in favor of #2924

PR #2621, opened 2023-05-31
Escape full stops in Filter.__str__

ChrisLovering:
Discord sees a number followed by a dot as a potential numbered list, so messes up the numbering when on sequential lines. Escaping the dot disables this discord behaviour.


shtlrs (approved):
Baguette au fromage très délicieuse.

PR #2633, opened 2023-06-10
Remove unnecessary pinning of nomination review message in #nomination-voting

vivekashok1221:
p.s: `bot.utils.messages.pin_no_system_message()` isn't used anywhere anymore but I've not removed it.

PR #2639, opened 2023-06-13
Fix missing message unpinning on adding ticket emoji to nomination review.

vivekashok1221:
This (d99310c391a251adb8245a2aa6220c43fa399323) is something I forgot to include, or rather did not think of, in my previous PR (#2633)- please forgive.

Trying to unpin a message that's not already pinned does not create any errors, though, from my testing.


Xithrius (approved):
You have been forgiven.

PR #2641, opened 2023-06-14
Increase help thread timeout 30 -> 60

swfarnsworth:
We decided after a staff meeting to have help threads time out after 60 minutes instead of 30, as the forum system allows us to have more concurrent help sessions than under the dynamic channel system.

image


vivekashok1221 (approved):
Looks good to me.

Xithrius (approved):
Yes.

PR #2652, opened 2023-06-21
Allow use of message reply feature to add additional context to incident reports

### LINKED ISSUE #2647: Use reply feature or a different prefix to avoid conflict with Discord markdown feature in #incidents
`#` is used to provide additional context to an incident report without creating a new ticket. However, due to the new Discord markdown feature,  the messages starting with `#` turns into heading text
# and the messages end up looking huge.

We could use a different prefix or, as suggested  by @import-pandas-as-numpy (sudorem on discord), use the message reply feature.

I personally prefer replacing `#` prefix with message replies.

vivekashok1221:
Closes #2647.
You can now include additional context to an incident reply by replying to a message and it will not be treated as a separate incident ticket.

I've not removed the previous method of adding additional context (iie, by prefixing `#`  to the message) for now to make transition easier (reporter muscle memory and all that).
Howeverm , IMO, removing it immediately will not inconvenience helpers much and  it may also assist in the discovery  of this change.

Screenshot:
image

### DIFF
--- a/bot/exts/moderation/incidents.py
+++ b/bot/exts/moderation/incidents.py
@@ -134,7 +134,8 @@ def is_incident(message: discord.Message) -> bool:
         message.channel.id == Channels.incidents,  # Message sent in #incidents
         not message.author.bot,                    # Not by a bot
         not message.content.startswith("#"),       # Doesn't start with a hash
-        not message.pinned,                        # And isn't header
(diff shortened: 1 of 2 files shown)


wookie184 (approved):
Nice, I think it's fine to still allow a `#` for now personally.

shtlrs (approved):

PR #2654, opened 2023-06-21
Remove casing in domain filtering

mbaruh:
The extraction of domains from a message in the filterlist removes casing, but the checks of individual filters kept theirs, which led to some filters not working.

This assumes that two URLs that differ only by casing are identical, which isn't necessarily true, but a conflict is very unlikely.


bot/exts/filtering/_filters/domain.py
```diff
@@ -40,7 +40,7 @@ async def triggered_on(self, ctx: FilterContext) -> bool:
 
         for found_url in ctx.content:
             extract = tldextract.extract(found_url)
-            if self.content in found_url and extract.registered_domain == domain:
+            if self.content.lower() in found_url and extract.registered_domain == domain:
```
wookie184, line 43: Should `domain`  be lowercased too, since it's derived from `self.content`?

mbaruh: Yeah you're right 743f2ce

wookie184 (approved):
LGTM

ChrisLovering (approved):

PR #2662, opened 2023-07-03
Optimise startup time

ChrisLovering:
The 4 cogs that I have modified here were taking ~1 second to load. As only so many tasks can run at once, this slowed down startup time by a decent amount.

I approached this by loading each extension one at a time during startup and picking out the worst offenders.


jchristgit (approved):
It's great.

wookie184 (approved):
Nice, looks good.

PR #2663, opened 2023-07-04
Additional potential symbols in valid invites

mbaruh:
Add underscores and dashes as valid symbols in invite codes.


minalike (approved):

vivekashok1221 (approved):

wookie184 (approved):
Looks good, had to double check that `-` wouldn't be interpreted as a range but seems to work fine since it's at the end.

PR #2669, opened 2023-07-05
Phishing button

mbaruh:
Adds a button to the mod alert when there's a chance this is a phishing attempt. Pressing the button will compban the user and add the appropriate content as an autoban filter.

Additionally, when adding an invite filter, if the given content isn't in the discord.gg/<invite> pattern, it will instead take it as the invite code itself as long as it doesn't contain spaces.


bot/exts/filtering/_ui/ui.py
```diff
@@ -531,12 +532,91 @@ async def cancel(self, interaction: Interaction, button: discord.ui.Button) -> N
...
+    async def confirm(self, interaction: Interaction, button: discord.ui.Button) -> None:
+        """Auto-ban the user and add the phishing content to the appropriate filter list as an auto-ban filter."""
+        await interaction.response.edit_message(view=None)
+
+        if self.offender:
+            compban_command = bot.instance.get_command("compban")
+            if not compban_command:
+                await interaction.followup.send(':warning: Could not find the command "compban".')
+            else:
+                ctx = FakeContext(interaction.message, interaction.channel, compban_command)
```
wookie184, line 561: This makes the actor of the ban the bot, but I think it should be the moderator instead.

wookie184 (changes requested):

bot/exts/filtering/_ui/ui.py
mbaruh: Good call fa4f233

bot/exts/filtering/_ui/ui.py
```diff
@@ -531,12 +532,91 @@ async def cancel(self, interaction: Interaction, button: discord.ui.Button) -> N
...
+    If confirmed, comp-ban the offending user, and add the appropriate domain or invite as an auto-ban filter.
+    """
+
+    def __init__(self, offender: User | Member | None, phishing_content: str, target_filter_list: FilterList):
+        super().__init__(emoji="🎣")
+        self.offender = offender
+        self.phishing_content = phishing_content
+        self.target_filter_list = target_filter_list
+
+    @lock_arg("phishing", "interaction", lambda interaction: interaction.message.id)
```
ChrisLovering, line 593: lambda 😢 
```suggestion
    @lock_arg("phishing", "interaction", attrgetter("message.id"))
```
Needs `from operator import attrgetter`

ChrisLovering (approved):
LGTM, 1 nit pick but don't care if it is done or not

bot/exts/filtering/_ui/ui.py
mbaruh: lambda good

bot/exts/filtering/_ui/ui.py
```diff
@@ -531,12 +532,91 @@ async def cancel(self, interaction: Interaction, button: discord.ui.Button) -> N
...
+                ctx = FakeContext(interaction.message, interaction.channel, compban_command, author=self.mod)
+                await compban_command(ctx, self.offender)
+
+        compf_command = bot.instance.get_command("compfilter")
+        if not compf_command:
+            message = ':warning: Could not find the command "compfilter".'
+            await interaction.followup.send(message)
+        else:
+            ctx = FakeContext(interaction.message, interaction.channel, compf_command)
+            await compf_command(ctx, self.target_filter_list.name, self.phishing_content)
```
wookie184, line 570: It would probably be unlikely in practice, but if the invite is not valid this will raise a `BadArgument` error which isn't handled. We could probably add a global error handler for interaction callbacks at some point, though i'm not sure exactly how that would work.

Probably fine to just catch the `BadArgument` here for now and tell the user it failed.

wookie184 (changes requested):
One minor comment, other than that all looks good.

bot/exts/filtering/_ui/ui.py
mbaruh: c998571

ChrisLovering (approved):

wookie184 (approved):
👍

PR #2677, opened 2023-07-10
Use utils already available in bot-core

shtlrs:
> Note: This relies on [this](https://github.com/python-discord/bot-core/pull/187) bot-core PR.

We have a couple of utils that are already available in `bot-core` like `AsyncCache` , etc.

This PR uses the ones provided by bot-core & deletes the duplicate code in `bot`.

This is a first step in cleaning up our projects & porting the commun utils in `bot-core`.

The commits are atomic, so you can view them one by one.


wookie184:
The linter changes shouldn't be necessary if you have the latest version of `ruff` installed (and they are not correct as they change the behaviour of the converters, see https://github.com/python-discord/bot/pull/2651#discussion_r1237109667).

shtlrs:
> The linter changes shouldn't be necessary if you have the latest version of `ruff` installed (and they are not correct as they change the behaviour of the converters, see [#2651 (comment)](https://github.com/python-discord/bot/pull/2651#discussion_r1237109667)).

Good catch red ranger, I've undone it.

wookie184 (changes requested):
The [`get_or_fetch_channel` in `pydis_core`](https://bot-core.pythondiscord.com/main/output/pydis_core.utils.channel.html#pydis_core.utils.channel.get_or_fetch_channel) requires passing the bot instance to the function, which the implementation here does not. I assume the usages will need changing.

Also the lock file needs updating.

shtlrs:
> The [`get_or_fetch_channel` in `pydis_core`](https://bot-core.pythondiscord.com/main/output/pydis_core.utils.channel.html#pydis_core.utils.channel.get_or_fetch_channel) requires passing the bot instance to the function, which the implementation here does not. I assume the usages will need changing.
> 
> Also the lock file needs updating.

Well spotted, thanks ! It's been dealt with in 2a6119854bcc6a229cd9c284bbcb280d78d1521d

Also, the bot-core version change has been reverted in favor of waiting for Chris' latest release where I'll just include the new paste service as well (that is if Chris didn't do it before)

shtlrs:
Closed in favor of #2685

PR #2682, opened 2023-07-14
Use `LinePaginator` from `bot-core`

shtlrs:
This works in conjunction with [bot-core#189](https://github.com/python-discord/bot-core/pull/189)

It serves as an example of how the new client code consuming bot-core's `LinePaginator` would look like.


bot/constants.py
```diff
@@ -702,3 +703,5 @@ class _Keys(EnvConfig):
     "Noooooo!!",
     "I can't believe you've done this",
 ]
+
+PaginationEmojis = PaginationEmojisSettings()
```
mbaruh, line 707: If this is using defaults, wouldn't it be better for the paginator to use those settings by default? Meaning, you wouldn't need to explicitly send it settings for it to work, only when you want to use something custom. This will save another import and function arg every time you want to you the paginator.

shtlrs: The entire problem lies within the trashcan. It's the thing that causes problems for contribs, since that emoji is only available in our server.

So if you don't pass those to the paginator when you're running bot, the paginator will crash since you can't use that emoji.

ChrisLovering: If that's the case, it might be better to just use https://emojipedia.org/wastebasket/ instead, to avoid all those problems.

shtlrs: You mean use it in our server as well ?
Tbh, I didn't consider that as an option, as I thought it was put in place on purpose.

If we were to use that, then we wouldn't be needing the pydnatinc model, import & extra parameter indeed.

ChrisLovering: Yea, not sure why we decided on the custom emoji instead of this unicode one. We also have the option to use https://emojipedia.org/stop-sign/ too. Using a unicode emoji would remove a lot of hassle.

shtlrs: Actually, now that I think of it, not only is it going to remove that hassle. But also, when it comes to making the `ErrorHandler` Cog part of bot-core, it will remove alot of not so pretty instantiation that would've been necessary to load it.

I'm all in for removing it, a waste basket or a stop sign will do the same thing. We might want to stick with the waste basket though as it's the one that resembles what we have already the most. The stop sign implies that you can't interact with the embed anymore, whereas the basket would imply better the deletion of the embed.

wookie184: I would be against changing the emoji, as it's there since it looks better than the unicode bin. I'm also not a fan of needing to pass in the constants everywhere they're needed. 

One simple way around this for now would be to just create a wrapper here that passes in the necessary constants and we can just use that instead. In the future we may want to think of other ways we can have configuration for bot-core.

shtlrs: Yeah the Emoji cannot change since it has been set to the custom one for legacy purposes.

However, regarding the Emojis as parameter part, I'm also not a biggest fan but that was one of the solutions.

I don't understand why is the `paginate` method a static method in the first place.

The way I'd want to do this is to instantie the Paginator in the parent package's `__init__.py` module and have it as a singleton. The constructor would take in the Emojis that need to be used the entire lifetime of the paginator, and we'd conserve the previously mentioned method's signature.

wookie184: `paginate` is a classmethod, and it's there because it needs to run async code which can't be done in the `__init__`.

It is still a bit strange since it doesn't actually return the class, so it would probably make sense to just expose `paginate` as a function by itself which creates a class internally. The current class cannot just be made into a singleton though as it contains state specific to an individual pagination.

shtlrs: I meant class method *

Well, i don't remember what it does specifically now but I'll see what I can do about it

shtlrs: @wookie184 

Hey ! I'm trying to pick this up again and I was wondering what you meant exactly by 
> One simple way around this for now would be to just create a wrapper here that passes in the necessary constants and we can just use that instead. In the future we may want to think of other ways we can have configuration for bot-core.

Do you mean that we'd keep the pagination emojis as part of the parameters, and then make a decorator that will pass them for us ?
Something like

```py

def pass_parameters(**kwargs_to_pass):
    def inner(func):
         def wrapper(*args, **kwargs):
              kwargs.update(kwargs_to_pass)
              return func(*args, **kwargs)
          retun wrapper
    return inner

``` 

Then we'd decorate the `paginate` method and pass the emojis using the previous deco ?

shtlrs: Ah, you meant that kind of wrapper.
Okay, I see.

>  Not sure if there's a nice way around that other than just copying the parameter list.
Don't think there is, even with `PEP 692` it wouldn't be quite feasible.

I think just copying the parameter list is fine, this will be used by us only, and anyone who wishes to not used the default emojis we provide can fallback to `_LinePaginator`

shtlrs:
Superseded by #2879

bot/constants.py
shtlrs: @wookie184 Implemented in #2879

PR #2685, opened 2023-07-19
Bump pydis_core to v10

shtlrs:
* Bumps pydis_core to 10.1.0
* Uses the utils already available in pydis_core instead of duplicating them here.


ChrisLovering (approved):
LGTM.

Now that bot-core can support multiple files, we should update the snekbox cog, in another PR, so that it uploads the text file outputs alongside all of the other outputs, rather than creating many pastefiles

bot/exts/moderation/dm_relay.py
```diff
@@ -53,14 +53,14 @@ async def dmrelay(self, ctx: Context, user: discord.User, limit: int = 100) -> N
             f"User: {user} ({user.id})\n"
             f"Channel ID: {user.dm_channel.id}\n\n"
         )
+        file = PasteFile(content=metadata+output, lexer="text")
```
wookie184, line 56: Should have spaces around operators as per PEP 8
```suggestion
        file = PasteFile(content=metadata + output, lexer="text")
```

wookie184 (changes requested):
One nit

wookie184 (approved):

PR #2687, opened 2023-07-20
Remove unused dependencies.

shtlrs:
We have a couple of unused dependencies that dependabot is trying to bump.

This is useless & a waste of time and effort.

I've searched for the import names of the removed packages.

Then I found [deptry](https://github.com/fpgmaas/deptry), which I later used to confirm my findings.


pyproject.toml
```diff
@@ -14,17 +14,14 @@ pydis_core = { version = "9.9.2", extras = ["async-rediscache"] }
 aiohttp = "3.8.4"
 arrow = "1.2.3"
 beautifulsoup4 = "4.12.2"
-colorama = { version = "0.4.6", markers = "sys_platform == 'win32'" }
```
ChrisLovering, line 17: This is needed by `coloredlogs` for windows coloured logging. https://coloredlogs.readthedocs.io/en/latest/readme.html?highlight=colorama#optional-dependencies

pyproject.toml
```diff
@@ -14,17 +14,14 @@ pydis_core = { version = "9.9.2", extras = ["async-rediscache"] }
...
 coloredlogs = "15.0.1"
 deepdiff = "6.3.0"
 emoji = "2.6.0"
 feedparser = "6.0.10"
-lxml = "4.9.3"
```
ChrisLovering, line 22: We use lxml as the parser for parsing xml files using `BeautifulSoup` in two places https://github.com/search?q=repo%3Apython-discord%2Fbot%20lxml&type=code

pyproject.toml
```diff
@@ -14,17 +14,14 @@ pydis_core = { version = "9.9.2", extras = ["async-rediscache"] }
...
 markdownify = "0.11.6"
 more-itertools = "9.1.0"
 python-dateutil = "2.8.2"
 python-frontmatter = "1.0.0"
-pyyaml = "6.0"
```
ChrisLovering, line 27: Yup, this is good to remove now that we don't use yml for constants,

ChrisLovering (changes requested):

pyproject.toml
shtlrs: Ok, I  saw that one. I just didn't think that it would be an extension that required a package.

pyproject.toml
ChrisLovering: Yea I think we've got it this way around as we only want it installed on windows hosts, hence the markers, rather than marking it as an extra on colored logs so it installs for everyone

pyproject.toml
ChrisLovering: Yea, it requires you to choose the parser and install separately. https://www.crummy.com/software/BeautifulSoup/bs4/doc/#installing-a-parser

shtlrs: Fixed in the forcepush. 9743c5b

pyproject.toml
shtlrs: Understood.
I've restored them both.

Xithrius (approved):

PR #2695, opened 2023-07-25
Migrate to Pydantic v2

ChrisLovering:
### Summary
All commits starting with `breaking:` are breaking changes from Pydantic and will be squashed into 1 commit before merging. (Hence the do not merge tag)

Other commits are kaizen style commits that I noticed while working on this PR.

This will be far easier to review commit-by-commit.

### Todo:
Squash commits before merging

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -8,7 +8,7 @@ By default, the values defined in the classes are used, these can be overridden 
 import os
 from enum import Enum
 
-from pydantic import BaseModel, model_validator
+from pydantic import BaseModel
 from pydantic_settings import BaseSettings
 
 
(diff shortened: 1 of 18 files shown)


shtlrs:
image

ChrisLovering:
> > Required investigation
> > I have not investigated if @field_validator("_", mode="before") works the same as @validator("_", pre=True) as far as the "*" is concerned to run the validator on all fields. relevant commit: [46bf3cc](https://github.com/python-discord/bot/commit/46bf3cc3daca24981f915f6518702a9b00b4c0af)
> 
> According to their docs, it looks like a wildcard will do the job. image

Ah nice, thanks.

wookie184 (approved):
🚀

shtlrs (approved):
Great work.

Would you like to do the same thing as 3a279b3d3faa679684a5b1bb5abda925abf16c38 for `Colours` and `Emojis`, I don't think these are things that anyone will ever configure, i don't even see the use of configuring them. (Might as well look into other classes that we ourselves do not quite setup)

ChrisLovering:
> Great work.
> 
> Would you like to do the same thing as [3a279b3](https://github.com/python-discord/bot/commit/3a279b3d3faa679684a5b1bb5abda925abf16c38) for `Colours` and `Emojis`, I don't think these are things that anyone will ever configure, i don't even see the use of configuring them. (Might as well look into other classes that we ourselves do not quite setup)

I didn't touch colours as it has the validator, but that just seems to convert hexstrings into ints, so that might be something we can move to a basic class. Emojis needs to stay he same way as it is, mostly for the trash & incidents emojis.

PR #2696, opened 2023-07-26
Change DM users get when warned

swfarnsworth:
Create separate message for warnings without duration, expiry, or "to appeal this infraction ..."

Since warnings involve no change in permissions, the DM that users receive said "Duration: Permanent \\ Expires: Never", which some interpreted to mean that we hold a permanent grudge. This is not intended, so this commit creates a separate template that omits those fields. Some users also thought they could appeal a warning to have it removed from the record, but this is not the case, so this commit creates a separate footer that removes "or appeal" from the instruction to contact the ModMail bot.

Showing the new message received when warned:
image

Showing the message received when muted (which has not changed):
image

(The ModMail bot is not in the user cache for the dummy account that I muted.)


ChrisLovering (approved):

Xithrius (approved):
Very good, thank you

PR #2712, opened 2023-08-10
Add a workflow to block fixup commits

shtlrs:
The idea is to block commits that have the `fixup` prefix, and make sure the authors squash those commits before merging.


shtlrs:
Superseded by  #2717

PR #2717, opened 2023-08-12
Block fixup commits

shtlrs:
This uses the organisation's fixup commit blocker workflow.

The purpose is to make sure all commits prefixed with `fixup` need to be squashed before merging.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,10 @@ concurrency:
 
 
 jobs:
+
+  fixup-commit-blocker:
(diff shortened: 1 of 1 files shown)


shtlrs:
> great

Not great enough apparently, I don't get why it's failing. Even when trying in debug mode, there are no useful logs.

shtlrs:
> > great
> 
> Not great enough apparently, I don't get why it's failing. Even when trying in debug mode, there are no useful logs.

Issue identified and fixed [here](https://github.com/python-discord/.github/pull/20)

MarkKoz (approved):

PR #2719, opened 2023-08-13
Make admins confirm edits of other users' reminders

### LINKED ISSUE #2353: Add confirmation step when deleting or modifying someone else's reminders
Once upon a time, an admin who shall not be named [modified](https://discord.com/channels/267624335836053506/635950537262759947/957013405494038558) a reminder which was not her own. Apparently admins can do this! 

Let's put in a button confirmation prompt when messing with reminders that don't belong to the command invoker, so that future admins can be prevented from making such heinous mistakes!

n0Oo0Oo0b:
Closes #2353

## Screenshots:

Confirmation:
image
When cancelled:
image
When confirmed:
image

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -38,14 +38,14 @@ log = get_logger(__name__)
 LOCK_NAMESPACE = "reminder"
 WHITELISTED_CHANNELS = Guild.reminder_whitelist
 MAXIMUM_REMINDERS = 5
-CONFIRMATION_TIMEOUT = 60
+REMINDER_EDIT_CONFIRMATION_TIMEOUT = 60
 
 Mentionable = discord.Member | discord.Role
 ReminderMention = UnambiguousUser | discord.Role
 
 
-class ModifyConfirmationView(discord.ui.View):
-    """A view to confirm modifying someone else's reminder."""
+class ModifyReminderConfirmationView(discord.ui.View):
+    """A view to confirm modifying someone else's reminder by admins."""
 
     def __init__(self, author: discord.Member | discord.User):
         super().__init__(timeout=CONFIRMATION_TIMEOUT)
@@ -572,11 +572,11 @@ class Reminders(Cog):
             raise e
 
         if api_response["author"] == ctx.author.id:
-            log.debug(f"{ctx.author} is the reminder author and passes the check.")
+            log.debug(f"{ctx.author} is the reminder's author and passes the check.")
             return True
 
         if await has_any_role_check(ctx, Roles.admins):
-            log.debug(f"{ctx.author} is an admin, asking for confirmation to modify the reminder.")
+            log.debug(f"{ctx.author} is an admin, asking for confirmation to modify someone else's.")
 
             confirmation_view = ModifyConfirmationView(ctx.author)
             await ctx.send(
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -48,7 +48,7 @@ class ModifyReminderConfirmationView(discord.ui.View):
     """A view to confirm modifying someone else's reminder by admins."""
 
     def __init__(self, author: discord.Member | discord.User):
-        super().__init__(timeout=CONFIRMATION_TIMEOUT)
+        super().__init__(timeout=REMINDER_EDIT_CONFIRMATION_TIMEOUT)
         self.author = author
         self.result: bool | None = None
 
@@ -578,7 +578,7 @@ class Reminders(Cog):
         if await has_any_role_check(ctx, Roles.admins):
             log.debug(f"{ctx.author} is an admin, asking for confirmation to modify someone else's.")
 
-            confirmation_view = ModifyConfirmationView(ctx.author)
+            confirmation_view = ModifyReminderConfirmationView(ctx.author)
             await ctx.reply(
                 "Are you sure you want to modify someone else's reminder?",
                 view=confirmation_view,
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -47,7 +47,7 @@ ReminderMention = UnambiguousUser | discord.Role
 class ModifyReminderConfirmationView(discord.ui.View):
     """A view to confirm modifying someone else's reminder by admins."""
 
-    def __init__(self, author: discord.Member | discord.User):
+    def __init__(self, author: discord.Member):
         super().__init__(timeout=REMINDER_EDIT_CONFIRMATION_TIMEOUT)
         self.author = author
         self.result: bool | None = None
@@ -581,7 +581,7 @@ class Reminders(Cog):
 
             confirmation_view = ModifyReminderConfirmationView(ctx.author)
             await ctx.reply(
-                "Are you sure you want to modify <@{owner_id}>'s reminder?",
+                f"Are you sure you want to modify <@{owner_id}>'s reminder?",
                 view=confirmation_view,
             )
             await confirmation_view.wait()
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -571,17 +571,30 @@ class Reminders(Cog):
                     return False
             raise e
 
-        if await has_any_role_check(ctx, Roles.admins):
+        if api_response["author"] == ctx.author.id:
+            log.debug(f"{ctx.author} is the reminder author and passes the check.")
             return True
 
-        if api_response["author"] != ctx.author.id:
-            log.debug(f"{ctx.author} is not the reminder author and does not pass the check.")
-            if send_on_denial:
-                await send_denial(ctx, "You can't modify reminders of other users!")
-            return False
+        if await has_any_role_check(ctx, Roles.admins):
+            log.debug(f"{ctx.author} is an admin, asking for confirmation to modify the reminder.")
 
-        log.debug(f"{ctx.author} is the reminder author and passes the check.")
-        return True
+            confirmation_view = ModifyConfirmationView(ctx.author)
+            await ctx.send(
+                "Are you sure you want to modify someone else's reminder?",
+                view=confirmation_view,
+            )
+            await confirmation_view.wait()
+
+            if confirmation_view.result:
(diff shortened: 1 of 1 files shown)


bot/exts/utils/reminders.py
```diff
@@ -37,11 +38,43 @@
 LOCK_NAMESPACE = "reminder"
 WHITELISTED_CHANNELS = Guild.reminder_whitelist
 MAXIMUM_REMINDERS = 5
+CONFIRMATION_TIMEOUT = 60
```
shtlrs, line 41:
```suggestion
REMINDER_EDIT_CONFIRMATION_TIMEOUT = 60
```

bot/exts/utils/reminders.py
```diff
@@ -37,11 +38,43 @@
...
 
 Mentionable = discord.Member | discord.Role
 ReminderMention = UnambiguousUser | discord.Role
 
 
+class ModifyConfirmationView(discord.ui.View):
```
shtlrs, line 47:
```suggestion
class ModifyReminderConfirmationView(discord.ui.View):
```

bot/exts/utils/reminders.py
```diff
@@ -37,11 +38,43 @@
...
+    """A view to confirm modifying someone else's reminder."""
```
shtlrs, line 48:
```suggestion
    """A view to confirm modifying someone else's reminder by admins."""
```

bot/exts/utils/reminders.py
```diff
@@ -538,17 +571,30 @@ async def _can_modify(self, ctx: Context, reminder_id: str | int, send_on_denial
...
-                await send_denial(ctx, "You can't modify reminders of other users!")
-            return False
+        if await has_any_role_check(ctx, Roles.admins):
+            log.debug(f"{ctx.author} is an admin, asking for confirmation to modify the reminder.")
 
-        log.debug(f"{ctx.author} is the reminder author and passes the check.")
-        return True
+            confirmation_view = ModifyConfirmationView(ctx.author)
+            await ctx.send(
+                "Are you sure you want to modify someone else's reminder?",
```
shtlrs, line 583: Maybe mentioning who that person is would be useful?

bot/exts/utils/reminders.py
```diff
@@ -538,17 +571,30 @@ async def _can_modify(self, ctx: Context, reminder_id: str | int, send_on_denial
...
+            )
+            await confirmation_view.wait()
+
+            if confirmation_view.result:
+                log.debug(f"{ctx.author} has confirmed reminder modification.")
+            else:
+                log.debug(f"{ctx.author} has cancelled reminder modification.")
+            return confirmation_view.result
+
+        log.debug(f"{ctx.author} is not the reminder author and does not pass the check.")
```
shtlrs, line 594:
```suggestion
        log.debug(f"{ctx.author} is not the reminder's author and thus does not pass the check.")
```

bot/exts/utils/reminders.py
```diff
@@ -538,17 +571,30 @@ async def _can_modify(self, ctx: Context, reminder_id: str | int, send_on_denial
                     return False
             raise e
 
-        if await has_any_role_check(ctx, Roles.admins):
+        if api_response["author"] == ctx.author.id:
+            log.debug(f"{ctx.author} is the reminder author and passes the check.")
```
shtlrs, line 575:
```suggestion
            log.debug(f"{ctx.author} is the reminder's author and passes the check.")
```

bot/exts/utils/reminders.py
```diff
@@ -538,17 +571,30 @@ async def _can_modify(self, ctx: Context, reminder_id: str | int, send_on_denial
...
             return True
 
-        if api_response["author"] != ctx.author.id:
-            log.debug(f"{ctx.author} is not the reminder author and does not pass the check.")
-            if send_on_denial:
-                await send_denial(ctx, "You can't modify reminders of other users!")
-            return False
+        if await has_any_role_check(ctx, Roles.admins):
+            log.debug(f"{ctx.author} is an admin, asking for confirmation to modify the reminder.")
```
shtlrs, line 579:
```suggestion
            log.debug(f"{ctx.author} is an admin, asking for confirmation to modify someone else's.")
```

shtlrs (changes requested):
Mostly are nitpicks, and this is only a code review, i'll test it later.

bot/exts/utils/reminders.py
n0Oo0Oo0b: I think replying to the original message that invoked the command would be better.

shtlrs: I actually meant the owner of the reminder.

So a reply + mentioning the actual author would be a good combo

n0Oo0Oo0b: I see, it has been added now.

bot/exts/utils/reminders.py
```diff
@@ -537,18 +570,32 @@ async def _can_modify(self, ctx: Context, reminder_id: str | int, send_on_denial
...
-                await send_denial(ctx, "You can't modify reminders of other users!")
-            return False
+        if await has_any_role_check(ctx, Roles.admins):
+            log.debug(f"{ctx.author} is an admin, asking for confirmation to modify someone else's.")
 
-        log.debug(f"{ctx.author} is the reminder author and passes the check.")
-        return True
+            confirmation_view = ModifyReminderConfirmationView(ctx.author)
+            await ctx.reply(
+                "Are you sure you want to modify <@{owner_id}>'s reminder?",
```
shtlrs, line 584: This currently says `Are you sure you want to modify {{mention}}'s reminder?` when you're trying to delete it, which isn't quite explicit and doesnt point the invoker to the action.

It's better to make this clearer, and IIRC you'll have access to that from `ctx.command.invoked_subcommand`

bot/exts/utils/reminders.py
```diff
@@ -37,11 +38,43 @@
...
+REMINDER_EDIT_CONFIRMATION_TIMEOUT = 60
 
 Mentionable = discord.Member | discord.Role
 ReminderMention = UnambiguousUser | discord.Role
 
 
+class ModifyReminderConfirmationView(discord.ui.View):
+    """A view to confirm modifying someone else's reminder by admins."""
+
+    def __init__(self, author: discord.Member | discord.User):
```
shtlrs, line 50: The `author` can only be a `Member` to invoke this since we don't allow commands in DMs.

bot/exts/utils/reminders.py
```diff
@@ -37,11 +38,43 @@
...
+        super().__init__(timeout=REMINDER_EDIT_CONFIRMATION_TIMEOUT)
+        self.author = author
+        self.result: bool | None = None
+
+    async def interaction_check(self, interaction: Interaction) -> bool:
+        """Only allow interactions from the command invoker."""
```
shtlrs, line 56: The first sentence of a function's docstring should complete this sentence: `This function will...`

shtlrs (changes requested):
Works great ! 

Just a nitpick or two, and a small bug to fix.

bot/exts/utils/reminders.py
n0Oo0Oo0b: Maybe the action being performed could be passed into the function instead? Although it might be redundant since a Context object is already being passed in.

I think I will stick with using `ctx.command.invoked_subcommand` for now.

n0Oo0Oo0b: Also, would it be worth changing the message in L597 to also include the modification type? I feel like doing so might imply that editing/deleting other people's reminders have different permissions when they actually don't.

bot/exts/utils/reminders.py
n0Oo0Oo0b: Could you please elaborate on this? What could the docstring be changed to instead? Would something like "Prevent other users from using the view buttons." be better?

shtlrs: Actually, it could be left as is, so don't mind me.

bot/exts/utils/reminders.py
shtlrs: > Also, would it be worth changing the message in L597 to also include the modification type? I feel like doing so might imply that editing/deleting other people's reminders have different permissions when they actually don't.

No you don't need to, as that message is destined to people who don't have the rights to do anything with a reminder.

The reason I asked you to make it explicit for adminds is because it will pinpoint the action they're about to do better.

bot/exts/utils/reminders.py
```diff
@@ -37,11 +38,43 @@
...
+    async def confirm(self, interaction: Interaction, button: discord.ui.Button) -> None:
+        """Confirm the reminder modification."""
+        await interaction.response.edit_message(view=None)
+        self.result = True
+        self.stop()
+
+    @discord.ui.button(label="Cancel", row=0)
+    async def cancel(self, interaction: Interaction, button: discord.ui.Button) -> None:
+        """Cancel the reminder modification."""
+        await interaction.response.edit_message(content="🚫 Operation canceled.", view=None)
```
n0Oo0Oo0b, line 73: This can now cause ghost pings for the reminder owner. Would sending a second message instead be better?

bot/exts/utils/reminders.py
```diff
@@ -37,11 +38,43 @@
...
+        self.author = author
+        self.result: bool | None = None
+
+    async def interaction_check(self, interaction: Interaction) -> bool:
+        """Only allow interactions from the command invoker."""
+        return interaction.user.id == self.author.id
+
+    async def on_timeout(self) -> None:
+        """Default to not modifying if the user doesn't respond."""
+        self.result = False
```
n0Oo0Oo0b, line 61: If the view times out, the view doesn't get deleted so the buttons fail with "application didn't respond."
Should we manually delete the view if it timed out (with the return value of `View.wait()`), or should all deleting (regardless of whether it timed out) happen in `_can_modify`?

shtlrs: We recently had a discussion about deleting messages like this. And it's not something we want to keep on doing.
Let's keep the view, but delete the buttons.

wookie184 (approved):
Nice, thanks for working on this.

shtlrs (approved):
Great work, thank you!

PR #2726, opened 2023-08-21
feat: use new markdown feature in new help channel message

### LINKED ISSUE #2646: Use new markdown features in messages
The bot should use new markdown features instead hard coded unordered lists for example:

image

uses `•` it can be replaced by `-` and markdown would format it same way

DownDev:
Closes #2646

### DIFF
--- a/bot/exts/filtering/_ui/ui.py
+++ b/bot/exts/filtering/_ui/ui.py
@@ -155,13 +155,13 @@ def format_response_error(e: ResponseCodeError) -> Embed:
     """Format the response error into an embed."""
     description = ""
     if isinstance(e.response_json, list):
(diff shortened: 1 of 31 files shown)


ChrisLovering:
Does this PR cover everything required in that issue? If not, we shouldn't mark this PR as closing that issue.

DownDev:
I guess i can go over commands and see if there is something to change.

DownDev:
@ChrisLovering Actually many tags dont use correct syntax for markdown. Also the `!patreon supporters` command is what i found for now.

All of the above are using the `•` instead of `-`

Xithrius (approved):
LGTM. Might want to look around again after this branch is updated.

ChrisLovering (approved):

PR #2737, opened 2023-08-31
Improve the `in-place` tag

n0Oo0Oo0b:
Closes python-discord/meta#213

### DIFF
--- a/bot/resources/tags/in-place.md
+++ b/bot/resources/tags/in-place.md
@@ -23,5 +23,4 @@ On the other hand, the `sorted()` function is out-of-place, so it returns a new 
 >>> my_list
 [5, 2, 3, 1]
 ```
-In general, methods of mutable objects tend to be in-place (since it can be expensive to create a new object), whereas operations on immutable objects are always out-of-place (since they cannot be modified)
-.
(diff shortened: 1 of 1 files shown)


bot/resources/tags/in-place.md
```diff
@@ -1,22 +1,27 @@
...
+```python
+>>> my_list = [5, 2, 3, 1]
+>>> sorted_list = sorted(my_list)
+>>> sorted_list
+[1, 2, 3, 5]
+>>> my_list
+[5, 2, 3, 1]
+```
+In general, methods of mutable objects tend to be in-place (since it can be expensive to create a new object), whereas operations on immutable objects are always out-of-place (since they cannot be modified)
+.
```
wookie184, line 27: The full stop here will appear by itself
image

Should move it to the previous line, or split somewhere on a single space so no newline is rendered.

wookie184 (changes requested):

bot/resources/tags/in-place.md
n0Oo0Oo0b: My bad, I thought it was on the previous line because of line wrapping.

wookie184 (approved):
Looks good, thanks!

PR #2749, opened 2023-09-10
fix: split docs command on whitespace characters

import-pandas-as-numpy:
The !d command, when invoked, can result in the item not being found if the character immediately following the library reference is a new line. 
For instance: 
```
!d collections.deque
This is a phrase.
```

Removing the specific space instance within the split field causes this to split on all whitespace characters, so a newline character will now be appropriately handled for parsing.


Xithrius (approved):
Original behavior:
image

Fixed behavior:
image

ChrisLovering (approved):

PR #2766, opened 2023-09-21
Bump bot core

ChrisLovering:
This bumps bot core and updates bot to make use of the new logging utils.

The idea with this is to have a consistent place where all bots can pull the logging form at from.

### DIFF
--- a/bot/log.py
+++ b/bot/log.py
@@ -1,63 +1,29 @@
 import logging
 import os
 import sys
-from logging import Logger, handlers
+from logging import handlers
 from pathlib import Path
-from typing import TYPE_CHECKING, cast
 
 import coloredlogs
 import sentry_sdk
+from pydis_core.utils import logging as core_logging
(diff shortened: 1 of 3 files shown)


bot/log.py
```diff
@@ -69,18 +35,11 @@ def setup() -> None:
         }
 
     if "COLOREDLOGS_LOG_FORMAT" not in os.environ:
-        coloredlogs.DEFAULT_LOG_FORMAT = format_string
+        coloredlogs.DEFAULT_LOG_FORMAT = core_logging.log_format._fmt
```
ChrisLovering, line 38: This is the only thing I'm slightly sad about. Since I didn't surface the logging string used by bot core, this is the only way to get the raw logging string.

I think it's fine, but if anyone feels strongly about it I can update bot core to make this available in a variable.

Xithrius (approved):
yej

PR #2772, opened 2023-10-01
Give help thread owners ability to use `!title` command.

swfarnsworth:
Previously, only staff members could use the command. The implementation is a bit fragile as it depends on short-circuiting `commands.has_any_role(constants.Roles.helpers).predicate(ctx)` in the case that the user is not the OP, which would raise an exception for non-staffers instead of returning False.

image


bot/exts/help_channels/_cog.py
```diff
@@ -108,11 +108,14 @@ async def rename_help_post(self, ctx: commands.Context, *, title: str) -> None:
...
-            # Silently fail for non-helpers
-            return
-
-        await ctx.channel.edit(name=title)
+        # First check that the user is the OP as `has_any_role` will only return True
+        # or raise an exception instead of False
+
+        if (
+            ctx.author.id == ctx.channel.owner_id
+            or await commands.has_any_role(constants.Roles.helpers).predicate(ctx)
```
Robin5605, line 116: Perhaps we could use [this](https://github.com/python-discord/bot/blob/main/bot/utils/checks.py#L97-L107) utility function? Suggested by @TizzySaurus 
```suggestion
            or has_any_role_check(ctx, constants.Roles.helpers)
```
`has_any_role_check` would need to be imported `from bot.utils.checks import has_any_role_check`

Robin5605 (approved):

shenanigansd (approved):
+1 for @Robin5605 's suggestion but LGTM

bot/exts/help_channels/_cog.py
swfarnsworth: If that solves the "The implementation is a bit fragile as it depends on short-circuiting ..." problem, then I support it.

swfarnsworth:
I'm going to refactor and push another commit--do not merge yet.

swfarnsworth:
image

image

bot/exts/help_channels/_cog.py
```diff
@@ -108,8 +109,10 @@ async def rename_help_post(self, ctx: commands.Context, *, title: str) -> None:
...
 
-        if not await commands.has_any_role(constants.Roles.helpers).predicate(ctx):
-            # Silently fail for non-helpers
+        user_is_op = ctx.author.id == ctx.channel.owner_id
+        user_is_staff = await has_any_role_check(ctx, constants.Roles.helpers)
+
+        if not (user_is_op or user_is_staff):
             return
 
         await ctx.channel.edit(name=title)
```
Robin5605, line 118:
```suggestion
        if user_is_op or user_is_staff:
            await ctx.channel.edit(name=title)
```
I think not using a negative here makes the intent clearer

Robin5605 (changes requested):

ChrisLovering:
No longer needed since https://github.com/python-discord/bot/pull/2774 was merged.

Issue #2760, opened 2023-09-17
Turn `!raw` into a context menu command


n0Oo0Oo0b:
I think `!raw` would benefit from being a context menu command (that sends an ephemeral message). From what I've seen, it's mostly used in `#bot-commands` with a message ID/link, so being able to use it from the context menu without having to go to a different channel would be a QoL improvement.

Issue #1037, opened 2020-07-11
raw command doesn't trim lines that are too long for the paginator


MarkKoz:
Sentry Issue: [BOT-6H](https://sentry.io/organizations/python-discord/issues/1757973940/?referrer=github_integration)

The command uses discord.py's default `Paginator`. Can that be replaced with our `LinePaginator`, which automatically moves long text to a new page?

```
RuntimeError: Line exceeds maximum page size 1992
  File "discord/ext/commands/core.py", line 83, in wrapped
    ret = await coro(*args, **kwargs)
  File "bot/cogs/information.py", line 412, in raw
    add_content(title, transformer(item))
  File "bot/cogs/information.py", line 396, in add_content
    paginator.add_line(content.replace('```', '`` `'))
  File "discord/ext/commands/help.py", line 127, in add_line
    raise RuntimeError('Line exceeds maximum page size %s' % (max_page_size))

Error executing command invoked by <REDACTED>: !raw <REDACTED>
```

MarkKoz:
I got it to work with `LinePaginator` but it the presentation is not very readable, especially since embeds are fixed to a narrow width. How does everyone feel about sending a hastebin link if the content is over 2000 characters?

wookie184:
Sending a hastebin link sounds like a good idea.

PR #2773, opened 2023-10-01
feat: context menu for !raw

VirdanTheBurden:
Closes Issue #2760.

The `!raw` command has been restricted to only prefix command usage since inception, and has not been touched since. With the introduction of new interfaces for executing commands (i.e. application context menus) and the inconvenience of copying message links to check their raw content, this pull request hopes to simplify the process by adding a context menu button for users to click.


bot/exts/info/information.py
```diff
@@ -480,19 +485,29 @@ def format_fields(self, mapping: Mapping[str, Any], field_width: int | None = No
         # remove trailing whitespace
         return out.rstrip()
 
-    async def send_raw_content(self, ctx: Context, message: Message, json: bool = False) -> None:
+    async def send_raw_content(self, ctx: Context | Interaction, message: Message, json: bool = False) -> None:
```
Robin5605, line 488: I think this function should be simplified into just a function that accepts a message, and returns a `str` or `list[str]` (to accommodate for pagination). Accepting both a Context and Interaction feels a bit janky in my opinion.

bot/exts/info/information.py
```diff
@@ -528,6 +543,9 @@ async def raw(self, ctx: Context, message: Message) -> None:
         """Shows information about the raw API response."""
         await self.send_raw_content(ctx, message)
 
+    async def _raw_context_menu_callback(self, interaction: discord.Interaction) -> None:
```
Robin5605, line 546:
```suggestion
    async def _raw_context_menu_callback(self, interaction: discord.Interaction, message: discord.Message) -> None:
```
Needs to accept a message argument in order to be classified as a message context menu

Robin5605 (changes requested):

bot/exts/info/information.py
VirdanTheBurden: This has already been fixed in a commit downstream. I was just writing a placeholder for the time.

bot/exts/info/information.py
VirdanTheBurden: I didn't feel quite the same way at first, but due to other reasons, this route of implementation is frankly *garbage.* I'll be moving the message sending responsibilities to the... overlying(?) methods.

Robin5605: > I'll be moving the message sending responsibilities to the... overlying(?) methods.
That sounds like a good idea, now that we have two different methods that send the message in two different ways (normal `!raw` command and context-menu command. Former needs to use `ctx.send` while the latter needs to use `interaction.response.send_message`). It would make sense to practice principle of least responsibility here and have one function that's just responsible for fetching the raw content.

VirdanTheBurden: I've pushed some new changes that made `send_raw_content` only fetch and return a `Paginator` object instead. Feel free to take a look at the new commits while I go fix the tests. I think they're borked with the `ContextMenu` being there now.

VirdanTheBurden: Er, take that back. Tests passed with flying colors, haha!

bot/exts/info/information.py
```diff
@@ -518,15 +520,26 @@ def add_content(title: str, content: str) -> None:
...
+
+        paginator = await self.get_raw_content(message)
+
+        for page in paginator.pages:
+            await ctx.send(page, allowed_mentions=AllowedMentions.none())
+
+    async def _raw_context_menu_callback(self, interaction: Interaction, message: Message) -> None:
+        """The callback to be ran when the Raw Text context menu button is used."""
+        paginator = await self.get_raw_content(message)
+        await interaction.response.send_message(paginator.pages[1], ephemeral=True)
```
Robin5605, line 542: Perhaps we could trail out some button-based pagination here? It's a shame that users can only see one page with the context menu version. I think one of the largest (pun intended) issues people had with buttons for pagination is their large vertical size, though it may be less of an issue in ephemeral messages.

VirdanTheBurden: Yeah, okay. I'll try adding a view with some buttons to see if we can scroll through.

wookie184: An alternative would be to do what was suggested in https://github.com/python-discord/bot/issues/1037 and just send a pastebin link if the message is too long, maybe in a similar way to the eval command where it displays as much as it can first.

PR #2774, opened 2023-10-02
Remove cooldown role

ChrisLovering:
This PR is an alternative to https://github.com/python-discord/bot/pull/2772 and is in discussions (internally) [here](https://canary.discord.com/channels/267624335836053506/1129018680404090970)

## Summary
The new forum-based help system implements its own cooldown system which is currently set to restrict members to a new post every 2 minutes.

This restriction in addition to the much higher allowance of threads in our server, means that this help cooldown role is no longer useful.

In fact, it is detrimental. Discord, in its infinite wisdom, tied the ability to change a post's name and tags to the ability to the 'Send messages' permission, meaning users with the help cooldown role were unable to complete these actions.

## Role out steps
After merging & deploying this change, the `Help cooldown` role should be removed from all users. The role itself and the help cooldown channel can be kept in case a rollback is desired. A staff announcement & changelog message should be made too


ChrisLovering:
Marking as do not merge until internal discussions are complete, please feel free to do a code review though.

ChrisLovering:
A few people have said that this is something that we want. Removing the do not merge label.

Robin5605 (approved):
Looks good! I checked and it doesn't seem like `constants.Roles.help_cooldown` is being referenced anywhere else in the code so this shouldn't break anything.

Xithrius (approved):
:shipit:

PR #2777, opened 2023-10-03
Update all references of 3.11 in eval to 3.12

ChrisLovering:
snekbox now runs 3.12 by default, so the user output should reflect that.

Once snekbox support multiple versions, version detection will be done programmatically based on snekbox's response.


vivekashok1221 (approved):
LGTM
I'm assuming you left [this](https://github.com/python-discord/bot/blob/main/bot/exts/utils/snekbox/_cog.py#L185) untouched because we currently do not support switching between python versions.

ChrisLovering:
> I'm assuming you left [this](https://github.com/python-discord/bot/blob/main/bot/exts/utils/snekbox/_cog.py#L185) untouched because we currently do not support switching between python versions.

Yea, it's unused anyway.

Xithrius (approved):
I ain't see nothing missing, very nice.

PR #2797, opened 2023-10-27
New Tag: "Avoid System Python for Development"

BruDriguezz:
This new tag, "Avoid System Python for Development," highlights the drawbacks of using system Python for development work, providing solutions for developers. It promotes best practices to create isolated development environments and includes a helpful audio resource for further insights.

[Relevant thread in PyDis](https://discord.com/channels/267624335836053506/1167598403014885426)


n0Oo0Oo0b:
Was this tag approved anywhere? You can suggest new tags for the bot by opening an issue at python-discord/meta.

BruDriguezz:
> Was this tag approved anywhere? If it was approved somewhere like the `#comunity-meta` channel, it would be helpful to link it here for more context. Otherwise, you can suggest new tags for the bot by opening an issue at our meta repository [here](https://github.com/python-discord/meta/issues).

Yeah, Stelercus has advised me to [open a draft PR](https://discord.com/channels/267624335836053506/429409067623251969/1167559904026951724) for this particular issue. 
I'll edit the PR's description for further clarity.

swfarnsworth:
Thank you for writing this and opening this PR.

It's important that tags be as terse as possible, so as to drive a very specific point before losing the reader's attention. It's also important that the instructions are easy to follow for the target audience.

If we go forward with this tag, I'd like to first condense it significantly, to something along these lines:

> Operating systems often come with a system Python interpreter that is used internally. While it may be suitable for executing one-off scripts, modifying it might interfere with its internal use, especially on Linux or macOS.
> 
> If you need to install packages or use a different Python version than the system's interpreter, try installing a separate interpreter.

And then list instructions for how to do this on Windows, macOS, and Linux.

Does this sound alright to you?

BruDriguezz:
> Thank you for writing this and opening this PR.
> 
> It's important that tags be as terse as possible, so as to drive a very specific point before losing the reader's attention. It's also important that the instructions are easy to follow for the target audience.
> 
> If we go forward with this tag, I'd like to first condense it significantly, to something along these lines:
> 
> > Operating systems often come with a system Python interpreter that is used internally. While it may be suitable for executing one-off scripts, modifying it might interfere with its internal use, especially on Linux or macOS.
> > If you need to install packages or use a different Python version than the system's interpreter, try installing a separate interpreter.
> 
> And then list instructions for how to do this on Windows, macOS, and Linux.
> 
> Does this sound alright to you?

Your concerns about retaining the reader's attention are valid, but I still have some concerns about completely removing the observations about the consequences of tampering.
The point which the tag wishes to address is not only to instruct the reader about how to install/use alternatives, but to effectively push them away from possibly tampering with the system interpreter because of the ignorance of consequences associated with it - and I fear that, by removing those observations, the reader may disregard the tag as unnecessary advice and not feel encouraged to follow it.
Perhaps, we can compromise; I can condense the tag into something smaller and concise, while still retaining the bulk of the original observations that I find necessary to address.
Thoughts?

BruDriguezz:
I revisited this draft and incorporated the tag updates as per my previous suggestions. I've also condensed it for brevity.
Feedback from anyone interested in chipping in would be greatly appreciated!

bot/resources/tags/system-python.md
```diff
@@ -0,0 +1,18 @@
+---
+embed:
+    title: "System Python"
+---
+
+*Why Avoid System Python for Development on Unix-like Systems:*
+
+- **Critical System Operations Dependency:** Altering the system interpreter may harm system functioning irreversibly.
```
import-pandas-as-numpy, line 8: Are we altering the interpreter at all? Breaking interactions with system Python tend to create dependency conflicts, which is the principle reason that we tend to dissuade individuals from utilizing the System Python installation.

bot/resources/tags/system-python.md
```diff
@@ -0,0 +1,18 @@
...
+  
+- **Stability and Security Concerns:** System interpreters lag behind current releases, lacking the latest features and security patches.
+  
+- **Limited Package Control:** External package management restricts control over versions, leading to compatibility issues with outdated packages.
+
+*Recommended Approach:*
+
+- **Install Independent Interpreter:** Install Python from source or utilize a virtual environment for flexibility and control.
+
+- **Utilize Pyenv or Similar Tools:** Manage multiple Python versions and create isolated development environments for smoother workflows.
```
import-pandas-as-numpy, line 18: I'd link to pyenv here for posterity.
```suggestion
- **Utilize [Pyenv](https://github.com/pyenv/pyenv) or Similar Tools:** Manage multiple Python versions and create isolated development environments for smoother workflows.
```

bot/resources/tags/system-python.md
BruDriguezz: They may attempt to alter it by 'updating' it or deleting files and libraries they find unnecessary, both of which have already happened in PyDis. 
Although the chances of it happening are significantly smaller than a simple dependency break, I don't see the harm in highlighting those risks as a stronger way to dissuade them from even trying.

bot/resources/tags/system-python.md
BruDriguezz: That would be nice, yeah. I agree.

bot/resources/tags/system-python.md
import-pandas-as-numpy:
```suggestion
- **Critical Operation System Dependencies:** Altering the system Python installation may harm internal operating system dependencies. 
```

import-pandas-as-numpy (approved):
I'm okay with this. I think we can debate some of the wording a little bit, but ultimately I think it accomplishes the goal of succinctly answering the following: 
"Why shouldn't I use the Python distributed with my OS?" and "What should I do instead?"

Technically installing a separate interpreter is done by Pyenv or similar tools, but I'm not sure engaging in that level of elaboration is going to make this any clearer to newcomers. 

LGTM.

PR #2817, opened 2023-11-18
Tag updates

swfarnsworth:
This PR updates two tags and introduces a new one about comparisons to True and False.

Regarding the new tag, I'm open to suggestions (and destructive criticism) about using vertical space more sparingly in the code section. I don't like having `my_func(user_input)` four times, but I'm worried that having `...` on the same line as the if statements will be confusing for beginners.


bot/resources/tags/equals-true.md
```diff
@@ -0,0 +1,24 @@
+---
+embed:
+    title: "Comparisons to `True` and `False`"
+---
+It's tempting to think that if statements always need a comparison operator like `==` or `!=`, but this isn't true.
+If you're just checking if a value is truthy or falsey, you don't need `== True` or `== False`.
```
Xithrius, line 6: > truthy or falsey

Should we expand on what these mean?

bot/resources/tags/codeblock.md
```diff
@@ -1,11 +1,11 @@
 ---
 embed:
-    title: "Formatting code on discord"
+    title: "Formatting code on Discord"
 ---
 Here's how to format Python code on Discord:
 
 \`\`\`py
-print('Hello world!')
+print('Code goes here on a new line')
```
camcaswell, line 8: Maybe this is overthinking, but are people going to think they need to put their code inside the `print('...')`?

bot/resources/tags/equals-true.md
camcaswell: I lean towards no. It would make the scope of the tag a lot bigger and is not necessarily relevant to the reader. As is, newbies can make an assumption about what they mean and still understand the rest or they can ask what they mean. Either are good outcomes.

camcaswell (commented):
👍

bot/resources/tags/codeblock.md
wookie184: I think `print("Hello world!")` was more clearly a placeholder, yeah.

swfarnsworth: I reverted this change.

wookie184 (approved):
Looks good to me, thanks

PR #2828, opened 2023-11-25
Don't shadow user object with dict when searching for infractions

ChrisLovering:
Solves #2827 Solves BOT-3P0


Xithrius (approved):
Another wonderful PR

PR #2838, opened 2023-12-07
Fix broken hyperlink rendering in windows-path tag

thegamecracks:
This is a minor fix for the !windows-path tag incorrectly rendering this hyperlink:

Screenshot of Discord not formating hyperlink to numpy


TizzySaurus (approved):
Thanks!

PR #2839, opened 2023-12-09
Support editing of timeout durations.

shtlrs:
All of the infractions that we have are scheduled for expiry thanks to our own Scheduler.

However, when it comes to `timeouts/mutes`, the expiry is handled by Discord itself, and the duration that we set in our databse serves no purposes besides their notifying the user about re-acquiring the ability to send messages again.

So this particular case needs to be handled on its own whenver the infraction type is a `timeout` and a new `expiry` datetime has been assigned.

### DIFF
--- a/bot/exts/moderation/infraction/_utils.py
+++ b/bot/exts/moderation/infraction/_utils.py
@@ -1,5 +1,9 @@
+
+import datetime
+
 import arrow
 import discord
+from dateutil.relativedelta import relativedelta
 from discord import Member
 from discord.ext.commands import Context
 from pydis_core.site_api import ResponseCodeError
@@ -60,6 +64,8 @@ INFRACTION_DESCRIPTION_WARNING_TEMPLATE = (
     "**Reason:** {reason}\n"
 )
 
+
+MAXIMUM_TIMEOUT_DAYS = datetime.timedelta(days=28)
 
 async def post_user(ctx: Context, user: MemberOrUser) -> dict | None:
     """
@@ -299,6 +305,22 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
             "The user either could not be retrieved or probably disabled their DMs."
         )
         return False
+
+
+def cap_timeout_duration(duration: datetime.datetime | relativedelta) -> tuple[bool, datetime.datetime]:
+    """Caps the duration of a duration to Discord's limit."""
+    now = arrow.utcnow()
+    capped = False
+    if isinstance(duration, relativedelta):
+        duration += now
+
+    if duration > now + MAXIMUM_TIMEOUT_DAYS:
+        duration = now + MAXIMUM_TIMEOUT_DAYS - datetime.timedelta(minutes=1)  # Duration cap is exclusive.
+        capped = True
+    elif duration > now + MAXIMUM_TIMEOUT_DAYS - datetime.timedelta(minutes=1):
+        # Duration cap is exclusive. This is to still allow specifying "28d".
+        duration -= datetime.timedelta(minutes=1)
+    return capped, duration
 
 
 async def confirm_elevated_user_ban(ctx: Context, user: MemberOrUser) -> bool:
--- a/bot/exts/moderation/infraction/_utils.py
+++ b/bot/exts/moderation/infraction/_utils.py
@@ -5,17 +5,17 @@ import arrow
 import discord
 from dateutil.relativedelta import relativedelta
 from discord import Member
-from discord.ext.commands import Context
+from discord.ext.commands import Bot, Context
 from pydis_core.site_api import ResponseCodeError
 
 import bot
-from bot.constants import Categories, Colours, Icons, MODERATION_ROLES, STAFF_PARTNERS_COMMUNITY_ROLES
+from bot.constants import Categories, Channels, Colours, Icons, MODERATION_ROLES, STAFF_PARTNERS_COMMUNITY_ROLES
 from bot.converters import DurationOrExpiry, MemberOrUser
 from bot.errors import InvalidInfractedUserError
 from bot.exts.moderation.infraction._views import InfractionConfirmationView
 from bot.log import get_logger
 from bot.utils import time
-from bot.utils.channel import is_in_category
+from bot.utils.channel import is_in_category, is_mod_channel
 from bot.utils.time import unpack_duration
 
 log = get_logger(__name__)
@@ -66,6 +66,11 @@ INFRACTION_DESCRIPTION_WARNING_TEMPLATE = (
 
 
 MAXIMUM_TIMEOUT_DAYS = datetime.timedelta(days=28)
+TIMEOUT_CAP_MESSAGE = (
+    f"The timeout for {{0}} can't be longer than {MAXIMUM_TIMEOUT_DAYS.days} days."
+    " I'll pretend that's what you meant."
+)
+
 
 async def post_user(ctx: Context, user: MemberOrUser) -> dict | None:
     """
@@ -308,7 +313,7 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
 
 
 def cap_timeout_duration(duration: datetime.datetime | relativedelta) -> tuple[bool, datetime.datetime]:
-    """Caps the duration of a duration to Discord's limit."""
+    """Cap the duration of a duration to Discord's limit."""
     now = arrow.utcnow()
     capped = False
     if isinstance(duration, relativedelta):
@@ -355,3 +360,13 @@ async def confirm_elevated_user_ban(ctx: Context, user: MemberOrUser) -> bool:
         return False
 
     return True
+
+
+async def notify_timeout_cap(bot: Bot, ctx: Context, user: discord.Member) -> None:
+    """Notify moderators about a timeout duration being capped."""
+    cap_message_for_user = TIMEOUT_CAP_MESSAGE.format(user.mention)
+    if is_mod_channel(ctx.channel):
+        await ctx.reply(f":warning: {cap_message_for_user}")
+    else:
+        await bot.get_channel(Channels.mods).send(
+            f":warning: {ctx.author.mention} {cap_message_for_user}")
--- a/bot/exts/moderation/infraction/infractions.py
+++ b/bot/exts/moderation/infraction/infractions.py
@@ -230,10 +230,8 @@ class Infractions(InfractionScheduler, commands.Cog):
         if duration is None:
             duration = await Duration().convert(ctx, "1h")
         else:
-            now = arrow.utcnow()
-            if isinstance(duration, relativedelta):
-                duration += now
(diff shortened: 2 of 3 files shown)


bot/exts/moderation/infraction/management.py
```diff
@@ -232,6 +236,9 @@ async def infraction_edit(
             # If the infraction was not marked as permanent, schedule a new expiration task
             if request_data["expires_at"]:
                 self.infractions_cog.schedule_expiration(new_infraction)
+                # Timeouts are handled by Discord itself, so we need to edit the expiry in Discord as well
+                if user and infraction["type"] == "timeout":
+                    await user.edit(reason=reason, timed_out_until=expiry)
```
vivekashok1221, line 241: We need to cap expiry at 28 days since that is the maximum timeout duration supported by Discord. The moderator should also be notified of this.

vivekashok1221, line 241: If the user is not present in the server, the infraction duration is updated in the database and a confirmation message is sent- but the actual timeout duration is not changed Discord-side (basically the same undesirable behaviour we have for editing timeout duration right now) .


We should either:
a) notify that the user has left the server
b) apply the new timeout duration when the user rejoins the server (might have to bring back some of  logic we had for `on_member_join` event from back when we used `@muted` role for muting users)

vivekashok1221 (changes requested):
Few wrinkles to iron out but code works otherwise.

bot/exts/moderation/infraction/management.py
vivekashok1221: I'm assuming it's not possible to change timeout duration for users while they are not present in the server.

shtlrs: 5abcfaf564897b10234799d924bf1e8ca7fc7034

shtlrs: 7d9bde13815dac0b8b3a71ebb90597ab06e03b25

bot/exts/moderation/infraction/infractions.py
```diff
@@ -664,6 +645,26 @@ async def cog_command_error(self, ctx: Context, error: Exception) -> None:
...
+        """
+        active_timeouts = await self.bot.api_client.get(
+            endpoint="bot/infractions",
+            params={"active": "true", "type": "timeout", "user__id": member.id}
+        )
+        if active_timeouts and not member.is_timed_out():
+            reason = f"Applying active timeout for returning member: {active_timeouts[0]['id']}"
+
+        async def action() -> None:
+            await member.edit(timed_out_until=arrow.get(active_timeouts[0]["expires_at"]).datetime, reason=reason)
```
vivekashok1221, line 664: I get the following error when my test account that I muted rejoins the server (while the timeout is still active) because `member.is_timed_out()` is `True` and hence `reason` is not assigned a value:
```
bot-1        |   File "/bot/bot/exts/moderation/infraction/infractions.py", line 664, in action
bot-1        |     await member.edit(timed_out_until=arrow.get(active_timeouts[0]["expires_at"]).datetime, reason=reason)
bot-1        |                                                                                                    ^^^^^^
bot-1        | NameError: cannot access free variable 'reason' where it is not associated with a value in enclosing scope
```

vivekashok1221 (changes requested):

bot/exts/moderation/infraction/infractions.py
vivekashok1221, line 664: This line triggers an `IndexError` whenever a user without an active timeout infraction joins the server.

vivekashok1221 (changes requested):
Sorry for the late review, I was teaching my houseplants Morse code. Please forgive.

bot/exts/moderation/infraction/infractions.py
shtlrs: Good find, both were fixed in the [force push](https://github.com/python-discord/bot/compare/67b8c8ab641529d90aa39756e69e980b41428387..36ee066ff883d56ef31d129a45d37e35a18849ab)

bot/exts/moderation/infraction/infractions.py
```diff
@@ -664,6 +645,26 @@ async def cog_command_error(self, ctx: Context, error: Exception) -> None:
...
+            endpoint="bot/infractions",
+            params={"active": "true", "type": "timeout", "user__id": member.id}
+        )
+
+        if active_timeouts and not member.is_timed_out():
+            reason = f"Applying active timeout for returning member: {active_timeouts[0]['id']}"
+
+            async def action() -> None:
+                await member.edit(timed_out_until=arrow.get(active_timeouts[0]["expires_at"]).datetime, reason=reason)
+            await self.reapply_infraction(active_timeouts[0], action)
```
vivekashok1221, line 666: Doesn't this overlook users whose timeout duration was edited while they were not present on the server but are still under a timeout when they rejoin the server? Once they rejoin the server, they will still be under the old timeout duration and not the new edited duration.

vivekashok1221 (changes requested):

bot/exts/moderation/infraction/infractions.py
shtlrs: How does [this](https://github.com/python-discord/bot/compare/36ee066ff883d56ef31d129a45d37e35a18849ab..8e1ff85c94afd86e6aee42de2ea7c3fdc995938e) look

bot/exts/moderation/infraction/infractions.py
```diff
@@ -664,6 +645,31 @@ async def cog_command_error(self, ctx: Context, error: Exception) -> None:
...
+        active_timeouts = await self.bot.api_client.get(
+            endpoint="bot/infractions",
+            params={"active": "true", "type": "timeout", "user__id": member.id}
+        )
+
+        if active_timeouts:
+            timeout_infraction = active_timeouts[0]
+            expiry = arrow.get(timeout_infraction["expires_at"], tzinfo=UTC).datetime
+            if member.is_timed_out():
+                expiry = max(expiry, member.timed_out_until)
```
vivekashok1221, line 665: Shouldn't we always use the expiry from the database?
With your current approach, if a moderator reduces the timeout duration, the infraction log will display the new duration. However, when the user rejoins the server, the old (original) longer timeout duration will be applied.

vivekashok1221 (changes requested):

vivekashok1221:
I've noticed something that could warrant it's own issue. When attempting to unmute a user who left the server, the bot will reply that the it has failed to pardon the infraction since the user has left the server. However, it will also mark the infraction as inactive in the db. This means that once the user rejoining, someone will have to manually remove the timeout.

bot/exts/moderation/infraction/infractions.py
shtlrs: I'm not on a streak this period of the year...

You're right, I'll fix that.

shtlrs:
> I've noticed something that could warrant its own issue. When attempting to unmute a user who left the server, the bot will reply that the it has failed to pardon the infraction since the user has left the server. However, it will also mark the infraction as inactive in the db. This means that once the user rejoins, someone will have to manually remove the timeout.

Yes, i think this should be dealt with in a separate issue.

bot/exts/moderation/infraction/infractions.py
```diff
@@ -664,6 +645,29 @@ async def cog_command_error(self, ctx: Context, error: Exception) -> None:
...
+
+        if active_timeouts:
+            timeout_infraction = active_timeouts[0]
+            expiry = arrow.get(timeout_infraction["expires_at"], tzinfo=UTC).datetime
+
+            reason = f"Applying active timeout for returning member: {timeout_infraction['id']}"
+
+            async def action() -> None:
+                await member.edit(timed_out_until=expiry, reason=reason)
+            await self.reapply_infraction(timeout_infraction, action)
```
vivekashok1221, line 669: The only thing I *miiiiight* change would be to check if `member.timed_out_until`  is not same as the expiry time in the database before applying the timeout. This could potentially save an API call in the handful of scenarios when a user under a timeout rejoins the server and the timeout duration is unchanged (*although*, it is more likely that in the event a timed-out user does rejoin the server, the duration has not been edited).
Again, this case does not happen that often, and even if it did, that extra API call is not that big of a deal. So, you have my approval for merging this PR.

vivekashok1221 (approved):
LGTM :star2:  :1st_place_medal: 

I do have a minor comment but I'm fine with merging this PR as it is.

bot/exts/moderation/infraction/infractions.py
shtlrs: I'll accept this only because your houseplants are talking good of you about your Morse code teaching methodology.

Implemented [here](https://github.com/python-discord/bot/compare/eba77402114dab719c5b9f45d0478aa19d9f9fcf..129705a5b195b36e5b2c64fa2beb5822cefeb9de)

vivekashok1221 (approved):
### LGTM :star2: :1st_place_medal:

bot/exts/moderation/infraction/_utils.py
```diff
@@ -298,3 +308,29 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
             "The user either could not be retrieved or probably disabled their DMs."
         )
         return False
+
+
+def cap_timeout_duration(duration: datetime.datetime | relativedelta) -> tuple[bool, datetime.datetime]:
+    """Caps the duration of a duration to Discord's limit."""
```
MarkKoz, line 314:
```suggestion
    """Cap a duration to Discord's limit."""
```

bot/exts/moderation/infraction/_utils.py
```diff
@@ -298,3 +308,29 @@ async def send_private_embed(user: MemberOrUser, embed: discord.Embed) -> bool:
...
+        duration = now + MAXIMUM_TIMEOUT_DAYS - datetime.timedelta(minutes=1)  # Duration cap is exclusive.
+        capped = True
+    elif duration > now + MAXIMUM_TIMEOUT_DAYS - datetime.timedelta(minutes=1):
+        # Duration cap is exclusive. This is to still allow specifying "28d".
+        duration -= datetime.timedelta(minutes=1)
+    return capped, duration
+
+
+async def notify_timeout_cap(bot: Bot, ctx: Context, user: discord.Member) -> None:
+    """Notifies moderators about a timeout duration being capped."""
```
MarkKoz, line 330: Imperative mood is our prevailing style.
```suggestion
    """Notify moderators about a timeout duration being capped."""
```

bot/exts/moderation/infraction/infractions.py
```diff
@@ -664,6 +645,32 @@ async def cog_command_error(self, ctx: Context, error: Exception) -> None:
                 await ctx.send(str(error.errors[0]))
                 error.handled = True
 
+    @commands.Cog.listener()
+    async def on_member_join(self, member: Member) -> None:
+        """
+
```
MarkKoz, line 651: Extra newline here
```suggestion
```

MarkKoz (approved):
Just some nits

PR #2845, opened 2023-12-13
Add a cog to manage tracking a user's alt accounts

ChrisLovering:
### Important note
This must be merged after https://github.com/python-discord/site/pull/1171

# Summary

This allows moderators, in mod channels, to add, remove & list a user's alt accounts.

When searching for infractions by user id, the number of al accounts is also shown, if there are any.

# Screenshot
Wording can be discussed...
image

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -342,7 +342,7 @@ class Information(Cog):
         return embed
 
     async def user_alt_count(self, user: MemberOrUser) -> tuple[str, int | str]:
-        """Get the number oif alts for the given member."""
+        """Get the number of alts for the given member."""
         resp = await self.bot.api_client.get(f"bot/users/{user.id}")
         return ("Associated accounts", len(resp["alts"]) or "No associated accounts")
 
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -323,6 +323,7 @@ class Information(Cog):
         if is_mod_channel(ctx.channel):
             fields.append(await self.expanded_user_infraction_counts(user))
             fields.append(await self.user_nomination_counts(user))
+            fields.append(await self.user_alt_count(user))
         else:
             fields.append(await self.basic_user_infraction_counts(user))
 
@@ -339,6 +340,12 @@ class Information(Cog):
         embed.colour = user.colour if user.colour != Colour.default() else Colour.og_blurple()
 
         return embed
+
+    async def user_alt_count(self, user: MemberOrUser) -> tuple[str, int | str]:
+        """Get the number oif alts for the given member."""
+        resp = await self.bot.api_client.get(f"bot/users/{user.id}")
+        return ("Associated accounts", len(resp["alts"]) or "No associated accounts")
+
 
     async def basic_user_infraction_counts(self, user: MemberOrUser) -> tuple[str, str]:
         """Gets the total and active infraction counts for the given `member`."""
--- a/bot/exts/moderation/alts.py
+++ b/bot/exts/moderation/alts.py
@@ -88,7 +88,7 @@ class AlternateAccounts(commands.Cog):
             return
         await ctx.send(f"✅ {user_1.mention} and {user_2.mention} successfully marked as alts.")
 
-    @association_group.command(name="edit ", aliases=("e",))
+    @association_group.command(name="edit", aliases=("e",))
     async def edit_association_command(
         self,
         ctx: commands.Context,
--- a/bot/exts/moderation/alts.py
+++ b/bot/exts/moderation/alts.py
@@ -1,0 +1,169 @@
+import gettext
+
+import discord
+from discord.ext import commands
+from pydis_core.site_api import ResponseCodeError
+from pydis_core.utils.members import get_or_fetch_member
+
+from bot import constants
+from bot.bot import Bot
+from bot.converters import UnambiguousMemberOrUser
+from bot.log import get_logger
+from bot.pagination import LinePaginator
+from bot.utils.channel import is_mod_channel
+from bot.utils.time import discord_timestamp
+
+log = get_logger(__name__)
+
+
+class AlternateAccounts(commands.Cog):
+    """A cog used to track a user's alternative accounts across Discord."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @staticmethod
+    def error_text_from_error(error: ResponseCodeError) -> str:
+        """Format the error into a user-facing message."""
+        if resp_json := error.response_json:
+            errors = ", ".join(
+                resp_json.get("non_field_errors", []) +
+                resp_json.get("source", []) +
+                resp_json.get("detail", [])
+            )
+            if errors:
+                return gettext.ngettext("Error from site: ", "Errors from site: ", len(errors)) + errors
+            return str(error.response_json)
+        return error.response_text
+
+    async def alts_to_string(self, alts: list[dict[str: str | int | list[int]]]) -> list[str]:
+        """Convert a list of alts to a list of string representations."""
+        lines = []
+        guild = self.bot.get_guild(self.bot.guild_id)
+        for idx, alt in enumerate(alts):
+            alt_obj = await get_or_fetch_member(guild, alt["target"])
+            alt_name = str(alt_obj) if alt_obj else alt["target"]
+            created_at = discord_timestamp(alt["created_at"])
+            updated_at = discord_timestamp(alt["updated_at"])
+
+            edited = f" edited on {updated_at}\n" if "edited" in alt else "\n"
+            num_alts = len(alt["alts"])
+            lines.append(
+                f"**Association #{idx} - {alt_name}**\n"
+                f"<@{alt['target']}> - {alt['target']}\n"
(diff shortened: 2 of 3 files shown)


bot/exts/moderation/alts.py
```diff
@@ -0,0 +1,166 @@
...
+            created_at = discord_timestamp(alt["created_at"])
+            updated_at = discord_timestamp(alt["updated_at"])
+
+            edited = f" edited on {updated_at}\n" if "edited" in alt else "\n"
+            lines.append(
+                f"**Association #{idx} - {alt_name}**\n"
+                f"<@{alt['target']}> - {alt['target']}\n"
+                f"Issued by: <@{alt['actor']}> ({alt['actor']}) on {created_at}{edited}"
+                f"Context: {alt['context']}\n"
+                f"<@{alt['target']}> has {len(alt['alts'])} associated account(s)"
```
jchristgit, line 53: How about `gettext.ngettext` to pluralize this like a king?

ChrisLovering: Is [this](https://github.com/python-discord/bot/compare/1152d3c5e57db137e6420634ee4e8c6b28cffee8..56d8fe9fec46b8b31462fac927ee9a966208784c) what you were expecting?

jchristgit: That looks like pluralization done right. A great move by the Python Discord Community

bot/exts/moderation/alts.py
```diff
@@ -0,0 +1,169 @@
...
+            )
+        except ResponseCodeError as e:
+            error = self.error_text_from_error(e)
+            await ctx.send(f":x: {error}")
+            return
+        await ctx.send(f"✅ {user_1.mention} and {user_2.mention} successfully marked as alts.")
+
+    @association_group.command(name="edit ", aliases=("e",))
+    async def edit_association_command(
+        self,
```
shtlrs, line 88: Two things I've noticed here
1. When I use the `edit` alias, it tried to convert `edit` to an `UnambiguousMemberOrUser`

2. When you pass alts that don't exist yet, the error message isn't quite "pretty"
image

bot/exts/moderation/alts.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+class AlternateAccounts(commands.Cog):
+    """A cog used to track a user's alternative accounts across Discord."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @staticmethod
+    def error_text_from_error(error: ResponseCodeError) -> str:
```
shtlrs, line 26: Wouldn't it be better to handle each response code gracefully for each endpoint ?

bot/exts/moderation/infraction/management.py
```diff
@@ -311,6 +312,10 @@ async def search_user(self, ctx: Context, user: MemberOrUser | discord.Object) -
         )
         # Manually form mention from ID as discord.Object doesn't have a `.mention` attr
         prefix = f"<@{user.id}> - {user.id}"
+        # If the user has alts show in the prefix
```
shtlrs, line 315: Just asking here, currently if the user has no infractions at all but has alt accounts, it won't show anything. 

It might not be intuitive at first to call the `alt info` command as the first reflex is to call the `i s` one.

shtlrs (changes requested):

bot/exts/moderation/alts.py
```diff
@@ -0,0 +1,169 @@
...
+                resp_json.get("non_field_errors", []) +
+                resp_json.get("source", []) +
+                resp_json.get("detail", [])
+            )
+            if errors:
+                return gettext.ngettext("Error from site: ", "Errors from site: ", len(errors)) + errors
+            return str(error.response_json)
+        return error.response_text
+
+    async def alts_to_string(self, alts: list[dict[str: str | int | list[int]]]) -> list[str]:
```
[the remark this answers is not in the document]
ChrisLovering: I was going to leave it as `list[dict]` but thought I might as well expand it further. Didn't want to make a full model since site might change and I don't want ot have to maintain it in two places

bot/exts/moderation/infraction/management.py
ChrisLovering: I'll add this info to the user command when ran in mod channels

bot/exts/moderation/alts.py
```diff
@@ -0,0 +1,169 @@
...
+        await ctx.send(f"✅ Context for association between {user_1.mention} and {user_2.mention} updated.")
+
+    @association_group.command(name="remove", aliases=("r",))
+    async def alt_remove_command(
+        self,
+        ctx: commands.Context,
+        user_1: UnambiguousMemberOrUser,
+        user_2: UnambiguousMemberOrUser,
+    ) -> None:
+        """Remove the alt association between the two users."""
```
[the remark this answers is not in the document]
ChrisLovering: Maybe this can be something we add in the future if we find outselves needing it often

bot/exts/moderation/alts.py
ChrisLovering: 1. There was a space after `edit` in the name, so that's why it failed. I only ever tested with the `e` alias. Fixed now.
2. Fixed this too

bot/exts/moderation/alts.py
```diff
@@ -0,0 +1,169 @@
...
+                f"<@{alt['target']}> has {num_alts} associated {gettext.ngettext('account', 'accounts', num_alts)}"
+            )
+        return lines
+
+    @commands.group(name="association ", aliases=("alt", "alts", "assoc"), invoke_without_command=True)
+    async def association_group(
+        self,
+        ctx: commands.Context,
+        user_1: UnambiguousMemberOrUser,
+        user_2: UnambiguousMemberOrUser,
```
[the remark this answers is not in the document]
ChrisLovering: Yea might as well, blocked that.

bot/exts/moderation/alts.py
ChrisLovering: There's overlapping between what errors can be returned by what endpoint, so this was the easiest, and most extensible way i could think of doing this

bot/exts/moderation/alts.py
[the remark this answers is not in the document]
shtlrs: Why would you have to maintain it in 2 places ? 
Reading this as it is is "painful" and one would need to squeeze their toughts in order to understand what's going on, and it isn't pleasant. Giving it structure would be more readable/intuitive/understandable.

bot/exts/moderation/alts.py
```diff
@@ -0,0 +1,174 @@
...
+            if errors:
+                return gettext.ngettext("Error from site: ", "Errors from site: ", len(errors)) + errors
+            return str(error.response_json)
+        return error.response_text
+
+    async def alts_to_string(self, alts: list[dict[str: str | int | list[int]]]) -> list[str]:
+        """Convert a list of alts to a list of string representations."""
+        lines = []
+        guild = self.bot.get_guild(self.bot.guild_id)
+        for idx, alt in enumerate(alts):
```
shtlrs, line 44: Might be better to start enumerating from `1` using the `start` kwarg here

bot/exts/info/information.py
```diff
@@ -337,6 +338,12 @@ async def create_user_embed(self, ctx: Context, user: MemberOrUser, passed_as_me
 
         return embed
 
+    async def user_alt_count(self, user: MemberOrUser) -> tuple[str, int | str]:
+        """Get the number oif alts for the given member."""
```
janine9vn, line 342: itty bitty typo `oif` -> `of`

janine9vn (approved):

bot/exts/info/information.py
mbaruh: That's just his accent

bot/exts/moderation/alts.py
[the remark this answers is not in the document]
wookie184: I think that the current type isn't very useful ,`str | int | list[int]` makes the usages below technically incorrect since whenever accessing a value they should handle it being any of those types.

I would be find with any of these, since they're all "correct", it just depends what you consider in-scope for the PR given our typing elsewhere with infractions isn't great either afaik:
1) `list[dict[str, typing.Any]]`
2) Use a `TypedDict`
3) Use a Pydantic model

bot/exts/moderation/alts.py
ChrisLovering: We use 0-based indexing around here!

bot/exts/info/information.py
ChrisLovering: I have corrected my accent in https://github.com/python-discord/bot/pull/2845/commits/98eea891c25f8beb7a99685fe1daa023654d2866

bot/exts/moderation/alts.py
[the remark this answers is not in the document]
ChrisLovering: Simplified the type annotation in https://github.com/python-discord/bot/pull/2845/commits/a380d2e030cf3f206e5e4d359db228e06cac81f5 (option 1)

shtlrs (approved):

jb3 (approved):

PR #2852, opened 2023-12-15
Use the new mailing lists API to track news

shtlrs:
> WARNING: This needs to be merged along with [site#1174](https://github.com/python-discord/site/pull/1174)

In the effort of removing the `bot-settings` table that has both news & defcon settings, we've started migrating these to their own tables.

This updates the `PythonNews` Cog to use the new endpoint to fetch/update news that have been tracked/posted.

We used to do syncing of schemas as we relied on a "dynamic" JSON, but now the mailing lists are predefined so no syncing is required.


bot/exts/info/python_news.py
```diff
@@ -216,16 +201,21 @@ async def post_maillist_news(self) -> None:
...
 
-                if msg.channel.is_news():
-                    log.trace("Publishing mailing list message because it was in a news channel")
-                    await msg.publish()
+                    if msg.channel.is_news():
+                        log.trace("Publishing mailing list message because it was in a news channel")
+                        await msg.publish()
+                except ResponseCodeError as e:
+                    non_field_errors = e.response_json.get("non_field_errors", [])
+                    if non_field_errors == ["Seen item already known."]:
```
jchristgit, line 215: What does this event cover? Inconsistent bot data?

bot/exts/info/python_news.py
```diff
@@ -139,26 +118,31 @@ async def post_pep_news(self) -> None:
...
-        # Apply new sent news to DB to avoid duplicate sending
-        await self.bot.api_client.put("bot/bot-settings/news", json=payload)
+                if msg.channel.is_news():
+                    log.trace("Publishing PEP announcement because it was in a news channel")
+                    await msg.publish()
+            except ResponseCodeError as e:
+                non_field_errors = e.response_json.get("non_field_errors", [])
+                if non_field_errors == ["Seen item already known."]:
+                    continue
+                raise e
```
jchristgit, line 136: This is kind of duplicated with the other block below. Can we turn it into a function? We could update the log.trace call to contain the mailing list name, so you could imply it's a PEP for those.

bot/exts/info/python_news.py
shtlrs: Yes, I'm not sure it'll hit that point but it's there for safety.
I ran into the issue when debugging, but I think it only happens because the loop is paused in the breakpoints.

bot/exts/info/python_news.py
shtlrs: Available in b93972d7644c42637f1e985c979bb8de77f7c352

I didn't want to add the message publishing thing because it's out of scope of the method.

jchristgit (approved):

Xithrius (approved):

PR #2862, opened 2023-12-23
Recognize `pyi` as `py` in code snippets

### LINKED ISSUE #2843: Recognise `pyi` snippets as `py` snippets when sharing github/gitlab links
Useful when referencing stubs

ad-chaos:
Closes #2843 

This special cases `pyi` only, I thought of adding a `lang_map` dictionary, but that seemed unnecessary at the moment.

### DIFF
--- a/bot/exts/info/code_snippets.py
+++ b/bot/exts/info/code_snippets.py
@@ -210,6 +210,9 @@ class CodeSnippets(Cog):
         if not is_valid_language:
             language = ""
 
+        if language == "pyi":
+            language = "py"
+
         # Adds a label showing the file path to the snippet
         if start_line == end_line:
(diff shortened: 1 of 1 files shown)


shenanigansd (approved):
+1 for the language map dict, there will only be one special case for the foreseeable future, so this is fine.

ChrisLovering (changes requested):
Could you move this a few lines down so that it's done after the traimming and validation? Will cover a few more cases that way.

ad-chaos:
Done

ChrisLovering (approved):

ChrisLovering:
Great stuff, thanks @ad-chaos!

Issue #2761, opened 2023-09-17
Allow using replies to `!raw` command


Robin5605:
I have on numerous occasions replied to a message with just the `!raw` command, but the command seems to expect a link. It would be useful if `!raw` could be used in the same way `.bookmark` command can, simply by replying to a message. Perhaps this is also related to #2760 ?

PR #2864, opened 2023-12-25
Support reply as a message reference for !raw [json]

ichard26:
Resolves GH-2761.

image


ichard26:
I didn't break the referenced message fallback out into its own function as I am not convinced it would be a net readability improvement. I don't have a strong opinion on this though.

bot/exts/info/information.py
```diff
@@ -524,13 +525,29 @@ def add_content(title: str, content: str) -> None:
...
+    async def raw(self, ctx: Context, message: Message | None = None) -> None:
         """Shows information about the raw API response."""
+        if message is None:
+            if (reference := ctx.message.reference) and isinstance(reference.resolved, Message):
+                message = reference.resolved
+
+        if message is None:
+            await send_denial(ctx, "Missing message argument. Please provide a message ID/link or reply to a message.")
+            return
+
```
MrHemlock, line 537: Is there a particular reason we're doing this in two checks?
```suggestion
        if message is None:
            if (reference := ctx.message.reference) and isinstance(reference.resolved, Message):
                message = reference.resolved

            else:
                await send_denial(ctx, "Missing message argument. Please provide a message ID/link or reply to a message.")
                return

```

bot/exts/info/information.py
```diff
@@ -524,13 +525,29 @@ def add_content(title: str, content: str) -> None:
...
+    async def json(self, ctx: Context, message: Message | None = None) -> None:
         """Shows information about the raw API response in a copy-pasteable Python format."""
+        if message is None:
+            if (reference := ctx.message.reference) and isinstance(reference.resolved, Message):
+                message = reference.resolved
+
+        if message is None:
+            await send_denial(ctx, "Missing message argument. Please provide a message ID/link or reply to a message.")
+            return
+
```
MrHemlock, line 550: Same as above, is there a specific reason for making these separate checks?


```suggestion
        if message is None:
            if (reference := ctx.message.reference) and isinstance(reference.resolved, Message):
                message = reference.resolved

            else:
                await send_denial(ctx, "Missing message argument. Please provide a message ID/link or reply to a message.")
                return

```

MrHemlock (changes requested):
Not sure if I'm being specifically picky or not, but it just seems redundant to check `if message is None:` twice

wookie184 (approved):
All looks good to me

ChrisLovering:
we should really get around to writing this as a d.py convertor in bot-core one of these days. This pattern seems to be used quite often now.

PR #2871, opened 2023-12-31
Fix test code that was doing nothing

wookie184:
In `push_scope_mock.set_tag.has_calls(set_tag_calls)`, `has_calls` is not a testing function so it just made a new mock and didn't actually test anything.

This makes the test use the correct method, `assert_has_calls`, which required as small change to the mock


ChrisLovering (approved):

lemonsaurus (approved):
The changes you're making seem reasonable to me, but I the thing I don't understand is why the tests that were run in the CI here don't include this fixed one. I mean, you're not adding the test, it should've been running even on main, right? It just wasn't a good test yet. So what am I missing? Cause I looked at the CI run and I don't see it there.

wookie184:
> The changes you're making seem reasonable to me, but I the thing I don't understand is why the tests that were run in the CI here don't include this fixed one. I mean, you're not adding the test, it should've been running even on main, right? It just wasn't a good test yet. So what am I missing? Cause I looked at the CI run and I don't see it there.

I don't think CI outputs each test that was run, just some dots: https://github.com/python-discord/bot/actions/runs/7371282307/job/20058499764?pr=2871#step:7:22. It has the coverage report though so you can at least see that the file was run.

lemonsaurus:
Guess I just missed it, sorry. 🧠 💊 🧑‍⚖️ 🏪

PR #2877, opened 2024-01-03
Move modlog method to util

### LINKED ISSUE #303: Proposal to change send_log_message calls to event dispatches.
Rather than fetching the ModLog cog and then calling or shortcutting the `send_log_message` calls in every cog, the idea is to utilise the event-based dispatch system (`bot.dispatch`) to emit a new event type that provides all the log args, and the ModLog cog will simple define a listener corresponding to that event which will trigger on emit.

It's likely more DRY, doesn't require importing a cog for type hints like are currently occurring, and will allow us to maybe control where each log type is sent within the ModLog logic instead of changing it in all the other cogs.

Any thoughts on the idea are welcome.

wookie184:
Closes #303

After this we could consider also making ignoring a modlog event a utility function. It would still need to fetch the cog internally, but it could do this more gracefully than current usages which would fail if the cog was unloaded (though the modlog cog currently can't be unloaded because of this).


shtlrs (approved):
Thanks !

Xithrius (approved):

PR #2880, opened 2024-01-04
Update unambiguous converter to not support usernames

### LINKED ISSUE #2878: Update "unambiguous" converters given Discord username changes
The current logic for detecting a username in the unambigous username converter checks if it contains a discriminator, but since discord removed discriminators it doesn't make sense any more:

https://github.com/python-discord/bot/blob/da6997aab25b57152eec5d8c706977c472563853/bot/converters.py#L425-L433

We can probably safely just remove support for passing a username, since the converters are mostly used for moderation commands where IDs are pretty much all that is used anyway.

wookie184:
Closes #2878

### DIFF
--- a/bot/converters.py
+++ b/bot/converters.py
@@ -423,23 +423,19 @@ class HushDurationConverter(Converter):
 
 
 def _is_an_unambiguous_user_argument(argument: str) -> bool:
-    """Check if the provided argument is a user mention, user id, or username (name#discrim)."""
-    has_id_or_mention = bool(IDConverter()._get_id_match(argument) or RE_USER_MENTION.match(argument))
+    """Check if the provided argument is a user mention or user id."""
+    user_id = IDConverter()._get_id_match(argument)
+    user_mention = RE_USER_MENTION.match(argument)
 
-    # Check to see if the author passed a username (a discriminator exists)
-    argument = argument.removeprefix("@")
-    has_username = len(argument) > 5 and argument[-5] == "#"
-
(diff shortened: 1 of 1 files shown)


wookie184:
There are some accounts that still have discriminators so there are some other bits of code referring to discriminators that we might not want to change yet, but I don't think it really matters here.

vivekashok1221 (approved):
I haven't tested but code looks good to me.

shtlrs:
I don't think it's quite used that much either but, just to be sure, how about we make a poll in mod-meta?

wookie184:
> I don't think it's quite used that much either but, just to be sure, how about we make a poll in mod-meta?

I've asked there, though I assume if nobody has had an issue so far it'll be fine.

bot/converters.py
```diff
@@ -423,23 +423,19 @@ async def convert(self, ctx: Context, argument: str) -> int:
 
 
 def _is_an_unambiguous_user_argument(argument: str) -> bool:
-    """Check if the provided argument is a user mention, user id, or username (name#discrim)."""
-    has_id_or_mention = bool(IDConverter()._get_id_match(argument) or RE_USER_MENTION.match(argument))
+    """Check if the provided argument is a user mention or user id."""
+    user_id = IDConverter()._get_id_match(argument)
```
shtlrs, line 427: `_get_id_match` is a static method, so no need to instantiate the converter.

shtlrs (approved):
Looks good & works well, thanks !

PR #2896, opened 2024-01-25
Update URL regex in embed filter (Fix #2895)

### LINKED ISSUE #2895: False alerts caused by incorrect regex url for `rich_embed` filter
Due to discord adding hyperlinking for users, the bot may mistakenly alert moderators about user embed.
Hyperlinks looks as follows: `[example](https://google.com)`

The regex we use [here](https://github.com/python-discord/bot/blob/main/bot/exts/filtering/_filters/unique/rich_embed.py#L11) would match the string `https://google.com)` and check it against `embed.url` [here](https://github.com/python-discord/bot/blob/main/bot/exts/filtering/_filters/unique/rich_embed.py#L40) which would cause a mismatch since Discord's `embed.url` would be `https://google.com`.

The solution might be as simple as adding `[^\)]` after `\S+` to the regex pattern.

lemonyte:
Fix #2895

### DIFF
--- a/bot/exts/filtering/_filters/unique/rich_embed.py
+++ b/bot/exts/filtering/_filters/unique/rich_embed.py
@@ -8,7 +8,7 @@ from bot.utils.helpers import remove_subdomain_from_url
 
 log = get_logger(__name__)
 
-URL_RE = re.compile(r"(https?://\S+)", flags=re.IGNORECASE)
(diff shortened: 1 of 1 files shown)


wookie184 (approved):
Tested, works nicely. 

Just a note for reviewers, a good test case for this is a twitter link. Not all links will produce the issue because it has to be a rich embed. I tested with this: `[example](https://twitter.com/ThePSF/status/1526895412732370945)`

ChrisLovering (approved):

PR #2897, opened 2024-01-27
Add Spammer field to `!user` info embed for moderators

swfarnsworth:
If the command is evoked in a mod channel, the resultant embed will display whether the user in question has been flagged by Discord as a spammer. This is potentially useful for evaluating reports of suspicious DMs.

Here it is when run in a moderator channel

image

And in a public channel 

image


Xithrius (approved):
At the current state, I don't see any flaws in this PR, besides the failing test.

After that's fixed, we should be good to go.

swfarnsworth:
As it turns out, the spammer flag can't be relied upon.

Xithrius:
Darn.

PR #2907, opened 2024-02-01
add backticks message to _get_no_ticks_message

zeptofine:
Adds the `"**These are backticks, not quotes.** ...` message to the response for pasting code with no ticks. This message is relevant for the context and will hopefully mean the bot/helpers/other users won't have to send *another* set of messages to convey the difference between backticks and quotes

### DIFF
--- a/bot/exts/info/codeblock/_instructions.py
+++ b/bot/exts/info/codeblock/_instructions.py
@@ -76,7 +76,10 @@ def _get_no_ticks_message(content: str) -> str | None:
             "Discord has support for Markdown, which allows you to post code with full "
             "syntax highlighting. Please use these whenever you paste code, as this "
             "helps improve the legibility and makes it easier for us to help you.\n\n"
-            f"**To do this, use the following method:**\n{example_blocks}"
+            f"**To do this, use the following method:**\n{example_blocks}\n"
+            "**These are backticks, not quotes.** Check "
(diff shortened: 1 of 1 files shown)


Xithrius:
Hello,

Was this discussed beforehand with a core developer, or in an issue? If not, please next time get approval from a core developer before creating a PR.

See [this page](https://www.pythondiscord.com/pages/guides/pydis-guides/contributing/contributing-guidelines/) on how to contribute to this repository.

Thanks!

zeptofine:
> Hello,
> 
> Was this discussed beforehand with a core developer, or in an issue? If not, please next time get approval from a core developer before creating a PR.
> 
> See [this page](https://www.pythondiscord.com/pages/guides/pydis-guides/contributing/contributing-guidelines/) on how to contribute to this repository.
> 
> Thanks!

ok my bad, should I close this?

wookie184:
@zeptofine it was alright, you can reopen it if you're still interested in this.

Currently if someone uses the wrong symbol a different message should be sent indicating that they're using the wrong symbol.

The trade-off here depends on how many people use the wrong symbol when seeing the current message. If it's most people than this change would be worth it. If it's only a few people then it might be better to keep the message shorter and accept that in a few cases they'll need the second message to be corrected. I don't have a great idea of the current state of things so this would probably need another opinion on whether it's worth doing.

PR #2908, opened 2024-02-01
Move url regex to constants

ChrisLovering:
This also updates the URL to ignore > charaters

https://regex101.com/r/yVs4Wj/1 can be used for testing


Xithrius (approved):
Perfection

mbaruh (changes requested):
There are subtleties here I'm not sure were addressed. You created a single constant in a separate file, but some of the constants it replaces aren't the same, and some of them are used in specific ways:

ChrisLovering:
I'm not going to pursue this further

PR #2914, opened 2024-02-04
Snekbox: truncate blocked file extensions

### LINKED ISSUE #2464: BOT-3EQ: long file extensions cause `!eval` result responses to breach the message length limits
# Summary

Sending this code -
```py
# https://discord.com/channels/267624335836053506/267659945086812160/1084221018119352350
for x in range(100):
  open(f"test.{'a' * x}", "w").close()```
```
Caused this error -
```
# https://discord.com/channels/267624335836053506/267659945086812160/1084221021281865788
In content: Must be 4000 or fewer in length.
```
It was triggered in -
https://github.com/python-discord/bot/blob/e21cd3ecea9afc7cf5d08bf75c944dd6049f4d28/bot/exts/utils/snekbox/_cog.py#L427

Looks like the blocked extensions are all appended to the text that is returned by the bot -
https://github.com/python-discord/bot/blob/e21cd3ecea9afc7cf5d08bf75c944dd6049f4d28/bot/exts/utils/snekbox/_cog.py#L386-L387
So the [extreme] file extension length caused the bot's response length to go over Discord's limit and get removed automatically.

Sentry Issue: [BOT-3EQ](https://python-discord.sentry.io/issues/3997032596/?referrer=github_integration)

MarkKoz:
Avoid Discord's character limit for messages.

Fix #2464

### DIFF
--- a/bot/exts/utils/snekbox/_cog.py
+++ b/bot/exts/utils/snekbox/_cog.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import contextlib
 import re
+from collections.abc import Iterable
 from functools import partial
 from operator import attrgetter
 from textwrap import dedent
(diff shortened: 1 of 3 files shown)


bot/exts/utils/snekbox/_cog.py
```diff
@@ -289,6 +290,69 @@ async def format_output(
 
         return output, paste_link
 
+    async def format_file_text(self, text_files: list[FileAttachment], output: str) -> str:
```
MarkKoz, line 293: This is the same as before. I just refactored it into its own function.

bot/exts/utils/snekbox/_cog.py
```diff
@@ -289,6 +290,69 @@ async def format_output(
...
+            blocked_msg = "Files with no extension can't be uploaded."
+        # Both
+        elif "" in blocked_sorted:
+            blocked_str = self.join_blocked_extensions(ext for ext in blocked_sorted if ext)
+            blocked_msg = (
+                f"Files with no extension or disallowed extensions can't be uploaded: **{blocked_str}**"
+            )
+        else:
+            blocked_str = self.join_blocked_extensions(blocked_sorted)
+            blocked_msg = f"Files with disallowed extensions can't be uploaded: **{blocked_str}**"
```
MarkKoz, line 339: If all extensions are too long then it will just show `...`. I think this is fine but it's also possible to completely omit `: **{blocked_str}**"` in such case.

shtlrs (approved):

bot/exts/utils/snekbox/_cog.py
```diff
@@ -289,6 +290,69 @@ async def format_output(
...
+
+        return f"\n{Emojis.failed_file} {blocked_msg}"
+
+    def join_blocked_extensions(self, extensions: Iterable, delimiter: str = ", ", char_limit: int = 100) -> str:
```
wookie184, line 343:
```suggestion
    def join_blocked_extensions(self, extensions: Iterable[str], delimiter: str = ", ", char_limit: int = 100) -> str:
```

bot/exts/utils/snekbox/_cog.py
```diff
@@ -289,6 +290,69 @@ async def format_output(
...
+        joined = ""
+        for ext in extensions:
+            cur_delimiter = delimiter if joined else ""
+            if len(joined) + len(cur_delimiter) + len(ext) >= char_limit:
+                joined += f"{cur_delimiter}..."
+                break
+
+            joined += f"{cur_delimiter}{ext}"
+
+        return joined
```
wookie184, line 353: It would be better to have `joined` as a list and then `"".join()` it at the end, so we're not relying on CPython optimisation for it not to have O(n) complexity.

Also, I believe this could go over `char_limit` by a few characters in some cases. If that's intentional for simplicity though then feel free to leave it as is.

wookie184 (changes requested):

bot/exts/utils/snekbox/_cog.py
MarkKoz: Yeah we have some leeway with our limits so I aimed to keep it simple. I also find `+=` more readable - I don't see this function as a performance bottleneck especially given we won't see large (in computer terms) numbers of extensions.

wookie184 (approved):
Works nicely, thanks

PR #2920, opened 2024-02-10
Reminders: show updated timestamp

### LINKED ISSUE #2172: Show new timestamp in confirmation embed after editing reminder duration

MarkKoz:
Resolve #2172

When a reminder's duration is updated, show the new timestamp in the confirmation message that is sent.

image


mbaruh (approved):

minalike (approved):
My hero 📆

PR #2924, opened 2024-02-18
Add upload date to !pypi

### LINKED ISSUE #2613: !pypi should include publish date in response

JelmerKorten:
Closes #2613 after discussion in #2614

pypi command with date in footer


bot/exts/info/pypi.py
```diff
@@ -22,6 +25,12 @@
 
 log = get_logger(__name__)
 
+def _get_latest_distribution_timestamp(data: dict[str, typing.Any]) -> datetime | None:
+    """Get upload date of last distribution, or `None` if no distributions were found."""
+    if not data["urls"]:
+        return None
+
+    return max(arrow.get(dist["upload_time"]) for dist in data["urls"]).date()
```
hedyhli, line 33: Looks good to me. However, would've been great if there was a way to just get the date for the most recently uploaded file directly. I wonder how bad it would be if we'd just get the `upload_time` for the last element in the urls, and use the date+time from the API directly, without needing `arrow`.

hedyhli (approved):

bot/exts/info/pypi.py
```diff
@@ -63,6 +72,10 @@ async def get_package_info(self, ctx: Context, package: str) -> None:
                     else:
                         embed.description = "No summary provided."
 
+                    upload_date = _get_latest_distribution_timestamp(response_json)
+                    if upload_date:
+                        embed.set_footer(text=f"Uploaded to PyPI on {upload_date}")
```
wookie184, line 77:
```suggestion
                        embed.set_footer(text=f"Latest release uploaded to PyPI on {upload_date}")
```
Thoughts on this to remove ambiguity of it referring to first release? Or do you think it's obvious?

wookie184 (approved):

bot/exts/info/pypi.py
GabeMillikan: Why not more concisely `Released on {date}`? It should be reasonably obvious that this represents the time when the "current upload" of that version was published. After all, this is what's displayed on the pypi website:

```suggestion
                        embed.set_footer(text=f"Released on {upload_date}")
```

image

hedyhli: And since we're modifying the text I'd like to add that explicitly stating "to PyPI" is somewhat unnecessary since it's a PyPI command. @GabeMillikan's suggestion also sounds reasonable to me.

JelmerKorten: I'm happy with the footer text to be just a Released / Uploaded + the date. However I do like to say that I added the "Uploaded to PyPI" after previous comments on that someone said it was unclear as to what it was uploaded. I do agree however that it's enough to just state "Released on <date>", so I will revert it back to that.

bot/exts/info/pypi.py
JelmerKorten: I'm running some tests to see if all are the same format, which I think it should be. It does seem however that not all versions are released on the same date. For now the only example I could find is pygame which has 2 different dates in the urls. 
As far as I'm aware we can just skip the arrow by providing datetime.date objects:
`return max(datetime.strptime(dist["upload_time"], "%Y-%m-%dT%H:%M:%S").date() for dist in data['urls'])`
In which case comparing still works and the latest date in the list of urls will be provided by the command. However, I don't really know the differences between the distributions and why there's a difference in date.

JelmerKorten: Okay I did some digging. I think we can, and should, return the last element in the list. `data["urls"][-1]` is the top element on the downloads page. This will be the "source" distribution. Which is actually when the current version of  the package was released. Later dates inside the urls are actually updated built distributions. So in either way, we'll need to use `min()` instead of `max()`, but as stated above I think this is unnecessary.
So I'm going with that and will await your feedback. 
I do not really like the strptime as I think it can be prone to issues if format is different for some reason. Any suggestions for that?

GabeMillikan: Just use the [dedicated ISO-formatted field](https://warehouse.pypa.io/api-reference/json.html#get--pypi--project_name--json), which can be parsed natively with [`datetime.fromisoformat`](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat).

```suggestion
    return min(datetime.fromisoformat(dist["upload_time_iso_8601"]) for dist in data["urls"]).date()
```

JelmerKorten: Yeah I found that later, thanks!
I've now switched away from the footer and utilising the discord timestamp together with the bot.utilts time.py
It should be here soon, would love your feedback. :)

hedyhli: Since the upload time provided by API is already in ISO 8601, we can just use `min()` on the `upload_time_iso_8601` value and have it use string comparisons rather than with `datetime`, can't we?

> However, I don't really know the differences between the distributions and why there's a difference in date.

Perhaps it's with the files being uploaded one by one during publishing, hence the delay of a few seconds between each. The last one, the source, as sorted alphabetically[^1], being uploaded first might explain why `data["urls"][-1]` could be good enough.

[^1]: I might be wrong here. This is from briefly looking into the source of how data is returned from the API

However take my words with a large grain of salt as I'm not an expert in packaging.

If my theory is right, though, I think `data["urls"][-1]` could do.

FrogsNCode:
hi bro!

PR #2951, opened 2024-03-11
Update configuration tests

shtlrs:
The old test relied on the old system where we loaded config from a yaml file, which ended up doing nothing.

### DIFF
--- a/tests/bot/test_constants.py
+++ b/tests/bot/test_constants.py
@@ -1,53 +1,41 @@
-import inspect
-import typing
-import unittest
+import os
+from pathlib import Path
+from unittest import TestCase, mock
 
-from bot import constants
+from pydantic import BaseModel
 
+from bot.constants import EnvConfig
 
-def is_annotation_instance(value: typing.Any, annotation: typing.Any) -> bool:
-    """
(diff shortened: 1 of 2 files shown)


tests/bot/test_constants.py
```diff
@@ -1,53 +1,43 @@
...
 
-def is_annotation_instance(value: typing.Any, annotation: typing.Any) -> bool:
-    """
-    Return True if `value` is an instance of the type represented by `annotation`.
 
-    This doesn't account for things like Unions or checking for homogenous types in collections.
-    """
-    origin = typing.get_origin(annotation)
+class TestEnvConfig(
+    BaseSettings,
```
ChrisLovering, line 13: Would it be better to use `EnvConfig` from our constants.py here (overwriting env_file)? That way we're testing how our constants actually get loaded, rather than relying on updating this file if we change that in the future.

shtlrs: Sounds good to me, I've done it [here](https://github.com/python-discord/bot/compare/913021d785a5139720500c7c99ea37bd7456684e..3927440ebec8a9bd354933bc4d1c85a0c8c48e75)

PR #2959, opened 2024-03-16
Continually check for posts that need to be closed

ChrisLovering:
This PR adds a check that runs every 5 minutes to look for idle help channels and close them.

This is to cover the case where a Discord outage causes us to miss a help post creation.
I purposefully didn't use the discord.py task system here as this cog already has it's own scheduler and deals with closing all active tasks itself.

This PR also has two small changes to the help cog I noticed while writing & testing this PR.


shtlrs (approved):
LGTM

bot/exts/help_channels/_channel.py
```diff
@@ -235,9 +234,10 @@ async def maybe_archive_idle_post(post: discord.Thread, scheduler: scheduling.Sc
         await _close_help_post(post, closing_reason)
         return
 
-    if has_task:
+    if post.id in scheduler:
+        # Cancel any existing close task
         scheduler.cancel(post.id)
```
MarkKoz, line 239: Isn't this going to cancel itself or does the scheduler remove the ID before running the task? Do we need an `asyncio.shield`?

bot/exts/help_channels/_cog.py
```diff
@@ -38,9 +38,18 @@ async def cog_load(self) -> None:
         self.help_forum_channel = self.bot.get_channel(constants.Channels.python_help)
         if not isinstance(self.help_forum_channel, discord.ForumChannel):
             raise TypeError("Channels.python_help is not a forum channel!")
+        await self.check_all_open_posts_have_close_task()
```
MarkKoz, line 41: While waiting for this, will the cog still be able to handle events or is the cog not considered ready yet? There are two concerns:

1. If this takes a long time, then we may miss events
2. If the cog does handle events during this, then are there any race conditions where multiple tasks may get scheduled for the same post?

bot/exts/help_channels/_channel.py
ChrisLovering: The [bot-core Scheduler](https://github.com/python-discord/bot-core/blob/main/pydis_core/utils/scheduling.py#L170) already wraps scheduled coros in asyncio.sheild so they can't cancel themselves.

I think we should be good here from how I understand this works.

bot/exts/help_channels/_cog.py
ChrisLovering: The cog isn't considered ready yet, so won't handle events.

Looking at what this does, it will only make API requests if the channel isn't cached, or if a post needs to be closed, so it shouldn't take that long in total.

The channel was actually always fetched via an API requesat here, which meant this cog ook a while to load. I just changed that in [this commit](https://github.com/python-discord/bot/pull/2959/commits/b04362048da8703554e4025e6535c628359d4bfb) in this PR.

bot/exts/help_channels/_channel.py
MarkKoz: Yes, I believe you are correct.

bot/exts/help_channels/_cog.py
MarkKoz: I'm not too concerned. Let's just try it and see if we end up with missed events - it's not catastrophic if we do.

MarkKoz (approved):

ChrisLovering:
forcepush is just a rebase onto main with no conflicts

PR #2968, opened 2024-03-24
Remove helpcog dm features

### LINKED ISSUE #1481: Opt-out of DMs when opening a help channel
Some users seem to not like this feature, from DM logs. It might be a good feature to have the ability to opt-out of receiving a DM from the bot when you open a channel.

ChrisLovering:
Closes #1481


Xithrius (approved):
thanks

MarkKoz (approved):

wookie184:
Part of the idea behind DMing authors when they open a help thread was to provide a way for them to look back on past threads they've opened, which is an issue that forum channels didn't really help with.

It's not vital since there are other ways of finding your threads (e.g. searching for pinned messages sent by you in the help forum channel), but since it's a fairly harmless feature and I don't think this use case was discussed I figured I'd mention it.

ChrisLovering:
> Part of the idea behind DMing authors when they open a help thread was to provide a way for them to look back on past threads they've opened, which is an issue that forum channels didn't really help with.
> 
> It's not vital since there are other ways of finding your threads (e.g. searching for pinned messages sent by you in the help forum channel), but since it's a fairly harmless feature and I don't think this use case was discussed I figured I'd mention it.

Quite a few people have been annoyed with the DMs from the bot from that feature specifically. I wdoner if we merge this PR, and add an issue to add it back in a form that can be opt-ed out of?

wookie184:
> Quite a few people have been annoyed with the DMs from the bot from that feature specifically. I wdoner if we merge this PR, and add an issue to add it back in a form that can be opt-ed out of?

Yeah that sounds good, I guess we can see if people mention that they still find it useful once we've removed it and add it back as opt-out or opt-in if that is the case.

PR #2976, opened 2024-03-26
Capture cog load times using sentry performance tracking

ChrisLovering:
I ran this once locally to make sure it all worked, and it looks great https://python-discord.sentry.io/performance/?project=2724195&statsPeriod=14d


mbaruh (approved):

wookie184 (approved):
Nice, I'd thought about using this for commands, but figured it would be too noisy given some commands wait for user responses etc.

This seems good though.

PR #2982, opened 2024-03-26
Fix case sensitivity bug in phishing button

vivekashok1221:
This PR addresses the following bugs introduced in the implementation of PR #2669 :

- https://github.com/python-discord/bot/pull/2982/commits/7569771398a986d85c2dc90c282cc1bc3b2475c5 Fixed a  bug which prevented moderators from adding guild invites of the form 'https://discord.gg/aBcDeFgH' (starting with "https://" AND has uppercase characters in the invite part) to the denylist by clicking on the phishing button. This bug only affected invite links with https:// prefix and so invite links of the type "discord.gg/aBcDeFgH" were unaffected by this. 

 -  https://github.com/python-discord/bot/pull/2982/commits/3996fa6aa64574c9698f1578d47d1178f93efc5e A bug that included the trailing closing parenthesis if the domain was inside parentheses when processing URLs with the phishing button.

- https://github.com/python-discord/bot/pull/2982/commits/3996fa6aa64574c9698f1578d47d1178f93efc5e A bug where  the phishing button would suggest adding, for example, "example.com/abcdef" to the comp-filter-list even if "example.com/aBcDeF" is already present in the filter list.


mbaruh (dismissed):

bot/exts/filtering/_filter_lists/domain.py
```diff
@@ -56,7 +58,7 @@ async def actions_for(
 
         triggers = await self[ListType.DENY].filter_list_result(new_ctx)
         ctx.notification_domain = new_ctx.notification_domain
-        unknown_urls = urls - {filter_.content for filter_ in triggers}
+        unknown_urls = urls - {filter_.content.lower() for filter_ in triggers}
```
vivekashok1221, line 61: Without this, the phishing button would suggest adding `"example.com/abcdef"` to the comp-filter-list even if `"example.com/aBcDeF"` is already present in the filter list.

bot/exts/filtering/_ui/ui.py
```diff
@@ -617,6 +618,9 @@ class AlertView(discord.ui.View):
     def __init__(self, ctx: FilterContext, triggered_filters: dict[FilterList, list[str]] | None = None):
         super().__init__(timeout=ALERT_VIEW_TIMEOUT)
         self.ctx = ctx
+        if "banned" in self.ctx.action_descriptions:
+            # If the user has already been banned, do not attempt to add phishing button
```
wookie184, line 622: Is this added because if they were autobanned then it's probably because the URL they sent is already added as a filter? If so, I think the comment should explain that.

bot/exts/filtering/_filter_lists/domain.py
wookie184: Is there a reason we don't normalise urls to lowercase before adding them to the filter list?

wookie184 (commented):
No issues I can see, just some questions/clarifications

bot/exts/filtering/_filter_lists/domain.py
vivekashok1221: I'm not sure. One potential reason I can think of is that that normalizing URLs to lowercase might lead to loss of distinction between case-sensitive URLs (example.com/aBcD vs example.com/abcd) so we will not be able to visit the website ourselves later (, perhaps, for verification reasons). Although I guess  we could search for the message with the original link but that would be tedious.

bot/exts/filtering/_ui/ui.py
```diff
@@ -681,12 +685,14 @@ def _extract_potential_phish(
...
-                if filter_list.name == "domain" and "discord" in content:  # Leave invites to the invite filterlist.
+                current_content = next(iter(content_list))
+                if filter_list.name == "domain" and re.fullmatch(DISCORD_INVITE, current_content):
+                    # Leave invites to the invite filterlist.
                     continue
                 if encountered:
                     return "", None
                 target_filter_list = filter_list
+                content = current_content
                 encountered = True
```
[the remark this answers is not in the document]
vivekashok1221: The former.

This change fixe the first bug I've mentioned in the PR description.
Invite links of the form `https://discord.gg/aBcDeFgH`  will be present in `content_list` twice. Once as an invite filter and once as a domain filter (..and in this order from my testing. This is why we `continue` to the next iteration of the loop if it's domain filter and the content of the filter matches the pattern of a guild invite.

Before this change, at the point where we encounter `continue`, `content` will be the lowercased version of the link. If we are at the last element of `content_list` we return this lowercased link when we should have actually reverted to the previous value of `content` when we hit the `continue` statement.


Also not directly related to your question: `encountered` is kind of redundant now because we can now simply check if `content` truthy- but I've left it unchanged for code clarity.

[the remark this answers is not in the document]
vivekashok1221: The former.

This change fixe the first bug I've mentioned in the PR description.
Invite links of the form `https://discord.gg/aBcDeFgH`  will be present in `content_list` twice. Once as an invite filter and once as a domain filter (..and in this order from my testing. This is why we `continue` to the next iteration of the loop if it's domain filter and the content of the filter matches the pattern of a guild invite.

Before this change, at the point where we encounter `continue`, `content` will be the lowercased version of the link. If we are at the last element of `content_list` we return this lowercased link when we should have actually reverted to the previous value of `content` when we hit the `continue` statement.


Also not directly related to your question: `encountered` is kind of redundant now because we can now simply check if `content` truthy- but I've left it unchanged for code clarity.

bot/exts/filtering/_ui/ui.py
vivekashok1221:
```suggestion
            # If the user has already been banned, do not attempt to add phishing button since the URL or guild invite
            # is probably already added as a filter
```
something like this?

bot/exts/filtering/_filter_lists/domain.py
wookie184: Fair, not something worth changing in this PR at least.

bot/exts/filtering/_ui/ui.py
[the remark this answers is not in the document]
wookie184: Right, right. Thanks for explaining, that makes sense.

bot/exts/filtering/_ui/ui.py
wookie184: that looks good to me 👍

vivekashok1221: https://github.com/python-discord/bot/pull/2982/commits/dce89a9ab3be38a4f88fab449a7158c6c75989fc

wookie184 (approved):
Looks good

GriceTurrble (approved):

PR #2987, opened 2024-03-29
Reference the `/tag` slash-command instead of the old `!tags` in the footer of tags listings

hedyhli:
I have written a way too long PR title but I think that'll explain it 😄

image image

We don't have a `!tags` anymore. It should reference `/tag` instead.


wookie184 (approved):
Looks good, thanks!

jchristgit (approved):
Thanks!

PR #2989, opened 2024-03-30
Log to #voice-log for passed and failed attempts at voice verification

### LINKED ISSUE #2922: Log indication of someone trying to voice verify.
In previous version of voice verification, users had to run a command which then would be deleted by the bot in order to get a voice verification. Deletion of said message would be logged in a channel, as a side effect that allowed moderators to later tell apart people that spammed in the server for sake of spamming from those that spammed to get voice verified and allowed them to take more appropriate actions. 

With new implementation of voice verification, this process is silent (cannot be seen by moderators) as such, when someone spams it is up to moderators guess based on other factors whether it was related to voice verification.

Ideally bot should have some sort of log when someone tries to voice verify (when they press a button).
2 proposed solutions:
- Send a message to a channel
This is how the previous system worked (deleted message would be logged to a channel).
- `log.info` the button press and its result

Ideally, we should implement both for clarity sake. Channel that could be used for this can be `640292421988646961` (`#voice-log`)

jb3:
This PR adds a new #voice-log type for passing and failing the voice gate.

image

The conditions used are the same that the user will receive in their failure message.

Icons are re-used from DEFCON, as they are pretty and apply to what the voice gate is! As a result though, the constant names are the same, do we care about this? Shall I add duplicate entries to the `Icons` constants area? I think it's minor as if we are updating/removing in future then it will get picked up on anyway.

Closes #2922

### DIFF
--- a/bot/exts/moderation/voice_gate.py
+++ b/bot/exts/moderation/voice_gate.py
@@ -9,9 +9,11 @@ from pydis_core.site_api import ResponseCodeError
 from pydis_core.utils.channel import get_or_fetch_channel
 
 from bot.bot import Bot
-from bot.constants import Channels, MODERATION_ROLES, Roles, VoiceGate as GateConf
+from bot.constants import Channels, Icons, MODERATION_ROLES, Roles, VoiceGate as GateConf
 from bot.log import get_logger
 from bot.utils.checks import InWhitelistCheckFailure
+from bot.utils.messages import format_user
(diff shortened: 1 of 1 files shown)


ChrisLovering (approved):

shtlrs (approved):
💯

wookie184 (approved):
Works nicely, I agree that there's no need to rename or duplicate the icon constants.

PR #2991, opened 2024-04-01
Also sync user's display_name

ChrisLovering:
This updates the sync cog to also sync user's display_names.

This PR also moves the initial sync to 10 seconds after cog start up, to give other cogs a chance to start up.

### DIFF
--- a/bot/exts/backend/sync/_cog.py
+++ b/bot/exts/backend/sync/_cog.py
@@ -1,10 +1,11 @@
 import asyncio
 from typing import Any
 
-from discord import Member, Role, User
+from discord import Guild, Member, Role, User
 from discord.ext import commands
 from discord.ext.commands import Cog, Context
 from pydis_core.site_api import ResponseCodeError
+from pydis_core.utils.scheduling import create_task
 
 from bot import constants
 from bot.bot import Bot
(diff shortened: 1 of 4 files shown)


bot/exts/backend/sync/_cog.py
```diff
@@ -20,33 +21,39 @@ class Sync(Cog):
 
     def __init__(self, bot: Bot) -> None:
         self.bot = bot
+        self.guild: Guild = None
```
jchristgit, line 24: Shouldn't this be `Guild | None`?

bot/exts/backend/sync/_cog.py
```diff
@@ -20,33 +21,39 @@ class Sync(Cog):
...
 
             if attempts == MAX_ATTEMPTS:
                 log.info("Guild not chunked after %d attempts, calling chunk manually.", MAX_ATTEMPTS)
-                await guild.chunk()
+                await self.guild.chunk()
                 break
 
             log.info("Attempt %d/%d: Guild not yet chunked, checking again in 10s.", attempts, MAX_ATTEMPTS)
             await asyncio.sleep(10)
+        create_task(self.sync())
```
jchristgit, line 49: I think we should save this task somewhere. Unless I'm mistaken, this may be garbage collected and die.

bot/exts/backend/sync/_cog.py
```diff
@@ -20,33 +21,39 @@ class Sync(Cog):
...
+
+    async def sync(self) -> None:
+        await asyncio.sleep(10)  # Give time to other cogs starting up
```
jchristgit, line 52: Which other cogs do we depend on? Or is it just to prevent ratelimits?

jb3 (approved):

bot/exts/backend/sync/_cog.py
wookie184: Our util handles keeping it alive https://github.com/python-discord/bot-core/blob/88c2eb6a2f065ee6055db0cf4533065cc95c87c4/pydis_core/utils/scheduling.py#L248

bot/exts/backend/sync/_cog.py
ChrisLovering: only so many tasks can be ran at a time, so this just allows other cogs to startup instead of the syncing cog taking up all processing power. No dependencies.

bot/exts/backend/sync/_cog.py
ChrisLovering: Yea, fixed in https://github.com/python-discord/bot/commit/094e76e432054d713859f15883da63e70f59f76f

bot/exts/backend/sync/_cog.py
wookie184: hmm, it shouldn't make much of a difference given the sync cog would mostly be waiting on network tasks so other things can run, are we sure this actually helps?

PR #2994, opened 2024-04-01
Use a d.py task for checking for necro help posts

ChrisLovering:


shtlrs (approved):

bot/exts/help_channels/_cog.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, bot: Bot):
         self.bot = bot
         self.scheduler = scheduling.Scheduler(self.__class__.__name__)
         self.help_forum_channel: discord.ForumChannel = None
+        self.check_all_open_posts_have_close_task.start()
```
wookie184, line 30: Should this be at the end of `cog_load` to ensure `self.help_forum_channel` is set first?

wookie184 (changes requested):

bot/exts/help_channels/_cog.py
ChrisLovering: Yea good point, lets do that just to be safe. Updated.

wookie184 (approved):
lgtm

PR #2999, opened 2024-04-02
Reply to eval command

### LINKED ISSUE #2382: Reply to !eval message instead of pinging

wookie184:
Closes #2382, supersedes #2402 

Some small changes were required to the mock helpers to ensure the attributes of `MockContext` are consistent with each other.

If the original message is deleted, the response message will still be sent with a mention or reply. I could make it send a mention in this case, but since it's an edge case I don't think it really matters.


bot/exts/utils/snekbox/_cog.py
```diff
@@ -435,7 +435,21 @@ async def send_job(self, ctx: Context, job: EvalJob) -> Message:
...
+                message = ctx.message.to_reference(fail_if_not_exists=False)
+                response = await ctx.send(
+                    msg,
+                    allowed_mentions=allowed_mentions,
+                    view=view,
+                    files=files,
+                    reference=message
+                )
+            else:
+                # The command was directed so a reply wont work, send a normal message with a mention.
```
wookie184, line 450:
```suggestion
                # The command was redirected so a reply wont work, send a normal message with a mention.
```

ChrisLovering (approved):

vivekashok1221 (approved):
LGTM :star2:

PR #3005, opened 2024-04-06
Fix showing `!source` on tags when tags cog is reloaded

### LINKED ISSUE #2022: Incorrect handling of tags by the `!source` command

hedyhli:
Closes #2022

image

---

image

---

### TLDR

Invoking `!source <tag>` should now work even after the tags cog is reloaded.

### Long description

Previously the source command checks for the source object's class after getting it from the SourceType converter. When the tags cog is reloaded, TagIdentifier is redefined. We were checking whether the old TagIdentifier class (the `type()` of an instance of the old TagIdentifier) equals the newly defined TagIdentifier class. This returns false because the TagIdentifier class is redefined when the tags cog is reloaded, which caused tags, correctly identified by Source converter, to be unable to be identified in the source cog.

The fix takes advantage of the fact that the source converter could correctly identify tags objects even after the tags cog reloads, and use enum comparisons rather than `isinstance`/`type()` to obtain the source type of whatever the source converter returns.

Since we're no longer using the source converter as an actual converter, the function is moved to the source cog instead, and it still works fine.


bot/exts/info/source.py
```diff
@@ -33,20 +43,51 @@ async def source_command(
...
             return
 
-        embed = await self.build_embed(source_item)
+        obj, source_type = await self.get_source_object(ctx, source_item)
+        embed = await self.build_embed(obj, source_type)
         await ctx.send(embed=embed)
 
-    def get_source_link(self, source_item: SourceType) -> tuple[str, str, int | None]:
+    @staticmethod
+    async def get_source_object(ctx: commands.Context, argument: str) -> (SourceObject, SourceType):
```
wookie184, line 51:
```suggestion
    async def get_source_object(ctx: commands.Context, argument: str) -> tuple[SourceObject, SourceType]:
```

wookie184 (changes requested):

wookie184 (approved):
Looks great, thanks!

vivekashok1221 (approved):
LGTM :star2:

PR #3017, opened 2024-04-16
Fix watchchannel bugs

### LINKED ISSUE #999: Empty data for watched user
Sentry Issue: [BOT-3A](https://sentry.io/organizations/python-discord/issues/1613515206/?referrer=github_integration)

There are over 500 events of this in Sentry. The events tend to be clumped within a relatively small time range of several hours, and these clumps occur weeks or months apart.

My theory is that a user was unwatched but there was a consumption ongoing for them. So despite being popped from the watched_users dictionary, by other methods (like send_header) attempting to retrieve data for the user, the unwatched user got reintroduced into the dictionary (because it is a `defaultdict') with empty data.

```
KeyError: 'actor'
  File "bot/cogs/watchchannels/watchchannel.py", line 189, in consume_messages
    await self.relay_message(msg)
  File "bot/cogs/watchchannels/watchchannel.py", line 226, in relay_message
    await self.send_header(msg)
  File "bot/cogs/watchchannels/watchchannel.py", line 268, in send_header
    actor = guild.get_member(self.watched_users[user_id]['actor'])

The message queue consume task has failed with:
```

### LINKED ISSUE #1068: Watchchannels KeyError: 'inserted_at'
Sentry Issue: [BOT-7B](https://sentry.io/organizations/python-discord/issues/1800202379/?referrer=github_integration)

This may be related to the user choosing to not updated the cache when invoking the command, but it's odd that any response, even cached ones, would be missing this key; Django adds this to every model. It may have been a bad response (like, non 200) that was cached.

```
KeyError: 'inserted_at'
  File "bot/cogs/watchchannels/bigbrother.py", line 49, in watched_command
    await self.list_watched_users(ctx, oldest_first=oldest_first, update_cache=update_cache)
  File "bot/cogs/watchchannels/watchchannel.py", line 308, in list_watched_users
    inserted_at = user_data['inserted_at']

Error executing command invoked by <REDACTED>: !bigbrother watched False False
```

### LINKED ISSUE #1184: Watch channel queue dictionary changes size during iteration
Sentry Issue: [BOT-84](https://sentry.io/organizations/python-discord/issues/1847909973/?referrer=github_integration)

I suppose it just needs to create another copy for iteration. Is there a more elegant solution?

```
RuntimeError: dictionary changed size during iteration
  File "bot/cogs/watchchannels/watchchannel.py", line 183, in consume_messages
    for user_channel_queues in self.consumption_queue.values():

The message queue consume task has failed with:
```

wookie184:
Fixes 3 related issues, details in commit messages.

Closes #999
Closes #1068
Closes #1184

I decided not to typehint more precisely than `dict` because the rest of the cog doesn't do so and I wanted to keep this change as focused as possible.

### DIFF
--- a/bot/exts/moderation/watchchannels/_watchchannel.py
+++ b/bot/exts/moderation/watchchannels/_watchchannel.py
@@ -62,7 +62,7 @@ class WatchChannel(metaclass=CogABCMeta):
         self.log = logger  # Logger of the child cog for a correct name in the logs
 
         self._consume_task = None
-        self.watched_users = defaultdict(dict)
+        self.watched_users = {}
         self.message_queue = defaultdict(lambda: defaultdict(deque))
         self.consumption_queue = {}
         self.retries = 5
@@ -154,7 +154,7 @@ class WatchChannel(metaclass=CogABCMeta):
             self.log.exception("Failed to fetch the watched users from the API", exc_info=err)
             return False
 
-        self.watched_users = defaultdict(dict)
+        self.watched_users.clear()
 
         for entry in data:
             user_id = entry.pop("user")
--- a/bot/exts/moderation/watchchannels/_watchchannel.py
+++ b/bot/exts/moderation/watchchannels/_watchchannel.py
@@ -185,7 +185,7 @@ class WatchChannel(metaclass=CogABCMeta):
             self.consumption_queue = self.message_queue.copy()
             self.message_queue.clear()
 
-        for user_id, channel_queues in self.consumption_queue.values():
+        for user_id, channel_queues in self.consumption_queue.items():
             for channel_queue in channel_queues.values():
                 while channel_queue:
                     msg = channel_queue.popleft()
--- a/bot/exts/moderation/watchchannels/_watchchannel.py
+++ b/bot/exts/moderation/watchchannels/_watchchannel.py
@@ -185,13 +185,16 @@ class WatchChannel(metaclass=CogABCMeta):
             self.consumption_queue = self.message_queue.copy()
             self.message_queue.clear()
 
-        for user_channel_queues in self.consumption_queue.values():
-            for channel_queue in user_channel_queues.values():
+        for user_id, channel_queues in self.consumption_queue.values():
+            for channel_queue in channel_queues.values():
                 while channel_queue:
                     msg = channel_queue.popleft()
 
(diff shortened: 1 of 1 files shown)


ChrisLovering (approved):

bot/exts/moderation/watchchannels/_watchchannel.py
```diff
@@ -366,8 +369,6 @@ async def prepare_watched_users_data(
     def _remove_user(self, user_id: int) -> None:
         """Removes a user from a watch channel."""
         self.watched_users.pop(user_id, None)
-        self.message_queue.pop(user_id, None)
-        self.consumption_queue.pop(user_id, None)
```
mbaruh, line 370: The watched_users pop here remains, but can't the same issue occur if you're trying to remove a user and list the users at the same time? Maybe it's worth converting the iterator in `prepare_watched_users_data` into a list before looping over it.

For the other two dicts, does this definitely resolve the issue? Looking at `on_message`, you could theoretically have multiple `consume_messages` tasks. If one task clears a queue while another is iterating over it, won't it create the same bug?

mbaruh (changes requested):

bot/exts/moderation/watchchannels/_watchchannel.py
wookie184: > The watched_users pop here remains, but can't the same issue occur if you're trying to remove a user and list the users at the same time? Maybe it's worth converting the iterator in prepare_watched_users_data into a list before looping over it.

I guess so, I'll do that.

> For the other two dicts, does this definitely resolve the issue? Looking at on_message, you could theoretically have multiple consume_messages tasks. If one task clears a queue while another is iterating over it, won't it create the same bug?

As soon as one task is created, the `consuming_messages` check will stop any more from being created while it's running: https://github.com/python-discord/bot/blob/bfd9e446d4d242dd8948057ed3e29faaa699267f/bot/exts/moderation/watchchannels/_watchchannel.py#L76-L90

wookie184: https://github.com/python-discord/bot/pull/3017/commits/25acba42f18b7f5ef813dd7ed5d62cead9f3940b

mbaruh (approved):

PR #3018, opened 2024-04-16
feat: tag loop-remove

GriceTurrble:
Details the gotcha of removing items from a collection while iterating that collection inside a for loop.

This should add tags for `!loop-remove`, `!loop-add`, and `!loop-modify`, all pointing to the same embed.


bot/resources/tags/loop-remove.md
```diff
@@ -0,0 +1,31 @@
...
+    title: "Removing items inside a for loop"
+---
+Avoid removing items from a collection, such as a list, as you iterate that collection in a `for` loop:
+```py
+data = [1, 2, 3, 4]
+for item in data:
+    data.remove(item)
+print(data)  # [2, 4] <-- every OTHER item was removed!
+```
+`for` loops track the index of the current item with a kind of pointer. Removing an element causes all other elements to shift, but the pointer is not changed:
```
wookie184, line 12: Technically it's not the for loop tracking it, it's the list's iterator. I don't think this really matters, but I also think it's not too hard to word it to avoid suggesting that. I'm also not sure how much "pointer" will mean to someone without knowledge in a language like C. Maybe something along the lines of this would be clearer?
```suggestion
While iterating with a for loop, an index tracks the current position in the list. If the list is modified, this index may no longer refer to the same element, causing elements to be repeated or skipped.
```

vivax3794:
I made a gif for this as discussed on discord, you ofc have my full permission to use it. it has a similar quality/fps to the return-gif one (actually slightly bigger).

DefaultTemplate_ManimCE_v0 18 0 post0

mbaruh:
> I made a gif for this as discussed on discord, you ofc have my full permission to use it. it has a similar quality/fps to the return-gif one (actually slightly bigger).

That's a really cool GIF. One thing is that I think it's better to show on the right what's being printed, to clarify that the result is "2\n4\n". Currently it looks like it ends too soon before I can understand what you're showing.

ChrisLovering:
Tag looks great, lets merge this. If we want to add a gif we can do so in a later PR.

PR #3021, opened 2024-04-17
Use correct key to check if infraction has expiry

### LINKED ISSUE #2559: Temporary bans are incorrectly assumed permanent when infractions overlap
If a user is temporarily banned, and a moderator attempts to permanently ban said user, the following code will never return anything but None.

https://github.com/python-discord/bot/blob/f94505ac975888b259ea2d07b4442d6d908352f8/bot/exts/moderation/infraction/infractions.py#L490

This is because the site api does not return a `duration_or_expiry` field, it's named `expires_at`. This prevents us from re-banning a temporary banned user to permanent without unbanning them first or editing the original infraction. 

#2234 updated this code to the newer name, but doesn't update returned `get_active_infraction` to alter the returned dictionary. 

https://github.com/python-discord/bot/blob/0921838112120355aed0936923983b545320edef/bot/exts/moderation/infraction/_utils.py#L159 here

This seems like something we should be using dataclasses of a sort for, so that this can be type-checked?

wookie184:
Closes #2559 

Preserve existing behaviour of not allowing a permaban to override a tempban


vivekashok1221 (approved):
Looks good to me :star2:

Xithrius (approved):
:shipit:

PR #3022, opened 2024-04-18
Notify help threads if the owner is no longer on the server

### LINKED ISSUE #2692: Automatically close help thread on ban
It's probably a rare occurrence but it happened today (https://discord.com/channels/267624335836053506/1133167276401643630) where a user who had an open help thread got banned in a different channel. This left the people helping confused for a bit as to where the user went. It would be a nice feature if the bot would automatically close any open help thread (with a message indicating user has 'decided to pursue opportunities at other servers' or similar). That way the people helping don't have to notice they were banned.

wookie184:
Closes #2692 

Happy to discuss if you think a different behaviour (e.g. closing automatically) would still be better.


bot/exts/help_channels/_cog.py
```diff
@@ -141,3 +143,17 @@ async def new_post_message_listener(self, message: discord.Message) -> None:
...
+        for thread in self.help_forum_channel.threads:
+            if thread.owner_id != member.id:
+                continue
+
+            if thread.archived:
+                continue
+
+            log.debug(f"Notifying help thread {thread.id} that owner {member.id} has left the server.")
+            with contextlib.suppress(discord.NotFound):
+                await thread.send(":warning: The owner of this post has left the server.")
```
hedyhli, line 159: A slight nit - the rest looks good as a code review 👍 

Looking at the one instance this was needed from the linked issue, it may not always (or even likely) be the owner deliberately leaving the server rather than getting a ban in such cases. How's this instead?
```suggestion
                await thread.send(":warning: The owner of this post is no longer in the server.")
```

wookie184: Yeah, that's better 👍

wookie184: Actually i'll change the log above too so it matches.

wookie184: done

vivekashok1221 (approved):
LGTM :star2:

Xithrius:
I'm for the auto-close feature, but not instant. Perhaps we could send a message saying that the thread will close in a minute instead? Such that if a user gets banned within a thread, messages can still go through for a bit.

In addition to that, perhaps an :x: could be added to the sent message to prevent the thread from closing. Might be out of scope for this PR though.

vivekashok1221:
> I'm for the auto-close feature, but not instant. Perhaps we could send a message saying that the thread will close in a minute instead? Such that if a user gets banned within a thread, messages can still go through for a bit.
> 
> In addition to that, perhaps an ❌ could be added to the sent message to prevent the thread from closing. Might be out of scope for this PR though.

I'm mostly not in favour of closing the help thread unless the helpee has been banned- but I can probably be convinced otherwise.

wookie184:
The initial reasoning behind the issue was to remove the confusion on where the user went if someone left/was banned, and I think sending a message is the simplest and least intrusive way of solving that.

I don't think auto-closing is really worth the extra complexity given it's such an edge case (and it could *potentially* be unwanted in some cases) unless there's a reason for it separate to the original motivation of the issue.

Xithrius (approved):
True. Just the message will do then. If we really want to add the auto-closing feature, we can always do it later.

jahrules:
@wookie184 today I saw the bot post the departed message twice. Any idea why?

https://discord.com/channels/267624335836053506/1237066212593242163

[
Screenshot_20240506-115343
](url)

wookie184:
> @wookie184 today I saw the bot post the departed message twice. Any idea why?
> 
> https://discord.com/channels/267624335836053506/1237066212593242163
> 
> [ Screenshot_20240506-115343 ](url)

It looks like they left, rejoined, and then left again pretty much straight away. We could add another message saying if they rejoined (unless we think it's so rare it's not worth it).

jahrules:
Or maybe a timeout? Could someone theoretically use this behavior to time
the bot out by leaving and rejoing real fast?

vivekashok1221:
> Or maybe a timeout? Could someone theoretically use this behavior to time the bot out by leaving and rejoing real fast?

I think the bot should be able to handle it. We already log whenever a user joins or leaves the server in a mod-only log channel. This additional message in the help thread shouldn't be much of a concern.

PR #3024, opened 2024-04-18
Remove trailing line from System Python tag

BruDriguezz:
As per @import-pandas-as-numpy's request, this PR removes the trailing line on the System Python's tag.


wookie184:
The trailing newline is necessary for linting and shouldn't affect how the tag looks. I believe the suggestion was to remove some of the empty lines in the tag (e.g. between bullet points).

import-pandas-as-numpy:
> The trailing newline is necessary for linting and shouldn't affect how the tag looks. I believe the suggestion was to remove some of the empty lines in the tag (e.g. between bullet points).

@BruDriguezz 
This is correct. Apologies if that wasn't clear. The embed takes up a significant amount of white space when posted: 
image

This may be a result of Discord's native list formatting (I'm unsure, to be honest) but we should review this, as the information being portrayed does not warrant the level of whitespace this tag takes.

Issue #3045, opened 2024-05-11
!eval output and pastebin content are same when the output is 11 lines


vivekashok1221:
If the output generated by snekbox has 11 lines, the output sent to Discord as well as the contents uploaded to our pastebin is the exact same.
![image](https://github.com/python-discord/bot/assets/37447267/4bbbf92e-c51a-43b2-8fc7-5513aebfa1a2)
![image](https://github.com/python-discord/bot/assets/37447267/dc80ba0f-0a7d-42c0-820e-b0457beace76)

This is because we always [truncate the output to 11 lines](https://github.com/python-discord/bot/blob/10a57366f4243ce5d22c6f868f0545f678fe21e3/bot/exts/utils/snekbox/_cog.py#L272), but we [check if the number of lines is more than 10](https://github.com/python-discord/bot/blob/10a57366f4243ce5d22c6f868f0545f678fe21e3/bot/exts/utils/snekbox/_cog.py#L275) to decide if we need upload to the pastebin. Basically, the current logic for deciding if the output is uploaded to pastebin or not relies on the extra 11th line of output. This works when the output has 12 lines or more (11 lines displayed, full output sent to pastebin), or if it the output has 10  lines or less. But when it is 11, the 11th line is included in the output as well as uploaded to the pastebin (we truncate to 11 lines but upload to the pastebin if the output has more than 10 lines). I'd say this is because of the incorrect fix for #709 but don't quote me on that.

I have a fix locally which I can push if this issue is approved and assigned to me.

Another thing I noticed is that we display 11 lines in Discord when `MAX_OUTPUT_BLOCK_LINES` is 10. I initially thought this was introduced in [this commit](https://github.com/python-discord/bot/commit/3d19133d70334eb9a46b4788f95e3c5ac44cc41d) when we switched from `str.split("\n")` to `str.splitlines()` but it seems like ever since we [started truncating eval output](https://github.com/python-discord/bot/commit/9b2312e2f5395ca8c3f9e7e4af53aeddc2a0f7e5), we've been [displaying 11 lines](https://discord.com/channels/267624335836053506/303906556754395136/466902066225217546).  If we indeed intend to display 11 lines of output, we should consider updating MAX_OUTPUT_BLOCK_LINES. In my fix, I've truncated the output to 10 lines

wookie184:
Is this the same as https://github.com/python-discord/bot/issues/2492?

vivekashok1221:
I didn't notice that issue, you're right.

[...]

PR #3046, opened 2024-05-12
Don't upload to pastebin when no truncation

vivekashok1221:
See issue #3045 

New behaviour:
image

### DIFF
--- a/bot/exts/utils/snekbox/_cog.py
+++ b/bot/exts/utils/snekbox/_cog.py
@@ -266,14 +266,14 @@ class Snekbox(Cog):
         truncated = False
         lines = output.splitlines()
 
-        if len(lines) > 1:
-            if line_nums:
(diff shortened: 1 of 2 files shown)


ChrisLovering (approved):

wookie184 (commented):
If we're adding a line to say that the message was truncated, but we're only truncating one line, wouldn't it be better to not truncate and the result would be the same number of lines?

That's the proposal I put forward here https://github.com/python-discord/bot/pull/2498#issuecomment-1493339577, essentially, for `max_lines=10`:
- 9 lines: not truncated
- 10 lines: not truncated
- 11 lines: truncated to 9 lines, plus the truncation message.

This is an improvement though, i'm happy to approve if you'd rather leave it like this for now.

wookie184 (approved):
Perfect, thanks!

PR #3050, opened 2024-05-13
Check length of nomination reason before attempting to send to site API

jb3:
This ensures we will not attempt to send a reason that is longer than
the REASON_MAX_CHARS value, which can quickly be reached through the
addition of message links in context.


ChrisLovering (approved):

wookie184 (approved):
👍

PR #3056, opened 2024-05-17
Allow for more than 3 nomination votes if others are ticketed

jb3:
This changes how we calculate whether we are ready for another vote in
the talentpool autonomination system.

Instead of hard-locking at 3 votes in the pool, we factor in whether an
admin-task has been created for the vote (signified by a reaction with
the :ticket: emoji).

We allow for up to 10 votes to be in the pool total, which in theory
allows 3 active votes and 7 votes waiting on the admin-task
completion (though we should never hit this threshold realistically).

This should allow for voting to be more or less continuous and ensure
there is always new votes in the pool and we are not blocked on the
completion of a Helpering/waiting for a Helper candidate to be available
for a new vote to start.

The thresholds are based on those suggested in the nomination-discussion
moderator channel.

### DIFF
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -33,6 +33,8 @@ MAX_EMBED_SIZE = 4000
 
 # Maximum number of active reviews
 MAX_ONGOING_REVIEWS = 3
+# Maximum number of total reviews
+MAX_TOTAL_REVIEWS = 10
 # Minimum time between reviews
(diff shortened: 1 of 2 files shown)


wookie184 (approved):
Looks good, [the tests](https://github.com/python-discord/bot/blob/1e73aa4741c6d682c67eeaacf6d53cac7b006286/tests/bot/exts/recruitment/talentpool/test_review.py#L62) could maybe do with updating but up to you.

jb3:
Good spot on the tests, I've updated them now to add 2 new cases for "above active, below total" and "above active, above total".

PR #3063, opened 2024-05-21
Post nomination entries into the voting thread instead of the root

jb3:
This change moves us from posting multiple messages in the
nomination-voting channel to instead posting all entries into the voting
thread.

This makes the nomination process guaranteed to be contained within one
root message and one channel, instead of splitting into multiple
messages in the root if we went over the character limit.

It greatly simplifies both the creation and the archival of nomination
votes as we only need to handle one message.

It also means that all nominations are guaranteed to be visible in the
voting thread.

### DIFF
--- a/bot/exts/recruitment/talentpool/_review.py
+++ b/bot/exts/recruitment/talentpool/_review.py
@@ -1,22 +1,20 @@
-import asyncio
 import contextlib
 import random
 import re
-import textwrap
 import typing
 from collections import Counter
 from datetime import UTC, datetime, timedelta
 
 import discord
 from async_rediscache import RedisCache
-from discord import Embed, Emoji, Member, Message, NotFound, PartialMessage, TextChannel
+from discord import Embed, Emoji, Member, NotFound, PartialMessage
 from pydis_core.site_api import ResponseCodeError
 from pydis_core.utils.channel import get_or_fetch_channel
 from pydis_core.utils.members import get_or_fetch_member
 
 from bot.bot import Bot
 from bot.constants import Channels, Colours, Emojis, Guild, Roles
-from bot.exts.recruitment.talentpool._api import Nomination, NominationAPI
+from bot.exts.recruitment.talentpool._api import Nomination, NominationAPI, NominationEntry
 from bot.log import get_logger
 from bot.utils import time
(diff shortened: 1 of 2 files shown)


bot/exts/recruitment/talentpool/_review.py
```diff
@@ -46,7 +45,7 @@
 # The higher this is, the lower the effect of review age. At 1, age and number of entries are weighted equally.
 REVIEW_SCORE_WEIGHT = 1.5
 
-# Regex for finding the first message of a nomination, and extracting the nominee.
+# Regex for finding the a nomination, and extracting the nominee.
```
wookie184, line 48:
```suggestion
# Regex for finding a nomination, and extracting the nominee.
```

wookie184 (changes requested):
The `get_review` command will need updating now that `make_review` returns another value:
https://github.com/python-discord/bot/blob/7437d98bfa8287a91857f860567efd748e38e438/bot/exts/recruitment/talentpool/_cog.py#L766
(I don't mind what you make the command do, or even if you want to remove it instead, I don't think it's that useful).

Not tested yet but looking pretty solid, thanks 💪

wookie184 (approved):
Works like a charm

bot/exts/recruitment/talentpool/_review.py
```diff
@@ -367,34 +372,36 @@ async def archive_vote(self, message: PartialMessage, passed: bool) -> None:
...
+        body = f"{nominations}\n\n{activity}\n\n{infractions}"
         if prev_nominations:
             body += f"\n\n{prev_nominations}"
         return body
 
+    async def _nominations_review(self, nomination: Nomination) -> str:
+        """Format a brief summary of how many nominations in this voting round the nominee has."""
+        plural = len(nomination.entries) > 1
+
+        return f"They have **{len(nomination.entries)}** nomination{'s' if plural else ''} this round."
```
Xithrius, line 403: Perhaps?
```suggestion
        entry_count = len(nomination.entries)

        return f"They have **{entry_count}** nomination{'s' if entry_count > 1 else ''} this round."
```

Xithrius (approved):
Yes.

PR #3064, opened 2024-05-21
Relay creations and amendments to nominations to vote threads

jb3:
If a nomination is actively being voted on and a staff member creates, updates
or attaches a message to their existing nomination we currently do not relay
that to the vote thread.

This PR relays any of these events to the vote thread so they are not lost.

This also allows moderators to use the message context menu command to attach
any relevant messages they find whilst reviewing into the active voting thread.

### DIFF
--- a/bot/exts/recruitment/talentpool/_cog.py
+++ b/bot/exts/recruitment/talentpool/_cog.py
@@ -41,10 +41,12 @@ class NominationContextModal(discord.ui.Modal, title="New Nomination"):
         max_length=REASON_MAX_CHARS - 110
     )
 
-    def __init__(self, api: NominationAPI, message: discord.Message, noms_channel: discord.TextChannel):
+    def __init__(self, cog: "TalentPool", message: discord.Message, noms_channel: discord.TextChannel):
         self.message = message
-        self.api = api
+        self.api = cog.api
         self.noms_channel = noms_channel
+        self.bot = cog.bot
+        self.cog = cog
 
         super().__init__()
 
@@ -77,7 +79,7 @@ class NominationContextModal(discord.ui.Modal, title="New Nomination"):
         )
 
         noms_channel_message = (
-            f":star: {interaction.user.mention} has nominated "
+            f":new: {interaction.user.mention} has nominated "
             f"{self.message.author.mention} from {self.message.jump_url}"
         )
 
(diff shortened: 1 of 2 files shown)


bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -369,6 +372,27 @@ async def list_nominations(
...
+        for nomination in nominations:
+            if nomination.thread_id:
+                found_thread_id = nomination.thread_id
+
+        if found_thread_id:
+            log.debug(f"Found thread ID for {nominee_id} nomination, relaying new context.")
+            thread = await get_or_fetch_channel(bot, found_thread_id)
+
+            await thread.send(update)
+
```
wookie184, line 395: If we added an `api.get_active_nomination(user_id) -> Nomination | None` function, I think this could be made a lot cleaner. I'd been thinking of suggesting that anyway as it'd be useful in quite a few other places as well.
```suggestion
        nomination = await api.get_active_nomination(nominee_id)
        
        if nomination and nomination.thread_id:
            log.debug(f"Found thread ID for {nominee_id} nomination, relaying new context.")
            thread = await get_or_fetch_channel(bot, nomination.thread_id)

            await thread.send(update)

```

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -41,10 +41,11 @@ class NominationContextModal(discord.ui.Modal, title="New Nomination"):
         max_length=REASON_MAX_CHARS - 110
     )
 
-    def __init__(self, api: NominationAPI, message: discord.Message, noms_channel: discord.TextChannel):
+    def __init__(self, api: NominationAPI, message: discord.Message, noms_channel: discord.TextChannel, bot: Bot):
```
wookie184, line 44: Since it's taking the `api`, `bot`, and calling methods on the cog, it may be cleaner to make it take the cog itself as an argument instead (and not have `maybe_relay_update` as a staticmethod).

I don't really mind though.

wookie184 (changes requested):
Looks good, just some ideas to make it cleaner that i'd like your thoughts on.

wookie184 (approved):
LGTM 🔥

bot/exts/recruitment/talentpool/_cog.py
```diff
@@ -449,14 +467,20 @@ async def _nominate_context_callback(self, interaction: discord.Interaction, mes
                 ":white_check_mark: Existing nomination updated",
                 ephemeral=True
             )
+
+            await self.maybe_relay_update(
+                message.author.id,
+                f":new: <@{interaction.user.id}> has attached {message.jump_url} to their nomination"
```
etrotta, line 473: Maybe use `.mention` for consistency

jb3: Good call, fixed in https://github.com/python-discord/bot/compare/7df00e07417b06f2f5646698fc5049894a03ebcf..e710d9f2741dfe5c3b9b34cd1226a0cec9088820

Xithrius (approved):

PR #3065, opened 2024-05-22
Support events that span from one year to another

### LINKED ISSUE #2035: Allow events across multiple years
**See python-discord/branding#171.**

Currently, at least according to the validation in [events/validation.py](https://github.com/python-discord/branding/blob/main/events/validation.py) in the branding repo, events can't go from December to January as the end date is earlier than the start date (meaning this only affects New Year's).

This may or may not need some changes to the bot as well, but it's definitely worth some time and consideration as we still have a full year left to implement this.

wookie184:
Closes #2035

### DIFF
--- a/bot/exts/backend/branding/_repository.py
+++ b/bot/exts/backend/branding/_repository.py
@@ -230,7 +230,17 @@ class BrandingRepository:
 
         for event in available_events:
             meta = event.meta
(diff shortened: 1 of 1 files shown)


hedyhli (approved):

vivekashok1221 (approved):
LGTM

PR #3071, opened 2024-05-23
Fix faulty hyperlink in charinfo command.

LeoCx1000:
Give a default name of "Name not found" if `unicodedata` is unable to find one, to fix the faulty hyperlinks.

image


ChrisLovering (approved):
Looks good, thanks!

wookie184 (approved):
Thanks for fixing this!

PR #3100, opened 2024-06-21
Skip line if no match against inventory line regex

jb3:
We currently have an issue on bot (BOT-3VH on Sentry) which is failing to parse
the sympy docs due to a malformed object inventory.

This looks to be caused by multi-line figure descriptions or something similar
which are creating lines that do not match our regex.

Ignoring these lines is fine as they likely provide no additional context
anyway, and the relevant symbol is still parsed as normal and links to the
original docs.


mbaruh (approved):
Legend

ChrisLovering (approved):

PR #3120, opened 2024-07-13
Suppress ResponseCodeErrors when fetching alts in !user

jb3:
Restores previous behaviour that allows the `!user` command to look up users who
are not in the guild.

When the alts feature was implemented we added a new API call to
`/bot/users/{user_id}`, which was fine, but it meant that if that 404'd and a
ResponseCodeError was raised then the bot would start returning erroneous
responses that implied a user does not exist, when they did exist, they were
just not in our site database.

### DIFF
--- a/bot/exts/info/information.py
+++ b/bot/exts/info/information.py
@@ -343,8 +343,16 @@ class Information(Cog):
 
     async def user_alt_count(self, user: MemberOrUser) -> tuple[str, int | str]:
         """Get the number of alts for the given member."""
-        resp = await self.bot.api_client.get(f"bot/users/{user.id}")
-        return ("Associated accounts", len(resp["alts"]) or "No associated accounts")
+        try:
+            resp = await self.bot.api_client.get(f"bot/users/{user.id}")
+            return ("Associated accounts", len(resp["alts"]) or "No associated accounts")
+        except ResponseCodeError as e:
+            # If user is not found, return a soft-error regarding this.
+            if e.response.status == 404:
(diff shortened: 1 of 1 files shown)


mbaruh (approved):
This works, although it means the line won't be added to the embed, whereas adding a try-except inside the function would allow for something to still be displayed

jb3:
Good call, amending.

bot/exts/info/information.py
```diff
@@ -323,7 +323,15 @@ async def create_user_embed(self, ctx: Context, user: MemberOrUser, passed_as_me
...
-            fields.append(await self.user_alt_count(user))
+            try:
+                fields.append(await self.user_alt_count(user))
+            except ResponseCodeError as e:
+                if e.response.status == 404:
+                    fields.append(
+                        ("Associated accounts", "User not found in site database.")
+                    )
+                else:
+                    raise e
```
vivekashok1221, line 334: Can we consider handling this in `self.user_alt_account`? 

Or,  we could document in `self.user_alt_account` that a `ResponseCodeError` will be raised if the user is not found in the database (have I been using too much Java?) .

vivekashok1221 (changes requested):
I have a single comment. I'm okay with not actioning it, and I can approve the PR if you decide not to address it.

bot/exts/info/information.py
jb3: Changed to this way in latest force push!

vivekashok1221 (approved):
LGTM!

PR #3145, opened 2024-08-13
Log and retry OT renaming if Discord returns an error

ChrisLovering:


AaritDutta:
Nice Work!!✨✨

bot/exts/fun/off_topic_names.py
```diff
@@ -61,9 +61,18 @@ async def update_names(self) -> None:
...
-        await channel_2.edit(name=OTN_FORMATTER.format(number=2, name=channel_2_name))
+        try:
+            await channel_0.edit(name=OTN_FORMATTER.format(number=0, name=channel_0_name))
+            await channel_1.edit(name=OTN_FORMATTER.format(number=1, name=channel_1_name))
+            await channel_2.edit(name=OTN_FORMATTER.format(number=2, name=channel_2_name))
+        except HTTPException:
+            await self.bot.get_channel(Channels.mod_meta).send(
+                ":exclamation: Got an error trying to update OT names to"
+                f"{channel_0_name}, {channel_1_name} and {channel_2_name}."
+                "\nI will retry with new names, but please delete the ones that sound bad."
```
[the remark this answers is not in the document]
ChrisLovering: Yeah, Discord frequentently updates their filter of allowed channel names

bot/exts/fun/off_topic_names.py
```diff
@@ -61,9 +61,18 @@ async def update_names(self) -> None:
...
-        await channel_1.edit(name=OTN_FORMATTER.format(number=1, name=channel_1_name))
-        await channel_2.edit(name=OTN_FORMATTER.format(number=2, name=channel_2_name))
+        try:
+            await channel_0.edit(name=OTN_FORMATTER.format(number=0, name=channel_0_name))
+            await channel_1.edit(name=OTN_FORMATTER.format(number=1, name=channel_1_name))
+            await channel_2.edit(name=OTN_FORMATTER.format(number=2, name=channel_2_name))
+        except HTTPException:
+            await self.bot.get_channel(Channels.mod_meta).send(
+                ":exclamation: Got an error trying to update OT names to"
+                f"{channel_0_name}, {channel_1_name} and {channel_2_name}."
```
vivekashok1221, line 71: Wouldn't it be better if we edited the channel names one by one in different `try`-blocks, storing any invalid names in a list, and then including only the invalid names in this message?

- In the current approach, if the first name is invalid, the other two names won't be used. However, I believe they will still be marked as inactive in the database, putting the valid names out of rotation for no reason.
- It will also eliminate the guesswork in identifying which names are invalid, especially when it isn't obvious from the name (although I guess you could simply check which channel name wasn't changed).

vivekashok1221 (changes requested):
In the end, the greatest snowball isn't a snowball at all. It's fear.

bot/exts/fun/off_topic_names.py
ChrisLovering: yes, but this was a quick fix that is good enough, without rewriting a lot

jb3 (approved):

PR #3165, opened 2024-09-03
General refactor of bot.py, log.py and pagination.py

nerd-bear:
There are a few small changes that I made thought these files.
For the log file:

> Broken down the setup function to be more understandable and readable.
> Improved name of the _set_trace_loggers function and it now uses a single underscore, which is more conventional for internal use.
> Added an early return in the _set_trace_loggers function. Now a early return is called if level_filter is false, making it more readable.
> This goes for all the files, added consistent spacing and indentation to everything.
> Also added a bit more type hits for new contributors to better understand the code base. 

For the bot file:
 
> Switched "from sys import exception" with "from sys import exc_info", because the exc_info() function gives more detailed information.
> Changed new_scope to push_scope, because its the more used and recommended method.
> Changed the __init__ method to include the base exceptions message in the new exception.
> Removed empty __init__ method from Bot class as it was just calling the superclass constructor without any changes.
> Improved the while loop and made it a for loop with a range, making it more 'Pythonic' if you will and easier to read.

And finally for the pagination.py file:

> Both the class and the paginate method now have more info filled docstrings that explain their meaning and usage.
> The parameters in the paginate method signature are now aligned for better readability.
> The custom PaginationEmojis instance is now created within the method body, making it easier to modify if needed in the future.
> The parameters are passed directly to the superclass method, reducing redundancy and improving maintainability.

I hope this passes the style guide lines and I hope this change is not to small to be ignored!

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -1,33 +1,24 @@
 import asyncio
 import contextlib
-from sys import exception
-
+from sys import exc_info
 import aiohttp
 from discord.errors import Forbidden
 from pydis_core import BotBase
 from pydis_core.utils.error_handling import handle_forbidden_from_block
-from sentry_sdk import new_scope, start_transaction
-
+from sentry_sdk import push_scope, start_transaction
 from bot import constants, exts
 from bot.log import get_logger
 
 log = get_logger("bot")
 
-
 class StartupError(Exception):
     """Exception class for startup errors."""
-
     def __init__(self, base: Exception):
-        super().__init__()
+        super().__init__(str(base))
         self.exception = base
-
 
 class Bot(BotBase):
     """A subclass of `pydis_core.BotBase` that implements bot-specific functions."""
-
-    def __init__(self, *args, **kwargs):
-
-        super().__init__(*args, **kwargs)
 
     async def load_extension(self, name: str, *args, **kwargs) -> None:
         """Extend D.py's load_extension function to also record sentry performance stats."""
@@ -36,17 +27,13 @@ class Bot(BotBase):
 
     async def ping_services(self) -> None:
         """A helper to make sure all the services the bot relies on are available on startup."""
-        # Connect Site/API
-        attempts = 0
-        while True:
+        for attempt in range(1, constants.URLs.connect_max_retries + 1):
             try:
-                log.info(f"Attempting site connection: {attempts + 1}/{constants.URLs.connect_max_retries}")
+                log.info(f"Attempting site connection: {attempt}/{constants.URLs.connect_max_retries}")
                 await self.api_client.get("healthcheck")
-                break
-
+                return
             except (aiohttp.ClientConnectorError, aiohttp.ServerDisconnectedError):
-                attempts += 1
-                if attempts == constants.URLs.connect_max_retries:
+                if attempt == constants.URLs.connect_max_retries:
                     raise
                 await asyncio.sleep(constants.URLs.connect_cooldown)
 
@@ -57,23 +44,18 @@ class Bot(BotBase):
 
     async def on_error(self, event: str, *args, **kwargs) -> None:
         """Log errors raised in event listeners rather than printing them to stderr."""
-        e_val = exception()
-
-        if isinstance(e_val, Forbidden):
+        _, error, _ = exc_info()
+        if isinstance(error, Forbidden):
             message = args[0] if event == "on_message" else args[1] if event == "on_message_edit" else None
-
             with contextlib.suppress(Forbidden):
-                # Attempt to handle the error. This reraises the error if's not due to a block,
-                # in which case the error is suppressed and handled normally. Otherwise, it was
-                # handled so return.
-                await handle_forbidden_from_block(e_val, message)
+                await handle_forbidden_from_block(error, message)
                 return
 
         self.stats.incr(f"errors.event.{event}")
-
-        with new_scope() as scope:
+        with push_scope() as scope:
             scope.set_tag("event", event)
-            scope.set_extra("args", args)
-            scope.set_extra("kwargs", kwargs)
-
+            scope.set_extras({
+                "args": args,
+                "kwargs": kwargs
+            })
             log.exception(f"Unhandled exception in {event}.")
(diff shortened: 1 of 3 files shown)


bot/bot.py
```diff
@@ -57,23 +44,18 @@ async def setup_hook(self) -> None:
...
         """Log errors raised in event listeners rather than printing them to stderr."""
-        e_val = exception()
-
-        if isinstance(e_val, Forbidden):
+        _, error, _ = exc_info()
+        if isinstance(error, Forbidden):
             message = args[0] if event == "on_message" else args[1] if event == "on_message_edit" else None
-
             with contextlib.suppress(Forbidden):
-                # Attempt to handle the error. This reraises the error if's not due to a block,
```
shtlrs, line 66: I guess you could leave the comment.

bot/log.py
```diff
@@ -63,29 +38,50 @@ def setup_sentry() -> None:
...
+
+    if level_filter.startswith("*"):
+        get_logger().setLevel(core_logging.TRACE_LEVEL)
+    elif level_filter.startswith("!"):
+        get_logger().setLevel(core_logging.TRACE_LEVEL)
+        for logger_name in level_filter.strip("!,").split(","):
+            get_logger(logger_name).setLevel(logging.DEBUG)
+    else:
+        for logger_name in level_filter.strip(",").split(","):
+            get_logger(logger_name).setLevel(core_logging.TRACE_LEVEL)
```
shtlrs, line 87: I thought ruff would complain once you have execution paths that do not return none when you have at least one that does. surprising.

n0Oo0Oo0b:
We appreciate your effort into this PR, but there are some issues with the changes you've proposed here:

---

> Broken down the setup function to be more understandable and readable.

I don't see the benefit of this, especially considering the docstring of `setup()` already indicates that it's meant to "set up loggers". If you find readability to be an issue in the existing code (IMO it's fine as it is), labeling each section of the function with a comment would be more reasonable.

> Improved name of the `_set_trace_loggers` function and it now uses a single underscore, which is more conventional for internal use.

I'm not sure what change you're referring to, this function was (and still is) prefixed with a single underscore.

> Added an early return in the _set_trace_loggers function. Now a early return is called if level_filter is false, making it more readable.

I'm fine with this change, but one might argue that the existing code would be easier to work with if something else needed to be added in the future.

> This goes for all the files, added consistent spacing and indentation to everything.

Consistent with what? We use `ruff` to format code in our repositories, so the formatting was already perfectly consistent before. Changing the formatting broke this, which is why merging of this PR is currently blocked.

> Also added a bit more type hits for new contributors to better understand the code base.

Again, I'm not sure what change this is referring to.

> Switched "from sys import exception" with "from sys import exc_info", because the exc_info() function gives more detailed information.

The additional info isn't needed in this case, so it seems pretty pointless to change this.

> Changed new_scope to push_scope, because its the more used and recommended method.

According to the [sentry docs](https://docs.sentry.io/platforms/python/enriching-events/scopes/#changing-the-scope), `push_scope` looks like it's for an older version so it shouldn't even exist anymore.

> Removed empty init method from Bot class as it was just calling the superclass constructor without any changes.

Looks alright, I don't have any objections on this.

> Improved the while loop and made it a for loop with a range, making it more 'Pythonic' if you will and easier to read.

Looks okay to me, but there might have been some reason it was like that before. A third opinion on this would be nice.

> Both the class and the paginate method now have more info filled docstrings that explain their meaning and usage.

I think the existing docstring deferring to the superclass is better since it's one less change if that API is changed down the line. Also note that function docstrings should complete the sentence "This function/method ..." (i.e. should start with a verb).

> The parameters in the paginate method signature are now aligned for better readability.

I'm okay with this as long as ruff is happy with it.

> The custom PaginationEmojis instance is now created within the method body, making it easier to modify if needed in the future.

This could easily be changed later if required, I'm not sure why this needs to be changed now, especially for something as simple as this.

>  The parameters are passed directly to the superclass method, reducing redundancy and improving maintainability.

Assuming by this you mean forwarding `**kwargs` to `super().paginate()`, I don't mind this, although there isn't a clear need for it.

---

If you wish to continue contributing to our projects, I would recommend re-reading our [contributing guidelines](https://www.pythondiscord.com/pages/guides/pydis-guides/contributing/contributing-guidelines/) to ensure your PRs are up to standard and are heading in the right direction.

bot/pagination.py
```diff
@@ -28,20 +24,26 @@ async def paginate(
...
 
-        Acts as a wrapper for the super class' `paginate` method to provide the pagination emojis by default.
+        This method wraps the superclass's `paginate` method, providing custom pagination emojis by default.
+        All parameters are passed directly to the superclass method.
 
-        Consult the super class's `paginate` method for detailed information.
+        Returns:
+            discord.Message | None: The message object if pagination is successful, None otherwise.
         """
+        pagination_emojis = PaginationEmojis(delete=Emojis.trashcan)
```
shtlrs, line 43: You don't really need the variable if it's only being used to pass it as argument to the `pagination_emojis` param.

bot/bot.py
```diff
@@ -1,52 +1,39 @@
...
+                log.info(f"Attempting site connection: {attempt}/{constants.URLs.connect_max_retries}")
                 await self.api_client.get("healthcheck")
-                break
-
+                return
             except (aiohttp.ClientConnectorError, aiohttp.ServerDisconnectedError):
-                attempts += 1
-                if attempts == constants.URLs.connect_max_retries:
+                if attempt == constants.URLs.connect_max_retries:
                     raise
```
shtlrs, line 37: We can make use of the `StartupError` exception here and raise it instead of a bare exception.

shtlrs (commented):
Thanks for the changes, couple of small comments.

bot/log.py
etrotta: All paths in that function return None, just implicitly (one from a bare return statement, one from having no explicit return statement)

shtlrs: > All paths in that function return None, just implicitly 

Yes that i know, it's just that i thought ruff would either impose that all returns be implicit, and when you have at leat one explicit, the rest need to become as such as well

PR #3198, opened 2024-11-14
Enable 3.13's colorized output (bot half)

thurisatic:
One of the features new to Python 3.13 is the use of colored traceback output. These colors *greatly* aid in locating smaller bugs and are overall quite handy, but they are not always enabled, including for the bot's output message. This changes that.

Care was taken to achieve a significant degree of backwards compatibility. Long tracebacks, whose full forms are uploaded to pinnwand, go through a regular expression that removes ANSI escape codes. Pinnwand does not support ANSI, so if this is not done, the raw characters are included, easily making the traceback unintelligible. This will not delete escaped ANSI sequences, but it does provide a slight regression in that colored output that makes it to pinnwand is rendered monochrome (previously, the characters would be present, causing the legibility issue mentioned earlier - though whether this was a problem may be debated).

This PR is the bot side. There is a snekbox PR (python-discord/snekbox#225) that causes the colored output, which should probably be merged *after* this one.

This was not done alone. Credit is shared with
@ChrisLovering (for pointing to snekbox)
@shenanigansd and @parrrate (for help with docker)
@godlygeek (with the de-ANSIfication regex, much simpler than i thought it'd be)

a demo of this feature, with a code block by PR author and eval output featuring ANSI text and a colored traceback
paste mentioned in image is https://paste.pythondiscord.com/NEP7SBFPWJVETSCU5XJDLVG6Q4

### DIFF
--- a/bot/exts/utils/snekbox/_cog.py
+++ b/bot/exts/utils/snekbox/_cog.py
@@ -29,6 +29,7 @@ if TYPE_CHECKING:
 
 log = get_logger(__name__)
 
+ANSI_REGEX = re.compile(r"\N{ESC}\[[0-9;:]*m")
 ESCAPE_REGEX = re.compile("[`\u202E\u200B]{3,}")
 
 # The timeit command should only output the very last line, so all other output should be suppressed.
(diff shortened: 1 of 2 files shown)


shenanigansd (approved):
_oooh pretty_

ChrisLovering (approved):

PR #3226, opened 2025-01-03
Update SECURITY.md

Killer2B:


Killer2B:
bad

wookie184:
What is the purpose of this change? A single `#` seems correct as it is a top level header.

Issue #3172, opened 2024-09-13
Single newlines in tags get rendered as-is in Discord


n0Oo0Oo0b:
According to the markdown spec, lines separated by a single newline should be joined with a space.

Discord embeds doesn't seem to respect this, as demonstrated by the `!traceback` tag:

https://github.com/python-discord/bot/blob/13927ec4791a3f87193724ea3d1ac24143a8684c/bot/resources/tags/traceback.md?plain=1#L5-L8

![image](https://github.com/user-attachments/assets/af114303-716a-4e98-a43a-e3afdf446d26)

We could resolve this by formatting paragraphs on a single line, but I think joining the lines manually as the tag is loaded is better (something along the lines of `re.sub(r"(?<!\n)\n(?!\n)", " ", content)`) since it leaves open more formatting options.

PR #3232, opened 2025-01-08
Abbreviate codeblock instructions; remove ability to dismiss the inst…

swfarnsworth:
…ructions with emoji.

The more concise instructions are intended to be easier to read and increase the rate of followthru. That the instructions cannot be dismissed is intended to make them harder to ignore.

The spelling of "followthru" is very intentional.

image

image

image

I should probably remove the "It looks like you ..." from this one.

image

I'm not sure if this one should have the example. I should add the "Check this out to find the backtick key" from the `!code` embed.

### DIFF
--- a/bot/exts/info/codeblock/_instructions.py
+++ b/bot/exts/info/codeblock/_instructions.py
@@ -52,10 +52,10 @@ def _get_bad_ticks_message(code_block: _parsing.CodeBlock) -> str | None:
         log.trace("Language specifier issue found; appending additional instructions.")
 
         # The first line has double newlines which are not desirable when appending the msg.
-        addition_msg = addition_msg.replace("\n\n", " ", 1)
+        addition_msg = addition_msg.replace("\n\n", " ", 1).strip()
 
         # Make the first character of the addition lower case.
-        instructions += "\n\nFurthermore, " + addition_msg[0].lower() + addition_msg[1:]
+        instructions += "\n\nAlso, " + addition_msg[0].lower() + addition_msg[1:]
     else:
         log.trace("No issues with the language specifier found.")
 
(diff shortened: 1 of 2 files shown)


bot/exts/info/codeblock/_instructions.py
```diff
@@ -177,7 +164,4 @@ def get_instructions(content: str) -> str | None:
             if not instructions:
                 instructions = _get_no_lang_message(block.content)
 
-    if instructions:
-        instructions += "\nYou can **edit your original message** to correct your code block."
```
MarkKoz, line 181: Do you think this line was ineffective? The intent was to avoid spamming the channel with several attempts to fix their message. I don't know a better way to communicate this.

bot/exts/info/codeblock/_instructions.py
```diff
@@ -135,12 +126,8 @@ def _get_no_lang_message(content: str) -> str | None:
         example_blocks = _get_example("py")
 
         # Note that _get_bad_ticks_message expects the first line to have two newlines.
-        return (
-            "It looks like you pasted Python code without syntax highlighting.\n\n"
-            "Please use syntax highlighting to improve the legibility of your code and make "
-            "it easier for us to help you.\n\n"
-            f"**To do this, use the following method:**\n{example_blocks}"
-        )
+        return f"Please add a `py` after the three backticks.\n\n{example_blocks}"
```
MarkKoz, line 129: The embed title already says "please", so saying it again feels redundant to me.

MarkKoz (approved):
I like the additional ideas you left in the PR description. This is still a good incremental improvement on its own.

bot/exts/info/codeblock/_instructions.py
swfarnsworth: I moved this instruction to the title for the embed.

MarkKoz: Oh I need to read more carefully, sorry!

swfarnsworth: No problem; thank you for having written the original code in such a way that the changes in this PR were straightforward to make.

swfarnsworth:
image

This is just documenting an edge case for the parser (it seems to think the docstring is the code block).

swfarnsworth:
Updated cases:

image

image

swfarnsworth:
@mbaruh do you have any conscientious objections to the changes?

n0Oo0Oo0b:
Considering single newlines are rendered in embeds too (#3172), replacing the double newlines should help reduce the vertical height a bit. I also think the "this will result in the following" section is unnecessary when the user (incorrectly) attempts to use a codeblock since it's often triggered by users right after being told to use a codeblock at all.

On a related note, I think adding something along the lines of "you can copy-paste the example in this embed if you're having trouble with formatting" might help as well.

lemonsaurus (approved):
Big improvement. 🦄

PR #3241, opened 2025-01-19
Apply filtering to text attachments; offer to auto-upload text attachments to paste bin

swfarnsworth:
This is a work-in-progress intended to allow us to lift the embargo on text file attachments. Whether we use this functionality, or anything like it, is subject to further discussion. I'm submitting it as a draft PR to expose what I've done so far.

Allowing text file attachments is intended to remove friction in making a user's code available to question-answerers during help sessions. If a user uploads their py file, and it's deleted by the bot, the user might become confused and not follow the instructions to use the paste bin. Instructing users on how to use the paste bin can also be grating for question-answerers, especially when they would have been able to read the py file in the file embed that appears in the desktop client, which provides syntax highlighting. 

However, the file embeds do not appear in the mobile version of Discord. Our filtering system also does not consider the content of text-based file attachments. This PR implements a solution for the former and the beginnings of a solution for the latter.

## Offer to auto-upload text attachments to the paste bin
This functionality, which is already fully implemented, is intended to benefit mobile users who can't see the text file embeds, while upholding our promise not to copy user content to an external system (ie, our paste bin) without their permission. It consists of these steps, following a user sending a message containing text attachments:
- The bot replies to the message, asking the user to react with a checkmark to have the attachments automatically uploaded to the paste bin.
- If the user complies within a fixed duration, the bot does so, and replaces its own message with a link to the paste bin.
- The bot DMs the user the delete link for the paste.
- Alternatively, the bot also waits for the user to react with a trashcan emoji, deleting the paste. This option times out, but the delete link sent via DMs does not.

## Apply token filters to text attachments
This one is much trickier. The simplest solution I could think of (after trying a few more complicated ones) is to simply treat the content of text attachments as a continuation of the message itself, though this could make it not-immediately-obvious to moderators when a filter is triggered by an attachment rather than the true message content. We also don't want to waste compute time applying the filters to especially long text attachments (which is relatively likely for CSV and JSON attachments).

We might decide on a constant size limit for text-based attachments and modify the extension filters to consider both the file extension and the size of the file, while also not appending attachment content to the message if the file size exceeds that same constant. We might also not apply the token filter to CSVs and JSONs, since those are usually posted as part of a minimal reproducible example and are probably not intended to be read.

### DIFF
--- a/bot/exts/filtering/_filter_lists/extension.py
+++ b/bot/exts/filtering/_filter_lists/extension.py
@@ -4,7 +4,7 @@ import typing
 from os.path import splitext
 
 import bot
-from bot.constants import Channels, Emojis
+from bot.constants import Channels
 from bot.exts.filtering._filter_context import Event, FilterContext
 from bot.exts.filtering._filter_lists.filter_list import FilterList, ListType
 from bot.exts.filtering._filters.extension import ExtensionFilter
@@ -14,26 +14,11 @@ from bot.exts.filtering._settings import ActionSettings
 if typing.TYPE_CHECKING:
     from bot.exts.filtering.filtering import Filtering
 
-PASTE_URL = "https://paste.pythondiscord.com"
-PY_EMBED_DESCRIPTION = (
-    "It looks like you tried to attach a Python file - "
-    f"please use a code-pasting service such as {PASTE_URL}"
-)
-
-TXT_LIKE_FILES = {".txt", ".csv", ".json", ".py"}
-TXT_EMBED_DESCRIPTION = (
-    "You either uploaded a `{blocked_extension}` file or entered a message that was too long. "
-    f"Please use our [paste bin]({PASTE_URL}) instead."
-)
-
 DISALLOWED_EMBED_DESCRIPTION = (
     "It looks like you tried to attach file type(s) that we do not allow ({joined_blacklist}). "
     "We currently allow the following file types: **{joined_whitelist}**.\n\n"
     "Feel free to ask in {meta_channel_mention} if you think this is a mistake."
 )
-
-PASTEBIN_UPLOAD_EMOJI = Emojis.check_mark
-DELETE_PASTE_EMOJI = Emojis.trashcan
 
 
 class ExtensionsList(FilterList[ExtensionFilter]):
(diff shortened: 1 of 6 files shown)


bot/exts/filtering/_filter_lists/extension.py
```diff
@@ -20,7 +28,7 @@
     f"please use a code-pasting service such as {PASTE_URL}"
 )
 
-TXT_LIKE_FILES = {".txt", ".csv", ".json"}
+TXT_LIKE_FILES = {".txt", ".csv", ".json", ".py"}
```
lemonsaurus, line 31: If we're lifting the ban on `.py`, then anything adjacent feels like it might as well also be added. Maybe `.js`,  `.css`, `.html` so that anyone using Python to build a webapp of some sort won't run into a brick wall? What about stuff like `.yaml`, `.toml`?

bot/exts/filtering/_filter_lists/extension.py
```diff
@@ -116,3 +127,80 @@ async def actions_for(
...
         return actions, [f"`{ext}`" if ext else "`No Extension`" for ext in not_allowed], {ListType.ALLOW: triggered}
+
+
+class EmbedFileHandler(commands.Cog):
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @staticmethod
+    async def _convert_attachment(attachment: discord.Attachment) -> paste_service.PasteFile:
```
lemonsaurus, line 138: coroutines deserve ✨d o c s t r i n g s ✨

bot/exts/filtering/_filter_lists/extension.py
```diff
@@ -116,3 +127,80 @@ async def actions_for(
...
+
+        bot_reply = await message.reply(f"React with {PASTEBIN_UPLOAD_EMOJI} to upload your file to our paste bin")
+        await bot_reply.add_reaction(PASTEBIN_UPLOAD_EMOJI)
+
+        def wait_for_upload_permission(reaction: discord.Reaction, user: discord.User) -> bool:
+            return (
+                reaction.message.id == bot_reply.id
+                and str(reaction.emoji) == PASTEBIN_UPLOAD_EMOJI
+                and user == message.author
+            )
```
lemonsaurus, line 157: wait, why are we defining this function inside the coroutine? Is that necessary? It strikes me as seems needlessly nested / convoluted, I'd just make this a helper function and give it a leading underscore, make it part of the cog class, and use ?`self._wait_for_upload_permission` below as the callback.

bot/exts/filtering/_filter_lists/extension.py
```diff
@@ -116,3 +127,80 @@ async def actions_for(
...
+            await self._convert_attachment(f)
+            for f in message.attachments
+            if f.content_type.startswith("text")
+        ]
+
+        try:
+            async with aiohttp.ClientSession() as session:
+                paste_response = await paste_service.send_to_paste_service(files=files, http_session=session)
+        except (paste_service.PasteTooLongError, ValueError):
+            # paste is too long
```
lemonsaurus, line 179: this comment is a bit WET 🍋 

it's not a surprise that `PasteTooLongError` means the paste is too long 😁

lemonsaurus (approved):
Tentatively approved. The feature looks solid, but I have stylistic concerns.
- Classes, methods and coroutines should have docstrings
- The cognitive complexity (e.g. mccabe) of  `on_message` is too high. It's too deeply nested, too long, and doesn't break the code into logical blocks with helpful block comments headlining them as a longer piece of code ought to. I'd like to see a refactor here to move helper functions out into the class, and perhaps break up the different pieces into a few function calls with helpful names. I think generally that it's better that `on_message` itself is aggressively short and very easy to read and understand, while the implementation details are abstracted away into other functions.
- Somewhat inconsistent adherence to PEP257. Comments should start with capital letters and end with punctuation, and should be imperative mood.

Looks like a great addition to the bot, though <3 thank you for writing it! I can't wait to see it merged.

bot/exts/filtering/_filter_lists/extension.py
swfarnsworth: I'll have to see if that's possible, since it's reading `bot_reply` and `message` from the outer scope, and I think the signature of `wait_for_upload_permission` is mandated by `self.bot.wait_for("reaction_add"`.

bot/exts/filtering/_filter_lists/extension.py
swfarnsworth: I realized later that that code has no effect on which file types are allowed.

bot/exts/filtering/_filter_lists/extension.py
swfarnsworth: I think that comment is an artefact of when I found out that a long paste can also raise ValueError.

swfarnsworth:
@lemonsaurus thanks for your feedback. I thought I was in the finishing-touches stage for this, but there's actually still more to be done. I'm pushing what I've worked on this evening for additional scrutiny/ridicule.

bot/exts/filtering/_filter_lists/extension.py
lemonsaurus: hm, okay I see. yeah that's a good reason, then :)

PR #3255, opened 2025-02-07
Send the log of purged messages to #message-change-log too

### LINKED ISSUE #3055: Post log from `!clean` (and simialr commands) in #message-change-log
Currently, if we do `!cban` or similar clean commands, the record of the messages being cleared gets output to `#mods`.

This can cause some confusion if two moderators are investigating and messages suddenly disappear, but there isn't a record in `#message-change-log`. If messages get deleted or edited, the logical place for a record of that to appear would be in `#message-change-log`. 

So we should change where the record of deleted messages get sent to and send it to `#message-change-log` instead.

vivekashok1221:
Resolves #3055.

I am sending the same message deletion logs currently sent to `#mod-log `to `#message-change-log` as well.

image

### DIFF
--- a/bot/exts/moderation/clean.py
+++ b/bot/exts/moderation/clean.py
@@ -368,14 +368,15 @@ class Clean(Cog):
             f"A log of the deleted messages can be found [here]({log_url})."
         )
 
-        await send_log_message(
-            self.bot,
-            icon_url=Icons.message_bulk_delete,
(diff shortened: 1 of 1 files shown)


swfarnsworth (approved):
Works on my test server.

KeithTheEE (approved):
Looks good, nothing much changes and is already tested so I think it's ready to go

PR #3259, opened 2025-02-13
Permit !remindme with no specified content

bast0006:
Change !remindme so that it doesn't fail if it hasn't been provided either text or a reply to copy from. I'm pretty sure I'm not the only person who gets annoyed when the bot wants a placeholder, when really I just want a reference to the current conversation. 

There are no tests for this part, but I don't have time to write them at the moment, hopefully that's okay.

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -252,9 +252,9 @@ class Reminders(Cog):
         await self.bot.api_client.delete(f"bot/reminders/{reminder['id']}")
 
     @staticmethod
-    async def try_get_content_from_reply(ctx: Context) -> str | None:
+    async def try_get_content_from_reply(ctx: Context) -> str:
         """
-        Attempts to get content from the referenced message, if applicable.
+        Attempts to get content from the referenced message, if applicable, or provides a default.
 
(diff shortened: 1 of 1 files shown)


bot/exts/utils/reminders.py
```diff
@@ -263,13 +263,9 @@ async def try_get_content_from_reply(ctx: Context) -> str | None:
...
-        if content is None:
-            await send_denial(ctx, "Your reminder must have a content and/or reply to a message.")
-            return None
-
-        # If the replied message has no content (e.g. only attachments/embeds)
-        if content == "":
+        # If the replied message has no content, we couldn't get the content, or no content was provided
+        # (e.g. only attachments/embeds)
+        if content is None or content == "":
             content = "*See referenced message.*"
```
vivekashok1221, line 269: The return type of this static method can be changed from `str | None` to `str`.

vivekashok1221 (changes requested):
Looks good, I've left a minor comment,

bot/exts/utils/reminders.py
bast0006: Oop, whoops, knew I was missing something

vivekashok1221 (approved):
LGTM

ChrisLovering (approved):

PR #3260, opened 2025-02-14
fix: `has_lines` returns empty str instead of bool

Strengthless:
Minimal reproducible example:
```py
def has_lines(string: str, count: int) -> bool:
    """Return True if `string` has at least `count` lines."""
    # Benchmarks show this is significantly faster than using str.count("\n") or a for loop & break.
    split = string.split("\n", count - 1)

    # Make sure the last part isn't empty, which would happen if there was a final newline.
    return split[-1] and len(split) == count

print(has_lines("test\n", 1)) # returns True correctly
print(has_lines("test\n", 3)) # returns "" instead of False
```

Of course, `""` in an if-statement will still evaluate to `False`, but it's not a good practice and probably unsafe.

This PR fixes that by adding an explicit `!= ""` check, so that booleans are returned at all times.

Unit tests for the whole file `bot/utils/helpers.py` have also been included, and the test coverage for that file is now at 100%.

### DIFF
--- a/bot/utils/helpers.py
+++ b/bot/utils/helpers.py
@@ -25,7 +25,7 @@ def has_lines(string: str, count: int) -> bool:
     split = string.split("\n", count - 1)
 
     # Make sure the last part isn't empty, which would happen if there was a final newline.
-    return split[-1] and len(split) == count
+    return split[-1] != "" and len(split) == count
 
 
 def pad_base64(data: str) -> str:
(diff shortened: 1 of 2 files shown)


vivekashok1221 (approved):
Nice catch, I'm curious how you found this.

ChrisLovering (approved):

PR #3262, opened 2025-02-16
Bump deps to latest

ChrisLovering:


KeithTheEE (approved):
I don't have an environment to test this but the updates look functional for the dependency upgrades
The added note about `pip-licenses` not being `PEP-639` compliant helped me follow that addition. 

Looks good from all I can tell.

PR #3263, opened 2025-02-16
Doc: add initial report files (#1)

Dudjfy:
Added report template file, with the name and URL of the project.

Added the assignment description file.


Dudjfy:
Did an oopsie, sorry, igonre or delete this 😅

PR #3295, opened 2025-03-14
Compact doc feature

anand2312:
As per @decorator-factory 's suggestion:
> `!d` provides a pretty huge embed that takes a lot of space, and you need to post a message for each documentation entry. I almost always click the docs link anyway, because the text gets cut off for all but the most trivial items.
> It's common for gaming communities on Reddit to have a bot that scans messages for specially formatted mentions of items and gives some very brief information on all of them (and most importantly a documentation link) in a single message: see screenshot. I think it would be pretty cool to be able to type something like
>
> > You'll need to use `[[zip]]` or `[[itertools.zip_longest]]`
>
> and have the bot reply with a _compact_ message linking to the documentation, like: 
>
> > - [`zip(*iterables, strict=False)`](<https://docs.python.org/3/library/functions.html#zip>)
> > - [`itertools.zip_longest(*iterables, fillvalue=None)`](<https://docs.python.org/3/library/itertools.html#itertools.zip_longest>)
> 

- [Link to discussion thread](https://discord.com/channels/267624335836053506/1326336848636936265)

This PR implements the feature described, and also reacts to message edits and edits the bot's response.

### DIFF
--- a/bot/exts/info/doc/_cog.py
+++ b/bot/exts/info/doc/_cog.py
@@ -268,26 +268,22 @@ class DocCog(commands.Cog):
                 return "Unable to parse the requested symbol."
         return markdown
 
-    async def _get_symbols_items(self, symbols: list[str]) -> list[tuple[str, DocItem | None]]:
-        """
-        Get DocItems for the given list of symbols, and update the stats for the fetched doc items.
-
-        Returns (symbol, None) for any symbol for which a DocItem could not be found.
-        """
+    async def _get_symbols_items(self, symbols: list[str]) -> dict[str, DocItem | None]:
+        """Get DocItems for the given list of symbols, and update the stats for the fetched doc items."""
         if not self.refresh_event.is_set():
             log.debug("Waiting for inventories to be refreshed before processing item.")
             await self.refresh_event.wait()
 
         # Ensure a refresh can't run in case of a context switch until the with block is exited
         with self.symbol_get_event:
-            items: list[tuple[str, DocItem | None]] = []
-            for symbol_name in set(symbols):
+            items: dict[str, DocItem | None] = {}
+            for symbol_name in symbols:
                 symbol_name, doc_item = self.get_symbol_item(symbol_name)
                 if doc_item:
-                    items.append((symbol_name, doc_item))
+                    items[symbol_name] = doc_item
                     self.bot.stats.incr(f"doc_fetches.{doc_item.package}")
                 else:
-                    items.append((symbol_name, None))
+                    items[symbol_name] = None
 
         return items
 
@@ -301,7 +297,7 @@ class DocCog(commands.Cog):
         """
         log.trace(f"Building embed for symbol `{symbol_name}`")
         _items = await self._get_symbols_items([symbol_name])
-        symbol_name, doc_item = _items[0]
+        doc_item = _items[symbol_name]
 
         if doc_item is None:
             log.debug(f"{symbol_name=} does not exist.")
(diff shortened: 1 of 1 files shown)


anand2312:
Sorry for the formatting changes to the unrelated lines. I think they're because I ran `ruff format` on the file; I can revert those changes if needed

decorator-factory:
> Sorry for the formatting changes to the unrelated lines. I think they're because I ran `ruff format` on the file; I can revert those changes if needed

Yeah, I think it's better not to change unrelated lines if we're not checking exactly matching `ruff format` in the CI

bot/exts/info/doc/_cog.py
```diff
@@ -273,33 +296,107 @@ async def create_symbol_embed(self, symbol_name: str) -> discord.Embed | None:
...
+        """
+        Re-eval a message for symbols if it gets edited.
+
+        The edit logic is essentially the same as that of the code eval (snekbox) command.
+        """
+        def edit_check(before: discord.Message, after: discord.Message) -> bool:
+            return (
+                original.id == before.id
+                and before.content != after.content
+                and len(re.findall(COMPACT_DOC_REGEX, after.content.replace("`", ""))) > 0
```
decorator-factory, line 353:
```py
and re.search(COMPACT_DOC_REGEX, after.content.replace("`", "")) is not None
```

bot/exts/info/doc/_cog.py
```diff
@@ -42,6 +44,10 @@
 
 COMMAND_LOCK_SINGLETON = "inventory refresh"
 
+REDO_EMOJI = "\U0001f501"  # :repeat:
+REDO_TIMEOUT = 30
+COMPACT_DOC_REGEX = r"\[\[(?P<symbol>\S*)\]\]"
```
decorator-factory, line 49: The reason I suggested matching specifically 
```py
[[`item`]]
```
is that this won't match Python lists like `[[]]` or `[[initial]]` or `[[1,2,3]]` and lead to confusing/noisy messages from the bot. Maybe this could be changed to
```py
r"\[\[`(?P<symbol>[^\s\]`]*)`\]\]"
```
? This would also remove the need to strip \` from matches

bot/exts/info/doc/_cog.py
```diff
@@ -264,6 +268,25 @@ async def get_symbol_markdown(self, doc_item: DocItem) -> str:
                 return "Unable to parse the requested symbol."
         return markdown
 
+    async def _get_symbols_items(self, symbols: list[str]) -> dict[str, DocItem | None]:
+        """Get DocItems for the given list of symbols, and update the stats for the fetched doc items."""
+        if not self.refresh_event.is_set():
+            log.debug("Waiting for inventories to be refreshed before processing item.")
+            await self.refresh_event.wait()
+
+        # Ensure a refresh can't run in case of a context switch until the with block is exited
```
decorator-factory, line 278: But a context switch cannot happen if there's no `await`/`async for`/`async with`

decorator-factory (changes requested):

bot/exts/info/doc/_cog.py
```diff
@@ -439,8 +537,7 @@ async def refresh_command(self, ctx: commands.Context) -> None:
             removed = "- " + removed
 
         embed = discord.Embed(
-            title="Inventories refreshed",
-            description=f"```diff\n{added}\n{removed}```" if added or removed else ""
+            title="Inventories refreshed", description=f"```diff\n{added}\n{removed}```" if added or removed else ""
```
vivekashok1221, line 540: You forgot to revert this formatting change.

vivekashok1221 (changes requested):

Issue #3289, opened 2025-03-07
!zen command does not allow slicing.


PeckyChicken:
The `!zen {line}` command treats the line parameter (if it is a number) as a list index, i.e. it starts from 0 and allows negative indices to start from the end of the list.
However, it does not allow list slicing. I think it should allow a slice parameter like `2:4` and it would return 
```
Simple is better than complex.
Complex is better than complicated.
```
being lines 2 and 3 of the Zen of Python.
It wouldn't have too much use but I still think it is something worth implementing.

PR #3297, opened 2025-03-16
Zen slicing

LeandroVandari:
Implemented the feature in issue #3289.

Now, the `!zen` command accepts arguments in the form `start:stop`, and will return the corresponding zen lines.
Additionally, since the indices can now no longer simply be passed as integers, the `zen_rule_index` argument to the zen command was removed, and now indices are parsed from a regex expression.

### DIFF
--- a/bot/exts/utils/utils.py
+++ b/bot/exts/utils/utils.py
@@ -111,7 +111,7 @@ class Utils(Cog):
         zen_lines = ZEN_OF_PYTHON.splitlines()
 
         # Prioritize checking for an index or slice
-        match = re.search(r"^(\d+)(:\d+)?", search_value.split(" ")[0])
+        match = re.match(r"(-?\d+)(:-?\d+)?", search_value.split(" ")[0])
         if match:
             upper_bound = len(zen_lines) - 1
             lower_bound = -1 * len(zen_lines)
--- a/bot/exts/utils/utils.py
+++ b/bot/exts/utils/utils.py
@@ -88,16 +88,14 @@ class Utils(Cog):
     async def zen(
         self,
         ctx: Context,
-        zen_rule_index: int | None,
-        *,
-        search_value: str | None = None
+        search_value: str | None,
     ) -> None:
         """
         Show the Zen of Python.
 
         Without any arguments, the full Zen will be produced.
-        If zen_rule_index is provided, the line with that index will be produced.
-        If only a string is provided, the line which matches best will be produced.
+        If an index or a slice is provided, the corresponding lines will be produced.
+        Otherwise, the line which matches best will be produced.
         """
         embed = Embed(
             colour=Colour.og_blurple(),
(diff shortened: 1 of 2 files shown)


bot/exts/utils/utils.py
```diff
@@ -88,40 +88,53 @@ def get_info(char: str) -> tuple[str, str]:
...
             return
 
         zen_lines = ZEN_OF_PYTHON.splitlines()
 
-        # Prioritize passing the zen rule index
-        if zen_rule_index is not None:
-
+        # Prioritize checking for an index or slice
+        match = re.search(r"^(\d+)(:\d+)?", search_value.split(" ")[0])
+        if match:
```
vivekashok1221, line 115: To regex needs to be adjusted to account for negative indices:
```suggestion
        match = re.search(r"^(-?\d+)(:-?\d+)?", search_value.split(" ")[0])
        if match:
```
(Rest of the logic already handles negative indexing fine)

vivekashok1221: With the current regex, the match fails and `search_value` is assumed to be a string query rather than as an index or a slice.

vivekashok1221 (changes requested):
Thank you for opening this PR :snake:

bot/exts/utils/utils.py
decorator-factory: I'll add a small nitpick: `re.search(r"^PATTERN")` can be written as `re.match(r"PATTERN")`

LeandroVandari: right, I thought about negative indices everywhere but in the regex 😅. I'll add these changes later on today

decorator-factory: Maybe it's possible to write a test for this? It seems like a mostly-pure function

LeandroVandari:
This latest commit fixes the problems you mentioned, but i noticed it's impossible to include the last line when slicing (i.e. `!zen 1:18` and `!zen 1:-1` will only include up to the semi-last line, and `!zen 1:19` will error out. I was considering some options to fix that, and I wanted to get your opinions:

1. Make slicing end inclusive  -- i don't like this because it feels unpythonic and just uncoherent with the language
2. Allowing for `end_index == len(zen_lines)` -- again, weird and unpythonic, but less bad than the previous option.
3. Add the possibility for an open range (i.e. 1:) which will include up to the last line. -- this feels like the best option to me.

Additionally, I was wondering: Do we ever want to support `step`s, now that we're going the slice way? That feature would probably never be used, but it's just consistent with slicing I feel like.

vivekashok1221:
> 2. Allowing for end_index == len(zen_lines) -- again, weird and unpythonic, but less bad than the previous option.
> 3. Add the possibility for an open range (i.e. 1:) which will include up to the last line. -- this feels like the best option to me.

 I like option 2 since it's consistent with how slicing in Python works. If we deviate from it, people might think it's a bug.
 Maybe we can do option 2 AND option 3.


re: writing tests
> draft a PR for it later on (should I open an issue about it?)

Sure (you don't need to open an issue IMO but you can if you want to)
Writing tests is optional btw, but highly appreciated.

bot/exts/utils/utils.py
```diff
@@ -88,40 +88,53 @@ def get_info(char: str) -> tuple[str, str]:
...
+                return
+
+            end_index=  int(match.group(3)) if match.group(3) else len(zen_lines)
+
+            if not ((lower_bound <= start_index <= upper_bound) and (lower_bound <= end_index <= len(zen_lines))):
+                raise BadArgument(f"Please provide valid indices between {lower_bound} and {upper_bound}.")
+            if not (start_index % len(zen_lines) < end_index % (len(zen_lines) + 1)):
+                raise BadArgument("The start index for the slice must be smaller than the end index.")
+
+            embed.title += f" (lines {start_index%len(zen_lines)}-{end_index%len(zen_lines)}):"
```
vivekashok1221, line 136:
```suggestion
            embed.title += f" (lines {start_index%len(zen_lines)}-{(end_index-1)%len(zen_lines)}):"
```
This is to prevent embed titles like `The Zen of Python (lines 0-0):` when the end index is 19. Also makes more sense to display it this way since end index is exclusive.

vivekashok1221 (changes requested):
Noticed one more thing. Everything else looks good to me.

vivekashok1221 (approved):
LGTM

LeandroVandari:
I drafted some tests  for the zen command. I tried to keep a balance between hard-coding / developing answers myself. I don't think it's ready yet, but I'd love some opinions on it :)

vivekashok1221 (approved):
LGTM

ChrisLovering (approved):

PR #3311, opened 2025-04-11
Add step to zen command slicing

MeGaGiGaGon:
This adds the long awaited ability to step while slicing to the `!zen` command, so you can finally do stupid things like getting every other line in reverse.

The regex was completely reworked to allow for the steps (I finally got to use `++` for something) (groups are now named instead of indexed for clarity)

Line numbers for display are now found via `zen_lines.index` instead of math. I am aware this is worse for performance, but it is a massive win for both ease of implementation and code clarity, since with negative steps the clamping and offset math for line numbers gets very hard very quick. The main downside of this approach is the math can offer very fine-grained error messaging, whereas the current code gives a generic `"Slice returned 0 lines."` message.

Due to how the change is implemented, more slicings are now valid than before, ie `!zen :100` will now work, so those test case fails were removed.

I am aware this will probably fail the formatting and linting CI passes, but I could not run them locally as my computer's OS is on it's deathbed and poetry refuses to work correctly. 

Edit:
All things now fixed. The tests for full zen (`0:19`, `:`, `::`) were failing because the source/default description included an extra `\n`, so I added a `.strip()` to fix that.
One other thing I forgot to say above, with the easier implementation from indexing I was able to clean up the messages that bugged me, so now if only one line is returned after slicing it will always say `(line x)` instead of `(lines x-x)`, and if the slice returns the full thing, it will be `, by Tim Peters` instead of `(lines 0-18)`


jb3 (approved):

vivekashok1221 (approved):
LGTM

PR #3331, opened 2025-06-02
Implemented optional duration parameter in slowmode command

### LINKED ISSUE #3327: Slowmode timeout argument
Supersedes #3325.

This issue is for the addition of a new argument to the `!slowmode set` command to have the bot time out slowmodes.

Proposed syntax: `!slowmode set [channel] <delay> <duration>`

Example use: `!slowmode set #python-discussion 10s 30m`

After expiration the slowmode should probably restore to whatever the previous value is, that way Moderators can still set a permanent value and have that be preserved (i.e. temporarily have a 10s slowmode before reverting to 2s after 2 hours).

We should persist this in Redis as it is non-critical if this information gets lost but will continue across reboots. This can likely be integrated with the existing scheduler.

b0nes1:
Closes #3327.

Implemented an optional `duration` argument for the slowmode command. If the duration argument is present, the intended expiration datetime and the original slowmode interval are saved in redis cache. After the duration is elapsed, the original slowmode interval is grabbed from redis and restored to the channel.

The expiration datetime is cached so that it can persist bot restart.


tests/bot/exts/moderation/test_slowmode.py
```diff
@@ -95,6 +96,113 @@ async def test_reset_slowmode_sets_delay_to_zero(self) -> None:
...
+
+        # check that _reschedule function was called upon cog reload.
+        self.cog._reschedule.assert_called()
+
+        # check that a task was created for every cached slowmode.
+        for channel in channels:
+            self.assertIn(channel.id, self.cog.scheduler)
+
+        # check that one channel with slowmode expiration in the past was edited immediately.
+        channels[0].edit.assert_awaited_once_with(slowmode_delay=channels[0].slowmode_delay)
```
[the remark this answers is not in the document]
wookie184: You are creating `MockTextChannel`s, but `_revert_slowmode` will create it's own mocks from calling `self.bot.get_channel(id)`. Those are the mocks that will have `edit` called on them, not your ones.

Not sure what the easiest fix would be.

bot/exts/moderation/slowmode.py
```diff
@@ -26,8 +30,15 @@
 class Slowmode(Cog):
     """Commands for getting and setting slowmode delays of text channels."""
 
+    # Stores the expiration timestamp in POSIX format for active slowmodes, keyed by channel ID.
+    slowmode_expiration_cache = RedisCache()
+
+    # Stores the original slowmode interval by channel ID, allowing its restoration after temporary slowmode expires.
+    original_slowmode_cache = RedisCache()
```
jb3, line 37: Given these are both structured and known values, I would maybe consider serializing them together into one store?

This is a nit though and others may disagree, it just appears that we use them more or less in the same place so whilst if we were just working in dict-land this would be preferred when working in a place with an underlying store I prefer it being the same.

jb3 (changes requested):
One minor comment, though if others disagree then let me know, it is more a nit based on me not wanting to add dependence onto intricacies of Redis and hoping multiple keys exist referring to the same slowmode delay etc.

bot/exts/moderation/slowmode.py
```diff
@@ -66,37 +89,98 @@ async def set_slowmode(
...
 
         # Ensure the delay is within discord's limits
-        if slowmode_delay <= SLOWMODE_MAX_DELAY:
-            log.info(f"{ctx.author} set the slowmode delay for #{channel} to {humanized_delay}.")
-
-            await channel.edit(slowmode_delay=slowmode_delay)
-            if channel.id in COMMONLY_SLOWMODED_CHANNELS:
-                log.info(f"Recording slowmode change in stats for {channel.name}.")
-                self.bot.stats.gauge(f"slowmode.{COMMONLY_SLOWMODED_CHANNELS[channel.id]}", slowmode_delay)
+        if not slowmode_delay <= SLOWMODE_MAX_DELAY:
```
mbaruh, line 92:
```suggestion
        if slowmode_delay > SLOWMODE_MAX_DELAY:
```

bot/exts/moderation/slowmode.py
```diff
@@ -42,17 +53,29 @@ async def get_slowmode(self, ctx: Context, channel: MessageHolder) -> None:
...
+        else:
+            await ctx.send(f"The slowmode delay for {channel.mention} is {humanized_delay}.")
 
     @slowmode_group.command(name="set", aliases=["s"])
     async def set_slowmode(
         self,
         ctx: Context,
         channel: MessageHolder,
         delay: DurationDelta | Literal["0s", "0seconds"],
+        duration: DurationDelta | None = None
```
mbaruh, line 71: Use `Duration` instead of `DurationDelta`, which handles the conversion to datetime for you.

bot/exts/moderation/slowmode.py
```diff
@@ -66,37 +89,98 @@ async def set_slowmode(
...
+                f"{Emojis.cross_mark} The slowmode delay must be between 0 and 6 hours."
             )
+            return
 
-        else:
+        if duration is not None:
+            slowmode_duration = time.relativedelta_to_timedelta(duration).total_seconds()
+            humanized_duration = time.humanize_delta(duration)
+
+            expiration_time = datetime.now(tz=UTC) + timedelta(seconds=slowmode_duration)
```
mbaruh, line 107: This section is then simplified with the usage of `Duration`.

bot/exts/moderation/slowmode.py
```diff
@@ -66,37 +89,98 @@ async def set_slowmode(
...
+            expiration_timestamp = discord_timestamp(expiration_time, TimestampFormats.RELATIVE)
+
+            # Only update original_slowmode_cache if the last slowmode was not temporary.
+            if not await self.slowmode_expiration_cache.contains(channel.id):
+                await self.original_slowmode_cache.set(channel.id, channel.slowmode_delay)
+            await self.slowmode_expiration_cache.set(channel.id, expiration_time.timestamp())
+
+            self.scheduler.schedule_at(expiration_time, channel.id, self._revert_slowmode(channel.id))
```
mbaruh, line 115: The task ID has to be unique. If you're scheduling with a channel ID that already has a schedule, this creates a clash and the new coroutine won't be registered.

https://github.com/python-discord/bot-core/blob/16835478af6b842d776d634e53aab8ee31bd350c/pydis_core/utils/scheduling.py#L65-L66

You're already checking whether the current slowmode is scheduled to expire just above, can cancel the current task while you're at it (although this logic will probably change if you're changing to a single cache).

bot/exts/moderation/slowmode.py
```diff
@@ -66,37 +89,98 @@ async def set_slowmode(
...
             log.info(
-                f"{ctx.author} tried to set the slowmode delay of #{channel} to {humanized_delay}, "
-                "which is not between 0 and 6 hours."
+                f"{ctx.author} set the slowmode delay for #{channel} to"
+                f"{humanized_delay} which expires in {humanized_duration}."
```
mbaruh, line 118:
```suggestion
                f"{humanized_delay} which expires {humanized_duration}."
```
Same here, can specify what it's going to revert to.

bot/exts/moderation/slowmode.py
```diff
@@ -66,37 +89,98 @@ async def set_slowmode(
...
             )
+            await channel.edit(slowmode_delay=slowmode_delay)
+            await ctx.send(
+            f"{Emojis.check_mark} The slowmode delay for {channel.mention}"
+            f" is now {humanized_delay} and expires in {expiration_timestamp}."
```
mbaruh, line 123: And here

bot/exts/moderation/slowmode.py
```diff
@@ -66,37 +89,98 @@ async def set_slowmode(
...
+
+    async def _revert_slowmode(self, channel_id: int) -> None:
+        original_slowmode = await self.original_slowmode_cache.get(channel_id)
+        slowmode_delay = time.humanize_delta(seconds=original_slowmode)
+        channel = self.bot.get_channel(channel_id)
+        log.info(f"Slowmode in #{channel} ({channel.id}) has expired and has reverted to {slowmode_delay}.")
+        await channel.edit(slowmode_delay=original_slowmode)
+        await channel.send(
+            f"{Emojis.check_mark} A previously applied slowmode has expired and has been reverted to {slowmode_delay}."
+        )
```
mbaruh, line 156: Slowmode changes are usually silent. Unless the mods agreed that this is what they want I would send this message to #mods

bot/exts/moderation/slowmode.py
```diff
@@ -66,37 +89,98 @@ async def set_slowmode(
...
+        for channel_id, expiration in await self.slowmode_expiration_cache.items():
+            expiration_datetime = datetime.fromtimestamp(expiration, tz=UTC)
+            channel = self.bot.get_channel(channel_id)
+            log.info(f"Rescheduling slowmode expiration for #{channel} ({channel_id}).")
+            self.scheduler.schedule_at(expiration_datetime, channel_id, self._revert_slowmode(channel_id))
+
+    async def _revert_slowmode(self, channel_id: int) -> None:
+        original_slowmode = await self.original_slowmode_cache.get(channel_id)
+        slowmode_delay = time.humanize_delta(seconds=original_slowmode)
+        channel = self.bot.get_channel(channel_id)
```
mbaruh, line 151: I would use `get_or_fetch_channel` here

https://bot-core.pythondiscord.com/main/output/pydis_core.utils.channel.html#pydis_core.utils.channel.get_or_fetch_channel

bot/exts/moderation/slowmode.py
```diff
@@ -66,37 +89,98 @@ async def set_slowmode(
...
     @slowmode_group.command(name="reset", aliases=["r"])
     async def reset_slowmode(self, ctx: Context, channel: MessageHolder) -> None:
         """Reset the slowmode delay for a text channel to 0 seconds."""
         await self.set_slowmode(ctx, channel, relativedelta(seconds=0))
+        if channel is None:
+            channel = ctx.channel
+        if await self.slowmode_expiration_cache.contains(channel.id):
+            await self.slowmode_expiration_cache.delete(channel.id)
+            await self.original_slowmode_cache.delete(channel.id)
+            self.scheduler.cancel(channel.id)
```
mbaruh, line 169: Shouldn't it all be already handled in `set_slowmode`?

mbaruh (changes requested):

bot/exts/moderation/slowmode.py
b0nes1: And if i send it to mod-log, should `get_or_fetch_channel` be used here also? Should the mod-log channel be defined as an instance attribute inside __init__ or is it fine to define it inside of the command provided that `get_or_fetch_channel` is used?

jb3 (approved):

jb3 (approved):

b0nes1:
> Sorry it took me so long to respond, solid PR overall. I made a small improvement to make the code a bit cleaner (which fixed a small bug as a side effect)

Looks much nicer, thanks! Just out of curiosity, what was the bug?

jb3 (approved):
LGTM :shipit:

PR #3339, opened 2025-06-23
Delete automated infraction messages after a period of time

jb3:
This PR automatically deletes messages generated by the bot as part of dealing out automated infractions (e.g. spam filters). These have a habit of filling up channels which are frequently abused but have low traffic otherwise (some channels appear as just a wall of automated infractions).

As of now, the timer is set to 8 hours, though we can increase this if we feel we need the longer persistence (though obviously the full record and log is still preserved in the #mod-alerts channel).

Deletion times are persisted to Redis to ensure we try to delete stuff after the bot restarts.

Closes #2317.

PR #3340, opened 2025-06-24
Moderator changes minor refactor

jb3:
Refactors some items from #3339 and #3338.

- Use partial objects to avoid unnecessary API calls when tidying up infraction messages
  - I hadn't noticed the partial functions when I went on my hunt, I was working under the assumption I'd have to hook into the HTTP stuff or create my own partials, but this is much cleaner and cheaper.
- Don't tidy-up infractions outside mod channels (ensures that infractions issued by the bot automatically in mod-alerts stay as part of the context)
- Exclude the bot-commands channel from the `!rule` usage alerts.


wookie184 (approved):
Looks good

PR #3345, opened 2025-07-15
Poetry to UV migration

jb3:
- Migrate from using poetry to uv
- Update CI/CD to use uv over poetry

Once more projects have migrated we will need to update Contributing guidance on site to reflect this.


ChrisLovering (approved):
https://www.youtube.com/watch?v=i7aQig-wjYA

MarkKoz (approved):

Issue #3376, opened 2025-08-30
Preserve invocation context when redirecting output of command


jb3:
For some commands such as `!eval` if executed in a non-snekbox category we redirect the output of the command and delete the invocation message.

This can be annoying for users when the original invocation cannot be recovered and to respond in #bot-commands with a new invocation is effortful.

Instead, we should amend the decorator that redirects the output of commands to #bot-commands to also upload the full invocation message to the paste service (using the util from pydis_core).

https://github.com/python-discord/bot/blob/cb4401b45978d90bb53cdbe4c03d7b89747562b2/bot/decorators.py#L161-L174

Originally touched upon by @algmyr in #3360:

> This kinda also applies to the code running commands where if you accidentally run in the wrong channel your invocation is nuked and the code you typed is lost, but that might need to be handled differently by posting the code in the channel where the execution results goes.

DragonSenseiGuy:
I can take a look @jb3 at this just gimme a day or two.

PR #3377, opened 2025-08-31
Preserve Invocation Context

DragonSenseiGuy:
Fix for #3376 
Added:
- Whenever a user sends a !eval command in a non-snekbox channel the redirect will include the original invocation command.

### DIFF
--- a/bot/exts/utils/snekbox/_cog.py
+++ b/bot/exts/utils/snekbox/_cog.py
@@ -450,7 +450,18 @@ class Snekbox(Cog):
                 )
             else:
                 # The command was redirected so a reply wont work, send a normal message with a mention.
-                msg = f"{ctx.author.mention} {msg}"
+                try:
+                    paste_response = await send_to_paste_service(
+                        files=[PasteFile(content=ctx.message.content, lexer="markdown")],
+                        http_session=self.bot.http_session,
+                        paste_url=BaseURLs.paste_url,
+                    )
+                    paste_link = paste_response.link
+                    msg = f"Here's the output of your command\n{paste_link}:\n{msg}"
+                except paste_service.PasteUploadError:
+                    # Fallback if paste service fails
+                    msg = f"{ctx.author.mention} {msg}"
+                    log.warning("Pastebin Upload Error")
                 response = await ctx.send(msg, allowed_mentions=allowed_mentions, view=view, files=files)
             view.message = response
 


jb3 (approved):

jb3:
Note to whoever merges, this one can be squashed since it's a one file and one area change.

bot/decorators.py
```diff
@@ -154,8 +156,34 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
             log.trace(f"Redirecting output of {ctx.author}'s command '{ctx.command.name}' to {redirect_channel.name}")
             ctx.channel = redirect_channel
 
+            paste_link = None
+            async with ClientSession() as session:
+                try:
+                    paste_response = await send_to_paste_service(
+                            files=[PasteFile(content=ctx.message.content, lexer="markdown")],
+                            http_session=session,
```
swfarnsworth, line 164: It looks like these lines are indented an extra level.

bot/decorators.py
```diff
@@ -154,8 +156,34 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
+                    )
+                    paste_link = paste_response.link
+                except PasteUploadError:
+                    log.exception(
+                        "Failed to upload message %d in channel %d to paste service when redirecting output",
+                        ctx.message.channel.id, ctx.message.id
+                        )
```
swfarnsworth, line 171: This one line should be indented one level less.

bot/decorators.py
```diff
@@ -154,8 +156,34 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
             if ping_user:
-                await ctx.send(f"Here's the output of your command, {ctx.author.mention}")
+                msg += f"{ctx.author.mention}, "
+
+            msg += "Here's the output of "
+
+            if paste_link:
+                msg += f"[your command]({paste_link}):"
+            else:
+                msg += "your command:"
```
swfarnsworth, line 183: This could be one line using a ternary condition.

swfarnsworth:
@jb3, am I correct in my understanding that this copies messages from the server to our pastebin, possibly unexpectedly?

bot/decorators.py
```diff
@@ -154,8 +156,34 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
+
+            await ctx.send(msg)
```
MarkKoz, line 185: Would this potentially send an empty message?

bot/decorators.py
```diff
@@ -154,8 +156,34 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
+                    log.exception(
+                        "Failed to upload message %d in channel %d to paste service when redirecting output",
+                        ctx.message.channel.id, ctx.message.id
+                        )
+
+            msg = ""
+
             if ping_user:
-                await ctx.send(f"Here's the output of your command, {ctx.author.mention}")
+                msg += f"{ctx.author.mention}, "
```
MarkKoz, line 176: Why are we changing the order and putting the mention first?

jb3:
> @jb3, am I correct in my understanding that this copies messages from the server to our pastebin, possibly unexpectedly?

Yes, but with it being a bot command I'm less worried about that. Previously we've worried about that when things such as automatically uploading unformatted code have been mentioned, or when we are talking about uploading messages which are not direct interactions with the bot.

In this case, it's reasonable to temporarily upload after a user has directly used our features (bear in mind we automatically upload the results if they are too long).

DragonSenseiGuy:
> Perhaps we can send the user the link to delete the paste? You can largely copy the code here:
> 
> https://github.com/python-discord/bot/blob/cb4401b45978d90bb53cdbe4c03d7b89747562b2/bot/exts/utils/attachment_pastebin_uploader.py#L130-L133

I mean if they did a command in #python-discussion and then tried to find it later it would be hard, as that channel get lots of messages.

bot/decorators.py
DragonSenseiGuy: Fixed.

bot/decorators.py
DragonSenseiGuy: Fixed.

bot/decorators.py
DragonSenseiGuy: Do you want it changed? It can be done.

MarkKoz: I don't have a strong preference, but I figure we should leave it alone if there is no compelling reason to change it.

jb3: Yeah this was probably something I changed inadvertedly when I wrote something on mobile, it can be changed back.

DragonSenseiGuy: Should I change and commit then? Or should we just leave it as is?

bot/decorators.py
DragonSenseiGuy: Fixed

bot/decorators.py
DragonSenseiGuy: Resolved with the ternary changes.

bot/decorators.py
```diff
@@ -154,8 +156,34 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
+            print(paste_response.link)
+            msg += f"[your command]({paste_response.link})" if paste_response else "your command"
+            msg += f", {ctx.author.mention}:" if ping_user else ":"
+
+            await ctx.send(msg)
+
+            #Send a DM to the user about the redirect and pastebin removal
+            await ctx.author.send(
+                content=
+                f"Your command output was redirected to <#{Channels.bot_commands}>. [Click here](<{paste_response.removal}>) to delete the automatically uploaded copy of your original command."
```
swfarnsworth, line 184: When a string is too long, the preferred format is like this
```py
content = (
    "Lorem ipsum dolor sit amet, consectetur "
    f"adipiscing {elit}, sed do eiusmod tempor "
    "incididunt ut labore et dolore magna aliqua."
)
```
This leverages implicit string concatenation. Note that segments using f-string need to start with an f. The trailing spaces at the end of each line are literal single spaces that are not line breaks.

DragonSenseiGuy: Yeah, Just saw the lint errors, pushed a commit.

bot/decorators.py
```diff
@@ -154,8 +156,34 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
+
+            await ctx.send(msg)
+
+            #Send a DM to the user about the redirect and pastebin removal
+            await ctx.author.send(
+                content=
+                f"Your command output was redirected to <#{Channels.bot_commands}>."
+                f" [Click here](<{paste_response.removal}>) to delete the automatically"
+                "uploaded copy of your original command."
+            )
```
jb3, line 185:
```suggestion
            # Send a DM to the user about the redirect and paste removal
            await ctx.author.send(
                    f"Your command output was redirected to <#{Channels.bot_commands}>."
                    f" [Click here](<{paste_response.removal}>) to delete the pasted"
                    " copy of your original command."
            )
```

jb3 (changes requested):
Formatting and wording.

bot/decorators.py
```diff
@@ -154,8 +156,33 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
+            msg = ""
+
+            msg = "Here's the output of "
+            msg += f"[your command]({paste_response.link})" if paste_response else "your command"
+            msg += f", {ctx.author.mention}:" if ping_user else ":"
+
+            await ctx.send(msg)
+
+            # Send a DM to the user about the redirect and paste removal
+            await ctx.author.send(
```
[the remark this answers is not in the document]
jb3: I'd check the docs for what can be raised, Forbidden is definitely the one we need to catch.

bot/decorators.py
```diff
@@ -154,8 +157,34 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
+                try:
+                    paste_response = await send_to_paste_service(
+                        files=[PasteFile(content=ctx.message.content, lexer="markdown")],
+                        http_session=session,
+                    )
+                except PasteUploadError:
+                    log.exception(
+                        "Failed to upload message %d in channel %d to paste service when redirecting output",
+                        ctx.message.channel.id, ctx.message.id
+                    )
```
jb3, line 170: Arguments wrong way around, this was probably my snippet that I wrote on mobile, apologies.

```suggestion
                    log.exception(
                        "Failed to upload message %d in channel %d to paste service when redirecting output",
                        ctx.message.id, ctx.message.channel.id
                    )
```

bot/decorators.py
```diff
@@ -154,8 +157,37 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
+                try:
+                    paste_response = await send_to_paste_service(
+                        files=[PasteFile(content=ctx.message.content, lexer="markdown")],
+                        http_session=session,
+                    )
+                except PasteUploadError:
+                    log.exception(
+                        "Failed to upload message %d in channel %d to paste service when redirecting output",
+                        ctx.message.id, ctx.message.channel.id
+                    )
```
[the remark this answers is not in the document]
jb3: Previously, you defined `paste_link = None` such that if the network request failed the variable would still be defined, it would be picked up that there was no link and no hyperlink would be inserted.

In the linked commit the variable was removed instead of being changed to the new variable (`paste_response`). You need to add back an initialisation of that variable to `None` before with `ClientSession` operations.

See also the additional comment from Mark & myself regarding supporting the invocation deletion configuration flag.

DragonSenseiGuy:
> > Thanks. Looks good besides the comment about exception handling.
> > Maybe we should only be doing this when `RedirectOutput.delete_invocation` is true. But in practice this is a global option and the default is true, so it's probably not worth our time. I'm not sure why we even have a config option for that.
> 
> If we have the option we may as well implement the functionality. I think just wrapping the network request section in an if statement should cover it? Once the definition for `paste_response` is added back then as long as that remains set to `None` then the bot will just form the generic no-link message and send that.

So an `if` statement over the entire `aysnc with ClientSession() as session`?

jb3:
> [...]
> 
> So an `if` statement over the entire `aysnc with ClientSession() as session`?

Yes, as well as the variable declaration.

jb3 (approved):

bot/decorators.py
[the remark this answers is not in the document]
DragonSenseiGuy: @wookie184 Done, check it out!

bot/decorators.py
```diff
@@ -154,8 +157,39 @@ async def inner(self: Cog, ctx: Context, *args, **kwargs) -> None:
...
+            msg += f"[your command]({paste_response.link})" if paste_response else "your command"
+            msg += f", {ctx.author.mention}:" if ping_user else ":"
+
+            await ctx.send(msg)
+
+            try:
+                # Send a DM to the user about the redirect and paste removal
+                await ctx.author.send(
+                        f"Your command output was redirected to <#{Channels.bot_commands}>."
+                        f" [Click here](<{paste_response.removal}>) to delete the pasted"
```
decorator-factory, line 184: this will break if `paste_response` is `None`

(smh why is there no type checker in CI)

DragonSenseiGuy: I think the logs stop it from executing?

PR #3385, opened 2025-09-12
Update captization of PyPI for consistency with PyPI trademark and branding

onerandomusername:
https://pypi.org/trademarks/


wookie184 (approved):
Thanks

PR #3396, opened 2025-10-07
Fix incorrect variable used in mentions logic

wookie184:
resolved is always None here, as the logic is for the case when resolved is None. Instead we want to use the channel and message ID from ref directly, which are always defined here.

Closes https://github.com/python-discord/bot/issues/3374


jb3:
Merged in 89849c7aa

PR #3409, opened 2025-10-13
Implement URL normalization in code snippet handler

### LINKED ISSUE #3315: Unsafe URL construction in the `CodeSnippets` cog
Exploitation example:

```
<https://github.com/python/cpython/blob/main/../../../../users/octocat#L1>
```

Image

(discord unhelpfully resolves the `..`s)

The reason this is happening is [this regex](https://github.com/python-discord/bot/blob/46cbc19f9a1c7af0677b29e9a85d2d790e24c9e0/bot/exts/info/code_snippets.py#L18-L21) matching `../../../../users/octocat` as the filename and then blindly substituting it into [this URL template](https://github.com/python-discord/bot/blob/46cbc19f9a1c7af0677b29e9a85d2d790e24c9e0/bot/exts/info/code_snippets.py#L99-L103). `..` in a URL has a similar meaning to UNIX file paths: the parent directory. So the resulting URL (`https://api.github.com/repos/python/cpython/contents/../../../../users/octocat?ref=main`) means the same as `https://api.github.com/users/octocat?ref=main`.

Not sure if there's a good fix for this besides just not allowing `(\.|%2[eE]){2}` in the URL. (Yes, `.%2E` and `%2e%2E` do both mean "parent directory", don't you love URI)

onerandomusername:
Add URL normalization checks using yarl in code snippets.

closes #3315 

Thanks to @decorator-factory for the idea.

---

tested on @pydis-bot and deployed on @Monty-Python-Bot

### DIFF
--- a/bot/exts/info/code_snippets.py
+++ b/bot/exts/info/code_snippets.py
@@ -274,7 +274,7 @@ class CodeSnippets(Cog):
         for pattern, handler in self.pattern_handlers:
             for match in pattern.finditer(content):
                 # ensure that the matched URL meets url normalization rules.
-                # parsing with yarl resolves all parent urls such as `/../`,
+                # parsing an absolute url with yarl resolves all parent urls such as `/../`,
                 # we then check the regex again to make sure our groups stay the same
                 unsanitized = match.group(0)
                 normalized = str(yarl.URL(unsanitized))


jb3 (changes requested):
Looks good, should we add & pin yarl as we are now explicitly using it? (I know it's present as a dependency of aiohttp but we should probably pin it ourselves).

wookie184 (approved):
There is some potential for false positives here as there is nothing inherently wrong with an unnormalised URL. I don't think it's really an issue because I can only think of examples that you would have to craft on purpose, e.g. if you percentage encode part of the URL (e.g. if you had a file called `#` or `?` :D) and decide to do it in lowercase `%3f`, yarl will normalise to uppercase `%3F`.

This is definitely better than allowing arbitrary requests to be made, and I can't really think of a neater approach.

Ideally we would probably have a helper we can use everywhere when substituting stuff into URLs (possibly a nice use case for PEP 750...). Even this ends up being a bit difficult to do perfectly though, as some URL parameters need to be more permissive than others, and it's not always clear when to quote vs when to just reject something.

onerandomusername:
Tested on this branch, seems to work:
image

Meanwhile, on main:
image


raw url tested: `https://github.com/onerandomusername/ghre-tester/blob/9be3025d879813db5f6524338002539ad0264e45/README.md%3Fplain%3D1#L1`

PR #3417, opened 2025-10-21
refactor: update botstrap to be more beginner-friendly

### LINKED ISSUE #3414: More user friendly botstrap
Botstrap can be enhanced to provide a few additional useful features for a user using botstrap for the first time.

- if the bot isn't in the configured guild, it can print an invite link for itself
- it can update its application flags to support message and member intents

There's also the following issues:

 - If botstrap quits early, it only shows a warning: paired with the rest of the logging this isn't apparent that execution stopped early. This should be upped to an error or even fatal.
 - silence the 3rd party loggers from httpcore and httpx: they're very noisy and hide what isn't configured

Furthermore (though needing to be updated in the template, I received the following warnings while following the contributing guide.

Image

(this is with logging temporarily lowered for third parties)

This aspect would need to be fixed with an update to the template.

onerandomusername:
Update botstrap to be much more userfriendly. This silences noisy logs that we don't want to see or care about when downloading new server.

We also provide a success message when the entire system is successful.

I've also implemented some small caching in the botcore client to only make requests when necessary, and to only make one request to get all of the webhooks. While yes, webhooks are free, I left request logging enabled so the user could see what requests were made. I didn't see a good reason to silence those, both for transparency and because they already provide logging. 

We also provide an invite link if the bot isn't in the guild already, but it is configured.

I've also added support which sets the app's flags to enable message content intents and member intent if they aren't already enabled.

closes #3414

### DIFF
--- a/botstrap.py
+++ b/botstrap.py
@@ -278,6 +278,8 @@ class BotStrapper:
         self.client: DiscordClient = DiscordClient(guild_id=guild_id, bot_token=bot_token)
         self.env_file: Path = env_file
 
+        self.client_id = self.client.get_app_info()["id"]
+
     def __enter__(self):
         return self
 
@@ -289,6 +291,34 @@ class BotStrapper:
     ) -> None:
         self.client.__exit__(exc_type, exc_value, traceback)
 
+    def _find_webhook_for_channel(
+        self,
+        *,
+        webhook_name: str,
+        webhook_id: int | str,
+        channel_id: int | str,
+        webhooks: list[dict[str, object]],
+    ) -> dict[str, object] | None:
+        """Find a usuable webhook by its name or ID from a list of existing webhooks."""
+        matches: list[dict[str, object]] = []
+        # This matches to a list to prefer webhooks created by this application first,
+        # which may be encountered AFTER reaching a webhook with the same name.
+        for webhook in webhooks:
+            if not webhook.get("token"):
+                continue  # Webhook is unusable without a token
+            if webhook["id"] == str(webhook_id):
+                return webhook  # Keep existing configuration
+            if webhook["channel_id"] != str(channel_id):
+                continue  # Exclude webhooks from other channels
+            if webhook["application_id"] == str(self.client_id):
+                matches.insert(0, webhook)  # Prefer webhooks created by this application
+            elif webhook["name"] == webhook_name:
+                if webhook in matches:
+                    return webhook  # owned, and the name is the name, Use this one
+                matches.append(webhook)  # Fallback to matching by name
+
+        return matches[0] if matches else None
+
     def upgrade_client(self) -> bool:
         """Upgrade the application's flags if necessary."""
         if self.client.upgrade_application_flags_if_necessary():
@@ -299,12 +329,11 @@ class BotStrapper:
     def check_guild_membership(self) -> None:
         """Check the bot is in the required guild."""
         if not self.client.check_if_in_guild():
-            client_id = self.client.get_app_info()["id"]
             log.error("The bot is not a member of the configured guild with ID %s.", self.guild_id)
             log.warning(
                 "Please invite with the following URL and rerun this script: "
                 "https://discord.com/oauth2/authorize?client_id=%s&guild_id=%s&scope=bot+applications.commands&permissions=8",
-                client_id,
+                self.client_id,
                 self.guild_id,
             )
             raise BotstrapError("Bot is not a member of the configured guild.")
@@ -379,21 +408,19 @@ class BotStrapper:
         existing_webhooks = self.client.get_all_guild_webhooks()
         for webhook_name, configured_webhook in Webhooks.model_dump().items():
             formatted_webhook_name = webhook_name.replace("_", " ").title()
-            configured_webhook_id = str(configured_webhook["id"])
+            webhook_channel_id = all_channels[webhook_name]
 
-            for existing_hook in existing_webhooks:
-                existing_hook_id: str = existing_hook["id"]
+            webhook = self._find_webhook_for_channel(
+                webhook_name= formatted_webhook_name,
+                webhook_id=configured_webhook["id"],
+                channel_id=webhook_channel_id,
+                webhooks=existing_webhooks,
+            )
 
-                if existing_hook_id == configured_webhook_id or (
-                    existing_hook["name"] == formatted_webhook_name
-                    # This requires the normalized channel name matches the webhook attribute
-                    and existing_hook["channel_id"] == str(all_channels[webhook_name])
-                ):
-                    webhook_id = existing_hook_id
-                    break
-            else:
-                webhook_channel_id = all_channels[webhook_name]
+            if not webhook:
                 webhook_id = self.client.create_webhook(formatted_webhook_name, webhook_channel_id)
+            else:
+                webhook_id = webhook["id"]
 
             data[webhook_name + "__id"] = webhook_id
 


botstrap.py
```diff
@@ -1,16 +1,23 @@
...
+# we want to filter out the patching send typing notifications from bot core when we import to get constants
+logging.getLogger("pydis_core").setLevel(logging.WARNING)
+
+from bot.constants import Webhooks, _Categories, _Channels, _Roles  # noqa: E402
+from bot.log import get_logger  # noqa: E402
 
 load_dotenv()
-log = get_logger("Config Bootstrapper")
+log = get_logger("botstrap")
+# silence noisy httpcore logger
```
jb3: https://peps.python.org/pep-0008/#comments
```suggestion
# Silence noisy httpcore logger
```

botstrap.py
```diff
@@ -51,10 +76,10 @@ def __getitem__(self, item: str):
         try:
             return super().__getitem__(item)
         except KeyError:
-            log.warning(f"Couldn't find key: {item} in dict: {self.name} ")
+            log.fatal(f"Couldn't find key: {item} in dict: {self.name} ")
```
jb3:
```suggestion
            log.fatal(f"Couldn't find key: {item} in dict: {self.name}")
```

botstrap.py
```diff
@@ -65,64 +90,113 @@ class DiscordClient(Client):
...
+        )
         forum_channel_id = response.json()["id"]
         log.info(f"New forum channel: {channel_name_} has been successfully created.")
         return forum_channel_id
 
     def is_forum_channel(self, channel_id_: str) -> bool:
         """A boolean that indicates if a channel is of type GUILD_FORUM."""
-        response = self.get(f"/channels/{channel_id_}")
-        return response.json()["type"] == GUILD_FORUM_TYPE
+        return next(filter(lambda c: c["id"] == channel_id_, self.guild_channels)).get("type", None) == GUILD_FORUM_TYPE
```
jb3: This isn't very clean and channel `type` is a mandatory attribute from the API, split it up and remove the unnecessary `get()` usage.

I suggest a new helper method to return a channel from `self.guild_channels` and then this method just has to use that and check the type key.

I'd also remove the unnecessary `_` suffixes from the arguments to these helper methods.

botstrap.py
```diff
@@ -147,37 +220,47 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
...
 with DiscordClient(guild_id=GUILD_ID) as discord_client:
+    discord_client.upgrade_application_flags_if_necessary()
+
+    if not discord_client.check_if_in_guild():
+        client_id = discord_client.app_info.get("id")
+        log.error(f"The bot is not a member of the configured guild with ID {GUILD_ID}.")
+        log.warning(
+            f"Please invite with the following URL and rerun this script: https://discord.com/oauth2/authorize?client_id={client_id}&guild_id={GUILD_ID}&scope=bot+applications.commands&permissions=8"
+        )
+        sys.exit(42)
```
jb3: Why 42?

jb3 (changes requested):
Few review comments.

Can we also squash down the history a little bit? I don't think we need 21 commits for the changes here. Obviously keep the constants one separate but I think we can squash down a lot of the other ones.

botstrap.py
```diff
@@ -225,29 +306,40 @@ def create_webhook(self, name: str, channel_id_: int) -> str:
...
+            webhook_channel_id = int(all_channels[webhook_name])
+            webhook_id = discord_client.create_webhook(formatted_webhook_name, webhook_channel_id)
         config_str += f"webhooks_{webhook_name}__id={webhook_id}\n"
-        config_str += f"webhooks_{webhook_name}__channel={all_channels[webhook_name]}\n"
 
     config_str += "\n#Emojis\n"
-    config_str += "emojis_trashcan=🗑️"
+
+    # we add some replacements for some of the emojis used by the bot
+    # this is a best-effort and doesn't replace every emote
```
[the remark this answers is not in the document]
onerandomusername: I was working on reworking this segment to instead clone all of the configured emojis. No longer only making a best-effort approach, it is now full effort.

botstrap.py
onerandomusername: To be fair, given the template now has the forum channel in it, this could be removed entirely. In my testing the code that upgrades python-help doesn't run when a guild is created from the template.

botstrap.py
```diff
@@ -1,16 +1,23 @@
...
 from pathlib import Path
+from typing import Any, cast
 
 from dotenv import load_dotenv
-from httpx import Client, HTTPStatusError, Response
+from httpx import Client, Response
 
-from bot.constants import Webhooks, _Categories, _Channels, _Roles
-from bot.log import get_logger
+# we want to filter out the patching send typing notifications from bot core when we import to get constants
```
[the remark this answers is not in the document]
onerandomusername: Updated in https://github.com/python-discord/bot/pull/3417/commits/00c597e28255d2c84f69f7bee81433b2c997a9ec

botstrap.py
onerandomusername: https://github.com/python-discord/bot/pull/3417/commits/00c597e28255d2c84f69f7bee81433b2c997a9ec

onerandomusername:
@jb3 I've compressed the history and addressed your comments in the updated history. Comments should now follow pep8, formatted with ruff, and silenced pyright.

botstrap.py
```diff
@@ -147,37 +219,82 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
...
         return new_webhook["id"]
 
+    def list_emojis(self) -> list[dict[str, Any]]:
+        """Lists all the emojis for the guild."""
+        response = self.get(f"/guilds/{self.guild_id}/emojis")
+        return response.json()
+
+    def get_emoji_contents(self, emoji_id: str | int) -> bytes | None:
+        """Fetches the image data for an emoji by ID."""
+        # emojis are located at https://cdn.discordapp.com/emojis/821611231663751168.png?size=4096
```
jb3: I'd use path-params to document this and where that snowflake comes from (even though it's obvious). Also, not sure the size is necessary given we don't then use it for the subsequent request.
```suggestion
        # Emojis are located at https://cdn.discordapp.com/emojis/{emoji_id}.png
```

botstrap.py
```diff
@@ -137,8 +210,7 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
         channels = SilencedDict(name="Channels dictionary")
         categories = SilencedDict(name="Categories dictionary")
 
-        response = self.get(f"guilds/{self.guild_id}/channels")
-        server_channels = response.json()
+        server_channels = self.guild_channels
 
         for channel in server_channels:
```
jb3:
```suggestion
        for channel in self.guild_channels:
```

botstrap.py
```diff
@@ -63,66 +79,123 @@ def __getitem__(self, item: str):
...
+            self._app_info = cast("dict[str, Any]", resp.json())
+            return True
+
+        return False
+
+    def check_if_in_guild(self) -> bool:
+        """Check if the bot is a member of the guild."""
+        try:
+            _ = self.guild_info
+        except HTTPStatusError:
```
jb3: Maybe we can also check the status here? Avoids us having problems from a transient Discord error that we should surface in a different way.

jb3 (changes requested):

jb3 (changes requested):
Sorry for these comments all being in parts, just keep doing scans and picking up new bits.

Can we also standardise on whether you're using f-strings in this file or using `%s` interpolation? There is a mix currently (I appreciate it was f-strings before which is not correct, but if you're adding some f-strings and some %s strings we may as well unify).

botstrap.py
```diff
@@ -137,117 +218,252 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
...
+
+        payload = {
+            "name": new_name,
+            "image": f"data:image/png;base64,{base64.b64encode(emoji_data).decode('utf-8')}",
+        }
+
+        response = self.post(
+            f"/guilds/{self.guild_id}/emojis",
+            json=payload,
+            headers={"X-Audit-Log-Reason": f"Creating {new_name} emoji as part of PyDis botstrap"},
```
jb3:
```suggestion
            headers={"X-Audit-Log-Reason": "Creating emoji as part of PyDis botstrap"},
```

botstrap.py
```diff
@@ -137,117 +218,252 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
...
+        config["webhooks"] = self.sync_webhooks()
+        config["emojis"] = self.sync_emojis()
+
+        self.write_config_env(config, self.env_file)
+
+
+if __name__ == "__main__":
+    botstrap = BotStrapper(guild_id=GUILD_ID, env_file=ENV_FILE)
+    with botstrap:
+        botstrap.run()
```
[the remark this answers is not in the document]
onerandomusername: I would prefer to manage the DiscordClient internally on the botstrap class, even if we use contextlib to manage the context manager. 

DiscordClient should not be used by anything other than the BotStrap class, and an instance of it outside of botstrap doesn't need to exist.

botstrap.py
```diff
@@ -137,117 +221,259 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
...
+                log.info("Creating emoji %s", emoji_name)
+                emoji_id = self.client.clone_emoji(new_name=emoji_name, original_emoji_id=emoji_id)
+
+            data[emoji_config_name] = f"<:{emoji_name}:{emoji_id}>"
+
+        return data
+
+    def write_config_env(self, config: dict[str, dict[str, Any]], env_file: Path) -> None:
+        """Write the configuration to the specified env_file."""
+        # in order to support commented sections, we write the following
```
jb3:
```suggestion
        # In order to support commented sections, we write the following
```

botstrap.py
```diff
@@ -51,78 +71,142 @@ def __getitem__(self, item: str):
...
         self.guild_id = guild_id
+        self._app_info: dict[str, Any] | None = None
+        self._guild_info: dict[str, Any] | None = None
+        self._guild_channels: list[dict[str, Any]] | None = None
 
     @staticmethod
     def _raise_for_status(response: Response) -> None:
         response.raise_for_status()
 
+    @property
```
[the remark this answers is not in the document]
jb3: Maybe a good case for a switch to a method with https://docs.python.org/3/library/functools.html#functools.cache

botstrap.py
```diff
@@ -137,117 +221,259 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
...
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None = None,
+        exc_value: BaseException | None = None,
+        traceback: TracebackType | None = None,
+    ) -> None:
+        self.client.__exit__(exc_type, exc_value, traceback)
 
-    config_str += "\n#Categories\n"
+    def upgrade_client(self) -> bool:
```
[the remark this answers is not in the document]
onerandomusername: Added the code that consumes it.

botstrap.py
```diff
@@ -83,11 +83,37 @@ def __init__(self, guild_id: int | str):
...
+
+    def upgrade_application_flags_if_necessary(self) -> bool:
+        """
+        Set the app's flags to allow the intents that we need.
+
+        Returns a boolean defining whether changes were made.
+        """
+        # Fetch first to modify, not overwrite
+        current_flags = self.app_info.get("flags", 0)
+        new_flags = current_flags | 1 << 15 | 1 << 19
```
shtlrs: Those bit shifts are also "magic numbers", aren't they ? it'd better to assign each one to a variable indicating the flag.

botstrap.py
```diff
@@ -210,6 +220,17 @@ def create_webhook(self, name: str, channel_id_: int) -> str:
     if discord_client.upgrade_application_flags_if_necessary():
         log.info("Application flags upgraded successfully, and necessary intents are now enabled.")
 
+    if not discord_client.check_if_in_guild():
+        client_id = discord_client.app_info["id"]
```
shtlrs: I think it's better to encapsulate this with a property

botstrap.py
```diff
@@ -137,117 +193,289 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
...
+
+if __name__ == "__main__":
+    botstrap = BotStrapper(guild_id=GuildConstants.id, env_file=ENV_FILE, bot_token=BotConstants.token)
+    with botstrap:
+        changes_made = botstrap.run()
+    if changes_made:
+        log.info("Botstrap completed successfully. Updated configuration has been written to %s", ENV_FILE)
+    else:
+        log.info("Botstrap completed successfully. No changes were necessary.")
+    sys.exit(changes_made)
```
jb3: Why exit with an error if we make changes?

botstrap.py
```diff
@@ -137,117 +193,289 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
...
+
+
+if __name__ == "__main__":
+    botstrap = BotStrapper(guild_id=GuildConstants.id, env_file=ENV_FILE, bot_token=BotConstants.token)
+    with botstrap:
+        changes_made = botstrap.run()
+    if changes_made:
+        log.info("Botstrap completed successfully. Updated configuration has been written to %s", ENV_FILE)
+    else:
+        log.info("Botstrap completed successfully. No changes were necessary.")
```
jb3: Space.
```suggestion
    with botstrap:
        changes_made = botstrap.run()

    if changes_made:
        log.info("Botstrap completed successfully. Updated configuration has been written to %s", ENV_FILE)
    else:
        log.info("Botstrap completed successfully. No changes were necessary.")

```

botstrap.py
```diff
@@ -137,117 +193,289 @@ def get_all_channels_and_categories(self) -> tuple[dict[str, str], dict[str, str
...
 
+    def list_emojis(self) -> list[dict[str, Any]]:
+        """Lists all the emojis for the guild."""
+        response = self.get(f"/guilds/{self.guild_id}/emojis")
+        return response.json()
+
+    def get_emoji_contents(self, id_: str | int) -> bytes | None:
+        """Fetches the image data for an emoji by ID."""
+        # Emojis are located at https://cdn.discordapp.com/emojis/{emoji_id}.{ext}
+        response = self.get(f"{self.CDN_BASE_URL}/emojis/{id_!s}.webp")
```
jb3: Shouldn't fetch webp if you then submit `image/png`, even if Discord is accepting it it's not right.

jb3 (approved):

jb3 (approved):

decorator-factory (approved):

PR #3430, opened 2025-10-30
Python 3.14 and dependency upgrades

jb3:
- Updates to Python 3.14 in pyproject & Docker
- Upgrades uv to 0.9
- Upgrades all dependencies to latest available version
- Implements fixes for 3.13 -> 3.14 changes
- Implements linting fixes for newer versions of Ruff & Python

### DIFF
--- a/bot/exts/info/doc/_doc_item.py
+++ b/bot/exts/info/doc/_doc_item.py
@@ -4,11 +4,20 @@ from typing import NamedTuple
 class DocItem(NamedTuple):
     """Holds inventory symbol information."""
 
-    package: str  # Name of the package name the symbol is from
-    group: str  # Interpshinx "role" of the symbol, for example `label` or `method`
-    base_url: str  # Absolute path to to which the relative path resolves, same for all items with the same package
-    relative_url_path: str  # Relative path to the page where the symbol is located
-    symbol_id: str  # Fragment id used to locate the symbol on the page
+    package: str
+    """Name of the package name the symbol is from"""
+
+    group: str
+    """Interpshinx "role" of the symbol, for example `label` or `method`"""
+
+    base_url: str
+    """Absolute path to to which the relative path resolves, same for all items with the same package"""
+
+    relative_url_path: str
+    """Relative path to the page where the symbol is located"""
+
+    symbol_id: str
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -3,28 +3,28 @@ authors = [
...
 dependencies = [
     "pydis-core[all]==11.8.0",
-    "aiohttp==3.12.14",
-    "arrow==1.3.0",
-    "beautifulsoup4==4.12.3",
+    "aiohttp==3.13.2",
+    "arrow==1.4.0",
+    "beautifulsoup4==4.14.2",
     "colorama==0.4.6; sys_platform == \"win32\"",
     "coloredlogs==15.0.1",
```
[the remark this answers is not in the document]
jb3: [a7b28c9](https://github.com/python-discord/bot/pull/3430/commits/a7b28c9da607d9cc8d069d3c786497bb4462650c)

.github/workflows/lint-test.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         uses: astral-sh/setup-uv@v7
         with:
           enable-cache: true
+          resolution-strategy: "lowest"
```
onerandomusername: What is the motivation behind this in ci only?

jb3: Where else would it be? It's a parameter for the action, not for uv itself (despite uv having a identically named parameter).

https://github.com/astral-sh/setup-uv/blob/30ce38e20653a26dd16caa082b010cec334ccf56/src/download/download-version.ts#L180-L183

We want to:
- Always use the lowest confirmed working version of `uv` in CI (for us it'll use `0.9.7` since our specifier is `>=0.9.7`)
- Simultaneously, we don't want to annoy people developing locally with newer versions of uv by adding an upper bound. We don't want CI to magically switch to 0.10.0 but it doesn't matter if we do locally.

onerandomusername: Ah, thanks! Its a slightly confusingly named parameter.

onerandomusername:
I agree with jc, bumping ruff can happen separately.

ChrisLovering:
can we revert this Pr and separate into multiple PRs so i can properly review?

PR #3450, opened 2025-12-12
Support unicode urls filtering

mbaruh:


bot/exts/filtering/_filters/domain.py
```diff
@@ -36,11 +37,11 @@ class DomainFilter(Filter):
 
     async def triggered_on(self, ctx: FilterContext) -> bool:
         """Searches for a domain within a given context."""
-        domain = tldextract.extract(self.content).registered_domain.lower()
+        domain = tldextract.extract(self.content).top_domain_under_public_suffix.lower()
```
L3viathan: `top_domain_under_public_suffix` does `"foo.bar.com" -> "bar.com"`, but `"foo.bar.co.uk" -> "bar.co.uk"` I'm guessing?

What does this have to do with decoding urlencoded stuff?

mbaruh: > What does this have to do with decoding urlencoded stuff?

It doesn't, which is why it's in a separate commit :D just replaced a deprecated attribute

PR #3461, opened 2026-01-25
Add user-facing message when deprecated cban alias is used

neevets:
This adds a helpful response when users attempt to use the deprecated `cban`
alias, explaining why it was removed and suggesting the correct alternatives
(`cleanban` or `compban`).

This should reduce confusion, especially for long-time moderators.


jb3 (changes requested):
After speaking with Moderators, we want to bring back the `cban` alias for the `cleanban` command.

Please can you adjust this PR to implement this (should just be undoing the change & adding the alias to the list in the cleanban command).

PR #3482, opened 2026-03-24
fix: Handling of site connection issues during outage 

### LINKED ISSUE #2918: Handling of site connection issues during outage.
Currently, under somewhat specific circumstances the bot might result in few Cogs/Extensions not being loaded if site will drop the connection or otherwise will fail to return relevant information to the bot. Most notably the bot will not load filters cog, superstarify, reminders and python_news.

This issue happens with no indication of a failure to moderators as such bot might end up in a state where filters will be disabled and mods will be none the wiser. 

The bot should:
- Alert moderators that such a thing occurred (filters weren't loaded)
- Potentially try to gracefully handle the issue and try to reconnect / retry later.

rippyboii:
## Summary
This PR improves bot startup reliability and moderator visibility when extensions/cogs fail to load.

- Aggregate extension + cog load failures during startup and report them as a **single** alert in `#mod-log`.
- Add **retry + exponential backoff** for cogs that depend on external sites/APIs (e.g., temporary 5xx/429/timeouts), with clear **mod notifications** on final failure.
- Add unit tests to validate retry behavior, error classification, and startup reporting.

During `setup_hook`, extensions are loaded concurrently. When an extension/cog fails due to a transient outage (rate limits, 5xx, timeouts), failures can either:
- stop startup unexpectedly, or
- fail noisily/fragmentedly, making it hard to see *what* broke and *why*.

This change standardizes both **resilience** (retry when appropriate) and **visibility** (one clean startup report + targeted alerts).

### 1) Startup failure aggregation (single `#mod-log` alert)
- Added `utils/startup_reporting.py` to format a standardized startup failure message.
- Updated `bot.py` to:
  - collect extension + cog load failures (import/setup/add_cog)
  - wait for all load tasks to complete
  - send one aggregated alert summarizing all failures
- Reporting is defensive: it does **not** crash if the log channel is unavailable.
- Startup continues for non-critical failures.

### 2) Retry + backoff for external/API-dependent cogs
Implemented retry logic with exponential backoff and explicit “retriable vs non-retriable” classification, plus moderator notifications when retries are exhausted.

Covered cogs include:
- **Filtering**: 3 attempts with backoff (1s, 2s, 4s); retries on `HTTP 429`, `HTTP 5xx`, `TimeoutError`, `OSError`; final failure logs + alerts `#mod-alerts`.
- **Reminders**: retry count is configured via `URLs.connect_max_retries`; warns are logged to Sentry; final failure posts to `#mod-log`.
- **PythonNews**: retries on `408`, `429`, `5xx`, `TimeoutError`, `OSError`; on max retries logs + alerts `mod_alerts` and re-raises to stop startup.
- **Superstarify cogs**: added retry + notification and corresponding tests.

## Tests / Verification
- Added unit tests covering:
  - retry-then-success
  - max-retries then alert + failure behavior
  - non-retriable errors
  - retry classification logic
  - aggregated startup failure reporting for faulty extensions/cogs

Suggested checks:
1. `uv run task test`
2. Run the bot and simulate a faulty extension/cog load to confirm a **single** aggregated `#mod-log` startup alert.

## Moderator Alert in Discord:
mod_alert

Closes #2918


bot/exts/filtering/filtering.py
```diff
@@ -108,7 +110,31 @@ async def cog_load(self) -> None:
...
+                backoff_seconds = URLs.connect_initial_backoff * (2 ** (attempt - 1))
+                log.warning(
+                    "Failed to load filtering data (attempt %d/%d). Retrying in %d second(s): %s",
+                    attempt,
+                    URLs.connect_max_retries,
+                    backoff_seconds,
+                    error
+                )
+                await asyncio.sleep(backoff_seconds)
+
```
rf20008, line 137: I think you repeated this backoff logic many times. Consider moving it to self.bot.api_client.get()

bot/exts/info/python_news.py
```diff
@@ -46,19 +49,45 @@ def __init__(self, bot: Bot):
...
+                    self.seen_items[mailing_list["name"]] = set(mailing_list["seen_items"])
+
+                with sentry_sdk.start_span(description="Update site with new mailing lists"):
+                    for mailing_list in ("pep", *constants.PythonNews.mail_lists):
+                        if mailing_list not in self.seen_items:
+                            await self.bot.api_client.post("bot/mailing-lists", json={"name": mailing_list})
+                            self.seen_items[mailing_list] = set()
+
+                self.fetch_new_media.start()
+                return
```
rf20008, line 69: Why is this in the `try`?

bot/utils/startup_reporting.py
```diff
@@ -0,0 +1,55 @@
...
+    """Formats and sends one aggregated startup failure alert to moderators."""
+
+    async def notify(self, bot: Bot, failures: Mapping[str, BaseException], channel_id: int = Channels.mod_log) -> None:
+        """Notify moderators of startup failures."""
+        if not failures:
+            return
+
+        if bot.get_channel(channel_id) is None:
+            # Can't send a message if the channel doesn't exist, so log instead
+            log.warning("Failed to send startup failure report: mod_log channel not found.")
```
rf20008: should the failures be logged in the error?

bot/bot.py
```diff
@@ -77,3 +85,53 @@ async def on_error(self, event: str, *args, **kwargs) -> None:
...
+
+        async def _load_one(extension: str) -> None:
+            try:
+                await self.load_extension(extension)
+                log.info(f"Extension successfully loaded: {extension}")
+
+            except BaseException as e:
+                self.extension_load_failures[extension] = e
+                log.exception(f"Failed to load extension: {extension}")
+                raise
```
rf20008, line 126: is there a reason this is defined within _load_extensions instead of it being a method of Bot?

bot/exts/filtering/filtering.py
rippyboii: Hi, Thank you for your review, 

I’m a bit hesitant to put retries directly into `api_client.get()` because that may affect all GET calls globally, including cases where retries may be not necessary or need different behavior/logging.

If you think this might not be the problem, I can definately move it to the `self.bot.api_client.get()`

bot/exts/info/python_news.py
rippyboii: Hi, 

The intent was to make the whole startup sync step atomic and retry on transient API failures from both the `get("bot/mailing-lists")` and `post("bot/mailing-lists")` calls.

`except` only retries `is_retryable_api_error(error)` so the local logic errors still raise immediately.

also the `try`  covers all the startup sequence, so the transient failure  in either get or the post should trigger the clean retry method. right?

bot/utils/startup_reporting.py
rippyboii: Ah right! Thank you for pointing it out. 

I think a missing `mod_log channel` means the failure report was silently dropped, which is an error, not just a warning. I will update it soon.

rippyboii: I have updated it with proper error logging. Could you check it again?

bot/bot.py
rippyboii: Hi,

I defined `_load_one` inside `_load_extensions` because it’s only used there.  I didn't realized we could move it to private `Bot` method. 

If you want, I’m happy to move it to a private `Bot` method (e.g. `_load_single_extension`). Shall I?

PR #3496, opened 2026-04-24
Fix `AttributeError` when reminder `jump_url` is `None`

oliveman-au:
`send_reminder` crashes with an `AttributeError` when `jump_url` is `None`. This can happen for older reminders created before the field was added to the API.

`reminder.get("jump_url")` returns `None` if the key is absent, but `.split()` was being called on it unconditionally, which causes the reminder to never be delivered.

This fix guards against `None` before calling `.split()` and skips the jump URL embed line if there's no URL, falling back to a plain `channel.send` instead.

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -383,11 +383,15 @@ class Reminders(Cog):
             mentionable.mention async for mentionable in self.get_mentionables(reminder["mentions"])
         ])
 
-        jump_url = reminder.get("jump_url")
-        embed.description += f"\n[Jump back to when you created the reminder]({jump_url})"
-        partial_message = channel.get_partial_message(int(jump_url.split("/")[-1]))
+        jump_url = reminder.get("jump_url") or ""
+        if jump_url:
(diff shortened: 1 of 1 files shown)


L3viathan:
Does this issue actually occur? When was the field added to the API?

oliveman-au:
> Does this issue actually occur? When was the field added to the API?

yeah it can, any reminder made before `jump_url` was added to the api would have it as `None`. the `.get()` instead of direct key access kinda hints the original dev knew it might be missing but just didnt guard the `.split()` call on the next line

L3viathan:
The reminders cog was added in 2018, the `jump_url` in 2019.
There's also no reminders in the actual DB without a `jump_url`, so no, it can't happen anymore.

oliveman-au:
> The reminders cog was added in 2018, the `jump_url` in 2019. There's also no reminders in the actual DB without a `jump_url`, so no, it can't happen anymore.

thanks for checking that, good to know. i'll close this one

oliveman-au:
closing this as the maintainers confirmed all reminders in the db have jump_url set so this can't actually occur anymore

PR #3499, opened 2026-04-24
Fix `AttributeError` in `log_uncached_deleted_message` when channel is `None`

oliveman-au:
`self.bot.get_channel()` can return `None` if the channel isn't in the bot's cache, but `channel.category` was being accessed immediately after with no None check, causing an `AttributeError`.

This is particularly impactful because `log_uncached_deleted_message` is called for every deleted message that isn't in the cache, which on a large server happens constantly. Every one of those events would silently crash the handler instead of logging.

Added an early return with a trace log if the channel can't be found, matching the pattern used elsewhere in the codebase.

### DIFF
--- a/bot/exts/moderation/modlog.py
+++ b/bot/exts/moderation/modlog.py
@@ -589,6 +589,9 @@ class ModLog(Cog, name="ModLog"):
             return
 
         channel = self.bot.get_channel(event.channel_id)
+        if channel is None:
+            log.trace(f"Channel {event.channel_id} not found, skipping uncached message delete log.")
+            return
 
         if channel.category:
             response = (


jb3 (changes requested):
This could probably be a higher log level, we should have some way of seeing how frequently this is happening.

wookie184:
I believe this is unnecessary as the check is already done in `is_channel_ignored`:
https://github.com/python-discord/bot/blob/a04ae25152102734e9f5ca13930ec83e0fba9b33/bot/exts/moderation/modlog.py#L476-L481

It's nice to have this code written in a way that a type checker can understand, but adding a check that will never trigger makes the code less clear.

I would support making this more type checker friendly, but that would need more discussion and I don't think this is the neatest approach so I will close this.

PR #3500, opened 2026-04-25
Fix false error message when deleting reminders with duplicate IDs

oliveman-au:
When running `!remind delete` with duplicate IDs (e.g. `!remind delete 123 123 123`), the reminder is deleted successfully but the bot incorrectly shows "The other reminder(s) could not be deleted as they're either locked, belong to someone else, or don't exist."

This happens because the check compares `len(deleted_ids)` against `len(ids)`, but the loop iterates over `set(ids)` to deduplicate. So passing 3 identical IDs results in 1 deletion but `len(ids)` is still 3, making it look like 2 failed.

Fixed by comparing against `len(set(ids))` instead, which matches what was actually attempted.

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -678,7 +678,7 @@ class Reminders(Cog):
             title = random.choice(POSITIVE_REPLIES)
             deletion_message = f"Successfully deleted the following reminder(s): {', '.join(deleted_ids)}"
 
-            if len(deleted_ids) != len(ids):
+            if len(deleted_ids) != len(set(ids)):
                 deletion_message += (
                     "\n\nThe other reminder(s) could not be deleted as they're either locked, "
                     "belong to someone else, or don't exist."


oliveman-au:
I Reproduced The Bug In The Image

Screenshot 2026-04-25 205620

lemonsaurus (approved):
one-liner fix, code looks good, you've proven it's a real bug, even added a screenshot of the bug being reproduced - nice! Good PR. Thanks for the contribution! 👍🏼

jb3 (changes requested):
We should reuse the already created set from earlier in the command, that's going to be a cleaner solution here and avoids a set creation twice for the same list (and solves other issues like how passing the same reminder ID in would go above the 5 reminder limit, even if it's the same reminder).

I suggest instead you do the set conversion earlier in the command (pretty much the first thing) and replace subsequent `set(ids)` invocations with that.

jb3:
Should be combined into one PR across all reminder functionality. Also please see my review comment.

PR #3501, opened 2026-04-26
Fix silent placeholder when editing reminder content with empty reply

oliveman-au:
When running `!remind edit content <id>` while replying to a message with no text content (e.g. an image-only message), the bot silently sets the reminder content to `*See referenced message.*` and confirms "That reminder has been edited successfully!" — giving the user no indication their content was replaced with a useless placeholder.

When the reminder fires, the user just gets pinged with `*See referenced message.*` and has no idea what they were supposed to remember.

Fixed by returning `None` from `try_get_content_from_reply` when the referenced message has no content, sending an error message to the user, and returning early from `edit_reminder_content` so the edit never goes through.

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -678,7 +678,7 @@ class Reminders(Cog):
             title = random.choice(POSITIVE_REPLIES)
             deletion_message = f"Successfully deleted the following reminder(s): {', '.join(deleted_ids)}"
 
-            if len(deleted_ids) != len(ids):
+            if len(deleted_ids) != len(set(ids)):
                 deletion_message += (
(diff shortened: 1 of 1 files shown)


oliveman-au:
reproduced this - ran `!remind edit content` while replying to an image-only message with no text. bot said "That reminder has been edited successfully!" with no warning, but the content was silently replaced with `*See referenced message.*`. when the reminder fires the user would just get pinged with that placeholder and have no idea what they were supposed to remember.

Screenshot 2026-04-26 102921

oliveman-au:
also reproduced in `new_reminder` — replying to an image-only message with `!remind 1d` (no content) silently creates the reminder with `*See referenced message.*` as the content. added a second commit to fix that path too.

Screenshot 2026-04-26 103431

jb3:

PR #3502, opened 2026-04-26
Reject past expiration dates when editing reminder duration

oliveman-au:
`!remind edit duration` accepts past dates with no validation. The bot confirms "That reminder has been edited successfully!" and then immediately fires the reminder since the expiration is already in the past.

Added a check to reject the edit if the provided expiration is before the current time, matching the expected behaviour of the original `!remind new` command.


oliveman-au:
reproduced - edited a reminder to expire in 1 second, bot confirmed the edit with a date in the past then immediately fired the reminder. no validation that the new expiration is in the future.

Screenshot 2026-04-26 104252

jb3:

PR #3509, opened 2026-05-07
Fix off-by-one errors and silent reminder deletion in reminders.py

### LINKED ISSUE #3507: Multiple off-by-one errors and silent reminder deletion with no user notification in reminders.py
The reminder limit for regular users is defined as:

```python
MAXIMUM_REMINDERS = 5
```

However, the check in `new_reminder` uses a strict greater-than comparison:

```python
if len(active_reminders) > MAXIMUM_REMINDERS:
    await send_denial(ctx, "You have too many active reminders!")
    return
```

Because `> 5` only triggers when a user already has **6** reminders, a user can successfully create a **6th reminder** before being blocked.

The fix would be changing `>` to `>=`.

**Steps to reproduce (as a regular member in `#bot-commands`):**

1. Run `!remind new 7d reminder 1` through `!remind new 7d reminder 6` six times
2. All six succeed — the denial is never triggered

**Expected:** denial on the 6th attempt
**Reproduced on:** `bot/exts/utils/reminders.py`, line 478

Proof shown below in images or can be recreated with steps above

Image
Image

oliveman-au:
Found a few bugs in `bot/exts/utils/reminders.py` while going through the reminders codebase. Two off-by-one errors and a silent deletion issue when the bot loses channel cache.

---

### 1. Off-by-one — users could create 6 reminders despite a limit of 5

The check was using `>` instead of `>=`, so the 6th reminder always slipped through before the denial fired.

```python
# Before
if len(active_reminders) > MAXIMUM_REMINDERS:

# After
if len(active_reminders) >= MAXIMUM_REMINDERS:
```

---

### 2. Off-by-one — users could delete 6 reminders at once despite a limit of 5

Same mistake in `delete_reminder`, which also directly contradicted the docstring saying "up to and including 5".

```python
# Before
if len(ids) > 5:

# After
if len(ids) >= 5:
```

---

### 3. Reminders silently deleted with no user notification when channel cache is lost

`get_channel()` only checks the local cache, so after a bot restart the channel returns `None` and the reminder gets permanently deleted with no DM, no message, nothing to the user. Switched to `get_or_fetch_channel()` as suggested by wookie184, and added a DM to the user if the channel truly can't be found so they at least get their reminder content back. Also updated both call sites to `await` the method since it's now async.

```python
# Before
def ensure_valid_reminder(self, reminder: dict) -> tuple[bool, discord.TextChannel]:
    channel = self.bot.get_channel(reminder["channel_id"])
    ...

# After
async def ensure_valid_reminder(self, reminder: dict) -> tuple[bool, discord.TextChannel | None]:
    channel = await get_or_fetch_channel(self.bot, reminder["channel_id"])
    ...
```

### DIFF
--- a/bot/exts/utils/reminders.py
+++ b/bot/exts/utils/reminders.py
@@ -10,6 +10,7 @@ from discord import Interaction
 from discord.ext.commands import Cog, Context, Greedy, group
 from pydis_core.site_api import ResponseCodeError
 from pydis_core.utils import scheduling
+from pydis_core.utils.channel import get_or_fetch_channel
 from pydis_core.utils.members import get_or_fetch_member
 from pydis_core.utils.scheduling import Scheduler
 
@@ -232,7 +233,7 @@ class Reminders(Cog):
         now = datetime.now(UTC)
 
         for reminder in response:
-            is_valid, *_ = self.ensure_valid_reminder(reminder)
+            is_valid, *_ = await self.ensure_valid_reminder(reminder)
             if not is_valid:
                 continue
 
@@ -244,19 +245,28 @@ class Reminders(Cog):
             else:
                 self.schedule_reminder(reminder)
 
-    def ensure_valid_reminder(self, reminder: dict) -> tuple[bool, discord.TextChannel]:
-        """Ensure reminder channel can be fetched otherwise delete the reminder."""
-        channel = self.bot.get_channel(reminder["channel_id"])
-        is_valid = True
+    async def ensure_valid_reminder(self, reminder: dict) -> tuple[bool, discord.TextChannel | None]:
+        """Ensure reminder channel can be fetched otherwise notify the user and delete the reminder."""
+        try:
+            channel = await get_or_fetch_channel(self.bot, reminder["channel_id"])
+        except discord.HTTPException:
+            channel = None
+
         if not channel:
-            is_valid = False
             log.info(
                 f"Reminder {reminder['id']} invalid: "
-                f"Channel {reminder['channel_id']}={channel}."
+                f"Channel {reminder['channel_id']} could not be fetched."
             )
+            user = self.bot.get_user(reminder["author"])
+            if user:
+                await user.send(
+                    f"Your reminder could not be delivered because the channel it was set in "
+                    f"is no longer accessible.\n**Reminder content:** {reminder['content']}"
+                )
             scheduling.create_task(self.bot.api_client.delete(f"bot/reminders/{reminder['id']}"))
+            return False, None
 
-        return is_valid, channel
+        return True, channel
 
     @staticmethod
     async def _send_confirmation(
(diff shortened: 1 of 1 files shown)


bot/exts/utils/reminders.py
```diff
@@ -659,7 +669,7 @@ async def _delete_reminder(self, ctx: Context, id_: int) -> bool:
     @remind_group.command("delete", aliases=("remove", "cancel"))
     async def delete_reminder(self, ctx: Context, ids: Greedy[int]) -> None:
         """Delete up to (and including) 5 of your active reminders."""
-        if len(ids) > 5:
+        if len(ids) >= 5:
             await send_denial(ctx, "You can only delete a maximum of 5 reminders at once.")
             return
```
wookie184, line 674: Actually, surely this change is wrong...

If you can delete up to 5 reminders, you only deny if there are `>5` reminders, so the original code is correct.

jchristgit:
@oliveman-au did you use LLM assistance to code these changes? If so, how much?

bot/exts/utils/reminders.py
```diff
@@ -244,19 +245,28 @@ async def cog_load(self) -> None:
...
-    def ensure_valid_reminder(self, reminder: dict) -> tuple[bool, discord.TextChannel]:
-        """Ensure reminder channel can be fetched otherwise delete the reminder."""
-        channel = self.bot.get_channel(reminder["channel_id"])
-        is_valid = True
+    async def ensure_valid_reminder(self, reminder: dict) -> tuple[bool, discord.TextChannel | None]:
+        """Ensure reminder channel can be fetched otherwise notify the user and delete the reminder."""
+        try:
+            channel = await get_or_fetch_channel(self.bot, reminder["channel_id"])
+        except discord.HTTPException:
+            channel = None
```
wookie184, line 253: Maybe we should be more specific about the exceptions we catch here, e.g. just notfound and forbidden, to try and avoid cancelling reminders in the case of a discord outage.

bot/exts/utils/reminders.py
```diff
@@ -244,19 +245,28 @@ async def cog_load(self) -> None:
...
                 f"Reminder {reminder['id']} invalid: "
-                f"Channel {reminder['channel_id']}={channel}."
+                f"Channel {reminder['channel_id']} could not be fetched."
             )
+            user = self.bot.get_user(reminder["author"])
+            if user:
+                await user.send(
+                    f"Your reminder could not be delivered because the channel it was set in "
+                    f"is no longer accessible.\n**Reminder content:** {reminder['content']}"
+                )
```
wookie184, line 265: Should we also be attempting to DM people that were also scheduled to be mentioned in the reminder? (in reminder['mentions']`)?

wookie184 (changes requested):
One of the changes seems incorrect from a quick look. Have not fully reviewed/tested the rest yet, but some initial comments.

Issue #2328, opened 2022-11-21
Simplify bot responses to detected unformatted code


Robin5605:
[Link to internal discussion](https://discord.com/channels/267624335836053506/551789653284356126/1044087416815894651). We were discussing how certain responses from the bot could be simplified. Especially the responses that are sent when the bot detected unformatted or improperly formatted code, as these happen quite a bit and can be disruptive to chat when there's such a large embed containing a lot of information to take in at once.

Please leave your thoughts on if and how we should condense these responses to be more concise and short.

PR #3513, opened 2026-06-01
Auto-formatting for messages with improperly formatted code blocks and extended code block detection

SamuelRoettgermann:
Addressing the long-standing issue https://github.com/python-discord/bot/issues/2328 .

This PR tries to solve this by:
1. Extending what kind of code-block mistakes are detected
2. Adding auto-formatting for easy-to-fix messages

========

Reason for why 1 is necessary: 
Our current system only detects unformatted code blocks if they
A) have 3 backticks at the front and back, but do not specify a language or
B) don't have any backticks at all (and the entire message is valid Python code)

This PR extends this to also detect unformatted code blocks if they:
C) have N (N != 3) backticks at the front and back or
D) have any sort of whitespace after the language specifier rather than just a newline

With these changes the bot now also detects common mistakes such as:

\`¶
x·=·4¶
y·=·3¶
z·=·4¶
print("abc")¶
\`¶

or:

\`\`\`py·¶
x·=·4¶
y·=·3¶
z·=·4¶
print("abc")¶
\`\`\`¶

========

Reason for why 2 is necessary: 
Rather than "spamming" an explanatory message, that may or may not then be applied. Even if the advice is applied, it is a common pattern for users to post an entirely new message, rather than editing their old message, which only then would actually remove the explanatory embed, thereby leading to even more "spam".

In the process of Markdown-formatting the code, I also use Black to properly format the Python code. I chose Black for its simplicity, but eventually it might be an idea to exchange this for ruff, possibly even reusing our own linting rules.

This PR only fixed messages with multiple code-blocks if all code-blocks are incorrectly formatted, thus keeping the expectation that "if the user has one properly formatted code block we can assume he knows what he's doing, even if there are other broken code blocks.

========

Please tell me what you think about the approach and the idea in general. But please also tell me what things are necessary still. Like, for example, a button to let the recipient (or a mod) delete the message again, or if we need a cooldown, or if you find any bugs. Maybe make the process opt-in by the bot reacting with some emoji and a very brief message/embed informing them of their option?


L3viathan:
I think detecting more cases is good, but deleting the user's message and formatting it with black is bad, for several reasons. One of them is that I don't like prescribing how something _must_ look like, the other one is that it prevents the user from deleting the message. The only other case that I can think of where we do this is posting large amounts of code to our pastebin, but that only happens after user confirmation.

PR #3514, opened 2026-06-05
Detect More Faulty Codeblocks

SamuelRoettgermann:
## Preface
Originally part of https://github.com/python-discord/bot/pull/3513 , but the main feature of that PR still needs some tuning.

## Summary
This PR extends what kind of codeblock mistakes are detected.

## Current system
Our current system only detects unformatted code blocks if they
A) have 3 backticks at the front and back, but do not specify a language or
B) don't have any backticks at all (and the entire message is valid Python code)

## Proposed system
This PR extends this to also detect unformatted code blocks if they:
C) have N (N != 3) backticks at the front and back or
D) have any sort of whitespace after the language specifier rather than just a newline

## Examples
With these changes the bot now also detects common mistakes such as:

\`¶
x·=·4¶
y·=·3¶
z·=·4¶
print("abc")¶
\`¶

or:

\`\`\`py·¶
x·=·4¶
y·=·3¶
z·=·4¶
print("abc")¶
\`\`\`¶

that were previously not detected.

### DIFF
--- a/bot/exts/info/codeblock/_parsing.py
+++ b/bot/exts/info/codeblock/_parsing.py
@@ -182,7 +182,7 @@ def parse_bad_language(content: str) -> BadLanguage | None:
     )
 
 
-def _get_leading_spaces(content: str) -> int:
+def _get_leading_spaces(content: str) -> int | None:
     """Return the number of spaces at the start of the first line in `content`."""
     leading_spaces = 0
     for char in content:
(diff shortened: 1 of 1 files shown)


L3viathan (approved):
Looks good to me.

ChrisLovering (approved):
Agreed, looks good to me too!

PR #3516, opened 2026-06-06
Implement `!tunnel` command

### LINKED ISSUE #3512: A `!tunnel` command for ease of redirecting conversations
This issue is mostly for keeping track of what was proposed (by me) in [#community-meta](https://discord.com/channels/267624335836053506/429409067623251969/1508570756160815194)

The first iteration of the implementation (what this issue is for) would target the following [idea by](https://discord.com/channels/267624335836053506/429409067623251969/1508590046696177781) @jb3:
> I'm envisaging it being fairly lightweight:
> 
> ➡️ Conversation continued in ⁠[#ot0-yes-do-as-i-sayǃ]()
> 
> ↩️ Conversation tunneled by [@joe]() from ⁠[#python-discussion]()

> (where those are message links, not channel links) 

[Original proposal](https://discord.com/channels/267624335836053506/429409067623251969/1508570756160815194):
> I have a proposal for a `!tunnel` feature for Lancebot
> 
> Suppose a discussion in [#python-discussion](https://discord.com/channels/267624335836053506/267624335836053506) is starting to veer off topic, so you'd like to move to one of the ot channels, but there isn't an easy way to cross-link the now-separated strands of conversation for some future reference (somebody either trying to find the beginning of the conversation going from ot backwards, or someone trying to find where the conversation went going from python-discussion forward). But this feature would create the bi-directional "portal" links, a link to a message in the ot channel in python-discussion and that message in the ot channel containing a link back to the tunnel-entry message that points back to itself.
> 
> Implementation-wise, `!tunnel` would by default pick one of the ot channels with the oldest recent message and use that, but otherwise it'd be possible to also specify a channel you want to tunnel to. The source would naturally always be the current channel the command is invoked in (though I guess, it could support a "from" keyword-only argument for the the source too, I don't see it being used that often? I guess if you wanted to invoke this command in one of the ot channels to bring a conversation from say python-discussion to there?).
> 
> So then, after the source and destination channels are sorted out, it would post a message in the source channel, post a message in the destination channel with a link to the message in the source channel, and then edit the message in the source channel with the link to the message in the destination channel.
> Mildly important, but the messages should contain text that is portal-technology-themed.
> 
> This might help reduce friction in moving conversations to another channel.

See also: [#dev-contrib](https://discord.com/channels/267624335836053506/635950537262759947/1509092857557024920)

Matiiss:
Fixes https://github.com/python-discord/bot/issues/3512

### DIFF
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -55,8 +55,8 @@ class Tunnel(commands.Cog):
         if source_channel.id == destination_channel.id:
             raise BadArgument("Source and destination channels cannot be the same")
 
-        source_message_template = "➡️ Conversation tunneled to {location}"
-        destination_message_template = "↩️ Conversation tunneled from {location}"
+        source_message_template = f"➡️ Conversation continued at {{location}} (by <@{ctx.author.id}>)"
+        destination_message_template = f"↩️ Conversation continued from {{location}} (by <@{ctx.author.id}>)"
 
         source_message = await source_channel.send(
             content=source_message_template.format(location=destination_channel.jump_url)
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -1,6 +1,6 @@
 import discord
 from discord.ext import commands
-from discord.ext.commands import BadArgument
+from discord.ext.commands import BadArgument, guild_only
 
 from bot.bot import Bot
 from bot.constants import Channels
@@ -33,6 +33,7 @@ class Tunnel(commands.Cog):
             self.channel_id_to_timestamp[channel_id] = last_message.created_at.timestamp()
 
     @commands.command()
+    @guild_only()
     async def tunnel(
         self,
         ctx: commands.Context,
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -17,8 +17,12 @@ class Tunnel(commands.Cog):
         self.bot = bot
         self.channel_id_to_timestamp: dict[int, float] = dict.fromkeys(CHANNEL_IDS, 0)
 
+    async def cog_load(self) -> None:
+        """Initialize channel timestamps."""
+        await self.bot.wait_until_guild_available()
+
         for channel_id in CHANNEL_IDS:
-            channel = bot.get_channel(channel_id)
+            channel = self.bot.get_channel(channel_id)
             if channel is None:
                 continue
 
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -22,7 +22,7 @@ class Tunnel(commands.Cog):
         await self.bot.wait_until_guild_available()
 
         for channel_id in CHANNEL_IDS:
-            channel = self.bot.get_channel(channel_id)
+            channel = await self.bot.fetch_channel(channel_id)
             if channel is None:
                 continue
 
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -26,9 +26,6 @@ class Tunnel(commands.Cog):
             if channel is None:
                 continue
 
-            if not isinstance(channel, discord.TextChannel):
-                raise AssertionError
-
             last_message = channel.last_message
             if last_message is None:
                 continue
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -1,5 +1,3 @@
-from operator import itemgetter
-
 import discord
 from discord.ext import commands
 from discord.ext.commands import BadArgument
@@ -87,7 +85,7 @@ class Tunnel(commands.Cog):
         """Gets least active off-topic channel."""
         return min(
             (channel for channel in self.channel_id_to_timestamp if channel != current_channel_id),
-            key=self.channel_id_to_timestamp.get
+            key=lambda c: self.channel_id_to_timestamp[c]
         )
 
 
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -26,7 +26,11 @@ class Tunnel(commands.Cog):
             if channel is None:
                 continue
 
-            last_message = channel.last_message
+            last_message_id = channel.last_message_id
+            if last_message_id is None:
+                continue
+
+            last_message = await channel.fetch_message(last_message_id)
             if last_message is None:
                 continue
 
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -38,7 +38,6 @@ class Tunnel(commands.Cog):
         self,
         ctx: commands.Context,
         destination_channel: discord.TextChannel | None,
-        source_channel: discord.TextChannel | None,
     ) -> None:
         """Creates a tunnel."""
         if destination_channel is None:
@@ -46,14 +45,7 @@ class Tunnel(commands.Cog):
             least_active_channel = await ctx.guild.fetch_channel(least_active_channel_id)
             destination_channel = least_active_channel
 
-        if source_channel is None:
-            if not isinstance(ctx.channel, discord.TextChannel):
-                raise BadArgument(
-                    f"The current channel of type '{ctx.channel.type}' is not a valid source channel "
-                    "and no explicit source channel was provided"
-                )
-
-            source_channel = ctx.channel
+        source_channel = ctx.channel
 
         if not source_channel.permissions_for(ctx.author).send_messages:
             raise BadArgument(f"You don't have permission to send messages in {source_channel.jump_url}")
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -19,8 +19,6 @@ class Tunnel(commands.Cog):
 
     async def cog_load(self) -> None:
         """Initialize channel timestamps."""
-        await self.bot.wait_until_guild_available()
-
         for channel_id in CHANNEL_IDS:
             channel = await self.bot.fetch_channel(channel_id)
             if channel is None:
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -42,15 +42,9 @@ class Tunnel(commands.Cog):
         source_channel: discord.TextChannel | None,
     ) -> None:
         """Creates a tunnel."""
-        if ctx.guild is None:
-            raise AssertionError
-
         if destination_channel is None:
             least_active_channel_id = self.get_least_active_channel_id(ctx.channel.id)
             least_active_channel = await ctx.guild.fetch_channel(least_active_channel_id)
-            if not isinstance(least_active_channel, discord.TextChannel):
-                raise AssertionError
-
             destination_channel = least_active_channel
 
         if source_channel is None:
@@ -61,9 +55,6 @@ class Tunnel(commands.Cog):
                 )
 
             source_channel = ctx.channel
-
-        if not isinstance(ctx.author, discord.Member):
-            raise AssertionError
 
         if not source_channel.permissions_for(ctx.author).send_messages:
             raise BadArgument(f"You don't have permission to send messages in {source_channel.jump_url}")
--- a/bot/exts/utils/tunnel.py
+++ b/bot/exts/utils/tunnel.py
@@ -1,0 +1,107 @@
+from operator import itemgetter
+
+import discord
+from discord.ext import commands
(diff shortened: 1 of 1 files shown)


Matiiss:
I do have a couple of questions though, first, so currently I have made it so that only `discord.TextChannel` are supported, which means you can't use this command inside those voice channel text channels or threads, but I suppose it'd be fine to allow its usage there, too.

Another thing is that if I change the command method signature to use `discord.abc.MessageableChannel`, then suddenly `!tunnel` is being interpreted as a "tag" and it's trying to find a match for it among the other tags, instead of running this command. I'd like some advice on how to make it behave like it does with `discord.TextChannel` or like make the tag search skip it or something.

Then also, if I use the `discord.TextChannel` annotation (and presumably it would be similar if I fixed it so that `discord.abc.MessageableChannel` would work), if I provide a wrong ID (same length, just changed some digit), it doesn't appear to complain at any level, it seems it just sets that argument to `None`, which, I'm not sure if that's the behavior that I'd like. That is, you either provide a correct argument for the channel or it uses the default behavior without letting you know that you may have messed up.

Should there be checks for whether the given channel is a DM channel or some such? Currently it just seems to treat it as incorrect channel ID (probably because it's not a `discord.TextChannel`), so it just sets it to `None`, but if switching to `discord.abc.MessageableChannel`, that should probably be explicitly checked?

Similarly, what to do about `ctx.guild`, should that throw some more specific error or should it realistically never be `None` (which is my assumption, hence the `AssertionError)? I used `raise AssertionError` mostly for typing purposes here (in places where it also should never be realistically raised anyway).

Is there a simple way to do overloads for this command? So that with one argument it's `!tunnel [destination_channel]`, but then with two it's `!tunnel [source_channel] [destination_channel]`?

Any shortcuts/convenience functions I should use for say permission checks from the existing code base?

Does this need unit or integration tests?

bot/exts/utils/tunnel.py
```diff
@@ -0,0 +1,107 @@
...
+                continue
+
+            if not isinstance(channel, discord.TextChannel):
+                raise AssertionError
+
+            last_message = channel.last_message
+            if last_message is None:
+                continue
+
+            self.channel_id_to_timestamp[channel_id] = last_message.created_at.timestamp()
```
jb3: This should be done in cog load listener, not in the cog init (see other cogs for prior art)

bot/exts/utils/tunnel.py
```diff
@@ -0,0 +1,107 @@
...
+    @commands.command()
+    async def tunnel(
+        self,
+        ctx: commands.Context,
+        destination_channel: discord.TextChannel | None,
+        source_channel: discord.TextChannel | None,
+    ) -> None:
+        """Creates a tunnel."""
+        if ctx.guild is None:
+            raise AssertionError
```
jb3: Use a guild-only check here instead, also maybe worth adding a cooldown decorator as we do to prevent abuse of this feature.

bot/exts/utils/tunnel.py
```diff
@@ -0,0 +1,107 @@
...
+        """Gets least active off-topic channel."""
+        channel_id, _ = min(
+            [
+                (channel_id, timestamp)
+                for channel_id, timestamp in self.channel_id_to_timestamp.items()
+                if channel_id != current_channel_id
+            ],
+            key=itemgetter(1),
+        )
+        return channel_id
```
jb3: Feels cleaner to just do this:

```suggestion
    def get_least_active_channel_id(self, current_channel_id: int) -> int:
        """Gets least active off-topic channel."""
        return min(
            (channel for channel in self.channel_id_to_timestamp if channel != current_channel_id),
            key=self.channel_id_to_timestamp.get
        )
```

bot/exts/utils/tunnel.py
```diff
@@ -0,0 +1,107 @@
...
+    ) -> None:
+        """Creates a tunnel."""
+        if ctx.guild is None:
+            raise AssertionError
+
+        if destination_channel is None:
+            least_active_channel_id = self.get_least_active_channel_id(ctx.channel.id)
+            least_active_channel = await ctx.guild.fetch_channel(least_active_channel_id)
+            if not isinstance(least_active_channel, discord.TextChannel):
+                raise AssertionError
```
jb3: I understand why you've got a few `AssertionError`s throughout your implementation here, but this is not a pattern we have through the rest of this project and I'm not sure how I feel about implementing it just for this cog.

We have found that trying to do precise typing is almost impossible with how many updates the upstream discord.py types get as well as when new features are released to the Discord API that suddenly then start breaking assertions.

I'd rather see any genuine failures handled but otherwise do not think we should have these in the implementation.

jb3 (changes requested):
Nice initial implementation, a few feedback comments. Will address your other comments in a second.

jb3:
> so currently I have made it so that only discord.TextChannel are supported, which means you can't use this command inside those voice channel text channels or threads, but I suppose it'd be fine to allow its usage there, too.

Sounds fine — with the necessary cooldown added as mentioned so users cannot spam this in places we have less visibility.

> Another thing is that if I change the command method signature to use discord.abc.MessageableChannel, then suddenly !tunnel is being interpreted as a "tag" and it's trying to find a match for it among the other tags, instead of running this command. I'd like some advice on how to make it behave like it does with discord.TextChannel or like make the tag search skip it or something.

This sounds weird, do you get the same if you try use `GuildChannel` instead?

> That is, you either provide a correct argument for the channel or it uses the default behavior without letting you know that you may have messed up.

I wonder if using a https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.TextChannelConverter might help?

> Should there be checks for whether the given channel is a DM channel or some such? Currently it just seems to treat it as incorrect channel ID (probably because it's not a discord.TextChannel), so it just sets it to None, but if switching to discord.abc.MessageableChannel, that should probably be explicitly checked?

You can use a guild-only decorator to enforce this.

> Similarly, what to do about ctx.guild, should that throw some more specific error or should it realistically never be None (which is my assumption, hence the AssertionError)? I used raise AssertionError` mostly for typing purposes here (in places where it also should never be realistically raised anyway).

Same as above.

> Is there a simple way to do overloads for this command? So that with one argument it's !tunnel [destination_channel], but then with two it's !tunnel [source_channel] [destination_channel]?

Not easily without doing most of the argument handling yourself.

I was going to suggest that for this initial implementation though we ditch the source channel support. Especially since the user who did the tunneling isn't mentioned in the tunneling messages it seems ripe for abuse (e.g. someone spamming in bot commands with another source & dest).

For now, let's keep things simple and just implement an optional destination channel.

> Does this need unit or integration tests?

You can if you want to (and if you remove the source stuff it should be easier), we do have some utilities for mocking Discord components here and it's not too hard to add new bits.

jb3:
Also, on the last point about abuse: can we include the user who tunneled on both the source and destination message (username is fine, doesn't need to be a mention), so that we can easily see any problems without having to look through message logs.

bot/exts/utils/tunnel.py
```diff
@@ -0,0 +1,107 @@
...
+
+        for channel_id in CHANNEL_IDS:
+            channel = bot.get_channel(channel_id)
+            if channel is None:
+                continue
+
+            if not isinstance(channel, discord.TextChannel):
+                raise AssertionError
+
+            last_message = channel.last_message
```
wookie184: I don't think this will work. From what I can see the Discord API returns a `last_message_id` along with the channel, but discord.py requires the message to be in the cache for `last_message` to be populated. If the bot has just loaded, the message wont be in cache.

The easiest fix is probably to use  `channel.last_message_id` directly with https://discordpy.readthedocs.io/en/latest/api.html#discord.utils.snowflake_time

bot/exts/utils/tunnel.py
wookie184: Also, just to add, `AssertionError` is [documented as](https://docs.python.org/3/library/exceptions.html#AssertionError) "Raised when an assert statement fails.". It's not an exception I would raise manually.

bot/exts/utils/tunnel.py
```diff
@@ -0,0 +1,107 @@
...
+        if not source_channel.permissions_for(ctx.author).send_messages:
+            raise BadArgument(f"You don't have permission to send messages in {source_channel.jump_url}")
+        if not destination_channel.permissions_for(ctx.author).send_messages:
+            raise BadArgument(f"You don't have permission to send messages in {destination_channel.jump_url}")
+
+        if source_channel.id == destination_channel.id:
+            raise BadArgument("Source and destination channels cannot be the same")
+
+        source_message_template = "➡️ Conversation tunneled to {location}"
+        destination_message_template = "↩️ Conversation tunneled from {location}"
```
wookie184: Just a thought, would it be clearer to say "Conversation continued at {location}"/"Conversation continued from {location}"? It might not be immediately clear what "tunneled" means.

wookie184 (changes requested):
Thanks for working on this :)

Potential simplification:

1. Get the channels in the `cog_load` and store them in in a list e.g. `self.channels`
2. replace `get_least_active_channel_id` with `get_least_active_channel` that looks something like
    ```
    return min(self.channels, key=lambda c: thattimeutilthing(c.last_message_id)
    ```
3. Then the `channel_id_to_timestamp` logic + `on_message` aren't needed at all, and it's a bit simpler as you're passing the channel around instead of the channel id (don't need to fetch it).

bot/exts/utils/tunnel.py
Matiiss: @wookie184 I would have used plain `assert`s but ruff was complaining that they get removed in optimized builds (IMO, for purposes of type narrowing it should be fine still, but oh well), so I emulated that via a `raise AssertionError` instead. I suppose I could instead raise some other exception.

But yeah, since it's here done purely for the purposes of type narrowing for typing purposes, I'll just get rid of these emulated asserts entirely

bot/exts/utils/tunnel.py
Matiiss: Presumably you mean the `discord.ext.commands.guild_only` decorator?

Matiiss:
> This sounds weird, do you get the same if you try use GuildChannel instead?

Yes

> I wonder if using a https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.TextChannelConverter might help?

Yes, it does, it seems I then just annotate the destination channel as `str` and any argument would be passed directly to it without some weird implicit magic going on. That is nice, currently the command is set up to expect a `discord.TextChannel` since that is the only converter used here. I think it's mostly fine, if you give it a thread id, like a help thread id, it will throw a `BadArgument` and that will be the response from the bot as well, so it's clear. It does mean that essentially you can't tunnel into a thread, but I don't think that's super necessary at the moment? You can tunnel out of one though.

Now, since the text uses the word "continued", should the command have an alias called "continue" as well? I quite like "tunnel", I think it's fun, but if the text now uses "continued", it might be more intuitive to have an option to call it via such an alias as well? I'm not sure, tbh, if the command is called "tunnel", then maybe it's not necessary to use the word "continued" in the text and we can use the word "tunneled" in the text as well, since it shouldn't be too confusing on what it means given the invocation of the command is probably gonna be pretty much nearby and also one could infer from the fact that there are two messages that link to each other in a "tunnel-like fashion". Not sure.

I saw there was a decorator for cooldown where some roles can be excluded from that (`bot.utils.checks.cooldown_with_role_bypass`), but I'm not sure that's super necessary here, besides that function had a comment saying that it _might_ break at some point. So the cooldown is there just mainly for abuse prevention I suppose, I don't imagine this command would actually get used extensively in practice.

Otherwise, I think this is ready for another review, I might eventually maybe add those integration tests as well, but who knows, that might take another month to get there :sweat_smile:

bot/exts/utils/tunnel.py
```diff
@@ -0,0 +1,74 @@
...
+        if destination_channel_raw is None:
+            least_active_channel = self.get_least_active_channel(ctx.channel)
+            destination_channel = least_active_channel
+        else:
+            destination_channel = await commands.TextChannelConverter().convert(ctx, destination_channel_raw)
+
+        source_channel = ctx.channel
+
+        if not source_channel.permissions_for(ctx.author).send_messages:
+            raise BadArgument(f"You don't have permission to send messages in {source_channel.jump_url}")
```
Sacul0457, line 46: Can this ever happen? Doesn't the user have to run the command in the source channel to even execute it?

Matiiss: I don't think it can, it is effectively dead code, I think I sort of kept it for some future compatibility when the source channel can also be specified, but I guess it can then be added back at that point.

PR #3517, opened 2026-06-06
Fix Codeblock False Positives

### LINKED ISSUE #3515: False positive faulty code block detection
This seems related to #3514. The message that triggered it:
```
I'd prefer to configure OIDC where possible so every token is short lived.

If that's not possible I'd store it in AWS secret manager or SSM parameters and fetch it on program setup.


If that's not possible I'd fallback to env:/
```

https://discord.com/channels/267624335836053506/267624335836053506/1512843612990214235

SamuelRoettgermann:
Fixes https://github.com/python-discord/bot/issues/3515 

Fixes the described issue and adds some unit tests to cover common cases.

Also updates the bad_ticks error message to work with the new, more flexible, variable ticks approach.

### DIFF
--- a/bot/exts/info/codeblock/_parsing.py
+++ b/bot/exts/info/codeblock/_parsing.py
@@ -36,7 +36,7 @@ _RE_IPYTHON_REPL = re.compile(r"^((In|Out) \[\d+\]: |\s*\.{3,}: ?)")
 _RE_CODE_BLOCK = re.compile(
     fr"""
     (?:^|                             # the ticks need to start at the front of a line to be recognized
-    \s)                               # or need to have a preceeding whitespace (to avoid detection of words like I'll).
+    \s)                               # or need to have a preceding whitespace (to avoid detection of words like I'll).
     (?P<ticks>
         (?P<tick>[{''.join(_TICKS)}]) # Put all ticks into a character class within a group.
         \2*                           # Match previous group up to N more times to ensure the same char.
--- a/bot/exts/info/codeblock/_parsing.py
+++ b/bot/exts/info/codeblock/_parsing.py
@@ -182,7 +182,7 @@ def parse_bad_language(content: str) -> BadLanguage | None:
     )
 
 
-def _get_leading_spaces(content: str) -> int:
+def _get_leading_spaces(content: str) -> int | None:
     """Return the number of spaces at the start of the first line in `content`."""
     leading_spaces = 0
     for char in content:
@@ -190,6 +190,7 @@ def _get_leading_spaces(content: str) -> int:
             leading_spaces += 1
         else:
             return leading_spaces
+
     return None
 
 
--- a/bot/exts/info/codeblock/_parsing.py
+++ b/bot/exts/info/codeblock/_parsing.py
@@ -35,11 +35,11 @@ _RE_CODE_BLOCK = re.compile(
     fr"""
     (?P<ticks>
         (?P<tick>[{''.join(_TICKS)}]) # Put all ticks into a character class within a group.
-        \2{{2}}                       # Match previous group 2 more times to ensure the same char.
+        \2*                           # Match previous group up to N more times to ensure the same char.
     )
-    (?P<lang>[A-Za-z0-9\+\-\.]+\n)?   # Optionally match a language specifier followed by a newline.
+    (?P<lang>[A-Za-z0-9+\-.]+\s)?     # Optionally match a language specifier followed by a whitespace.
     (?P<code>.+?)                     # Match the actual code within the block.
-    \1                                # Match the same 3 ticks used at the start of the block.
+    \1                                # Match the same N ticks used at the start of the block.
     """,
     re.DOTALL | re.VERBOSE
 )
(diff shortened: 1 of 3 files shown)


ShaharNaveh (dismissed):
LGTM!

bot/exts/info/codeblock/_parsing.py
```diff
@@ -33,6 +35,8 @@
 
 _RE_CODE_BLOCK = re.compile(
     fr"""
+    (?:^|                             # the ticks need to start at the front of a line to be recognized
+    \s)                               # or need to have a preceeding whitespace (to avoid detection of words like I'll).
```
jb3:
```suggestion
    \s)                               # or need to have a preceding whitespace (to avoid detection of words like I'll).
```

jb3 (changes requested):
Minor typo, else LGTM

jb3 (approved):

bot/exts/info/codeblock/_parsing.py
```diff
@@ -43,6 +47,7 @@
     """,
     re.DOTALL | re.VERBOSE
 )
+_RE_CODE_BLOCK_REGEX = regex.compile(_RE_CODE_BLOCK.pattern, regex.DOTALL | regex.VERBOSE)
```
MarkKoz: Am I missing something here or is this just a copy of `_RE_CODE_BLOCK`?

bot/exts/info/codeblock/_parsing.py
```diff
@@ -92,11 +98,18 @@ def find_code_blocks(message: str) -> Sequence[CodeBlock] | None:
...
             return None
 
-        if has_lines(groups["code"], constants.CodeBlock.minimum_lines):
-            code_block = CodeBlock(groups["code"], language, groups["tick"])
+        if not has_lines(groups["code"], constants.CodeBlock.minimum_lines):
+            log.trace("Skipped a code block shorter than 4 lines.")
+            continue
+
+        if (groups["tick"] == BACKTICK
+                or (language in PY_LANG_CODES and is_python_code(groups["code"]))
```
MarkKoz: We also call `is_python_code` in `_get_no_lang_message`. An optimisation could be to save the result in `CodeBlock`.

MarkKoz: Do we have test coverage for the case where the language matches but the content isn't valid Python code? Is that `test_should_not_recognize_normal_double_quotes_python_text`?

tests/bot/exts/info/codeblock/test_parsing.py
```diff
@@ -0,0 +1,153 @@
+import unittest
+
+from bot.exts.info.codeblock import _parsing as parsing
+
+
+class FindFaultyCodeblocksTest(unittest.TestCase):
```
MarkKoz, line 6: Thank you for adding tests!

bot/exts/info/codeblock/_parsing.py
```diff
@@ -70,9 +76,9 @@ class BadLanguage(NamedTuple):
     has_terminal_newline: bool
 
 
-def find_code_blocks(message: str) -> Sequence[CodeBlock] | None:
+def find_faulty_code_blocks(message: str) -> Sequence[CodeBlock] | None:
```
MarkKoz, line 81: Good rename

MarkKoz (changes requested):

bot/exts/info/codeblock/_parsing.py
SamuelRoettgermann: Yes and no.

So, it is a copy, but it uses the `regex` module, rather than the `re` module, because `regex.finditer` allows to pass an additional `overlapped` parameter, which I need in the `find_faulty_code_blocks` function.

The reason why I didn't omit the `_RE_CODE_BLOCK` variable entirely just comes down to code highlighting. Let me show you with this image: 
image

As you can see, at least in PyCharm, there is no default regex-highlighting when using the `regex.compile` function, but there is when using the `re.compile` function.

bot/exts/info/codeblock/_parsing.py
SamuelRoettgermann: That indeed sounds like a good idea

SamuelRoettgermann: Exactly. If you want, I can also add a test case for the same thing with single quotes, but I don't think that's necessary.

Maybe the names for the tests aren't all super good, if you have a naming suggestion/improvement, please do share.

SamuelRoettgermann: iirc we also call it in smth like `_get_bad_lang_message` or something alike, but I'll need to properly check that again, that just what I recall from memory rn

MarkKoz: I think the coverage and names are fine, thanks

bot/exts/info/codeblock/_parsing.py
MarkKoz: Oh yeah I missed that it's different library. I don't find syntax highlighting to be a compelling reason to create this confusion, but don't feel strongly.

SamuelRoettgermann: I added a comment explaining the decision to avoid further confusion

MarkKoz (approved):

PR #3519, opened 2026-06-12
New UniqueFilter for images associated with compromised accounts.

swfarnsworth:
Currently, the perceptual hashes of known images are hard-coded, but this is a temporary solution.

image

In response to only the second image:

image

Upon merging this change, someone with requisite permissions must run `!filter add unique image` and then configure the desired behavior for positive identifications from this filter.

### DIFF
--- a/bot/exts/filtering/_filters/unique/image.py
+++ b/bot/exts/filtering/_filters/unique/image.py
@@ -8,6 +8,7 @@ from bot.log import get_logger
 
 log = get_logger(__name__)
 
+# Maximum perceptual hash difference for positive predictions
 _THRESHOLD = 4
 # Maximum number of seconds to wait for Rhodium API
 _TIMEOUT = 0.5
@@ -22,17 +23,9 @@ _KNOWN_IMAGE_HASHES = [
 ]
 
 
-def _hamming_distance(a: int, b: int) -> int:
-    return bin(a ^ b).count("1")
-
-
-def _is_similar(hash_a: int, hash_b: int, max_distance: int = 3) -> bool:
-    return _hamming_distance(hash_a, hash_b) <= max_distance
-
-
 def _is_match(image_hash: int) -> bool:
     return any(
-        _is_similar(image_hash, candidate_hash, max_distance=_THRESHOLD)
+        int.bit_count(image_hash ^ candidate_hash) <= _THRESHOLD
         for candidate_hash in _KNOWN_IMAGE_HASHES
     )
 
--- a/bot/exts/filtering/_filters/unique/image.py
+++ b/bot/exts/filtering/_filters/unique/image.py
@@ -1,8 +1,9 @@
 import asyncio
 import io
 
+import discord
 import imagehash
-from PIL import Image
+from PIL import Image, UnidentifiedImageError
 
 from bot.exts.filtering._filter_context import Event, FilterContext
 from bot.exts.filtering._filters.filter import UniqueFilter
@@ -36,8 +37,16 @@ class ImageFilter(UniqueFilter):
             ):
                 continue
 
-            image_bytes = io.BytesIO(await attachment.read())
-            image = await asyncio.to_thread(Image.open, image_bytes)
+            try:
+                image_bytes = io.BytesIO(await attachment.read())
+            except (discord.Forbidden, discord.DiscordServerError, discord.NotFound):
+                continue
+
+            try:
+                image = await asyncio.to_thread(Image.open, image_bytes)
+            except UnidentifiedImageError:
+                continue
+
             if await _image_is_match(image):
                 return True
 
--- a/bot/exts/filtering/_filters/unique/image.py
+++ b/bot/exts/filtering/_filters/unique/image.py
@@ -61,7 +61,7 @@ class ImageFilter(UniqueFilter):
             if (
                 attachment.content_type is None
                 or not attachment.content_type.startswith("image")
-                or attachment.size > 3e7  # 30 mb
+                or attachment.size > 5e6  # 5mb
             ):
                 continue
 
--- a/bot/exts/filtering/_filters/unique/image.py
+++ b/bot/exts/filtering/_filters/unique/image.py
@@ -1,5 +1,7 @@
 import os
 
+import aiohttp
+
 import bot
 from bot import constants
 from bot.exts.filtering._filter_context import Event, FilterContext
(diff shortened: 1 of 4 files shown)


bot/exts/filtering/_filters/unique/image.py
```diff
@@ -0,0 +1,38 @@
...
+class ImageFilter(UniqueFilter):
+    """Filter messages that contain an image attachment whose perceptual hash matches images associated with scams."""
+
+    name = "image"
+    events = (Event.MESSAGE, )
+
+    async def triggered_on(self, ctx: FilterContext) -> bool:
+        """Return whether the message has an attached image that is known to be posted by compromised accounts."""
+        for attachment in ctx.attachments:
+            if not attachment.content_type.startswith("image"):
```
wookie184: attachment.content_type can be none https://discordpy.readthedocs.io/en/latest/api.html?highlight=content_type#discord.Attachment.content_type

bot/exts/filtering/_filters/unique/image.py
```diff
@@ -0,0 +1,38 @@
...
+                continue
+            image = Image.open(io.BytesIO(await attachment.read()))
+            if _image_is_match(image):
+                return True
```
wookie184, line 90: We should use `asyncio.to_thread` to avoid blocking the event loop.

I'm not sure what errors `Image.open`/`imagehash.phash` can raise but they should be caught (and probably ignored, e.g. if it isn't an image or is a corrupted image)

wookie184 (changes requested):
Is it a requirement that this is a perceptual hash, or would a normal file hash be sufficient for our current use case.

The reasons I suggest this is:
- Currently this brings in some chunky dependencies (numpy, scikit) - more than doubling the .venv size. Not a massive issue, but nice to avoid if possible.
- Image processing is a bit slow. At the very least we would need to use `asyncio.to_thread` to avoid blocking the event loop (though would potentially make sense to use that for a file hash too).
- Image processing is complex and creates a larger attack surface for malicious files.  See https://pillow.readthedocs.io/en/latest/handbook/security.html 

Using a file hash like md5 is simple, fast, and can be used on any type of file.

bot/exts/filtering/_filters/unique/image.py
swfarnsworth: Thank you for pointing this out. I'll push a commit containing the fix shortly.

bot/exts/filtering/_filters/unique/image.py
swfarnsworth: I've applied `asyncio.to_thread` to `Image.open` and `imagehash.phash`.

I'll also handle exceptions for `Image.open` and `attachment.read`. I don't see any exceptions that `imagehash.phash` could expectedly raise if the correct types are passed.

swfarnsworth:
@wookie184 thank you for your review. I think I've addressed all your concerns.

As we discussed on Discord (restating here for others), the images used in the ongoing attack differ slightly each time. 

It's disappointing that the new dependencies double the size of the venv. We could instead have this as a microservice, but that might not be worth the tradeoffs. In either case, I'm fine with ultimately removing this functionality and its dependencies when the ongoing attack ends.

swfarnsworth:
Oh, I meant to also say that I don't have any logging statements in either of the except blocks, though I considered it. Let me know if you want me to do that. I think "info" would be the correct logging level here?

swfarnsworth:
@jb3 the two most recent commits address concerns that you raised in messages in Discord.

The filter will only consider attachments less than 30mb and I've added comments describing the image for each perceptual hash.

There were two instances of this attack while I typed this message 🫠

wookie184:
It's not so much the file size that matters, but the image resolution. You can craft a very small image that acts as a decompression bomb.

See the warning on [Image.open](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.open). We should:
- Turn the `Image.DecompressionBombWarning` into an error.
- Lower [MAX_IMAGE_PIXELS](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.MAX_IMAGE_PIXELS)

bot/exts/filtering/_filters/unique/image.py
```diff
@@ -1,35 +1,51 @@
...
-    """Return whether the one image matches any of those known to be posted by compromised accounts."""
-    incoming_image_hash = await asyncio.to_thread(imagehash.phash, image)
-    has_match = any(
-        incoming_image_hash - known_image_hash < _THRESHOLD
-        for known_image_hash in _KNOWN_IMAGE_HASHES
+def _hamming_distance(a: int, b: int) -> int:
+    return bin(a ^ b).count("1")
+
+
+def _is_similar(hash_a: int, hash_b: int, max_distance: int = 3) -> bool:
```
wookie184: It's a bit confusing to have a default value that is overridden when there is only one use of the function, and it passes a different value.

bot/exts/filtering/_filters/unique/image.py
```diff
@@ -1,35 +1,51 @@
...
 
 
-async def _image_is_match(image: Image.Image) -> bool:
-    """Return whether the one image matches any of those known to be posted by compromised accounts."""
-    incoming_image_hash = await asyncio.to_thread(imagehash.phash, image)
-    has_match = any(
-        incoming_image_hash - known_image_hash < _THRESHOLD
-        for known_image_hash in _KNOWN_IMAGE_HASHES
+def _hamming_distance(a: int, b: int) -> int:
+    return bin(a ^ b).count("1")
```
wookie184: Use `int.bit_count`

bot/exts/filtering/_filters/unique/image.py
```diff
@@ -0,0 +1,72 @@
...
+
+def _is_match(image_hash: int) -> bool:
+    return any(
+        _is_similar(image_hash, candidate_hash, max_distance=_THRESHOLD)
+        for candidate_hash in _KNOWN_IMAGE_HASHES
+    )
+
+async def _get_hash(image_url: str) -> int:
+    async with (
+        aiohttp.ClientSession() as session,
```
wookie184: Creating a new session for every request isn't ideal. You can get the bot's session with `from bot import instance`, `instance.http_session`.

bot/exts/filtering/_filters/unique/image.py
```diff
@@ -0,0 +1,72 @@
...
+        session.post(
+            url=os.getenv("RHODIUM_HOST_URL"),
+            headers={"Authorization": f"Bearer {os.getenv('RHODIUM_AUTH_TOKEN')}"},
```
wookie184: See https://github.com/python-discord/bot/blob/main/bot/constants.py for how we handle constants.

wookie184 (changes requested):
We should be careful with error handling here. My understanding is that any errors raised here will cause the filter system to fail for the sent message, potentially blocking other filters from triggering.

Network errors need to be handled+logged according to the type of error, and not reraised.

We should also consider performance here, as other filters will be blocked while this one is executing. I would suggest:
- Consider running requests for multiple attachments in parallel.
- Add a timeout in the case of the worker being slow or other network issues.

wookie184 (approved):
lgtm

ChrisLovering (approved):

ChrisLovering (approved):

PR #3525, opened 2026-06-24
Add additional hashes to ImageFilter

swfarnsworth:
I identified several images whose hashes should have been included in the prior PR. While these attacks often involve a few images, this PR includes only hashes for images that contain the malicious instructions, as all such attacks must contain at least that image.


swfarnsworth:
I didn't realize the prior PR had been squashed--I'll open a new one that accounts for this. Apologies.

PR #3527, opened 2026-06-24
Fix issue with apparent codeblock within other codeblock

### LINKED ISSUE #3522: Code block recognition issue.
The Python bot has an issue with code blocks that are formatted with ` ``` ` but not ` ```py`. It replies with a [message](https://discord.com/channels/267624335836053506/267624335836053506/1518362996844527627) that says:

> You are using the wrong character instead of backticks.

It should say something to the effect of "While you are using backticks you are not specifying the language. Please use ` ```py` for Python code."

brass75:
If there is a triple quoted string inside a `codeblock` that appears invalid the current implementation will trigger the the "Back ticks message". This is confusing to the user and generally wrong because that happens to be correct. What needs to happen is that the we should isolate the `codeblock`s and make sure that the violating `codeblock` is notcontained within a `codeblock`.
This change handles that scenario by checking all of the invalid `codeblock`s to generate the correct message based on the invalid `codeblock`. If a message contains multiple invalid `codeblock`s and there is at least one that is using invalid ticks the invalid ticks message will be shown. If there is no `codeblock` with invalid ticks the missing language and no language logic will be followed without change.

Fixes #3522


brass75:
Is this going to be looked at?

jb3 (changes requested):
Thanks for the PR.

This seems to now have the correction message not fire at all for the message that was initially sent (the one mentioned in #3522).

The first message below is with the bot running on `main`, the second is running on your PR branch.

Image

We still want the message to fire here, we just want it to fire with the "add `py`" bit instead of the wrong backticks, I think? (if the message fired before it still should fire, just with the right wording)

Link to the test message I am using (30 day expiry): https://paste.pythondiscord.com/TNUA

brass75:
It doesn't fire here because there's a leading space before the first character so it's not recognized as Python. If the original message is changed to remove that space it does fire as expected. 

Essentially what was happening was:

- Code block without a language specification. 
- Inability to determine the language as Python because the AST parser failed because of the leading space. 
- Detection of the `"""` block inside the code block triggering the invalid ticks message. 

What could be done to try to get the correct message out is to try to parse again removing a single leading space and see if it then parses as Python but, IMO, this is a somewhat exceptional case.

jb3 (approved):
Seems all good then, and agree that's likely an edge case that we can watch out for. Thanks again!

PR #3528, opened 2026-06-25
Fix error in botstrap when attempting to upgrade to community server

wookie184:
Previously it was sending the entire guild info payload in the patch. Now it only sends the fields we want to change.

Also send a helpful error message if the bot doesn't have admin permisssions on the server.

Testing:
- Confirmed the friendly message is sent without admininstrator permission.
- Confirmed bootstrap now works when reinvited with administrator permission.

### DIFF
--- a/botstrap.py
+++ b/botstrap.py
@@ -82,10 +82,6 @@ class SilencedDict[T](dict[str, T]):
             sys.exit(-1)
 
 
-class BotstrapError(Exception):
-    """Raised when an error occurs during the botstrap process."""
-
-
 class DiscordClient(Client):
     """An HTTP client to communicate with Discord's APIs."""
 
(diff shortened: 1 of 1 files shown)


botstrap.py
```diff
@@ -330,20 +331,32 @@ def check_guild_membership(self) -> None:
...
-            announcements_channel_id=announcements_channel_id,
-        )
+        try:
+            return self.client.upgrade_server_to_community_if_necessary(
+                rules_channel_id=rules_channel_id,
+                announcements_channel_id=announcements_channel_id,
+            )
+        except HTTPStatusError as e:
+            if e.response.status_code == 403:
+                log.error("Unable to upgrade to community guild as he bot does not have Administrator permissions")
```
brass75, line 351:
```suggestion
                log.error("Unable to upgrade to community guild as the bot does not have Administrator permissions.")
```

botstrap.py
```diff
@@ -330,20 +331,32 @@ def check_guild_membership(self) -> None:
...
+                log.info(
+                    "Please re-invite with the following URL and rerun this script: "
```
brass75, line 353:
```suggestion
                    "Please re-invite with the following URL and run this script again: "
```
This wording might be a bit more understandable to non-native English speakers.

botstrap.py
```diff
@@ -335,20 +331,32 @@ def check_guild_membership(self) -> None:
...
-            announcements_channel_id=announcements_channel_id,
-        )
+        try:
+            return self.client.upgrade_server_to_community_if_necessary(
+                rules_channel_id=rules_channel_id,
+                announcements_channel_id=announcements_channel_id,
+            )
+        except HTTPStatusError as e:
+            if e.response.status_code == 403:
+                log.error("Unable to upgrade to community guild as he bot does not have Administrator permissions")
```
shtlrs, line 351:
```suggestion
                log.error("Unable to upgrade to community guild as the bot does not have Administrator permissions")
```

shtlrs (approved):
Thanks

PR #3530, opened 2026-06-28
Tighter mention filtering in auto-ban mod-alerts

jb3:
For filters that resolve to `infraction_type=BAN`, the resolver no longer removes the entire mentions action. Instead, it now strips only:

 - the Moderators role ping
 - `@here`
 - `@everyone`

All other configured guild_pings/dm_pings are preserved and still included in mod alerts (e.g. if a user has explicitly added themselves to a specific filter).

Also added a test for _clean_ban_mentions to verify moderator/here/everyone mentions get removed and other mentions remain.

### DIFF
--- a/bot/exts/filtering/filtering.py
+++ b/bot/exts/filtering/filtering.py
@@ -47,7 +47,7 @@ from bot.exts.filtering._ui.ui import (
     build_mod_alert,
     format_response_error,
 )
-from bot.exts.filtering._utils import past_tense, repr_equals, starting_value, to_serializable
+from bot.exts.filtering._utils import past_tense, repr_equals, resolve_mention, starting_value, to_serializable
 from bot.exts.moderation.infraction.infractions import COMP_BAN_DURATION, COMP_BAN_REASON
 from bot.exts.utils.snekbox._io import FileAttachment
 from bot.log import get_logger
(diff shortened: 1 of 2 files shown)


bot/exts/filtering/filtering.py
```diff
@@ -66,6 +66,21 @@
...
+    cleaned_mentions = set()
+
+    for mention in mentions:
+        if mention.casefold() == "moderators":
+            continue
+        if resolve_mention(mention) in blocked_mentions:
+            continue
+        cleaned_mentions.add(mention)
+
+    return cleaned_mentions
```
swfarnsworth, line 81: I believe this whole function could be a set comprehension, but I don't think it's necessary to change it.

swfarnsworth:
Desired behavior confirmed; screenshot shows two evocations of the filter before and after adding BAN as the infraction response.

image

swfarnsworth:
image

Confirmed also for `@here`, for whatever that's worth.

swfarnsworth (approved):

swfarnsworth:
Thank you @jb3 for such a quick turnaround on this QoL improvement!

bot/exts/filtering/filtering.py
jb3: It most probably could be — but I think for ease of unit testing & the fact there's two conditions it's proooobably cleaner this way, so will leave for now (if we want to unit test it then it'd likely have to be a function anyway, at which point the benefit of doing a set comprehension is slightly lost). Will leave for now as you say :+1:

shtlrs (approved):
Dear Joe.

That's it, no monologue.

PR #3532, opened 2026-06-29
Add Image Hash filter list

jb3:
This PR migrates image scam detection from the hardcoded unique image filter to a dynamic `image_hash` filter list.

- Added a new image_hash filter type and image_hash deny list flow in filtering.
- Removed the unique image filter implementation
- Added shared Rhodium hash helpers and updated hash parsing to use Rhodium's hex response key.
- Added a moderator utility command: `!filter imagehash` (alias: `!filter hash`) to hash an attached image and return both hex and i64 hash
- Updated image hash alert formatting so #mod-alert output always includes a description (with fallback text if missing)

A manual migration of the currently known image hashes is required after deploy. Hashes must be added as image_hash deny filters (this time using hex values instead of the integer hashes).


swfarnsworth:
I get expected behavior when running `!filter hash` but can't find a way to add a new image hash filter.

image

swfarnsworth:
It appears that one needs to run `!filterlist add deny image_hash` in a server when running this functionality for the first time. One must also ensure that the filter list itself is set to enabled (it is not by default), or none of the filters in the list will work, even if they themselves are activated.

Separately from the goal of this PR, I may have found a bug in `!filterlist edit` command. `!filterlist edit noui image_hash enabled=True` causes `RuntimeError: dictionary keys changed during iteration`. I'll investigate and potentially open another PR.

Standby while I continue to test this functionality.

swfarnsworth:
I've confirmed that the filter list works, though I wonder if it would be possible for the mod alert to tell us what the similarity score was between the posted image and the match. What do you think, @jb3?

image

ChrisLovering (commented):
thanks

PR #3535, opened 2026-07-04
Clean improvements

mbaruh:


bot/utils/async_utils.py
```diff
@@ -49,6 +49,34 @@ async def wait(self) -> None:
...
 
+class AsyncExecutor:
+    """Execute coroutines with a pool limit for concurrent execution."""
+
+    def __init__(self, *, limit: int):
+        self._semaphore = asyncio.Semaphore(limit)
+        self._running_tasks = set()
+
+    def submit[T](self, coro: Awaitable[T]) -> asyncio.Task[T]:
+        """Wraps the coroutine with _semaphore logic, schedules it on the event loop, and ensures cleanup."""
```
MarkKoz: Nit: Use imperative mood for docs "Wrap the coroutine" rather than "Wraps the coroutine".

Also, `_semaphore` is an internal detail so it shouldn't be directly referenced in the docstring.

bot/utils/async_utils.py
```diff
@@ -49,6 +49,34 @@ async def wait(self) -> None:
...
+        return task
+
+    async def execute[T](self, coro: Awaitable[T]) -> T:
+        """Executes the coroutine once there is an available slot in the _semaphore."""
+        async with self._semaphore:
+            return await coro
+
+    async def gather(self, return_exceptions: bool = False) -> list[Any]:
+        """Waits for all submitted coroutines to finish execution."""
+        if self._running_tasks:
```
MarkKoz: Nit: I think this check is redundant

bot/utils/async_utils.py
```diff
@@ -49,6 +49,34 @@ async def wait(self) -> None:
         await self._event.wait()
 
 
+class AsyncExecutor:
```
MarkKoz, line 52: Should there be a public interface for cancelling the tasks? Might be needed if the bot is shutting down.

bot/exts/moderation/clean.py
```diff
@@ -217,57 +232,84 @@ async def _delete_invocation(self, ctx: Context) -> None:
...
                 log.info("Tried to delete invocation message, but it was already deleted.")
 
-    def _use_cache(self, limit: datetime) -> bool:
-        """Tell whether all messages to be cleaned can be found in the cache."""
-        return self.bot.cached_messages[0].created_at <= limit
+    def _earliest_cache_datetime(self) -> datetime:
+        """Return the datetime of the earliest message cached, or now if the cache is empty."""
+        return self.bot.cached_messages[0].created_at if self.bot.cached_messages else arrow.utcnow().datetime
+
+    def _use_cache(self, most_recent_limit: datetime | None) -> bool:
```
MarkKoz, line 239: Nit: Inconsistent terminology in this file. There's earliest/latest, lower/upper, first/second, and least/most recent.

bot/exts/moderation/clean.py
```diff
@@ -217,57 +232,84 @@ async def _delete_invocation(self, ctx: Context) -> None:
...
-        channels: set[TextChannel],
+        channels: list[TextChannel],
         to_delete: Predicate,
         lower_limit: datetime
-    ) -> tuple[defaultdict[TextChannel, list], list[int]]:
+    ) -> dict[TextChannel, list[Message]]:
         """Helper function for getting messages from the cache."""
-        message_mappings = defaultdict(list)
-        message_ids = []
+        message_mappings = {channel: [] for channel in channels}
```
MarkKoz, line 254: Why not leave it as a defaultdict?

bot/exts/moderation/clean.py
```diff
@@ -217,57 +232,84 @@ async def _delete_invocation(self, ctx: Context) -> None:
...
 
-        return message_mappings, message_ids
+        return message_mappings
 
-    async def _get_messages_from_channels(
+    async def _get_messages_from_channels_and_delete(
         self,
-        channels: Iterable[TextChannel],
+        executor: AsyncExecutor,
+        *,
```
MarkKoz, line 269: Nit: why make them keyword-only?

bot/exts/moderation/clean.py
```diff
@@ -217,57 +232,84 @@ async def _delete_invocation(self, ctx: Context) -> None:
...
+                    return deleted, {}
 
                 if to_delete(message):
-                    message_mappings[message.channel].append(message)
+                    if self.is_older_than_14d(message):
+                        old_messages[channel].append(message)
+                        continue
+
+                    messages_to_delete.append(message)
                     message_ids.append(message.id)
```
MarkKoz: Nit: message IDs can be derived from `messages_to_delete`. I think it'd be cleaner to not maintain parallel lists, and that it wouldn't be a significant change in performance.

bot/exts/moderation/clean.py
```diff
@@ -281,64 +323,68 @@ def is_older_than_14d(message: Message) -> bool:
...
+                await asyncio.sleep(0)  # Avoid hogging the event loop, and allow the clean to be cancelled.
                 if not self.cleaning:
                     # Means that the cleaning was canceled
-                    return deleted
+                    return deleted, {}
 
                 if self.is_older_than_14d(message):
                     # Further messages are too old to be deleted in bulk
-                    delete_old = True
+                    old_messages[channel] = messages[current_index:]
```
MarkKoz, line 362: Should add to the docstring that this function expects messages to be sorted by timestamp.

bot/exts/moderation/clean.py
```diff
@@ -416,32 +462,45 @@ async def _clean_messages(
...
+            for channel, messages in api_old_messages.items():
+                old_messages[channel].extend(messages)
+
+        await executor.gather(return_exceptions=True)  # Ignore NotFound errors etc.
 
         if not self.cleaning:
             # Means that the cleaning was canceled
             return None
+        if old_messages:
+            old_deleted = await self._delete_messages_individually(old_messages)
```
MarkKoz, line 499: Why do old messages need to be deleted individually?

MarkKoz (changes requested):

bot/exts/moderation/clean.py
mbaruh: To preserve channel order

bot/exts/moderation/clean.py
mbaruh: API limitation

bot/exts/moderation/clean.py
mbaruh: I don't like having this many distinct positional arguments, and the function was being called with kw already anyway

bot/utils/async_utils.py
mbaruh: oh I thought an empty asyncio.gather raised an error, I guess not

bot/utils/async_utils.py
mbaruh: hah the underscore is because of a refactor, but yeah don't have to mention how it's implemented.

bot/utils/async_utils.py
mbaruh: good point

bot/utils/async_utils.py
```diff
@@ -57,24 +57,27 @@ def __init__(self, *, limit: int):
...
-            self._running_tasks.clear()
-            return result
-
-        return []
+        """Wait for all submitted coroutines to finish execution."""
+        result = await asyncio.gather(*self._running_tasks, return_exceptions=return_exceptions)
+        self._running_tasks.clear()
+        return result
+
+    def cancel_all(self) -> None:
```
MarkKoz, line 76: Should we wait for the tasks to be cancelled or is it fine to fire and forget?

bot/exts/moderation/clean.py
```diff
@@ -722,6 +725,10 @@ async def cog_command_error(self, ctx: Context, error: Exception) -> None:
         """Safely end the cleaning operation on unexpected errors."""
         self.cleaning = False
 
+    def cog_unload(self) -> None:
+        """Stop any ongoing clean."""
+        self.cleaning = False
```
MarkKoz, line 730: Why not cancel the tasks in here? Is setting `cleaning` to False guaranteed to cancel the tasks if any exist?

bot/exts/moderation/clean.py
```diff
@@ -471,34 +467,35 @@ async def _clean_messages(
...
-            # Means that the cleaning was canceled
+            log.trace("Cleaning was cancelled, cancelling tasks.")
             executor.cancel_all()
             return None
         await executor.gather(return_exceptions=True)  # Ignore NotFound errors etc.
 
         if not self.cleaning:
             return None
         if old_messages:
+            log.trace("Some of the found messages are older than 14d, and will deleted individually.")
```
MarkKoz, line 498: Typo: "will deleted" -> "will be deleted"

bot/exts/moderation/clean.py
mbaruh: If there's a running `_clean_messages` coro, it will stop the cleaning process, hit the check in line 489 and cancel the tasks. I can also cancel it here for good measure, but that function is the central point for handling this flag. What do you think?

bot/utils/async_utils.py
mbaruh: Do we have anything to do if we wait? Tasks that haven't started yet are cleared, and tasks that are already ongoing get cancelled. But if the task e.g. made an API call to Discord, we have no choice but let it play out. Is there something beyond cancelling it that we can realistically do? In the case of the clean cog, I think it's ok even if we let tasks that already started finish.

bot/exts/moderation/clean.py
mbaruh: I also don't have the tasks here, because I've found holding in the cog tasks specific to a command call awkward

bot/utils/async_utils.py
MarkKoz: This primarily impacts situations like bot showdown or reloading the cog. Waiting for cancellation may avoid race conditions if the bot is restarted or the cog is reloaded, but whether it's an actual risk depends on what the tasks actually do.

bot/exts/moderation/clean.py
MarkKoz: Cancelling may stop it sooner than reaching the condition you mentioned. I'm not gonna block the PR on this, but would be nice to follow up on it.

bot/utils/async_utils.py
mbaruh: By waiting for cancellation, do you mean awaiting the task after calling `.cancel`? I guess I can add a flag that will make it gather the tasks after cancelling them, or add an `async def cancel_and_wait` so that cancel can stay synchronous.

Issue #3537, opened 2026-07-06
Add support for replies to help snippets


Saber0324:
I've noticed that a decent amount of new users ignore the snippets of the bot that are responding to commands such as !learn or !code. 

A simple yet effective solution to this would make the bot reply to the same message that the caller replies, and if the command invoke isn't a reply, the bot maintains its current behavior. 

This would make it so when replying to a "how can I learn python?" with an "!learn" will draw the new users attention to the snippet thanks to the bot's replying to their message instead of just sending a message like it currently does.

Saber0324:
I would like to work in this issue myself.

jb3:
@Saber0324 Issue is approved and assigned, feel free to PR when ready.

PR #3538, opened 2026-07-06
Add support for replies to help snippets

Saber0324:
PR for issue #3537  

This implementation makes the bot reply to the *same message that the caller replies*, and if the command invoke isn't a reply, the bot maintains its current behavior, sending the snippet without replying to the caller.
If the caller replies to himself, the bot also maintains the no-reply behavior.

It is currently implemented for (all aliases of):
- !learn
- !kind
- !pep
- !src
- !code
- !res
- !gif

Due to the use of `LinePaginator.paginate` in `!rule` command, I wasn't able to implement it for all of its variants, so I left it untouched for consistency. (I'll open an issue about it if the PR gets merged)
image

### DIFF
--- a/bot/exts/info/pep.py
+++ b/bot/exts/info/pep.py
@@ -94,7 +94,7 @@ class PythonEnhancementProposals(Cog):
                 colour=Colour.red(),
             )
 
-            await send_or_reply(ctx, embed)
+        await send_or_reply(ctx, embed)
 
 async def setup(bot: Bot) -> None:
     """Load the PEP cog."""
--- a/bot/exts/info/pep.py
+++ b/bot/exts/info/pep.py
@@ -6,7 +6,7 @@ from discord.ext.commands import Cog, Context, command
 
 from bot.bot import Bot
 from bot.log import get_logger
-from bot.utils.messages import get_reference_message
+from bot.utils.messages import send_or_reply
 
 log = get_logger(__name__)
 
@@ -94,11 +94,7 @@ class PythonEnhancementProposals(Cog):
                 colour=Colour.red(),
             )
 
-        if message_reference := await get_reference_message(ctx):
-            await message_reference.reply(embed=embed)
-        else:
-            await ctx.send(embed=embed)
-
+            await send_or_reply(ctx, embed)
 
 async def setup(bot: Bot) -> None:
     """Load the PEP cog."""
--- a/bot/exts/info/pep.py
+++ b/bot/exts/info/pep.py
@@ -6,6 +6,7 @@ from discord.ext.commands import Cog, Context, command
 
 from bot.bot import Bot
 from bot.log import get_logger
+from bot.utils.messages import get_reference_message
 
 log = get_logger(__name__)
 
@@ -93,7 +94,10 @@ class PythonEnhancementProposals(Cog):
                 colour=Colour.red(),
             )
 
-        await ctx.send(embed=embed)
+        if message_reference := await get_reference_message(ctx):
+            await message_reference.reply(embed=embed)
+        else:
+            await ctx.send(embed=embed)
 
 
 async def setup(bot: Bot) -> None:
--- a/bot/exts/info/source.py
+++ b/bot/exts/info/source.py
@@ -9,6 +9,7 @@ from discord.utils import escape_markdown
 from bot.bot import Bot
 from bot.constants import URLs
 from bot.exts.info.tags import TagIdentifier
+from bot.utils.messages import get_reference_message
 
 SourceObject = commands.HelpCommand | commands.Command | commands.Cog | TagIdentifier | commands.ExtensionNotLoaded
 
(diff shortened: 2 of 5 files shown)


jb3:
For various reasons we probably don't want to add this functionality to `!rule`, so I wouldn't worry about that limitation.

It is not users responsibility to moderate others and the rule command serves as a reference rather than a tool for non-staff members.

bot/exts/info/source.py
```diff
@@ -40,12 +41,20 @@ async def source_command(
             embed = Embed(title="Bot's GitHub Repository")
             embed.add_field(name="Repository", value=f"[Go to GitHub]({URLs.github_bot_repo})")
             embed.set_thumbnail(url="https://avatars1.githubusercontent.com/u/9919")
-            await ctx.send(embed=embed)
+
+            if message_reference := await get_reference_message(ctx):
+                await message_reference.reply(embed=embed)
+            else:
+                await ctx.send(embed=embed)
```
brass75: I see this same code block a bunch of times in here - probably should refactor into a function and use that.

Saber0324: It was left this way for ease of use, not all commands implement bare `ctx.send`, some of them include it inside a `wait_for_deletion` call, this makes it more flexible, even if it's more verbose.

ShaharNaveh (dismissed):
I approve if it's worth anything

brass75 (dismissed):

bot/utils/messages.py
```diff
@@ -285,3 +285,27 @@ async def upload_log(
...
+        return None
+    try:
+        referenced_message = await ctx.fetch_message(ctx.message.reference.message_id)
+    except (discord.Forbidden, discord.NotFound):
+        referenced_message = None
+    if referenced_message is None:
+        return None
+    if referenced_message.author == ctx.author:
+        return None
+    return referenced_message
```
Sacul0457: You can use `ctx.message.reference.resolved` or `ctx.message.reference.cached_message` so that an api fetch is used only when necessary. See the d.py docs if needed.

also I think it's cleaner if it was structured this way:
```py
try:
        ...
except ...:
        return None
else:
         ...
```

bot/utils/messages.py
```diff
@@ -285,3 +285,28 @@ async def upload_log(
...
+    try:
+        referenced_message = ctx.message.reference.resolved
+    except (discord.Forbidden, discord.NotFound):
+        return None
+    else:
+        if referenced_message is None or isinstance(referenced_message, DeletedReferencedMessage):
+            return None
+        if referenced_message.author == ctx.author:
+            return None
+        return referenced_message
```
Sacul0457, line 307: sorry if wasn't very clear,
it should look something like this

```suggestion
async def get_reference_message(ctx: Context) -> Message | None:
    """Return a message reference if the reference exists and it does not refer to the author of the message."""
    if ctx.message.reference is None:
        return None
    try:
        referenced_message = .resolved or .cached_message or await fetch_message(...)
    except (discord.Forbidden, discord.NotFound):
        return None
    else:
        if isinstance(referenced_message, DeletedReferencedMessage):
            return None
        if referenced_message.author == ctx.author:
            return None
        return referenced_message
```

`NotFound` and `Forbidden` won't ever be raised unless an API call is made.
Basically the idea is to check the data in `resolved`, if empty, check `cached_message`, if also empty finally make an api call.

bot/utils/messages.py
```diff
@@ -285,3 +285,33 @@ async def upload_log(
...
+    try:
+        referenced_message = (
+        ctx.message.reference.resolved
+        or ctx.message.reference.cached_message
+        or await ctx.fetch_message(ctx.message.reference.message_id)
+        )
+    except (discord.Forbidden, discord.NotFound):
+        return None
+    else:
+        if referenced_message is None or isinstance(referenced_message, DeletedReferencedMessage):
```
Sacul0457, line 303: Yes this looks gud now, just one minor nit: `referenced_message` can never be `None` here as if `resolved` or `cached_message` is None, we will fetch using the API, which will raise `NotFound` etc. if the message can't be fetched.

bot/utils/messages.py
```diff
@@ -285,3 +285,33 @@ async def upload_log(
...
+        return referenced_message
+
+async def send_or_reply(ctx: Context, embed: Embed) -> Message:
+    """
+    Sends the bot message as a reply if the callers message has a reference.
+
+    If the callers message does not have a reference, the bot messsage is sent without reply
+    """
+    if message_reference := await get_reference_message(ctx):
+        return await message_reference.reply(embed=embed)
```
onerandomusername: This can fail if the referenced message is deleted between the fetch and send. 

This is undesirable behavior since the bot will just fail to respond in that case. The command has been invoked, a reply being deleted shouldn't cause the bot to not reply either.

One of two options:
- You'll need to construct a new reference with `fail_if_not_exists` set to False and provide that to .send()
- try, except a failure, try again without the reference.

onerandomusername (changes requested):

bot/utils/messages.py
Saber0324:
```py
    reference = None
    if message_reference := await get_reference_message(ctx):
        reference = message_reference
    return await ctx.send(embed=embed, reference=reference)
```

Would this approach work? I tested the `None` case by commenting the line that sets the `message_reference` and it's sent as a normal message. This is the simplest solution I came across.

Actually, looking at how you reconstruct references, the first approach you mentioned is better, mine would fail anyways since an exception is raised.

```py
    if message_reference := await get_reference_message(ctx):
        reference = message_reference.to_reference(fail_if_not_exists=False)
        return await ctx.send(embed=embed, reference=reference)
    return await ctx.send(embed=embed)
```

I don't like the duplication, but this avoids what you mentioned.

Saber0324: image

I confirmed that this version works, is there anything else I could upgrade? ^^

bot/exts/info/pep.py
```diff
@@ -93,8 +94,7 @@ async def pep_command(self, ctx: Context, pep_number: int) -> None:
                 colour=Colour.red(),
             )
 
-        await ctx.send(embed=embed)
-
+            await send_or_reply(ctx, embed)
```
ShaharNaveh: Was this indented inside the else on purpose?

ShaharNaveh (approved):

L3viathan (approved):

jb3 (approved):

PR #3539, opened 2026-07-06
Use log formatting over f strings

ChrisLovering:
For context, as per the first commit, this is so that sentry errors get grouped together correctly, as sentry's logging integration doesn't work nicely with f-string'ed logs

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -40,7 +40,7 @@ class Bot(BotBase):
         attempts = 0
         while True:
             try:
-                log.info(f"Attempting site connection: {attempts + 1}/{constants.URLs.connect_max_retries}")
+                log.info("Attempting site connection: %s/%s", attempts + 1, constants.URLs.connect_max_retries)
                 await self.api_client.get("healthcheck")
                 break
 
@@ -55,7 +55,7 @@ class Bot(BotBase):
         await super().setup_hook()
         await self.load_extensions(exts)
 
-    async def on_error(self, event: str, *args, **kwargs) -> None:
+    async def on_error(self, event: str, /, *args, **kwargs) -> None:
         """Log errors raised in event listeners rather than printing them to stderr."""
         e_val = exception()
 
@@ -76,4 +76,4 @@ class Bot(BotBase):
             scope.set_extra("args", args)
             scope.set_extra("kwargs", kwargs)
 
-            log.exception(f"Unhandled exception in {event}.")
+            log.exception("Unhandled exception in %s.", event)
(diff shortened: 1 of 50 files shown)


MarkKoz:
Hi Chris. I hope this message finds you well. Thanks for your continued determination to raise the bar on our code quality. I assure you it has not gone unnoticed here. As for the change itself: I cannot deny the feeling in my heart that f-strings are more readable, but this is an excellent PR and I am okay with this change. 

I've long thought about this rule. Yes, indeed, I have applied this to other repositories I work on. I thought it was a great rule that would serve us and our users well. Who could say "no" to more performance, am I right? I was praised. I was congratulated. And yet, after it was all said and done, I felt... uneasy; something was amiss. Had I... made a mistake? A profound thought struck me: am I prematurely optimising, compromising the readability I've fought tirelessly to achieve? At that moment, I had the sinking feeling that I had not only sullied the code, but also my trust and reputation. I implore you to ask yourself: are we writing the rules, or are the rules writing us?

jb3:
Mark, hi, huge fan of your work, particularly admire that compiler you wrote blindfolded in '99.

I believe the intention with this rule is less to submit to a style dictator or change readability (I do concur that the `%`-formatted log lines are considerably less legible, they make me think of an angry fruit salad), the intention is more to ensure that Sentry (the error reporting system, not the component of the military industrial complex) is able to accurately group log items as variable log fragments will be excluded from groupings.

One user on Reddit suggests: "delete your entire project and start over again"

Another user on [StackOverflow suggests](https://stackoverflow.com/a/59762507) using a custom formatter to retain `{` formatting whilst ensuring Sentry can accurately group exceptions, though I fear this isn't particularly clear still and adds additional complexity.

Team, we should establish a multi-national panel amongst the Core Developers to decide a mutually acceptable route forward here with no blockers and no friction. Please revert once anyone has suggestions or information, relying on you.

LLAP.

ChrisLovering:
Morning gents, this was done for sentry, so that with its bad eyesight it can merge issues from the same log line.

p.s. say hello to your ai agents for me.

wookie184:
This seems like it would be a cool use case of PEP 750 template strings. It wouldn't work out of the box, but maybe sentry would want to add support. I see https://github.com/getsentry/sentry-python/issues/4982.

MarkKoz (approved):
Of course, how could I have missed it? Sentry is a brilliant piece of software, and we should accommodate it.

jb3 (approved):

PR #3544, opened 2026-08-01
Add PEP slash command

### LINKED ISSUE #3543: PEP Slashcommand with autocomplete
The autocomplete allows people to search for peps by name, which is much more convenient and common.

This is how I would handle the autocomplete:
- If the cached peps is an empty, fetch it and cache it.
- Implement some fuzzy searching to find related peps

Image

But I don't think I would PR this since I can't test it, so someone can take it up if they would like too.

Sacul0457:
Notes:
- I set the limit to `10` as I think `25` would be too many and clutter the screen
- I have not tested the code
- `get_pep_embed` needs a better name...

Issue: #3543

### DIFF
--- a/bot/exts/info/pep.py
+++ b/bot/exts/info/pep.py
@@ -14,7 +14,6 @@ log = get_logger(__name__)
 ICON_URL = "https://www.python.org/static/opengraph-icon-200x200.png"
 PEP_API_URL = "https://peps.python.org/api/peps.json"
 
-
 class PEPInfo(TypedDict):
     """
     Useful subset of the PEP API response.
@@ -47,7 +46,9 @@ class PythonEnhancementProposals(Cog):
         log.trace("Started refreshing PEP data.")
         async with self.bot.http_session.get(PEP_API_URL) as resp:
             if resp.status != 200:
-                log.warning("Fetching PEP data from PEP API failed with code %s", resp.status)
+                log.warning(
+                    "Fetching PEP data from PEP API failed with code %s",
+                resp.status)
                 return
             listing = await resp.json()
 
@@ -73,8 +74,11 @@ class PythonEnhancementProposals(Cog):
         return embed
 
     async def get_pep_embed(self, pep_number: int) -> Embed:
-        if self.last_refreshed_peps is None or (
-            (self.last_refreshed_peps + timedelta(hours=1)) <= datetime.now(tz=UTC) and len(str(pep_number)) < 5
+        if (
+            self.last_refreshed_peps is None or (
+                (self.last_refreshed_peps + timedelta(hours=1)) <= datetime.now(tz=UTC)
+                and len(str(pep_number)) < 5
+            )
         ):
             await self.refresh_pep_data()
 
--- a/bot/exts/info/pep.py
+++ b/bot/exts/info/pep.py
@@ -106,7 +106,7 @@ class PythonEnhancementProposals(Cog):
 
     @app_commands.command(name="pep")
     @app_commands.guild_only()
-    @app_commands.describe(pep_number="The pep number or the autocompleted pep")
+    @app_commands.describe(pep_number="PEP number or title")
     async def pep_slash_command(self, interaction: Interaction, pep_number: int) -> None:
         """Fetches information about a PEP and sends it to the channel."""
         embed = await self.get_pep_embed(pep_number)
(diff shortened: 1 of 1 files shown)


bot/exts/info/pep.py
```diff
@@ -93,9 +91,34 @@ async def pep_command(self, ctx: Context, pep_number: int) -> None:
...
+    @app_commands.guild_only()
+    @app_commands.describe(pep_number="The pep number or the autocompleted pep")
+    async def pep_slash_command(self, interaction: Interaction, pep_number: int) -> bool:
+        embed = await self.get_pep_embed(pep_number)
+        await interaction.response.send_message(embed=embed)
+
+    @pep_slash_command.autocomplete("pep_number")
+    async def pep_slash_command_autocomplete(self, interaction: Interaction, current: str) -> list[app_commands.Choice]:
+        if self.last_refreshed_peps is None:
+            await self.refresh_pep_data()
```
ChrisLovering: This doesn't do the `(self.last_refreshed_peps + timedelta(hours=1)) <= datetime.now(tz=UTC)` check that's in `get_pep_embed`. So `self.peps` below could be out of date.

Maybe we should move this logic into a shared function now, one that checks if `self.last_refreshed_peps` is over an hour out of date, if so refresh, if not just return.

bot/exts/info/pep.py
```diff
@@ -93,9 +91,34 @@ async def pep_command(self, ctx: Context, pep_number: int) -> None:
...
+
+        choices = {pep_id: pep_info["title"] for pep_id, pep_info in self.peps.items()}
```
ChrisLovering: This could be a instance variable, updated in `self.refresh_pep_data()` so that we don't have to run this every time they type a character.

ChrisLovering (changes requested):
nice change, a few comments, also I'd suggest installing the pre-commit hooks so that your changes get linted when you commit `uv run pre-commit install`

bot/exts/info/pep.py
Sacul0457: I didn't add that guard because it would require `and len(str(pep_number)) < 5` to be true too...
And tbh I'm not sure what that does here?

bot/exts/info/pep.py
Sacul0457: Sounds gud

Sacul0457: Done

bot/exts/info/pep.py
ChrisLovering: Since there aren't any peps over 5 digits long, that check means we don't refresh peps if that pep has no chance of existing anyway.

ChrisLovering (approved):

Sacul0457:
Thanks for approving! 

Someone should probably test this to see if it works tho

bot/exts/info/pep.py
```diff
@@ -93,9 +95,33 @@ async def pep_command(self, ctx: Context, pep_number: int) -> None:
...
+    @command(name="pep", aliases=("get_pep", "p"))
+    async def pep_command(self, ctx: Context, pep_number: int) -> None:
+        """Fetches information about a PEP and sends it to the channel."""
+        # Refresh the PEP data up to every hour, as e.g. the PEP status might have changed.
+        embed = await self.get_pep_embed(pep_number)
         await send_or_reply(ctx, embed)
 
+    @app_commands.command(name="pep")
+    @app_commands.guild_only()
+    @app_commands.describe(pep_number="The pep number or the autocompleted pep")
```
jb3: Could be a better description and use the correct casing for PEP, since this is describing what the user should input (not what it is coerced to)


```suggestion
    @app_commands.describe(pep_number="PEP number or title")
```

jb3 (commented):
This works fine from local testing, just one comment and the below

One thing that would be nice to change: if a user is filling out the autocomplete and enters a number (e.g. `522`) it'd be nice to shortcut and put that PEP to the top of the suggestion list, right now it tries to fuzzy find on it which means the results don't appear in the list at all.

Image

Ideally, we either add the result to the top of the list when a match is entered here, or we return only that PEP (only fuzzy finding when there are letters or spaces in the text).

jb3 (changes requested):
Meant to request changes, oops.

Sacul0457:
Done! I've chose to implement this:
> return only that PEP (only fuzzy finding when there are letters or spaces in the text).

jb3 (approved):

PR #3549, opened 2026-08-17
Fix getting of the "bad ticks" instructions for codeblocks

### LINKED ISSUE #3547: No "wrong ticks" instruction for one or more identical code blocks with the wrong ticks
## Description

When a user have sent one or more identical code block with ticks other than backticks with or without any language specifier, the bot says to "add a `py` after the three backticks", even when the language specifier is `py`, and don't mention about the wrong ticks.

## Reasons

The PR #3517 replaced the `_RE_CODE_BLOCK` pattern to `_REGEX_CODE_BLOCK` and set a `overlapped` flag to `True` on the [line 94](https://github.com/python-discord/bot/pull/3517/changes#diff-89ba6b33c402902497f3db64cc7d61e6170bc8ece860f476100513a252f8a014R94). As a result, a new bug (#3522) occurred: when a user had used the quotes that were placed in the formatted code block, the bot anyway considered it as the wrong-formatted code block. The bug was fixed then by the PR #3527 by doing a following check:

1. If there are code blocks with bad ticks, start to compare it with the other wrong code blocks
2. Consider the first found code with the wrong ticks block when comparing with other wrong code blocks as a valid one when:
    1. They're not the same
    2. They're have different content
    3. The comparable blocks doesn't contain the content of the block with the wrong ticks

Therefore, when a user have sent one code block with the wrong ticks, the bot sets `block` to `None` and the condition leads to the block with getting of the instruction for the missing language.

### Without the language specifier

An example of the processing of a code block without the language specifier

### With the `py` language specifier

An example of the processing of a code block with the py language specifier

### With the `c` language specifier

An example of the processing of a code block without the c language specifier

### When two code blocks are identical

An example of the processing of identical code blocks

## Expected behavior

An example of the expected processing of a code block

## Possible Solution

The solution I have found is to remove the `overlapped` flag and PR #3527's check. I'm not sure about the purpose of the `overlapped` flag, but seems it breaks nothing. The only case that depends of this is the [`test_should_recognize_contained_codeblock_even_if_that_breaks_formatting` unit test](https://github.com/python-discord/bot/pull/3517/changes#diff-46160e0afd9060bc82b837d1e74c07077c35f512a3ce98cf6034b5abca70fb65R33) created in the PR #3517. When I removed the `overlapped` flag, all still worked and no issues weren't noticed. After I removed the PR #3527's check, the instruction becomes correct: the bot says about the wrong ticks and the missing language specifier.

roli2py:
Also, because RE finding isn't overlapping now, the `test_should_recognize_contained_codeblock_even_if_that_breaks_formatting` unit test is failing. Not sure it's critical or not, so I marked it as expected to fail.

Closes #3547

### DIFF
--- a/bot/exts/info/codeblock/_instructions.py
+++ b/bot/exts/info/codeblock/_instructions.py
@@ -148,23 +148,7 @@ def get_instructions(content: str) -> str | None:
         instructions = _get_no_ticks_message(content)
     else:
         log.trace("Searching results for a code block with invalid ticks.")
-        bad_ticks = [block for block in blocks if block.tick != _parsing.BACKTICK]
-        block = None
-        if bad_ticks:
-            block = next(
-                (
-                    bad_tick
-                    for bad_tick in bad_ticks
-                    if any(
-                        block
-                        for block in blocks
-                        if block != bad_tick
-                        and bad_tick.content != block.content
-                        and bad_tick.content not in block.content
-                    )
-                ),
-                None,
-            )
+        block = next((block for block in blocks if block.tick != _parsing.BACKTICK), None)
 
         if block:
             log.trace("A code block exists but has invalid ticks.")
--- a/bot/exts/info/codeblock/_instructions.py
+++ b/bot/exts/info/codeblock/_instructions.py
@@ -148,7 +148,7 @@ def get_instructions(content: str) -> str | None:
         instructions = _get_no_ticks_message(content)
     else:
         log.trace("Searching results for a code block with invalid ticks.")
-        block = next((block for block in blocks if block.tick != _parsing.BACKTICK), None)
+        block = _parsing._get_block_with_invalid_ticks(blocks)
 
         if block:
             log.trace("A code block exists but has invalid ticks.")
--- a/tests/bot/exts/info/codeblock/test_instructions.py
+++ b/tests/bot/exts/info/codeblock/test_instructions.py
@@ -1,0 +1,93 @@
+import unittest
+from textwrap import dedent
+
+from bot.exts.info.codeblock import _instructions as instructions, _parsing as parsing
+
+
+class ProvideBadTicksInstructionsTest(unittest.TestCase):
+    def __assert_is_instructions_for_message_bad_ticks_one(self, message: str) -> None:
+        code_blocks = parsing.find_faulty_code_blocks(message)
+        self.assertIsNotNone(code_blocks)
+
+        # Type narrowing
+        if code_blocks is None:
+            return
+
+        code_block = next((block for block in code_blocks if block.tick != parsing.BACKTICK), None)
+        self.assertIsNotNone(code_block)
+
+        # Type narrowing
+        if code_block is None:
+            return
+
+        expected_instructions_text = instructions._get_bad_ticks_message(code_block)
+        self.assertIsInstance(expected_instructions_text, str)
+
+        # Type narrowing
+        if not isinstance(expected_instructions_text, str):
+            return
+
+        instructions_text = instructions.get_instructions(message)
+        self.assertIsInstance(instructions_text, str)
+
+        # Type narrowing
+        if not isinstance(instructions_text, str):
+            return
+
+        self.assertEqual(instructions_text, expected_instructions_text)
+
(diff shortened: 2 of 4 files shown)


bot/exts/info/codeblock/_instructions.py
```diff
@@ -148,23 +148,7 @@ def get_instructions(content: str) -> str | None:
...
-                        block
-                        for block in blocks
-                        if block != bad_tick
-                        and bad_tick.content != block.content
-                        and bad_tick.content not in block.content
-                    )
-                ),
-                None,
-            )
+        block = next((block for block in blocks if block.tick != _parsing.BACKTICK), None)
```
MarkKoz: Doesn't reverting this regress the fix for this issue? https://github.com/python-discord/bot/issues/3522

That issue may not have been fully covered by the unit tests.

tests/bot/exts/info/codeblock/test_parsing.py
```diff
@@ -30,6 +30,7 @@ def test_should_recognize_contained_codeblock(self):
         faulty_code_blocks = parsing.find_faulty_code_blocks(message)
         self.assertIsNone(faulty_code_blocks)
 
+    @unittest.expectedFailure
     def test_should_recognize_contained_codeblock_even_if_that_breaks_formatting(self):
```
MarkKoz, line 35: What would be the faulty code block here? I think this test shouldn't e an expected failure. Though if this is an edge case that's hard to fix, then I'm fine with letting this be a failure.

tests/bot/exts/info/codeblock/test_instructions.py
```diff
@@ -0,0 +1,93 @@
+import unittest
+from textwrap import dedent
+
+from bot.exts.info.codeblock import _instructions as instructions, _parsing as parsing
+
+
+class ProvideBadTicksInstructionsTest(unittest.TestCase):
+    def __assert_is_instructions_for_message_bad_ticks_one(self, message: str) -> None:
```
MarkKoz: Use a single underscore as a prefix, not a double underscore.

MarkKoz: This scope of this test is too large.

- Don't call `find_faulty_code_blocks` because that should be tested separately. Instead, construct code block objects yourself and provide them as inputs
- Don't do `next((block for block in code_blocks if block.tick != parsing.BACKTICK), None)` because this is copying what's in the implementation.
- Don't rely on calling a private method `_get_bad_ticks_message`. Find some heuristic to use when asserting the instructions text rather than doing an exact match.

tests/bot/exts/info/codeblock/test_instructions.py
```diff
@@ -0,0 +1,93 @@
...
+        instructions_text = instructions.get_instructions(message)
+        self.assertIsInstance(instructions_text, str)
+
+        # Type narrowing
+        if not isinstance(instructions_text, str):
+            return
+
+        self.assertEqual(instructions_text, expected_instructions_text)
+
+    def test_should_provide_when_no_lang_spec_and_bad_ticks_are_used(self) -> None:
```
MarkKoz: If find these test names confusing. What does "should provide" mean? Provide what?

MarkKoz (changes requested):

tests/bot/exts/info/codeblock/test_instructions.py
roli2py: Why? This method is test-case-specific and I think we can just leave with double underscore. As I understand, a single underscore is used when we need to use objects internally, for example, within the module and not only in the class.

UPD: rephrased the last sentence

tests/bot/exts/info/codeblock/test_parsing.py
roli2py: I think it's an edge case, but I don't really know. I tried to figure out how to fix it, but it seems adding a lot of code and, therefore, an overwhelm.

tests/bot/exts/info/codeblock/test_instructions.py
roli2py: I'll fix it

PR #3550, opened 2026-08-20
Fix !pypi replies

### LINKED ISSUE #3542: Replying to a post with `!pypi` does not cause the bot to reply to the replied post.
## Description
If you invoke `!pypi` as a response to a message it should trigger the bot to reply to the original message as it does with `!learn`, `!resources`, and others. (I understand that `!rules` was explicitly not included in this but AFAIK that was the only explicit exclusion.)

## Steps to Reproduce
Reply to a message while invoking  the `!pypi` command.

## Expected Behaviour
The message that was replied to should also be replied to by the bot with the response to the `!pypi` command.

## Actual Behaviour
The bot does not reply to the original message and the output of `!pypi` is a standalone message.

## Known Impacted Platforms

- [x] Web
- [x] Desktop
- [x] Android App
- [x] iOS App

## Possible Solutions
Port the functionality from other, similar, commands to `!pypi`

## Would you like to implement a fix?

***Note: For high-priority or critical bugs, fixes may be implemented by staff.***

- [ ] I'd like to implement the bug fix
- [X] Anyone can implement the bug fix

Kaztral-ar:
Fixes an issue where the !pypi command does not reply to the message that the user is replying to.

This changes the command to use the same reply behavior as other commands like !learn and !resources.

Fixes #3542

### DIFF
--- a/bot/exts/info/pypi.py
+++ b/bot/exts/info/pypi.py
@@ -13,7 +13,7 @@ from bot.bot import Bot
 from bot.constants import Colours, NEGATIVE_REPLIES, RedirectOutput
 from bot.log import get_logger
 from bot.utils import time
-from bot.utils.messages import wait_for_deletion
+from bot.utils.messages import send_or_reply, wait_for_deletion
 
 URL = "https://pypi.org/pypi/{package}/json"
 PYPI_ICON = "https://cdn.discordapp.com/emojis/766274397257334814.png"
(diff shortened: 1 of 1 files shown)


jb3:
Is this something we see people trying to do?

Worth noting the original issue was not yet approved because I don't think we figured out what the criteria should be for enabling this.

I'm happy to merge this but in future please check issues have the approved label to save yourself work if we decide not to accept a contribution, and also I don't want to set a precedent that every command must support replies unless we see it being used or people think it will genuinely be useful.

jb3 (approved):

PR #3552, opened 2026-08-20
Improve resource topic title. 

### LINKED ISSUE #2510: Display the resource topic when using the `!resources` command
The `!resources` command accepts an optional argument to specify the resource topic in the URL. For example, `!resources data science` will result in the following link: https://www.pythondiscord.com/resources/?topics=data-science

However, there is no difference in the embed (other than the hyperlink itself) to indicate that a topic has been specified. I propose that when a topic is specified, it will be added to the title of the embed, like "**Resources for: Data science**".

(I would like to implement this myself)

Kaztral-ar:
When a topic is provided with the !resources command, show it in the embed title.

For example:

!resources games

will now show:

Resources for: games

This makes it clearer which resources the command is showing.

Fixes #2510

### DIFF
--- a/bot/exts/info/pypi.py
+++ b/bot/exts/info/pypi.py
@@ -13,7 +13,7 @@ from bot.bot import Bot
 from bot.constants import Colours, NEGATIVE_REPLIES, RedirectOutput
 from bot.log import get_logger
 from bot.utils import time
-from bot.utils.messages import wait_for_deletion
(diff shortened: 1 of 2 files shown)


jb3 (changes requested):
Can we title case the topic so it's a bit cleaner?

Kaztral-ar (commented):
Title case resource topic

jb3 (approved):

